PR #575, opened 2019-03-19
Add project header background image

### LINKED ISSUE #556: Add ProjectHeader background
**Package** 

app-project

ProjectHeader should use the project background if available, falling back to teal if not.

cc #44

rogerhutchings:
Package:

app-project

Closes #556.

Adds the project background image to the header component.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/src/components/ProjectHeader/components/Background/Background.js
+++ b/packages/app-project/src/components/ProjectHeader/components/Background/Background.js
@@ -1,5 +1,5 @@
-import PropTypes from 'prop-types'
-import { string } from 'react'
+import { string } from 'prop-types'
+import React from 'react'
 import { Box } from 'grommet'
 import styled from 'styled-components'
 
--- a/packages/app-project/src/components/ProjectHeader/components/Background/Background.js
+++ b/packages/app-project/src/components/ProjectHeader/components/Background/Background.js
@@ -3,32 +3,32 @@ import React from 'react'
 import { Box } from 'grommet'
 import styled from 'styled-components'
 
-const StyledBox = styled(Box)`
-  height: 100%;
-  overflow: hidden;
+const BackgroundBox = styled(Box)`
   position: absolute;
-  width: 100%;
   z-index: -1;
 `
 
-const StyledImg = styled.img`
-  filter: blur(4px);
-  object-fit: cover;
-  opacity: 0.5;
+const BackgroundImage = styled.div`
+  background-blend-mode: multiply;
+  background-color: rgba(0,93,105,0.3);
+  background-image: url("${props => props.src}");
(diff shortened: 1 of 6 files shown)


packages/app-project/src/components/ProjectHeader/components/Background/Background.js
```diff
@@ -0,0 +1,38 @@
+import PropTypes from 'prop-types'
+import { string } from 'react'
```
mcbouslog: first two lines I think mixed up, just need `import { string } from 'prop-types'`?

packages/app-project/src/components/ProjectHeader/components/Background/Background.js
```diff
@@ -0,0 +1,38 @@
...
+  position: absolute;
+  width: 100%;
+  z-index: -1;
+`
+
+const StyledImg = styled.img`
+  filter: blur(4px);
+  object-fit: cover;
+  opacity: 0.5;
+`
```
mcbouslog: This doesn't feel right on an `img` tag, I thought `object-fit: cover` would do this, but adding `width: 100%` stretched the background image out similar to PFE's current style.

packages/app-project/src/components/ProjectHeader/components/Background/Background.js
rogerhutchings: 🤦‍♂️

packages/app-project/src/components/ProjectHeader/components/Background/Background.js
eatyourgreens: Why not use a background image, and `background-size: cover` like we do on zooniverse.org? I'm not understanding why the background has been changed to use a content image.

rogerhutchings: Because I was writing this after a hospital visit with Jake, and all I could remember at the time was that you can't apply CSS filters directly to background images. I'll copy over the existing PFE stuff

eatyourgreens: :+1:

packages/app-project/src/components/ProjectHeader/components/Background/BackgroundContainer.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+}
+
+@inject(storeMapper)
+@observer
+class BackgroundContainer extends Component {
+  render () {
+    return (
+      <Background
+        aria-hidden={true}
```
eatyourgreens: Minor point but I think `aria-hidden` is redundant on an empty div.

PR #549, opened 2019-03-15
Add feedback modal

mcbouslog:
Package: lib-classifier

Towards #205, after #531 and #540 

Describe your changes:
- add feedback modal

https://github.com/mobxjs/mobx-state-tree/blob/master/docs/middleware.md
https://github.com/mobxjs/mobx-state-tree/blob/master/docs/async-actions.md

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?
- [ ] Is the changelog updated?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?

PR #585, opened 2019-03-21
Refactor LCV annotation to use a tools array

srallen:
Package: lib-classifier

Describe your changes:
I was looking over what the aggregation code expects for extraction and noticed that we aren't 100% conforming to the expectation. The shape extractor _can_ be reused for the LCV annotation, however, it's also expecting a tool annotation so it knows which shape the annotation is and thus which extractor to use. (see: https://github.com/zooniverse/aggregation-for-caesar/blob/master/panoptes_aggregation/extractors/tool_wrapper.py. The is a decorator for the shape extractor: https://github.com/zooniverse/aggregation-for-caesar/blob/master/panoptes_aggregation/extractors/shape_extractor.py)

This is a refactor on the structure of the workflow task for the LCV to use a tools array similar to the drawing task. I've renamed it to `DataVisAnnotationTask` and then moved the specific task into a tool, `Graph2dRangeXTool`. The annotation for this task and tool now also includes the array index of the selected tool as well as its string type. The string type is not necessary for the aggregation, but for resolving the types in the MST tree for validation (aggregation can just ignore this or use it if they wish). 

Long term, this is also a good idea to begin decoupling the task from the LCV. The upcoming Variable Star LCV will not likely be using the same task as TESS. Some of the function now hard coded into the viewer component can be extracted out into their own set of helper functions and/or into the new `DataVisAnnotatingStore` (which is analogous in purpose to the `DrawingStore`)

Additionally, I think it's a good idea to have consistent UI so this task now looks similar to the drawing task in the classifier UI with a set of radio buttons to select which tool to use for the annotation. This also partially preps for adding the drawing tool task because we now have a similar working variant of it. 

Unfortunately the original LCV code does not have robust tests, but I've updated and added tests where I could. 

To test it, I did update the workflow selection code to allow for selecting by query param in the most basic way: http://localhost:8080/?project=1862&workflow=3316 (we're also missing good tests for the workflow selection at the moment)

So you should be able to load up the workflow 3316 on the staging project 1862 that uses a task that is structured the way this branch expects. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/store/DataVisAnnotatingStore.js
```diff
@@ -0,0 +1,37 @@
...
+
+  .actions(self => {
+    function afterAttach() {
+      createClassificationObserver()
+    }
+
+    function createClassificationObserver() {
+      const classificationDisposer = autorun(() => {
+        onAction(getRoot(self), (call) => {
+          if (call.name === 'completeClassification') self.reset()
```
srallen: I just realized that this should probably reset between steps, not just at the end of the classification.

eatyourgreens: Is there a reason that this listens to the whole tree, and not just the classifications store? Elsewhere, I've been adding listeners to a particular node rather than root.

srallen: Nope. This bit needs improvement.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DataVisAnnotationTask/DataVisAnnotationTask.js
```diff
@@ -0,0 +1,96 @@
...
+import en from './locales/en'
+counterpart.registerTranslations('en', en)
+
+export const StyledFieldset = styled.fieldset`
+  border: none;
+`
+
+function storeMapper (stores) {
+  const annotations = stores.classifierStore.classifications.currentAnnotations
+  const { active, setActive } = stores.classifierStore.dataVisAnnotating
```
eatyourgreens: Not a blocking comment but should task stores be children of the classifications store, or maybe attached to a generic tasks store rather than root? I can see our flat store structure getting hard to manage as more tasks are added and it grows outwards.

srallen: Yeah I can see us breaking that up like I did with the WorkflowStore and WorkflowStepStore. Annotations could be its own store handling the add/remove annotation calls.

srallen:
Also, I can try to break this PR up if requested. I know it's pretty big.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DataVisAnnotationTask/DataVisAnnotationTask.js
eatyourgreens: I think I'd go for `classifierStore.tasks.dataVisAnnotating`. Similarly, I think workflow steps could be `classifierStore.workflow.steps` which would give the workflow store access to actions on the workflowStep store. At the moment, the structure's completely flat, and I'm thinking that loses us some of the advantages of the state tree.

Not a high priority, just something I've been thinking about since working with feedback and looking at how the different models interact.

srallen: Could you open an RFC issue about it? It could be an agenda item to discuss for the week I'm in Oxford

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DataVisAnnotationTask/DataVisAnnotationTask.js
```diff
@@ -0,0 +1,96 @@
...
+        
+        {task.tools.map((tool, index) => {
+          const checked =  active === index
+          return (
+            <TaskInputField
+              checked={checked}
+              index={index}
+              key={`${task.taskKey}_${index}`}
+              label={tool.label}
+              labelIcon={<InputIcon icon={<Graph2dRangeXIcon />} />}
```
mcbouslog: Getting `React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: object.` related to one of these components.

packages/lib-classifier/src/store/DataVisAnnotatingStore.js
```diff
@@ -0,0 +1,37 @@
+import { types } from 'mobx-state-tree'
```
mcbouslog: I think missing a few imports?
```
import { addDisposer, getRoot, onAction, types } from 'mobx-state-tree'
```

packages/lib-classifier/src/store/DataVisAnnotatingStore.js
```diff
@@ -0,0 +1,37 @@
...
+
+const DataVisAnnotatingStore = types
+  .model('DataVisAnnotatingStore', {
+    active: types.optional(types.number, 0),
```
eatyourgreens: Having a model for a single number feels heavy-handed. Is the general idea that each task will have a store to hold the state of the form for that task?

srallen: I'm expecting some of the functions that are in the `LightCurveViewer` react component will move to this store when that is refactored to decouple the task from the viewer.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DataVisAnnotationTask/DataVisAnnotationTask.js
srallen: I might have screwed up a renaming of something

packages/lib-classifier/src/store/DataVisAnnotatingStore.js
srallen: MST warns about doing this... but our use case seems ok?

packages/lib-classifier/src/store/DataVisAnnotatingStore.js
srallen: Also, I think we do want to store the annotation form data in MST stores because then it validates against the model. We are doing this for the single and multiple choice task type annotations, too.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DataVisAnnotationTask/DataVisAnnotationTask.js
srallen: I've opened #603 for this.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DataVisAnnotationTask/DataVisAnnotationTask.js
srallen: I've fixed this and also the export of InputIcon.

packages/lib-classifier/src/store/DataVisAnnotatingStore.js
srallen: Thanks. I have no idea how this got messed up in my last commit, but they had to be there at one point for this to be working when I first opened the PR. 🤷‍♀️

srallen:
The coveralls coverage decreased because of the addition of the disposers and autoruns for listening to the actions in the DataVisAnnotatingStore. I'm not sure how to test those parts of MST yet... Perhaps we can circle back to that?

eatyourgreens:
In #549 I exported the lifecycle action `onSubjectAdvance` just so it could be tested, but I’m not a massive fan of that approach.

PR #587, opened 2019-03-21
Make project title a link when not on the project home page

### LINKED ISSUE #567: Avatar and project name should link back to project home
**Package**

app-project

The avatar and project title should serve as a link back to the project home page.

rogerhutchings:
Package: app-project

Closes #567.

Makes the project title a link back to the project home when on another page. Copies PFE, in that only the title and not the avatar work as a link.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/src/components/ProjectHeader/ProjectHeader.js
+++ b/packages/app-project/src/components/ProjectHeader/ProjectHeader.js
@@ -1,5 +1,5 @@
 import counterpart from 'counterpart'
-import { Box, Heading } from 'grommet'
+import { Anchor, Box, Heading } from 'grommet'
 import { string } from 'prop-types'
 import React from 'react'
 import styled from 'styled-components'
@@ -20,8 +20,24 @@ const StyledHeading = styled(Heading)`
   text-shadow: 0 2px 4px rgba(0,0,0,0.5);
 `
 
+const StyledAnchor = styled(Anchor)`
+  border-bottom: 3px solid transparent;
+  &:focus,
+  &:hover {
+    text-decoration: none;
+    border-color: white;
+  }
+`
+
 function ProjectHeader (props) {
-  const { className, title } = props
+  const { className, href, title } = props
+
+  const Title = () => (
+    <StyledHeading color='white' margin='none' size='small'>
+      {title}
+    </StyledHeading>
+  )
+
   return (
     <StyledBox>
       <Background />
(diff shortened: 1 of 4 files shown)


packages/app-project/src/components/ProjectHeader/ProjectHeaderContainer.spec.js
```diff
@@ -27,4 +39,19 @@ describe('Component > ProjectHeaderContainer', function () {
   it('should pass down the project title', function () {
     expect(componentWrapper.prop('title')).to.equal(PROJECT.display_name)
   })
+
+  it('should pass down the project homepage href if on another page', function () {
+    expect(componentWrapper.prop('href')).to.equal(`/projects/${OWNER}/${PROJECT_SLUG}`)
```
srallen: At first read this was confusing because of what project slug means from the API, `:owner/:display_name`... so same comment about renaming the variable here.

packages/app-project/src/components/ProjectHeader/ProjectHeaderContainer.spec.js
```diff
@@ -9,11 +9,23 @@ let wrapper
 let componentWrapper
 
 const PROJECT = projects.mocks.resources.projectOne
+const OWNER = 'foo'
+const PROJECT_SLUG = 'bar'
```
srallen: Perhaps we shouldn't be calling this the slug, because that's the concatenation of the owner and display_name serialized from the API for us on the project resource. This could be confusing. Maybe `PROJECT_DISPLAY_NAME`?

packages/app-project/src/components/ProjectHeader/ProjectHeader.js
```diff
@@ -20,8 +20,24 @@ const StyledHeading = styled(Heading)`
   text-shadow: 0 2px 4px rgba(0,0,0,0.5);
 `
 
+const StyledAnchor = styled(Anchor)`
```
srallen: Just an aside, but maybe this is a candidate for moving into the shared components lib since the standard in the design seems to be using borders instead of text-decoration.

packages/app-project/src/components/ProjectHeader/ProjectHeader.js
```diff
@@ -20,8 +20,24 @@ const StyledHeading = styled(Heading)`
...
+  const { className, href, title } = props
+
+  const Title = () => (
+    <StyledHeading color='white' margin='none' size='small'>
+      {title}
+    </StyledHeading>
+  )
+
   return (
     <StyledBox>
```
srallen: `className` needs to be set passed on to the outer component here to avoid class mismatch errors.

Issue #588, opened 2019-03-21
SubjectId prop to CollectionsModalContainer is undefined


rogerhutchings:
**Package**

app-project

**Feature or Issue Description**

- This is firing off complaint a warning that `subjectId` is a required prop, but is undefined. Logging it while opening / closing the modal and general futzing around with the classifier shows that it's always undefined. Could it be left over from a previous iteration?

eatyourgreens:
Yep. A quick look at the code shows it’s stored in component state now. https://github.com/zooniverse/front-end-monorepo/blob/master/packages/app-project/src/screens/ClassifyPage/components/CollectionsModal/CollectionsModalContainer.js

rogerhutchings:
Cool, that's what I thought - just wanted to check.

PR #590, opened 2019-03-21
Fix warning about required prop in CollectionsModalContainer

rogerhutchings:
Package: app-project

Fixes the console warning about a missing required prop, and adds propTypes for the props that are still being used. 

Note that I've used `isRequired` _and_ provided a default prop; I think the store hasn't finished instantiating itself by the time this component is mounted, so this avoids further warnings. We've avoided doing this in the past, but I think this will need to be done throughout the app, as these props are strictly speaking required by the component for it to work.


rogerhutchings:
fixes #588

PR #584, opened 2019-03-21
Add ADR for production docker setup

rogerhutchings:
Adds an ADR for the current production build setup, and updates the README.

PR #595, opened 2019-03-21
Bump to node 10

rogerhutchings:
Package: all

Following on from https://github.com/zooniverse/front-end-monorepo/pull/584#discussion_r267945685, this bumps the required node version to 10.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-async-states/package.json
+++ b/packages/lib-async-states/package.json
@@ -30,6 +30,9 @@
     "snazzy": "~8.0.0",
     "standard": "~12.0.1"
   },
+  "engines": {
+    "node": ">=0.12.18"
+  },
   "publishConfig": {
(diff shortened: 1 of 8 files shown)


packages/lib-async-states/package.json
```diff
@@ -30,6 +30,9 @@
     "snazzy": "~8.0.0",
     "standard": "~12.0.1"
   },
+  "engines": {
+    "node": ">=0.12.18"
```
eatyourgreens: node 0.12?

rogerhutchings: That's all it needs, this library is tiny enough that I could manually check exactly what's required on node.green

PR #552, opened 2019-03-15
Tutorial modal

srallen:
Package: lib-classifier, lib-react-components

Describe your changes:
Adds the modal UI component, with some refactors to the Modal component in lib-react-components. The styles look broken, but I anticipate will be fixed after #551. Still to do:

- [x] ~The actual PUT request to update the UPP~ (in #572)
- [x] ~Figure out why roughly 1 out 5 times of loading the page the classifier crashes because MST can't resolve the active medium~ (Fixed in #600)
- [x] Update/write tests as needed

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?
- [ ] Is the changelog updated?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?

PR #600, opened 2019-03-22
Wait for the media fetch

srallen:
Package: lib-classifier

Describe your changes:
Branched off of #552. This fixes an issue where sometimes the reference to the media would not resolve because the request for it hadn't finished before the tutorial modal opened up. We now wait for the request for the media to finish before setting the tutorial and calling to show the modal. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/TutorialStore.js
+++ b/packages/lib-classifier/src/store/TutorialStore.js
@@ -111,7 +111,6 @@ const TutorialStore = types
     }
 
     function * fetchMedia (tutorial) {
-      const upp = getRoot(self).userProjectPreferences
       const { tutorials } = getRoot(self).client
       if (tutorial) {
         self.loadingState = asyncStates.loading
@@ -119,10 +118,6 @@ const TutorialStore = types
           const response = yield tutorials.getAttachedImages({ id: tutorial.id })
           const { media } = response.body
           if (media && media.length > 0) self.setMediaResources(media)
-          self.loadingState = asyncStates.success
-          if (upp.loadingState === asyncStates.success) {
-            self.showTutorialInModal()
-          }
         } catch (error) {
           // We're not setting the store state to error because
           // we do not want to prevent the tutorial from rendering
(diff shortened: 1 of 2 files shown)


packages/lib-classifier/src/store/TutorialStore.js
```diff
@@ -140,14 +135,19 @@ const TutorialStore = types
...
       const tutorialsClient = getRoot(self).client.tutorials
+      const upp = getRoot(self).userProjectPreferences
 
       self.loadingState = asyncStates.loading
       try {
         const response = yield tutorialsClient.get({ workflowId: workflow.id })
         const { tutorials } = response.body
         if (tutorials && tutorials.length > 0) {
-          tutorials.forEach(tutorial => self.fetchMedia(tutorial))
+          yield Promise.all(tutorials.map(tutorial => self.fetchMedia(tutorial)))
```
eatyourgreens: I don't think `fetchMedia` is called anywhere else in the code so it can be a private method.
```suggestion
          yield Promise.all(tutorials.map(tutorial => flow(fetchMedia)(tutorial)))
```

eatyourgreens:
The code works when I test this in a browser, so it looks good. However, the tests fail if `fetchMedia` isn't stubbed, which makes me worried.

srallen:
They were failing because the panoptes client isn't being stubbed to return a mocked medium resource for those get requests for those tests. We now have a `yield` statement on those requests and it'll wait for a promise to resolve or reject. If we aren't returning a promise from `fetchMedia` then the test will fail. It was easier to just stub `fetchMedia` there to just return a promise to get the tests passing then to make what I think is a hard to follow mocking of the client even more complicated. `fetchMedia` itself is being tested too, so imo, we aren't missing test coverage here.

eatyourgreens:
Weird, because I was seeing a media resource coming back from the mocked client, but tests were failing anyway.

srallen:
It's being mocked in a different describe block I think, but I could be wrong about that... this is what I mean, it's confusing/hard to follow how the panoptes client is being mocked.

eatyourgreens:
Yeah. This makes me nervous because `fetchMedia` is only being exported in order to pass the tests, and I'm not 100% sure that the tests are actually testing the code the way that we think.

My thinking was that as long as `tutorials.getAttachedImages()` is stubbed in the tests, then they should still work. As far as I can tell, it is stubbed and resolves with a mock image, but the tests fail if we actually run that stub via the `fetchMedia` call.

eatyourgreens:
Nevermind, I completely missed that `Promise.all` isn't appropriate here. I'll open a PR.

eatyourgreens:
I can't figure out why `Promise.all` breaks the tests, unless `fetchMedia` is stubbed with `Promise.resolve()`, but the behaviour I'm seeing is that the expectations run before `setTutorial` has run, or before `showTutorialInModal` has run. Look at #601 for an example.

srallen:
I looked it over and the last two tests for the describe block for `Actions > showTutorialInModal` didn't actually need fetchMedia's stub calling a fake function to return the promise. It is being stubbed correctly in the client for those two tests. It's the first test that didn't and needs the `callsFake` because `getAttachedImages` is not being stubbed.

srallen:
I think it's the confusing way I'm mocking panoptes.js. I'd like to revisit that and get a standard way that this done and a part of panoptes.js itself.

eatyourgreens:
Tests still pass without this change, so there's a test missing to catch the original bug.

PR #601, opened 2019-03-25
Remove self.fetchMedia from tutorial store

eatyourgreens:
Package:
lib-classifier

Remove the public `fetchMedia` action. It was only being exported so that it could be stubbed in tests.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/Classifier.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.js
@@ -14,6 +14,7 @@ import {
 import { registerWorkers, unregisterWorkers } from '../../workers'
 import RootStore from 'src/store'
 import Layout from './components/Layout'
+import ModalTutorial from './components/ModalTutorial'
 import { isBackgroundSyncAvailable } from '../../helpers/featureDetection'
 
 const client = {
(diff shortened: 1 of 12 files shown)


eatyourgreens:
`fetchMedia` tests still pass, so I think this works.

Previously, I think I was seeing different results between the code and the tests because the stubbed `fetchMedia` method returned a Promise, but the actual `fetchMedia` method is a generator.

srallen:
The generators wrapped by flow always return promises according to MST docs: https://github.com/mobxjs/mobx-state-tree#asynchronous-actions

eatyourgreens:
Weird, so why are the tests failing with `Promise.all`?

srallen:
I cleaned up the stubs a little bit (see: https://github.com/zooniverse/front-end-monorepo/pull/600#issuecomment-476298333). The test for `it should show the tutorial for logged out users` still needs `fetchMedia`'s stub to call a fake because the panoptes client stub for that test doesn't stub `getAttachedImages`. Does that clarify it at all? Could this update close? Do you still want at least the update to not make `fetchMedia` public?

srallen:
It looks like there's some additional changes coming in with this that shouldn't be, too. I went ahead and merged #600 so if there are any more changes you'd like, please comment on #552. Thanks.

PR #606, opened 2019-03-26
Remove self.fetchMedia from tutorial store

eatyourgreens:
Package:
lib-classifier

Removes `self.fetchMedia` from the tutorial store, since it isn't used outside the store. Can anyone see why doing this breaks the tests?

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/TutorialStore.js
+++ b/packages/lib-classifier/src/store/TutorialStore.js
@@ -142,7 +142,7 @@ const TutorialStore = types
         const response = yield tutorialsClient.get({ workflowId: workflow.id })
         const { tutorials } = response.body
         if (tutorials && tutorials.length > 0) {
-          const mediaRequests = tutorials.map(tutorial => fetchMedia(tutorial))
+          const mediaRequests = tutorials.map(fetchMedia)
           yield Promise.all(mediaRequests)
           self.setTutorials(tutorials)
           self.loadingState = asyncStates.success
--- a/packages/lib-classifier/src/store/TutorialStore.js
+++ b/packages/lib-classifier/src/store/TutorialStore.js
@@ -224,7 +224,6 @@ const TutorialStore = types
 
     return {
       afterAttach,
-      fetchMedia: flow(fetchMedia),
       fetchTutorials: flow(fetchTutorials),
       resetActiveTutorial,
       resetSeen,
--- a/packages/lib-classifier/src/store/TutorialStore.js
+++ b/packages/lib-classifier/src/store/TutorialStore.js
@@ -142,7 +142,8 @@ const TutorialStore = types
         const response = yield tutorialsClient.get({ workflowId: workflow.id })
         const { tutorials } = response.body
         if (tutorials && tutorials.length > 0) {
-          yield Promise.all(tutorials.map(tutorial => self.fetchMedia(tutorial)))
+          const mediaRequests = tutorials.map(tutorial => flow(fetchMedia)(tutorial))
+          yield Promise.all(mediaRequests)
           self.setTutorials(tutorials)
           self.loadingState = asyncStates.success
           if (upp.loadingState === asyncStates.success) {
--- a/packages/lib-classifier/src/store/TutorialStore.js
+++ b/packages/lib-classifier/src/store/TutorialStore.js
@@ -110,7 +110,7 @@ const TutorialStore = types
       addDisposer(self, uppDisposer)
     }
 
-    function * fetchMedia (tutorial) {
+    const fetchMedia = flow(function * fetchMedia (tutorial) {
       const { tutorials } = getRoot(self).client
       if (tutorial) {
         self.loadingState = asyncStates.loading
@@ -124,7 +124,7 @@ const TutorialStore = types
           console.error(error)
         }
       }
-    }
+    })
 
     function setMediaResources (media) {
       media.forEach(medium => self.attachedMedia.put(medium))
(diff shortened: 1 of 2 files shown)


packages/lib-classifier/src/store/TutorialStore.js
```diff
@@ -142,7 +142,8 @@ const TutorialStore = types
         const response = yield tutorialsClient.get({ workflowId: workflow.id })
         const { tutorials } = response.body
         if (tutorials && tutorials.length > 0) {
-          yield Promise.all(tutorials.map(tutorial => self.fetchMedia(tutorial)))
+          const mediaRequests = tutorials.map(tutorial => flow(fetchMedia)(tutorial))
+          yield Promise.all(mediaRequests)
```
eatyourgreens: Test expectations are being evaluated here, before the tutorial is set and `showTutorialInModal` is called. The workaround seems to be to stub `fetchMedia` to return `Promise.resolve()` and I can't see why that's needed to pass the tests.

rogerhutchings: via https://github.com/mobxjs/mobx-state-tree/blob/master/docs/async-actions.md#using-generators

> `Yield` always needs to return a `Promise`.

...but there are several states in which `fetchMedia` doesn't (e.g. no tutorial), unless it's always stubbed to return a resolved promise 💡

eatyourgreens: But `tutorials.getAttachedImages` is stubbed to return a resolved promise in the tests. That's what's confusing me. https://github.com/zooniverse/front-end-monorepo/blob/ebd450a09dd5540ff62f8771d1b5409b70acba1a/packages/lib-classifier/src/store/TutorialStore.spec.js#L73-L79

eatyourgreens: I'm seeing that resolve with the mock media, then the expectation being evaluated, then `setTutorial` being called.

rogerhutchings: Hmmm. I've just tried running `it('should request for the media if there are tutorials')` in isolation with `.only`, and it passed; could the test state be dirty?

eatyourgreens: That's not one of the failing tests.

Try this one:
```
Model > TutorialStore
       should set the tutorial if there is a workflow:
     TypeError: Cannot read property 'toJSON' of undefined
```

eatyourgreens: I think the failing tests are all ones that evaluate the outcome of `setTutorials()` or `showTutorialInModal()`, because those calls aren't waiting for done. 🤔 

I also think this is worrying me because the code change is harmless (open http://localhost:8080/?project=1233 in a browser and it works just fine) and working code shouldn't break the tests.

eatyourgreens: Hang on, all the failing tests run when `rootStore.workflows.setActive()` resolves. Am I right in thinking that they should be waiting for `rootStore.tutorials.fetchTutorials()` to resolve, which happens slightly later? That might be the problem.

rogerhutchings: I was going to say something like that! I've just picked it up again (sorry, am still feeling pretty rough), and looking at `it('should set the tutorial if there is a workflow')` - it's asserting that once a workflow is loaded, a tutorial _is also_ loaded, even though fetching a tutorial is happening in response to workflow loading

eatyourgreens: I think those calls should be `rootStore.workflows.setActive(workflow).then(rootStore.tutorials.fetchTutorials).then…`.

rogerhutchings: the other approach would be to create an observer on `tutorials.active`, look for an update where the tutorial id matches the expected one, and then call `done`; which might be preferable since you're not testing the exact implementation, just that a change happens and the expected thing happens in response

eatyourgreens: No worries! Thanks very much for taking a look at this. It's probably a common case in the classifier tests. A promise resolves, triggering an observer that runs another promise and we don't want to test until the second promise has resolved.

Hope you feel better soon!

eatyourgreens: Ha! Making my change doesn't fix all the tests, because spies are reset asynchronously, so tests for `calledOnce` start to break.

srallen:
Thanks for tackling these tests. It'd be good to know what I wasn't doing quite right here.

packages/lib-classifier/src/store/TutorialStore.js
eatyourgreens: My dumb solution, in the end, was to stub `fetchTutorials`, run `workflow.setActive` with the stub, then restore it and run `fetchTutorials` by hand. That stops the tests from running until after we've fetched tutorials, and preserves the spy history.

eatyourgreens:
@srallen the problem seems to be that the tests were set to run when the active workflow was set, but that doesn't wait for the tutorials and media to be fetched, unless we stub them to resolve instantly.

Separate issue: if I remove the `Promise.all`, I reintroduce the unresolved medium bug but I think the tests still pass. So there could be a missing test for that bug.

eatyourgreens:
This should be passing tests and ready for review now.

eatyourgreens:
I think I caught all the ones that need it to be stubbed when the workflow is set, but I’m not sure.

eatyourgreens:
@srallen I've gone through and stubbed `fetchTutorials` so that it's only ever invoked once for each test.

PR #610, opened 2019-03-26
Test tutorial media errors

eatyourgreens:
Package:
lib-classifier

This builds off #606 because I made changes to the tutorial tests there. It mocks the tutorial client to reject with an error on `tutorials.getAttachedImages()`, and expects the tutorial not to be shown in that case.

It also tests the case where media requests resolve after tutorial requests, to make sure that we aren't showing the tutorial before it's ready.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/TutorialStore.js
+++ b/packages/lib-classifier/src/store/TutorialStore.js
@@ -122,7 +122,6 @@ const TutorialStore = types
           // We're not setting the store state to error because
           // we do not want to prevent the tutorial from rendering
           console.error(error)
-          self.setMediaResources([])
         }
       }
     })
--- a/packages/lib-classifier/src/store/TutorialStore.js
+++ b/packages/lib-classifier/src/store/TutorialStore.js
@@ -122,6 +122,7 @@ const TutorialStore = types
           // We're not setting the store state to error because
           // we do not want to prevent the tutorial from rendering
           console.error(error)
+          self.setMediaResources([])
         }
       }
     })
--- a/packages/lib-classifier/src/store/TutorialStore.spec.js
+++ b/packages/lib-classifier/src/store/TutorialStore.spec.js
@@ -91,8 +91,7 @@ const authClientStubWithUser = {
   checkBearerToken: sinon.stub().callsFake(() => Promise.resolve(token))
 }
 
-
-describe.only('Model > TutorialStore', function () {
+describe('Model > TutorialStore', function () {
   function fetchTutorials () {
     sinon.stub(rootStore.tutorials, 'fetchTutorials')
     return rootStore.workflows.setActive(workflow.id)
--- a/packages/lib-classifier/src/store/TutorialStore.spec.js
+++ b/packages/lib-classifier/src/store/TutorialStore.spec.js
@@ -50,10 +50,10 @@ const tutorialMiniCourseKind = TutorialFactory.build(
 )
 
 const upp = UPPFactory.build()
-const uppWithTutorialTimeStamp = UPPFactory.build({ 
+const uppWithTutorialTimeStamp = UPPFactory.build({
   preferences: {
     tutorials_completed_at: {
-      [tutorial.id]: seenMock 
+      [tutorial.id]: seenMock
     }
   }
 })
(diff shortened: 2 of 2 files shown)


eatyourgreens:
The new test fails with `Error: [mobx-state-tree] Failed to resolve reference '1' to type 'MediumResource' (from node: /tutorials/activeMedium)`. Does that mean the code is still trying to reference the linked media, even if they don't load?

packages/lib-classifier/src/store/TutorialStore.js
```diff
@@ -122,6 +122,7 @@ const TutorialStore = types
           // We're not setting the store state to error because
           // we do not want to prevent the tutorial from rendering
           console.error(error)
+          self.setMediaResources([])
```
srallen: `attachedMedia` is a map and I'd rather leave it as undefined rather than set it to an empty array if there is no attached media for the tutorial

packages/lib-classifier/src/store/TutorialStore.spec.js
```diff
@@ -495,73 +495,39 @@ describe('Model > TutorialStore', function () {
...
+      before(function () {
+        tutorialsClient.tutorials.getAttachedImages = sinon.stub().callsFake(() => {
+          return Promise.reject('Media could not be loaded')
+        })
       })
-    })
 
-    describe('for logged-in users', function () {
-      it('should show the tutorial if it has not been seen', function (done) {
+      it('should not show the tutorial', function (done) {
```
srallen: I don't think this should happen. We should allow the tutorial to load if the tutorial does not have any media. This is the current PFE behavior. Do you think that should change?

srallen:
Just to give context, we have launched approved projects that have tutorials that don't use the attached media resources. The American Solider is one: https://www.zooniverse.org/projects/tkotwim/the-american-soldier/classify

packages/lib-classifier/src/store/TutorialStore.spec.js
eatyourgreens: No. My reading of the store code was that the tutorial wasn’t intended to be shown on error. Either way, the code to show a tutorial is crashing if the media map is empty.

eatyourgreens:
That’s not a case where media exist but the request errors.

packages/lib-classifier/src/store/TutorialStore.spec.js
srallen: Hm, I think that's an unintended consequence of the yield for the media requests. I think we should render the tutorial anyway even if the media request errors.

eatyourgreens: I think it’s crashing because the active tutorial step is linked to an image that didn’t load, giving that unresolved reference error.

srallen:
I think the bug fix is checking that `self.attachedMedia` exists first too before setting the `activeMedium` here: https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/TutorialStore.js#L170

eatyourgreens:
Tests are passing now when tutorial media fail to load, but the classifier crashes with an unresolved reference error in that case. Which is the bug I was trying to test in this PR. 😕

eatyourgreens:
@srallen This should be ready for review again.

If you comment out the `Promise.all` in the tutorial store, you should find that the tests fail with an unresolved medium error, which is the bug I'm trying to catch.

packages/lib-classifier/src/store/TutorialStore.js
eatyourgreens: @roger pointed out that this generator doesn't yield anything on error, but I'm not sure what it should yield, if anything.

eatyourgreens:
That said, the classifier crashes with a white screen of death if `fetchMedia` errors, but the tests don't trigger that bug. 🤔

srallen:
> That said, the classifier crashes with a white screen of death if fetchMedia errors, but the tests don't trigger that bug.

Hm, I'm not sure if it is the line I pointed out in the previous comment, but it shouldn't try to set the active medium if for some reason the media resources wasn't set and is undefined.

Issue #205, opened 2018-10-26
Add feedback


rogerhutchings:
**Package**

classifier

**Feature or Issue Description**

The TESS beta will require feedback adding to the new classifier, along with a `range` feedback strategy. The majority of the code can be ported over from PFE [(relevant code)](https://github.com/zooniverse/Panoptes-Front-End/tree/master/app/features/feedback), with some modification to use Mobx instead of Redux.

PR #611, opened 2019-03-26
Add Graph2dRange feedback strategy

mcbouslog:
Package: lib-classifier

Towards #205.

Describe your changes:
- adds Graph2dRange strategy (`create-rule` and `reducer`, `feedback-mark` will be in subsequent PR)

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/store/feedback/strategies/datavis/graph2drange/feedback-mark.js
```diff
@@ -0,0 +1,41 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+
+function FeedbackMark ({ rule }) {
+  const color = (rule.success) ? 'green' : 'red'
+  const tolerance = parseInt(rule.tolerance, 10)
+  return (
+    <g>
```
eatyourgreens: The classifier is using d3 to draw the subject so I think you’ll need to use d3 to add marks to it. I’m not sure, but I don’t think you can use React components.

packages/lib-classifier/src/store/feedback/strategies/datavis/validations.js
```diff
@@ -0,0 +1,13 @@
+function validateDefaultTolerance (formState) {
```
eatyourgreens: I think this is meant to be added to the project builder.

packages/lib-classifier/src/store/feedback/strategies/datavis/graph2drange/feedback-mark.js
mcbouslog: Ah yes thanks for heads up! I removed for now, will revisit in next PR to show markings.

PR #612, opened 2019-03-26
Set env by query during initial server side load

srallen:
Package: app-project, lib-panoptes-js

Describe your changes:
This another attempt at setting the correct env by using url query param during the initial server side load in the app. I've decided to pass the env along as a query param, so it's still panoptes.js's responsibility to set the host based on the env for the requests, but it just adds another route to set it.

This makes loading a staging project in production work:
- Locally run `npm run build && npm start` to start the app in production mode
- Load up a staging project like: `http://localhost:3000/projects/nora-dot-test/planet-finders-beta/classify?env=staging`

Or:
- locally run `npm run dev` to start the app in staging mode 
- Load up a production project like: `http://localhost:3000/projects/zookeeper/galaxy-zoo/classify?env=production`

Either should work now. 

I updated the error handling in the app-project too because we were eating the server side errors. At the very least we'll now get a server side console log, it won't attempt to set the state when there's no project in the response, and it'll show a generic error message in the UI when there's no project. I couldn't figure out how to pass along the project not found error from the server to the client, but that's a separate issue.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/stores/Project.js
+++ b/packages/app-project/stores/Project.js
@@ -14,7 +14,7 @@ const Project = types
     completeness: types.optional(types.number, 0),
     description: types.optional(types.string, ''),
     display_name: types.maybeNull(types.string),
-    error: types.maybeNull(types.frozen({})),
+    error: types.optional(types.frozen({}), {}),
     experimental_tools: types.frozen([]),
     id: types.maybeNull(numberString),
     launch_approved: types.optional(types.boolean, false),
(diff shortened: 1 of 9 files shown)


packages/app-project/stores/Project.js
```diff
@@ -40,37 +40,42 @@ const Project = types
...
+              'retired_subjects_count',
+              'slug',
+              'subjects_count',
+              'urls'
+            ]
+            properties.forEach(property => { self[property] = project[property] })
+            self.loadingState = asyncStates.success
+          }
+
+          if (!project) throw new Error(`${slug} could not be found`)
```
eatyourgreens: If this line is moved up before the previous `if` block, then you shouldn't need the `if (project) {}` because the function will abort before trying to process the response.

srallen:
Travis CI appears to be having issues right now and the tests are passing locally...
https://www.traviscistatus.com/

I'm going to merge since this will fix what is currently broken in the app.

Issue #391, opened 2019-01-24
Add Field Guide


srallen:
**Package**
lib-classifer

**Feature or Issue Description**
Field Guide store could be similar to how Tutorial is being handled. Store and UI needs to be added. Modal itself is in react-components

**To Dos**

- [ ] Example

PR #614, opened 2019-03-27
FieldGuide store

srallen:
Package: lib-classifier

For #391

Describe your changes:
This is adding the store and tests for the field guide in the classifier. UI will be added in a separate PR

~WIP because we're working on improving the tests for the tutorial still and the field guide is fairly similar, so we should learn from that first.~

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/FieldGuideStore.js
+++ b/packages/lib-classifier/src/store/FieldGuideStore.js
@@ -44,7 +44,7 @@ const FieldGuideStore = types
     // TODO: this might need to paginate for field guides that have 20+ items
     const fetchMedia = flow(function * fetchMedia (fieldGuide) {
       const { type } = self
-      const { client } = getRoot(self)
+      const client = getRoot(self).client.panoptes
       if (fieldGuide) {
         self.loadingState = asyncStates.loading
         try {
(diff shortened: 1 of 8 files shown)


packages/lib-classifier/src/store/FieldGuideStore.js
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    return {
+      afterAttach,
+      fetchFieldGuide: flow(fetchFieldGuide),
+      reset,
+      setActiveItem,
+      setMediaResources,
+      setModalVisibility
+    }
```
eatyourgreens: Are `afterAttach`, `setActiveItem` and `setMediaResources` ever accessed from outside the model? If not, you could remove them from the export, which would simplify testing the model, and reduce the amount of stubbing the tests need to do.

srallen: `setActiveItem` definitely will be used by a click handler. The others probably not. Is it ok to update in the PR for the UI?

srallen: Since it's approved. I'll just plan on updating the exported actions in the next PR. Thanks!

PR #615, opened 2019-03-27
Focus accessibility for tutorial navigation

### LINKED ISSUE #607: Tutorial keyboard interaction
**Package** lib-classifier

From comments in #552 

> When a tutorial pops up on zooniverse.org, I can change slides by pressing the cursor keys. When the tutorial pops up here, pressing the cursor keys does nothing and there's no indicator to show me where focus is.

Also note that Grommet's RadioButton does not behave the way we expect:

> RadioButtonGroup handles focus accessibly. RadioButton on it's own does not. We recommend RadioButtonGroup as the preferred means of handling radio button input. You can file an issue on RadioButton. However, I suspect it will not receive much attention from the grommet core team unless an example arises where RadioButtonGroup isn't sufficient.

RadioButton needs focus or we should not use it.

srallen:
Package: lib-classifier

Closes #607.

Describe your changes:
I gave it another go to try using Grommet's `RadioButtonGroup` which does have accessible focus, and this seems to be working. @eatyourgreens if this is better, then let me know and I'll update tests accordingly. Is there anything else needed for accessibility here?

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.js
@@ -46,7 +46,7 @@ class SlideTutorial extends React.Component {
               height={200}
               src={medium.src}
             />}
-          <StyledMarkdownWrapper autoFocus isThereMedia={isThereMedia} overflow='auto'>
+          <StyledMarkdownWrapper aria-live='polite' autoFocus isThereMedia={isThereMedia} overflow='auto'>
             {/* TODO: translation */}
             <Markdownz>{step.content}</Markdownz>
           </StyledMarkdownWrapper>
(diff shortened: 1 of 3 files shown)


eatyourgreens:
This is a great improvement. If I focus the radio buttons, I can navigate the slides using the cursor keys. In PFE, I think we autofocus the group but I'm indifferent about that. On the plus side, it allows keyboard control of the slides as soon as the tutorial opens. On the minus side, it's an unannounced and confusing change of reading focus, particularly if you don't realise that the focus has skipped past the slide content.

In VoiceOver, I can navigate the first slide as soon as it opens, hearing the content in order. When I get to the navigation and press a key, I hear the step number change but the new slide content isn't announced. Maybe the slide contents need to be a live region too?

srallen:
@eatyourgreens I updated the tests and added a live area for the tutorial content, so this is ready for review again.

eatyourgreens:
Looks good. VoiceOver isn't announcing the change in media on step change, but I do hear the new text after changing the radio buttons.

srallen:
We set `alt=''` on the media here since we don't know what it is, so that might be why it's not read out. I'd like to open an issue with Panoptes to add a couple of fields to media resources to support adding a description and a credit field when we're redoing that part of the lab.

eatyourgreens:
Sorry, I meant that VoiceOver is skipping the media entirely on step change, so I don’t know there’s a video or image there. Empty alt would still be read as an image.

PR #616, opened 2019-03-27
Center button label text for need help

srallen:
Package: lib-classifier

Describe your changes:
Centers the button label. This is a regression from some point, but not sure when. Might have been from a grommet version bump.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
Changes look harmless to me, but tests are failing with `Error: Method “simulate” is meant to be run on 1 node. 0 found instead.` so I guess the component name change broke a test that looks for the button.

srallen:
Thanks for the fix!

eatyourgreens:
Just waiting for Travis to do its thing now.

PR #617, opened 2019-03-27
Fix Styling Bugs

### LINKED ISSUE #543: Classifier needs to respond to narrow screen widths
**Package**

lib-classifier

Screenshot 2019-03-13 21 30 39

### LINKED ISSUE #561: Styling bug in task area
**Package**

lib-classifier

The task area height matches that of the subject viewer, instead of matching its own content.

Screenshot 2019-03-18 10 41 58

wgranger:
Package:
lib-classifier

Closes #543 
Closes #561 

Describe your changes:
Fixes a couple bugs in the classifier. The first of which allows the task area to wrap under the subject area on narrow screen sizes. The second fix keeps the `TaskArea` and `ImageToolbar` heights equal. 

On issue #561, the error occurs due to the `pad: small` designation on the Grommet `Box` changing heights when the 768 responsive breakpoint is hit. The SVGs oddly try to compensate by growing in size, thus stretching the SVG icon, and toolbar, to the size shown in that issue.

# Review Checklist

## General

## Apps

- [x] Does it work on mobile?


packages/lib-classifier/src/components/Classifier/components/MetaTools/MetaTools.js
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { Box } from 'grommet'
 import { inject, observer } from 'mobx-react'
+import styled from 'styled-components'
```
rogerhutchings: I don't think you need this

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
```diff
@@ -3,12 +3,17 @@ import { Box, Paragraph } from 'grommet'
...
 import React from 'react'
-import { ThemeProvider } from 'styled-components'
+import styled, { ThemeProvider } from 'styled-components'
 
 import getTaskComponent from './helpers/getTaskComponent'
 import TaskHelp from './components/TaskHelp'
 import { default as TaskNavButtons } from './components/TaskNavButtons'
 
+const StyledBox = styled(Box)`
+  max-height: 600px;
```
rogerhutchings: Isn't this just going to cause content clipping if we have a task with lots of choices?

wgranger: Ah yes. I meant to add an overflow here.

srallen: I realized this was a bad idea in my comment: https://github.com/zooniverse/front-end-monorepo/issues/561#issuecomment-474437971

Volunteers might miss tasks this way. It's should be a min-height instead that allows the container to grow.

wgranger: Gotcha. I was thinking we were using a max-height and encouraging users to stick to max task numbers with overflow as a safety net. I've pushed a new commit removing max height. The container should grow as needed.

srallen: We can maybe do that once the workflow editor page is redesigned, but that's a ways off still.

srallen:
I've rebased this to fix the conflicts

srallen:
@rogerhutchings I think I fixed those things. We might want to consider refactoring the default layout to use Grommet's ResponsiveContext so that it uses the same breakpoints as other components we've started to use it with.

rogerhutchings:
Cool, will do another check. We should also add those breakpoints into the theme at some point

srallen:
Just rebased with the latest master.

srallen:
I'm making sure a few things in #695 get into this PR.

srallen:
You may notice there's an extra gap in the metatools area between the subject metadata button and the favorites button. This is because we're hiding the modal there. I'd like to refactor that to get rid of this and have a single container fragment component for the button and modal like I did with the field guide. 

@rogerhutchings could I get another review on this? There's a few additional fixes for #695 here.

PR #637, opened 2019-04-01
Field guide modal

### LINKED ISSUE #391: Add Field Guide

srallen:
Package:

Closes #391.

Describe your changes:
Adds the UI for the field guide

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


srallen:
My project 1233 has a field guide. The default project loaded by the classifier does not.

srallen:
I don't have this setup for the light/dark themes yet. This is an already large PR so I'll note in an issue.

srallen:
This also has the issue of the size of the modal changing between the full list of items and selected item. I haven't come up with a better way to have it be responsive and have a max-width yet...

eatyourgreens:
Testing this out in VoiceOver, I'm wondering if the item buttons should be links to the individual items? Give it a go and let me know what you think. When I press a button, it's not immediately obvious that I've navigated to the item, although I can continue navigating with VoiceOver and read the new text.

There's an annoying jump in height between an item loading and the linked image loading. Can we show a placeholder for images, or set a default height on that box.

packages/lib-classifier/src/store/FieldGuideStore.js
```diff
@@ -92,11 +92,18 @@ const FieldGuideStore = types
...
     }
 
-    function setActiveItem (index) {
+    function setActiveItem (item) {
       const fieldGuide = self.active
-      if (fieldGuide && fieldGuide.items.length === index + 1 && fieldGuide.items[index]) {
-        self.activeItem = index
-        if (fieldGuide.items[index].icon) self.activeMedium = fieldGuide.items[index].icon
+      if (fieldGuide) {
+        const items = fieldGuide.items
```
eatyourgreens:
```suggestion
        const { items } = fieldGuide
```

packages/lib-classifier/src/store/FieldGuideStore.js
```diff
@@ -92,11 +92,18 @@ const FieldGuideStore = types
...
+        const index = items.indexOf(item)
+        if (items[index]) {
+          const icon = items[index].icon
```
eatyourgreens:
```suggestion
          const { icon } = items[index]
```

packages/lib-classifier/src/store/FieldGuideStore.js
```diff
@@ -92,11 +92,18 @@ const FieldGuideStore = types
...
+          if (icon) self.activeMedium = icon
+          return self.activeItem = index
```
eatyourgreens: Why is `self.activeItem` an index, and not the item? I find that confusing.

srallen:
The new suggested changes feature is neat.

eatyourgreens:
Yeah. I haven't tried it out for changes that are more than a few characters. Big changes might break things.

srallen:
I updated to use an anchor instead and I like it a lot better for the hover state than what I came up with for the button.

eatyourgreens:
VoiceOver says there's a missing counterpart translation in the link text somewhere.
Screenshot of the field guide item links, showing a missing translation error for the link text.

srallen:
@eatyourgreens Oops! That's fixed in the latest commit.

eatyourgreens:
I'm also getting `setActiveItem is not a function` when I click those links.

srallen:
In addition to the refactor to use Anchor instead of Button for the items list, I've:
- Created an icon component that uses a placeholder fallback consistently between the list items view and the item details view
- Added Grommet's ResponsiveContext component to set the widths for the outer Box container. This fixes it between the item list view and item detail view and also makes it responsive. Note Enzyme does not yet support React's `createContext`, so I cannot test how `FieldGuide` works until Enzyme updates this. See: https://github.com/airbnb/enzyme/issues/1908
- Refactored how I was using Grid. It was just wrong before. I'm still learning!
- Finished the fix for renaming the `setActiveItemIndex` and `activeItemIndex`
- Fix the tests

packages/lib-classifier/src/store/FieldGuideStore.js
```diff
@@ -92,19 +92,27 @@ const FieldGuideStore = types
       self.showModal = boolean
     }
 
-    function setActiveItem (index) {
+    function setActiveItemIndex (item) {
```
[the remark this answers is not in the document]
srallen: The action is now using index consistently.

srallen:
The focus bug I'll open as a separate issue. It might be a Grommet bug.

eatyourgreens:
I've pulled down the rebased branch and the item links aren't working any more. I'll have a look to see if I can figure out why.

packages/lib-classifier/src/store/FieldGuideStore.js
```diff
@@ -92,20 +92,14 @@ const FieldGuideStore = types
...
-        const { items } = fieldGuide
-        const index = items.indexOf(item)
-        if (items[index]) {
-          const { icon } = items[index]
-          if (icon) self.activeMedium = icon
-          return self.activeItemIndex = index
-        }
-
-        return self.activeItemIndex = undefined
+      if (fieldGuide && fieldGuide.items.length === index + 1 && fieldGuide.items[index]) {
```
eatyourgreens: `fieldGuide.items.length === index + 1` is only true for the last item can be active at the moment.

eatyourgreens: This should have been caught by a test too.

srallen: I've fixed this and updated the tests too.

packages/lib-classifier/src/store/FieldGuideStore.spec.js
```diff
@@ -32,7 +36,7 @@ const fieldGuideWithoutIcon = FieldGuideFactory.build({
 
 const project = ProjectFactory.build()
 
-describe('Model > FieldGuideStore', function () {
+describe.only('Model > FieldGuideStore', function () {
```
eatyourgreens: Stray `.only`

srallen: Whoops! That should be removed now.

eatyourgreens:
Just waiting on Travis.

PR #644, opened 2019-04-02
Remove the default transparent background for drop from theme

### LINKED ISSUE #628: Weirdness in add to collections dropdown
When you try and add a subject to an existing collection the dropdown is appearing all weird and transparent:

image

srallen:
Package: lib-grommet-theme, app-project

Closes #628

Describe your changes:
The issue was the transparent background added for the drop component in our theme. The Select component uses Drop. I set transparent as the default background to fix an issue I had in an earlier version of Grommet v2, but that seems to be fixed, so the transparent background default can be removed. I also set a height for the drop container. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
The text input shouldn’t render underneath the combo box either.

srallen:
@rogerhutchings I've removed `shadowSize: 'none'` from our theme too, which was also a part of the fix I did a while ago that may not apply anymore. This should help differentiate the dropdown from the background.

@eatyourgreens Unfortunately that's the design of the Select component which uses DropButton: https://storybook.grommet.io/?path=/story/select--search
I'm guessing they wanted to keep the search input separate from the currently selected item in its UI?

srallen:
Whoops, I didn't get that pushed up when I commented. You'll see my latest commit now.

Issue #486, opened 2019-02-21
Rewrite Tabs component


rogerhutchings:
**Package**

lib-classifier

**Feature or Issue Description**

The code for the custom Tabs component in the Tasks area is a bit finicky and could be DRYed up some.

rogerhutchings:
Fixed by #646 🎉

PR #646, opened 2019-04-03
Set tabs active state

### LINKED ISSUE #620: Active tab doesn't have a different colour
**Package**

lib-classifier

The active tab in the workflow area doesn't have a different colour to denote it.

rogerhutchings:
Package: lib-classifier

Closes #620, #486.

Rewrites the tabs component to simplify a lot of the styling by passing in a custom theme, and fix the active tab colour not being set.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
VoiceOver announces the selected tab, so can we use `[aria-selected]` to style the tabs? Might cut down the code some more.

rogerhutchings:
> VoiceOver announces the selected tab, so can we use [aria-selected] to style the tabs? Might cut down the code some more.

Good call, I've made that change

PR #648, opened 2019-04-03
Init add of feedback brushes to feedback modal

mcbouslog:
Package: lib-classifier

Towards #205.

Describe your changes:
- adds feedback related brushes to `LightCurverViewer`

**Refactor in progress:**
- extract feedback specific code from `LightCurveViewer`
- move feedback specific brush code elsewhere
- add to LCV something like `renderAlternativeBrushes` that takes custom brush render code prop

^ should be done soon, but don't want to hold this up for beta testing

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/Feedback/components/Graph2dRangeFeedback.js
+++ b/packages/lib-classifier/src/components/Classifier/components/Feedback/components/Graph2dRangeFeedback.js
@@ -33,6 +33,7 @@ class Graph2dRangeFeedback extends Component {
   }
 
   drawFeedbackBrushes (d3annotationsLayer, repositionBrush) {
+    console.log('d3annotationsLayer = ', d3annotationsLayer)
     const annotationBrushes = []
     this.props.annotations.forEach(annotation => {
       const { task, value } = toJS(annotation)
--- a/packages/lib-classifier/src/components/Classifier/components/Feedback/components/Graph2dRangeFeedback.js
+++ b/packages/lib-classifier/src/components/Classifier/components/Feedback/components/Graph2dRangeFeedback.js
@@ -33,7 +33,6 @@ class Graph2dRangeFeedback extends Component {
   }
 
   drawFeedbackBrushes (d3annotationsLayer, repositionBrush) {
-    console.log('d3annotationsLayer = ', d3annotationsLayer)
     const annotationBrushes = []
     this.props.annotations.forEach(annotation => {
       const { task, value } = toJS(annotation)
--- a/packages/lib-classifier/src/components/Classifier/components/Feedback/components/Graph2dRangeFeedback.js
+++ b/packages/lib-classifier/src/components/Classifier/components/Feedback/components/Graph2dRangeFeedback.js
@@ -1,0 +1,116 @@
+import * as d3 from 'd3'
+import { toJS } from 'mobx'
+import { inject, observer } from 'mobx-react'
+import React, { Component } from 'react'
+
+import LightCurveViewer from '../../SubjectViewer/components/LightCurveViewer'
+
+function storeMapper (stores) {
+  const annotations = stores.classifierStore.classifications.currentAnnotations
+  const { 
+    applicableRules,
+    showModal: feedback
+  } = stores.classifierStore.feedback
+  const {
+    active: subject
+  } = stores.classifierStore.subjects
+
+  return {
+    annotations,
+    applicableRules,
+    feedback,
+    subject
+  }
+}
+
+@inject(storeMapper)
+@observer
+class Graph2dRangeFeedback extends Component {
+  constructor () {
+    super()
+
+    this.drawFeedbackBrushes = this.drawFeedbackBrushes.bind(this)
+  }
+
+  drawFeedbackBrushes (d3annotationsLayer, repositionBrush) {
+    const annotationBrushes = []
+    this.props.annotations.forEach(annotation => {
+      const { task, value } = toJS(annotation)
+      value.forEach((marking, i) => {
+        const markingBrush = {
+          id: i,
+          brush: d3.brushX(),
+          maxX: (marking.x + (marking.width / 2)),
+          minX: (marking.x - (marking.width / 2))
+        }
+        annotationBrushes.push(markingBrush)
+      })
+    })
+
+    const ruleBrushes = []
+    this.props.applicableRules.forEach(rule => {
+      const ruleBrush = {
+        id: rule.id,
+        brush: d3.brushX(),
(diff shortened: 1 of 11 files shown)


packages/lib-classifier/src/store/FeedbackStore.js
```diff
@@ -15,6 +15,14 @@ const FeedbackStore = types
     onHide: () => true
   }))
   .views(self => ({
+    get applicableRules () {
+      return flatten(Array.from(self.rules.values()))
+        .map(rule => {
+          if ((rule.success && rule.successEnabled) || (!rule.success && rule.failureEnabled)) {
```
eatyourgreens: I might be remembering this wrongly, but aren't the rules already filtered for this when they're first set up?

eatyourgreens: Also, this is better done with `.filter` rather than `.map`.

eatyourgreens:
I really dislike how this hacks a specific feedback strategy into the subject viewer. My reading of the code is that it will only work for the TESS workflow and not for any other. Is that right? 

Rendering code for a strategy should be abstracted out into a function for that strategy, like it is in the PFE feedback code. The subject viewer should then be using a strategy pattern to render the feedback marks.

packages/lib-classifier/src/store/FeedbackStore.js
```diff
@@ -15,6 +15,14 @@ const FeedbackStore = types
     onHide: () => true
   }))
   .views(self => ({
+    get applicableRules () {
```
eatyourgreens: What are these rules applicable to? I thought it might be a specific task, but the code isn't filtering on task key.

mcbouslog:
💯 agree should be abstracted out, have similar comment you made Jim in Slack pinned in my notes, working on that refactor, but wanted to get this out there if needed asap for beta testing.

rogerhutchings:
@mcbouslog We've got some time still - let's keep the feedback code out of the light curve viewer (@eatyourgreens suggested doing it with a decorator, which I really like), as the rest of it will need to be refactored and we probably don't want to make that job bigger than we have to :)

eatyourgreens:
Looking at how [feedback marks were implemented in PFE](https://github.com/zooniverse/Panoptes-Front-End/blob/95bb51fe7a0de27bc90022399b61cbd3c98e3d71/app/features/feedback/classifier/helpers/open-feedback-modal.jsx#L22-L34), the equivalent here would be a computed property on the model, `feedback.marks` that loops through the feedback rules and returns the marks, or functions that generate marks.

packages/lib-classifier/src/store/FeedbackStore.js
mcbouslog: Maybe could use a better name, but essentially this is getting the rules that have messages (success or failure) that will be shown. The updated test rules show an example where I'd expect the `classification.metadata.feedback` to include three rules, one of which a message was never shown, but `applicableRules` is designed to return the 2 rules for which a message will be shown.

packages/lib-classifier/src/components/Classifier/components/Feedback/components/Graph2dRangeFeedback.js
```diff
@@ -0,0 +1,116 @@
...
+      repositionBrush(feedbackBrush, d3brush)
+    })
+  }
+
+  render () {
+    const { feedback, subject } = this.props
+    return (
+      <LightCurveViewer 
+        drawFeedbackBrushes={this.drawFeedbackBrushes}
+        feedback={feedback}
```
mcbouslog: I don't like adding `drawFeedbackBrushes` or `feedback` as props, considered adding to the feedback store, but the thought of a workflow with different types of feedback strategies (i.e. "single answer question" and "radial") haunts me, which would complicate what would be added to the feedback store so different types of feedback markings could be shown in the subject viewer.

eatyourgreens: Looking at the feedback code in PFE, I think what you’re looking for here is a Strategy pattern. 

So, the viewer loops through annotations and invokes an abstract renderer that should render a mark for that annotation, based on the annotation type. Each feedback strategy then defines a method that renders a feedback mark for an annotation.

mcbouslog: I think [this code in PFE](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/drawing/markings-renderer.cjsx#L49-L53) is missing from the current LCV, or maybe the DataVisAnnotationTask, but I'm doing something similar albeit not ideally with `drawFeedbackBrushes`. 

I feel like the drawing tasks and the subject viewer are decoupled differently in PFE than the `dataVisAnnotationTask` and the LCV, but maybe I'm wrong or it shouldn't matter.

eatyourgreens: That's exactly the code I was thinking of. The same pattern should apply here, I think.

mcbouslog: So I think I'd need to refactor the LCV, I don't think it's currently configured to create markings from existing annotations? per https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js#L214

eatyourgreens: `annotations` are passed in as props, which is all you need to get the strategies right?

mcbouslog: I don't think the annotations coming into the LCV from the classifier store include feedback information, I think in PFE we [edit the annotations so they do include feedback](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/features/feedback/classifier/helpers/open-feedback-modal.jsx#L51-L55), then pass those edited annotations (with feedback) down through the tree as props.

eatyourgreens: Oh, that's a pain. I was thinking the code would be something like
```javascript
annotations.forEach(function (annotation) {
  strategies[annotation.strategy].drawFeedback()
}
```
but I hadn't thought about how rules map to annotations.

eatyourgreens:
>We'll just have to be disciplined to make sure this gets done

With this PR, feedback only supports the TESS light-curve marking task, so feedback in other projects will make sure this gets rewritten.

PR #659, opened 2019-04-05
Remove banners from DOM when inactive

### LINKED ISSUE #657: 'Already seen' banners are present when subject hasn't been seen
**Package**
lib-classifier

Testing in VoiceOver revealed that four copies of the 'Already Seen' banners are present on new, unseen subjects.

They should only be present for seen subjects, and only one should be present for a given subject.
Screenshot showing that VoiceOver finds four Already Seen banners on every subject.

rogerhutchings:
Package: lib-classifier

Closes #657.

Adds `react-pose` and to enable entry and exit animations, and removes banners from the DOM when not active.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
Those unlabelled buttons should all be labelled, even if they don't have tooltips, but that's a different issue.

rogerhutchings:
#664

PR #661, opened 2019-04-06
Fix linting errors

lancesnider:
Package: App Project

Describe your changes:

Cleaning up linting errors. 

There's one error left, but I wanted to double check before I fix it. In `CreateCollection.js` there's a non-camelCase identifier named `display_name`. It's used in several places so I wanted to double check that it wasn't on purpose before renaming it. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/app-project/src/screens/ClassifyPage/components/CollectionsModal/CollectionsModalContainer.js
```diff
@@ -29,7 +34,7 @@ class CollectionsModalContainer extends Component {
     this.state = {
       active: false,
       newCollection: {
-        display_name: '',
+        displayName: '',
```
srallen: This change is causing a couple of tests to fail. I think we might want to leave it as `display_name` despite the linter complaining. The resources' attributes are snake case.

lancesnider: Oops! Looks like the linter automatically converted that one. I'll figure out how to make it ignore those errors and not automatically change this one.

lancesnider: @srallen Fixed. I also told `standard` to ignore the lines that were throwing errors: https://github.com/zooniverse/front-end-monorepo/pull/661/files#diff-1277d75a3bea17c9e1e77ad4b9dc4dc3R40

srallen:
I think this branch is having the same problem as we've seen in other repos since we don't allow travis to run on forks.

eatyourgreens:
I'm not sure why Travis isn't running for forked branches, but I've turned off the requirement for branch tests to pass. PR tests still need to pass before merging is allowed.

PR #665, opened 2019-04-08
Subject metadata modal fixes

### LINKED ISSUE #649: Subject metadata is invisible in VoiceOver
**Package**
lib-classifier

Subject metadata is announced as a table with one column and zero rows in VoiceOver and can't be entered or read out.
Screenshot showing the Subject Metadata table in VoiceOver.

srallen:
Package: lib-classifier

Closes #649 

Describe your changes:
I removed the `size` prop from the DataTable and the table is now accessible. This seems to be a Grommet bug, opened issue at https://github.com/grommet/grommet/issues/3006

To still have the table overflow, I wrapped it with a Box and set the height and overflow. 

I reduced the size of the rows. I'm not 100% what the size should be since the InDesign prototype is not inspectable: https://projects.invisionapp.com/d/#/console/12923997/270495352/preview so I tried to eyeball it. @beckyrother what should the size of the rows be here?

I extracted out the function that filters what is displayed to be usable by the store. Project 272 has subjects that only have hidden metadata, so the button should be hidden. (This is actually a bug in current PFE now btw). The store view function isn't tested.

Also, the original implementation of the filters supports `//` as a filter, but I do not know of any project that uses this? I think we should drop support for it. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/MetaTools/MetaTools.js
+++ b/packages/lib-classifier/src/components/Classifier/components/MetaTools/MetaTools.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import { Box } from 'grommet'
 import { inject, observer } from 'mobx-react'
 import MetadataButton from './components/MetadataButton'
-import MetadataModal from './components/MetadataModal'
+import { MetadataModal } from './components/MetadataModal'
 import FavouritesButton from './components/FavouritesButton'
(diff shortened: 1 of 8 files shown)


beckyrother:
The left column is 100px, the gutter between the columns is 20px, the right column is the remainder

srallen:
@eatyourgreens  would you review the new tests I added in the newest commits? Thanks!

Issue #667, opened 2019-04-09
Mark up lists of links as unordered lists


eatyourgreens:
**Package**

**Feature or Issue Description**

**To Dos**

- [ ] Example

eatyourgreens:
[`aria-current`](https://tink.uk/using-the-aria-current-attribute/) looks useful too.

rogerhutchings:
Zooniverse site navigation:

<img width="508" alt="Screenshot 2019-09-14 11 44 27" src="https://user-images.githubusercontent.com/2725841/64907013-467d9180-d6e5-11e9-9d1a-0914b8383307.png">

rogerhutchings:
User profile dropdown menu:

<img width="988" alt="Screenshot 2019-09-14 11 48 41" src="https://user-images.githubusercontent.com/2725841/64907051-b4c25400-d6e5-11e9-82e6-ff59caa02750.png">

eatyourgreens:
Dropdown menus should be navigable with up/down keys too. The user menu does do this in PFE.

eatyourgreens:
Lists of links on the publications page aren’t marked up as lists either.

PR #670, opened 2019-04-09
Render link lists as ul in footer

rogerhutchings:
Package: lib-react-components

cc #667 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-react-components/src/ZooFooter/components/LinkList/LinkList.js
+++ b/packages/lib-react-components/src/ZooFooter/components/LinkList/LinkList.js
@@ -1,34 +1,47 @@
 import { Anchor, Box } from 'grommet'
-import { arrayOf, string} from 'prop-types'
+import { arrayOf, string } from 'prop-types'
 import React from 'react'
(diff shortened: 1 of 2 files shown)


eatyourgreens:
The failing test needs to be fixed, but I think that might not be related to this PR.

eatyourgreens:
Here's the test that's blocking this from merging.
```
1) Subjects Helpers
       buildQueuedSubjectResource
         should return a mocked subject with a random id > 0 and <= 100:
     AssertionError: expected 0 to be within 1..100
      at Context.<anonymous> (src/resources/subjects/helpers.spec.js:13:40)
```
I guess that test is going to fail ~1% of the time and just happened to fail by chance on this PR.

eatyourgreens:
Rerunning the Travis build generated a valid, random subject ID this time.

srallen:
Thanks for catching that test failure. I'd like to copy or move the rosie factory setup that I've started to use in the lib-classifier to lib-panoptes-js.

Issue #666, opened 2019-04-08
Design: Field Guide


beckyrother:
- [x] Is it possible to drag the field guide around the screen? One of the main goals for moving this to a modal was to allow it to stay open while the user makes classifications. (interim solution in #697)

- [x] All modals should have a shadow to indicate that they are 'floating' above the main site content `box-shadow: 0px 10px 20px #00000030;`

- [x] Margins should be even at 30px between left and right sides. (This is my fault as the InVision did a bad job)

- [x] Decrease margin between the H2s and the paragraphs under them – both elements do not need margins.

- [ ] Thumbnail title: `line-height: 18px`

eatyourgreens:
Anything that’s draggable with a mouse should also be moveable with the keyboard. 

Modals, by definition, block interaction with anything else while they’re open. I don’t know if our Modal component will allow you to classify while it’s open.

beckyrother:
I recognize that this is different to our other modals; the reason to make this a modal rather than a sliding panel was to be able to classify at the same time rather than some volunteers keeping two browser windows open.

eatyourgreens:
Grommet’s Layer component lets you turn off modal behaviour, so that’s good. I don’t know if Layers can be moved, though.

eatyourgreens:
@beckyrother Grommet is opinionated about modals. If you want the classifier to be available while the guide is open, the Field Guide can’t be modal.

beckyrother:
Is there another component that could be used instead?

rogerhutchings:
Not in grommet

eatyourgreens:
You can have floating layers that aren’t modal in Grommet, but I don’t know if they can be moved around.

eatyourgreens:
I’m leaning towards the Field Guide being a separate thing that can be linked to from the classifier and Talk, per #662.

beckyrother:
Is it possible to pop out the component into a separate window? This current modal is likely going to make classification worse if the modal covers up the subject image. 

In the future, what is a good way to communicate design intent like this? Since I'm not in the Oxford office to drop by and see progress and the Zoominar about this was so long ago, I'm not sure where best to have included this information.

srallen:
[...]

PR #672, opened 2019-04-09
Field guide style fixes

srallen:
Package: lib-classifier, lib-react-components, lib-grommet-theme

For #666

Describe your changes:
Fixes:

- heading and paragraph styling in selected item content
- item heading no longer being cutoff when overflow happens
- adds box-shadow to the react-components Modal
- set paragraph default in theme to take a max-width of 100% and more specific styles can be defined in components

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/FieldGuide/components/FieldGuide/components/FieldGuideItem/FieldGuideItem.js
+++ b/packages/lib-classifier/src/components/Classifier/components/FieldGuide/components/FieldGuide/components/FieldGuideItem/FieldGuideItem.js
@@ -28,6 +28,23 @@ const FieldGuideItemHeader = styled(Box)`
   }
 `
 
+const FieldGuideItemContent = styled(Box)`
+  > h1, h2, h3, h4, h5, h6 {
+    font-size: 14px;
+    letter-spacing: 1.5px;
+    line-height: 17px;
+    margin: 5px 0;
+  }
+
+  > p {
+    margin: 5px 0;
+
+    &:last-of-type {
+      margin: 5px 0 20px;
+    }
+  }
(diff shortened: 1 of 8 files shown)


srallen:
I'll branch off this one to update the field guide to not be a modal and use the Grommet Layer directly to open on the side.

packages/lib-classifier/src/components/Classifier/components/FieldGuide/components/FieldGuide/components/FieldGuideItem/FieldGuideItem.js
```diff
@@ -28,6 +28,19 @@ const FieldGuideItemHeader = styled(Box)`
   }
 `
 
+const FieldGuideItemContent = styled(Box)`
+  > h1, h2, h3, h4, h5, h6 {
+    font-size: 14px;
```
eatyourgreens: Should pixel sizes like this be set in the theme? It seems odd, to me, to have components that use non-standard spacings/sizes.

srallen: The theme's defaults are set to what Becky's defined here: https://projects.invisionapp.com/d/main#/projects/prototypes/12923997/assets/77727645

The field guide's design though shows the same size for all the sub-headings in the markdown, so, I'm setting it here, since it's the only thing so far doing something like this (the tutorial is only resetting the size for h1 and h2 only). If we find ourselves needing to repeat this somewhere else, we can either use the `extend` theme option or maybe set the small size headings in the theme to do something like this?

Issue #636, opened 2019-04-01
Rebuild subject queue on login / logout


rogerhutchings:
**Package**

lib-classifier

**Feature or Issue Description**

We need to rebuild the subject queue when the user logs in or out, as it will contain states that are no longer accurate (e.g. `already_seen`)

eatyourgreens:
One way to do this would be to key the classifier on user ID, so that it unmounts and remounts when the user changes.

eatyourgreens:
I think that's currently handled by keying on workflow ID, so the key could be expanded to `workflow-${workflow.id}-user-${user.id}` when there's someone logged in.

eatyourgreens:
Turns out the classifier is already set to unmount when the user changes.
https://github.com/zooniverse/front-end-monorepo/blob/2b776f73966ad53fd629feddbed7bf768efcd991/packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapperContainer.js#L81-L92

PR #690, opened 2019-04-10
Disable add to collections for anonymous users

### LINKED ISSUE #658: Add to favourites and Add to collection should require sign-in
**Package**
lib-classifier

'Add to favourites' and 'Add to collection' should only be available if you've logged in.

eatyourgreens:
Disable 'add to collections' or 'add to favourites' if there's no logged-in user.

Package:
lib-classifier

Closes #658.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/MetaTools/MetaTools.js
+++ b/packages/lib-classifier/src/components/Classifier/components/MetaTools/MetaTools.js
@@ -9,9 +9,11 @@ import CollectionsButton from './components/CollectionsButton'
 
 function storeMapper(stores) {
   const { active: subject, isThereMetadata } = stores.classifierStore.subjects
+  const upp = stores.classifierStore.userProjectPreferences.active
   return {
     isThereMetadata,
-    subject
+    subject,
+    upp
   }
 }
 
@@ -45,7 +47,7 @@ export default class MetaTools extends React.Component {
   }
 
   render () {
-    const { className, isThereMetadata, subject } = this.props
+    const { className, isThereMetadata, subject, upp } = this.props
     return (
       <Box className={className} direction="row">
         {isThereMetadata &&
@@ -58,9 +60,11 @@ export default class MetaTools extends React.Component {
           />}
         <FavouritesButton
           checked={subject && subject.favorite}
+          disabled={!upp}
           onClick={this.toggleFavourites}
         />
         <CollectionsButton
+          disabled={!upp}
           onClick={this.addToCollection}
         />
       </Box>
--- a/packages/lib-classifier/src/components/Classifier/components/MetaTools/components/CollectionsButton/CollectionsButton.js
+++ b/packages/lib-classifier/src/components/Classifier/components/MetaTools/components/CollectionsButton/CollectionsButton.js
@@ -20,25 +20,28 @@ export const Collect = styled(CollectionsIcon)`
 `
 
 export default function CollectionsButton (props) {
-  const { onClick, theme } = props
+  const { disabled, onClick, theme } = props
   return (
     <ThemeProvider theme={{ mode: theme }}>
       <PlainButton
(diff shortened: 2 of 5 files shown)


eatyourgreens:
Not the greatest solution to #658 but it works.

shaunanoordin:
## PR Review

This PR fixes an issue where non-logged-in users could _attempt to_ add a Subject to their favourites or to a collection. These features should _only_ be available to users that are signed in.

Screen Shot 2019-04-10 at 17 02 35
_app-project showing a Classifier with the "Add to Favourites" and "Add to Collections" buttons_

With this fix, if a non-logged-in user clicks on the "Add to Favourites" 🇬🇧 or "Add to Collections" button on the Classifier, the user is prompted with the _Sign-In modal._ A signed-in user can click these two buttons to execute their functions as normal.

Without this fix, a non-logged-in user who clicks on "Add to Favourites" would immediately crash the app, and one who clicks on "Add to Collections" will show the Collections modal _before_ crashing the app.

### Testing

Tested on _app-project_ (note to self: do I need to check on lib-classifier?), on the localhost+OSX+Chrome, using the [I Fancy Cats](http://localhost:3000/projects/brooke/i-fancy-cats/classify) staging project.

This PR's changes to the button behaviours work mostly as intended, save for one quirk.

Screen Shot 2019-04-10 at 17 00 39

Quirk: clicking _"Add To Favourites"_ 🇬🇧 as a non-logged-in user correctly displays the _Sign In modal,_ but ALSO incorrectly marks the Subject as Favourited (i.e. the heart turns red)

This may be confusing if the user cancels out of the Sign In modal, because the Favourite heart _continues_ to be red even if there's logically no user profile to add a "favourite" to.

Further tests: determine if Subject is correctly added to Favourites after Signing In.

### Status

Code review is still ongoing.

eatyourgreens:
Logging in rebuilds the subject queue (#636), which is why I'm not 100% happy with this solution.

shaunanoordin:
> Logging in rebuilds the subject queue (#636), which is why I'm not 100% happy with this solution.

Aye, I just noticed it actually basically reloads the page.

A non-logged-in user clicking "Add To Favourite" or "Add to Collection" causes the Sign In modal to appear, and when the user logs in, the page reloads - meaning the Subject they were looking at is lost, and isn't added to favourites/collections.

While this is definitely a bad user experience, this fix _does_ solve the more pertinent issue of the app crashing, so hmm.

eatyourgreens:
I think my ideal solution would be that logging in refreshes the incoming subject queue, but doesn’t change the active subject. Pinging @srallen and @rogerhutchings for opinions about how the subject store should change when you log in.

shaunanoordin:
> I think my ideal solution would be that logging in refreshes the incoming subject queue, but doesn’t change the active subject.

Agreed that this is the ideal situation, though there are some edge cases that I'm not sure are solved. _(e.g. consider a project with an A-B split. If a non-logged-in user sees a Subject from the A-split, the logs into a user account in the B-split, will that user still see that the active Subject is from the A-split? Same question for projects that allow "levelling up" to new workflows like Gravity Spy.)_

If we're tight on time, a faster solution would be to do what PFE currently does: hide the _"Add to Favourite"_ and _"Add to Collections"_ buttons until the user signs in, with a _"you should sign in"_ prompt where the buttons should be.

eatyourgreens:
@shaunanoordin I replaced this branch with a version that disables those buttons if you aren't logged in.

eatyourgreens:
There still seems to be a bug where the classifier crashes if you log in and Add To Collections after the page has loaded.

eatyourgreens:
>There still seems to be a bug where the classifier crashes if you log in and Add To Collections after the page has loaded.

That bug should be fixed by #706.

PR #698, opened 2019-04-11
Add selected_at to subjects

eatyourgreens:
Add selected_at to the subject model, and copy it across to new classifications.

Package:
lib-classifier

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/Classification.js
+++ b/packages/lib-classifier/src/store/Classification.js
@@ -19,6 +19,7 @@ const ClassificationMetadata = types.model('ClassificationMetadata', {
     already_seen: types.optional(types.boolean, false),
     finished_workflow: types.optional(types.boolean, false),
     retired: types.optional(types.boolean, false),
+    selected_at: types.maybe(types.string),
     selection_state: types.maybe(types.string),
     user_has_finished_workflow: types.optional(types.boolean, false)
   }),
--- a/packages/lib-classifier/src/store/ClassificationStore.js
+++ b/packages/lib-classifier/src/store/ClassificationStore.js
@@ -74,6 +74,7 @@ const ClassificationStore = types
             already_seen: subject.already_seen,
             finished_workflow: subject.finished_workflow,
             retired: subject.retired,
+            selected_at: subject.selected_at,
             selection_state: subject.selection_state,
             user_has_finished_workflow: subject.user_has_finished_workflow
           },
--- a/packages/lib-classifier/src/store/Subject.js
+++ b/packages/lib-classifier/src/store/Subject.js
@@ -11,6 +11,7 @@ const Subject = types
     locations: types.frozen(),
     metadata: types.frozen(),
     retired: types.optional(types.boolean, false),
+    selected_at: types.maybe(types.string),
     selection_state: types.maybe(types.string),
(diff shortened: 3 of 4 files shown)


eatyourgreens:
@camallen I wasn't sure what type the new attribute is, so I've guessed at Date.

camallen:
> @camallen I wasn't sure what type the new attribute is, so I've guessed at Date.

It is a DateTime string in UTC `selected_at: "2019-04-11T11:16:29.063Z"`

eatyourgreens:
Hopefully that's covered by `types.Date`. If this doesn't work, we can change the model to use a string.

shaunanoordin:
> Hopefully that's covered by types.Date. If this doesn't work, we can change the model to use a string.

Would it help if we updated the tests, e.g. so `Classification.spec.js` checks if `"2019-04-11T11:16:29.063Z"` doesn't choke up `types.Date`? This might also be useful for other devs who'd otherwise need to guess what the input would look like.

srallen:
I think `types.Date` will fail with the UTC strings. It's expecting `Date` instance (i.e. `new Date()`)

srallen:
Which is a bit annoying. We could define our own custom types to use here as well as the tutorial seen timestamp...

eatyourgreens:
@shaunanoordin that's a good idea about mocking up a Panoptes subject in the tests and passing it to a new Subject model.

eatyourgreens:
Trying out Shaun's suggestion broke the model, so I'll change the type.
```
Error: [mobx-state-tree] Error while converting `{"id":"3","already_seen":false,"favorite":false,"finished_workflow":false,"locations":[{"image/png":"https://foo.bar/kctnr.png"},{"image/png":"https://foo.bar/nzvao.png"}],"metadata":{},"retired":false,"selected_at":"2019-04-11T11:16:29.063Z","selection_state":"","user_h......:false}` to `SubjectResource`:

    at path "/selected_at" value `"2019-04-11T11:16:29.063Z"` is not assignable to type: `(Date | undefined)` (No type is applicable for the union).
    at path "/selected_at" value `"2019-04-11T11:16:29.063Z"` is not assignable to type: `(Date | undefined)` (Value is not a Date or a unix milliseconds timestamp).
    at path "/selected_at" value `"2019-04-11T11:16:29.063Z"` is not assignable to type: `(Date | undefined)` (Value is not a undefined).
```

eatyourgreens:
I've updated this to add `selected_at` to the tests too.

packages/lib-classifier/src/store/Classification.js
```diff
@@ -19,6 +19,7 @@ const ClassificationMetadata = types.model('ClassificationMetadata', {
     already_seen: types.optional(types.boolean, false),
     finished_workflow: types.optional(types.boolean, false),
     retired: types.optional(types.boolean, false),
+    selected_at: types.maybe(types.Date),
```
srallen: I think this will need to change to `types.maybe(types.string))` too?

eatyourgreens: Should be fixed now.

eatyourgreens: Should be fixed now.

PR #700, opened 2019-04-11
Update line heights for paragraphs and span texts

### LINKED ISSUE #666: Design: Field Guide

srallen:
Package:

Closes #666 

Describe your changes:
Updates the theme's line-height settings for paragraphs and span texts according to @beckyrother's input in #666 

Medium size is Grommet's default and with a font-size: 14px and line-height of 18px, I set the rest of the sizes to follow a 4px difference. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-grommet-theme/src/index.js
+++ b/packages/lib-grommet-theme/src/index.js
@@ -353,22 +353,22 @@ const theme = deepFreeze({
     },
     medium: {
       size: '14px',
-      height: '22px',
+      height: '18px',
       maxWidth: '100%'
     },
     large: {
       size: '18px',
-      height: '26px',
+      height: '22px',
       maxWidth: '100%'
     },
     xlarge: {
       size: '22px',
-      height: '30px',
+      height: '26px',
       maxWidth: '100%'
     },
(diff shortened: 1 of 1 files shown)


beckyrother:
@srallen in line 389, `medium` is 14px with 28px line-height. was that intentional? it feels like that's going to be too much, unless there's a specific place that's going to be used?

srallen:
@beckyrother That's a typo. It should be 18px line-height there. I'll open another PR to fix it. Thanks!

Issue #695, opened 2019-04-10
Design: please fix before beta launch


beckyrother:
### Tutorial / all modals
- [x] Pagination dots are too far apart, arrows are even farther. There should be 5px between each element
- [x] Between the teal bar and the top of the H2 should just be 30px. 
- [x] Why does 'Planet Finders BETA Tutorial' appear on each slide
- [x] The entire contents of the tutorial should scroll, not just the text part
- [ ] Modal text (need help, add to collections, etc) remains underlined after the modal has been closed

### Classify interface
- [x] 5px padding between 'info' etc icon and label (in #617)
- [x] 5px padding between these buttons and the bottom of the subject image: (in #617)
![Classifier_2018_Preview_Mode_-_InVision](https://user-images.githubusercontent.com/1392130/55896718-88c60400-5b84-11e9-9337-6a062577e83f.png)
- [x] Padding within the Task box should be 30px all the way around (in #617)
![Screenshot_4_10_19__11_34_AM](https://user-images.githubusercontent.com/1392130/55896802-c7f45500-5b84-11e9-90bc-37bd45a5d04c.png)
- [x] Line up the bottom of the Task module with the bottom of the subject image. (in #617) The buttons and 'need some help' should be anchored to the bottom (with 30px padding between them and the bottom of the module), additional space can be between the 'transit' button and 'need help' (in #617
![Planet_Finders_BETA___Zooniverse_-_People-powered_research](https://user-images.githubusercontent.com/1392130/55898034-61bd0180-5b87-11e9-8b57-4f9daa914ab7.png)
- [ ] When resizing the window, there's a slight lag when the data resizes to fit. Can there be an overlay with a loader to indicate that something is happening – I think this is used elsewhere in the site so if that component can be reused that would be ideal
- [x] It feels a little awkward that the background of the data is the same color as the buttons – please make it a bit darker: #003941 (better contrast for accessibility as well)
- [x] the 'tutorial' button text should be #5C5C5C while 'task' is #000 to indicate which one is selected #724
- [x] 'task' should not have a hover state, nothing happens when it's clicked #724
- [x] the tutorial doesn't scroll when it's in the tab state
![Planet_Finders_BETA___Zooniverse_-_People-powered_research](https://user-images.githubusercontent.com/1392130/55898761-e9efd680-5b88-11e9-8117-8b038e54c776.png)

### Field guide
(also in #666 but here's more)
- [x] all icons should be the same dimensions (square)
[...]

PR #704, opened 2019-04-11
Additional tutorial style fixes

srallen:
Package: lib-classifier, lib-react-components

Describe your changes:
Fixes these items from #695

- Pagination dots are too far apart, arrows are even farther. There should be 5px between each element
- Between the teal bar and the top of the H2 should just be 30px.
- Why does 'Planet Finders BETA Tutorial' appear on each slide
- The entire contents of the tutorial should scroll, not just the text part
- the tutorial doesn't scroll when it's in the tab state

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.js
@@ -33,15 +33,18 @@ function SlideTutorial (props) {
       pad={pad}
       width={width}
     >
-      {isThereMedia &&
-        <Media
-          alt={counterpart('SlideTutorial.alt', { activeStep })}
-          fit='cover'
-          height={200}
-          src={medium.src}
-        />}
-      <Heading level='3' margin='20px 0 10px 0'>{counterpart('SlideTutorial.heading', { projectDisplayName })}</Heading>
-      <StyledMarkdownWrapper aria-live='polite' autoFocus height='100%' overflow='auto'>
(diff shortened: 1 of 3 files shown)


packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.js
```diff
@@ -33,15 +33,18 @@ function SlideTutorial (props) {
...
+      <StyledMarkdownWrapper aria-live='polite' autoFocus height='100%' overflow={{ horizontal: 'hidden', vertical: 'auto' }}>
+        {isThereMedia &&
+          <Media
+            alt={counterpart('SlideTutorial.alt', { activeStep })}
+            fit='cover'
+            height={200}
+            src={medium.src}
+          />}
+        {activeStep === 0 &&
+          <Heading level='3' margin='20px 0 10px 0'>
```
rogerhutchings: we can use the named margins here! `margin={{ bottom: 'xsmall', top: 'small' }}`

https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-grommet-theme/src/index.js#L110

packages/lib-classifier/src/components/Classifier/components/SlideTutorial/components/StepNavigation/StepNavigation.js
```diff
@@ -60,7 +73,7 @@ class StepNavigation extends React.Component {
         }
       })
       return (
-        <Box as='nav' align='center' className={className} direction='row' justify='center' margin={{ top: '30px' }}>
+        <Box align='center' className={className} direction='row' justify='center' margin={{ top: '30px' }}>
```
rogerhutchings: and here

PR #711, opened 2019-04-12
Fix Next.js document so it correctly renders <Head>

### LINKED ISSUE #710: Project Header styles (and possibly other styles) aren't being processed properly
## Visual Style Issue

**Package:** `app-project`

**Issue:** When viewing `app-project` on _certain dev environments,_ the Project Header styles (and possibly other styles) aren't being processed properly. This causes weirdness such as a huge cat head staring judgementally at you, instead of a tiny cat head staring judgementally at you.

Screen Shot 2019-04-12 at 13 59 41
_Current problematic UI: Project header is unstyled_

Screen Shot 2019-04-12 at 14 10 54
_Expected UI: Project header is correctly styled_

May be related to #327, I'm not sure.

### Investigation

**Diagnosis:**
- Strangely, this is caused by changes to `app-project/pages/_document.js` in #702
- If the website is hosted on a server where `NODE_ENV` _isn't_ `production`, the GA tags are added, and that... mysteriously breaks certain styles.
- Because magic?

**Testing:**
- On the [master branch](https://github.com/zooniverse/front-end-monorepo/tree/2d97b24e7d59fc3edabf0b9a27b3024edbb7be7b), edit `app-project/pages/_document.js` so `const isProduction = process.env.NODE_ENV === 'production'` becomes `const isProduction = true` or `const isProduction = false` and compare results.
- Tests have show that when `true`, the Project Header displays fine, and when `false`, the Project Header becomes a huge scary cat face. Same can be determined by changing your NODE_ENV variable, of course.

### Status

Currently investigating _why_ adding the GA code breaks styles.

shaunanoordin:
## PR Overview

Package: `app-project`
Fixes #710 

This PR fixes an issue where - if the website is hosted on a server that _isn't_ set to `NODE_ENV=production` - then visual styling elements aren't displayed properly.

Basically, you see this...

Screen Shot 2019-04-12 at 13 59 41

...instead of this...

Screen Shot 2019-04-12 at 14 10 54

### Details

Currently, in `app-project/pages/_document.js`, the `<Head>` element isn't rendered at all if the web server's NODE_ENV isn't set to production. The lack of a Next.js `<Head>` in these instances basically screwed up style injections for the `app-project` website, causing all sorts of visual shenanigans.

Side note, this means that the website always looks fine if the dev has `NODE_ENV` set to production on their machine, or on the actual production server.

### Status

Ready for review!

### DIFF
--- a/packages/app-project/pages/_document.js
+++ b/packages/app-project/pages/_document.js
@@ -34,11 +34,11 @@ export default class MyDocument extends Document {
   render () {
     return (
       <html>
-        {isProduction && (
-          <Head>
+        <Head>
+          {isProduction && (
             <script dangerouslySetInnerHTML={{ __html: GA_TRACKING_SCRIPT }} />
-          </Head>
-        )}
+          )}
+        </Head>
         <body>
           {isProduction && (
             <noscript>


srallen:
Thanks @shaunanoordin! This is fixing the FOUC.

PR #726, opened 2019-04-16
Reduce tutorial height

eatyourgreens:
Change the tutorial height from a fixed 465px to 53vh.

Package:
lib-classifier

This is a quick fix that stops the tutorial from resizing the subject viewer.
Screenshot showing the tutorial much shorter next to the subject viewer.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


shaunanoordin:
Nyerp - I just realised that this PR affects _two_ Tutorial instances:
1. the Tutorial that pops up when the Classifier loads
2. the Tutorial that appears on the right-hand panel of the Classifier, when the user wants to re-view the Tutorial.

Both look fine on a second test, but I should have noticed the scope of the changes earlier.

eatyourgreens:
#740 reduced the height of the subject viewer, so the height set here is now too high for the available space.

PR #727, opened 2019-04-16
Add keyboard zoom to the image tools

eatyourgreens:
Add +/- keys as zoom controls when the image toolbar has focus.
Stub out handlers for the left and right cursor keys.
Add a keydown listener to the light curve viewer.

Package:
lib-classifier

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/ImageToolbar.js
+++ b/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/ImageToolbar.js
@@ -1,6 +1,5 @@
 import { Box } from 'grommet'
-import { inject } from 'mobx-react'
-import { func, shape, string } from 'prop-types'
+import { shape, string } from 'prop-types'
 import React, { Component } from 'react'
 import { withTheme } from 'styled-components'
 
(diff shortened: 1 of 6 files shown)


shaunanoordin:
## Quick Review

- At the moment, the code is using `onKeydown(e) => { e.which... }` to determine which key is pressed. [KeyboardEvent.which](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which) has been deprecated, and the recommendation is that we use `KeyboardEvent.key` instead.

  - I'll double check what's the value for "key plus" and "key minus", and see if we need to factor in any differences between physical keyboards (e.g. is the code for the plus-equals key on my number row producing the same value as the plus key on my numpad?) and browsers.

- Curiosity question: why `onKeyDown` instead of `onKeyPress`?

shaunanoordin:
Sorry about faffing over the details of KeyboardEvent - I remember going slightly crazy creating [standardised cross-browser keyboard input](https://github.com/zooniverse/anti-slavery-manuscripts/blob/master/src/lib/Utility.js) for my video games, so I think it's best we use the current standard of [keyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key).

Not [keyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code), mind you - that one returns the physical location of the keys.

shaunanoordin:
Some additional notes:

- ⚠️ do we want the Zoom In/Out keyboard actions to only work when the ImageToolbar has focus? (c.f. moving the focus to the SubjectViewer)
- Oh wait, these are _React_ KeyboardEvents, not native KeyboardEvents. Just realised this means `KeyboardEvent.code` wouldn't be available even if we wanted it.
- I noted that's little practical difference between `onKeyPress(e)` and `onKeyDown(e)` in this project.
  - Both types of key event auto-repeat when the key is held down (note: this might lead to way too much zooming if the native repeat of keys is too high) 🤔 
  - (Trivia) Strangely, `e.which` for the `+=` and `-_` keys are different between keyPress and keyDown. Uh? 🤷‍♂️ 

Technical:

- If we're using `keyboardEvent.key`, the values tested should be `'=' || '+'` for zooming in, and `'-' || '_'` for zooming out.
  - i.e. `if (e.key === '=' || e.key === '+') { zoomIn() }`
  - This has the side effect of listening to the +/- keys of the numpad, though might result in weirdness if there's a keyboard out there where `=` or `_` is paired with, say, the number keys instead of with `+` or `-` .

OK, I'm going to stop digging any further into this for now.

eatyourgreens:
Thanks @shaunanoordin! I've switched to using `event.key`. I've also rewritten the zoom handler as a decorator function, so that it can be extracted and applied to the subject viewer too.

PR #731, opened 2019-04-18
Start app-about

rogerhutchings:
Package:

Adds a new app for the about pages, which grabs team and publication data from Contentful. This is a side project I started while working on the Publications CMS hack project - the reasoning being I might as well create this (simple) app rather than make more changes to existing PFE.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


shaunanoordin:
## Quick Thoughts

package: `app-about` 🆕

- Basic checks look good - I managed to compile and run this app just fine 👍 
  - (Tested on OSX+Chrome, though it took me a second to realise the useful path to check is localhost:3000/about)
- I haven't set up Contentful for myself - and I couldn't find the login details in Passbolt yet to put into my `.env` - so I'm going to have to guess on the main functionality (listing in Publications and Team)
  - As expected, I'm getting  `Request failed with status code 404` error messages when visiting `http://localhost:3000/about/publications`
  - ❓ I'm curious to know how confident we are about the structure of the data that we're getting back from Contentful - for example, in `shared/stores/Publications.js`, we have `item.fields.project.fields.category.fields.name`. I wanted to check that, e.g., every Publication will have a `Publication.project.category` value, otherwise I'd be worried about a single Publication's `undefined` data field ruining the party for everyone.

My overall thoughts on this PR is a general 👌 though I really need to get into Contentful to provide more Meaningful feedback.

rogerhutchings:
1. My mistake - the login details were in Passbolt, but not shared. I've also updated the README with instructions on how to get the credentials.
1. Data structure is defined by the content model in Contentful, so if we break that, then yes, it'll break everything else :trollface: I've made essential fields required, however, so you can't add a Publication with a Project, for example.

shaunanoordin:
Awesome, I just pulled the details from Passbolt - thanks, Roger! 👍 The About app works perfectly now (notably Publications and Team) that I've set up `.env` with the proper digital sigils and sorcerous passkeys.

**Contentful**

Since you've already set up the Zooniverse organisation at Contentful, I'm going to try to create my own Contentful user account and invite myself to the org. 👌  Also, if you're in the office tomorrow, I'd like to check with you on whether we want to set a generic Zooniverse account as the org's owner. (This is long-term planning and not really relevant to this specific PR.)

**Data Structure**

> Data structure is defined by the content model in Contentful, so if we break that, then yes, it'll break everything else :trollface:

Ohhh nooo! 😱 

> I've made essential fields required, however, so you can't add a Publication with a Project, for example.

Ohhh that's okay then. 😐 

Still, the paranoid part of me wants to make a note of this somewhere, maybe add a note in the README that says _"lol don't mess w implicit data structure at Contentful or else u break site"_

Anyway, this PR is looking quite good for now, so please ping me when that WIP tag is done and I'll run a final review + approval. 👍

rogerhutchings:
Okay, I've added a few more tests, tidied up and whatnot - if someone can review and merge, I'll issue up the remaining things for this.

Also, I was wondering about renaming this `app-static-content`, since it would make sense to put things like Ts&Cs here

PR #743, opened 2019-04-23
Set theme via cookie and SSR correct theme mode

### LINKED ISSUE #737: Set cookie to store theme mode preference
**Package**

app-project

Following #653, the mode toggle status is lost when refreshing; probably the best way to do it is to set a cookie containing the preference, extract it from the request and use that setting to populate the store for SSR

rogerhutchings:
Package:

app-project

Closes #737

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


rogerhutchings:
Whoops - I should have added the `path` param (https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#Write_a_new_cookie)

rogerhutchings:
@shaunanoordin it should be working now

shaunanoordin:
Specifying a path, of course! 👍 

The recent changes you've added are close to fixing the issue, but the following line...

```
document.cookie = `mode=${self.mode}; path="/"; max-age=31536000`
```

...needs to be tweaked to...

```
document.cookie = `mode=${self.mode}; path=/; max-age=31536000`
```

...for full functionality. I've tested both variations on Chrome (Incognito) and discovered that `path` won't work with `"` quotes. Apparently, them cookies don't like the looks of them sensible string delimiters in this here town. 🤷‍♂️

rogerhutchings:
:+1: that's been updated

PR #355, opened 2019-01-14
[WIP] Toolbar tooltips

srallen:
Package: lib-classifier

Closes #76.

Describe your changes:
Started this a while ago, so now just finishing up. This is a refactor of both the buttons and their icons to use more of Grommet and native browser behavior and css where it makes sense. I ended up not using the `Tooltip` component I wrote a while ago and added to the react-components lib because Grommet v2 ended up releasing with a `DropButton` which almost does what we want it to except it's currently breaking keyboard interaction and I have an open issue with them about it. We may want to choose to wait for them to fix that. 

There's also a slightly odd behavior with the focus, but I'm not sure if that's related to the tabindex issue I've already reported with them. 

TO DO
- [ ] Fix any broken tests
- [ ] Wait for Grommet fix

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?
- [ ] Is the changelog updated?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?

PR #747, opened 2019-04-23
Bump Grommet

mcbouslog:
Package: fe-project, lib-classifier, lib-grommet-theme, lib-react-components

Upgrades Grommet to v2.6.6, which includes:
- fixed theme issue on grommet button
- DataTable overflow issue
- active theme support
- custom drop overflow

Might help with a few little issues we're having, at minimum good to stay current.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2186,7 +2186,7 @@
     },
     "duplexer": {
       "version": "0.1.1",
-      "resolved": "https://registry.npmjs.org/duplexer/-/duplexer-0.1.1.tgz",
+      "resolved": "http://registry.npmjs.org/duplexer/-/duplexer-0.1.1.tgz",
       "integrity": "sha1-rOb/gIwc5mtX0ev5eXessCM0z8E=",
       "dev": true
     },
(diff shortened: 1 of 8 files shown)


srallen:
I'm not sure why coveralls is now complaining about coverage dropped. I looked into the report and I don't see any files with actually decreased coverage.

srallen:
> That said, I'm not sure of which specific Grommet-theme-button issues or other UI quirks this update was intended to resolve (do we have links to issues/screenshots?), so it might be a good idea to get a second set of eyes on this.

The update isn't specific to known issues we have, but the hope is that the latest might fix the keyboard focus issue with the DropButton (being used as a tooltip) as seen in #355, despite no mention of it in their changelog as well as just trying to stay current.

mcbouslog:
@shaunanoordin it looks like you're using Chrome? interestingly, also with Chrome, I don't see the "Finished for the day?" issue.

srallen:
@mcbouslog I'm not seeing it either. @shaunanoordin if you're still seeing it, could you open an issue? Thanks!

PR #765, opened 2019-04-24
Fix font sizes and theme in ZooFooter

rogerhutchings:
This PR fixes the wrong font sizes being shown in the footer, and adds dark theme support.


packages/lib-react-components/package.json
```diff
@@ -72,6 +72,7 @@
     "grommet": "~2.6.6",
     "grommet-icons": "~4.2.0",
     "jsdom": "~13.0.0",
+    "lodash": "^4.17.11",
```
srallen: Could this use tilde instead?

Issue #746, opened 2019-04-23
Design: Mobile issues


beckyrother:
### Tutorial
- [x] add a 'get started' button on the last slide of the tutorial (also applies to the desktop version)
- [ ] the circles within the page indicators are not centered. this is because the circle is an even px width; increase the outside circle by 1px 
![IMG_4692](https://user-images.githubusercontent.com/1392130/56610134-6694a300-65d4-11e9-9283-9f407664799b.png)

---

### Interface
- [x] left align info, favourites, collections buttons to the left edge of the subject
- [x] adjust padding so it's even between buttons
- [x] Task/Tutorial tabs should be 40px tall
![IMG_4693](https://user-images.githubusercontent.com/1392130/56610004-1a496300-65d4-11e9-9346-e1432de8dc5e.png)

---

### Secondary modules
- [x] Stack 'Planet Finders Beta Statistics' and 'view more stats' buttons
       - Title left aligned, button right aligned
- [x] Add padding between 'keep track of progress' and the copy (currently lipsum)
- [x] Add padding between the content and bottom of the module
![IMG_4694](https://user-images.githubusercontent.com/1392130/56609932-fa19a400-65d3-11e9-83c1-6017d1f63ffd.png)

---

### Footer
- [x] Text should be two columns
![IMG_4695](https://user-images.githubusercontent.com/1392130/56609856-c179ca80-65d3-11e9-9556-115e035ed2ba.PNG)

srallen:
@beckyrother did the footer not collapse correctly when dragging the window to be smaller? What if you refresh the page when at the smaller size? Or was this in the dev tools emulator? I'm asking because it renders ok in the storybook and emulators for me.

beckyrother:
These are screenshots from my iPhone 6s

srallen:
@beckyrother Thanks! I'll coordinate with you to test with an actual device because I'm not seeing all these issues with the emulator.

srallen:
Footer fix in #786

srallen:
A few notes from viewing on my iPhone SE:

- Grommet's Layer in modal form is still scrollable for the tutorial, field guide, and need some help. Is this intended?

<img src="https://user-images.githubusercontent.com/5016731/56926581-adccd900-6a97-11e9-974f-60c34e78ceed.PNG" width="200px" />
[...]

PR #767, opened 2019-04-24
Tutorial get started button

srallen:
Package: lib-classifier

Adds the get started button requested for in #746 and #695. Is this ok @beckyrother?

Describe your changes:
- The last slide shows the get started button
- The click handler varies depending on the context, modal vs tabbed
- The tabs had to be refactored to be controlled so the active tab index could be set by the button's click handler
- Adds and updates tests

Note this does not update the height of the buttons as requested in in #695. I'll update that separately.

How it looks:

Tabbed:

Screen Shot 2019-04-24 at 1 48 20 PM

Modal:

Screen Shot 2019-04-24 at 1 49 21 PM

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/components/Classifier/components/ModalTutorial/ModalTutorial.js
```diff
@@ -23,21 +23,26 @@ function storeMapper (stores) {
...
   render () {
-    const { loadingState, showModal, setModalVisibility, tutorial } = this.props
+    const { loadingState, showModal, tutorial } = this.props
     if (loadingState === asyncStates.success && tutorial) {
       return (
         <Modal
           {...this.props}
           active={showModal}
-          closeFn={() => { setModalVisibility(false) }}
+          closeFn={this.onClose.bind(this)}
```
eatyourgreens: Minor point, but passing on some advice from @rogerhutchings: this might create a new bound function on each call to `render` so I've been switching back to binding event handlers in the component constructor.

packages/lib-classifier/src/components/Classifier/components/ModalTutorial/ModalTutorial.js
```diff
@@ -23,21 +23,26 @@ function storeMapper (stores) {
...
           title={counterpart('ModalTutorial.title')}
         >
           <ResponsiveContext.Consumer>
             {size => {
               const width = (size === 'small') ? '100%' : '330px'
               return (
-                <SlideTutorial pad='none' width={width} />
+                <SlideTutorial onClick={this.onClose.bind(this)} pad='none' width={width} />
```
eatyourgreens: Styling: you sometimes list props, sometimes write them inline like here. Should they always be listed vertically?

packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.js
```diff
@@ -41,29 +52,37 @@ function SlideTutorial (props) {
...
+        {isFirstStep &&
           <Heading level='3' margin={{ bottom: 'xsmall', top: 'small' }}>
             {counterpart('SlideTutorial.heading', { projectDisplayName })}
           </Heading>}
         {/* TODO: translation */}
         <Markdownz>{step.content}</Markdownz>
       </StyledMarkdownWrapper>
       <StepNavigation />
+      {isLastStep &&
+        <Button label={counterpart('SlideTutorial.getStarted')} onClick={onClick} margin={{ top: 'medium' }} primary />}
```
eatyourgreens: List props here too? The linter might fix this, I'm not sure.

packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.spec.js
```diff
@@ -44,4 +46,25 @@ describe('SlideTutorial', function () {
     const wrapper = shallow(<SlideTutorial stepWithMedium={{ step, medium }} />)
     expect(wrapper.find(Media)).to.have.lengthOf(1)
   })
+
+  it('should only render the Heading on the first step', function () {
+    const wrapper = shallow(<SlideTutorial activeStep={0} stepWithMedium={{ step, medium }} isFirstStep />)
+    expect(wrapper.find(Heading)).to.have.lengthOf(1)
+    wrapper.setProps({ activeStep: 1, isFirstStep: false })
```
eatyourgreens: Minor point about the component API, but it seems weird to me to that it's possible to pass in a combination like `activeStep: 1, isFirstStep: true`.

packages/lib-classifier/src/store/TutorialStore.js
```diff
@@ -78,6 +78,23 @@ const TutorialStore = types
 
       if (miniCourse) return lastStepSeen === miniCourse.steps.length - 1
 
+      return false
+    },
+
+    get isFirstStep () {
+      if (!!self.active && !!self.activeStep.toString()) {
```
eatyourgreens: Why the `!!` here? I thought the if statement would automatically treat the conditionals as booleans? Is it important to actually cast them to Boolean?

eatyourgreens: Oh wait, is this catching a case where `self.activeStep.toString()` returns '0' and JavaScript evaluates that as false? I'm confused as to what the conditional is actually checking, and why `self.active` isn't a boolean by default.

packages/lib-classifier/src/store/TutorialStore.js
```diff
@@ -78,6 +78,23 @@ const TutorialStore = types
...
+        return self.activeStep === 0
+      }
+
+      return false
+    },
+
+    get isLastStep () {
+      if (!!self.active && !!self.activeStep.toString()) {
```
eatyourgreens: Same comments about the conditional here. I don't quite follow what it's checking.

packages/lib-classifier/src/store/TutorialStore.js
srallen: Yes, you're correct about the '0' issue in javascript. I need it to treat '0' as truthy in this case. However, I don't remember why I was casting `self.active` to boolean here. Let me try removing it... but I think I vaguely remember doing it because I found a scenario where it could fail in a test without doing this. I'll confirm and can leave a comment when I re-figure it out again.

packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.spec.js
srallen: Yeah, I see how that is a bit confusing. The activeStep prop is only used in constructing the aria-label for the radio buttons to be read aloud "Go to step 3" for instance.

packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.js
srallen: I definitely forgot to run the linter. Will do.

packages/lib-classifier/src/components/Classifier/components/ModalTutorial/ModalTutorial.js
srallen: Oh, good point. I thought this was the preferred way according to the React docs, but I just re-read it and it does point out the potential performance issue here. Binding inline is only necessary to pass arguments. Thanks.

packages/lib-classifier/src/components/Classifier/components/ModalTutorial/ModalTutorial.js
rogerhutchings: I tend to list them vertically when there are 3+ / some are objects

packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskArea.js
```diff
@@ -24,17 +24,37 @@ function storeMapper (stores) {
...
+    this.setState({ activeIndex })
   }
 
   render () {
     const { disableTutorialTab } = this.props
 
     return (
       <Tabs
+        activeIndex={this.state.activeIndex}
         className={this.props.className}
```
rogerhutchings: Not related to this PR, but I don't think there's a `className` prop to pass as we're not wrapping anything in a styled component: https://www.styled-components.com/docs/advanced#styling-normal-react-components

packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.spec.js
srallen: Also the alt text for the media, "Tutorial step 3", which I'm not sure how useful that is.

packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskArea.js
srallen: Might be cruft from a previous version of the tabs?

packages/lib-classifier/src/store/TutorialStore.js
srallen: I can't determine why I did that, so I'll remove it from `self.active`

Issue #759, opened 2019-04-24
Dark theme: Next, Done and Done & Talk dark theme styles


srallen:
**Package** lib-grommet-theme, lib-classifier

@rogerhutchings I'm assigning you since you've done the dark theme work most recently.

**Feature or Issue Description**
The dark theme styles mean switching to a outline border style for the task navigation and slight color changes:

<img width="396" alt="Screen Shot 2019-04-24 at 10 08 02 AM" src="https://user-images.githubusercontent.com/5016731/56670662-e2482b80-6678-11e9-941a-1018f9569d95.png">
<img width="397" alt="Screen Shot 2019-04-24 at 10 08 38 AM" src="https://user-images.githubusercontent.com/5016731/56670702-f2f8a180-6678-11e9-8c7a-bf689859e3c1.png">

How they appear in the new app:

<img width="387" alt="Screen Shot 2019-04-24 at 10 09 27 AM" src="https://user-images.githubusercontent.com/5016731/56670827-1e7b8c00-6679-11e9-9b03-aa304d5caf68.png">
<img width="389" alt="Screen Shot 2019-04-24 at 10 09 51 AM" src="https://user-images.githubusercontent.com/5016731/56670832-21767c80-6679-11e9-88a4-737cbcaa7cf2.png">

@beckyrother any changes here or should we replicate the current implementation?

beckyrother:
Sorry which is which in those images?

srallen:
@beckyrother first two is how it is in current PFE. The next two is how it is with the new app: https://fe-project.preview.zooniverse.org/projects/nora-dot-test/planet-finders-beta?env=staging (it's using grommet defaults I think)

beckyrother:
Got it. Please continue to use the outline versions with the font size that's shown, they're more legible.

PR #769, opened 2019-04-25
Style `Next` button in Classifier

rogerhutchings:
cc #759

Fixes the styling for the `Next` button in the classifier task area.

Light:

Screenshot 2019-04-25 10 27 15

Light + hover:

Screenshot 2019-04-25 10 27 27

Dark:

Screenshot 2019-04-25 10 27 45

Dark + hover:

Screenshot 2019-04-25 10 27 37

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButton.js
```diff
@@ -1,98 +1,50 @@
...
+)
 
-function NextButton ({ autoFocus, disabled, classifierTheme, onClick }) {
+function NextButton (props) {
+  const { autoFocus, disabled, onClick, theme } = props
+  const mergedThemes = merge({}, theme, nextButtonTheme)
   return (
-    <ThemeProvider theme={{ mode: classifierTheme }}>
-      <StyledNextButton
+    <Grommet theme={mergedThemes}>
```
srallen: I was looking over the Grommet docs again today and maybe we should be using `ThemeContext` instead of the Grommet component instead to be doing stuff like this? https://v2.grommet.io/themecontext

rogerhutchings: Done :+1:

rogerhutchings:
I much prefer that method, creating a whole new Grommet context feels heavy-handed. 

It was mostly painless, although I did have to do a separate export for testing; I was getting errors about `colors` not being defined, and without digging further I'm guessing that trying to use `ThemeContext` outside of a Grommet context is breaking.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButton.spec.js
```diff
@@ -2,9 +2,10 @@ import React from 'react'
 import { shallow } from 'enzyme'
 import { expect } from 'chai'
 import sinon from 'sinon'
-import NextButton, { StyledNextButton } from './NextButton'
 
-describe('NextButton', function () {
+import { NextButton } from './NextButton'
+
+describe.only('NextButton', function () {
```
eatyourgreens: Stray `.only`.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButton.js
```diff
@@ -1,98 +1,50 @@
...
-  })};
-  }
-
-  &:disabled {
-    cursor: not-allowed;
-  }
-  `
+const Label = (
+  <Text size='medium'>
+    {counterpart('NextButton.next')} →
```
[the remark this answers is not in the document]
rogerhutchings: Good idea: I've used CSS content and added a prop to determine where and which arrow to insert. I've also added a `TODO` to remind us that we'll need to hook this up to something :)

rogerhutchings:
I'm going to extract the ThemeContext HOC out to its own component in the Tabs PR, as it seems like it's a pretty common use

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButton.js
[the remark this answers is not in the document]
eatyourgreens: Thanks, but the arrow should always be after the word Next. You don't need to add logic to change the order. Your change would always display the arrow on the right of the word Next, but it should be on the left in an RTL language.

rogerhutchings:
You're right, I've just tried it out and pushed a fix.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButton.js
```diff
@@ -1,98 +1,56 @@
...
-    light: `solid thin ${adjustHue(-7, gold)}`
-  })};
-    box-shadow: none;
-    color: ${theme('mode', {
-    dark: 'black',
-    light: 'black'
-  })};
-  }
+// TODO: This will need hooking up to counterpart / a helper function to
+// determine whether to use an RTL language or not
```
eatyourgreens: In PFE, we keep a list of RTL languages and check the stored locale against that.
https://github.com/zooniverse/Panoptes-Front-End/blob/2f7c55797bc9837ae5f0e6407d7c28d78b7bbe9b/app/redux/ducks/translations.js#L57-L60

A CSS selector like `html[lang=he]` handles most of the basic project display though.

rogerhutchings: ooh, there's also https://developer.mozilla.org/en-US/docs/Web/CSS/:lang (didn't know about this one)

eatyourgreens: That's new to me too. It'd be really nice if we didn't have to maintain a big list of RTL language codes, but could let the browser do that for us.

PR #779, opened 2019-04-26
Refactor metadata modal to be contained in a fragment container class

srallen:
Package: lib-classifier

Describe your changes:
This fixes the extra gap Grommet was adding. Even though the modal was hidden, Grommet was putting a gap after it. So now instead of:

Screen Shot 2019-04-26 at 10 23 04 AM

It renders as:

Screen Shot 2019-04-26 at 10 22 29 AM

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/components/Classifier/components/MetaTools/components/Metadata/Metadata.js
```diff
@@ -0,0 +1,45 @@
...
+
+    this.state = {
+      showMetadataModal: false
+    }
+
+    this.toggleMetadataModal = this.toggleMetadataModal.bind(this)
+  }
+
+  toggleMetadataModal () {
+    this.setState((prevState) => { return { showMetadataModal: !prevState.showMetadataModal } })
```
rogerhutchings: this can be shortened for es6 style points:

```
this.setState(prevState => ({ showMetadataModal: !prevState.showMetadataModal }))
```

PR #780, opened 2019-04-26
Update theme for the radio buttons

srallen:
Package: lib-grommet-theme

Describe your changes:
Fixes the regression to the circle size after #747. I'm not sure if this fixes the issue in #746 because I can't replicate it using the emulators. We'll have to start testing with actual devices. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-grommet-theme/src/index.js
+++ b/packages/lib-grommet-theme/src/index.js
@@ -426,7 +426,7 @@ const theme = deepFreeze({
   radioButton: {
     icon: {
       size: '15px',
-      extend: props => `circle { r: 10px; }`
+      extend: 'circle { r: 10px; }'
     },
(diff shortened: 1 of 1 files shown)


packages/lib-grommet-theme/src/index.js
```diff
@@ -425,7 +425,8 @@ const theme = deepFreeze({
   },
   radioButton: {
     icon: {
-      size: 'small'
+      size: '15px',
+      extend: props => `circle { r: 10px; }`
```
rogerhutchings: Not a blocker, but this doesn't need to be a function, as it's not using `props` - it can just be:

```
extend: 'circle { r: 10px; }'
```

PR #784, opened 2019-04-29
Set default padding for button in the theme

### LINKED ISSUE #772: Design: Button height
**Package** lib-classifier, app-project

Originally from #695 

- [ ] Check button height, all should be 40px tall

srallen:
Package: lib-grommet-theme

Closes #772

Describe your changes:
Sets the default padding for the button component in the theme.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-grommet-theme/src/index.js
+++ b/packages/lib-grommet-theme/src/index.js
@@ -168,6 +168,9 @@ const theme = deepFreeze({
   button: {
     border: {
       radius: '0px'
+    },
+    padding: {
+      vertical: '9px'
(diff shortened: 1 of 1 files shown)


rogerhutchings:
I'm not sure I'd worry about trying to account for border width, especially when the designs are all using single pixel borders, but this is :+1:

srallen:
I did it since the original Grommet base theme does it and since we haven't overwritten all the settings, there may be styles still being applied assuming that we're doing that.

Issue #758, opened 2019-04-24
Font size style regression


rogerhutchings:
It looks like without a specific prop of `size='medium'`, font sizes in text-based components are reverting to `18px`

srallen:
Grommet is showing the default for Paragraph is `medium`: https://v2.grommet.io/paragraph

But doesn't show a default for Text (the docs just might not be right). Is this happening with Paragraph or Text? Or places where we're missing using either?

rogerhutchings:
List items in the footer, for example, and content box headings etc are all rendering smaller (12px), weirdly. I can't remember which component they're using off-hand, but I wonder if a global default has been accidentally set

srallen:
Was this fixed in #765?

srallen:
As noted in #786, the 'small' sizes in the theme for text, paragraph, and headings should use a line-height of 14px. The footer section headings need to be updated to use 'small'.

PR #786, opened 2019-04-29
Footer in two columns on mobile view

srallen:
Package: lib-react-components

Fix for #746

Describe your changes:
Makes footer two columns:
IMG_3492

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


srallen:
@rogerhutchings I see them have a line-height of 18px for a font-size of 14px, which are the base medium sizes set in the theme. I used @beckyrother guidance in this doc for these: https://projects.invisionapp.com/d/main#/projects/prototypes/12923997/assets/77727645

According to that doc, it should be 14px/17px. @beckyrother do you want the pixel decrease for the line-height in the theme? Do you want the footer to use these base sizes or something else?

rogerhutchings:
I'm sure they're accurate to the design, but look at them - the gap between lines in the section titles is almost the same size as the gap between section items.

beckyrother:
Agreed @rogerhutchings, they're a bit tall. Can we adjust the style to be 14px / 14px line-height? This would be a global change for that type style.

srallen:
Ok, I'm not sure where we should set that in the theme...

@rogerhutchings we could make the small paragraph setting 14px/14px? 

https://github.com/zooniverse/front-end-monorepo/blob/c177dd59f5f9d47ba16ee7932542f3db44613bcd/packages/lib-grommet-theme/src/index.js#L372-L375

beckyrother:
This is for the teal labels right?

srallen:
@beckyrother they're anchors wrapped by spans, which is reusing the theme's defaults for 'medium' sized paragraphs and spans. We could set the 'small' spans/paragraph's defaults to 14px/14px. Then any paragraph or span (and child components) we set to `size="small"` will be this.

rogerhutchings:
@srallen heading levels can be set separately: https://v2.grommet.io/heading

srallen:
Yep, headings have their sizes defined in the theme too, but those aren't headings in the markup. They could be however. I'll make notes on what to change in the issue for #758 and let's keep this as just the grid update for the footer.

srallen:
I should note that the collision should be fixed and this is ready for another review.

beckyrother:
Is there a way to see this live? I'm sorry I definitely should know this by now.

srallen:
@beckyrother if you're willing to pull down the repo and run locally, but we don't have staged deployments setup for this yet. Or you could come bug me and take a look on my screen. We'll try to get staged deployments working soon!

srallen:
Oh I should add, the react components repo we want to deploy the storybook to github pages, so that'd be another way to take a look, but that hasn't been done yet either.

PR #464, opened 2019-02-15
Your project activity

eatyourgreens:
Package:
app-project

Towards #38.

Add a YourStats store. Request user prefs and daily classification counts on initial user login.
Test that the store initialises, fetches activity counts on user change and increments counts properly for both authenticated and anonymous users.
Actions:
- `yourStats.fetchActivityCount()` gets your total classification count from your user preferences.
- `yourStats.fetchDailyCounts()` fetches your daily classification counts from the stats service
- `yourStats.increment()` increments your classification count for this session.

Todo:
- [x] tests.
- [ ] Your Progress React component
- [ ] increment classification counts when the classifier completes a classification

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?
- [ ] Is the changelog updated?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?

Issue #38, opened 2018-06-07
"Your statistics" widget


rogerhutchings:
<img width="376" alt="screenshot 2018-06-07 16 50 12" src="https://user-images.githubusercontent.com/2725841/41111178-30fa07e2-6a73-11e8-9644-416390faf92d.png">

rogerhutchings:
@camallen This is the widget that would use the new stats API

eatyourgreens:
Those day names on the x-axis are going to need to be localised, aren't they?

eatyourgreens:
Sorry to be asking dumb questions but is this a modal box or is it displayed somewhere on the classify page?

srallen:
It's further down the page. Here are the redesign in invision: https://projects.invisionapp.com/d/main#/projects/prototypes/12923997

rogerhutchings:
On the classify page, between "Finished for the Day" and "Project Statistics":

<img width="678" alt="screenshot 2019-01-16 16 15 14" src="https://user-images.githubusercontent.com/2725841/51262272-f5789a80-19a9-11e9-87db-a2bc7e92b2cc.png">

camallen:
Noting there that the counts of volunteer activity (daily, total, etc) will ideally come from the new graphql stats API (currently down for db maintenance) instead of the old stats API (planned deprecation this year).

eatyourgreens:
Is this high priority for the TESS launch, or something that can be added once we have a working classify page up and running on a server?

Can the stats API send back day-of-the-week with the daily counts? I'm thinking about how the histogram might be rendered with the days along the x-axis for the last 7 days (or is it this week with future days blank if I'm classifying on Tuesday say?)

camallen:
> Can the stats API send back day-of-the-week with the daily counts? I'm thinking about how the histogram might be rendered with the days along the x-axis for the last 7 days (or is it this week with future days blank if I'm classifying on Tuesday say?)

Stats API will return the 'interval' (daily by default) histogram for the past only (all data i believe - a limit on this would be nice and easily supported i think). 

The new stats api code needs docs on the API itself and how to use it but you can see the code here https://github.com/zooniverse/zoo-stats-api-graphql/blob/9382005d6d2e33cd864fe8fe00125e292e13fb1f/app/graphql/types/query_type.rb#L7

eatyourgreens:
[...]

PR #787, opened 2019-04-30
Start the stats UI component 🎉

rogerhutchings:
Part of #38, this starts the user-facing components for the Your Stats widget:

Logged in:

Screenshot 2019-04-30 16 17 48

Logged out:

Screenshot 2019-04-30 16 18 02

On reviewing #464, it looks like the idea is to support some statistics for anonymous users, is that right? If so, then the "log in" message can be moved to the histogram, as I've written it as a HOC.

- Moves around some components to make them easier to reuse
- Adds login message HOC
- Adds basic component for showing stats

For future PRs:

- add the histogram
- wire up to the store


packages/app-project/src/shared/components/withRequireUser/withRequireUser.js
```diff
@@ -0,0 +1,49 @@
...
+import React, { Component } from 'react'
+
+import en from './locales/en'
+
+counterpart.registerTranslations('en', en)
+
+function storeMapper (stores) {
+  const { isLoggedIn } = stores.store.user
+  // return { isLoggedIn }
+  return { isLoggedIn: false }
```
shaunanoordin: This line forces the YourStats component to remain un-logged-in - is this a WIP?

I tested the code by removing `return { isLoggedIn: false }` and re-enabled `return { isLoggedIn }`, and that seems to get the YourStats component working properly.

rogerhutchings: WOOPS yeah that's totally left in from debug

rogerhutchings:
Yes, there's going to be another component on the right for Recently Viewed Subjects

PR #790, opened 2019-05-01
Mobile view design improvements

srallen:
Package: lib-grommet-theme, lib-classifier, app-project

For #746 

Describe your changes:
This PR fixes the following from #746 
- left align info, favourites, collections buttons to the left edge of the subject
- Task/Tutorial tabs should be 40px tall
- Stack 'Planet Finders Beta Statistics' and 'view more stats' buttons
- Add padding between 'keep track of progress' and the copy (currently lipsum)
- Add padding between the content and bottom of the module

I also stacked the 'Finished for the Day' image and text as a column.

Most of this was achieved by defining the `edgeSize` in the theme under the small breakpoint. Grommet switches to use these sizes for margins, paddings when the ResponsiveContext size becomes `'small'`

The sizes I picked to add to the theme here are arbitrary, so I'm open to any adjustments. Note Grommet defaults to the 'medium' size:
```
    breakpoints: {
      small: {
        edgeSize: {
          xxsmall: '5px',
          xsmall: `10px`,
          small: `15px`,
          medium: `20px`,
          large: `25px`,
          xlarge: `30px`,
        }
      }
    }
```

Prior to this change, `medium` was set in the base theme to be 12px, which was too small.

Here are screenshots of the results:

Screen Shot 2019-05-01 at 10 23 29 AM
Screen Shot 2019-04-30 at 3 43 59 PM
Screen Shot 2019-04-30 at 3 44 14 PM
Screen Shot 2019-04-30 at 3 44 21 PM

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


srallen:
There are a few broken tests that I still need to fix.

packages/app-project/src/screens/ClassifyPage/components/FinishedForTheDay/FinishedForTheDayContainer.js
```diff
@@ -19,11 +20,16 @@ class FinishedForTheDayContainer extends Component {
   render () {
     const { imageSrc, isLoggedIn, projectName } = this.props
     return (
-      <FinishedForTheDay
-        imageSrc={imageSrc}
-        isLoggedIn={isLoggedIn}
-        projectName={projectName}
-      />
+      <ResponsiveContext.Consumer>
```
rogerhutchings: Is it worth creating a HOC for ResponsiveContext, so we can export the unwrapped component and preserve the tests?

packages/app-project/src/shared/components/ProjectStatistics/components/Subtitle/Subtitle.js
```diff
@@ -1,12 +1,16 @@
...
 
-function Subtitle ({ mode, text, ...props }) {
+function Subtitle ({ margin, mode, text, ...props }) {
   return (
-    <Text {...props}>
-      {text}
-    </Text>
+    <ResponsiveContext.Consumer>
+      {(size) => (
+        <Text {...props} margin={(size === 'small') ? { top: 'none', bottom: 'xsmall' } : margin}>
```
rogerhutchings: Instead of changing props based on size, couldn't we change the values of `global.breakpoints.small.edgeSize` to be smaller so they're applied across the theme?

shaunanoordin:
## Quick Review
- Changes LGTM as far as the code goes,
- but UI-wise/visually, I'm seeing some quirks that would be best reviewed by a designer.
- Most notably is this: _ info, favourites, collections buttons are no longer left-aligned to the left edge of the subject._ (as a result of master-merges maybe?)
  Screen Shot 2019-05-07 at 15 35 57
- Snapshots and full review to follow

packages/app-project/src/screens/ClassifyPage/components/FinishedForTheDay/FinishedForTheDayContainer.js
srallen: Sure, but I would prefer as a separate update.

packages/app-project/src/shared/components/ProjectStatistics/components/Subtitle/Subtitle.js
srallen: This PR includes an update to the global small edge sizes to increase them slightly. This is the only case where it needed to vary. If we make the size smaller again, there's a cascading effect of the margins and paddings not being enough for other components.

beckyrother:
Definitely those buttons (info, favourites, etc) need to be left-aligned and with much less padding between. Should we remove the project name just on small screens to give the subject image a little more space?

srallen:
@beckyrother I was playing around with what to do with the dark theme toggle and the project name and I was going to suggest removing the project name from small sizes, however, we also discussed potentially using this area for displaying the workflow name instead, so do you have any ideas on positioning? I was trying out a column stacking, something like this (pardon the basic ascii-like drawing here):

```
-------------
header
-------------
project-name
-------------
|           |
|   subject |
|   viewer  |
-------------
meta-tools
dark toggle
-------------
```

But the project name there is very redundant because you see it in the header area already, but if this is the workflow name instead, then it should remain somewhere, I'm just not sure where.

srallen:
That being said, I'll probably update the layout for project name and dark toggle in a different PR to keep this one smaller

beckyrother:
@srallen that ascii layout looks good, but I don't think we'll need the project name. Let's take that out and just move the dark mode to below the subject so there's more space.

srallen:
> left-aligned Subject Info, Favourites, and Collections 

@shaunanoordin it's not just on your screen. I see it too. I broke it somehow or a regression came in with a rebase. I'm fixing it now. 

As for IE, don't worry bout it. We're dropping support for it. And Edge is moving to the Chromium engine very soon... for better or worse.

srallen:
@shaunanoordin the alignment of the metatools buttons should be fixed now, however, it required a refactor to create an underlying shared button component to make sure the styling was applied to all. It also provided opportunity to update how the theme is applied to the newer way of doing it (using `withTheme` HOC instead of `styled-theming` lib). I also updated and adding missing tests.

srallen:
Note that there are a couple of unrelated components that got the linter update.

shaunanoordin:
The updates are looking good! 👍 Thanks for fixing the alignment of those metatool buttons; as far as I can tell, that's everything that this PR aimed to address?

Becky, if you want to see the latest:
- [Classify page on Chrome, emulating iPhone X view](https://user-images.githubusercontent.com/13952701/57320545-0daa2c00-70f7-11e9-8ab9-e930069a53fe.png)
- [Classify page on Firefox, emulating iPhone X view](https://user-images.githubusercontent.com/13952701/57320547-0daa2c00-70f7-11e9-8a97-eefc8ef348e9.png)

packages/app-project/src/screens/ClassifyPage/components/FinishedForTheDay/FinishedForTheDayContainer.js
srallen: Actually I ended up adding a HOC to the react-components library much like the one for ThemeContext because I might use it again to fix the last padding issue.

srallen:
@beckyrother is this ok for spacing?
Screen Shot 2019-05-08 at 12 24 25 PM

@shaunanoordin and @rogerhutchings this is ready for final review. I've made some modifications to the react-components including updating PlainButton and adding the withResponsiveContext HOC helper. I've updated and added tests for the MetaTools component.

packages/lib-classifier/src/components/Classifier/components/Layout/components/DefaultLayout/DefaultLayout.js
```diff
@@ -25,6 +25,7 @@ const ContainerGrid = styled.div`
     grid-template-columns: 100%;
     grid-template-rows: auto auto;
     grid-auto-flow: column;
+    grid-gap: 1rem;
```
rogerhutchings: IIRC we're using multiples of `10px` for spacing in the theme?

packages/app-project/src/shared/components/ProjectStatistics/components/Subtitle/Subtitle.js
```diff
@@ -1,10 +1,11 @@
 import { Text } from 'grommet'
+import { withResponsiveContext } from '@zooniverse/react-components'
 import { object, oneOfType, string } from 'prop-types'
 import React from 'react'
 
-function Subtitle ({ mode, text, ...props }) {
+function Subtitle ({ margin, mode, screenSize, text, ...props }) {
   return (
-    <Text {...props}>
+    <Text {...props} margin={(screenSize === 'small') ? { top: 'none', bottom: 'xsmall' } : margin}>
```
rogerhutchings: I'd recommend moving this ternary out of the JSX for legibility, as you've done later on

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/CollectionsButton/CollectionsButton.js
```diff
@@ -1,47 +1,36 @@
...
-      <PlainButton
-        disabled={disabled}
-        icon={<Collect />}
-        margin={{ vertical: '5px', horizontal: 'none' }}
-        text={counterpart('CollectionsButton.add')}
-        onClick={disabled ? undefined : onClick}
-      />
-    </ThemeProvider>
+    <MetaToolsButton
+      disabled={disabled}
```
rogerhutchings: Do we need a `className` here?

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/FavouritesButton/FavouritesButton.js
```diff
@@ -1,54 +1,43 @@
...
 
 counterpart.registerTranslations('en', en)
 
 export const Favourite = styled(HeartIcon)`
-  fill: ${props => props.filled ? '#E45950' : 'none'};
-  stroke: ${theme('mode', {
-    dark: zooTheme.global.colors.text.dark,
-    light: zooTheme.global.colors.text.light
-  })};
+  fill: ${props => props.filled ? zooTheme.global.colors.statusColors.error : 'none'};
```
rogerhutchings:
```
fill: ${props => props.filled ? props.theme.global.colors.statusColors.error : 'none'};
```

...and then you can remove the `zooTheme` import

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/FavouritesButton/FavouritesButton.js
```diff
@@ -1,54 +1,43 @@
...
   width: 1em;
 `
 
 export default function FavouritesButton (props) {
-  const { checked, disabled, onClick, theme } = props
+  const { checked, disabled, onClick } = props
```
rogerhutchings: `className`?

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/MetaToolsButton/MetaToolsButton.js
```diff
@@ -0,0 +1,51 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import styled, { withTheme } from 'styled-components'
+import { PlainButton } from '@zooniverse/react-components'
+import zooTheme from '@zooniverse/grommet-theme'
+
+export const StyledPlainButton = styled(PlainButton)`
+  > div {
+    justify-content: flex-start;
+    line-height: ${zooTheme.paragraph.small.height};
```
rogerhutchings: You should have access to the `theme` via the prop again here:

```
line-height: ${props.theme.paragraph.small.height};
```

packages/lib-react-components/src/PlainButton/PlainButton.js
```diff
@@ -20,8 +18,8 @@ export const StyledPlainButton = styled(Button)`
 `
 
 function PlainButton (props) {
-  const { onClick, text,  } = props
-  // const labelColor = determineColor(theme)
+  const { onClick, labelSize, text } = props
```
rogerhutchings: `className`?

packages/lib-react-components/src/helpers/withResponsiveContext/README.md
```diff
@@ -0,0 +1,18 @@
...
+wrapped in Grommet's `ResponsiveContext` component, with the size passed into
+wrapped component's `size` prop.
+
+## Usage
+
+```js
+
+class MyComponent extends React.Component {
+  render () {
+    ...
```
rogerhutchings: A quick example might be handy, especially to illustrate the fact that `ResponsiveContext` returns breakpoint names rather than pixel values

packages/lib-classifier/src/components/Classifier/components/Layout/components/DefaultLayout/DefaultLayout.js
srallen: I used rem here since the layout is already using rem, but I can switch to pixels. I'd like to refactor it to use the Grommet Grid here at a later point, so we don't have to manually set this.

Issue #304, opened 2018-11-27
Add subject dimensions to classification metadata


srallen:
**Package** lib-classifier

**Feature or Issue Description**
This was left out during development because of #201. The subject's natural width and height as well as client width and height should be saved. The metadata should be updated on load of the image: https://github.com/zooniverse/Panoptes-Front-End/blob/6e1ed81cd338ec9e3d782318807a69607219d627/app/classifier/classifier.jsx#L200-L201

@shaunanoordin For non-image subject types like data, should we record the dimensions of the plot displayed?

shaunanoordin:
If by "non-image subject types", we mean the TESS light curve data, then I'm inclined to say the information is unnecessary.

My rationale is that knowing the client width/height of a subject only matters if we're trying to understand users' classification behaviour, and if that were the case, what we'd need to focus on is tracking how the user uses the Light Curve Viewer's _zoom_ tools. That said, I can easily be swayed on this position - adding the dimensions of the plot on Classification submit should be fairly trivial, and only cost us a few extra bytes per Classification in the long run.

I'd best pass this question to @noraeisner, in case there is some potential research question about how the _size of the data, as displayed in the browser window,_ affects user accuracy/etc.

srallen:
@shaunanoordin Did we have a definitive answer about this for TESS?

shaunanoordin:
I recall that the answer was _"no, we don't need browser metadata",_ but I'm pinging @noraeisner to confirm. I should have documented this, sorry.

noraeisner:
I'm inclined to say that if it's easy to store this information then we should (just in case we do ever want to use it) but if it's not easy to store then it's definitely not essential.

shaunanoordin:
👍

eatyourgreens:
Thinking about it, natural height/width aren't relevant for SVG, since it can be scaled to any size. Client height and width should be available from the SVG DOM node or its containing HTML element.

eatyourgreens:
That said, we record viewport dimensions so we've already got a measure of what size screen the classification was made on, which might be the important number for data analysis.

PR #791, opened 2019-05-01
Disable task area while subject is loading

### LINKED ISSUE #511: Tasks should be disabled while subject is loading
**Package**
lib-classifier

Testing the classifier with the network disconnected,I found that you can continue to classify even if the subject area is displaying a broken image icon.

For image subjects, the task area should be disabled/hidden while the image file is loading.

eatyourgreens:
- [x] Add a `ready` flag ro the subject viewer store. Reset it to false for each new subject. Add an `onReady` handler to subject viewers, which fires on load for images, or after JSON data has loaded and parsed for data subjects.
- [x] Add a disabled prop to task nav buttons.
- [x] Add a disabled prop to tasks.
- [x] Disable tasks while subject data is loading.

Package:
lib-classifier

Closes #511 .

Describe your changes:

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
@@ -63,31 +63,41 @@ class LightCurveViewerContainer extends Component {
       } else {
         // Get the JSON data, or (as a failsafe) parse the JSON data if the
         // response is returned as a string
-        this.setState({ loading: asyncStates.success })
         return response.body || JSON.parse(response.text)
       }
     } catch (error) {
-      this.setState({
-        loading: asyncStates.error,
-        data: null
-      })
-      return error
+      return this.onError(error)
     }
   }
 
   async handleSubject () {
     try {
       const rawData = await this.requestData()
-      this.setState({
-        dataExtent: {
-          x: d3.extent(rawData.x),
-          y: d3.extent(rawData.y)
-        },
-        dataPoints: zip(rawData.x, rawData.y)
-      })
+      this.onLoad(rawData)
+      
     } catch (error) {
       console.error(error)
     }
+  }
+
+  onLoad (rawData) {
+    this.setState({
+      dataExtent: {
+        x: d3.extent(rawData.x),
+        y: d3.extent(rawData.y)
+      },
+      dataPoints: zip(rawData.x, rawData.y),
+      loading: asyncStates.success
+    })
+    this.props.onReady()
+  }
+
+  onError (error) {
+    this.setState({
+      loading: asyncStates.error,
+      data: null
+    })
+    return error
   }
 
   render () {
@@ -108,7 +118,12 @@ class LightCurveViewerContainer extends Component {
   }
 }
 
+LightCurveViewerContainer.defaultProps = {
+  onReady: () => true
+}
+
 LightCurveViewerContainer.propTypes = {
+  onReady: PropTypes.func,
   subject: PropTypes.shape({
     locations: PropTypes.arrayOf(locationValidator)
   }),
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
@@ -81,6 +81,7 @@ class LightCurveViewerContainer extends Component {
   }
 
   onLoad (rawData) {
+    const { onReady } = this.props
     this.setState({
       dataExtent: {
         x: d3.extent(rawData.x),
@@ -88,8 +89,8 @@ class LightCurveViewerContainer extends Component {
       },
       dataPoints: zip(rawData.x, rawData.y),
       loading: asyncStates.success
-    })
-    this.props.onReady()
+    },
+    onReady)
   }
 
   onError (error) {
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
@@ -8,7 +8,6 @@ import locationValidator from '../../helpers/locationValidator'
 class SingleImageViewerContainer extends React.Component {
   constructor () {
     super()
-    this.onLoad = this.onLoad.bind(this)
     this.onError = this.onError.bind(this)
     this.state = {
       height: null,
@@ -52,10 +51,6 @@ class SingleImageViewerContainer extends React.Component {
     }
   }
 
-  onLoad (event) {
-    this.props.onReady(event)
-  }
-
   onError (error) {
     console.error(error)
     this.setState({ loading: asyncStates.error })
@@ -79,7 +74,7 @@ class SingleImageViewerContainer extends React.Component {
     return (
       <SingleImageViewer
         onError={this.onError}
-        onLoad={this.onLoad}
+        onLoad={onReady}
         url={imageUrl}
       />
     )
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
@@ -8,6 +8,8 @@ import locationValidator from '../../helpers/locationValidator'
 class SingleImageViewerContainer extends React.Component {
   constructor () {
     super()
+    this.onLoad = this.onLoad.bind(this)
+    this.onError = this.onError.bind(this)
     this.state = {
       height: null,
       width: null,
@@ -16,20 +18,7 @@ class SingleImageViewerContainer extends React.Component {
   }
 
   componentDidMount () {
-    if (this.props.subject) {
-      this.handleSubject()
-    }
-  }
-
-  componentDidUpdate (prevProps) {
-    // Casting to JSON fixes reference issue from MST store
-    // A more robust solution might be to have a getter view function defined on the model
-    const prevSubject = prevProps.subject.toJSON()
-    const subject = this.props.subject.toJSON()
-
-    if (subject && (!prevSubject || prevSubject.id !== subject.id)) {
-      this.handleSubject()
-    }
+    this.handleSubject()
   }
 
   // TODO: store the subject image's naturalWidth, naturalHeight, clientWidth, and clientHeight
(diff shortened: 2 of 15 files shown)


eatyourgreens:
Thanks. The problem I ran into with the approach here is that these changes break any project that uses images as subjects. The store never registers the subject files as having loaded.

eatyourgreens:
@srallen I've re-written this so the store stays agnostic about the type of files that a subject might be loading. It should work now for TESS data subjects and for single image subjects. Each subject viewer is required to fire an `onReady` callback when all linked subject files are loaded and ready.

eatyourgreens:
Tasks, and task nav buttons, don't support a `disabled` prop at the moment, so I've just hidden the task area while the subject is loading. I'm not sure if I like that, visually.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/MultipleChoiceTask/MultipleChoiceTask.js
```diff
@@ -54,14 +54,18 @@ class MultipleChoiceTask extends React.Component {
...
       task
     } = this.props
     let annotation
     if (annotations && annotations.size > 0) {
       annotation = annotations.get(task.taskKey)
     }
     return (
-      <StyledFieldset autoFocus={(annotation && annotation.value && annotation.value.length === 0)}>
+      <StyledFieldset
+        autoFocus={(annotation && annotation.value && annotation.value.length === 0)}
```
eatyourgreens: Fieldsets aren't focusable elements in HTML, so I suspect this a line that was copied over from PFE even though it does nothing here. It can probably be removed.

eatyourgreens:
I've updated the task components so that task nav buttons and question answers can be disabled by passing a `disabled` prop. I think this is ready to go now.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/MultipleChoiceTask/MultipleChoiceTask.js
srallen: Yep, it wouldn't just been a straight lift from PFE. Let's open an issue to update the auto focus for the task area.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
```diff
@@ -59,14 +48,22 @@ class SingleImageViewerContainer extends React.Component {
...
+
+  onError (error) {
+    console.error(error)
+    this.setState({ loading: asyncStates.error })
+  }
+
   render () {
     const { loadingState } = this.state
-    const { subject } = this.props
+    const { onReady, subject } = this.props
```
srallen: What's onReady doing here? I don't see it being used.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -63,33 +63,43 @@ class LightCurveViewerContainer extends Component {
...
+  onLoad (rawData) {
+    this.setState({
+      dataExtent: {
+        x: d3.extent(rawData.x),
+        y: d3.extent(rawData.y)
+      },
+      dataPoints: zip(rawData.x, rawData.y),
+      loading: asyncStates.success
+    })
+    this.props.onReady()
```
srallen: Perhaps this should be a callback of `setState`?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
```diff
@@ -59,14 +48,22 @@ class SingleImageViewerContainer extends React.Component {
...
       })
     } catch (error) {
-      console.error(error)
-      this.setState({ loading: asyncStates.error })
+      this.onError(error)
     }
   }
 
+  onLoad (event) {
+    this.props.onReady(event)
```
srallen: Does the event object need to be passed along here? The `onSubjectReady` store function is just setting a boolean and doesn't do anything with the event.

packages/lib-classifier/src/store/SubjectViewerStore.js
```diff
@@ -27,55 +29,79 @@ const SubjectViewer = types
...
+
+      disableFullscreen () {
+        self.fullscreen = false
+      },
+
+      onSubjectReady () {
+        self.ready = true
+      },
+
+      resetSubject () {
```
srallen: I think this could be slightly more clearly named, perhaps `resetSubjectReady`?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
eatyourgreens: My bad. It's supposed to be passed down as the `onLoad` handler for the SVG image.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
eatyourgreens: Sure. I don't think it makes much difference here if `onReady` is called before, after, or at the same time as the state change.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
eatyourgreens: I originally thought `onReady` would be a proxy handler for the DOM's `load` event but that's only true if the store ever needs to reflect on the event eg. to get info about the downloaded image file. The image loading code is slightly complicated anyway, because of the preloader that I'm assuming is there to get the images original height and width?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
eatyourgreens: I've rewritten this, and added a test.

eatyourgreens:
Looking back through my changes here, I'd also expect a test that the task area is disabled while the subject is loading, but I haven't added one. 😞

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.spec.js
```diff
@@ -26,21 +34,51 @@ describe('Tasks', function () {
...
+        <Tasks.wrappedComponent
+          loadingState={asyncStates.success}
+          ready={false}
+          tasks={tasks}
+        />
+      )
+      taskWrapper = wrapper.find('inject-SingleChoiceTask')
+    })
+
+    it('tasks should be disabled', function () {
```
eatyourgreens: Small thing but I'm not a big fan of wording like `it tasks should be disabled.` It tells me I haven't quite got my describe/it blocks set up correctly.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.spec.js
```diff
@@ -26,21 +34,51 @@ describe('Tasks', function () {
...
-      type: 'single'
-    }]
-
-    const wrapper = shallow(<Tasks.wrappedComponent loadingState={asyncStates.success} tasks={tasks} />)
+    const wrapper = shallow(<Tasks.wrappedComponent loadingState={asyncStates.success} ready={true} tasks={tasks} />)
     // Is there a better way to do this?
     expect(wrapper.find('inject-SingleChoiceTask')).to.have.lengthOf(1)
   })
+
+  describe('when the subject is loading', function () {
```
eatyourgreens: Maybe this should be describing a thing, rather than a state. Then the `it` block can test an expected state for that thing.

eatyourgreens:
@srallen this should be done now. I added some more tests, in addition to the PR comments.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
srallen: Yes, the original PFE code gets the original height and width as well as the client height and width  at the onload event and stores them in the classification metadata. We still need to add that data to the classification metadata as noted in #304, but it's hasn't been done yet because we're now using an svg `image` instead of an html `img` and it doesn't have a client height and width. I'm not sure if we should go forward with the change to svg `image` or not.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.spec.js
```diff
@@ -5,6 +5,14 @@ import { Tasks } from './Tasks'
 import asyncStates from '@zooniverse/async-states'
 
 describe('Tasks', function () {
+  const tasks = [{
```
srallen: There's a factory for a test mock of single choice task types if you want to use it. I'd like us to move toward using those so our mocks are consistent across tests. This isn't really a blocker for me though.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
srallen: I could see this being used to store the image's original height and width in the subject viewer store and the classification store could observe for that state and update the metadata.

eatyourgreens: Yeah, if you inspect the `image` element in the DOM, client height and width are 0 in Chrome. Although we've recorded those numbers historically, I don't think they've ever been used in data analysis. Caesar doesn't need the client size, as far as I know, although it will need the natural size of an SVG image in order to analyse drawing annotations.

>I'm not sure if we should go forward with the change to svg image or not.

I don't think you can have drawing tasks or zoom and rotation transforms without using SVG.

eatyourgreens: PFE renders an SVG rectangle just to get the client bounding box for drawing tasks:
https://github.com/zooniverse/Panoptes-Front-End/blob/fe0c87a8ef589ef6c662bbb350ed35f8b1b8e895/app/classifier/annotation-renderer/svg.jsx#L18-L31

Anti-Slavery Manuscripts is a drawing task on top of a jpeg image, so it might be worth looking at their code to see how they get the image dimensions in oder to map drawn points to positions on a page. They might have used a better approach than the one in PFE. AnnoTate might be worth looking at too.

eatyourgreens: I've started work on this in #804.

eatyourgreens: If we do need client height/width, they can be read from one of the containing DOM elements. Or we can get the client bounding box of the SVG element. #804 is pretty much done but I'm recording the client dimensions as 0 at the moment.

PR #792, opened 2019-05-01
Refactor Image Toolbar to use Grommet Button and Blank icon

mcbouslog:
Package: lib-classifier

Describe your changes:
- refactors image toolbar icons with Grommet `<Blank />` icon component
- refactors Button.js with styled Grommet `<Button />`
- adds `title` attribute for temporary tooltip until Grommet `<Drop />` fixed (see Grommet issue 2660)

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


srallen:
@mcbouslog and I will pair on resolving the Grommet bug this week, but we won't know the timeline on that being reviewed and published, so using HTML title attr is a good interim.

srallen:
@mcbouslog There's something wrong with the travis build: https://travis-ci.com/zooniverse/front-end-monorepo/builds/110393301

It looks like some components aren't being found?

mcbouslog:
TIL renaming a file with only a capitalization change may or may not be recognized by git depending on OS, in which case `git mv --force myFile MyFile` forces the update.

packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/Button/Button.js
```diff
@@ -1,65 +1,43 @@
...
-      width: `${sizeInPercent}%`
+    > svg {
+      fill: white;
     }
   }
 
+  > svg {
+    fill: ${props => props.active ? 'white' : (props.theme.dark ? 'white' : 'black')};
+    stroke: transparent;
+    width: ${pxToRem(18)};
```
srallen: Height needs to be set as well here. Right now, the icons are inheriting the Grommet default for height now, 24px, and it's making the circle buttons oval now. Height should be the same as width to have the inner icon be square and the outer button be a circle.

mcbouslog: Yes it looked a little off thank you for catching!

PR #804, opened 2019-05-08
Record subject image dimensions on load

### LINKED ISSUE #201: SingleSubjectImageViewer using SVG image, might want to switch to HTML img
**Package** lib-classifier

We store the subject image's `naturalWidth`, `naturalHeight`, `clientWidth`, `clientHeight` per image (stored by frame index) in the classification's metadata. This is to record the original image size versus how it is rendered in the browser. I'm not sure what the equivalent is for SVG `image`, but if we cannot do this, then we need to refactor the viewer to use HTML `img`.

### LINKED ISSUE #304: Add subject dimensions to classification metadata

eatyourgreens:
Get image dimensions from the HTML image preloader. Create a new, fake event and fake target on SVG image load, since the originals are read-only. Pass the fake event, with the HTML image dimensions, to the subject viewer store via the onReady event handler.

Todo:
- [x] Enable React lifecycle testing.
- [x] Mock window.Image and test the preloader.
- [x] ~~Record dimensions for SVG data images.~~
- [x] Pass image dimensions from the subject viewer to the current classification.
- [x] Record client rendered dimensions too (this might be optional.)

Package:
lib-classifier

Closes #201.
Closes #304.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
@@ -83,8 +83,7 @@ class LightCurveViewerContainer extends Component {
 
   onLoad (rawData) {
     const { onReady } = this.props
-    // TODO: find a cleaner way to get the container node for the viewer.
-    const target = this.viewer.current.wrappedInstance.svgContainer.current
+    const target = this.viewer.current
     this.setState({
       dataExtent: {
         x: d3.extent(rawData.x),
@@ -114,7 +113,7 @@ class LightCurveViewerContainer extends Component {
 
     return (
       <LightCurveViewer
-        ref={this.viewer}
+        forwardRef={this.viewer}
         dataExtent={this.state.dataExtent}
         dataPoints={this.state.dataPoints}
         drawFeedbackBrushes={this.props.drawFeedbackBrushes}
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
@@ -13,6 +13,7 @@ import withKeyZoom from '../../../withKeyZoom'
 class LightCurveViewerContainer extends Component {
   constructor () {
     super()
+    this.viewer = React.createRef()
     this.state = {
       loading: asyncStates.initialized,
       dataExtent: {
@@ -82,6 +83,8 @@ class LightCurveViewerContainer extends Component {
 
   onLoad (rawData) {
     const { onReady } = this.props
+    // TODO: find a cleaner way to get the container node for the viewer.
+    const target = this.viewer.current.wrappedInstance.svgContainer.current
     this.setState({
       dataExtent: {
         x: d3.extent(rawData.x),
@@ -90,7 +93,9 @@ class LightCurveViewerContainer extends Component {
       dataPoints: zip(rawData.x, rawData.y),
       loading: asyncStates.success
     },
-    onReady)
+    function () {
+      onReady({ target })
+    })
   }
 
   onError (error) {
@@ -109,6 +114,7 @@ class LightCurveViewerContainer extends Component {
 
     return (
       <LightCurveViewer
+        ref={this.viewer}
         dataExtent={this.state.dataExtent}
         dataPoints={this.state.dataPoints}
         drawFeedbackBrushes={this.props.drawFeedbackBrushes}
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
@@ -8,10 +8,11 @@ import locationValidator from '../../helpers/locationValidator'
 class SingleImageViewerContainer extends React.Component {
   constructor () {
     super()
+    this.onLoad = this.onLoad.bind(this)
     this.onError = this.onError.bind(this)
     this.state = {
-      height: null,
-      width: null,
+      naturalHeight: null,
+      naturalWidth: null,
       loading: asyncStates.initialized
     }
   }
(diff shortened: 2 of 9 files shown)


packages/lib-classifier/src/store/SubjectViewerStore.js
```diff
@@ -63,12 +69,21 @@ const SubjectViewer = types
...
+      onSubjectReady (event) {
+        const { target = {} } = event || {}
+        const {
+          clientHeight = 0,
+          clientWidth = 0,
+          naturalHeight = 0,
+          naturalWidth = 0
+        } = target
+        self.dimensions.push({ clientHeight, clientWidth, naturalHeight, naturalWidth })
         self.ready = true
```
eatyourgreens: Maybe use `self.ready = self.dimensions.length === subject.locations.length` here to account for subjects with multiple linked files.

eatyourgreens:
`clientWidth` and `clientHeight` are set to 0 at the moment, but I could update this to get the rendered dimensions of the subject viewer's container. I'm not 100% sure if these are useful numbers to have with each classification. `naturalHeight` and `naturalWidth` make sense for interpreting the annotations from drawing tasks.

eatyourgreens:
Client dimensions added now, by getting a ref to the LCV's container DOM node. The code is a bit fragile, though. 🤔

packages/lib-classifier/src/store/ClassificationStore.js
```diff
@@ -132,10 +132,12 @@ const ClassificationStore = types
 
     function completeClassification () {
       const classification = self.active
+      const subjectDimensions = getRoot(self).subjectViewer.dimensions.toJSON()
```
srallen: Could we use `toJS` instead of `toJSON`? `toJS` gives us options about how to convert maps, if the data structure is using one.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -82,6 +83,8 @@ class LightCurveViewerContainer extends Component {
 
   onLoad (rawData) {
     const { onReady } = this.props
+    // TODO: find a cleaner way to get the container node for the viewer.
+    const target = this.viewer.current.wrappedInstance.svgContainer.current
```
srallen: I'm planning on refactoring the LCV to not be the connected component to the store and instead move that up to the container here (or have another container specific to the store data connection), so that should make the ref cleaner.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
```diff
@@ -8,10 +8,14 @@ import locationValidator from '../../helpers/locationValidator'
...
+    this.imageViewer = React.createRef()
+    this.onLoad = this.onLoad.bind(this)
     this.onError = this.onError.bind(this)
     this.state = {
-      height: null,
-      width: null,
+      clientHeight: 0,
+      clientWidth: 0,
+      naturalHeight: null,
+      naturalWidth: null,
```
srallen: Should the natural sizes also be 0? Just to be consistent with type with the client sizes

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
```diff
@@ -36,29 +40,41 @@ class SingleImageViewerContainer extends React.Component {
...
       })
     } catch (error) {
       this.onError(error)
     }
   }
 
+  onLoad (event) {
+    const { onReady } = this.props
+    const { clientHeight, clientWidth, naturalHeight, naturalWidth } = this.state
+    const { target } = event || {}
```
srallen: When would the target be undefined?

packages/lib-classifier/src/store/SubjectViewerStore.js
srallen: Yes, we'll want to do something like that when multi-image subject support is added. Could you note that in an issue? We have a milestone for implementing the other subject viewers that it could be linked to.

packages/lib-classifier/src/store/ClassificationStore.js
eatyourgreens: Sure.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
eatyourgreens: I've changed this to use a forward ref, which is a lot cleaner.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
eatyourgreens: Sure. They were `null` in the original code but that is inconsistent with the model type.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
eatyourgreens: Video and audio, probably, from what I remember of the `onLoad` callbacks in PFE.

eatyourgreens: This line catches the case where the event is undefined, though.

eatyourgreens: Oh wait, this is the image loading code. Ignore my comments, I was thinking of the generic event handler in the subject viewer store.

packages/lib-classifier/src/store/SubjectViewerStore.js
eatyourgreens: Sure. There's also audio projects where each subject has two locations: one audio, one image; and video projects like Chimp & See where the first location is a video and the rest of the locations array are images.

HTML5 media don't fire a load event so we'll have to come back to this anyway.

eatyourgreens:
@srallen changes should all be taken care of now.

PR #808, opened 2019-05-09
Deploy production version of app-project to kubernetes

rogerhutchings:
- Deploy production version of app-project to kubernetes
- Rename the staging version for clarity

### DIFF
--- a/kubernetes/deployment.tmpl
+++ b/kubernetes/deployment.tmpl
@@ -1,21 +1,48 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: front-end-staging
+  name: front-end-project-staging
   labels:
-    app: front-end-staging
+    app: front-end-project-staging
 spec:
   replicas: 2
   selector:
     matchLabels:
-      app: front-end-staging
+      app: front-end-project-staging
   template:
     metadata:
       labels:
-        app: front-end-staging
+        app: front-end-project-staging
     spec:
       containers:
         - name: fe-project-staging
           image: zooniverse/fe-project:__IMAGE_TAG__
           ports:
             - containerPort: 3000
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: front-end-project-production
+  labels:
+    app: front-end-project-production
+spec:
+  replicas: 2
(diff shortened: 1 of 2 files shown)


kubernetes/deployment.tmpl
```diff
@@ -1,21 +1,48 @@
...
       containers:
         - name: fe-project-staging
           image: zooniverse/fe-project:__IMAGE_TAG__
           ports:
             - containerPort: 3000
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: front-end-project-production
```
camallen: `front-end-project-production` doesn't really tell me which service this, could this be better named, e.g. `zooniverse-org-production`

adammcmaster: I agree. At the moment it's obvious that the "front end project" is the rebuild, but that won't make sense a couple of years from now.

adammcmaster:
> Will changing a service name require us to manually remove the service and re-create it?

Yes I think so – we won't have to manually create the new one but we'll need to remove the old one.

kubernetes/deployment.tmpl
rogerhutchings: Okay - how does that fit in with specifying that this is the project app as opposed to the static pages / admin / whatever app, as these will be built out separately? Or will they share the same service?

rogerhutchings:
> With this setup, production and staging will both always be deployed at the same time and will always run the same version of the code. Is that what we want at this stage?

I _think_ so - what we'd want ideally is automatic branch builds at some point for testing purposes, but I'll settle for a staging and production build

kubernetes/deployment.tmpl
camallen: `zooniverse-org-project-production` vs `zooniverse-org-static-production` etc?

rogerhutchings: Yeah - `zooniverse.org` will consist of multiple apps, which will have >1 environment (staging, production) etc, so would `zooniverse-org-project-production` for this one be more appropriate?

rogerhutchings:
I'm game - let me see if I can get `kubectl` set up

Issue #650, opened 2019-04-04
Breaking up the LCV


rogerhutchings:
**Package**

lib-classifier

**Feature or Issue Description**

I'm not happy that the LCV is a big ball of code with no tests, so here are some thoughts:

- We break up the LCV into a plain viewer (for Talk), viewer with non-interactive annotations (for feedback) and marking tools
- We add tests
- We add an event bus for the classifier to handle button interactions; there's a discussion to be had as to whether viewer state is kept in the viewer or the store, but a common interface to trigger these would be a good first step
- We break off the LCV into a separate package, and dynamically import it when required. It comes with dependencies like D3 that won't be required 

cc #366

eatyourgreens:
#648 makes the big ball of untested code even bigger. Could feedback use a decorator: `withFeedback(SubjectViewer)` to add feedback marks to the viewer?

@rogerhutchings have you got time to comment over there about making feedback generic, and less coupled to the light curve viewer?

srallen:
@eatyourgreens Mark is working on that refactor but I asked him to push up what he had done now so that we can have the UI in place for beta testing with volunteers.

srallen:
> We break up the LCV into a plain viewer (for Talk), viewer with non-interactive annotations (for feedback) and marking tools

The LCV shouldn't even be coupled with marking tools. We have an incoming project with CSSI that will want to visualize the light curves of variable stars, but the task is to identify them using the survey task, so no marking will be happening at all.

srallen:
The LCV should be able to take props to set:
- Axis labels
- Zoom type: x-axis only or x and y (brush zoom)
- Axis inversion
- Initial zoom level

eatyourgreens:
In my mind, any feedback between task and subject should happen via a common parent or store. Or the subject viewer should be a child of the task component, so that the task can pass props to the subject.

PR #809, opened 2019-05-09
LCV container tests

srallen:
Package: lib-classifier

For #650

Describe your changes:
A first step toward the refactor by adding tests to the LCV container component. I plan on moving the store connection from the LCV to the LCV container and I want to make sure what is existing doesn't get broken in that process.

~I have a single failing test that is a bit of a mystery to me at the moment. It `calls the onReady prop on successful load` passes if isolated with `only`, but fails if run altogether with the block.~ I figured out that I was missing a config for nock to work for multiple tests. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
```diff
@@ -1,20 +1,173 @@
...
+      const nextSubjectD3YExtent = d3.extent(nextSubjectJSON.y)
+      const cduSpy = sinon.spy(LightCurveViewerContainer.wrappedComponent.prototype, 'componentDidUpdate')
+      wrapper = shallow(
+        <LightCurveViewerContainer.wrappedComponent subject={subject} />
+      )
+      wrapper.setProps({ subject: nextSubject })
+
+      // componentDidUpdate gets called three times
+      // only the third is the update to the new props we're expecting for this test
+      // Is there a better way to do this? This seems brittle
```
eatyourgreens: Spying on `wrapper.instance().onLoad` might be what you need here. 

By the way, the classifier app is set up to mount a new subject viewer for each new subject, so these updates may never happen in actual use.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -124,16 +124,20 @@ class LightCurveViewerContainer extends Component {
   }
 }
 
-LightCurveViewerContainer.defaultProps = {
-  onReady: () => true
+LightCurveViewerContainer.wrappedComponent.defaultProps = {
+  onReady: () => true,
+  subject: {
```
eatyourgreens: This is different from the default subject prop for the `SingleImageViewerContainer`.  Subject file viewers should probably be consistent in their props, so we should decide whether that component should set a default subject too.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -58,26 +59,25 @@ class LightCurveViewerContainer extends Component {
...
+      this.onError(error)
     }
   }
 
   async handleSubject () {
     try {
       const rawData = await this.requestData()
-      this.onLoad(rawData)
-      
+      if (rawData && rawData.x && rawData.y) this.onLoad(rawData)
```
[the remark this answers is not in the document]
eatyourgreens: I think the previous version of the code errored if the data was invalid. My reading of this change is that the component now fails silently if the data is invalid. I think we should signal that something's wrong if the subject data fails to load properly.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
```diff
@@ -1,20 +1,173 @@
...
+      nockScope.persist(false)
+    })
+
+    it('should error if a json or text subject location file can\'t be found', function (done) {
+      wrapper = shallow(
+        <LightCurveViewerContainer.wrappedComponent subject={imageSubject} />
+      )
+
+      expect(cdmSpy).to.have.been.calledOnce
+      cdmSpy.returnValues[0].then(() => {
```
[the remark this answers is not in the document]
eatyourgreens: I've opened #810 to try this out for `componentDidMount`.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
eatyourgreens: Maybe use a timeout here too, and expect the state to have updated after a short pause.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
```diff
@@ -1,20 +1,173 @@
...
+          expect(xDataPoint).to.equal(nextSubjectD3XExtent[index])
+        })
+
+        wrapper.state().dataExtent.y.forEach((yDataPoint, index) => {
+          expect(yDataPoint).to.equal(nextSubjectD3YExtent[index])
+        })
+
+        expect(wrapper.state().dataPoints).to.have.lengthOf(nextSubjectJSONZip.length)
+
+        expect(wrapper.state().loading).to.equal(asyncStates.success)
```
eatyourgreens: The loading state should be checked while the new subject is loading too, so immediately after the props change.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
```diff
@@ -1,20 +1,173 @@
...
+      nock.cleanAll()
+      nockScope.persist(false)
+    })
+
+    it('should error if a json or text subject location file can\'t be found', function (done) {
+      wrapper = shallow(
+        <LightCurveViewerContainer.wrappedComponent subject={imageSubject} />
+      )
+
+      expect(cdmSpy).to.have.been.calledOnce
```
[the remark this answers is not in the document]
srallen: Yes, it's cruft from when I was trying to figure out why componentDidMount was showing up in the spy as being called twice. I hadn't noticed that the parent describe block was mounting the wrapper for its tests in a `beforeEach` by that point.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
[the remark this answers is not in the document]
srallen: I did a bunch of reading about how to test an async `componentDidMount` and the maintainers seem to prefer having the promise returned to make it available for tests: https://github.com/airbnb/enzyme/issues/1587#issuecomment-400934989

Using `setTimeout` seems more brittle to me because it's arbitrary rather than explicitly using the resolve value from the Promise.

srallen:
Could we have waiting until merging #810 until I had a chance to respond?

srallen:
We discussed potentially using the `onReady` prop to spy on instead.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
```diff
@@ -1,20 +1,173 @@
...
+describe('Component > LightCurveViewerContainer', function () {
   it('should render without crashing', function () {
+    wrapper = shallow(<LightCurveViewerContainer.wrappedComponent />)
     expect(wrapper).to.be.ok
   })
 
   it('should render null if there is no subject prop', function () {
-    expect(wrapper.type()).to.be.null
+    wrapper = shallow(<LightCurveViewerContainer.wrappedComponent />)
+    expect(wrapper.html()).to.be.null
```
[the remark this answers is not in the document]
srallen: It did fail and I fixed the code, but did not rename the test description to match.

srallen:
A few minutes of experimentation I've found that spying on `onReady` isn't useful for testing error states since `onReady` won't be called. We could spy on `onError`, but we'll have the same problem of figuring out how to tell the test to wait for the async calls to resolve.

I experimented with disabling the lifecycle methods and found that it worked for testing on waiting for a single promise to resolve, but not a chain of them. For example, my test for `it 'should error if the location request response fails'` is failing doing this because the test expectations run before the await for the http request resolves. 

I really think returning the promise from `componentDidMount` is cleanest and most reliable in this case. I've found another example doing a similar approach: https://medium.com/@manastunga/unit-testing-api-calls-in-react-enzyme-and-jest-133b87aaacb4

One last approach could be flushing the promises, which I haven't tried yet: https://medium.com/@lucksp_22012/jest-enzyme-react-testing-with-async-componentdidmount-7c4c99e77d2d

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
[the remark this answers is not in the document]
eatyourgreens: If the exact timeout is the problem, I suppose you could set an explicit delay for the mock response with `nock.setDelay()` then expect state to change after that time.

My problem with using the promise is that you're changing the component code for the tests, so you're testing implementation, not behaviour (https://github.com/airbnb/enzyme/issues/1587#issuecomment-438876545). You've limited how much you can refactor the component, because you've tied the implementation of your tests to the implementation of `componentDidMount`. This might not be big deal, however.

eatyourgreens:
`async componentDidMount` seems much less brittle than conditionally returning a Promise if a specific prop is present. You don't have to check whether `componentDidMount().then` exists before you use it.

Should `onError` happen in the store, rather than the component? #815 has made me wonder if there should be a common error handler for these components.

eatyourgreens:
In #815 I'm splitting the viewer tests into `without a subject`, `with a valid subject` and `with an invalid subject`, which might be a useful way to organise the tests here too.

srallen:
> Should onError happen in the store, rather than the component? #815 has made me wonder if there should be a common error handler for these components.

If `ready` is renamed to be `loading` and used the async states and then we also stored the error message in the store, then that will be more consistent with the other store APIs. I'd be for this refactor.

srallen:
> async componentDidMount seems much less brittle than conditionally returning a Promise if a specific prop is present. You don't have to check whether componentDidMount().then exists before you use it.

I'll give this approach a try. I can see your point about using a return for the other promise being inconsistent since it is within a conditional while making `componentDidMount` async will guarantee that a promise is returned.

eatyourgreens:
I'm going to update #815 to add `loadingState` to the subject viewer store.

srallen:
This will be updated to use the improved error handling introduced in #815

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
srallen: The LCV container is using subject id in the componentDidUpdate lifecycle method, so we need a default. We could have a default empty string for id for all the viewers and check that?

srallen:
This is ready for review again. It's been updated to use the error handling in the subject viewer store.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -24,28 +23,36 @@ class LightCurveViewerContainer extends Component {
...
-  componentDidMount () {
-    if (this.props.subject) {
-      this.handleSubject()
+  async componentDidMount () {
+    const { onError, subject } = this.props
+    if (subject) {
+      try {
+        await this.handleSubject()
+      } catch (error) {
+        onError(error)
```
eatyourgreens: `handleSubject` catches errors so this will never be called.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -24,28 +23,36 @@ class LightCurveViewerContainer extends Component {
...
     const prevSubjectId = prevProps.subject && prevProps.subject.id
     const subjectChanged = subject && (subject.id !== prevSubjectId)
 
     if (subjectChanged) {
-      this.handleSubject()
+      try {
+        await this.handleSubject()
+      } catch (error) {
+        onError(error)
+      }
```
eatyourgreens: Same here.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -55,28 +62,28 @@ class LightCurveViewerContainer extends Component {
   }
 
   async requestData () {
+    const { onError } = this.props
     try {
-      this.setState({ loading: asyncStates.loading })
+      // this.setState({ loading: asyncStates.loading })
```
eatyourgreens: Should this be commented?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -55,28 +62,28 @@ class LightCurveViewerContainer extends Component {
...
       const url = this.getSubjectUrl()
       const response = await request.get(url)
-      if (!response.ok) {
-        throw new Error('Invalid response')
-      } else {
+      if (response.ok) {
```
eatyourgreens: If the response isn’t ok, what does this function do?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -55,28 +62,28 @@ class LightCurveViewerContainer extends Component {
...
         // Get the JSON data, or (as a failsafe) parse the JSON data if the
         // response is returned as a string
         return response.body || JSON.parse(response.text)
       }
     } catch (error) {
-      return this.onError(error)
+      onError(error)
```
eatyourgreens: What does this function return after catching an error?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -55,28 +62,28 @@ class LightCurveViewerContainer extends Component {
...
     }
   }
 
   async handleSubject () {
+    const { onError } = this.props
     try {
       const rawData = await this.requestData()
-      this.onLoad(rawData)
+      if (rawData) this.onLoad(rawData)
```
eatyourgreens: This fails silently if `rawData` is falsy. Should it message that the request failed?

eatyourgreens: Sorry, just to expand on that, it looks to me like `this.requestData()` returns `undefined` for bad responses or errors. Is that intentional? Previously, that would have thrown a new error here but the code now fails silently.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
eatyourgreens: I think I'd prefer if the default subject is null, but I'm not sure. Is there a reason for having a mock subject as the default? It seems to me that this means any code that handles the subject must first check if it's real and not the default.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
```diff
@@ -1,20 +1,187 @@
...
+      nock.cleanAll()
+      nockScope.persist(false)
+    })
+
+    it('should error if a json or text subject location file can\'t be found', function (done) {
+      wrapper = shallow(
+        <LightCurveViewerContainer.wrappedComponent onError={onErrorSpy} subject={imageSubject} />
+      )
+
+      cdmSpy.returnValues[0].then(() => {
```
eatyourgreens: If you set `onErrorSpy` to call `done()` after it runs then you don't need this.

eatyourgreens: Or rather, you could replace `cmdSpy.returnValues[0].then(…)` with `onError.callsFake(…)` to run your expectations and call `done()` when the callback is called.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
```diff
@@ -1,20 +1,187 @@
...
+        expect(onErrorSpy).to.have.been.calledOnce
+        expect(onErrorSpy.args[0][0].message).to.equal('No JSON url found for this subject')
```
eatyourgreens: Just a style comment, but would `expect(onErrorSpy.withArgs(new Error('No JSON url found for this subject'))).to.have.been.calledOnce` work here?

srallen: I did try that first and it does not work. Whatever underlying matching sinon does for the arguments does not match the error argument with a new Error object, so it'll comeback as not having been called with that argument.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
srallen: It should actually be removed because the state is tracked in the store now. I just forgot to do that.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
srallen: It's an error. Superagent rejects none 200 responses for us, and that error is then caught. I suppose technically the check for `response.ok` doesn't have to happen here because of superagent.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
srallen: Returning the error makes `rawData` truthy and then I have to check if it is actually the data or not an Error object like I was previously doing. If an error is returned by `this.requestData()` and I don't check that it's not an error object, then `this.onLoad` will attempt to use lodash and set state with that error object and then error. That error will be caught, but it's a lodash error, not the error from the data request. 

This is the bit of error handling and bubbling I found hard with async/await. I'm not sure if this is right, but it's how I got it to work.... returning undefined on error from `this.requestData` made it so I was sure that `onLoad` was only called with what I wanted it to.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
srallen: I think it makes sense to mock the subject because there _should_ be a subject by the time the specific subject viewer component loads. If there isn't, then something is really going wrong with the parent component or the store. 

We could instead set the subject prop to be required and remove the default and check for it being undefined in `componentDidUpdate`. I don't have a strong opinion either way, but this component shouldn't be mounting without a subject.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
srallen: I've made the lifecycles async and it makes sense to test that to me, which using the spies on them indirectly does.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
eatyourgreens: After catching an error and recovering from it, you’d expect `requestData` to return a `rawData` object. If it helps, think of how the return value from a `catch` block is passed to the next `then` block in a promise chain.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
eatyourgreens: Good point. I was thinking of the case where the default subject gets passed down the tree, but maybe that never happens.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
eatyourgreens: Fair enough, but this makes the LCV behave differently from viewers for images, audio and video, which won’t have this async `componentDidMount` behaviour.

srallen: I think this will be true for the data visualization viewers.

srallen:
@eatyourgreens I've tidied this up a bit. I think it's ready for review again.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -55,28 +54,26 @@ class LightCurveViewerContainer extends Component {
...
-        // response is returned as a string
-        return response.body || JSON.parse(response.text)
-      }
+
+      // Get the JSON data, or (as a failsafe) parse the JSON data if the
+      // response is returned as a string
+      return response.body || JSON.parse(response.text)
     } catch (error) {
-      return this.onError(error)
+      return onError(error)
```
eatyourgreens:
```suggestion
      onError(error)
      return { x: [], y: [] }
```

eatyourgreens: Log the error, then return a default value for the response data.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
eatyourgreens: Text files need custom code to be fetched from the server too. I think it's still possible to write a set of generic tests for the viewer container props, without expecting a specific implementation in the container code. That can be addressed in #811.

packages/lib-classifier/src/store/SubjectViewerStore.js
```diff
@@ -73,6 +73,7 @@ const SubjectViewer = types
       onError (error) {
         console.error(error)
         self.loadingState = asyncStates.error
+        return error
```
eatyourgreens:
```suggestion
        error
```

eatyourgreens: I don't think this returned value gets used anywhere.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
```diff
@@ -1,20 +1,187 @@
...
+
+const imageSubject = Factory.build('subject')
 
+const failSubject = Factory.build('subject', {
+  locations: [
+    { 'application/json': 'http://localhost:8080/failure.json' }
+  ]
+})
+
+describe.only('Component > LightCurveViewerContainer', function () {
```
eatyourgreens:
```suggestion
describe('Component > LightCurveViewerContainer', function () {
```

eatyourgreens:
There was a stray `.only()` in the tests, so I've removed it using the PR suggestion/commit feature.

packages/lib-classifier/src/store/SubjectViewerStore.js
srallen: Ok, I think this can be removed entirely then. I was misunderstanding your consistent return suggestion and thought you meant that the error should be returned.

srallen:
@eatyourgreens this is ready for review again

PR #810, opened 2019-05-10
Use a timeout for LCV async tests

eatyourgreens:
Remove the spy on componentDidMount. Set a timeout instead, to wait for component state to update on mock subject load.

Package:
lib-classifier

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
@rogerhutchings done!

srallen:
I don't agree with this approach.

PR #815, opened 2019-05-13
Add tests for errors when loading subject images

eatyourgreens:
Package:
lib-classifier

Mock a bad image for the single image viewer and stub console error logging. Test the component error handler for the cases where either the HTML `img` or the SVG `image` elements throw an error.

Rename `handleSubject` to `getImageSize`, to make its purpose clear.

Add an `onError` action to the subject viewer store and use that to handle subject viewer errors. Rename `SubjectViewerStore.ready` to `SubjectViewerStore.loadingState` and change it from a boolean to an enumeration of loading states.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/SubjectViewer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/SubjectViewer.js
@@ -7,12 +7,13 @@ import getViewer from './helpers/getViewer'
 
 function storeMapper (stores) {
   const { active: subject, loadingState } = stores.classifierStore.subjects
-  const { onSubjectReady, onError } = stores.classifierStore.subjectViewer
+  const { onSubjectReady, onError, loadingState: subjectReadyState } = stores.classifierStore.subjectViewer
   return {
     loadingState,
     onError,
     onSubjectReady,
-    subject
+    subject,
+    subjectReadyState
   }
 }
 
@@ -33,12 +34,13 @@ class SubjectViewer extends React.Component {
   }
 
   [asyncStates.success] () {
-    const { onError, onSubjectReady, subject } = this.props
+    const { onError, onSubjectReady, subject, subjectReadyState } = this.props
     const Viewer = getViewer(subject.viewer)
     return (
       <Viewer
         key={subject.id}
         subject={subject}
+        loadingState={subjectReadyState}
         onError={onError}
         onReady={onSubjectReady}
       />
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/SubjectViewer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/SubjectViewer.js
@@ -7,9 +7,10 @@ import getViewer from './helpers/getViewer'
 
 function storeMapper (stores) {
   const { active: subject, loadingState } = stores.classifierStore.subjects
-  const { onSubjectReady } = stores.classifierStore.subjectViewer
+  const { onSubjectReady, onError } = stores.classifierStore.subjectViewer
   return {
     loadingState,
+    onError,
     onSubjectReady,
     subject
   }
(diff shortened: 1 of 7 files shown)


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/SubjectViewer.js
```diff
@@ -7,11 +7,13 @@ import getViewer from './helpers/getViewer'
 
 function storeMapper (stores) {
   const { active: subject, loadingState } = stores.classifierStore.subjects
-  const { onSubjectReady } = stores.classifierStore.subjectViewer
+  const { onSubjectReady, onError, loadingState: subjectReadyState } = stores.classifierStore.subjectViewer
```
srallen: Could we get a rename for the above `loadingState` to make it extra clear the difference between the two being checked in this component? How about `subjectQueueState`?

eatyourgreens:
`componentDidMount` doesn’t fire the `onReady` callback. It fires when the `load` event is fired by the child component. Even after writing the tests, I’m not sure what the HTML img does in this component. I thought it was a preloader, but the HTML and SVG images load in parallel.

I’m thinking maybe create the HTML image element after the SVG has loaded, in order to get the image dimensions, then call the `onReady` callback once the image has loaded and we know how big it is.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/SubjectViewer.js
eatyourgreens: Sure. Is that the loading state on line 9?

eatyourgreens:
Testing this out with real subjects, there's a bug where the classification is submitted with the image dimensions set to 0. I've updated the tests to catch it. It happens when `onLoad` runs before the HTML image has loaded.

eatyourgreens:
@srallen the component doesn't do anything in `componentDidMount` any more, so I've updated the tests to call `done()` when the `onReady` and `onError` callbacks run.

srallen:
FWIW, based on my discussions on slack, I don't think we need to record client dimensions. I think there's potential for it to be confusing since project owners might not know or understand the difference between the natural, client, and viewport sizes we record. I think I'll write an ADR up proposing we drop recording the client dimensions in the classification metadata.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/SubjectViewer.js
srallen: Yes, that's the one.

srallen:
Just to note, removal of client dimensions can be after we discuss and agree on the ADR.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
```diff
@@ -37,44 +22,33 @@ class SingleImageViewerContainer extends React.Component {
     })
   }
 
-  async handleSubject () {
-    const { subject } = this.props
+  async getImageSize () {
+    const { onError, subject } = this.props
```
srallen: `onError` isn't being used in this method anymore. Why did the try/catch get removed? Is it bubbling up and being caught in `onLoad`? The test specs indicate it's being caught, so this isn't a blocker, just more of an understanding question because I find the async/await error handling and bubbling of events to be a harder to follow than promises.

eatyourgreens:
Dropping `clientHeight` and `clientWidth` sounds good to me. I can't think of a case where you might use the rendered size of the subject viewer in your analysis of classifications.

eatyourgreens:
I should also note that this new image viewer records the client height and width of the `<svg>` element, which is not quite the same as the rendered size of the `<image>` element since there may be margins, in order to fit the full image into the available space.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
eatyourgreens: Good catch. If we catch the error here, then the calling function will always run successfully. In this case, that means `onReady` always gets called and you get classifications with `{ naturalHeight: 0, naturalWidth: 0 }` if the error is trapped here. I thought it was better to throw an error here and catch it in `onLoad`, so that we can set the store to an error state.

PR #821, opened 2019-05-14
Hook up the daily, total counts to the stats widget

rogerhutchings:
Hook up the classification counts to the stats widget.

~~For some reason, I can't get this to update once the user store has been populated. It _will_ update properly on a hot reload, however 🤔~~

~~Steps to reproduce:~~

~~1. Fire up the app (we're hitting the stats API, so make sure you've set `PANOPTES_ENV=production`)~~
~~1. Navigate to a project you've classified on, e.g. `/projects/zookeeper/galaxy-zoo/classify`~~
~~1. Look at the console - you should see the correct stats being logged~~
~~1. Look at the actual counts in the stats widget, they will still be 0~~

~~I've tried the recommended approach of deferencing as late as possible, by creating a view returning the relevant counts and passing that through to the component. Not sure why this isn't working...~~

Thanks to @srallen for the catch, this is now grabbing the stats.

To do (in other issues):

- Set up an observer on the classification store to watch for submitted classifications and increment the counters. This will be stored in the `sessionCount` volatile, and will need to be added to the daily count, the total count, and the weekly stats used for the chart. (#860)
- Create the daily count bar chart (#38)

cc #787


rogerhutchings:
thanks @srallen, I'll finish this off tomorrow!

rogerhutchings:
This should be ready to go

rogerhutchings:
this should be ready to go @srallen

packages/app-project/src/shared/components/Stat/components/AnimatedNumber/AnimatedNumber.spec.js
```diff
@@ -1,10 +1,41 @@
...
   })
+
+  it('should render the `value` by default', function () {
+    const wrapper = render(<AnimatedNumber duration={DURATION} value={VALUE} />)
+    expect(wrapper.text()).to.equal(FORMATTED_VALUE)
+  })
+
+  describe('animation behaviour', function () {
+    it('should animate from zero on mount', function () {
+      const animateNumberSpy = sinon.spy(AnimatedNumber.prototype, 'animateValue')
```
srallen: You should call `animateNumberSpy.restore()` after the expectation runs, else this stays wrapped while the tests are run.

packages/app-project/src/shared/components/Stat/components/AnimatedNumber/AnimatedNumber.js
```diff
@@ -9,14 +9,15 @@ class AnimatedNumber extends Component {
...
+    this.animateValue(0)
   }
 
   componentDidUpdate (prevProps) {
     this.animateValue(prevProps.value)
   }
 
-  animateValue (prevValue = 0) {
+  animateValue (prevValue) {
+    const self = this
```
srallen: Non-blocking, but why did you need to do this?

rogerhutchings: Woops, not sure why it was marked resolved already. Anyway, it turns out `calledWithArgs` doesn't pick up default arguments

PR #828, opened 2019-05-15
Add check to Modal to see if rendering on client

### LINKED ISSUE #696: Modal component should check it's not rendering on the server
**Package**

react-component

The Modal component requires access to `document`, so if you try and have it open on the server, it'll break. We can add a state object to it, e.g. `{ client: false }`, have that set to true by `componentDidMount`, and only open the modal if both `client` and the `active` prop are `true`

srallen:
Package: lib-react-components

Closes #696

Describe your changes:
Adds a check and state for rendering only on client side for the Modal since it uses the Layer component. Layer uses `window.document` which is not available on the server. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
Could this be done in the functional component by checking for `window.document` with `useEffect()`?

srallen:
I'm not sure. I haven't dug into hooks yet.

rogerhutchings:
FYI I’m pretty sure this check exists in the various individual modals as well, so it may be worth removing them in this PR as well

srallen:
@rogerhutchings Maybe it should be a HOC?

srallen:
> FYI I’m pretty sure this check exists in the various individual modals as well, so it may be worth removing them in this PR as wel

I searched the src folder of app-project and lib-classifier for `document` and `window.document` and didn't find anything.

rogerhutchings:
I was sure this did exist: https://github.com/zooniverse/front-end-monorepo/tree/master/packages/app-project/src/components/AuthModals/shared/components/withOnlyRenderOnBrowser

PR #833, opened 2019-05-15
Set button gap for PlainButton

srallen:
Package: lib-react-components, lib-classifier, app-project

Describe your changes:
Sets the gap between the label and the icon for the PlainButton. Use the `size` prop on the icons to set their sizes. Styles the theme toggle to be like the other plain buttons and fixes the hover styling as noted in #738

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


srallen:
@rogerhutchings I see that underline issue in Chrome too. Maybe Chrome is buggy with the transform and underline rendering?

rogerhutchings:
That's what I thought. Swapping over to a `border` might fix it?

srallen:
It might. I'll open an issue.

PR #834, opened 2019-05-16
Refactor feedback for rule data numbers, modal styling

### LINKED ISSUE #831: Graph2dRange feedback breaks with numeric values for rule x, width and tolerance
**Package**
lib-classifier

Graph2dRange feedback breaks with numeric values for rule x, width and tolerance
See https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/feedback/strategies/datavis/graph2drange/reducer.js
**To Dos**

- [ ] create tests to catch bug
- [ ] add error logging
- [ ] check type, return or convert as needed

mcbouslog:
Package: lib-classifier

Closes #831, towards #832 

Describe your changes:
- adds tests for rule data as numbers
- updates rule reducer to use `parseFloat` as opposed to `parseInt`
- adds tests in FeedbackModal for showing subject viewer or not
- changes FeedbackModal styling for subject viewer and messages list 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/store/feedback/strategies/datavis/graph2drange/reducer.spec.js
```diff
@@ -2,89 +2,164 @@ import { expect } from 'chai'
...
-      id: '1234',
-      strategy: 'graph2dRange',
-      success: true,
-      successEnabled: true,
-      successMessage: 'Success!',
-      successfulClassifications: [annotationSuccess, annotationTolerance],
-      tolerance: '1',
-      width: '2',
-      x: '15'
+  describe('with rule data as numbers', function () {
```
eatyourgreens: There's a lot of repeated code from the previous block in this function, which could be moved into a shared function.
```javascript
describe('with rule data as strings', function () {
 testFeedbackRule(ruleWithStrings);
}
describe('with rule data as numbers', function () {
 testFeedbackRule(ruleWithNums);
}
```
where `testFeedbackRule` is a function that takes a mock feedback rule and passes it to a series of `it()` functions.

packages/lib-classifier/src/components/Classifier/components/Feedback/FeedbackModal.spec.js
```diff
@@ -5,6 +5,11 @@ import sinon from 'sinon'
 import { Button } from 'grommet'
 import FeedbackModal from './FeedbackModal'
 
+const applicableRules = [
+  { id: '1' },
+  { id: '2' }
+]
+
```
eatyourgreens: I could be wrong, but I thought the feedback viewer was set to only render a viewer if it was passed rules for the `graph2dRange` task. So I'm a bit confused as to how these mock rules will work.

mcbouslog:
Great questions/suggestions @eatyourgreens! I think this is looking better.

mcbouslog:
I think there are two issues with feedback when tested with the most recent TESS data:
1. rule data (x, width, tolerance) set as numbers as opposed to strings causes the feedback reducer to produce incorrect feedback information
2. lists of feedback messages greater than the modal size causes the LCV viewer to shrink to no height

This PR has tests and refactors the code for 1, and I think fixes 2, but I'm currently unable to reproduce a failing test for 2. I'm happy to continue trying, but know we want this fixed asap for additional beta testing.

packages/lib-classifier/src/components/Classifier/components/Feedback/FeedbackModal.spec.js
```diff
@@ -16,6 +22,28 @@ describe('FeedbackModal', function () {
     expect(wrapper.html()).to.be.null
   })
 
+  it('should show FeedbackViewer if hideSubjectViewer false and applicableRules', function () {
```
srallen: How about we add sub describe blocks for `when subject viewer is hidden` and for `when subject viewer is shown`? The it statements are a little hard to parse for this set.

packages/lib-classifier/src/store/feedback/strategies/datavis/graph2drange/reducer.spec.js
```diff
@@ -1,90 +1,82 @@
...
-    hideSubjectViewer: false,
-    id: '1234',
-    strategy: 'graph2dRange',
-    successEnabled: true,
-    successMessage: 'Success!',
-    tolerance: '1',
-    width: '2',
-    x: '15'
-  }
+const ruleWithStrings = {
```
srallen: We could make a feedback rules factory to make setup of this consistent in tests.

packages/lib-classifier/src/components/Classifier/components/Feedback/FeedbackModal.js
```diff
@@ -49,17 +49,21 @@ class FeedbackModal extends React.Component {
...
-                  <li key={Math.random()}>
-                    {message}
-                  </li>
-                )}
-              </ul>
+              {showViewer && (
+                  <Box height='100%'>
+                    <FeedbackViewer />
+                  </Box>)}
+              <Box overflow='scroll'>
```
srallen: Why don't we give this Box a fixed height if a viewer is displayed so the viewer can have more visual real estate? It's still very small... I arbitrarily picked 200px here (hopefully the markings go 100% if done properly in the code):

Screen Shot 2019-05-16 at 3 19 13 PM

PR #768, opened 2019-04-24
Finish up Publications

rogerhutchings:
Adds project logos to the publications page.

closes #756, #751

PR #829, opened 2019-05-15
WIP Add custom express server, lru caching

rogerhutchings:
Adds an LRU cache to speed up the publications page.

Note that it only works in production - to try it out locally, make sure you set `NODE_ENV` to `production`

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?
- [ ] Is the changelog updated?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?

PR #835, opened 2019-05-16
Set cache-control headers in production

rogerhutchings:
Adds a `Cache-Control` header in production, setting `max-age` to a year for JS files (which are cache-busted internally by filename), and 1 minute for other files.

This is to fix [a bug](https://github.com/zooniverse/front-end-monorepo/pull/768#issuecomment-492839022) noticed by @lcjohnso, where https://www.zooniverse.org/about/publications was getting 404s when trying fetch its JS bundles. This was caused by CloudFront caching the HTML for the page by default, causing it to try and fetch bundles that no longer exist when the app is redeployed.

[CF does respect any cache headers set upstream](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html#expiration-individual-objects), however, and since we can't set cache headers by object type in CF we have to do it in the server.


rogerhutchings:
Next thing to do is create a new distribution for `fe-*.zooniverse.org` subdomains so we can cache assets fetched during requests to `www.zooniverse.org`

camallen:
> CF does respect any cache headers set upstream, however, and since we can't set cache headers by object type in CF we have to do it in the server.

Can't CF paths match different mime targets? 
From the docs
> Specify which requests you want to route to the origin. For example, the path pattern "images/*.jpg" routes all requests for ".jpg" files in the images directory and in all subdirectories below the images directory to the Amazon S3 bucket or the web server that you specify in Origin.

packages/app-content-pages/server/server.js
```diff
@@ -22,3 +27,18 @@ app.prepare().then(() => {
     console.log(`> Ready on http://localhost:${port}`)
   })
 })
+
+function setCacheHeaders (req, res) {
+  let maxAge = 60 // 1 minute
```
camallen: Do you want these to be configured via ENV vars with a default fall back? I.e. change a deploy value instead of code to get changes into play?

rogerhutchings:
Okay, so yes we could 🤦‍♂I still need to dial down the expiry for the `/about/publications` path on the main zooniverse.org distribution, but I could the above on the fe-*.zooniverse.org distribution (which I'll create now)

packages/app-content-pages/server/server.js
```diff
@@ -22,3 +27,18 @@ app.prepare().then(() => {
...
+
+  if (isJsRequest(req)) {
+    maxAge = 31536000 // 1 year
+  }
+
+  res.setHeader('Cache-Control', `max-age=${maxAge}`)
+}
+
+function isJsRequest(req) {
+  const regex = /\.(?:js)$/i
```
camallen: The regex works, any reason for the non capture group? Could this be simplified to `/\.js$/i`? I'm probably missing something important here. 

Super minor - just thinking about this getting copied around / becoming the way it to check file extensions.

packages/app-content-pages/server/server.js
```diff
@@ -11,6 +11,11 @@ const handleNextRequests = app.getRequestHandler()
 app.prepare().then(() => {
   const server = new http.Server((req, res) => {
     app.setAssetPrefix(assetPrefix)
+
+    if (!dev) {
```
camallen: Does this carry over to staging? Is `dev` really indicative of all non-production environments?

packages/app-content-pages/server/server.js
rogerhutchings: No, that could absolutely be simplified to that

packages/app-content-pages/server/server.js
rogerhutchings: `dev` here is shorthand for `process.env.NODE_ENV !== production`. The `NODE_ENV` environment variable is set locally, and in production it's set in the `deployment.tmpl` file

camallen: How does caching work on staging env?

camallen: I assume there is no staging server for this app, right? Feel free to ignore this if that's the case but dev isn't just `!= 'production'` in most setups.

rogerhutchings: there isn't. that said, would it make more sense to actually have ENABLE_CACHE as an env var as well?

rogerhutchings:
Okay, following @srallen's comments on https://github.com/zooniverse/front-end-monorepo/pull/829, I want to add some tests for the server before merging this.

rogerhutchings:
@srallen can you take a look at the server tests please?

packages/app-content-pages/package.json
```diff
@@ -58,6 +58,7 @@
     "enzyme-adapter-react-16": "~1.11.2",
     "jsdom": "~13.0.0",
     "mocha": "~5.2.0",
+    "proxyquire": "^2.1.0",
```
srallen: Non-blocking, but the pesky ^ vs ~ strikes again.

packages/app-content-pages/server/configure-next-server.spec.js
```diff
@@ -0,0 +1,76 @@
...
+  it('should exist', function () {
+    expect(configureNextServer).to.be.a('function')
+  })
+
+  describe('assetPrefix setting', function () {
+    it('should default to an empty string', function () {
+      const assetPrefix = ''
+      configureNextServer(mockApp)
+      const result = mockApp.setAssetPrefix.calledWith(assetPrefix)
+      expect(result).to.be.true()
```
srallen: We've been using sinon-chai in other tests. It's a bit more readable, so this expectation instead could be written as: `expect(mockApp.setAssetPrefix).to.be.calledWith(assetPrefix)`

eatyourgreens:
```js
const prefixSpy = mockApp.setAssetPrefix.withArgs(assetPrefix)
expect(prefixSpy).to.have.been.called()
```
 would work too. I've been using `spy.withArgs()` in recent tests but I can't remember why it's preferred over `.calledWith()`. Maybe because it gives you a spy with call history etc. for those specific arguments?

srallen: Yeah, I think it matches the arguments first and then returns the call history only for that. I think it's useful if you already know that there may be many calls the spy is counting but only X for a specific argument and that's what you want to test for. However, I think in this case, they're basically doing the same thing.

PR #837, opened 2019-05-17
Add consistent-return rule to linter

eatyourgreens:
Package:
all

Adds the[ consistent-return rule](https://eslint.org/docs/rules/consistent-return), which requires functions that return values to return consistent values. It should warn about missing return statements on functions that return from inside blocks.

Updates `app-content-pages` to use the same linter config as other packages.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/tools-standard/options.js
+++ b/packages/tools-standard/options.js
@@ -18,6 +18,9 @@ var opts = Object.assign({}, stdOpts, {
     plugins: [
       'jsx-a11y'
     ],
+    rules: {
+      "consistent-return": "error"
+    },
     useEslintrc: false
   },
   homepage: pkg.homepage,
--- a/packages/tools-standard/options.js
+++ b/packages/tools-standard/options.js
@@ -12,6 +12,9 @@ var opts = Object.assign({}, stdOpts, {
     envs: [
       'mocha'
     ],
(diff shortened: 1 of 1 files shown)


eatyourgreens:
I notice it's complaining about functions where early returns are used to exit if args aren't present.
ie.
```javascript
function doSomething(args) {
 if (!args) return defaultAction()
 newState = getStateFromArgs(args)
}
```
instead of
```javascript
function doSomething(args) {
 if (!args) {
   defaultAction()
 } else {
  newState = getStateFromArgs(args) 
 }
}
```
Personally I find the second style clearer but that might just be me.

srallen:
Was this  a part of #898? Can this close?

eatyourgreens:
It wasn't in #898 but I can rebase this PR to add this rule to the new config. I think it will warn about any function that returns early without also having a return statement at the end of the function.

srallen:
Ok, sounds good to me

eatyourgreens:
@srallen should be ready now. I also enabled the recommended accessibility checks, which I think you have to do before the linter will run them. At least, we had that `extends` line in our old eslintConfig for the project and the classifier.

PR #863, opened 2019-05-22
Temporarily disable the stats widget

rogerhutchings:
Temporarily disables the stats widget, as it's not really required for the TESS production launch.

cc #38 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


rogerhutchings:
I talked to @camallen about setting up feature flags earlier, so once that's in place I suggest we use that - in the meantime we can just toggle `false` to `true` :P

PR #866, opened 2019-05-22
Add random number to data-mask ID

### LINKED ISSUE #851: LCV: Feedback doesn't always render all of the json points
**Package** lib-classifier

>  it looks like it just didn’t load properly because the LC shouldn’t finish at 18 days? And the data points look like they’ve been cut-off.

> I’ve been able to reproduce this, on the very first subject I got. I marked a transit at ~20 days to see what would happen.

Screen Shot 2019-05-17 at 11 42 03 AM

image

mcbouslog:
Package: lib-classifier

Closes #851 .

Describe your changes:
- the data-mask from the initial subject viewer was being applied to the feedback viewer, this adds a random number to the data-mask id to differentiate
- small edits to Feedback Modal sizing

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/Feedback/FeedbackModal.js
+++ b/packages/lib-classifier/src/components/Classifier/components/Feedback/FeedbackModal.js
@@ -38,7 +38,7 @@ class FeedbackModal extends React.Component {
     if (showViewer) {
       FeedbackViewer = getFeedbackViewer(applicableRules)
     }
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Looks good to me but you could add `id` as an optional prop to the viewer. Use the supplied ID if present, otherwise generate one.

Issue #876, opened 2019-05-23
Publications Page: issues with new content


lcjohnso:
Two issues cropped up when I tried to add new content for the new publications page via Contentful:

1) I tried to add a new Category of "Social Science", but when it rendered it included a translation error.  I'll note that all the Contentful content is tagged with "en-US" by default (and there are no options to alter this property).
![screenshot_publications-categories](https://user-images.githubusercontent.com/2006101/58273428-53e9c700-7d56-11e9-9cbc-ef117a498ddb.png)
For now I unpublished the category to remove the buggy category (also, "Social Science" is not totally necessary at this point (only one publication that is cross-listed in two other categories), so no immediate attention is needed for this bug.

2) For one of the new publications I added a new project (Sounds of New York City) along with the publication entry.  The project-specific avatar is not loading (showing default Zooniverse avatar).  Tried on Chrome and Firefox, in and out of private browsing, but not appearing.  When adding a new publication, is something additional required to trigger the avatar image load?

rogerhutchings:
1. I didn't think of this. And actually, since Contentful supports translations, I should just get the translated titles from there. I'll write a fix.
2. There is, it requires the project ID. There is a field for this in Contentful, but it's non-required since there's at least one project that doesn't have a project on Panoptes. I'll add a note to the help field and add a note to Stack Overflow to explain it.

lcjohnso:
Re: #2 - When I added the new Project entry, I did include the Project ID (for SONYC, 5352), but no avatar appeared.

rogerhutchings:
Right - I'll try that out tomorrow and see if I can see what's going on

rogerhutchings:
Out of interest, is the publication data all still in Contentful? If not, would you mind adding it here?

camallen:
> but it's non-required since there's at least one project that doesn't have a project on Panoptes.

which project doesn't have an id?

rogerhutchings:
Reverse the Odds

camallen:
what this hard cell? a mobile app the CRUK developed on our old API platform

eatyourgreens:
I'm not sure if it's Hard Cell, but it was a CRUK project: https://daily.zooniverse.org/2014/10/03/reverse-the-odds/

lcjohnso:
[...]

PR #878, opened 2019-05-23
Use contentful category titles

rogerhutchings:
cc #876

This changes the store to return the category titles from Contentful, rather than translate them in the app. This means that everything on that page can be edited via the CMS.

Once a language selector is in place, we'll need to send the language code along with the request in Contentful, or store translations locally (probably more likely) - this is captured in #877

### DIFF
--- a/packages/app-content-pages/src/screens/Publications/Publications.js
+++ b/packages/app-content-pages/src/screens/Publications/Publications.js
@@ -22,6 +22,17 @@ const FORM_URL = 'https://docs.google.com/forms/d/e/1FAIpQLSdbAKVT2tGs1WfBqWNrMe
 function Publications (props) {
   const { className, data, filters } = props
 
+  // `Show all` isn't a Contentful category, so we need to manually add the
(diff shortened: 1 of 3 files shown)


camallen:
I can't help but feel like this might make life harder for us in the future as it's fragmenting our translations setups.

rogerhutchings:
Yes, it would be fragmenting translations. However:

 - We're talking about 5-10 proper nouns, so manually adding these wouldn't be too much pain (especially since to my knowledge, I think this page only has one other translation)
- The point of using a CMS was to get away from requiring people to submit PRs to change stuff like this
- While it does split up where translation strings are stored, it could still use a common interface (with some changes to Pandora).

I think it's worth it.

eatyourgreens:
Just noting that Pandora is an app specifically for project translators, so doesn't handle UI labels anyway.

rogerhutchings:
I've just tested this out again locally, and it looks fine - can someone give it a 👍 or 👎?

PR #880, opened 2019-05-24
Move store connection from LCV to LCV container

srallen:
Package: lib-classifier

Describe your changes:
I see this as steps towards #650 and #680. The LCV is now a component that is only concerned with how to render itself using D3. We can now move it into a separate package of subject viewer view components and begin to define a standard set of props that those subject viewers will expect.

I've separated out the export of the LCV container and its decorated version of itself with the HOCs. This seemed to fix the issues of the tests running.

All of the tests are passing and functionally it seems to be working in manual tests, but I've left it as WIP to figure out a good way to add store mapper function tests for the connected LCV container.
 
# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


rogerhutchings:
The only way that I can really think of is to just export the `storeMapper` function as well, pass in a dummy store object and see what you get out. Depending on the size, you could also have a separate spec for it (`LightCurveViewerContainer.storeMapper.spec.js`)?

rogerhutchings:
Also, this looks great

eatyourgreens:
You could add a Provider to the tests and check that expected values are passed in from a mock store.
https://github.com/zooniverse/front-end-monorepo/blob/ed1a5ae79f8108ddba9a33cdcb791f1f2560f5f0/packages/lib-classifier/src/components/Classifier/components/withKeyZoom/withKeyZoom.spec.js#L27-L31

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -94,42 +128,82 @@ class LightCurveViewerContainer extends Component {
...
 }
 
-LightCurveViewerContainer.wrappedComponent.propTypes = {
+LightCurveViewerContainer.propTypes = {
+  addAnnotation: PropTypes.func,
+  drawFeedbackBrushes: PropTypes.func,
+  interactionMode: PropTypes.oneOf(['annotate', 'move']),
   loadingState: PropTypes.string,
   onError: PropTypes.func,
+  onKeyDown: PropTypes.func.isRequired,
```
eatyourgreens: Why is `onKeyDown` required?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -94,42 +128,82 @@ class LightCurveViewerContainer extends Component {
...
   onReady: PropTypes.func,
+  setOnPan: PropTypes.func.isRequired,
+  setOnZoom: PropTypes.func.isRequired,
```
eatyourgreens: Why are these two callbacks required too? Can't they default to functions that return null or true?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
```diff
@@ -32,12 +31,15 @@ const failSubject = Factory.build('subject', {
 
 describe('Component > LightCurveViewerContainer', function () {
   it('should render without crashing', function () {
-    wrapper = shallow(<LightCurveViewerContainer.wrappedComponent />)
+    wrapper = shallow(<LightCurveViewerContainer onKeyDown={() => {}} setOnPan={() => {}} setOnZoom={() => {}} />)
```
eatyourgreens: The tests run without these callback props so I'm not sure that they're needed.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
srallen: They didn't have defaults previously, so just being consistent about setting props that don't have defaults as required. It makes sense to me to have these be required later on when the view component itself gets put into its own library since the LCV functionally needs a zoom function to be set to be usable.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
srallen: The tests run, but then we get the props warnings about them not being defined since I now set them as required, which is annoying.

eatyourgreens: Yeah, and the component renders happily without them so I'm not seeing why they're required in order for it to run. My interpretation would be that the component should error if required props aren't present when it tries to render.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -94,42 +128,82 @@ class LightCurveViewerContainer extends Component {
...
+      enableAnnotate,
+      enableMove,
+      feedback,
+      interactionMode,
+      onKeyDown,
+      setOnPan,
+      setOnZoom,
+      subject,
+      toolIndex,
+    } = this.props
```
eatyourgreens: The LCV has a scary number of props. That runs through my mind whenever I work on it.

srallen: Yep, next steps will be refactor on some of this. Inevitable for this step though.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.spec.js
srallen: I'd like to refactor it to error if a zoom function isn't set. That's a next step I think.

srallen:
I've been experimenting with testing the store connection, so when I have something I'll open another PR. I'll go ahead and merge this one as is.

Issue #869, opened 2019-05-22
Build registration form


srallen:
**Package** app-project

**Feature or Issue Description**
The registration form is missing and needs to be built.

rogerhutchings:
I don't think we have any sign in / register modal designs - @beckyrother?

beckyrother:
Not specifically. It sounds like Grommet had some components that could work so I haven't put anything together; happy to do so if it helps.

As a reminder this is pinned in #design-thoughts: https://docs.google.com/document/d/131LXuIlBGMaaPW7bd7SvFRReeLsczXfMtCUNCnOucpI/edit#heading=h.i2qn31h1i89a

rogerhutchings:
There are form components, yeah, but it would be nice to have a design (for consistency if nothing else)

beckyrother:
Sign in: https://invis.io/UAN6AP0JTBC#/365296253_00-31_Sign_In_
Register: https://invis.io/UAN6AP0JTBC#/365299013_00-33_Register_Wide 
Register with forms filled in and error message: https://invis.io/UAN6AP0JTBC#/365302900_00-331_Register_Wide_Erros

beckyrother:
@rogerhutchings it turned out to be easier to just make the dark theme and upload it: https://invis.io/UAN6AP0JTBC#/365807868_08-9_Dark_Register_
that pink/red is #FFB6AA

PR #881, opened 2019-05-24
Move Tabs to lib-react-components

rogerhutchings:
Moves the Tabs component from the classifier into the shared components, as it's also used in the new auth modal designs.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


rogerhutchings:
cc #869

PR #892, opened 2019-05-30
Basic feedback reducer

rogerhutchings:
Adds a basic reducer to show each feedback message once, with a number of matches next to it.

Screenshot 2019-05-30 15 22 56

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/Feedback/FeedbackModal.js
+++ b/packages/lib-classifier/src/components/Classifier/components/Feedback/FeedbackModal.js
@@ -7,6 +7,7 @@ import counterpart from 'counterpart'
 import en from './locales/en'
 
 import getFeedbackViewer from './helpers/getFeedbackViewer'
+import reduceFeedbackMessages from './helpers/reduceFeedbackMessages'
 
 counterpart.registerTranslations('en', en)
 
@@ -35,10 +36,14 @@ class FeedbackModal extends React.Component {
     const { applicableRules, hideFeedback, hideSubjectViewer, messages, showModal } = this.props
     const showViewer = !hideSubjectViewer && applicableRules && applicableRules.length > 0
     let FeedbackViewer = null
+
     if (showViewer) {
       FeedbackViewer = getFeedbackViewer(applicableRules)
     }
+
     const messageContainerHeight = showViewer ? '400px' : '100%'
+
+    const reducedMessages = reduceFeedbackMessages(messages)
 
     if (showModal) {
       return (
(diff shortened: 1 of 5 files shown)


rogerhutchings:
In this case, no - the message is the same regardless of success or failure, or I would have done that.

eatyourgreens:
Could we reduce `applicableRules` to get the number of successes/failures? I'd have to go back and look at the code but I think messages are reduced down from that.

rogerhutchings:
I assumed not, as you wouldn't need to pass in the messages as a separate prop, but I'll take a look :+1:

rogerhutchings:
Update:

- The TESS feedback messages aren't _exactly_ the same:

    Success:

    > "Awesome! The green column shows where you’ve correctly marked a simulated planet transit! A red column means you missed one or drew a column slightly too large (try to only capture the transiting points in the columns that you draw), but don’t be discouraged, some are very difficult, or even impossible, to spot. This can happen when the planet is very small compared to its host star, meaning that the transit is very shallow and easily lost in the varying brightness of the star. You’re doing great, keep going!"

    vs failure: 

    > "A green column indicates that you’ve correctly marked a simulated planet transit! A red column means thst you missed one or that you drew a column slightly too large (try to only capture the transiting points in the columns that you draw), but don’t be discouraged, some are very difficult, or even impossible, to spot. This can happen when the planet is very small compared to its host star, meaning that the transit is very shallow and easily lost in the varying brightness of the star. You’re doing great, keep going!"

- `applicableRules` doesn't include results: 

    ```
    {
      failureEnabled: true
      failureMessage: "A green column indicates that you’ve correctly marked a simulated planet transit! A red column means thst you missed one or that you drew a column slightly too large (try to only capture the transiting points in the columns that you draw), but don’t be discouraged, some are very difficult, or even impossible, to spot. This can happen when the planet is very small compared to its host star, meaning that the transit is very shallow and easily lost in the varying brightness of the star. You’re doing great, keep going!"
      hideSubjectViewer: false
      id: "11110"
      strategy: "graph2drange"
      successEnabled: true
      successMessage: "Awesome! The green column shows where you’ve correctly marked a simulated planet transit! A red column means you missed one or drew a column slightly too large (try to only capture the transiting points in the columns that you draw), but don’t be discouraged, some are very difficult, or even impossible, to spot. This can happen when the planet is very small compared to its host star, meaning that the transit is very shallow and easily lost in the varying brightness of the star. You’re doing great, keep going!"
      tolerance: 0.260563125
      width: 0.3
      x: 0.596792229294433
    }

- Combining the `applicableRules` and `messages` we _could_ get the number of successes and failures (assuming there are no rules where the same text is used for both messages, or the success message in one rule matches the failure rule of another). But there is the argument that `success` and `failure` may have stronger connotations than we'd want; `matches` seems acceptably neutral.

eatyourgreens:
The problem I have with using matches is that it gives a false impression of success. The feedback example shown at the top of this PR shows 10 red misses on the light curve, but the text says '10 matches'. We can try that out with volunteers, though, to see if they find it confusing.

PR #893, opened 2019-05-30
Build out RegistrationFormContainer for validations

srallen:
Package: app-project

A WIP build out of the async API requests and validations. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


srallen:
I'm a bit stuck with the asynchronous validations. When the validation function was just the synchronous check for password matching, the error object was returned as expected and the error displayed in the DOM. However, with the async check, the error object is always empty.... I'm definitely doing something wrong, just not seeing it late in the day. FYI, link to formik's docs on validation: https://jaredpalmer.com/formik/docs/api/formik#validate-values-values-formikerrors-values-promise-any

srallen:
Also the actual onSubmit handler still needs updating.

srallen:
Ok, we now have an actual onSubmit handler and working validations. This just needs a general error catching handler. And tests.

srallen:
I think this is ready for review. This still doesn't have tests, but some of them are coming in a separate PR, so I'll hold off for now.

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.js
```diff
@@ -57,16 +59,27 @@ function RegisterForm (props) {
...
+            <Text>{counterpart('RegisterForm.underageConsent')}{' '}(<PrivacyPolicyLink />)</Text> :
+            <Text>{counterpart('RegisterForm.privacyAgreement')}{' '}(<PrivacyPolicyLink />)</Text>
+          const emailListSignUpLabel = (values.underageWithParent) ?
+            <Text>{counterpart('RegisterForm.underageEmail')}</Text> :
+            <Text>{counterpart('RegisterForm.emailListSignUp')}</Text>
+
+          return (
             <Box as='form' onSubmit={handleSubmit} margin={{ top: 'small' }}>
+              {generalError && 
+                <Text color={{ light: 'status-error', dark: 'status-error' }}>{generalError}</Text>}
```
eatyourgreens: This should maybe go outside the form tag. Is it showing error messages from the store, rather than error messages for a particular form input?

eatyourgreens: [MDN suggests adding `role="alert"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/alerts) too, which might announce the error in screenreaders without them having to realise they must go back and read the form again.

srallen: > Is it showing error messages from the store, rather than error messages for a particular form input?

This is the general error catch for the registration request after validations pass, so it's not from the store. I store any errors from that request in the container component state and display them at top. I'm imagining this might be like 400 or 500 errors, so not a specific input either unless someone somehow is able to attempt the registration on a really old browser that doesn't support the built in validations? Unfortunately if it is a specific input error coming back from the API, the panoptes-javascript-client joins the error array into a string which makes it really hard to parse. Notice the regex parsing I'm doing because of it for the other errors :(

Would you prefer we store these kinds of errors in the user store?

eatyourgreens: No, I was just wondering what the `generalError` prop referred to. Sounds like it should definitely be marked up as an alert, outside the form tag.

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterFormContainer.js
```diff
@@ -1,33 +1,126 @@
...
+      error: ''
+    }
+  }
+
+  async validate (values) {
+    const { email, password, passwordConfirm, username } = values
+    let errors = {}
+
+    errors.username = await this.checkNameForConflict(username)
+    errors.email = await this.checkEmailForConflict(email)
```
eatyourgreens: You're sending the register request twice here. Is there a reason not to register once, catch the error and then parse it?

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterFormContainer.js
```diff
@@ -1,33 +1,126 @@
...
+      login: username,
+      password,
+      project_email_communication: emailListSignUp
+    }
+
+    if (store && store.project && store.project.id) {
+      valuesToSubmit.project_id = store.project.id
+    }
+
+    authClient.register(valuesToSubmit)
```
eatyourgreens: Haven't we already made this API call during the validation?

rogerhutchings:
Wouldn't it be better to have rapid feedback on taken usernames / emails? I'm guessing that this wouldn't happen fast enough to hammer the API too badly

eatyourgreens:
True. Is `validate` run on every change to a field? The Formik docs say it is but that seems like it would generate excessive API traffic. @camallen any thoughts on repeated calls to the registration endpoint in order to check for existing usernames/emails?

rogerhutchings:
It's currently debounced to 1 / second: https://github.com/zooniverse/front-end-monorepo/pull/893/files#diff-7bca13c270676b39030a5333e185b45cR17

eatyourgreens:
My reading of the code is that it validates the whole form again on any change, which will make two calls to the registration endpoint. Am I missing something about Formik?

eatyourgreens:
A screenshot might help. When I fill out the registration form, I'm seeing repeated requests to `/users` and `/users/sign_in`. It seems like my username and email are being checked multiple times, even if they haven't changed.
A screenshot of the network requests panel showing repeated redundant requests from the registration form.

srallen:
I was thinking about this over the weekend and I think we might want to switch away from the debounced async checks because it seems like formik will drop built in support for async validations in the next major version: https://github.com/jaredpalmer/formik/issues/1524

srallen:
@eatyourgreens you are correct that the validation for those fields are running more often than they should because of the way the validations run with the formik API. I could potentially add a conditional check to see if the input is what we expect for each validation, but I'm leaning toward dropping the debounced async validation and rely on the response from the submission.

eatyourgreens:
@srallen I think Formik let’s you pass a `validate` callback to individual fields too, but it sounds like async support for that will also go away.

If we want instant feedback for email/username then debounced change handlers on those specific inputs might work too (separate from the form validation.) Depends whether we want people to submit the form before finding out their username is taken. That might be irritating.

eatyourgreens:
I think this is good to go if the redundant API requests are cut down, and if the general form error text is made accessibile to screenreaders.

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.js
eatyourgreens: Thinking about it, isn’t there a Grommet component for alerts/notifications in screenreaders? I can’t remember its name but I’m sure I’ve seen it.

srallen:
Yeah, I think I can at least cut down on the number of requests as a first pass on improving it. However, I'm not sure how much users expect nearly instant feedback on these kind of validations vs just get it on submission, basically, is it irritating? Dropping it will make the code for the form a lot simpler.

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.js
eatyourgreens: Here it is: https://v2.grommet.io/announcecontext

srallen: I'll try that out on another pass at this, but for now I've added the `role="alert"` and moved it outside of the form.

srallen:
Ok I've updated the async validations to only happen for the username and email inputs debounced during the onChange handler, so there are less API calls, but not as minimal as it could be if we just moved to just simply using the submission handler. Because I've moved the async validations out of the Formik validate function, though, this will be v2 ready in that sense.

I've tidied up the styling a bit more too. I'll post screenshots for review in a follow up comment.

One last thing though, is that the submission itself isn't actually working. The form doesn't seem to be calling onSubmit anymore and I'm not seeing what's wrong. I could use some help on this one since I'm a bit fried by this point.

srallen:
Massive screenshot post for @beckyrother to review. A few things to note: some of the validations I've written to use built in browser validation, this saves us a lot of time, but the trade off is that we can't style the messages. Using built in browser validation also gives us translations for free (no dev effort). 

A few of the validations are done in javascript: username availability, password confirmation, email availability, email confirmation, and privacy policy agreement. These are styled as you'd like.

Also, one more thing to note is that we had to make a change a few months ago to explicitly label which fields are required and which are optional (https://github.com/zooniverse/Panoptes-Front-End/issues/4539), so this has to continue here. I think this might be a GDPR requirement? (Or some other UK legal requirement?)

### Initial form view
Screen Shot 2019-06-03 at 3 33 36 PM

Screen Shot 2019-06-03 at 3 48 29 PM

### Username validations

#### Browser validation on allowed characters (shows on submit)
Screen Shot 2019-06-03 at 3 34 03 PM

Screen Shot 2019-06-03 at 3 30 40 PM

#### Javascript validation for already used usernames

Screen Shot 2019-06-03 at 3 47 15 PM

Screen Shot 2019-06-03 at 3 30 52 PM

### Password validations

#### Browser validations for password length

Screen Shot 2019-06-03 at 3 54 20 PM

#### Javascript validation for passwords not matching

Screen Shot 2019-06-03 at 3 34 19 PM

Screen Shot 2019-06-03 at 3 31 18 PM

### Email validations

#### Browser validations for email presence and format

Screen Shot 2019-06-03 at 3 34 42 PM

Screen Shot 2019-06-03 at 3 31 37 PM

Screen Shot 2019-06-03 at 3 51 36 PM

Screen Shot 2019-06-03 at 3 31 58 PM

#### Javascript email confirmation validation

Screen Shot 2019-06-03 at 3 35 03 PM

Screen Shot 2019-06-03 at 3 32 13 PM

#### Javascript email check for availability

Screen Shot 2019-06-03 at 3 35 55 PM

Screen Shot 2019-06-03 at 3 54 53 PM

beckyrother:
These look great @srallen thank you!! I'm totally fine with the built-in validation, it still looks nice and makes sense to not have to rebuild things that already exist. 

Only question: will the validation occur as soon as the field is active, or will it only appear after a user has clicked into and then out of the field? I ask because as a user it's frustrating for a form to ask you to enter a valid email address while you're still typing your email address.

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.js
eatyourgreens: No problem. I'm not even sure if that Grommet component does what I thought it did (marks a text element as a live region.)

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterFormContainer.js
```diff
@@ -91,7 +93,23 @@ class RegisterFormContainer extends Component {
     return regex.test(errorMessage)
   }
 
+  async onChange (event, formErrorHandler) {
+    const { target: { name, value } } = event
+    if (name === 'username') {
+      this.checkNameForConflict(value)
+        .then((errorMessage) => {
+          if (errorMessage) formErrorHandler('username', errorMessage)
+        })
```
eatyourgreens: Should this block be using `await`? Something like:
```
const errorMessage = await this.checkNameForConflict(value)
```

camallen:
>However, I'm not sure how much users expect nearly instant feedback on these kind of validations vs just get it on submission, basically, is it irritating? Dropping it will make the code for the form a lot simpler.

I'm not sure we've ever show the UX is better with JS email validation while you are filling in the form. Personally I prefer one validation round trip on submit button and API wise serving less requests is always better and I'd vote for form submission only, especially if it simplifies the front end code base.

How often do we foresee users signing up with clashing email addresses? Login field could certainly clash more considering the restrictions we have on them but folks to tend to use unique handles as well.

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.js
```diff
@@ -57,16 +71,26 @@ function RegisterForm (props) {
           handleBlur,
           handleSubmit,
           isSubmitting,
-          touched,
-          values,
-          /* and other goodies */
-        }) => (
+          setFieldError,
+          values
+        }) => {
```
eatyourgreens: This function should probably be extracted out into its own component.

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterFormContainer.js
```diff
@@ -1,33 +1,153 @@
...
+    }
+  }
+
+  validate (values) {
+    const { email, emailConfirm, password, passwordConfirm, privacyAgreement } = values
+    let errors = {}
+    errors.emailConfirm = this.validateEmailConfirmation(email, emailConfirm)
+    errors.passwordConfirm = this.validatePassword(password, passwordConfirm)
+    errors.privacyAgreement = this.validatePrivacyAgreement(privacyAgreement)
+    return errors
```
eatyourgreens: This is blocking `onSubmit` because Formik expects it to be an empty object for a valid form.

packages/app-project/src/components/AuthModal/components/LoginForm/LoginForm.spec.js
```diff
@@ -1,16 +1,34 @@
...
-    wrapper = shallow(<LoginForm />)
+    wrapper = shallow(<LoginForm onSubmit={onSubmitStub} />)
   })
 
   it('should render without crashing', function () {
     expect(wrapper).to.be.ok()
   })
+
+
+  it('should call the `onSubmit` prop on submission', function () {
```
eatyourgreens: If I'm understanding Formik correctly, then I think this test is testing Formik, since the form calls Formik's `handleSubmit` on submission.

`wrapper.find(Formik).simulate('submit', fakeEvent)` should call Formik's `onSubmit` prop, according to the enzyme docs.

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
```diff
@@ -0,0 +1,86 @@
...
+  before(function () {
+    onSubmitStub = sinon.stub()
+    wrapper = shallow(<RegisterForm onSubmit={onSubmitStub} />)
+  })
+
+  it('should render without crashing', function () {
+    expect(wrapper).to.be.ok()
+  })
+
+  it('should call the `onSubmit` prop on submission', function (done) {
```
eatyourgreens: Same comment about testing Formik here.

packages/app-project/src/components/AuthModal/components/LoginForm/LoginForm.spec.js
rogerhutchings: Isn't it also testing that the component is passing its `onSubmit` prop to the form (regardless of Formik)?

eatyourgreens: That function doesn't get passed to the form. The form calls `handleSubmit`, then Formik calls `onSubmit`.

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterFormContainer.js
eatyourgreens: There should also be a test that expects `validate` to return an empty object for valid forms.

packages/app-project/src/components/AuthModal/components/LoginForm/LoginForm.spec.js
eatyourgreens: I think it is slightly confusing that the form is decorated with Formik in the view component. Elsewhere, I think we've used a pattern where decorators/HOCs are applied in the container. That threw me a bit, initially, when I was reviewing this.

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.js
```diff
@@ -159,77 +208,86 @@ function RegisterForm (props) {
...
+  generalError: PropTypes.string,
+  onChange: PropTypes.func,
+  onSubmit: PropTypes.func.isRequired,
+  validate: PropTypes.func
 }
 
 RegisterForm.defaultProps = {
+  generalError: '',
+  onChange: () => true,
+  validate: () => true,
```
eatyourgreens:
```suggestion
  validate: () => {},
```

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
```diff
@@ -0,0 +1,86 @@
...
+    const form = wrapper.find(Formik).dive().find({ as: 'form' })
+
+    form.simulate('submit', {
+      preventDefault: () => { }
+    })
+
+    // on submit wasn't being called without the timeout
+    window.setTimeout(() => {
```
eatyourgreens:
```suggestion
    onSubmitStub.callsFake(() => {
```

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
```diff
@@ -0,0 +1,86 @@
...
+      expect(onSubmitStub).to.have.been.calledOnce
+      console.log(onSubmitStub.callCount)
+      done()
+    }, 0)
```
eatyourgreens:
```suggestion
    })
```

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
eatyourgreens: Calling `done()` after `onSubmit` runs would be more reliable than a timeout.

packages/app-project/src/components/AuthModal/components/LoginForm/LoginForm.spec.js
srallen: I was trying to follow their docs on how to test: https://github.com/jaredpalmer/formik/blob/f5d76609b222ce583663add279ac76e807e2d0ba/README.md#testing-formik

However, I had trouble applying the example to how we have it setup. Others have been having trouble too: https://github.com/jaredpalmer/formik/issues/937 

And tried out the timeout suggested in one of the comments. I don't think this is ultimately the right way to go.

srallen:
> The 'under 16' checkbox should be at the beginning of the form, and 'email address' should be 'Parent/guardian's email address' if that box is ticked.

@camallen Is having the under 16 checkbox at the top a legal requirement? I believe @beckyrother purposely moved it below as part of her design.

srallen:
> will the validation occur as soon as the field is active, or will it only appear after a user has clicked into and then out of the field? I ask because as a user it's frustrating for a form to ask you to enter a valid email address while you're still typing your email address.

@beckyrother it depends on the validation, some of them are synchronous while the input is being typed into, so will appear almost instantly, some are asynchronous and are delayed 1 second such as the username and email availability. All of the built in browser validations happen when the submit button is clicked.

I think I would prefer the username and email availability validation to happen on submit as well and it sounds better for the API to do this, so I think I will refactor at least those two to be on submit event validations.

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterFormContainer.js
srallen: Oh yeah, I think I started switching it to async/await and didn't finish. Thanks.

packages/app-project/src/components/AuthModal/components/LoginForm/LoginForm.spec.js
eatyourgreens: If you name the anonymous functional component that renders the form, I think you could find that by name rather than using `dive()`. It still feels like this test is triggering `handleSubmit` then testing that it calls `onSubmit`, which I'd expect to be tested by Formik anyway.

eatyourgreens:
The 'under 16' checkbox changes the form if it's ticked. I think it's at the top so that those changes are made before you fill out the form ie. to make sure you fill out the 'under 16' version of the form.

camallen:
> > The 'under 16' checkbox should be at the beginning of the form, and 'email address' should be 'Parent/guardian's email address' if that box is ticked.
> 
> @camallen Is having the under 16 checkbox at the top a legal requirement? I believe @beckyrother purposely moved it below as part of her design.

My understanding is we got approval for the UI components as they are, any changes would probably have to go through another review. @mrniaboc can provide clarity here.

mrniaboc:
It would be great if we could keep it at the top as this is what we agreed with our ethics board. Also, it saves an under16 from filling everything out first then realising at the end that they should have ticked that box and seen the text changes.

eatyourgreens:
Ticking the 'under 16' box here, the form visibly jumps so I think there are other text changes than the label for the email field that I flagged up.
https://signin.zooniverse.org/users/sign_up

eatyourgreens:
If you search through the old coffeescript form for `@state.underAge`, I think that will find all the bits that are different in the under-16 sign-up form. 
https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/partials/register-form.cjsx

beckyrother:
Hi all. I did indeed move the 'under-16' box to under the form as @srallen said, to keep the tick boxes together. When we were asked to add this box, there wasn't a lot of time for design thinking, so the top made the most sense at the time. If it's now too late to move it, we can certainly put it back at the top. Happy to provide a new invision if needed.

srallen:
Ok the under 16 checkbox is now on top and I think I got all the language switches, but it'd be good for someone else to check:

Normal:

Screen Shot 2019-06-04 at 2 53 46 PM

Under 16:

Screen Shot 2019-06-04 at 2 53 59 PM

We might want to increase the max width of the container so that it doesn't have to overflow scroll with the under 16 text which hides the submit button. What do you think @beckyrother ?

beckyrother:
Looks great! Thanks @srallen

beckyrother:
Would it be worth it to add 'your' to the 'real name' label, to make sure it's clear that it's not the parent/guardian's real name?

srallen:
@beckyrother I don't know. @mrniaboc's comment suggests to me that if we made that text change, it'd require another review by their ethics board.

srallen:
Closing. The merge conflict is nasty and the PR is grown to be big, so I'm going to break this up.

PR #894, opened 2019-05-31
Add basic input tests for registration form

rogerhutchings:
Adds some basic input tests to the registration form - I'll circle back to this later on @srallen

### DIFF
--- a/packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
+++ b/packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
@@ -5,7 +5,7 @@ import RegisterForm from './RegisterForm'
 
 let wrapper
 
-describe.only('Component > RegisterForm', function () {
+describe('Component > RegisterForm', function () {
   before(function () {
     wrapper = shallow(<RegisterForm />)
   })
--- a/packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
+++ b/packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
@@ -1,0 +1,65 @@
+import { shallow } from 'enzyme'
+import React from 'react'
+
+import RegisterForm from './RegisterForm'
+
+let wrapper
+
+describe.only('Component > RegisterForm', function () {
+  before(function () {
+    wrapper = shallow(<RegisterForm />)
+  })
+
+  it('should render without crashing', function () {
+    expect(wrapper).to.be.ok()
+  })
+
+  it('should call the `onSubmit` prop on submission', function () {
+    // https://github.com/jaredpalmer/formik/blob/f5d76609b222ce583663add279ac76e807e2d0ba/README.md#testing-formik
+  })
+
+  describe('fields', function () {
+    let rendered
+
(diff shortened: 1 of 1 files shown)


packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    it('should have a password field', function () {
+      const input = rendered.find('[name="password"]')
+      expect(input).to.have.lengthOf(1)
+      expect(input.attr('type')).to.equal('password')
+      expect(input.attr('required')).to.equal('required')
+      expect(input.prop('minlength')).to.equal('8')
+    })
+
+    it('should have a passwordConfirm field', function () {
+      const input = rendered.find('[name="password"]')
```
srallen: Wouldn't this be the same input from the above test. Could you find by id to be sure it's the one intended to be tested?

eatyourgreens: I think the name is wrong. It should be `passwordConfirm` for the confirmation input?

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
```diff
@@ -0,0 +1,65 @@
...
+      expect(input.attr('type')).to.equal('password')
+      expect(input.attr('required')).to.equal('required')
+      expect(input.prop('minlength')).to.equal('8')
+    })
+
+    it('should have an email field', function () {
+      const input = rendered.find('[name="email"]')
+      expect(input).to.have.lengthOf(1)
+      expect(input.attr('type')).to.equal('email')
+      expect(input.attr('required')).to.equal('required')
```
eatyourgreens: Should there be an input to confirm the email address too?

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
```diff
@@ -0,0 +1,65 @@
+import { shallow } from 'enzyme'
+import React from 'react'
+
+import RegisterForm from './RegisterForm'
+
+let wrapper
+
+describe.only('Component > RegisterForm', function () {
```
eatyourgreens:
```suggestion
describe('Component > RegisterForm', function () {
```

packages/app-project/src/components/AuthModal/components/RegisterForm/RegisterForm.spec.js
rogerhutchings: Possibly, although there isn't one in the base branch yet

srallen: That would probably help cut down on people submitting emails with typos in them.

eatyourgreens: We could spec it here then add it to the form.

srallen: I'll merge and spec it in the base branch before I add it in.

PR #898, opened 2019-06-03
Common linting package

rogerhutchings:
Creates a linting package based on standard that we can use across projects, with our current defaults.


eatyourgreens:
Is this still WIP? Looks pretty complete to me.

rogerhutchings:
It should be good to go. I couldn’t work out how to programmatically pipe the output through snazzy, but that’s a nice to have

PR #901, opened 2019-06-05
Fix Done And Talk when feedback isn't present

### LINKED ISSUE #900: 'Done and Talk' bug
**Package**

https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/talk/2108/1020172?comment=1682439&page=1 provides thread. In brief, 'Done and Talk' works for sims, but not for regular subjects. As Jim noted in the #tess slack thread, "Sounds like going to Talk fails unless there’s feedback, which points to a bug in the action code."

**To Dos**

- [ ] Example

eatyourgreens:
Add an openInTalk action to subjects and invoke it when Done & Talk is pressed. 
Remove the code from the DoneAndTalkButtonContainer that updated window.location to open a Talk page.
Instead, we now intercept the subject queue advance and look to see if we expect feedback or if Done & Talk has been pressed, then go to Talk if it's safe to do so.

Package:
lib-classifier

Closes #900.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
My thinking at the moment is that we open Talk from an action, `subject.openinTalk()`, then the feedback store can interrupt that action if necessary and invoke it again later. If there's no feedback, the action runs without interruption.

PR #889, opened 2019-05-29
[WIP] Registration form

srallen:
Package: app-project

WIP start of the registration form branched off of #888. No validations or API calls yet, just the UI bits and some styling work still needs to be done. 

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?
- [ ] Is the changelog updated?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?

PR #908, opened 2019-06-05
Registration Form

### LINKED ISSUE #869: Build registration form

srallen:
Package: app-project

Describe your changes:
An actually mergeable version of #889 that includes the updates from #893. I still need to address the last few comments from #893 so still WIP. #907 also should merge first.

Closes #869

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


srallen:
I was successfully able to register a new account via a manual test, but I want to note that I got a 422 on the UPP POST after it logged me in with the new account with this error:

```
"PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_user_project_preferences_on_project_id_and_user_id"↵DETAIL:  Key (project_id, user_id)=(1862, 1325970) already exists.↵: INSERT INTO "user_project_preferences" ("project_id", "user_id", "created_at", "updated_at") VALUES (1862, 1325970, '2019-06-06 21:19:17.036062', '2019-06-06 21:19:17.036062') RETURNING "id
```

srallen:
@eatyourgreens I've made an actual HOC for wrapping the form in Formik and separated the actual Form component into its own component that is getting decorated by the new `withFormik` HOC. This made it much easier to write tests for both the HOC and the inner form. Everything else about this PR should be like the original https://github.com/zooniverse/front-end-monorepo/pull/893

packages/app-project/src/components/AuthModal/components/RegisterForm/components/Form/Form.spec.js
```diff
@@ -0,0 +1,225 @@
...
+      wrapper.setProps({ errors })
+
+      expect(wrapper.find({ htmlFor: emailFieldId }).props().error).to.equal(errors.email)
+      expect(wrapper.find({ htmlFor: emailConfirmFieldId }).props().error).to.equal(errors.emailConfirm)
+      expect(wrapper.find({ htmlFor: passwordConfirmFieldId }).props().error).to.equal(errors.passwordConfirm)
+      expect(wrapper.find({ htmlFor: userNameFieldId }).props().error).to.equal(errors.username)
+    })
+  })
+
+  describe('when the under 16 checkbox is toggled', function () {
```
eatyourgreens: Nice to have tests for this, so that we can refer to them later.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.js
```diff
@@ -0,0 +1,64 @@
...
+            const {
+              handleBlur,
+              handleChange
+            } = innerProps
+
+            return (
+              <WrappedComponent
+                handleBlur={(event) => this.handleBlurWithCallback(event, handleBlur, innerProps)}
+                handleChange={(event) => this.handleChangeWithCallback(event, handleChange, innerProps)}
+                {...innerProps}
```
eatyourgreens: Will this overwrite `handleChange` and `handleBlur` with `innerProps.handleChange` and `innerProps.handleBlur`?

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    const formikPropsMock = {}
+    const wrapper = shallow(
+      <WrappedMockForm
+        initialValues={initialValues}
+        onChange={onChangeStub}
+        onSubmit={onSubmit}
+      />
+    )
+
+    wrapper.instance().handleChangeWithCallback(eventMock, handleChangeStub, formikPropsMock)
```
eatyourgreens: If you can call `WrappedComponent.props.handleChange` here, that would be a more robust test.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    const formikPropsMock = {}
+    const wrapper = shallow(
+      <WrappedMockForm
+        initialValues={initialValues}
+        onBlur={onBlurStub}
+        onSubmit={onSubmit}
+      />
+    )
+
+    wrapper.instance().handleBlurWithCallback(eventMock, handleBlurStub, formikPropsMock)
```
eatyourgreens: Ditto for `handleBlur` on the wrapped component.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.js
srallen: I'll write up a readme about what this is doing. This is adding support for adding custom callbacks to these two events if we want to do our own async validations on these events. Formik doesn't directly support this. I can see how it's confusing with the naming scheme going on.

eatyourgreens: One solution would be to remove `handleBlur` and `handleChange` from the spread props, so that's there no chance of accidental duplication.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.js
```diff
@@ -0,0 +1,64 @@
...
+      this.props.onChange(event, formikProps)
+    }
+
+    render () {
+      const { initialValues, onSubmit } = this.props
+      return (
+        <Formik
+          initialValues={initialValues}
+          onSubmit={onSubmit}
+          {...this.props}
```
eatyourgreens: I can never remember how the spread and rest operators work, but are you passing `initialValues` and `onSubmit` twice here, since they are included in `this.props`?

srallen: Checking the React dev tools I only see each expected prop once.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.spec.js
srallen: It would be in theory, but the `WrappedComponent` is the return value for Formik's render. Formik internals are hard to work with. Some async actions are happening which makes it hard to test without using a timeout, and I think an expectation inside a timeout in this case is potentially unreliable. Maybe this will change in the v2 release? If you have an idea to make this less brittle, I'm open to suggestion.

srallen:
I've added a readme for the `withFormik` HOC and clarified the prop names.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    handleBlurWithCallback (event, handleBlur, formikProps) {
+      handleBlur(event)
+      this.props.onBlur(event, formikProps)
+    }
+
+    handleChangeWithCallback (event, handleChange, formikProps) {
+      handleChange(event)
+      this.props.onChange(event, formikProps)
```
eatyourgreens: Should `formikProps.handleChange()` be called from this function too?

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.js
```diff
@@ -0,0 +1,68 @@
...
+  class FormikHOC extends React.Component {
+    constructor () {
+      super()
+
+      this.handleBlurWithCallback = this.handleBlurWithCallback.bind(this)
+      this.handleChangeWithCallback = this.handleChangeWithCallback.bind(this)
+    }
+
+    handleBlurWithCallback (event, handleBlur, formikProps) {
+      handleBlur(event)
```
eatyourgreens: Should `formikProps.handleBlur()` be called too?

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.js
eatyourgreens: Sorry, the readme has left me more confused. The function now passes new event handlers down in addition to the original handlers passed down in `...innerProps`. So `WrappedComponent` has a `handleChange` prop and an `onChangeEvent` prop and so on for blur, reset and submit. Is one handler supposed to handle async validation and the other synchronous validation. I'm finding this code very hard to understand.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.spec.js
eatyourgreens: Sorry, I missed this on my first pass, but the mock Formik props here should not be empty but should mock the expected inner props for these values of `initialValues`, `onSubmit` and `onChange`.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    const formikComponent = wrapper.find(Formik)
+
+    expect(formikComponent.props().validate).to.equal(validate)
+  })
+
+  it('should call onChange prop', function () {
+    const onChangeStub = sinon.stub()
+    const handleChangeStub = sinon.stub()
+    const eventMock = {}
+    const formikPropsMock = {}
```
eatyourgreens: I think this should be
```js
const formikPropsMock = wrapper.dive().find('MockForm').props()
```
so that you're testing the props that Formik generates for the wrapped form.

eatyourgreens:
Another idea for the HOC tests: add support for ref forwarding and use a ref to inspect the wrapped instance in tests? MobX's `inject` used to do this with the `.wrappedInstance` property but uses forward refs now.
https://github.com/mobxjs/mobx-react/blob/4904b624f46be2e14d3a3d1017c1fa9bcb334f02/src/inject.js#L20-L22

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.js
srallen: `handleChange` is the original Formik handler that is required to be used with the inputs on change event. Their internal API is handling the form state and updating the input during the on change event. I've added support for the `onChange` prop since Formik doesn't support this, so we can support things like async validation on the on change event.

Similarly I did this with on blur too because Formik itself doesn't support customization of the event handling. They do support customization of the submit and reset event, but I've explicitly set them to `onResetEvent` and `onSubmitEvent` to make the prop names consistent for the consuming form. 

I'll try another pass at the naming of things and the readme since it's still not clear.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Formik } from 'formik'
+
+function withFormik (WrappedComponent) {
```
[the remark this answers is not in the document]
srallen: Sure, I'll think about it.

srallen:
> Another idea for the HOC tests: add support for ref forwarding and use a ref to inspect the wrapped instance in tests? MobX's inject used to do this with the .wrappedInstance property but uses forward refs now.

We could do that when MobX and related libs is updated?

> I'm very confused by the withFormik() function, and not sure how it differs from the withFormik function that comes with Formik.

The difference is the added support I wrote for `props.onChange` and `props.onBlur` since Formik doesn't support customizing or having callbacks for those events. I'll rename the function to make it clear it had a difference.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.js
eatyourgreens: OK. What's confusing me most at the moment is that in the React dev tools I can see `handleChange` in the wrapped Form's props (from `{...innerProps}`) but the code for `Form.js` doesn't use that prop, unless I've missed it.

I _think_ it's fine to name it `handleChange`, as long as you don't also use `{...innerProps}` to pass `handleChange` down to the wrapped form.

eatyourgreens:
>The difference is the added support I wrote for props.onChange and props.onBlur since Formik doesn't support customizing or having callbacks for those events. I'll rename the function to make it clear it had a difference.

The example given for [`withFormik`](https://jaredpalmer.com/formik/docs/api/withformik) uses `onChange` on input fields but, yeah, it doesn't pass a `handleChange` prop to the wrapped form according to the props listed there.

eatyourgreens:
>We could do that when MobX and related libs is updated?

We already support forward refs, and use them in the subject viewer.
https://github.com/zooniverse/front-end-monorepo/blob/73d300b6441a632f888f20d60e5b7f66138a3a4f/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.js#L12-L25

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.js
srallen: Yeah, I need to fix how I'm destructuring and passing the props.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.spec.js
eatyourgreens: This afternoon, I experimented with using a named function to get the wrapped component in tests but couldn't get that to work. Refs might be the way to go if you want to test the props that are passed down to the wrapped form.

packages/app-project/src/components/AuthModal/shared/components/withFormik/withFormik.js
srallen: That's what the `handleChange` is calling in the above line. I wanted to make the rest of the props available to the props handler, like setting the error message of necessary. I see I can tidy this up though.

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.js
```diff
@@ -0,0 +1,69 @@
...
+              handleSubmit,
+              ...restOfInnerProps
+            } = innerProps
+
+            return (
+              <WrappedComponent
+                handleBlur={(event) => this.handleBlurWithCallback(event, innerProps)}
+                handleChange={(event) => this.handleChangeWithCallback(event, innerProps)}
+                {...restOfInnerProps}
+              />
```
eatyourgreens: `handleSubmit` and `handleReset` are missing here.

srallen: They're being passed down with the `restOfInnerProps`. Formik supports `onSubmit` and `onReset` custom handlers, so I don't really have to do the hack like I have for change and blur. I had them explicitly defined for the props passed down here in the previous commit because I wanted the prop namings to be consistent, however, since I reverted back to `handleChange`, `handleBlur`, etc, I don't really need to do that again.

eatyourgreens: You’ve declared them separately on the previous line. This is where it would be useful to have tests for the props that are passed down to the wrapped form.

srallen: Yep I did. I'll clean this up and look into adding forwarded refs

eatyourgreens: If you want to get this out of the way quickly, I think `wrapper.dive().find('MockForm').props()` should also work for getting a reference to the form and testing that the handlers have been added properly.

Then ref forwarding could be another PR. Depends how much more work it adds.

srallen: I've added the forwarded refs and refactored the tests for the onChange and onBlur handlers. The onChange handler is working, however, I could not get the onBlur one to work at all. I even manually tested to see if the original formik `handleBlur` was working and it does not appear to be firing at all. I have no idea why and I'd like to investigate this separately with the LoginForm in a follow up because I know the blur event handlers are working there with the direct use of the Formik API. For now, I removed the customized onBlur handling.

eatyourgreens: Changes look good to me. I’ll pull it down tomorrow and test it locally. The only blocking issue today was that submission had broken, and it looks like that’s fixed now.

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.js
```diff
@@ -0,0 +1,66 @@
...
+      this.props.onChange(event, formikProps)
+    }
+
+    render () {
+      const { forwardedRef, initialValues, onSubmit, ...rest } = this.props
+      return (
+        <Formik
+          initialValues={initialValues}
+          onSubmit={onSubmit}
+          {...rest}
```
eatyourgreens:
```suggestion
          {...optionalProps}
```
Maybe clarify the naming here.

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.js
```diff
@@ -0,0 +1,66 @@
...
+      this.handleChangeWithCallback = this.handleChangeWithCallback.bind(this)
+    }
+
+    handleChangeWithCallback (event, formikProps) {
+      formikProps.handleChange(event)
+      this.props.onChange(event, formikProps)
+    }
+
+    render () {
+      const { forwardedRef, initialValues, onSubmit, ...rest } = this.props
```
eatyourgreens:
```suggestion
      const { forwardedRef, initialValues, onSubmit, ...optionalProps } = this.props
```

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.js
```diff
@@ -0,0 +1,66 @@
...
+      return (
+        <Formik
+          initialValues={initialValues}
+          onSubmit={onSubmit}
+          {...rest}
+        >
+          {(innerProps) => {
+            const {
+              handleChange,
+              ...restOfInnerProps
```
eatyourgreens: `innerProps.handleBlur` will be passed down here, so filter it out here if you really do want to remove support for it.

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
```diff
@@ -0,0 +1,91 @@
...
+    const onChangeStub = sinon.stub()
+
+    const wrapper = mount(
+      <WrappedMockForm
+        initialValues={initialValues}
+        onChange={onChangeStub}
+        onSubmit={onSubmit}
+      />
+    )
+    const formikProps = wrapper.find(MockForm).props()
```
eatyourgreens: Out of interest, why did we have to use `wrapper.dive().find('MockForm')` to get the form before, but `wrapper.find(MockForm)` works here?

eatyourgreens: Oh wait, we were using shallow mounting before and now use a full render?

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
```diff
@@ -0,0 +1,91 @@
...
+    const eventMock = {
+      target: {
+        name: testInput.props().name,
+        value: 'foo'
+      }
+    }
+
+    expect(testInput.props().value).to.equal('')
+    testInput.simulate('change', eventMock)
+    // diry chai does not work with sinon chai
```
eatyourgreens: dirty-chai works with sinon-chai in other tests, so maybe be specific here about what doesn't work eg. `dirty-chai does not work with sinon's spy.withArgs().`

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
```diff
@@ -0,0 +1,91 @@
...
+    expect(onChangeStub.withArgs(eventMock, formikProps)).to.have.been.calledOnce
```
eatyourgreens: I _think_ this is passing because `.calledOnce` is a function, so you're checking that function name exists here.

eatyourgreens:
For the tests, I would add expectations to check that `handleChange` and `handleSubmit` are passed to the wrapped form, since one commit broke this and it wasn't picked up.

Maybe also call `mockForm.props.handleChange()` just to check that it calls your custom handler as expected. You probably don't need to test that Formik handles changes, since Formik should have tests for that.

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
eatyourgreens: Testing locally, `eventsMock` doesn't match the synthetic event that is actually passed to the event handler.

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
srallen: Yep

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.js
srallen: I'm not sure if I want to do that because I don't understand why it's broken yet.

srallen:
Tests were updated via #918

PR #913, opened 2019-06-10
Add drawing task inputs

mcbouslog:
Package: lib-classifier

First of a few PRs related to adding drawing tasks to lib-classifier.

Describe your changes:
- add drawing task tool inputs
- add drawing tool icons (wrapped in grommet-icons `Blank`)
- refactor `DrawingStore` for active drawing tool
- add `counterpart` to `InputStatus`

Questions
- see `Tasks.js`, changed tasks container box `basis` from `246px` to `auto`, unsure if acceptable across all use cases
- unsure if `icons.js` optimal way to organize drawing tool icons, `InputIcon` slightly different in lib-classifier (Icon included as prop) compared to PFE (icon set in [DrawingToolInputIcon](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/drawing/components/DrawingToolInputIcon/DrawingToolInputIcon.jsx))

Notes
- staging project with point, line, rectangle, another line: http://localhost:8080/?project=1876

# Review Checklist

## General

## Apps

- [x] Does it work on mobile?


packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DrawingTask/icons.js
```diff
@@ -0,0 +1,187 @@
+import React from 'react'
+import { Blank } from 'grommet-icons'
+
+const BlankWithViewBox = ({ children }) => (
+  <Blank viewBox='0 0 100 100'>
+    {children}
+  </Blank>
+)
+
+const icons = {
```
mcbouslog: maybe something like [getViewer](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/helpers/getViewer/getViewer.js) would be preferable? Therefore split icons into individual files.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
```diff
@@ -54,7 +54,7 @@ export class Tasks extends React.Component {
               const TaskComponent = getTaskComponent(task.type)
               if (TaskComponent) {
                 return (
-                  <Box key={task.taskKey} basis='246px'>
+                  <Box key={task.taskKey} basis='auto'>
```
srallen: I think the fixed size was to make it so that the box div had a min height. But it looks like this is causing issues, so we'll need to do this another way. Switching to auto is best for now.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DrawingTask/icons.js
srallen: I think so. At some point I think we're going to move the task types into their own package, so the consuming app only loads up what it needs.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/InputStatus/InputStatus.js
```diff
@@ -1,29 +1,36 @@
...
   text-align: right;
 `
 
-// TODO Add Translations
 export default function InputStatus ({ count, tool }) {
   const minStyleColor = (count < tool.min) ? 'red' : ''
   const maxStyleColor = (count === tool.max) ? 'orange' : ''
+  const drawn = counterpart('InputStatus.drawn')
+  const maximum = counterpart('InputStatus.maximum')
+  const of = counterpart('InputStatus.of')
```
srallen: I think `of` is a reserved word in javascript (or was it coffeescript)? Perhaps rename this to `ofTranslation`?

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/InputStatus/InputStatus.js
```diff
@@ -1,29 +1,36 @@
...
+  const required = counterpart('InputStatus.required')
   return (
     <StyledInputStatus>
       {count}{' '}
       {(!!tool.min || !!tool.max) &&
-        'of '}
+        `${of} `}
       {!!tool.min &&
-        <span style={{ color: minStyleColor }}>{tool.min} required</span>}
+        <span style={{ color: minStyleColor }}>{`${tool.min} ${required}`}</span>}
```
srallen: Let's use Grommet's `Text` component here, which renders as a span, then we can style it using the color prop. Also, we should check with Becky on these colors. The red is hard to read to me when the input is selected and has the teal background.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskInputField/components/TaskInputLabel/TaskInputLabel.js
```diff
@@ -30,7 +32,6 @@ export const StyledTaskInputLabelWrapper = styled(Box)`
   }
 
   img:only-child, svg:only-child {
-    background-color: ${zooTheme.global.colors.brand};
```
srallen: This shouldn't be removed. This is the background color for projects that are including a single image icon in their markdown, like galaxy zoo.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskInputField/components/TaskInputLabel/TaskInputLabel.js
```diff
@@ -7,6 +7,8 @@ import zooTheme from '@zooniverse/grommet-theme'
 import { Markdownz, pxToRem } from '@zooniverse/react-components'
 
 export const StyledTaskInputLabelWrapper = styled(Box)`
+  margin-right: 10px;
```
srallen: I think this can be set directly on the Box prop using `margin`. https://v2.grommet.io/box#margin

mcbouslog:
@beckyrother - finishing up the drawing tool inputs and could use your input on the following at minimum, but of course any notes welcome:
- tool min and max text color
- icon background color (inherit from state, i.e. hover/selected/unselected?)
Here are some screenshots, light theme:
Screen Shot 2019-06-24 at 5 25 12 PM
Screen Shot 2019-06-25 at 8 54 56 AM

srallen:
@mcbouslog just to clarify: the teal background is intended for projects that add a svg or transparent png image to their markdown for the input label, like Galaxy Zoo. I don't think we can use it for the drawing tools icons since they vary in color.

beckyrother:
Agreed with @srallen, the teal is going to make those icons very difficult to read. How does it look without a background?

srallen:
I think we can use a bit of CSS to make sure the shared styles of underlying input component only applies the background to the single and multiple choices task types.

mcbouslog:
Oh yes! Sorry didn't mean to imply would finish with tool icon teal background like that. I think this looks better:
dark theme:
Screen Shot 2019-06-25 at 10 51 00 AM
light theme:
Screen Shot 2019-06-25 at 10 51 15 AM

beckyrother:
Much better, thank you!

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DrawingTask/components/FreehandLine.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+const FreehandLine = () => {
+  return (
+    <path d='M10,50 Q25,10,50,50,75,90,90,50' fill={'none'} />
```
srallen: Not a blocker, but `fill` can just be `fill='none'`

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/InputStatus/InputStatus.js
```diff
@@ -1,29 +1,37 @@
...
 `
 
-// TODO Add Translations
 export default function InputStatus ({ count, tool }) {
   const minStyleColor = (count < tool.min) ? 'red' : ''
   const maxStyleColor = (count === tool.max) ? 'orange' : ''
+  const drawn = counterpart('InputStatus.drawn')
+  const maximum = counterpart('InputStatus.maximum')
+  const ofTranslation = counterpart('InputStatus.of')
+  const required = counterpart('InputStatus.required')
```
eatyourgreens: This assumes the sentence structure will be `x of y required` in another language. I'm not sure if that's true.

srallen: Perhaps the whole sentence should be processed by counterpart with variable placeholders for the counts?

eatyourgreens: That would be my preferred option. I'm trying to remember why we didn't do that in PFE. I'm assuming because of the styling of the coloured text, and the conditional logic that constructs the sentence in English.

eatyourgreens: Oh, and the logic here is complicated enough to consider moving this message into its own, testable, functional component.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
srallen: This change is causing a style issue:

On this branch:
Screen Shot 2019-07-01 at 10 54 36 AM

On production:
Screen Shot 2019-07-01 at 10 54 47 AM

So the top spacing between the buttons below the classifier are depending on a min-height of the task area. The buttons should be managing their own margins.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/InputStatus/InputStatus.js
srallen: The color styling looks bad with the active teal color state as is, so let's maybe drop the colors and better support translations instead.

eatyourgreens: Sounds good to me. Markdown strings might work better here anyway.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskInputField/components/TaskInputLabel/TaskInputLabel.js
mcbouslog: Is it ok if I move the `background-color` styling specifically to the single and multiple tasks in [Commit 4f47](https://github.com/zooniverse/front-end-monorepo/pull/913/commits/4f47250143588bd65881e521d83b2afa13cdc9c3)?

staging project/workflow `markb-panoptes/fem-drawing-tools/classify?workflow=3344` has question task with icons in label.

srallen: Yes, that works for me.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/InputStatus/InputStatus.js
mcbouslog: Not sure what I'm doing wrong with the following `en` json translation object:
```
{
  "InputStatus": {
    "drawn": "%(count) drawn"
  }
}
```
Then attempting to set the status as follows (though won't be hardcoded to 10, simplifying troubleshooting):
```
const status = counterpart('InputStatus.drawn', { count: '10' })
```
I get console error of:
```
Uncaught SyntaxError: [sprintf] unexpected placeholder
```

srallen: This trips me up every time. Counterpart is expecting a syntax of `%(variable)s` note the `s` at the end.

eatyourgreens: sprintf requires an `s` to specify a string, I think: `%(string)s`.

mcbouslog: ah ha thank you both!

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DrawingTask/DrawingTask.js
```diff
@@ -1,56 +1,101 @@
...
-    // if (annotations && annotations.size > 0) {
-    //   annotation = annotations.get(task.taskKey)
-    // }
     return (
-      <Text size='small' tag='legend'>
-        <Markdownz>
-          {task.instruction}
-        </Markdownz>
-      </Text>
+      <StyledFieldset>
```
mcbouslog: Discussion started in #1020:
@eatyourgreens:
I don't think Fieldset is the right element to use here. Fieldsets have simple, inline text legends but our task questions are block-level elements, sometimes with more than one paragraph.

I think this should just be a Grommet Box.

eatyourgreens:
Fieldset/Legend can't take block-level children, so can't wrap a Markdownz component.

eatyourgreens:
Here's the list of what's valid in a fieldset legend. I don't think we can easily switch to using these elements for task questions without breaking existing live workflows.
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content

@srallen:
This was my bad when first starting on the task area and I think it's being copied around. I was thinking of using valid form elements if in theory we could have a PWA that could work without javascript, but since there's markdown going here, we won't be able to guarantee the question is valid HTML for these elements.

eatyourgreens:
We could possibly mark up transcription tasks as forms that submit the values of the text inputs, or question tasks as forms that submit the answers. Drawing tasks would submit the value of the selected drawing tool, so I'm thinking either we move the form markup down to specific types of task, or drop it completely.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
mcbouslog: I think this has been fixed with the changes to TaskArea and adding `fill` or `flex` on related components.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DrawingTask/DrawingTask.js
mcbouslog: Addressed in #1039

beckyrother:
Would it be better to keep them the way they were at the beginning with a dark background (#2D2D2D instead of teal though) regardless of dark or light theme? 
These colors were chosen so they'd show up on the biggest range of backgrounds, which isn't going to solve for everything, but good idea to try to at least catch what we can control.

simensta:
Using a darker background color like #2D2D2D for the icon sounds good to me. For the green tool, it increases the contrast score to [10.03:1](https://webaim.org/resources/contrastchecker/?fcolor=00FF00&bcolor=2D2D2D) and will likely improve the contrast score for the other tool colors.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
mcbouslog: @srallen I might be mistaken but I think your "requested changes" status is related to this styling issue? if so, is this looking ok now? @simensta and I are looking to get this PR cleaned up and ready for merge early this week.

srallen: Screen Shot 2019-07-29 at 1 12 35 PM

This still is showing the metadata buttons right up against the subject viewer. As long as we note this and do a follow up PR addressing this, I can dismiss my review and allow @simensta to finish review for merging.

simensta:
Hey @beckyrother, @mcbouslog and I tried adding #2D2D2D as a background to the drawing tool icons. We thought it looked a bit clunky and in need of some further design input from you. 

icons_with_dark_background_2019-07-29

mcbouslog:
I've opened #1052 to capture any drawing task related design thoughts until more formal documentation or discussion. Optionally we can shift the icon background discussion there if we want to merge this and revisit.

simensta:
Good idea - let's shift the icon background discussion to #1052. @mcbouslog, can you rebase this branch?  Then I think this branch is ready to merge.

PR #917, opened 2019-06-12
Use withArgs for spies

eatyourgreens:
Use withArgs to create specific spies in tests, rather than using calledOnceWith.

Package:
app-project

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/src/screens/ClassifyPage/components/CollectionsModal/CollectionsModalContainer.spec.js
+++ b/packages/app-project/src/screens/ClassifyPage/components/CollectionsModal/CollectionsModalContainer.spec.js
@@ -63,7 +63,7 @@ describe('Component > CollectionsModalContainer', function () {
       })
 
       it('should fetch a list of collections filtered by name', function () {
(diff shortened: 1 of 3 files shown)


eatyourgreens:
So I expected this to break the tests, as per the same change in #908, but it doesn't. 🤔

srallen:
There's some ambiguity with the meaning of `calledOnceWith` and it seems that using `withArgs` is the more clear option of what we're intending to test? https://github.com/sinonjs/sinon/issues/1200

eatyourgreens:
I hadn’t thought of that. 

Today I found that the pattern here will fail with something like `expected spy to be called once but it was called 0 times` whereas `calledOnceWith` will list the expected args and actual args when it fails. 

That can be much more helpful when debugging failing tests, so I’m happy to close this without merging if that’s preferred.

eatyourgreens:
On ambiguity, if I call a function with`(1, 2)` then `calledOnceWith(1)` will pass but `withArgs(1)` might fail the test because it looks for an exact match.

PR #918, opened 2019-06-12
Fix withCustomFormik test failures

eatyourgreens:
Package:
app-project

Try to figure out why registration form tests are failing.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
+++ b/packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
@@ -70,18 +70,22 @@ describe.only('Higher Order Component > withCustomFormik', function () {
         onSubmit={onSubmit}
       />
     )
-    const mockForm = wrapper.find(MockForm)
-    const formikProps = mockForm.props()
     const eventMock = {
       target: {
         name: 'testInput',
         value: 'foo'
       }
     }
-    const changeSpy = onChangeStub.callsFake((event, props) => console.log(event === eventMock, props === formikProps))
+    const renderForm = wrapper.find('Formik').props().children
+    const mockInnerProps = {
+      values: {
+        testInput: 'foo'
+      },
+      handleChange: sinon.stub()
+    }
 
-    formikProps.handleChange(eventMock, formikProps)
-    console.log(changeSpy.callCount)
-    expect(changeSpy).to.have.been.calledOnceWith(eventMock, formikProps)
+    const mockForm = renderForm(mockInnerProps)
+    mockForm.props.handleChange(eventMock)
+    expect(onChangeStub.withArgs(eventMock, mockInnerProps)).to.have.been.calledOnce()
   })
 })
--- a/packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
+++ b/packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
@@ -26,7 +26,7 @@ const WrappedMockForm = withCustomFormik(MockForm)
 const initialValues = { testInput: '' }
 const onSubmit = sinon.spy()
 
-describe.only('Higher Order Component > withCustomFormik', function () {
+describe('Higher Order Component > withCustomFormik', function () {
   it('renders without crashing', function () {
     const wrapper = shallow(<WrappedMockForm initialValues={initialValues} onSubmit={onSubmit} />)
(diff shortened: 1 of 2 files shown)


eatyourgreens:
This isn't necessary for #908. It's splitting out the HOC tests to see if we can figure out why they're broken.

eatyourgreens:
Based on the code here, it seems the tests fail because the `props` argument to the change handler doesn't match?

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
```diff
@@ -60,32 +60,29 @@ describe('Higher Order Component > withCustomFormik', function () {
...
 
-    expect(testInput.props().value).to.equal('')
-    testInput.simulate('change', eventMock)
-    // diry chai does not work with sinon chai
-    expect(onChangeStub.withArgs(eventMock, formikProps)).to.have.been.calledOnce
-    testInput = wrapper.find('input[name="testInput"]')
-    // Formik's handleChange updates the input value for us based on the name attribute
-    // Test that the internal Formik handleChange has been called and worked
-    expect(testInput.props().value).to.equal('foo')
+    formikProps.handleChange(eventMock, formikProps)
```
eatyourgreens: Will this set up some weird recursive behaviour because the custom handler function calls `formikProps.handleChange`?

eatyourgreens: Maybe this is why the test fails, because that function has been modified so doesn't match?

eatyourgreens: Yes, this test is failing because we should be testing that the handler is called with the forms original props (`innerProps` in the code), not the modified props.

eatyourgreens:
This is ready to review now. My solution was to test the custom render function that is passed to Formik as `props.children`.

eatyourgreens:
I'll also note that I tried adding a ref to the wrapped form, which works in #914, but it did not work here so I abandoned that approach.

packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
```diff
@@ -60,32 +60,33 @@ describe('Higher Order Component > withCustomFormik', function () {
...
 
-    expect(testInput.props().value).to.equal('')
-    testInput.simulate('change', eventMock)
-    // diry chai does not work with sinon chai
-    expect(onChangeStub.withArgs(eventMock, formikProps)).to.have.been.calledOnce
-    testInput = wrapper.find('input[name="testInput"]')
-    // Formik's handleChange updates the input value for us based on the name attribute
-    // Test that the internal Formik handleChange has been called and worked
-    expect(testInput.props().value).to.equal('foo')
+    const mockForm = renderForm(mockInnerProps)
```
eatyourgreens: Should I write this in JSX to make it clear what’s happening here?

srallen: Yeah I think that'd help. I'm having a hard time understanding how we're testing this all.. maybe I shouldn't bother with the customization and just use the vanilla `withFormik` they provide?

eatyourgreens: I’ll rewrite this in JSX tomorrow. It’s a quick change but I had a hard time figuring this out until I realised that Formik expects a component declaration, not a component instance, as a child. 

I’ve been expecting it to wrap a component instance, because that’s more intuitive to me in JSX. Passing a function as `props.chlidren` made it harder to reason about.

If `renderForm` is renamed to `InnerForm` then this line becomes `mockForm = <InnerForm {...mockInnerProps} />`.

eatyourgreens:
I’m inclined to write an RFC expanding on my thoughts here and asking if we want to continue with Formik.  I haven’t looked at `withFormik()` in detail, but that might be easier to work with.

srallen:
@eatyourgreens That seems reasonable. I just had a look at their github and they're preparing to release version 2. The readme has changed to be about the pre-release. Perhaps there will be improvements? I know that they're dropping async validation support.

eatyourgreens:
I've added more tests, and refactored the code to try and make it easier to read. I've added a couple of comments too.

PR #922, opened 2019-06-13
Add back support for customization of onBlur

srallen:
Package: app-project

Describe your changes:
As a follow up, I checked to see if onBlur is on working with the `withCustomFormik` HOC with the refactor from #918 and it does now! Tests are passing for it, so I now I have confidence it can be used with the login form.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
+++ b/packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
@@ -61,10 +61,12 @@ describe('Higher Order Component > withCustomFormik', function () {
   })
 
   describe('the wrapped form', function () {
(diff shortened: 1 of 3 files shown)


packages/app-project/src/components/AuthModal/shared/components/withCustomFormik/withCustomFormik.spec.js
```diff
@@ -91,7 +94,7 @@ describe('Higher Order Component > withCustomFormik', function () {
     
     Object.keys(mockInnerProps).forEach(function (key) {
       it(`should pass ${key} to the wrapped form`, function () {
-        expect(mockForm.prop(key)).to.exist
+        expect(mockForm.prop(key)).to.exist()
```
eatyourgreens: good catch!

PR #924, opened 2019-06-14
[WIP] Update mobx

srallen:
Package: lib-classifier

Describe your changes:
This updates mobx to 5.9.0 and mobx-state-tree to 3.11.0. These are not the latest versions, but the versions when our stores break. The updates I've had to make are:

- Updates `types.reference` to `types.safeReference`. `types.reference` cannot be undefined. `types.safeReference` can be and will auto clean itself up, so we don't have to manually set it to undefined on clean up or reset actions for the stores.
- Updates how we instantiate the child node stores on the `RootStore`. We shouldn't be doing this: `types.optional(ClassificationStore, ClassificationStore.create({})`. MST will throw an error that you can't use an instance in the optional default value, but must set it to either a snapshot or a function that creates a snapshot or instance. So they've been changed to use an empty snapshot like: `types.optional(ClassificationStore, {})`. In the future, we might update this to use an actual snapshot dumped into the user's IndexedDB or something like that to pick up state where they last left it.
- I've added names to all of the autorun functions to make debugging easier. 

This is still WIP because some `SubjectStore` tests are still failing. If I can resolve this, then I'll do the last mobx related dependency update for mobx-react and see if that goes smoothly. If so, then we'll have a path to update mobx in the project app. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/store/FieldGuideStore.spec.js
```diff
@@ -54,8 +53,16 @@ describe('Model > FieldGuideStore', function () {
...
+      dataVisAnnotating: {},
+      drawing: {},
+      feedback: {},
+      subjects: {},
+      subjectViewer: {},
+      tutorials: {},
+      workflows: {},
+      workflowSteps: {},
+      userProjectPreferences: {}
     }, { client: { panoptes: { get: sinon.stub().callsFake(() => Promise.resolve(null)) } } })
```
eatyourgreens: These lines get repeated throughout these tests. Could the root store be returned by a function, `rootStore = getRootStore(options)`, where options is an object that can be merged into the default snapshot before it's passed to `create()`? Then you'd only be declaring the default values once.

srallen:
@eatyourgreens and @rogerhutchings could I get a mini-review specifically just on the `WorkflowStepStore` tests refactor (the most recent commit)? I purposely left in an `only` to scope to these tests. I had to refactor them because we cannot mock stores with setting an `active` attribute to be a plain resource object mock anymore. This is because I've had to add in reference validations using the new `isValidReference` API method and tests will fail if `active` is not a `safeReference` type. If the changes I've made to the `WorkflowStepStore` looks good, then I'll update the `ClassificationStore` tests accordingly (and elsewhere if needed).

srallen:
@eatyourgreens or @rogerhutchings whenever you can it'd be helpful to get that mini-review for the test restructuring and rewrite of the `WorkflowStepStore` so I can proceed when I can with this. I probably won't get to it for at least another week though (US holiday next week).

packages/lib-classifier/src/store/WorkflowStepStore.spec.js
```diff
@@ -108,10 +118,12 @@ describe('Model > WorkflowStepStore', function () {
...
-        workflows: {
-          active: WORKFLOW
-        }
+        projects: {},
+        subjects: {},
+        userProjectPreferences: {}
       })
+      ROOT_STORE_INSTANCE.workflows.setResource(WORKFLOW)
+      ROOT_STORE_INSTANCE.workflows.setActive(WORKFLOW.id)
     })
```
eatyourgreens: This is duplicated from line 47 so could maybe be moved into a function and declared once. `ROOT_STORE_INSTANCE = getRootStore()`.

packages/lib-classifier/src/store/WorkflowStepStore.spec.js
```diff
@@ -121,7 +133,8 @@ describe('Model > WorkflowStepStore', function () {
 
     it('should convert the tasks to steps and set the steps', function () {
       const { workflowSteps } = ROOT_STORE_INSTANCE
-      expect(ROOT_STORE_INSTANCE.workflows.active.steps).to.have.lengthOf(WORKFLOW.steps.length)
+      const stepsSize = workflowSteps.steps.size
```
eatyourgreens: Having both `workflows.active.steps` and `workflowSteps.steps` in the models is confusing.

packages/lib-classifier/src/store/WorkflowStepStore.spec.js
```diff
@@ -185,71 +187,50 @@ describe.only('Model > WorkflowStepStore', function () {
...
     })
 
     after(function () {
       ROOT_STORE_INSTANCE = null
+      subject = null
       WORKFLOW = null
     })
 
-    it('should return false if there is not an active workflow', function () {
+    function setupDoneAndTalkTest ({ workflow, createClassification }) {
```
eatyourgreens: Maybe rename `createClassification` to `withClassification`. I thought it was a reference to the function of the same name.

packages/lib-classifier/src/store/WorkflowStepStore.spec.js
srallen: Yes, I agree, but I'd like to try to keep this scoped only to changes needed to get MobX updates working.

eatyourgreens: No problem. I've been trying to figure out why this line changed and I'm still not 100% sure what's going on in this test, or what it was testing before it changed.

srallen: Yes it was a mistake, likely caused by the similar naming in the two stores as you pointed out. I don't recall if something else was failing with it with the MobX update, but I likely took a closer look at it because it was failing and then realized it was testing the wrong thing to begin with.

srallen:
I am closing this and will plan on redoing this in smaller PRs after the yarn switchover. I'm going to leave the branch, though, for reference as I redo it.

PR #928, opened 2019-06-17
Fix landmarks

### LINKED ISSUE #671: Page landmarks are inconsistent and confusing
**Package**
app-classifier

Screenshot showing landmarks for the classify page

Page landmarks are kind of confusing in the new app. The classify page shown here has a few things that could be improved:
- multiple navigation landmarks could be labelled with `aria-label` to distinguish them. (I think `aria-label` works on `nav` elements.)
- the footer probably doesn't need 8 navigation landmarks.
- there's no `main` landmark for the page content but there is a `complementary`, which marks the subject toolbar as a sidebar for the page content.
- the project header isn't a banner.
- the second banner, from the stats box, maybe shouldn't be there at all.

rogerhutchings:
Closes #671

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/src/components/ProjectHeader/ProjectHeader.js
+++ b/packages/app-project/src/components/ProjectHeader/ProjectHeader.js
@@ -23,7 +23,7 @@ function ProjectHeader (props) {
   const { className, navLinks, projectHomeLink, size, title } = props
 
   return (
-    <StyledBox className={className}>
+    <StyledBox as='header' className={className} aria-label={counterpart('ProjectHeader.ariaLabel')}>
       <Background />
       <Box
         align='center'
--- a/packages/app-project/src/components/ProjectHeader/ProjectHeader.js
+++ b/packages/app-project/src/components/ProjectHeader/ProjectHeader.js
@@ -23,7 +23,7 @@ function ProjectHeader (props) {
   const { className, navLinks, projectHomeLink, size, title } = props
 
   return (
-    <StyledBox as='header' className={className} role='presentation'>
+    <StyledBox as='header' className={className}>
       <Background />
       <Box
         align='center'
--- a/packages/app-project/src/components/ProjectHeader/ProjectHeader.js
+++ b/packages/app-project/src/components/ProjectHeader/ProjectHeader.js
@@ -23,7 +23,7 @@ function ProjectHeader (props) {
   const { className, navLinks, projectHomeLink, size, title } = props
 
   return (
-    <StyledBox as='header' className={className} aria-label={counterpart('ProjectHeader.ariaLabel')}>
+    <StyledBox as='header' className={className} role='presentation'>
       <Background />
       <Box
         align='center'
--- a/packages/lib-react-components/src/ZooHeader/ZooHeader.js
+++ b/packages/lib-react-components/src/ZooHeader/ZooHeader.js
@@ -63,6 +63,7 @@ export default function ZooHeader (props) {
       justify='between'
       pad='none'
       responsive={false}
+      role='presentation'
       tag='header'
       {...props}
     >
--- a/packages/lib-react-components/src/ZooHeader/ZooHeader.js
+++ b/packages/lib-react-components/src/ZooHeader/ZooHeader.js
@@ -69,7 +69,7 @@ export default function ZooHeader (props) {
     >
       <Box
         align='center'
-        aria-label={counterpart('ZooHeader.navAriaLabel')}
+        aria-label={counterpart('ZooHeader.ariaLabel')}
         direction='row'
         pad={{ horizontal: 'medium' }}
         responsive={false}
--- a/packages/lib-react-components/src/ZooHeader/ZooHeader.js
+++ b/packages/lib-react-components/src/ZooHeader/ZooHeader.js
@@ -55,6 +55,7 @@ export default function ZooHeader (props) {
 
   return (
     <StyledHeader
+      aria-label={counterpart('ZooHeader.ariaLabel')}
       background='black'
       direction='row'
       fill='horizontal'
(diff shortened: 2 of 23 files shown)


packages/app-project/src/components/ProjectHeader/ProjectHeader.js
```diff
@@ -23,7 +23,7 @@ function ProjectHeader (props) {
   const { className, navLinks, projectHomeLink, size, title } = props
 
   return (
-    <StyledBox className={className}>
+    <StyledBox as='header' className={className} aria-label={counterpart('ProjectHeader.ariaLabel')}>
```
eatyourgreens: Why is this labelled? I thought the project had a HTML heading already?

packages/lib-react-components/src/ZooHeader/ZooHeader.js
```diff
@@ -55,6 +55,7 @@ export default function ZooHeader (props) {
 
   return (
     <StyledHeader
+      aria-label={counterpart('ZooHeader.ariaLabel')}
```
eatyourgreens: I'll double-check, but we probably don't need to label these headers as long as the page heading structure is clearly marked up.

packages/app-project/src/components/ProjectHeader/ProjectHeader.js
rogerhutchings: There's the site header, and the project header, so it's there to differentiate them

eatyourgreens: Yes, but they have headings already, right?

eatyourgreens: This is what I was thinking of:
>That said, the header and footer elements are exceptions to the rule. These elements should only be exposed to screen readers when they are children of the body element. In other words screen readers should not be aware of multiple headers/footers on a page - so providing extended labels is not necessary.
https://accessibility.blog.gov.uk/2016/05/27/using-navigation-landmarks/#comment-2273

rogerhutchings: Site header _doesn't_ have a heading

eatyourgreens: Ok, but labelled headers aren't supported. Maybe have just one header landmark on the page, around the page heading?

rogerhutchings: From further testing / research:

- I can't see anything in the WCAG docs that say you _can't_ use a label on a header, their examples include labels on `div`s and whatnot;
- I was following [your suggested solution](https://github.com/zooniverse/front-end-monorepo/issues/671#issuecomment-482034433), which seemed reasonable?
-  I've just tested it in VoiceOver, and it looks like labelled headers are supported:

     

    (Sorry for the crappy photo, I couldn't get screenshots working at the same time as VO.) Are there screenreaders that don't, maybe?
- To prevent duplicate words, I'm going to remove them from the `aria-label`s, e.g. `<nav aria-label="Project" />` should produce `Project navigation` in a screenreader, instead of `Project navigation navigation` - is that cool with you?
- I have a healthy respect for the number of shortcuts used by those using assistive tech

eatyourgreens: My suggestion was wrong, sorry. Reading that discussion above, we're marking up the page banner and my understanding is there should be only one banner.

eatyourgreens: >We have a lot of uppercase links, which are both shouty and can apparently be read as abbreviations in some cases? I'll go through and add labels

Please don't do this.

rogerhutchings: Okay

rogerhutchings: > Please don't do this.

Any references for why not?

eatyourgreens: Because those links already have text, and the uppercase is CSS right? Don't add extra markup because VoiceOver happens to be the IE of screenreaders. I don't think we should be duplicating existing link text.

rogerhutchings: They do, and it is. It doesn't seem like a lot to add (especially since presumably Apple devices _only_ have VO?), but if you're confident that other screenreaders won't render it the same way, then I'll defer to your experience

rogerhutchings:
Okay, I've tweaked the labels / roles a bit, this should be ready for a final review

packages/app-project/src/components/ProjectHeader/ProjectHeader.js
eatyourgreens: The most common OS/screenreader combinations are probably still Windows/JAWS and Windows/NVDA.

eatyourgreens:
Looks good, but the footer isn't showing as a landmark when I try this with VO/Safari or VO/Chrome.

I still think it makes more sense for the page header to wrap the page heading, too.

packages/app-project/src/components/ProjectHeader/ProjectHeader.js
```diff
@@ -23,7 +23,7 @@ function ProjectHeader (props) {
   const { className, navLinks, projectHomeLink, size, title } = props
 
   return (
-    <StyledBox className={className}>
+    <StyledBox as='header' className={className} role='presentation'>
```
eatyourgreens: I'd take off the presentation role here, which removes the element completely from the page while browsing, and make this the page header.

rogerhutchings:
Yeah, I found that too. It's rendered as a footer, so `contentinfo` _should_ be implied; I can ignore it for VO as we've done with link labels, or add a role, your call.

eatyourgreens:
I'd ignore it. I don't see anything obviously wrong with the footer markup.

By the way, here's a more complete answer about why not to label links: https://stackoverflow.com/a/53825307/10951669

PR #931, opened 2019-06-19
Add domain to the cookie being set

### LINKED ISSUE #903: Color theme setting not respected
**Package** app-project

From Talk (https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/talk/2110/855185?comment=1682277):

> I’ve noticed a little problem with the new interface: The Dark Theme switches back to the Light Theme everytime I reload the page

srallen:
Package: app-project

Closes #903

Describe your changes:
Adds the domain to the cookie being set if in production. I'm not sure how to go about testing this.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/stores/UI.js
+++ b/packages/app-project/stores/UI.js
@@ -19,7 +19,11 @@ const UI = types
       const modeDisposer = autorun(() => {
         // process.browser doesn't exist in the jsdom test environment
         if (process.browser || process.env.BABEL_ENV === 'test') {
-          document.cookie = `mode=${self.mode}; path=/; max-age=31536000`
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Looks good, but why is the domain only being set in production? Surely you'd want it set in staging and development too?

srallen:
> Looks good, but why is the domain only being set in production? Surely you'd want it set in staging and development too?

The app is loading up on the same domain in staging and dev though, so I thought it wouldn't be necessary. I can change that for the sake of consistency.

eatyourgreens:
It's not a priority at the moment but that might be worth doing once we have staging set up.

PR #935, opened 2019-06-19
Add ADR 15

srallen:
Package: lib-classifier

Describe your changes:
Adds ADR 15 describing a high level API for the drawing tools based on our discussion today. Two more thoughts come to mind:

- How do we handle multi-image subjects and different streams for different frames?
- How do we handle selecting and editing a mark or deleting it?

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/docs/arch/adr-15.md
+++ b/docs/arch/adr-15.md
@@ -6,12 +6,10 @@ June 19, 2019
 
 The way the drawing tools currently function on Panoptes-Front-End (PFE) have numerous issues including:
 
-- Re-rendering the DOM more than necessary
-- Updating the classification annotation on each touch or pointer event
+- Updating the classification annotation on each touch or pointer event which causes unnecessary re-rendering of the DOM
+- The separation concerns are not clear between components and stores. Multiple components can update the annotation making it hard to debug or add new features to. 
+  - Example: The `MarkingsRenderer` and the `FrameAnnotator` both call change handlers that update the classification annotation? Can the drawing annotation be updated by both or is one solely responsible? It is unclear by reading the code. Why does something named `MarkingsRenderer` update the annotation?
 
-Both of these can cause significant performance issues.
-
-- The separation concerns are not clear between components and stores. Multiple components can update what is being rendered making it hard to debug or add new features to. 
 
 ## Decision
 
@@ -23,11 +21,12 @@ What we do not want to do:
 
 What we do want to do:
 - Have a component, the interactive layer, that manages the interaction with the marks and pointer and touch events.
-  - It should not allow events to bubble so the events are encapsulated to just the interaction with the subject. This is to help prevent browser scroll during drawing.
+  - The interactive layer should not allow events to bubble so the events are encapsulated to just the interaction with the subject. This is to help prevent browser scroll during drawing. An attempted fix on PFE for reference: zooniverse/Panoptes-Front-End#5411
   - Events will be observed and be streamed via an observable. We will use rx.js to create an observer/observable event stream.
   - The last state of the event stream will be used to add an annotation to the classification when next or done is clicked.
 - Have a component, the markings renderer, that manages the rendering of previous annotation marks as read only. It will hide them if hide previous marks is toggled.
 - These two components will initially use separate SVG layers that connect to the same stores. Later any duplicated code will be refactored into a higher order component.
+- Have a multi-image subject viewer that will create an interactive layer and markings renderer for each frame. Each interactive layer will initialize its own event stream. We will have to continue to support the configuration of having drawing marks across all frames and of independent marks for each frame. TODO: example of what these mean and how they will be accomplished.
 
 ## Status
 
@@ -37,4 +36,3 @@ Proposed
 
 - This is the first time rx.js will be used in our code and there will be a learning curve.
 - The current feedback code may need refactoring as it is written only for the D3.js interactive plot subject viewers.
-- How do we integrate this with multi-image subjects?
--- a/docs/arch/adr-15.md
+++ b/docs/arch/adr-15.md
@@ -7,7 +7,7 @@ June 19, 2019
 The way the drawing tools currently function on Panoptes-Front-End (PFE) have numerous issues including:
 
 - Re-rendering the DOM more than necessary
-- Updating the classification annotation on each touch or click event
+- Updating the classification annotation on each touch or pointer event
 
 Both of these can cause significant performance issues.
 
@@ -16,16 +16,16 @@ Both of these can cause significant performance issues.
 ## Decision
 
 What we do not want to do:
-- Re-render every time a click or touch event happens
-- Add to the annotations to the classification every time a click or touch event happens
-- Do not support more than one drawing task in a step
+- Re-render on every pointer or touch event.
+- update annotation state while drawing is in progress.
+- support more than one drawing task in a step.
 - Use D3.js since it has its own internal data store and it would be complicated to integrate that with a observable stream.
 
 What we do want to do:
-- Have a component, the interactive layer, that manages the interaction with the marks and the click and touch events.
+- Have a component, the interactive layer, that manages the interaction with the marks and pointer and touch events.
   - It should not allow events to bubble so the events are encapsulated to just the interaction with the subject. This is to help prevent browser scroll during drawing.
-  - The events will be observed and be streamed via an observable. We will use rx.js to create an observer/observable event stream.
-  - The last state of the event stream will be used to add annotations to the classification when next or done is clicked.
+  - Events will be observed and be streamed via an observable. We will use rx.js to create an observer/observable event stream.
+  - The last state of the event stream will be used to add an annotation to the classification when next or done is clicked.
 - Have a component, the markings renderer, that manages the rendering of previous annotation marks as read only. It will hide them if hide previous marks is toggled.
 - These two components will initially use separate SVG layers that connect to the same stores. Later any duplicated code will be refactored into a higher order component.
 
--- a/docs/arch/adr-15.md
+++ b/docs/arch/adr-15.md
@@ -9,7 +9,8 @@ The way the drawing tools currently function on Panoptes-Front-End (PFE) have nu
 - Updating the classification annotation on each touch or pointer event which causes unnecessary re-rendering of the DOM
 - The separation concerns are not clear between components and stores. Multiple components can update the annotation making it hard to debug or add new features to. 
   - Example: The `MarkingsRenderer` and the `FrameAnnotator` both call change handlers that update the classification annotation? Can the drawing annotation be updated by both or is one solely responsible? It is unclear by reading the code. Why does something named `MarkingsRenderer` update the annotation?
-
+- Drawing tools have a complex API that involves exposing static methods to be called by their parent component
+- Annotation / classification payloads have no consistent standards for describing data: some tools mark rotation in differing directions, for example.
 
 ## Decision
 
--- a/docs/arch/adr-15.md
+++ b/docs/arch/adr-15.md
@@ -28,7 +28,7 @@ What we do want to do:
 - Have a component, the markings renderer, that manages the rendering of previous annotation marks as read only. It will hide them if hide previous marks is toggled.
 - These two components will initially use separate SVG layers that connect to the same stores. Later any duplicated code will be refactored into a higher order component.
 - Have a multi-image subject viewer that will create an interactive layer and markings renderer for each frame. Each interactive layer will initialize its own event stream. We will have to continue to support the configuration of having drawing marks across all frames and of independent marks for each frame. TODO: example of what these mean and how they will be accomplished.
-
+- Have a schema, or set of schemas, describing annotations.
 ## Status
 
 Proposed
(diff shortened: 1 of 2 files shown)


docs/arch/adr-15.md
```diff
@@ -0,0 +1,40 @@
+# ADR 15: Drawing tools API
+
+June 19, 2019
+
+## Context
+
+The way the drawing tools currently function on Panoptes-Front-End (PFE) have numerous issues including:
+
+- Re-rendering the DOM more than necessary
+- Updating the classification annotation on each touch or click event
```
eatyourgreens:
```suggestion
- Updating the classification annotation on each touch or pointer event
```

eatyourgreens: I think this is what's causing the re-renders, so these two are probably the same problem stated two different ways.

docs/arch/adr-15.md
```diff
@@ -0,0 +1,40 @@
...
+
+Both of these can cause significant performance issues.
+
+- The separation concerns are not clear between components and stores. Multiple components can update what is being rendered making it hard to debug or add new features to.
```
eatyourgreens: Drawing is very difficult to debug but I'm not sure what this means, or which components it's referring to.

docs/arch/adr-15.md
```diff
@@ -0,0 +1,40 @@
...
+- Updating the classification annotation on each touch or click event
+
+Both of these can cause significant performance issues.
+
+- The separation concerns are not clear between components and stores. Multiple components can update what is being rendered making it hard to debug or add new features to. 
+
+## Decision
+
+What we do not want to do:
+- Re-render every time a click or touch event happens
```
eatyourgreens:
```suggestion
- Re-render on every pointer or touch event.
```

docs/arch/adr-15.md
```diff
@@ -0,0 +1,40 @@
...
+- Add to the annotations to the classification every time a click or touch event happens
```
eatyourgreens:
```suggestion
- update annotation state while drawing is in progress.
```

docs/arch/adr-15.md
```diff
@@ -0,0 +1,40 @@
...
+- Do not support more than one drawing task in a step
```
eatyourgreens:
```suggestion
- support more than one drawing task in a step.
```

docs/arch/adr-15.md
```diff
@@ -0,0 +1,40 @@
...
+- Use D3.js since it has its own internal data store and it would be complicated to integrate that with a observable stream.
+
+What we do want to do:
+- Have a component, the interactive layer, that manages the interaction with the marks and the click and touch events.
```
eatyourgreens:
```suggestion
- Have a component, the interactive layer, that manages the interaction with the marks and pointer and touch events.
```

docs/arch/adr-15.md
```diff
@@ -0,0 +1,40 @@
...
+  - It should not allow events to bubble so the events are encapsulated to just the interaction with the subject. This is to help prevent browser scroll during drawing.
```
eatyourgreens: What does 'It' refer to here?

eatyourgreens: Maybe refer to https://github.com/zooniverse/Panoptes-Front-End/pull/5411 which doesn't fix this but might be helpful to refer to.

docs/arch/adr-15.md
```diff
@@ -0,0 +1,40 @@
...
+  - The events will be observed and be streamed via an observable. We will use rx.js to create an observer/observable event stream.
```
eatyourgreens:
```suggestion
  - Events will be observed and be streamed via an observable. We will use rx.js to create an observer/observable event stream.
```

docs/arch/adr-15.md
```diff
@@ -0,0 +1,40 @@
...
+  - The last state of the event stream will be used to add annotations to the classification when next or done is clicked.
```
eatyourgreens:
```suggestion
  - The last state of the event stream will be used to add an annotation to the classification when next or done is clicked.
```

eatyourgreens: I could be wrong, but won't each drawing task generate a single annotation for that task key?

docs/arch/adr-15.md
```diff
@@ -0,0 +1,40 @@
...
+- Have a component, the markings renderer, that manages the rendering of previous annotation marks as read only. It will hide them if hide previous marks is toggled.
```
eatyourgreens: This should make Hide Previous Marks so much simpler to implement than the current hack.

docs/arch/adr-15.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Status
+
+Proposed
+
+## Consequences
+
+- This is the first time rx.js will be used in our code and there will be a learning curve.
+- The current feedback code may need refactoring as it is written only for the D3.js interactive plot subject viewers.
+- How do we integrate this with multi-image subjects?
```
eatyourgreens: I've been looking at how PFE does this and have some ideas. Should we add a description of the current implementation to this document?

docs/arch/adr-15.md
eatyourgreens: The current mark will have to render on each event, since we're changing it, but we want to avoid re-rendering the whole drawing surface and any static marks.

`React.memo()` and pure components, with sensibly designed props, will also help here.

docs/arch/adr-15.md
srallen: Yes, it is. I can combine this into one statement.

docs/arch/adr-15.md
srallen: I'll clarify this with an example.

docs/arch/adr-15.md
srallen: The interaction layer. I'll update that to state it explicitly.

docs/arch/adr-15.md
srallen: It's an array for the drawing task, so I was thinking of it as multiple annotations. I can clarify this a bit more.

docs/arch/adr-15.md
srallen: Yeah, I think we should.

docs/arch/adr-15.md
```diff
@@ -0,0 +1,38 @@
...
+  - Events will be observed and be streamed via an observable. We will use rx.js to create an observer/observable event stream.
+  - The last state of the event stream will be used to add an annotation to the classification when next or done is clicked.
+- Have a component, the markings renderer, that manages the rendering of previous annotation marks as read only. It will hide them if hide previous marks is toggled.
+- These two components will initially use separate SVG layers that connect to the same stores. Later any duplicated code will be refactored into a higher order component.
+- Have a multi-image subject viewer that will create an interactive layer and markings renderer for each frame. Each interactive layer will initialize its own event stream. We will have to continue to support the configuration of having drawing marks across all frames and of independent marks for each frame. TODO: example of what these mean and how they will be accomplished.
```
srallen: @eatyourgreens I've added this bullet point describing what we'll do with multi-image subjects. Feel free to edit this file to add your thoughts about it.

srallen:
TODO: we should add a description of how we'll handle mark editing and deletion.

docs/arch/adr-15.md
eatyourgreens: No problem. A single drawing layer, which filters the annotations by frame number before rendering, might be faster than multiple layers.

eatyourgreens:
TODO: how will subtasks be displayed?

docs/arch/adr-15.md
```diff
@@ -0,0 +1,38 @@
...
+June 19, 2019
+
+## Context
+
+The way the drawing tools currently function on Panoptes-Front-End (PFE) have numerous issues including:
+
+- Updating the classification annotation on each touch or pointer event which causes unnecessary re-rendering of the DOM
+- The separation concerns are not clear between components and stores. Multiple components can update the annotation making it hard to debug or add new features to. 
+  - Example: The `MarkingsRenderer` and the `FrameAnnotator` both call change handlers that update the classification annotation? Can the drawing annotation be updated by both or is one solely responsible? It is unclear by reading the code. Why does something named `MarkingsRenderer` update the annotation?
+
```
rogerhutchings:
```suggestion
- Drawing tools have a complex API that involves exposing static methods to be called by their parent component
- Annotation / classification payloads have no consistent standards for describing data: some tools mark rotation in differing directions, for example.
```

docs/arch/adr-15.md
```diff
@@ -0,0 +1,38 @@
...
+  - Events will be observed and be streamed via an observable. We will use rx.js to create an observer/observable event stream.
+  - The last state of the event stream will be used to add an annotation to the classification when next or done is clicked.
+- Have a component, the markings renderer, that manages the rendering of previous annotation marks as read only. It will hide them if hide previous marks is toggled.
+- These two components will initially use separate SVG layers that connect to the same stores. Later any duplicated code will be refactored into a higher order component.
+- Have a multi-image subject viewer that will create an interactive layer and markings renderer for each frame. Each interactive layer will initialize its own event stream. We will have to continue to support the configuration of having drawing marks across all frames and of independent marks for each frame. TODO: example of what these mean and how they will be accomplished.
+
```
rogerhutchings:
```suggestion
- Have a schema, or set of schemas, describing classifications
```

I can see arguments for both - if drawing tools are separate packages, then it may make more sense for the schema to live with the code. Alternatively, having one schema as a package provides a single canonical reference, and drawing tool packages can depend on the version of that package that supports them.

docs/arch/adr-15.md
eatyourgreens: Multi-image subjects are complicated by the option to show the images side-by-side or in a flipbook. If they're all visible together, then PFE renders a new drawing layer for each one. However, this causes headaches when handling events eg. deleting or moving a mark has to be copied across layers. Opening subtasks gets weird if each copy of a mark has its own task component.

See [Space Warps HSC](https://www.zooniverse.org/projects/aprajita/space-warps-hsc/classify) for a live project that draws on four frames simultaneously with the point tool.

eatyourgreens: In PFE, the drawing layer is also used to handle zoom, pan and rotation transformations, which also gets clunky with projects like Space Warps, where the transformations are not copied across frames.

docs/arch/adr-15.md
eatyourgreens: I've edited this to suggest a schema for annotations, rather than classifications.

eatyourgreens: When a developer adds a new tool, or modifies an existing tool, it would helpful if the code for new models and schemas didn't have to be modified in multiple places.

srallen:
> how will subtasks be displayed?

I don't think this has totally been decided, but the recommendation from the ILMS projects is to have the subtask be a movable modal rather than the tooltip display like on current PFE. The tooltip style blocks part of the subject image which is a problem for transcription projects where they might need to read the next lines for context.

docs/arch/adr-15.md
srallen: That Space Warp's classify page is crashing for me, so I can't check it out. I'm not sure what's wrong with it.

docs/arch/adr-15.md
srallen: I'm not sure how to avoid this because we should have a model for the tool type in the workflow resource and a model for the annotation, which is added to the classification annotation. As long as we document that models need to be added in both places, I don't think it's too big of a lift to ask developers to update both.

docs/arch/adr-15.md
srallen: I've updated the description of what we will do to include the two options proposed for multi-image subjects. Is this clear enough for an ADR?

srallen:
> TODO: we should add a description of how we'll handle mark editing and deletion.

I've added a bullet point suggesting we use the component state of the interactive layer to handle which mark is actively being edited and have the edit and delete events be component class methods.

srallen:
@mcbouslog is this ADR's description still in line with what you've been doing?

docs/arch/adr-15.md
eatyourgreens: A plug-in architecture might be appropriate, where the models are generated from a list of installed plugins.

eatyourgreens:
Looks good but I don’t think there’s any requirement for independent zoom controls across image frames. There’s an issue on PFE (might be stale) asking us to implement a consistent zoom across frames, which we can’t do with the PFE implementation.

eatyourgreens:
Here's the PFE Zoom issue, which should be implemented in the new classifier. https://github.com/zooniverse/Panoptes-Front-End/issues/3465

docs/arch/adr-15.md
eatyourgreens: I'm not sure what's going on with Space Warps, but the feedback code is crashing the classifier. That link worked when I posted it last week.

mcbouslog:
> @mcbouslog is this ADR's description still in line with what you've been doing?

Yes this is in-line with what I've been doing. I think more details can be worked out about multiple frames or the more complicated tools (like ellipse with multiple events) as we implement, and can update this accordingly.

PR #936, opened 2019-06-20
Add notification on failed builds

rogerhutchings:
Ping the `#frontend-rewrite` channel on Slack if the master build process fails on Jenkins


rogerhutchings:
yeah, I realised when I checked the jenkins status for this branch 🤦‍♂ i've just pushed a change that should work, but i'll try rewriting it with the declarative syntax

rogerhutchings:
No - I didn't want to accidentally push up the wrong image, although since it includes the branch name in the image title I'm going to quickly test it out without the `master` conditional

`get-app-folders.sh` was used by the old scripted Jenkinsfile - I've dispensed with the magic and just written a separate stage for each deployed app, which makes things a lot more explicit

rogerhutchings:
@camallen https://jenkins.zooniverse.org/job/Zooniverse%20GitHub/job/front-end-monorepo/job/PR-936/11/console 🎉

PR #937, opened 2019-06-20
Add roadmap link, badges to README

rogerhutchings:
Adds a link to the Roadmap Trello board, and a few badges for bling


eatyourgreens:
Yep, in a private window the link gives 'Board not found'.

rogerhutchings:
Hadn't thought about it - I don't actually mind whether it's public or private, but that won't block this PR

eatyourgreens:
The Readme is public, so links should probably go to public documents if we want potential contributors to read them.

PR #940, opened 2019-06-20
Add specs and refactor getCookie helper

### LINKED ISSUE #903: Color theme setting not respected

srallen:
Package: app-project

Fourth time is the charm? Closes #903

Describe your changes:
I initially added some logs which led me to add specs and debug the `getCookie` helper I had added. Sometimes calling `pop()` on an array doesn't actually get you what you think it does. If a cookie has a lot of key value pairs, sometimes the last one in the array is not a match to the regex, and will be an empty string, even if the key-value you're looking for is indeed in the cookie. I think the refactor I did will work, but happy to have any suggestions on making this more robust. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/stores/UI.js
+++ b/packages/app-project/stores/UI.js
@@ -6,7 +6,6 @@ const UI = types
   .model('UI', {
     mode: types.optional(types.enumeration('mode', ['light', 'dark']), () => {
       const mode = getCookie('mode')
-      console.log(`Instantiating UI store with ${(mode) ? mode : 'light'}`)
       return (mode) ? mode : 'light'
     })
   })
@@ -31,13 +30,11 @@ const UI = types
       if (process.browser || process.env.BABEL_ENV === 'test') {
         const storedMode = getCookie('mode')
         if (self.mode !== storedMode) {
-          console.log(`Color mode ${self.mode} does not match ${storedMode} in cookie`)
           if (process.env.NODE_ENV === 'production') {
             document.cookie = `mode=${self.mode}; path=/; domain=zooniverse.org; max-age=31536000`
           } else {
             document.cookie = `mode=${self.mode}; path=/; max-age=31536000`
           }
-          console.log(`Set ${self.mode} in cookie`)
         }
       }
     },
--- a/packages/app-project/stores/UI.js
+++ b/packages/app-project/stores/UI.js
@@ -6,7 +6,7 @@ const UI = types
   .model('UI', {
     mode: types.optional(types.enumeration('mode', ['light', 'dark']), () => {
       const mode = getCookie('mode')
-      return (mode) ? mode : 'light'
+      return mode || 'light'
     })
   })
 
--- a/packages/app-project/stores/UI.js
+++ b/packages/app-project/stores/UI.js
@@ -6,7 +6,8 @@ const UI = types
   .model('UI', {
     mode: types.optional(types.enumeration('mode', ['light', 'dark']), () => {
       const mode = getCookie('mode')
-      return mode || 'light'
+      console.log(`Instantiating UI store with ${(mode) ? mode : 'light'}`)
+      return (mode) ? mode : 'light'
     })
   })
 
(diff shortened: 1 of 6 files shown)


eatyourgreens:
If the bug is caused by more than one cookie being present, then my first step would be to set more than one cookie in the tests, so that I've got a failing test that catches the bug.

MDN has [a simple cookie framework](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie/Simple_document.cookie_framework) so maybe consider using that instead of rolling your own.

If the regex is causing problems, then maybe just [iterate over all cookies](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#Example_5_Check_a_cookie_existence) to find the mode cookie instead?
```js
const matchingCookies = document.cookie.split(';').filter(cookie => cookie.trim().startsWith('mode=')
```

packages/app-project/stores/UI.js
```diff
@@ -6,7 +6,7 @@ const UI = types
   .model('UI', {
     mode: types.optional(types.enumeration('mode', ['light', 'dark']), () => {
       const mode = getCookie('mode')
-      return mode || 'light'
+      return (mode) ? mode : 'light'
```
eatyourgreens: PFE prefers `||` to set defaults and warns agains using ternary expressions like this. Do we want to carry that over into the new app?
http://eslint.org/docs/rules/no-unneeded-ternary.html

packages/app-project/stores/helpers/cookie.js
```diff
@@ -1,10 +1,15 @@
+import { compact } from 'lodash'
+
 // Borrowed from https://stackoverflow.com/a/25490531/5372931
 export function getCookie (name) {
   // process.browser doesn't exist in the jsdom test environment
   if (process.browser || process.env.BABEL_ENV === 'test') {
-    const value = document.cookie.match('(^|[^;]+)\\s*' + name + '\\s*=\\s*([^;]+)')
+    const match = document.cookie.match('(^|[^;]+)\\s*' + name + '\\s*=\\s*([^;]+)')
+
+    const compactMatchArray = compact(match)
```
eatyourgreens:
```suggestion
    const compactMatchArray = match.filter(Boolean)
```
does the same thing (removes all falsy values from `match`.)

`compactMatchArray` could be better named. `matchingCookieValues` maybe? I prefer array names that describe what the arrays are storing.

eatyourgreens:
Sorry, one more thing: I'm not really understanding what the original bug was here. Is it that the array of matched items isn't being handled properly when it's empty, or that there are cases in production where there's more than one match for cookies named `mode`? A test case would definitely help me understand what this code is fixing.

>If a cookie has a lot of key value pairs, sometimes the last one in the array is not a match to the regex
`String.match(pattern)` always returns an array of matches to `pattern`, so I'm not following what this means. Is it that our pattern has an error in it?

srallen:
> I'm not really understanding what the original bug was here. Is it that the array of matched items isn't being handled properly when it's empty, or that there are cases in production where there's more than one match for cookies named mode?

The bug is more than one key value pair in the same cookie. Say you have a cookie like this:
`mode=light; another=pair; path=/; max-age=31536000`

With the current regex, this would return:
`['mode=light', 'light', '']`

I added a test using that example cookie. So, it's not that the array is empty. It's the last value is empty because the last key-value pair in the cookie is not a match, so you can't rely on `pop()` to return the value you are looking for.

packages/app-project/stores/UI.js
srallen: FWIW, I find `||` a bit harder to read for some reason and I changed it when I first noticed that sometimes `getCookie` was returning an empty string. I meant to have switched it back.

I'm still of the opinion that if we want to match PFE's linting rules, then we should drop standard and switch back to the airbnb rule set then. I don't have a strong opinion about which rule set we follow.

srallen:
I'm looking at the MDN docs simple cookie functions. It's using a regex too, but might be a bit more robust. I'm going to compare.

srallen:
The MDN library looks good, but it's GPL so we can't use it. I decided to switch to the iterative approach to finding the cookie rather than use a regex. Tests still pass.

PR #942, opened 2019-06-21
Switch to use js-cookie

srallen:
Package: app-project

Describe your changes:
Switches the retrieval of the value of the mode in the cookie to an external package js-cookie. Setting the cookie is fine, and the response for the page is including the mode in the cookie. The issue still seems to be with the function getting the value out of it on instantiation, so let's not try to use our own anymore because it might not be accounting for all of the variables of how a cookie can be. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
Looks good to me. Should it be using `Cookie.set(name, value)` to set the cookie too?

srallen:
I wanted to keep the scope of this PR to just fixing the retrieval of the cookie value. If this fixes it and we are happy with it, then I can update the setting in a separate PR.

Issue #755, opened 2019-04-24
Add alumni to team page


rogerhutchings:
- [ ] Add alumni filter to the bottom of the sidebar on the team page
- [ ] Ensure alumni members don't appear in other filters
- [ ] Add alumni data to Contentful

Issue #754, opened 2019-04-24
Finish team member profiles


rogerhutchings:
- [ ] Add team member photos to contentful
- [ ] Add team member photo to component
- [ ] Add team member bio to component
- [ ] Add team member twitter link to component

PR #944, opened 2019-06-24
Add the Our Team page to app-content

### LINKED ISSUE #753: Add remaining team members to Contentful

rogerhutchings:
Adds the Our Team page to app-content

- A lot of this is copied over from the publications page
- Tests are basic, more to come but I wanted to get this out in time for some updates happening (and it's a non-critical page)
- I'm getting some styled-components errors about className not being passed down BUT IT IS

Otherwise, should be good to go.

fix #753, #754, #755 

cc zooniverse/Panoptes-Front-End#5417

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/app-content-pages/src/screens/Team/Team.js
```diff
@@ -0,0 +1,84 @@
...
+          people={team.people}
+        />
+      ))}
+    </article>
+  )
+
+  const sidebar = (
+    <Box gap='small'>
+      {filters.map(filter => (
+        <div key={filter.name} >
```
srallen: Not sure what's best for accessibility, but this sidebar is a menu, so I was thinking initially that this should be an unordered list, but I've read that sometimes lists can be bad for accessibility. @eatyourgreens do you have a suggestion?

rogerhutchings:
> I just noticed that the team for everyone but Chicago is the place, not the institution name. Maybe it should be changed from The Adler Planetarium to Chicago? This is now in Contentful?

It is in Contentful, and this was a straight copy from the existing page - I don't have strong feelings one way or another

packages/app-content-pages/src/screens/Team/Team.js
eatyourgreens: Lists of links should be marked up as lists.

packages/app-content-pages/src/screens/Team/components/Person/components/TwitterLink/locales/en.json
```diff
@@ -0,0 +1,5 @@
+{
+  "TwitterLink": {
+    "a11yTitle": " %(name)s's Twitter feed"
```
eatyourgreens: I don’t think this is needed. The link text should be the Twitter username.

rogerhutchings:
At the moment, they're just icon links; we can always tweak this later, I just want to get the bulk of it in.

eatyourgreens:
I missed this in review, but tabbing through the page skips past the filter buttons. The first focussed item is Cliff's twitter link, but I'd expect it to be one of the filter buttons. Tab order should respect the visual layout on the page.

eatyourgreens:
The list of filters should maybe be a radio button group too, since only one can be selected at a time, but that's an enhancement.

PR #958, opened 2019-06-26
Remove standard as dep

rogerhutchings:
So the build is breaking, because `@zooniverse/standard` is not being resolved properly. This PR removes it as a dependency for the time being, just to get the build up and running again

### DIFF
--- a/bin/panic-button.sh
+++ b/bin/panic-button.sh
@@ -2,27 +2,18 @@
 
 # Trashes all build folders and `node_modules`
 
-printf 'Trashing all build folders and `node_modules` folders...\n'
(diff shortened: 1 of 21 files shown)


rogerhutchings:
Okay, tests are breaking because of enzyme, I'm going to fix this up here.

rogerhutchings:
testing the actual build in https://jenkins.zooniverse.org/job/Zooniverse%20GitHub/job/front-end-monorepo/job/PR-958/

eatyourgreens:
Do we even care abut the linter in a build anyway? It should be a dev dependency, and ignored in production.

PR #960, opened 2019-06-26
Add drawing task point tool model

mcbouslog:
Package: lib-classifier

Describe your changes:
One of a few tiny PRs related to adding the drawing task point tool. No required order to review/merge.
This adds the point tool model for the drawing store.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/store/tasks/DrawingTask.js
```diff
@@ -10,7 +11,14 @@ const Drawing = types.model('Drawing', {
   })),
   help: types.optional(types.string, ''),
   instruction: types.maybe(types.string),
-  tools: types.array(types.frozen()),
+  tools: types.array(types.union({
+    dispatcher: (snapshot) => {
+      switch (snapshot.type) {
+        case 'point':
+          return Point
```
eatyourgreens: Small stylistic point but you can surround cases in {} to isolate scope.
```js
case 'point': {
  return Point
}

File: packages/lib-classifier/src/store/tasks/drawingTools/Point.js [hunk 2]
```diff
@@ -0,0 +1,12 @@
+import { types, onSnapshot } from 'mobx-state-tree'
+
+const Point = types.model('Point', {
+  color: types.optional(types.string, '#00ff00'),
```
eatyourgreens: Do we want to code colours into the tools, or look them up from a shared lookup table?

mcbouslog: Shared lookup table sounds better, but maybe I'm not thinking about this quite right. And it's subject to change if we change how the tools are built in the lab. Hm, maybe for now I just change to `types.maybe(types.string)` and can revisit.

srallen: I think `types.optional(types.string, '')` might be better in this case? It depends on if `color` can ever be null or undefined.

packages/lib-classifier/src/store/tasks/DrawingTask.js
```diff
@@ -1,16 +1,22 @@
...
 const Drawing = types.model('Drawing', {
-  answers: types.array(types.frozen({
-    label: types.string,
-    next: types.maybe(types.string)
-  })),
   help: types.optional(types.string, ''),
   instruction: types.maybe(types.string),
-  tools: types.array(types.frozen()),
+  tools: types.array(types.union({
+    dispatcher: (snapshot) => {
```
simensta: yarn lint detects an error on line 11: `11:28  error  Expected to return a value at the end of arrow function 'dispatcher'`.

Adding a `default` clause that returns `null` should resolve the error.

Issue #823, opened 2019-05-14
Should we make annotation value types consistently iterables?


srallen:
**Package** lib-classifier

**Feature or Issue Description**
The variability of type of the annotation value has been a consistent source of issues and pain for the aggregation service and possibly other post-classification services. For example, this recent PR was to account for the possibility of the annotation value either being an array, object, number, or string: https://github.com/zooniverse/aggregation-for-caesar/pull/154

I'd like to propose that we always make the annotation value an array with the annotation values contained therein. So, the change I see for the following tasks are:

- Single choice: `value: 0` to `value: [0]`
- Multiple choice: already an array of numbers
- Drawing: already an array of objects
- Crop: `value: {}` to `value: [{}]`
- Text: `value: ""` to `value: [""]`
- Survey: already an array of objects
- Dropdown: already an array of object
- Slider: `value: 3` to `value: [3]`
- Highlighter: already an array of objects

In addition to this, I'm also thinking about adding a `taskType` which would be a string of the type like single, multiple, survey, etc so the type is explicitly sent in along with the annotation value rather than rely on type checking to try to figure out what the annotation task type was. For example:

```
{
  task: 'T1',
  value: [0],
  taskType: 'single'
}
```

Any reason this is a bad idea? cc @camallen @zwolf @CKrawczyk

eatyourgreens:
In https://github.com/zooniverse/Panoptes-Front-End/issues/4478 we considered adding a schema for annotations. So you'd publish a task with a URL that describes the annotations. Classifications would include that URL, so that Caesar etc. can look it up and find out what format annotations should be in.

If the task is changed/updated it should publish a new schema, then any new classifications use the updated schema URL and aggregation knows the format has changed.

It might be too heavy-handed of an approach, but it would allow us to do things like change the data format for freehand tasks, without breaking aggregation code that needs to be backwards compatible with old classifications.

Otherwise, I think we're locked into a single annotation type for each task type, because all annotations for that type re expected to be in that format.

CKrawczyk:
I like the idea of passing along the `taskType` inside the classification, that could make things like auto-configuration easier and reducer the need for checking data types when multiple task types share an extractor (e.g. https://github.com/zooniverse/aggregation-for-caesar/blob/master/panoptes_aggregation/extractors/question_extractor.py#L56-L60).

That should be more than enough information for the aggregation code to know what data shape to expect, I don't really think forcing everything into an array is needed.

[...]

PR #961, opened 2019-06-26
Add drawing task point tool annotation model

mcbouslog:
Package: lib-classifier

Describe your changes:
One of a few tiny PRs related to adding the drawing task point tool functionality. No required order to review/merge.
This adds a drawing task point tool annotation model for the classifications store.
Does not yet incorporate schema per [ADR-7](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-7.md), but does include `taskType` as per #823.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/store/annotations/DrawingToolAnnotations/PointAnnotation.js
```diff
@@ -0,0 +1,10 @@
+import { types } from 'mobx-state-tree'
+
+const PointAnnotation = types.model('PointAnnotation', {
+  tool: types.number,
+  toolType: types.string,
```
eatyourgreens:
```suggestion
  toolType: 'point',
```

packages/lib-classifier/src/store/annotations/DrawingToolAnnotations/PointAnnotation.spec.js
```diff
@@ -0,0 +1,10 @@
+import { types } from 'mobx-state-tree'
+
+const PointAnnotation = types.model('PointAnnotation', {
+  tool: types.number,
+  toolType: types.string,
+  x: types.number,
+  y: types.number
+})
+
+export default PointAnnotation
```
eatyourgreens: This should be a spec for the point annotation model, not a copy of the model definition.

packages/lib-classifier/src/store/annotations/DrawingAnnotation.js
```diff
@@ -0,0 +1,15 @@
+import { types } from 'mobx-state-tree'
+import Annotation from './Annotation'
+import { PointAnnotation } from './DrawingToolAnnotations'
+
+const Drawing = types.model('Drawing', {
+  value: types.array(types.union({
+    dispatcher: (snapshot) => {
+      if (snapshot.toolType === 'point') return PointAnnotation
```
eatyourgreens: This is fine for now but is going to get clunky when we want to add new tools.

packages/lib-classifier/src/store/annotations/DrawingToolAnnotations/PointAnnotation.spec.js
mcbouslog: 🤦‍♂

packages/lib-classifier/src/store/annotations/DrawingAnnotation.js
mcbouslog: Yes plan to refactor to switch statement at minimum when add next tool.

eatyourgreens: Longer term, this should be generated dynamically from the list of available tools, otherwise developers have to remember to touch the code in multiple places when adding a new drawing tool.

packages/lib-classifier/src/store/annotations/DrawingAnnotation.js
```diff
@@ -0,0 +1,15 @@
...
+    }
```
eatyourgreens: Drawing annotations should record the frame number that the drawing was made on. PFE records this in the `value` object but should it be declared on this model, since it applies to all drawing tools (and adding it to each annotation model would be clunky?)

eatyourgreens: `toolIndex` is another value key that's used by all drawing tools, so could also be declared in one place.

mcbouslog: Updated to use `Object.assign` to combine the generic attributes `frame` and `toolIndex` with what will be the unique tool attributes, but I'm not sure if there's a more elegant way utilizing MST types. I attempted to use `compose` and `union` unsuccessfully.

packages/lib-classifier/src/store/annotations/DrawingAnnotation.spec.js
```diff
@@ -0,0 +1,28 @@
...
+}
+
+describe('Model > DrawingAnnotation', function () {
+  it('should exist', function () {
+    const drawingAnnotationInstance = DrawingAnnotation.create(drawingAnnotation)
+    expect(drawingAnnotationInstance).to.exist()
+    expect(drawingAnnotationInstance).to.be.an('object')
+  })
+
+  it('should throw an error without frame number', function () {
```
mcbouslog: this and the 'should throw an error without toolIndex number' tests are failing, and I'm not sure why. I'll revisit soon, but pushed up if something quick jumps out.

packages/lib-classifier/src/store/annotations/DrawingAnnotation.js
```diff
@@ -0,0 +1,22 @@
+import { types } from 'mobx-state-tree'
+import Annotation from './Annotation'
+import { PointAnnotation } from './DrawingToolAnnotations'
+
+const Drawing = types.model('Drawing', {
+  value: types.array(Object.assign(
+    types.frozen({
+      frame: types.number,
+      toolIndex: types.number
```
srallen: This is just `tool` in PFE, so we should keep using that for consistency with post-classification analysis unless we have a reason to change it.

packages/lib-classifier/src/store/annotations/DrawingAnnotation.js
srallen: Instead of trying to merge it together here with `Object.assign`, you can have a generic model declaration in the annotations folder and compose them in the export of `PointAnnotation` similar to other compositions we do elsewhere. Also, `types.frozen` will cause an issue here because the frozen type is intended to be immutable, where this will change as someone is annotating, so `types.model` should be used instead. If free to ask me in person if this isn't making sense.

packages/lib-classifier/src/store/annotations/DrawingAnnotation.spec.js
srallen: I think we can have the model default tool and frame to 0 with the optional type. The task area of the drawing task should auto select the first tool and the subject viewer will default to the first frame.

packages/lib-classifier/src/store/annotations/DrawingAnnotation.js
mcbouslog: That worked, thanks! Updates incoming...

PR #962, opened 2019-06-28
Add logging for sentry

rogerhutchings:
- Add DSNs to kube deploy setup
- Add sentry package
- Add logger functions (I've kept them generic so we can swap out the internals without too much pain)

### DIFF
--- a/kubernetes/deployment.tmpl
+++ b/kubernetes/deployment.tmpl
@@ -53,6 +53,8 @@ spec:
               value: production
             - name: PANOPTES_ENV
               value: production
+            - name: SENTRY_DSN
+              value: https://2a50683835694829b4bc3cccc9adcc1b@sentry.io/1492691
 ---
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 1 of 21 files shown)


rogerhutchings:
It doesn't need an `.env` to run - it just allows you to have one (e.g. if you want to post to Sentry locally). `dotenv-webpack` can also pick up the system env vars, which is where the Sentry DSN will be set in production

srallen:
@rogerhutchings hmm, well then something is going on with this branch for me locally. When I try to run the project app, it fails to compile and has these errors:

`Failed to load /Users/sarah/code/front-end-monorepo/packages/app-project/.env`
`Module not found: Can't resolve '@sentry/browser' in '/Users/sarah/code/front-end-monorepo/packages/app-project/src/helpers/logger`

I've panic-buttoned and bootstrapped a couple of times trying this out.

rogerhutchings:
That's just a warning; the problem is I forgot to push a commit 🤦‍♂Pull, do a bootstrap and you should be okay

PR #967, opened 2019-07-01
Switch to yarn: Add standard back in

rogerhutchings:
~~Requires #966 merging first - I'll remove WIP after that and merge.~~

- Set up `src`-based directory structure (recommended while I was testing rush.js, and felt like a better structure)
- Adds `@zooniverse/standard` back into the repo
- Update some packages
- Fix erroneous `snazzy` require in @zooniverse/standard


rogerhutchings:
Okay, now you can `yarn install` from the top level and it'll install the relevant packages for its workspaces.

PR #969, opened 2019-07-01
Test for clip-path

srallen:
Package: lib-classifier

For #650

Describe your changes:
Updates `addDataLayer` test to also test for the addition of the clip-path attribute

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


srallen:
I've noticed that some of the d3 functions do have tests, but they take a different approach than what I've been adding recently. These are using fixtures with sinon just to check to see if the function is called with the parameters and returning what we expect. I instead had been creating a test SVG node in the jsdom environment and testing the  resulting DOM that the d3 functions create. Both work. Is there a preferred approach here?

srallen:
I'm taking a look at the `addDataMask` tests now and it's hard to create a fixture that is chainable and that function is using attr and append in a chainable way. So I'm choosing in this case to test against the DOM.

PR #970, opened 2019-07-01
Add unit tests for addDataMask

srallen:
Package: lib-classifier

For #650 

Describe your changes:
Adds unit tests for `addDataMask`

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/d3/addDataMask.spec.js
```diff
@@ -1,7 +1,41 @@
...
+    body.appendChild(node)
+    d3svg = d3.select(node)
+  })
+
+  after(function () {
+    body.removeChild(node)
+  })
+
+  afterEach(function () {
+    node.innerHTML = ''
```
mcbouslog: This is just for my understanding, no rush to respond, but what is the reason for resetting the innerHTML after each test? Also, in #968 the test svg node was built for each test compared to these tests where it's built with just before/after, and I'm not sure why one or the other.

srallen: It's because the d3 API being chainable. The `addDataMask` function adds a `clipPath`, then adds a `rect` as a child of `clipPath`. If `addDataMask` is updated to have more nodes added in the chain, I want to be sure that the `innerHTML` of the parent `clipPath` is reset for each test.

The other function in #968 is three `g` elements that are the child of the `svg` so I found it best to just remove the whole `svg` between each test.

PR #975, opened 2019-07-02
Update panoptes-js

rogerhutchings:
- Update dependencies
- Add linter
- Run lint (there are still a fair few things that require manual fixes)

### DIFF
--- a/packages/lib-panoptes-js/package-lock.json
+++ b/packages/lib-panoptes-js/package-lock.json
@@ -14,12 +14,13 @@
 			}
 		},
 		"@sinonjs/formatio": {
-			"version": "2.0.0",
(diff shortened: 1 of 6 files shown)


eatyourgreens:
The remaining linter errors are mostly unused variables in tests or complaints about tests accessing chai properties when they should be making function calls eg. https://github.com/zooniverse/front-end-monorepo/blob/53b5f8677e6f5d7c7afc2497c2d6061bfa881dd0/packages/lib-panoptes-js/src/panoptes.spec.js#L38

The solution would be to switch to dirty-chai.

NB. this is going to be a big problem in the classifier.

rogerhutchings:
Yeah, I ran the tests and some local integration tests - things looked good. I thought I'd leave the linter errors for another day, but I'm going to tackle them before too long.

PR #977, opened 2019-07-02
Update react components

rogerhutchings:
- Update dependencies, including storybook
- Tidy up scripts, `package.json`
- Add plugin for cleaning `dist` folder


srallen:
I've fixed the broken tests and did one minor refactor to get rid of the prop types warning `ThumbnailImage` was also throwing.

rogerhutchings:
thanks @srallen!

packages/lib-react-components/package.json
```diff
@@ -76,6 +76,7 @@
     "react-dom": "~16.8.6",
     "sinon": "~7.3.2",
     "sinon-chai": "~3.3.0",
+    "snazzy": "~8.0.0",
```
eatyourgreens: I installed snazzy globally when I last worked on the linter, so it should already be installed in the root package file.

srallen: I think this might be managed by yarn once we switch over, so I'm going to leave this here for now.

srallen:
Markdownz tests are failing after a linting change. Looking into it now.

srallen:
Ok tests should be passing now. We have to be careful with adding on things like sinon-chai and dirty-chai because there may be cases of tests passing or not passing when they should or shouldn't be with some of the underlying behavior changes.

srallen:
The tests for the react component lib are passing. Travis is now failing with the project app:
```
It looks like there are several instances of 'styled-components' initialized in this application. This may cause dynamic styles not rendering properly, errors happening during rehydration process and makes your application bigger without a good reason.
See https://s-c.sh/2BAXzed for more info.
/home/travis/build/zooniverse/front-end-monorepo/packages/app-project/node_modules/styled-components/dist/styled-components.cjs.js:479
      throw new StyledComponentsError(6);
      ^
Error: Trying to insert a new style tag, but the given Node is unmounted!
- Are you using a custom target that isn't mounted?
- Does your document not have a valid head element?
- Have you accidentally removed a style tag manually?
```

srallen:
Reverting the styled-components update to stay at 4.1.x gets tests running again. I'm going to recommend we hold off on doing minor version updates to styled-components so we can focus on the yarn switchover.

eatyourgreens:
Tilde should mean we only get patch version changes, right? I can never remember how semver works.

packages/lib-react-components/src/Markdownz/Markdownz.spec.js
```diff
@@ -87,7 +89,7 @@ describe('<Markdownz />', function () {
         const baseURL = ''
         const expectedReturnValue = `${baseURL}/talk/search?query=%23${resource}`
         wrapper.instance().buildResourceURL(resource, '#')
-        expect(buildResourceURLSpy).to.have.been.calledOnce()
+        expect(buildResourceURLSpy.withArgs(resource, '#')).to.have.been.calledOnce()
```
eatyourgreens: I don't understand this test. The preceding line called `buildResourceURL(resource, '#')` so what is this expectation checking?

packages/lib-react-components/src/Markdownz/Markdownz.spec.js
```diff
@@ -96,8 +98,8 @@ describe('<Markdownz />', function () {
         const projectSlug = 'zooniverse/snapshot-wakanda'
         const expectedReturnValue = `/projects/${projectSlug}/talk/tag/${resource}`
         wrapper.setProps({ projectSlug })
-        wrapper.instance().buildResourceURL('tigers', '#')
-        expect(buildResourceURLSpy).to.have.been.calledOnce()
+        wrapper.instance().buildResourceURL(resource, '#')
+        expect(buildResourceURLSpy.withArgs(resource, '#')).to.have.been.calledOnce()
```
eatyourgreens: Same question here about testing our test code, not the component.

packages/lib-react-components/src/Markdownz/Markdownz.spec.js
```diff
@@ -107,8 +109,8 @@ describe('<Markdownz />', function () {
         const baseURI = 'https://classrooms.zooniverse.org'
         const expectedReturnValue = `${baseURI}/talk/search?query=%23${resource}`
         wrapper.setProps({ baseURI })
-        wrapper.instance().buildResourceURL('tigers', '#')
-        expect(buildResourceURLSpy).to.have.been.calledOnce()
+        wrapper.instance().buildResourceURL(resource, '#')
+        expect(buildResourceURLSpy.withArgs(resource, '#')).to.have.been.calledOnce()
```
eatyourgreens: Here too. My reading of the code is that `buildResourceURLSpy` spies on `wrapper.instance().buildResourceURL`. I feel like I'm missing something.

Issue #981, opened 2019-07-02
Update docs on how to use Yarn


rogerhutchings:

rogerhutchings:
Fixed in #991

Issue #953, opened 2019-06-25
Production dockerfiles are installing devDependencies


rogerhutchings:
The production dockerfiles use the lerna bootstrap script, which runs a straight `npm install` - if we add the `--production` flag, or set the `NODE_ENV` to `production`, it'll only install actual dependencies, which will slim down the built images and shave some time off builds

rogerhutchings:
It's because we're building our libraries in the Dockerfile, so of course we need the devDependencies.

PR #990, opened 2019-07-03
First pass at ADR 16: Monorepo Tooling

rogerhutchings:
Adds an ADR for monorepo tooling, feedback welcome.

cc #981


docs/arch/adr-16.md
```diff
@@ -0,0 +1,40 @@
...
+When we first started the rebuild, @rogerhutchings set it up as a monorepo. The idea being that there are benefits to separating our code between multiple apps and packages:
+
+  - We can reuse code more easily
+  - Error boundaries are built-in (if we break one thing, we don't necessarily break everything)
+  - To make changes to one thing, we don't need to recompile and rebuild _everything_ in development, at least
+
+...but that working across multiple repos can be pretty tedious: adding a feature which requires changes to more than one package requires multiple pull requests to different repos, for example. So Roger set it up as a monorepo using Lerna. This started out fine, but we hit some issues in the development of Planet Hunters TESS:
+
+  - `styled-components` wasn't collect styling from symlinked dependencies, causing a FOUC on initial render: https://github.com/zooniverse/front-end-monorepo/issues/327
+  - Adding the `@zooniverse/standard` package broke the build because Lerna wasn't correctly linking it up: https://github.com/zooniverse/front-end-monorepo/pull/958
```
eatyourgreens: Something that's been puzzling me: why does the linter break a production build?

docs/arch/adr-16.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Context
+
+When we first started the rebuild, @rogerhutchings set it up as a monorepo. The idea being that there are benefits to separating our code between multiple apps and packages:
+
+  - We can reuse code more easily
+  - Error boundaries are built-in (if we break one thing, we don't necessarily break everything)
+  - To make changes to one thing, we don't need to recompile and rebuild _everything_ in development, at least
+
+...but that working across multiple repos can be pretty tedious: adding a feature which requires changes to more than one package requires multiple pull requests to different repos, for example. So Roger set it up as a monorepo using Lerna. This started out fine, but we hit some issues in the development of Planet Hunters TESS:
```
eatyourgreens:
```suggestion
Working across multiple repos can be pretty tedious: adding a feature which requires changes to more than one package requires multiple pull requests to different repos, for example, and publishing updated packages in the correct order. So Roger set this project up as a monorepo using Lerna. This started out fine, but we hit some issues in the development of Planet Hunters TESS:
```

docs/arch/adr-16.md
rogerhutchings: The production build script was doing a standard `npm install` without the `--production` flag. I _had_ added the `NODE_ENV=production` envinronment variable, but it seemed to be ignored (possibly as the install command was being run in a subshell? not bash-savvy enough, but I thought they inherited their environment). I started noticing the broken linking locally too, but transiently 🤦‍♂ see https://github.com/zooniverse/front-end-monorepo/issues/953

rogerhutchings: Also, the way the build process is set up, we'd have to install the build dependencies for the `lib-` folders, as we need to build a production version of the libraries for the apps to use.

rogerhutchings:
Thanks @eatyourgreens. I'm going to leave this open to let US folks have a look next week.

PR #995, opened 2019-07-08
Update scripts and docs

rogerhutchings:
- Update shell scripts to use yarn
- Update dockerfiles to use yarn
- Update docs

### DIFF
--- a/bin/panic-button.sh
+++ b/bin/panic-button.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Trashes all `node_modules` folders, lockfiles and build artifacts to give you
+# Trashes all `node_modules` folders and build artifacts to give you
 # a clean version of the project.
 
 printf 'Trashing `node_modules` folders and build artifacts...\n'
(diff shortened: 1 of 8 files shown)


bin/panic-button.sh
```diff
@@ -1,13 +1,15 @@
...
+# a clean version of the project.
 
-printf 'Trashing `node_modules` folders and build artifacts...\n'
+printf 'Trashing `node_modules` folders, lockfiles and build artifacts...\n'
 
 ROOT_DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && cd .. && pwd)"
 
 printf "Cleaning \`$ROOT_DIR\`..."
 rm -rf $ROOT_DIR/node_modules
+rm -rf $ROOT_DIR/yarn.lock
```
eatyourgreens: Why is the lock file being deleted?

rogerhutchings: Force of habit. In npm, it gets rewritten all the time anyway; I think yarn is a bit more conservative. I'll leave it in.

PR #996, opened 2019-07-08
Hook up AboutProject component

rogerhutchings:
Hooks up the `AboutProject` component to the store, and adds the project description to the store


eatyourgreens:
Looks like this is forked from #991 rather than `master`.

eatyourgreens:
Changes in the most recent commit look good but we really should be getting these language strings from the project's translation resource.

rogerhutchings:
> we really should be getting these language strings from the project's translation resource

My plan for this is ultimately to have translations as a node of the project model, and then accessor functions for each property looking up the string by the current translation - in theory, we wouldn't need to change the front-end code to accommodate translations.

eatyourgreens:
Cool. That should be written down somewhere so someone can pick it up and do it.

eatyourgreens:
Wait, that works for project titles and descriptions but doesn't work in general because Panoptes resources are nested objects whereas translations are flat dictionaries eg. `workflow.tasks[0].answers[1].label` becomes `workflowTranslation.strings['tasks.0.answers.1.label']`.

rogerhutchings:
We could use https://lodash.com/docs/4.17.11#set to convert those into a nested object?

eatyourgreens:
PFE converts to a nested object and I'd like to move away from that approach because it's horribly slow. Here's the PFE code.
https://github.com/zooniverse/Panoptes-Front-End/blob/8bdfc43f5c6cdf3832e7a1bfbe824df3434382a9/app/redux/ducks/translations.js#L15-L25

PR #997, opened 2019-07-08
Initialise ui.mode with a valid value

eatyourgreens:
Setting ui.mode to an empty string creates an invalid snapshot for the UI model.

Package:
app-project

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/pages/_app.js
+++ b/packages/app-project/pages/_app.js
@@ -38,7 +38,7 @@ export default class MyApp extends App {
 
     if (pageProps.isServer) {
       // cookie is in the next.js context req object
-      const mode = getCookie(context, 'mode') || ''
+      const mode = getCookie(context, 'mode') || undefined
       const store = initStore(pageProps.isServer, {
         ui: {
           mode


eatyourgreens:
Screenshot of the project app on localhost throwing an error about an invalid snapshot.

This PR should fix this error, which occurs when you run the project app without the mode cookie set.

packages/app-project/pages/_app.js
```diff
@@ -38,7 +38,7 @@ export default class MyApp extends App {
 
     if (pageProps.isServer) {
       // cookie is in the next.js context req object
-      const mode = getCookie(context, 'mode') || ''
+      const mode = getCookie(context, 'mode') || 'light'
```
rogerhutchings: This is essentially setting a default in two places. Instead, we could use:

```
const mode = getCookie(context, 'mode') || undefined
```

`getCookie(context, 'mode')` will return an empty string if not set, which will break the model; `undefined` won't.

eatyourgreens: That's fixed now.

PR #1003, opened 2019-07-09
[MST Upgrade] Update Stores

srallen:
Package: lib-classifier

Describe your changes:
Bug fix in https://github.com/mobxjs/mobx-state-tree/issues/1002 released in MST 3.3.0. This means we have to change how we're instantiating child node stores. 

Tests still fail because other changes are still needed and will be updated in separate PRs.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


rogerhutchings:
Travis is currently failing 😞

srallen:
The tests are expected to fail until all of the stores are fixed.

srallen:
This now includes the updated resource and project stores.

srallen:
I updated the name of this PR to make it more clear that I'm merging all of the individual store update PRs into this one.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -1,147 +1,150 @@
...
+    it('should make the next subject in the queue active when calling `advance()`', function () {
+      expect(rootStore.subjects.active.id).to.equal(rootStore.subjects.resources.values().next().value.id)
+      rootStore.subjects.advance()
+      expect(rootStore.subjects.active.id).to.equal(rootStore.subjects.resources.values().next().value.id)
+      expect(rootStore.subjects.resources.get('1')).to.be.undefined()
     })
 
     describe('with less than three subjects in the queue', function () {
-      let populateSpy
+      let rootStore
```
eatyourgreens: I missed this before, but `rootStore` here is shadowing `rootStore` in the outer function, which leads to two references with the same name but different values. Probably a bad idea: https://eslint.org/docs/rules/no-shadow.html

This will be true for all functions that declare `rootStore` inside this block.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -1,147 +1,150 @@
...
+    sinon.spy(store.subjects, 'populateQueue')
+
+    store.projects.setResource(project)
+    store.projects.setActive(project.id)
+    store.workflows.setResource(workflow)
+    store.workflows.setActive(workflow.id)
+    return store
   }
   describe('Actions > advance', function () {
+    let rootStore
```
eatyourgreens: Here's the declaration that's being shadowed by the inner functions.

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: Fixed by #1045.

srallen:
I'm working on another branch to fix some of the store tests that have mysteriously broke. Hope to have it up today.

eatyourgreens:
Travis CI is running on this branch now.

eatyourgreens:
Are all the stores done now? I think this might be ready to merge.

srallen:
Yes! It's now ready for final review and manual testing.

PR #1006, opened 2019-07-09
[MST Upgrade] Update ResourceStore

srallen:
Package: lib-classifier

Describe your changes:
This updates the reference to `safeReference` and fixes the headers to be an optional type with a default. Fixes one broken test, with `only` left in on purpose to show tests are now passing.

Note that setting active to undefined in the reset action is removed. This is because `safeReference` now handles this for us. When the referenced data is removed from the associated map, it'll automatically set the reference to undefined.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/ResourceStore.js
+++ b/packages/lib-classifier/src/store/ResourceStore.js
@@ -6,8 +6,8 @@ import Resource from './Resource'
 const ResourceStore = types
   .model('ResourceStore', {
     active: types.safeReference(Resource),
-    headers: types.optional(types.frozen({
-      // etag: types.string // setting this is causing the store to be set with a MST type and thus defined. Maybe a bug?
+    headers: types.optional(types.model({
+      etag: types.maybe(types.string)
     }), {}),
     resources: types.map(Resource),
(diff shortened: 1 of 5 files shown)


packages/lib-classifier/src/store/ResourceStore.js
```diff
@@ -5,10 +5,10 @@ import Resource from './Resource'
 
 const ResourceStore = types
   .model('ResourceStore', {
-    active: types.maybe(types.reference(Resource)),
-    headers: types.maybe(types.frozen({
+    active: types.safeReference(Resource),
+    headers: types.optional(types.frozen({
       // etag: types.string // setting this is causing the store to be set with a MST type and thus defined. Maybe a bug?
```
rogerhutchings: I don't think this is a bug - with `frozen`, you can pass an object to provide a default snapshot value, or a subtype, but not a model. See https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/types/utility-types/frozen.ts#L70

You _could_ just make `headers` a model...

srallen: Good point. I think I left that comment when I was still confused about the difference between frozen types and model types.

PR #992, opened 2019-07-05
Lint lib-classifier

eatyourgreens:
Package:
lib-classifier

Closes #978.

Describe your changes:
Convert chai expressions to function calls.
Fix all other linter warnings.

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?
- [ ] Is the changelog updated?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?

PR #1007, opened 2019-07-10
Cleanup classifier tests

eatyourgreens:
Package:
lib-classifier

Finishes off #992.

Also fixes the session ID tests so that they test expected behaviour when a stored session exists.

Note for future maintainers of this code: we're replacing boolean tests because `expected spy to be called with x but it was called with y` is a lot easier to debug than `expected false to be true`.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/store/utils/session.spec.js
```diff
@@ -87,7 +87,7 @@ describe('Store utils > sessionUtils', function () {
 
     it('should call generateSessionID', function () {
       sessionUtils.getSessionID()
-      expect(generateSessionIDStub.calledTwice).to.be.true()
+      expect(generateSessionIDStub).to.have.been.calledTwice()
```
eatyourgreens: Why is the session ID generated twice when you call `getSessionID()`?

eatyourgreens: Looking through [`session.js`](https://github.com/zooniverse/front-end-monorepo/blob/ab843296fb2b42fc8fd42024455c85e469d29383/packages/lib-classifier/src/store/utils/session.js), the code should only call `generateSessionID` once.

eatyourgreens: Fixed in 6814400.

packages/lib-classifier/src/store/utils/session.spec.js
```diff
@@ -69,25 +69,21 @@ describe('Store utils > sessionUtils', function () {
...
-      sessionStorage.removeItem('session_id')
-    })
+      after(function () {
+        sessionStorage.removeItem('session_id')
+      })
 
-    it('should call generateSessionID', function () {
-      sessionUtils.getSessionID()
-      expect(generateSessionIDStub.calledTwice).to.be.true()
+      it('should call generateSessionID', function () {
```
eatyourgreens:
```suggestion
      it('should generate a new session ID', function () {
```
Describe behaviour in BDD tests.

PR #1016, opened 2019-07-11
Hook up Researcher Message widget

rogerhutchings:
Adds the message from researcher widget.

Screenshot 2019-07-11 11 40 38

- shows message from named `owner`
- shows message from project
- shows a fallback message if no message is set

@beckyrother would a border around the avatar look a bit rubbish? I'm thinking for instances like this one where the avatar background matches the widget background

Also:

- the design implies the researcher name is a link. If so, is it supposed to link to the user profile, and is there any benefit to that given the current user profile page (I have _thoughts_ on that page)


eatyourgreens:
Looks good but this should be getting the message from the project translation. Here’s an example from Galaxy Zoo. https://www.zooniverse.org/projects/zookeeper/galaxy-zoo?language=zh-tw

packages/app-project/src/screens/ProjectHomePage/components/MessageFromResearcher/MessageFromResearcher.js
```diff
@@ -0,0 +1,71 @@
...
+
+counterpart.registerTranslations('en', en)
+
+const components = {
+  p: (nodeProps) => (
+    <Paragraph
+      children={nodeProps.children}
+      margin={{ top: 'none' }}
+      size='xlarge'
+    />
```
eatyourgreens: Out of interest, why isn't this written as:
```
<Paragraph>
  {nodeProps.children}
</Paragraph>
```

packages/app-project/src/screens/ProjectHomePage/components/MessageFromResearcher/locales/en.json
```diff
@@ -1,5 +1,6 @@
 {
   "MessageFromResearcher": {
+    "noMessage": "The researcher hasn't added a message.",
```
eatyourgreens: Isn't the default behaviour to show nothing if there's no message?

packages/app-project/src/screens/ProjectHomePage/components/MessageFromResearcher/MessageFromResearcher.js
rogerhutchings: No special reason, other than if it turns into a list I think this way is easier to read

packages/app-project/src/screens/ProjectHomePage/components/MessageFromResearcher/locales/en.json
rogerhutchings: I don't mind either way, but I can see the masonry-style design looking lopsided if we start taking out blocks. One for @beckyrother?

eatyourgreens: If we're always going to display it, then it should probably be added to the beta checklist too, so that projects don't go live with the default message.

beckyrother: Yes, it should always be there, so should be added to the beta checklist. 

If they've not added a message yet, let's say something like "Connect with the research team on Talk to learn more about this project." and take out the researcher name.

beckyrother:
Personally I kind of like the white space if the researcher's avatar is also white, but open to hearing suggestions if anyone feels strongly the other way

rogerhutchings:
No strong feelings, just thought I'd mention it :)

PR #499, opened 2019-02-22
Add API store to classifier

rogerhutchings:
Package: lib-classifier

Adds a store for env settings in the classifier

PR #1017, opened 2019-07-11
Zooniverse Talk widget

rogerhutchings:
Adds the Zooniverse talk widget. ~~Currently WIP while we decide the criteria for the subject images~~ see https://github.com/zooniverse/front-end-monorepo/pull/1017#issuecomment-511862805


eatyourgreens:
Is this a copy of the current Talk widget? If so, just get the three most recent subjects from Talk.
https://github.com/zooniverse/Panoptes-Front-End/blob/9a7138aa84814ff48c455766be3c7cae8f8ea6e5/app/pages/project/home/index.jsx#L58-L60

rogerhutchings:
Yup, that's what I'm doing after having spoken to @beckyrother

rogerhutchings:
Okay, this is ready for review. At the moment, it only displays a box with the subject ID - I want this to use our subject viewer components from the classifier, which still need some work, so that'll come in a future PR.

beckyrother:
Ooo I like that loader! Can the empty message be more directive, like "It looks like no one has commented on a subject yet. Go to Talk to be the first" – something like that?

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/JoinInButton/JoinInButton.js
```diff
@@ -0,0 +1,26 @@
...
+const StyledButton = styled(Button)`
+  text-align: center;
+`
+
+function JoinInButton (props) {
+  const { href } = props
+  const label = counterpart('ZooniverseTalk.button')
+  return (
+    <Link href={href} passHref>
+      <StyledButton label={label} />
```
srallen: Grommet's `Button` has an `href` prop that will render an anchor tag styled like a button. I think that's what the intention is here, so maybe you can set the `href` directly rather than wrap it with a `Link`?

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/components/MessageBox/MessageBox.js
```diff
@@ -0,0 +1,19 @@
+import { Box } from 'grommet'
+import React from 'react'
+
+function MessageBox (props) {
+  return (
+    <Box
+      align='center'
+      animation={['fadeIn', 'slideUp']}
+      background='light-1'
```
srallen: Not sure if dark theme is in scope of the initial PR, but this should have a dark theme variant. Doesn't look great at the moment with dark theme: 

Screen Shot 2019-07-29 at 11 58 34 AM

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/helpers/fetchRecentSubjects/fetchTalkData.js
```diff
@@ -0,0 +1,13 @@
+import talkClient from 'panoptes-client/lib/talk-client'
+
+async function fetchTalkData (projectId) {
+  return talkClient.type('comments')
```
srallen: I think you can use panoptes.js instead even for other hosts, so we don't have to rely on the old Talk client:

```
const query =  {
      focus_type: 'Subject',
      page_size: 10,
      section: `project-${projectId}`,
      sort: '-created_at'
  }

return panoptes.get('/comment', query, {}, 'https://talk.zooniverse.org')
```
the empty object parameter is for any headers you might need to set, like authorization.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/helpers/fetchRecentSubjects/getSubjectIdsFromTalkData.js
```diff
@@ -0,0 +1,14 @@
+import flow from 'lodash/fp/flow'
+import map from 'lodash/fp/map'
+import slice from 'lodash/fp/slice'
+import uniq from 'lodash/uniq'
+
+function getSubjectIdsFromTalkData (comments) {
+  return flow(
+    map(comment => comment.focus_id),
+    uniq,
+    slice(0, 3)
```
srallen: I'm fine with lodash being used for helper functions that don't have a native equivalent, but if we could use the native Array `map` and `slice`, I'd prefer it.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/JoinInButton/JoinInButton.js
rogerhutchings: In this case we can, as we're directing the user towards Talk. Other anchors will require `Link` to enable client side routing - https://github.com/zeit/next.js/#dynamic-routing

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/helpers/fetchRecentSubjects/getSubjectIdsFromTalkData.js
rogerhutchings: These aren't native Array methods, they're the FP versions. It allows them to be composed together to perform multiple transforms without method chaining / lots of clunky temporary variables.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/helpers/fetchRecentSubjects/fetchTalkData.js
rogerhutchings: Yup. The host value can be configured as an env var or kept in a store (or the former used to populate the latter). A similar store was rejected in #499. Do you have a preference?

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/components/MessageBox/MessageBox.js
rogerhutchings: Screenshot 2019-07-30 12 20 15

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/helpers/fetchRecentSubjects/fetchTalkData.js
srallen: I _think_ this is the only place in the new project app that'll be making a request to the Talk API. I would prefer the least bit of code that doesn't rely on the old javascript client, so env vars and using the panoptes.js get request? If we find that we'll be making more requests to Talk in the future, then we can decide on how we'll handle it (maybe a talk.js client?).

rogerhutchings: Sure. I'm going to argue against baking in more config again if we get to that point, but that's a discussion for another day :)

rogerhutchings: Done

rogerhutchings:
This should be good for another pass @srallen

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/helpers/fetchRecentSubjects/fetchTalkData.js
```diff
@@ -1,13 +1,15 @@
...
-      sort: '-created_at'
-    })
+  const query = {
+    focus_type: 'Subject',
+    page_size: 10,
+    section: `project-${projectId}`,
+    sort: '-created_at'
+  }
+
+  return panoptes.get('/comments', query, {}, process.env.TALK_HOST)
```
srallen: Non-blocking: I didn't think of this when suggesting it, but this is not respecting the query param env setting. I don't think it's important for now, but just something to keep in mind.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/ZooniverseTalk.js
```diff
@@ -0,0 +1,53 @@
...
+import RecentSubjects from './components/RecentSubjects'
+
+counterpart.registerTranslations('en', en)
+
+function ZooniverseTalk (props) {
+  const showSubjects = props.screenSize !== 'small'
+  const width = showSubjects ? '18%' : 'full'
+
+  return (
+    <ContentBox direction='row' gap='large'>
```
srallen: The flex direction here needs to be responsive. Here how this looks now:

Screen Shot 2019-07-31 at 10 45 04 AM

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsContainer.js
```diff
@@ -0,0 +1,76 @@
...
+
+    if (show) {
+      if (loading === asyncStates.success) {
+        result = (subjects.length < 1)
+          ? (<MessageBox children={counterpart('RecentSubjects.noSubjects')} />)
+          : (<RecentSubjects subjects={subjects} />)
+      } else if (loading === asyncStates.error) {
+        result = (<MessageBox children={counterpart('RecentSubjects.error')} />)
+      } else {
+        result = (<Placeholder />)
```
srallen: There's a flash of the placeholders and then disappear on small mobile screen sizes I think because the API request is still happening.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsContainer.js
```diff
@@ -0,0 +1,76 @@
...
+    super()
+    this.state = {
+      loading: asyncStates.initialized,
+      subjects: []
+    }
+  }
+
+  componentDidMount () {
+    this.setState({ loading: asyncStates.loading })
+    fetchRecentSubjects(this.props.projectId)
```
srallen: If the subjects are hidden on small screen sizes, then we shouldn't be making an API request for them. Could you check for this before making the request?

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/ZooniverseTalk.js
srallen: @beckyrother here's what this looks like with the flex direction rotated to column on the mobile size:

Screen Shot 2019-07-31 at 11 27 16 AM

beckyrother: Perfect! ty

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsContainer.js
rogerhutchings: The API request happens, but I think it's actually because the default `screenSize` provided by the `ResponsiveContext` component is `medium`

srallen:
@eatyourgreens any chance you'd be available to finish this up? I don't think there's much left to fix.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/ZooniverseTalk.js
srallen: The `screenSize` is injected into the props from the `withResponseContext` HOC, so that can be used to set the flex direction depending on the size.

eatyourgreens:
No problem.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/ZooniverseTalk.js
eatyourgreens: I've updated this and added a spec.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsContainer.js
eatyourgreens: We don't hide subjects on phones, we just show them at a smaller size. I've changed the code to reflect how the site works at the moment.

srallen: Check in with @beckyrother, but I don't think she wants to match the current site, but wants to hide them on mobile instead now with the new design?

eatyourgreens: Ok. There's nothing in this PR to indicate how the widget should look, so I assumed it would be the same as the current site.

eatyourgreens:
I've added thumbnail images for the subjects. Here's what it looks like for Galaxy Zoo. Alt text for each image is `subject 12345` where 12345 is the Panoptes subject ID.

Wide viewport:
Screenshot showing image previews displayed to the right of the Talk box.


Narrow viewport:
Screenshot showing subject thumbnails displayed beneath the Talk box.

beckyrother:
Is it possible to show the previous 4 subjects rather than 3 so that 'Join In' button isn't quite so awkwardly long?

eatyourgreens:
Sure, or those three images could be bigger, I guess. I chose the max height/width at random.

beckyrother:
@eatyourgreens I've sent you a link to the InVision for this page, so you should be able to find the exact dimensions from there. Sorry! I'd thought you had access to that page already. 
Here's a link to the Inspect mode where you'll be able to find all that info: https://projects.invisionapp.com/d/main?origin=v7#/console/12924056/344288106/inspect

eatyourgreens:
@beckyrother How's this?
Screenshot showing the widescreen view with adjusted sizing and larger subject thumbnails.

eatyourgreens:
The InVision link comes up blank for me, with a bunch of errors in the console. I'm wrapping up for the day but I can take another look tomorrow morning.

beckyrother:
@eatyourgreens Looks perfect! Thanks for updating. I think InVision is having some troubles today, I got an error earlier when I tried to look at a page ¯\_(ツ)_/¯

eatyourgreens:
@srallen this should be ready for another look now.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/ZooniverseTalk.js
eatyourgreens: I switched this to use a responsive grid, instead of flexbox. There's now a grid that has one column on small screens, two columns on larger displays.

eatyourgreens:
Here's what it looks like with text labels.
Screenshot showing Talk subjects with text labels overlaid on each subject.

eatyourgreens:
The design for mobile has an image carousel, which isn't implemented here.
Mobile design for the Talk widget, showing recent images in a carousel.

eatyourgreens:
Grommet has [a carousel component](https://v2.grommet.io/carousel), which could be used for the mobile view. I'm experimenting with it at the moment, it looks good but it doesn't support links on child images.

eatyourgreens:
I've opened #1097 for the carousel on small screens, because I'm not completely sure that the Grommet carousel does exactly what we want.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsContainer.js
```diff
@@ -0,0 +1,78 @@
...
+  render () {
+    const { carousel } = this.props
+    const { loading, subjects } = this.state
+    const href=`/projects/${this.props.slug}/talk`
+    let result = null
+    const ThumbnailComponent = carousel ? RecentSubjectsCarousel : RecentSubjects
+
+    if (loading === asyncStates.success) {
+      result = (subjects.length < 1)
+        ? (<MessageBox children={counterpart('RecentSubjects.noSubjects')} />)
```
srallen: The children message here should probably be wrapped by `Paragraph` or something to give it some padding. It looks a bit odd on smaller sizes without it:

iPad Pro size:
Screen Shot 2019-08-30 at 12 07 45 PM

Mobile:
Screen Shot 2019-08-30 at 12 07 52 PM

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsContainer.js
```diff
@@ -0,0 +1,78 @@
...
+        : (<ThumbnailComponent href={href} subjects={subjects} />)
+    } else if (loading === asyncStates.error) {
+      result = (<MessageBox children={counterpart('RecentSubjects.error')} />)
```
srallen: Same comment about the children here.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/components/Placeholder/Placeholder.js
```diff
@@ -0,0 +1,47 @@
...
+    : theme.global.colors['light-4']
+
+  return (
+    <Grid
+      columns={['1fr', '1fr', '1fr']}
+      fill
+      gap='small'
+      {...props}
+    >
+      {[null, null, null].map((element, index) => (
```
srallen: I didn't catch this on review to add the carousel, but perhaps on small sizes, this should only show a single loading placeholder, so it has the correct single column layout. Right now on mobile sizes this is what this looks like (took the screenshot while throttling):

Screen Shot 2019-08-30 at 12 12 52 PM

eatyourgreens:
>The TESS subject's aspect ratio is a bit awkward in the design and it becomes much more noticeable on smaller sizes, but before switching to the single column:

I've been wondering if responsibility for cropping the image should be handed off to the Media component, since the file here could be audio, video or an image (or maybe even SVG data.)

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/components/Placeholder/Placeholder.js
eatyourgreens: Yeah, the placeholder should be moved down the tree. Doesn't the Media component already have a placeholder?

eatyourgreens:
>The app appears to be reloading the store when navigating between the classify and home pages so even if I switch between dark and light theme, my selection gets reset. (My cookie state happened to have a mode stored into it twice and dark mode was in it first so that was loading first. The duplication might be a remnant of working on dark mode before and I hadn't cleared that out)

Dark mode persists for me across the home page, the Classify page and the About page on master.

eatyourgreens:
#1101 adds the theme toggle to the project home page.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsContainer.js
eatyourgreens: Fixed.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsContainer.js
eatyourgreens: Fixed.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/components/Placeholder/Placeholder.js
eatyourgreens: I've removed the placeholder.

eatyourgreens:
Screenshot of the TESS Talk subjects, showing the images cropped to a narrow slice of the full light curve.

I've set the thumbnail Box to fill its container, which I think looks better for those wide, short images but loses a lot of the light curve.

eatyourgreens:
The previous screenshot shows that the verified project tick icon is broken on small screen sizes, though.

eatyourgreens:
Tall images are enlarged to fill the available width, which can lead to them looking pixelated. This could be fixed by making the thumbnail dimensions bigger.
Screenshot of a Notes from Nature project, showing the subject thumbnails enlarged to fill the available width, pixelated and cropped to about half their total height.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/components/SubjectThumbnail/SubjectThumbnail.js
```diff
@@ -0,0 +1,62 @@
...
+
+function SubjectThumbnail ({ height, href, width, subject }) {
+  const subjectURLs = subject.locations.map(location => Object.values(location)[0])
+  const subjectURL = subjectURLs[0]
+  return (
+    <Anchor
+      href={`${href}/subjects/${subject.id}`}
+    >
+      <StyledBox
+        background='brand'
```
srallen: Let's remove the background color here... the grid is sizing this div to be slightly larger than the child image and we get the background color showing:

Screen Shot 2019-09-03 at 10 53 02 AM

eatyourgreens: Weird, the image should be filling the full width, like the screenshot from the NfN project above.

srallen: Depends on what the column fractions of the grid calculate to by the browser and the parent div with the background color is just width 100%. I think my laptop window might be at a higher res?

eatyourgreens: I can reproduce this at high resolution when the viewport width is ~1400px or more.
image

It's a bug in the Media component, which styles the image with a max height and width so won't fill its parent if the parent dimensions are larger than the height or width props that are passed in. I'll adjust the height and width props, otherwise that the spacing between the images will be wrong at high resolution (with the ugly teal background removed.)

eatyourgreens:
I've removed the teal background and adjusted the thumbnail size props so that the image continues to grow to fill the available space at larger viewport widths.

http://localhost.zooniverse.org:3000/projects/nora-dot-eisner/planet-hunters-tess (landscape images)
Screenshot of Planet Hunters TESS showing wide landscape thumbnail images.


http://localhost.zooniverse.org:3000/projects/zookeeper/galaxy-zoo (square images)
Screenshot of Galaxy Zoo showing large square thumbnails.


http://localhost.zooniverse.org:3000/projects/md68135/notes-from-nature-capturing-californias-flowers (portrait images)
Screenshot of Notes from Nature showing cropped square thumbnails.

I can open an issue for the Media component if the max height/width is a bug.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsContainer.js
eatyourgreens: I've removed the placeholder component, now that the carousel is implemented for small screens.

packages/app-project/README.md
```diff
@@ -15,12 +15,12 @@ This package should be cloned as part of the [front-end-monorepo](https://github
 
 Starts a development server on port 3000 and a Storybook server on port 9001 by default.
 
-####Docker
+#### Docker
```
eatyourgreens: The README is also fixed in #1088.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/helpers/fetchRecentSubjects/fetchTalkData.js
eatyourgreens: I think you can run `PANOPTES_ENV=production yarn dev` to develop against the production API.

PR #1020, opened 2019-07-11
[WIP] Add drawing tool point component and refactor interaction layer container

mcbouslog:
Package: lib-classifier

I'll be editing this initial comment shortly, opening to discuss some specific pieces.
This whole branch may get replaced after a few other drawing PRs merge, as this includes many files/changes from those PRs.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -16,24 +28,92 @@ function storeMapper (stores) {
...
+      id: Math.random(),
+      toolType: drawingTask.tools[activeDrawingTool].type
+    }, event)
+
+    return newMark
   }
 
   addToStream (event, type) {
-    this.props.drawing.addToStream({
+    const svgEventOffset = this.getEventOffset(event.clientX, event.clientY)
```
mcbouslog: Is it better to convert the event coords from client to svg before adding to the RxJs eventStream (current implementation) or when consuming?

mcbouslog:
PFE has a root drawing tool - https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/drawing-tools/root.jsx, I think we'll want as well.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -16,24 +28,92 @@ function storeMapper (stores) {
...
+    // const stream = this.props.eventStream
+    // stream.subscribe(z => console.log('z = ', z))
+    this.drawMark()
+  }
+
+  componentDidUpdate () {
+    this.drawMark()
+  }
+
+  drawMark () {
```
mcbouslog: This needs refactor, maybe should be `switch`.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -16,24 +28,92 @@ function storeMapper (stores) {
...
+
+    return newMark
   }
 
   addToStream (event, type) {
-    this.props.drawing.addToStream({
+    const svgEventOffset = this.getEventOffset(event.clientX, event.clientY)
+
+    this.props.addToStream({
       event: type,
```
srallen: Is this what is causing the `event.event` above? Perhaps this can be `type: type` or shorthand `type` instead? That way it'll be `event.type` in other places in the code?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -16,24 +28,92 @@ function storeMapper (stores) {
...
+
+    return svgEventOffset
+  }
+
+  getNewMark (event) {
+    const { activeDrawingTool, activeStepTasks } = this.props
+
+    const [drawingTask] = activeStepTasks.filter(task => task.type === 'drawing')
+    const newMark = Object.assign({
+      id: Math.random(),
```
srallen: The uuid package is already a dependency because I used it for a temporary unique id for classifications. Perhaps that can be used here instead?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -16,24 +28,92 @@ function storeMapper (stores) {
...
+      if (drawing && event.event === 'mouseup') {
+        this.setState({ drawing: false })
+      }
+      // subsequent mark
+      if (!drawing && currentMark !== null && event.event === 'mousedown') {
+        const filteredMarkings = markings.filter(marking => marking.id !== currentMark.id)
+        filteredMarkings.push(currentMark)
+        newMark = this.getNewMark(event)
+        this.setState({ currentMark: newMark, drawing: true, markings: filteredMarkings })
+      }
```
srallen: I think each of the above could be their own class methods, `onMouseUp`, `onMouseDown` etc. If you want to avoid a switch statement, you could define an object that's set to call the event handlers:
```
const handlers = {
  mousedown: () => this.onMouseDown()
}
```

and call them as so:
```
handlers[event.type]()
```

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
srallen: I'm not sure what would be best here. I'm going to do a bit of reading and get back to this question.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -16,24 +28,92 @@ function storeMapper (stores) {
 class InteractionLayerContainer extends Component {
   constructor () {
     super()
+
+    this.state = {
+      currentMark: null,
+      drawing: false,
+      markings: []
```
srallen: Maybe instead of an array we can use a map? That way you can leverage the ids of each mark to get/set as long as they're unique ids?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
srallen: You want want to pass the original event object in as an argument in my example depending on if it is needed by the handler or any other data.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
eatyourgreens: Why `markings` and not `marks`?

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DrawingTask/DrawingTask.js
```diff
@@ -1,56 +1,100 @@
...
-    // if (annotations && annotations.size > 0) {
-    //   annotation = annotations.get(task.taskKey)
-    // }
     return (
-      <Text size='small' tag='legend'>
-        <Markdownz>
-          {task.instruction}
-        </Markdownz>
-      </Text>
+      <StyledFieldset>
```
eatyourgreens: I don't think Fieldset is the right element to use here. Fieldsets have simple, inline text legends but our task questions are block-level elements, sometimes with more than one paragraph.

I think this should just be a Grommet Box.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/DrawingTask/DrawingTask.js
```diff
@@ -1,56 +1,100 @@
...
+        <StyledText size='small' tag='legend'>
```
eatyourgreens: Ditto here about fieldset/legend. Legend can't take block-level children, so can't wrap a Markdownz component.

eatyourgreens: Here's the list of what's valid in a fieldset legend. I don't think we can easily switch to using these elements for task questions without breaking existing live workflows.
https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content

srallen: This was my bad when first starting on the task area and I think it's being copied around. I was thinking of using valid form elements if in theory we could have a PWA that could work without javascript, but since there's markdown going here, we won't be able to guarantee the question is valid HTML for these elements.

eatyourgreens: We could possibly mark up transcription tasks as forms that submit the values of the text inputs, or question tasks as forms that submit the answers. Drawing tasks would submit the value of the selected drawing tool, so I'm thinking either we move the form markup down to specific types of task, or drop it completely.

mcbouslog: My mistake in opening this PR in a sloppy manner, but can I copy and paste this convo to #913 ? I'll edit this PR so it only reflects changes related to the InteractionLayerContainer and Point tool component.

mcbouslog: Moved to #913.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
srallen: After thinking about this, I think we want the stream to retain the original event in case we want to cancel it to prevent bubbling, so to that end, the conversion should happen when consuming.

mcbouslog:
Closing in favor of new PR, will duplicate open comments

PR #1022, opened 2019-07-11
[MST Upgrade] Update UPP store

srallen:
Package: lib-classifier

Describe your changes:
Updates to use `types.safeReference`. Updates tests. `only` left in on purpose to show UPP store tets are passing.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/UserProjectPreferencesStore.spec.js
+++ b/packages/lib-classifier/src/store/UserProjectPreferencesStore.spec.js
@@ -12,7 +12,6 @@ import {
 } from '../../test/factories'
 import stubPanoptesJs from '../../test/stubPanoptesJs'
 
-let rootStore
 const project = ProjectFactory.build()
 const upp = UPPFactory.build()
 const user = UserFactory.build()
@@ -40,7 +39,7 @@ const authClientStubWithUser = {
 }
 
 function setupStores (clientStub, authClientStub) {
-  rootStore = RootStore.create({
+  return RootStore.create({
     classifications: {},
     dataVisAnnotating: {},
     drawing: {},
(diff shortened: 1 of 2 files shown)


packages/lib-classifier/src/store/UserProjectPreferencesStore.spec.js
```diff
@@ -40,39 +39,49 @@ const authClientStubWithUser = {
   checkBearerToken: sinon.stub().callsFake(() => Promise.resolve(token))
 }
 
-describe('Model > UserProjectPreferencesStore', function () {
+function setupStores (clientStub, authClientStub) {
+  rootStore = RootStore.create({
```
eatyourgreens: Should this function return a new root store? Functions that modify global variables can be confusing.

packages/lib-classifier/src/store/UserProjectPreferencesStore.spec.js
```diff
@@ -40,39 +39,49 @@ const authClientStubWithUser = {
...
     expect(rootStore.userProjectPreferences.loadingState).to.equal(asyncStates.initialized)
+    rootStore = null
   })
 
   it('should set project preferences if there is a user and a project', function (done) {
-    rootStore = RootStore.create({
-      projects: ProjectStore.create(),
-      userProjectPreferences: UserProjectPreferencesStore.create()
-    }, { authClient: authClientStubWithUser, client: clientStub })
+    setupStores(clientStub, authClientStubWithUser)
```
eatyourgreens: I think this would be clearer as `rootStore = setupStores(clientStub, authClientStubWithUser)` or ` const rootStore = setupStores(clientStub, authClientStubWithUser)` to keep the scope clear.

PR #1024, opened 2019-07-12
[MST Upgrade] Update TutorialStore

srallen:
Package: lib-classifier

Describe your changes:
Updates store to use `types.safeReference`. Refactors tests to instantiate stores as needed for the tests. Purposely leaves in `only` to show `TutorialStore` tests are passing.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/TutorialStore.spec.js
+++ b/packages/lib-classifier/src/store/TutorialStore.spec.js
@@ -81,7 +81,7 @@ const authClientStubWithUser = {
   checkBearerToken: sinon.stub().callsFake(() => Promise.resolve(token))
 }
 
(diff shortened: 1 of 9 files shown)


packages/lib-classifier/src/store/TutorialStore.spec.js
```diff
@@ -91,24 +91,36 @@ describe('Model > TutorialStore', function () {
       })
   }
 
+  function setupStores(clientStub, authClientStub) {
+    rootStore = null
```
eatyourgreens: Same comment here as on the UPP tests. Functions should avoid modifying variables outside their own scope. Better to to return the new root store from this function.

PR #1026, opened 2019-07-16
Add docker-compose

eatyourgreens:
Set up docker-compose to run the Next apps in development.

`docker-compose build` to install dependencies and bootstrap.
`docker-compose up` should run the project app locally on port 3000.
`docker-compose down` to stop the server after interrupting it with ^C.
`docker-compose run --rm dev test` to run the tests.
`docker-compose run --rm dev test:ci` to run the CI tests.
`docker-compose run --rm --service-ports dev workspace @zooniverse/classifier dev` to override the default app and run the classifier.
`docker-compose run --rm --service-ports dev workspace @zooniverse/react-components start` to run the React components storybook.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
`yarn test` is failing for me locally with
`@zooniverse/related-projects-api: Error: No test files found: "./src/**/*.spec.js"`.

eatyourgreens:
This sets up docket to run `yarn`, with the project app as the default arguments. You can override those by passing new arguments to `docker-compose run --rm dev`, which will then be passed to `yarn` in the container.

rogerhutchings:
This needs adding to the README

eatyourgreens:
Consider adding config files to each package, so that `docker compose up` runs that package.

Dockerfile.dev
```diff
@@ -0,0 +1,16 @@
...
+COPY ./ .
+RUN chown -R node:node .
+
+USER node
+
+RUN yarn bootstrap
+
+EXPOSE 3000 6006 8080
+ENTRYPOINT ["yarn"]
+CMD ["workspace",  "@zooniverse/fe-project", "dev"]
```
adammcmaster: You can specify these last three lines in `docker-compose.yml` so it might be better to do that and avoid having a second Dockerfile.

eatyourgreens: That sounds useful in terms of having a different entry point for each package too. Is the syntax the same in the YAML file?

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+      - "8080:8080"
+    environment:
+      - HOST=0.0.0.0
+      - AWS_REGION=${AWS_REGION}
+      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
+      - AWS_SESSION_TOKEN=${AWS_SESSION_TOKEN}
+      - AWS_SECURITY_TOKEN=${AWS_SECURITY_TOKEN}
+    volumes:
+      - ${PWD}:/src
```
adammcmaster: This should mount to `/usr/src`

Dockerfile.dev
adammcmaster: It passes the whole command as a string rather than a list, but here's an example: https://github.com/zooniverse/hamlet/blob/master/docker-compose.yml#L37

eatyourgreens: Does that allow you to override the default args like ENTRYPOINT and CMD do? The string versions of those don't allow that in a Dockerfile.

adammcmaster: Yup I think they do.

eatyourgreens: Docs say lists are supported so I might use those anyway.
https://docs.docker.com/compose/compose-file/#command
https://docs.docker.com/compose/compose-file/#entrypoint

All the documentation I read yesterday said to prefer arrays but didn't really explain why.

eatyourgreens:
@adammcmaster thanks for the comments. I've updated this to use one Dockerfile and get the entrypoint from docker-compose.yml. I've also change `/src` to `/usr/src`.

eatyourgreens:
TODO: add `docker-compose.yml` for each package that we might run locally.

PR #1030, opened 2019-07-17
[MST Upgrade] Update SubjectStore

srallen:
Package: lib-classifier

Describe your changes:
This updates the SubjectStore and SubjectResource model as well as fix a few bugs:

For `ResourceStore`:
- Added check to make sure the response body has length before storing it

For `SubjectResource`:
- Added `isValidReference` existential checks for project and workflows
- Updated test and added tests for the viewer view getter
- ~Removed cruft describe blocks~

For `SubjectStore`:
- `types.reference` is now `types.safeReference`
- Removed unnecessary `types.optional` around the `types.map`
- Added `isValidReference` for the existential checks for the store's `active` property that is a reference
- Removed the actions set to `undefined` in the returned object for the actions. MST will just simply throw an undefined type error, so don't do this.
- A refactor to deal with bugs with the subject advance middleware, an explanation:

The original middleware:
```
    function onSubjectAdvance (call, next, abort) {
      const root = getRoot(self)
      const subject = self.active
      const shouldShowFeedback = root.feedback.isActive && root.feedback.messages.length && !root.feedback.showModal
      if (!shouldShowFeedback && subject && subject.shouldDiscuss) {
        const { url, newTab } = subject.shouldDiscuss
        openTalkPage(url, newTab)
      }
      next(call)
    }
```

now should have an existential check of the active reference otherwise MST will error on the first time this is called, so changed to:

```
    function onSubjectAdvance (call, next, abort) {
      const root = getRoot(self)
      const validSubjectReference = isValidReference(() => self.active)
      if (validSubjectReference) {
        const subject = self.active
        const shouldShowFeedback = root.feedback.isActive && root.feedback.messages.length && !root.feedback.showModal
        if (!shouldShowFeedback && subject && subject.shouldDiscuss) {
          const { url, newTab } = subject.shouldDiscuss
          openTalkPage(url, newTab)
        }
        next(call)
      }
    }
```

But, when `advance` is first called on the initial queue load and thus this middleware, there isn't a valid active reference yet by this point because the middleware intercepts before we let the `advance` action remove and set the new active subject. We have to intercept it first because we want to potentially browse to talk before the next subject is made active. Because there isn't an active reference yet, neither `next` or `abort` is called and MST will error.

So, to resolve this, you could add:

```
    function onSubjectAdvance (call, next, abort) {
      const root = getRoot(self)
      const validSubjectReference = isValidReference(() => self.active)
      if (validSubjectReference) {
        const subject = self.active
        const shouldShowFeedback = root.feedback.isActive && root.feedback.messages.length && !root.feedback.showModal
        if (!shouldShowFeedback && subject && subject.shouldDiscuss) {
          const { url, newTab } = subject.shouldDiscuss
          openTalkPage(url, newTab)
        }
        next(call)
      } else {
        next(call)
      }
    }
```

But in test, this caused an infinite loop: `populateQueue` called `advance` when there was no active reference, then interrupted by middleware, which calls `advance` when there is no active reference, which calls `populateQueue` when there's less than 3 subjects because this is called before the active reference is set, then `populateQueue` calls `advance`, etc.

To resolve this, I decided to remove the direct call to advance the queue from the `populateQueue` call and instead added an observer for the queue itself. The queue will only be called to be advanced if we have successfully populated the queue, we have a queue, and there is no active reference.

We might want to also consider removing the call to `populateQueue` from `advance` and instead have the subject store observe itself in this case as well, but this is working now.

Left in `only` for the relevant updated tests to show they're passing.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/ResourceStore.js
+++ b/packages/lib-classifier/src/store/ResourceStore.js
@@ -27,6 +27,7 @@ const ResourceStore = types
           self.loadingState = asyncStates.success
           return resource
         } else {
+          self.loadingState = asyncStates.success
           return undefined
         }
       } catch (error) {
--- a/packages/lib-classifier/src/store/ResourceStore.js
+++ b/packages/lib-classifier/src/store/ResourceStore.js
@@ -22,9 +22,13 @@ const ResourceStore = types
       try {
         const response = yield client.get(`/${type}/${id}`)
         self.headers = response.headers
-        const resource = response.body[type][0]
-        self.loadingState = asyncStates.success
-        return resource
+        if (response.body[type] && response.body[type].length > 0) {
+          const resource = response.body[type][0]
+          self.loadingState = asyncStates.success
+          return resource
+        } else {
+          return undefined
+        }
       } catch (error) {
         console.error(error)
         self.loadingState = asyncStates.error
--- a/packages/lib-classifier/src/store/Subject.spec.js
+++ b/packages/lib-classifier/src/store/Subject.spec.js
@@ -60,11 +60,11 @@ describe.only('Model > Subject', function () {
 
     it('should reutrn the light curve viewer if the workflow configuration is defined', function () {
       const dataSubject = SubjectFactory.build({ location: [{ 'application/json': 'https://foo.bar/data.json' }]})
-      const subjectStore = Subject.create(dataSubject)
-      subjectStore.workflows = WorkflowStore.create({})
-      subjectStore.workflows.setResource(workflowWithConfig)
-      subjectStore.workflows.setActive(workflowWithConfig.id)
-      expect(subjectStore.viewer).to.equal(subjectViewers.lightCurve)
+      const subjectResourceStore = Subject.create(dataSubject)
+      subjectResourceStore.workflows = WorkflowStore.create({})
+      subjectResourceStore.workflows.setResource(workflowWithConfig)
+      subjectResourceStore.workflows.setActive(workflowWithConfig.id)
+      expect(subjectResourceStore.viewer).to.equal(subjectViewers.lightCurve)
     })
   })
 
--- a/packages/lib-classifier/src/store/Subject.spec.js
+++ b/packages/lib-classifier/src/store/Subject.spec.js
@@ -58,7 +58,7 @@ describe.only('Model > Subject', function () {
       expect(subjectStore.viewer).to.equal(subjectViewers.singleImage)
     })
 
-    it('should reutrn the light curve viewer if the workflow configuration is defined', function () {
+    it('should return the light curve viewer if the workflow configuration is defined', function () {
       const dataSubject = SubjectFactory.build({ location: [{ 'application/json': 'https://foo.bar/data.json' }]})
       const subjectResourceStore = Subject.create(dataSubject)
       subjectResourceStore.workflows = WorkflowStore.create({})
--- a/packages/lib-classifier/src/store/Subject.spec.js
+++ b/packages/lib-classifier/src/store/Subject.spec.js
@@ -91,4 +91,35 @@ describe.only('Model > Subject', function () {
       expect(subject.onAddToCollection.withArgs(subject.id)).to.have.been.calledOnce()
     })
   })
+
+  describe('Actions > openInTalk', function () {
+    let url
+
+    before (function () {
+      url = `https://example.org/projects/${project.slug}/talk/subjects/${subject.id}`
+    })
+
+    function testOpenInTalk (newTab) {
+      const subject = Subject.create(stub)
+      subject.projects = ProjectStore.create({})
+      subject.projects.setResource(project)
+      subject.projects.setActive(project.id)
+      subject.openInTalk(newTab)
+      expect(subject.shouldDiscuss).to.eql({ newTab, url })
+    }
+
+    describe('in the same tab', function () {
+      
+      it('should set the shouldDiscuss property', function () {
+        testOpenInTalk(false)
+      })
+    })
+
+    describe('in a new tab', function () {
+
+      it('should set the shouldDiscuss property', function () {
+        testOpenInTalk(true)
+      })
+    })
+  })
 })
--- a/packages/lib-classifier/src/store/Subject.spec.js
+++ b/packages/lib-classifier/src/store/Subject.spec.js
@@ -1,20 +1,23 @@
 import sinon from 'sinon'
 import Subject from './Subject'
-import { ProjectFactory, SubjectFactory } from '../../test/factories'
+import ProjectStore from './ProjectStore'
+import WorkflowStore from './WorkflowStore'
+import { ProjectFactory, SubjectFactory, WorkflowFactory } from '../../test/factories'
+import subjectViewers from '../helpers/subjectViewers'
 
 const stub = SubjectFactory.build()
-const project = ProjectFactory.build()
+const workflow = WorkflowFactory.build()
+const workflowWithConfig = WorkflowFactory.build({ configuration: { subject_viewer: 'lightcurve' } })
+const project = ProjectFactory.build({}, { activeWorkflowId: workflow.id })
 
-describe('Model > Subject', function () {
+describe.only('Model > Subject', function () {
   let subject
 
   before(function () {
     subject = Subject.create(stub)
     subject.onToggleFavourite = sinon.stub()
     subject.onAddToCollection = sinon.stub()
-    subject.projects = {
-      active: project
-    }
+
   })
 
   it('should exist', function () {
@@ -26,11 +29,46 @@ describe('Model > Subject', function () {
     expect(subject.locations).to.deep.equal(stub.locations)
   })
 
-  it('should have a Talk URL', function () {
-    expect(subject.talkURL).to.equal(`https://example.org/projects/zooniverse/example/talk/subjects/${subject.id}`)
+  describe('Views > talkURL', function () {
+    before(function() {
+      subject.projects = ProjectStore.create({})
+      subject.projects.setResource(project)
+      subject.projects.setActive(project.id)
+    })
+
+    it('should have a Talk URL', function () {
+      expect(subject.talkURL).to.equal(`https://example.org/projects/${project.slug}/talk/subjects/${subject.id}`)
+    })
   })
 
-  describe('toggleFavorite', function () {
+  describe('Views > viewer', function () {
+    it('should return null as default', function () {
+      subject.workflows = WorkflowStore.create({})
+      subject.workflows.setResource(workflow)
+      subject.workflows.setActive(workflow.id)
+      expect(subject.viewer).to.be.null()
+    })
+
+    it('should return the single image viewer for subjects with a single image location', function () {
+      const singleImageSubject = SubjectFactory.build({ locations: [{ 'image/png': 'https://foo.bar/example.png' }] })
+      const subjectStore = Subject.create(singleImageSubject)
+      subjectStore.workflows = WorkflowStore.create({})
+      subjectStore.workflows.setResource(workflow)
+      subjectStore.workflows.setActive(workflow.id)
+      expect(subjectStore.viewer).to.equal(subjectViewers.singleImage)
+    })
+
+    it('should reutrn the light curve viewer if the workflow configuration is defined', function () {
+      const dataSubject = SubjectFactory.build({ location: [{ 'application/json': 'https://foo.bar/data.json' }]})
+      const subjectStore = Subject.create(dataSubject)
+      subjectStore.workflows = WorkflowStore.create({})
+      subjectStore.workflows.setResource(workflowWithConfig)
+      subjectStore.workflows.setActive(workflowWithConfig.id)
+      expect(subjectStore.viewer).to.equal(subjectViewers.lightCurve)
+    })
+  })
+
+  describe('Actions > toggleFavorite', function () {
     before(function () {
       subject.toggleFavorite()
     })
@@ -40,60 +78,17 @@ describe('Model > Subject', function () {
     })
 
     it('should call the onToggleFavourite callback', function () {
-      expect(subject.onToggleFavourite).to.have.been.calledOnceWith(subject.id, subject.favorite)
+      expect(subject.onToggleFavourite.withArgs(subject.id, subject.favorite)).to.have.been.calledOnce()
     })
   })
 
-  describe('addToCollection', function () {
+  describe('Actions > addToCollection', function () {
     before(function () {
       subject.addToCollection()
     })
 
     it('should call the onAddToCollection callback', function () {
-      expect(subject.onAddToCollection).to.have.been.calledOnceWith(subject.id)
-    })
-  })
-
-  describe('openInTalk', function () {
-    let feedback
-    let onHide = () => null
-
-    before(function () {
-      feedback = {
-        onHide,
-        setOnHide: sinon.stub().callsFake(callback => { onHide = callback })
-      }
-      subject.feedback = feedback
-    })
-
-    describe('in the same tab', function () {
-      before(function () {
-        subject.openInTalk(false)
-      })
-
-      after(function () {
-        feedback.setOnHide.resetHistory()
-        window.location.assign.resetHistory()
-      })
-    })
-
-    describe('in a new tab', function () {
-      let newTab = {
-        opener: null,
-        location: null,
-        target: null,
-        focus: sinon.stub()
-      }
-
-      before(function () {
-        window.open = sinon.stub().callsFake(() => newTab)
-        subject.openInTalk(true)
-      })
-
-      after(function () {
-        feedback.setOnHide.resetHistory()
-        window.location.assign.resetHistory()
-      })
+      expect(subject.onAddToCollection.withArgs(subject.id)).to.have.been.calledOnce()
     })
   })
 })
--- a/packages/lib-classifier/src/store/SubjectStore.js
+++ b/packages/lib-classifier/src/store/SubjectStore.js
@@ -1,6 +1,6 @@
 import asyncStates from '@zooniverse/async-states'
 import { autorun } from 'mobx'
-import { addDisposer, addMiddleware, flow, getRoot, onPatch, types } from 'mobx-state-tree'
+import { addDisposer, addMiddleware, flow, getRoot, isValidReference, onPatch, types } from 'mobx-state-tree'
 import { getBearerToken } from './utils'
 import { filterByLabel, filters } from '../components/Classifier/components/MetaTools/components/Metadata/components/MetadataModal'
 import ResourceStore from './ResourceStore'
@@ -20,14 +20,15 @@ function openTalkPage (talkURL, newTab = false) {
 
 const SubjectStore = types
   .model('SubjectStore', {
-    active: types.maybe(types.reference(Subject)),
-    resources: types.optional(types.map(Subject), {}),
+    active: types.safeReference(Subject),
+    resources: types.map(Subject),
     type: types.optional(types.string, 'subjects')
   })
 
   .views(self => ({
     get isThereMetadata () {
-      if (self.active) {
+      const validSubjectReference = isValidReference(() => self.active)
+      if (validSubjectReference) {
         const filteredMetadata = Object.keys(self.active.metadata)
           .filter((label) => filterByLabel(label, filters))
         return filteredMetadata.length > 0
@@ -39,35 +40,36 @@ const SubjectStore = types
 
   .actions(self => {
     function advance () {
-      if (self.active) {
+      const validSubjectReference = isValidReference(() => self.active)
+      if (validSubjectReference) {
         const idToRemove = self.active.id
         self.resources.delete(idToRemove)
-        self.active = undefined
       }
+
+      const nextSubject = self.resources.values().next().value
+      self.active = nextSubject && nextSubject.id
 
       if (self.resources.size < 3) {
         console.log('Fetching more subjects')
         self.populateQueue()
       }
-
-      const nextSubject = self.resources.values().next().value
-      self.active = nextSubject && nextSubject.id
     }
 
     function afterAttach () {
       createWorkflowObserver()
       createClassificationObserver()
       createSubjectMiddleware()
+      createQueueLoadObserver()
     }
 
     function createWorkflowObserver () {
       const workflowDisposer = autorun(() => {
-        const root = getRoot(self)
-        if (root.workflows && root.workflows.active) {
+        const validWorkflowReference = isValidReference(() => getRoot(self).workflows.active)
+        if (validWorkflowReference) {
           self.reset()
           self.populateQueue()
         }
-      })
+      }, { name: 'SubjectStore Workflow Observer autorun' })
       addDisposer(self, workflowDisposer)
     }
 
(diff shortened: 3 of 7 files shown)


packages/lib-classifier/src/store/Subject.spec.js
```diff
@@ -40,60 +78,17 @@ describe('Model > Subject', function () {
...
     before(function () {
       subject.addToCollection()
     })
 
     it('should call the onAddToCollection callback', function () {
-      expect(subject.onAddToCollection).to.have.been.calledOnceWith(subject.id)
-    })
-  })
-
-  describe('openInTalk', function () {
```
eatyourgreens: I don't think these tests for going to Talk should be removed. Looks like I started them but forgot to copy the expectations across when I moved the code. https://github.com/zooniverse/front-end-monorepo/pull/901

eatyourgreens: I've added these tests in #1031.

srallen: Thanks for catching this. When I glanced at it, I didn't realize that this was intended to test something different from `openTalkPage` in the `SubjectStore` spec file and then thought it was cruft.

packages/lib-classifier/src/store/Subject.spec.js
```diff
@@ -26,11 +29,46 @@ describe('Model > Subject', function () {
...
+      subject.workflows = WorkflowStore.create({})
+      subject.workflows.setResource(workflow)
+      subject.workflows.setActive(workflow.id)
+      expect(subject.viewer).to.be.null()
+    })
+
+    it('should return the single image viewer for subjects with a single image location', function () {
+      const singleImageSubject = SubjectFactory.build({ locations: [{ 'image/png': 'https://foo.bar/example.png' }] })
+      const subjectStore = Subject.create(singleImageSubject)
+      subjectStore.workflows = WorkflowStore.create({})
```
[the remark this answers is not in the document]
srallen: I'll rename to `subjectResourceStore` which matches its actual name in MST.

packages/lib-classifier/src/store/Subject.spec.js
eatyourgreens: I thought I'd tested these too, but I don't see any tests in the PR that added them.

srallen:
@eatyourgreens and @rogerhutchings this is ready for review again.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -51,97 +46,78 @@ describe('Model > SubjectStore', function () {
...
         sinon.stub(clientStub.panoptes, 'get').callsFake(() => Promise.resolve({ body: [] }))
-        while (rootStore.subjects.resources.size > 0) {
-          rootStore.subjects.advance()
-        }
+        rootStore = setupStores()
       })
 
       after(function () {
         clientStub.panoptes.get.restore()
+        rootStore = null
```
eatyourgreens: This isn't needed. `rootStore` is scoped to this function with `let`, so it will be destroyed when the describe block finishes.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -51,97 +46,78 @@ describe('Model > SubjectStore', function () {
...
       })
 
       it('should leave the active subject empty', function () {
+        while (rootStore.subjects.resources.size > 0) {
```
eatyourgreens: Shouldn't this be in the `before` block?

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -51,97 +46,78 @@ describe('Model > SubjectStore', function () {
...
-              })
-            }
-          }
-        }
-        }
-      )
+  describe('Views > isThereMetadata', function () {
+    let rootStore
+    afterEach(function () {
+      rootStore = null
```
eatyourgreens: Instead of resetting `rootStore` between tests, you could use `const rootStore = setupStores()` inside the tests and not have to worry about accidentally changing the mock for one test inside a previous test.

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: Re-reading this, shouldn't the before block set up the store, populate the queue then advance it until it's empty? At the moment, I think this block is testing the initial state of the queue.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -51,97 +46,78 @@ describe('Model > SubjectStore', function () {
 
     describe('with less than three subjects in the queue', function () {
       let populateSpy
+      let rootStore
 
       before(function () {
+        rootStore = setupStores()
         populateSpy = sinon.spy(rootStore.subjects, 'populateQueue')
```
eatyourgreens: Here, the queue should be populated, then advanced until there are less than 3 subjects.

Previously, a populated queue was set up in the outer function, then advanced to the required state by these inner functions. Now, each inner function is setting up a new queue to test so the initial conditions for each test have to be set up too. Otherwise, you're just testing the initial state of an empty queue.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -51,97 +46,78 @@ describe('Model > SubjectStore', function () {
...
-        }
       })
 
       after(function () {
         rootStore.subjects.populateQueue.restore()
+        rootStore = null
       })
 
       it('should request more subjects', function () {
+        while (rootStore.subjects.resources.size > 2) {
```
eatyourgreens: This loop should be in the before block.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -51,97 +46,78 @@ describe('Model > SubjectStore', function () {
...
-              get: () => {
-                return Promise.resolve({
-                  body: { subjects: [subjectWithHiddenMetadata] }
-                })
-              }
-            }
-          }
-        }
-      )
+      const getStub = stubPanoptesJs({ subjects: subjectWithHiddenMetadata })
```
eatyourgreens: Should `subjects` be an array here?

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -150,24 +126,11 @@ describe('Model > SubjectStore', function () {
...
-              get: () => {
-                return Promise.resolve({
-                  body: { subjects: [subjectWithMetadata] }
-                })
-              }
-            }
-          }
-        }
-      )
+      const getStub = stubPanoptesJs({ subjects: subjectWithMetadata })
```
eatyourgreens: Should `subjects` be an array here too?

packages/lib-classifier/src/store/Subject.spec.js
```diff
@@ -26,11 +29,46 @@ describe('Model > Subject', function () {
...
+      subjectStore.workflows.setResource(workflow)
+      subjectStore.workflows.setActive(workflow.id)
+      expect(subjectStore.viewer).to.equal(subjectViewers.singleImage)
+    })
+
+    it('should reutrn the light curve viewer if the workflow configuration is defined', function () {
```
eatyourgreens:
```suggestion
    it('should return the light curve viewer if the workflow configuration is defined', function () {
```

packages/lib-classifier/src/store/SubjectStore.js
```diff
@@ -77,19 +79,24 @@ const SubjectStore = types
...
+      if (validSubjectReference) {
+        const subject = self.active
+        const shouldShowFeedback = root.feedback.isActive && root.feedback.messages.length && !root.feedback.showModal
+        if (!shouldShowFeedback && subject && subject.shouldDiscuss) {
+          const { url, newTab } = subject.shouldDiscuss
+          openTalkPage(url, newTab)
+        }
+        next(call)
+      } else {
+        next(call)
```
eatyourgreens: Should this be `abort()`? If you always want to call `next(call)`, then it should be moved outside the if block.

eatyourgreens:
>But in test, this caused an infinite loop: populateQueue called advance when there was no active reference, then interrupted by middleware, which calls advance when there is no active reference, which calls populateQueue when there's less than 3 subjects because this is called before the active reference is set, then populateQueue calls advance, etc.

I think this is happening when one test mocks `populateQueue` so that the returned subject queue is always empty. Should that test be modified, or is this a case we should guard against (the API returning an empty array)? If the latter, maybe `populateQueue` should be modified to check that there are returned subjects before calling `self.advance()`?

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: `stubPanoptesJs` can take either a single resource stub or an array of resource stubs. A single stub will be put into an array.

packages/lib-classifier/src/store/ResourceStore.js
```diff
@@ -22,9 +22,13 @@ const ResourceStore = types
...
         self.headers = response.headers
-        const resource = response.body[type][0]
-        self.loadingState = asyncStates.success
-        return resource
+        if (response.body[type] && response.body[type].length > 0) {
+          const resource = response.body[type][0]
+          self.loadingState = asyncStates.success
+          return resource
+        } else {
+          return undefined
```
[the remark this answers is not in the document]
srallen: I don't think so. If the response isn't 400+ or 500+, then it's successful. I think it's possible for some endpoints to return a successful response that is an empty array for the body and resource type.

srallen:
> If the latter, maybe populateQueue should be modified to check that there are returned subjects before calling self.advance()?

This is the resolution included in the PR. The new queue load observer doesn't call `advance` unless there was a successful request, the resources size is greater than 0, and there is not an active subject resource. The classification submission observer handles subsequent advance calls.

packages/lib-classifier/src/store/ResourceStore.js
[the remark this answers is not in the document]
eatyourgreens: Isn’t the loading state still ‘loading’ here? It only gets set to success if subjects are returned.

eatyourgreens:
The old code only called `self.advance()` if the response wasn’t empty, so that’s what’s confusing me. I’m not grasping what the change is.

srallen:
In test, `advance` is being called directly, which gets interrupted by the middleware and continues the call, and the calls `populateQueue`, which then calls `advance` and the loop starts. Looking at this again, I think the tests can be refactored to maybe avoid this, but right now it's possible to call `advance` outside the store, so we should safe guard against a loop starting regardless I think.

eatyourgreens:
`populateQueue` doesn’t call `advance` if the returned queue is empty. I’m completely missing something here.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: The while loop doesn't run in the before block because it gets called before the async populate queue call is done in the setup. It does run before the expectation is called in the it statement. I could try putting it in a `beforeEach` instead.

packages/lib-classifier/src/store/ResourceStore.js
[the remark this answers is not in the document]
srallen: Ah, yes. You're correct. I'll push up a fix.

srallen:
I've refactored the test setup to make the queue observer unnecessary, so hopefully this is more clear to follow what's going on. I'm still concerned that it might be possible to have a loop scenario, but hopefully tests would catch that if it happened.

eatyourgreens:
I forgot that attaching the workflow starts an async request to populate the queue.

srallen:
@eatyourgreens this is ready for another review

eatyourgreens:
I'll have a look, but the test that needed fixing was for the case where we empty the queue without receiving any new subjects back from the API.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -1,147 +1,131 @@
...
     describe('after emptying the queue', function () {
+      let rootStore
       before(function () {
-        sinon.stub(clientStub.panoptes, 'get').callsFake(() => Promise.resolve({ body: [] }))
+        const clientStub = stubPanoptesJs({ workflows: workflow, subjects: [] })
+        rootStore = setupStores(clientStub)
+      })
+
+      beforeEach(function () {
         while (rootStore.subjects.resources.size > 0) {
```
eatyourgreens: The queue size is 0 here, because the root store is initialised with an empty queue, so this loop never runs and we don't test what happens when we advance a queue until it is empty.

Can the root store be initialised with subjects in the queue? On master, this test re-uses the root store from thr previous block, which has 2 subjects in the queue.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -1,147 +1,131 @@
...
 
       it('should request more subjects', function () {
-        expect(populateSpy).to.have.been.calledOnce()
+        expect(rootStore.subjects.populateQueue).to.have.been.calledOnce()
+      })
+
+      it('should request more subjects when the initial queue response only has less than 3 subjects', function () {
+        const clientStub = stubPanoptesJs({ workflows: workflow, subjects: shortListSubjects })
+        rootStore = setupStores(clientStub)
+        expect(rootStore.subjects.populateQueue).to.have.been.calledOnce()
```
eatyourgreens: I think `populateQueue` should be called twice here: once when the workflow is attached (to prepopulate the queue), then again when the first active subject is advanced (because the initial queue is only 2 subjects.)

eatyourgreens: Should this test call `rootStore.subjects.advance()` or does that happen automatically after the workflow is attached?

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -1,147 +1,131 @@
...
       after(function () {
         rootStore.subjects.populateQueue.restore()
       })
 
       it('should request more subjects', function () {
-        expect(populateSpy).to.have.been.calledOnce()
+        expect(rootStore.subjects.populateQueue).to.have.been.calledOnce()
+      })
+
+      it('should request more subjects when the initial queue response only has less than 3 subjects', function () {
```
eatyourgreens: Minor point for readability but maybe put this into its own `describe` block:
```js
describe('when the initial response has less than 3 subjects', function () {
  before(function () { /* mock the API to return 2 subjects */ })
  it('should request more subjects', function () {…})
})
```

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: I'm resetting the history before the expectation in the `beforeEach` because this is only testing for the behavior to call `populateQueue` again. I thought it'd be confusing to test for both calls, since that's not being stated here as being tested.

eatyourgreens: Is that because this block just observes what happens when we drop an existing queue from 3 to 2 subjects?

What might be missing here is that the `beforeEach` block doesn't repopulate the queue before advancing it down to 2 subjects, so that behaviour is only tested in the first test. It's hard to tell, though. This new test sets up a new, empty queue which will then populate when the workflow attaches and you might be testing for the initial call that happens then.

eatyourgreens: I've double-checked, and the queue is empty at this point in the test, so the expectation is being checked before the API promise has resolved with 2 new subjects.

srallen: I forgot about the asynchronicity of the initial store setup myself here, so moving this into its own describe block with before function setup should not only improve readability but ensure that we're testing what we think here.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: I haven't had much success setting a map via snapshot for test mocks, so it's been easier for me using the actual API of the stores. I've removed the reset history call from the spy to show that the calls happen as expected, once on initialization and once when the queue advances to below three subjects, so the while loop here does work. Instead I've left comments clarifying what's being tested.

srallen:
@eatyourgreens I believe I've resolved the last few comments, so this is ready for review again.

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: I'm running this locally, logging the while loop, and it isn't running. The test is setting up an empty queue, skipping the while loop then testing that the empty store is empty. Maybe it should expect `subjects.advance` to have been called at least once?

If I add that expectation, I get this failure.
```sh
1) Model > SubjectStore
       Actions > advance
         after emptying the queue
           should leave the active subject empty:
     AssertionError: expected advance to have been called at least once, but it was never called
      at Context.called (src/store/SubjectStore.spec.js:122:57)

```

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -64,84 +67,82 @@ describe('Model > SubjectStore', function () {
...
+        })
+
+        it('should request more subjects', function () {
+          // Once for initialization
+          expect(rootStore.subjects.populateQueue).to.have.been.calledOnce()
+        })
       })
     })
 
     describe('after emptying the queue', function () {
```
eatyourgreens:
```suggestion
    describe('when no subjects are returned from the API', function () {
```
In retrospect, this is a better description of this test case.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -64,84 +67,82 @@ describe('Model > SubjectStore', function () {
...
         while (rootStore.subjects.resources.size > 0) {
           rootStore.subjects.advance()
         }
       })
 
-      after(function () {
-        clientStub.panoptes.get.restore()
-      })
-
       it('should leave the active subject empty', function () {
```
eatyourgreens:
```suggestion
      it('should advance the queue until it is empty', function () {
```
I think this is also a better description of the test. I'm thinking maybe expect `rootStore.subjects.advance` to have been called here?

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -64,84 +67,82 @@ describe('Model > SubjectStore', function () {
...
+        before(function () {
+          const clientStub = stubPanoptesJs({ workflows: workflow, subjects: [] })
+          rootStore = setupStores(clientStub)
+        })
+
+        it('should request more subjects', function () {
+          // Once for initialization
+          expect(rootStore.subjects.populateQueue).to.have.been.calledOnce()
+        })
       })
```
eatyourgreens: Maybe test for `it should not advance the queue` too? I think that's the difference between a request that returns subjects and a request that doesn't.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -64,84 +67,82 @@ describe('Model > SubjectStore', function () {
...
+      describe('when the initial response has less than three subjects', function () {
+        let rootStore
+        before(function () {
+          const clientStub = stubPanoptesJs({ workflows: workflow, subjects: shortListSubjects })
+          rootStore = setupStores(clientStub)
+        })
+
+        it('should request more subjects', function () {
+          // Once for initialization and again since less than three subjects in initial response
+          expect(rootStore.subjects.populateQueue).to.have.been.calledTwice()
```
eatyourgreens: This is passing because `populateQueue` was called twice in an earlier test.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -1,59 +1,62 @@
...
-      let populateSpy
+      let rootStore
 
       before(function () {
-        populateSpy = sinon.spy(rootStore.subjects, 'populateQueue')
+        rootStore = setupStores()
+      })
+
+      beforeEach(function () {
         while (rootStore.subjects.resources.size > 2) {
```
eatyourgreens: This will not run before each test unless the queue is reset to 3 or more subjects. Otherwise, it only runs before the first test, then the queue size is 2 for subsequent tests.

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: I think I was confused by the nested variables here. This is a fresh instance of `rootStore`, with a new spy?

srallen: Each describe block has its own root store instantiated.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: Sorry, on first read through of your first comment, I missed that the issue was the setup. So I'll update the panoptes js stub to respond with a queue of subjects. Thanks.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: I don't think that's what we're testing here. We're testing to make sure active isn't being set when the resources size is 0. I think if we want to test advance is being called as its written in the store, we'll need integration tests testing the observers setup between the classification store and the subjects store. If we add an expectation that advance is called here, we're just testing the calls from the `beforeEach` which doesn't seem right to me.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: We're testing the initial API response having no subjects in the previous describe block. I think the confusion here happened with a bad panoptes js client stub setup.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: It was only applicable to the first test in this block, so I'll move the while loop into that test to make sure it's clear.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: I think that makes sense here. Good idea.

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: Sorry but II wrote this test and this is what it was intended to test. It should start with a populated queue, advance that queue until it’s empty then check that it really is empty. It’s a regression test to catch a bug that happened if you call advance on the queue when the number of subjects is small.

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: I intended this test to test the case where the queue is being advanced but isn’t being refilled from Panoptes.

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: I think that will break this test because we do want to advance the queue until it’s empty, and make sure that the store doesn’t crash when you advance a queue that has 2 or fewer subjects.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: Ok, we can include the suggested refactor here in the issue about improving these tests.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: Could you explain how it'd break the test? The stores are instantiated with an initial async request to populate the queue, the panoptes js stub is now correctly responding with a queue of subjects and then the while loop is called that calls advance directly, so the queue is being advanced until it is empty before the expectation is run.

eatyourgreens: The while loop will repopulate the queue, when the queue size is less than 3, so that we never reach 0 and don’t test the queue in the case where it isn’t being refilled.

srallen: If the test was asynchronous, then it would fail because of the reason you've given. I suppose relying on the synchronicity of the current expectation is a bit brittle and unclear.

srallen: The suggested refactor to separate the populate and advance actions I think will help with this.

eatyourgreens: Yes. What this test should really do is add 2 subjects to the initial queue during setup. 

At the moment it’s setting up the conditions for an infinite loop of `advance` calling `populateQueue` calling `advance` etc.

PR #1035, opened 2019-07-22
Add docker to app-content-pages

eatyourgreens:
Add a docker-compose file to the content pages app.
`docker-compose up` to run a server on  http://localhost:3000.
`docker-compose down` to stop the dev server.
`docker-compose run --rm dev test` to run the tests.
`docker-compose run --rm --service-ports dev storybook` to run the storybook on http://localhost:9001.

Package:
app-content-pages

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-content-pages/docker-compose.yml
+++ b/packages/app-content-pages/docker-compose.yml
@@ -2,8 +2,7 @@ version: '3.2'
 
 services:
   dev:
-    build:
-      context: ../../
+    image: front-end-monorepo_dev:latest
     entrypoint:
       - "yarn"
       - "workspace"
--- a/packages/app-content-pages/docker-compose.yml
+++ b/packages/app-content-pages/docker-compose.yml
@@ -3,6 +3,8 @@ version: '3.2'
 services:
   dev:
     image: front-end-monorepo_dev:latest
+    build:
+      context: ../../
     entrypoint:
       - "yarn"
       - "workspace"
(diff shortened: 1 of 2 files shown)


packages/app-content-pages/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+      - "yarn"
+      - "workspace"
+      - "@zooniverse/fe-content-pages"
+    command: ["dev"]
+    ports:
+      - "3000:3000"
+      - "9001:9001"
+    environment:
+      - HOST=0.0.0.0
+      - AWS_REGION=${AWS_REGION}
```
rogerhutchings: afaik none of these AWS env vars are required. you can see the required vars in `env-default` - and these don't need to be set here anyway, since [docker compose supports `.env` files](https://docs.docker.com/compose/environment-variables/#the-env-file)

eatyourgreens: Ok. This is just copied from the docker-compose file we use to run the project app.

eatyourgreens: I think HOST does need to be set, though.

eatyourgreens: webpack dev server doesn't work without HOST but express is fine. 🤷‍♂ computers, eh?

PR #1038, opened 2019-07-23
Fix client-side routing

rogerhutchings:
- Modify nav links to do proper dynamic routing
- Trim out some cruft in _app.js
- Fix helper method to be client/server-agnostic
- Add some additional pages to be filled in later
- Refactor some nav components to use correct link props, and fetch
  store values closer to where they're used
- Update tests
- Run a lint


packages/app-project/src/components/ProjectHeader/components/ProjectTitle/ProjectTitle.js
```diff
@@ -16,10 +17,25 @@ const StyledHeading = styled(Heading)`
...
+    </StyledHeading>
+  )
+
+  if (link) {
+    result = (
+      <Link as={link.as} href={link.href} passHref>
+        <StyledAnchor>
+          {result}
+        </StyledAnchor>
+      </Link >
```
eatyourgreens: Shouldn't the heading wrap the link, not the other way round? I'm not sure about the markup here.

eatyourgreens:
The error page seems to have broken, but I'm not sure if that's a big deal.
```sh
dev_1  | [ event ] client pings, but there's no entry for page: /_error
dev_1  | [ event ] disposing inactive page(s): /next/dist/pages/_error
dev_1  | [ event ] build page: /next/dist/pages/_error
dev_1  | [ wait ]  compiling ...
```

eatyourgreens:
Other than that, works for me browsing a test project locally 👍 

I haven't had time to go through and read the changes in detail.

packages/app-project/src/components/ProjectHeader/components/ProjectTitle/ProjectTitle.js
rogerhutchings: `h1` is flow content, which is a permitted tag inside an anchor: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Properties

rogerhutchings: In terms of interactivity, it doesn't make any difference here as it's laid out with flexbox. Under normal circumstances, it'd be the difference between making a `block` element clickable, or just the text.

rogerhutchings:
> The error page seems to have broken, but I'm not sure if that's a big deal.

Not sure why these changes would have affected that; there's some more work to do on making the app properly throw 404s if a project isn't found, so we'll need to circle back to that anyway.

eatyourgreens:
Oh, I was able to test this PR by running `docker-compose up` and browsing the project app, so that was convenient.

rogerhutchings:
Okay, no one has any further comments, so I'm going to merge this.

PR #1040, opened 2019-07-23
[WIP] simple BarChartViewer

srallen:
Package: lib-classifier

Describe your changes:
Initial pass at adding a read only bar chart. I'm taking a pretty different approach than what we did for the LCV. I've never felt totally comfortable handing over the rendering and manipulation of the DOM over the D3 and I find the code very hard to read, partly because of d3's chaining API.

Instead, I'm trying out only using React for DOM manipulation and rendering and only using D3 for the maths (which is more in line with a lot of the React+D3 libraries out there).

Conceptually, I'm borrowing really heavily from [vx](https://vx-demo.now.sh/). I believe it was potentially one of our candidates last fall when we were evaluating react+d3 libraries, but it was very young and was missing core features like zoom. Now it has at least zoom and may also have the rest of what we might need. 

This begs the question, do we continue rolling our own low level components and d3 use or do we adopt vx? If vx is missing anything, do we contribute back, fork it, or just continue our own code?

We're going to have more requests for complex data visualizations in the future and I'd like us to reevaluate adopting a react+d3 library to save us time. 

If we decide to try out vx in earnest, I could redo this PR with it. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/BarChartViewer/BarChartViewer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/BarChartViewer/BarChartViewer.js
@@ -1,0 +1,63 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import { scaleLinear, scaleBand } from '../../helpers/d3'
+import Chart from '../SVGComponents/Chart'
+import Background from '../SVGComponents/Background'
+import Group from '../SVGComponents/Group'
+import Bar from './components/Bar'
+
+const BarChartViewer = React.forwardRef(function BarChartViewer({ backgroundFill, data, height, width }, ref) {
+  const yMax = height - 120
+  const xScale = scaleBand({
+    domain: data.map(datum => datum.label),
+    rangeRound: [0, width],
+    padding: 0.5
+  })
+  const yScale = scaleLinear({
+    domain: [0, Math.max(...data.map(datum => datum.value))],
+    rangeRound: [yMax, 0]
+  })
+  return (
+    <Chart ref={ref}>
+      <Background fill={backgroundFill} />
+      <Group>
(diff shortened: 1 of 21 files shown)


srallen:
I quickly put together #1056 using vx for comparison.

srallen:
This doesn't have an axis, but using vx, I got one added to a basic bar chart in ~10 mins. This could be a big time saver.

mcbouslog:
My initial impression is that (this) react for rendering and D3 for maths and low level components would be preferable than D3 for rendering and the rest (LCV). I find the component and render code written for react easier to understand. Is it accurate to compare [this code in the LCV](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js#L389-L429) to BarCartViewer.js? That's largely what I'm basing my opinion on. I know not a direct comparison (plot vs bar), but the basics of initializing a chart with data.

I also think using vx (#1056) is a good idea. It looks like it handles the basic data visualization components well, while leaving what we'd want to control up to us.
- for easy reference: https://github.com/zooniverse/front-end-monorepo/compare/react-d3-bars...react-vx-bars

A lot of the code in the LCV deals with annotations and scaling everything appropriately. With vx's components it looks like that should be relatively straight-forward, but depending on difficulty might be good to test before formally deciding. [Most](https://vx-demo.now.sh/static/docs/vx-zoom.html) vx components look like they have some documation, but [some](https://vx-demo.now.sh/static/docs/vx-brush.html) might be missing, possibly relevant [vx brushes github issue](https://github.com/hshoff/vx/issues/14).

srallen:
I'm gonna close this because I think vx is gonna work out. I'll write up an ADR.

PR #1041, opened 2019-07-24
Add docker-compose to the project app

eatyourgreens:
Build from the monorepo image.
Start a dev server on http://localhost:3000 with `docker-compose up`.
Stop the dev server with `docker-compose down`.
Run the tests with `docker-compose run --rm dev test`.

Package:
app-project

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/docker-compose.yml
+++ b/packages/app-project/docker-compose.yml
@@ -10,6 +10,17 @@ services:
     command: ["dev"]
     ports:
       - "3000:3000"
+    volumes:
+      - ../../:/usr/src
+      - /usr/src/node_modules
+  storybook:
+    image: front-end-monorepo_dev:latest
+    entrypoint:
+      - "yarn"
+      - "workspace"
+      - "@zooniverse/fe-project"
+    command: ["storybook"]
+    ports:
       - "9001:9001"
     volumes:
       - ../../:/usr/src
--- a/packages/app-project/docker-compose.yml
+++ b/packages/app-project/docker-compose.yml
@@ -3,6 +3,8 @@ version: '3.2'
 services:
   dev:
     image: front-end-monorepo_dev:latest
+    build:
+      context: ../../
     entrypoint:
       - "yarn"
       - "workspace"
--- a/packages/app-project/docker-compose.yml
+++ b/packages/app-project/docker-compose.yml
@@ -1,0 +1,15 @@
+version: '3.2'
+
+services:
+  dev:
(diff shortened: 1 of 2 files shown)


packages/app-project/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
+version: '3.2'
+
+services:
+  dev:
+    image: front-end-monorepo_dev:latest
```
camallen: Looks like this is a locally built image that is then used to run. Normally you'd specify a build cmd/context vs an image here. 

I might be missing how this works in practice though.

https://docs.docker.com/compose/compose-file/#image

> If the image does not exist, Compose attempts to pull it, unless you have also specified build, in which case it builds it using the specified options and tags it with the specified tag.

https://docs.docker.com/compose/compose-file/#build
> If you specify image as well as build, then Compose names the built image with the webapp and optional tag specified in image:

eatyourgreens: I was finding that `docker-compose up` in each package directory ran a new build, eg. `app-project_dev:latest`, which is just an exact copy of the monorepo build. I haven't tried it with a clean install, though, and I could be misunderstanding how Docker caches/reuses images.

camallen: Right - gotcha. This might be a way around the multiple builds then. I assume `docker-compose dev build` works just fine but has to be done before running `up`. 

OOI does running `docker-compose up` with no built image kick off multiple builds? 

Nice work around :)

eatyourgreens: At the moment this just crashes if the monorepo image hasn't been built. I already had that image built on my local machine so didn't notice. `docker-compose build` in the top-level directory might be a prerequisite before running any of the packages in their package directories.

I can play around with getting build and image to work together. When I tried them last week, Docker ignored the existing image and ran the build anyway. Maybe order in the config file is important?

eatyourgreens: Adding a build context for `dev` fixes this, so I'll push that up.

`docker-compose up` builds and tags, if I delete the existing image.
```sh
Successfully built 4cf45fed9c0b
Successfully tagged front-end-monorepo_dev:latest
WARNING: Image for service dev was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
Creating app-project_storybook_1 ... done
Creating app-project_dev_1       ... done
Attaching to app-project_dev_1, app-project_storybook_1
```

PR #1044, opened 2019-07-25
Make the minimum subject queue size configurable

### LINKED ISSUE #1042: Update queue when queue length is less than 2
**Package** lib-classifier

Change the condition for additional subject requests to be for less than 2 in the queue. This will be consistent with PFE and it's important for workflows configured to use designator ratios to better guarantee serving up the ratios as intended.

eatyourgreens:
Adds MINIMUM_QUEUE_SIZE to the subject queue, sets it to 2 and updates the tests to use it.
Fixes a couple of errors that came up when changing the tests. It turns out I didn't declare `rootStore` properly in #909 and the tests were all sharing the same mock store state.
Adds a couple of tests to check that advancing the queue reduces the queue size by one, and that advancing the queue works even if the store is empty.

Package:
lib-classifier

Closes #1042 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/SubjectStore.spec.js
+++ b/packages/lib-classifier/src/store/SubjectStore.spec.js
@@ -1,7 +1,7 @@
 import sinon from 'sinon'
 import ProjectStore from './ProjectStore'
 import RootStore from './RootStore'
-import SubjectStore, { openTalkPage } from './SubjectStore'
+import SubjectStore, { openTalkPage, MINIMUM_QUEUE_SIZE } from './SubjectStore'
 import WorkflowStore from './WorkflowStore'
 import { ProjectFactory, SubjectFactory, WorkflowFactory } from '../../test/factories'
 import { Factory } from 'rosie'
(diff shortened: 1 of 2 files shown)


packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -68,78 +84,58 @@ describe('Model > SubjectStore', function () {
...
       })
 
-      it('should leave the active subject empty', function () {
+      it('should advance until empty without crashing', function () {
+        expect(rootStore.subjects.resources.size).to.equal(10)
+        while (rootStore.subjects.resources.size > 0) {
+          rootStore.subjects.advance()
+        }
+        expect(rootStore.subjects.resources.size).to.equal(0)
+        expect(rootStore.subjects.active).to.be.undefined()
```
eatyourgreens: Should the active subject be undefined here, or should it be the last subject that was in the initial queue?

PR #1045, opened 2019-07-25
Fix shadowing in the SubjectStore tests

eatyourgreens:
Adds a describe block to make sure every declaration of rootStore is locally scoped to a function.
Adds a test to check that advancing the queue reduces its size by one.

Package:
lib-classifier

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/SubjectStore.spec.js
+++ b/packages/lib-classifier/src/store/SubjectStore.spec.js
@@ -39,7 +39,7 @@ describe('Model > SubjectStore', function () {
   describe('Actions > advance', function () {
     describe('with a full queue', function () {
       let rootStore
-      let firstSubject
+      let previousSubject
       let initialSize
 
       before(function () {
@@ -47,7 +47,7 @@ describe('Model > SubjectStore', function () {
       })
 
       beforeEach(function () {
-        firstSubject = rootStore.subjects.resources.values().next().value.id
+        previousSubject = rootStore.subjects.resources.values().next().value.id
         initialSize = rootStore.subjects.resources.size
         rootStore.subjects.advance()
       })
@@ -59,8 +59,8 @@ describe('Model > SubjectStore', function () {
       it('should make the next subject in the queue active', function () {
         const currentSubject = rootStore.subjects.resources.values().next().value.id
         expect(rootStore.subjects.active.id).to.equal(currentSubject)
-        expect(rootStore.subjects.resources.get(firstSubject)).to.be.undefined()
-        expect(firstSubject).to.not.equal(currentSubject)
+        expect(rootStore.subjects.resources.get(previousSubject)).to.be.undefined()
+        expect(previousSubject).to.not.equal(currentSubject)
       })
     })
 
--- a/packages/lib-classifier/src/store/SubjectStore.spec.js
+++ b/packages/lib-classifier/src/store/SubjectStore.spec.js
@@ -12,7 +12,7 @@ const shortListSubjects = Factory.buildList('subject', 2)
 
 const clientStub = stubPanoptesJs({ subjects, workflows: workflow })
 
-describe.only('Model > SubjectStore', function () {
+describe('Model > SubjectStore', function () {
   function setupStores(panoptesClientStub = clientStub) {
     const store = RootStore.create({
       classifications: {},
--- a/packages/lib-classifier/src/store/SubjectStore.spec.js
+++ b/packages/lib-classifier/src/store/SubjectStore.spec.js
@@ -12,7 +12,7 @@ const shortListSubjects = Factory.buildList('subject', 2)
 
 const clientStub = stubPanoptesJs({ subjects, workflows: workflow })
 
-describe('Model > SubjectStore', function () {
+describe.only('Model > SubjectStore', function () {
   function setupStores(panoptesClientStub = clientStub) {
     const store = RootStore.create({
       classifications: {},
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Tests are failing because `advance()` doesn't change the size of the queue. That's weird. Am I testing a reference, instead of the value at any given instant?

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -37,35 +37,45 @@ describe('Model > SubjectStore', function () {
...
-      it('should request more subjects', function () {
-        while (rootStore.subjects.resources.size > 2) {
-          rootStore.subjects.advance()
-        }
-        // Once for initialization and once after the queue has been advanced to less than 3 subjects
-        expect(rootStore.subjects.populateQueue).to.have.been.calledTwice()
+      it('should reduce the queue size by one', function () {
+        const expectedSize = rootStore.subjects.resources.size - 1
+        rootStore.subjects.advance()
+        expect(rootStore.subjects.resources.size).to.equal(expectedSize)
```
eatyourgreens: Tests are failing because `rootStore.subjects.resources.size` doesn't change here, and I can't immediately see why that would happen.

eatyourgreens:
Working on this, I discovered that there isn't a subject with ID '1' in the mock queue, so I've updated that test to use subject IDs from the queue.

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: Manually calling `populateQueue` fixed this. I don't know why.

srallen: Because `populateQueue` is asynchronous. The call for it in the store instantiation's `afterAttach` wasn't done yet. This was resolved in other describe blocks by setting up the stores in a `before` or `beforeEach`. This test is now setting up the stores with an unfinished `populateQueue` call and then manually calls it for the expectations. 

It seems cleaner to me to do the store instantiation in a `before` or `beforeEach` since these are really integration tests now and we are testing the behavior of both actions here, but this may be more readable? I'm not sure what I'd prefer. What I'm concerned with is that people will read this and not realize we are calling `populateQueue` on store instantiation in its `afterAttach` lifecycle method.

srallen:
> Working on this, I discovered that there isn't a subject with ID '1' in the mock queue, so I've updated that test to use subject IDs from the queue.

Rosie generates the ids in the factories, so we do need to be sure to be correctly using the generated factories for the expectations.

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: At this point, the code was using `before`. If you check out that commit, you should see that the tests fail.

eatyourgreens: I’d much prefer to have this working with the initial setup in a `before` block. I noticed that the metadata tests had to use the same hack.

eatyourgreens:
Tip I learnt from Cam: when you test a state change, test the initial state as well as the final state.

We’ve had the occasional test that passed because state didn’t change but initial state just happened to meet the expectations.

srallen:
That's a good idea. I'm going to come back to this after I finish up some updates to the feedback store which I missed previously and I figured out that's why I'm having trouble with the classification store updates.

eatyourgreens:
>Rosie generates the ids in the factories, so we do need to be sure to be correctly using the generated factories for the expectations.

I'm noticing that the mock subjects are classified in ID order: `10, 11, 12…`. A real response from Panoptes won't be ordered by subject ID, so we should simulate that in the tests too, just in cse the queue is changing the subject order.

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: I've removed that hack, because I think it makes it confusing to try and understand what the initial mock store state will be.

What you should see now is that adding a new expectation causes previously passing tests to fail. Do you have any ideas why the active subject is undefined after advancing the queue?

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -37,35 +37,52 @@ describe('Model > SubjectStore', function () {
...
-        // Once for initialization and once after the queue has been advanced to less than 3 subjects
-        expect(rootStore.subjects.populateQueue).to.have.been.calledTwice()
+      it('should reduce the queue size by one', function () {
+        expect(rootStore.subjects.resources.size).to.equal(initialSize - 1)
+      })
+
+      it('should make the next subject in the queue active', function () {
+        const currentSubject = rootStore.subjects.resources.values().next().value.id
+        expect(rootStore.subjects.active.id).to.equal(currentSubject)
+        expect(rootStore.subjects.resources.get(previousSubject)).to.be.undefined()
```
eatyourgreens: This is failing because `previousSubject` is still the active subject.

srallen:
I've rebased this branch against the latest `update-store-instantiation` branch.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: I see this as passing, so the latest commits + rebase appears to be working.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: This is passing now, so the latest commits are working.

srallen:
> I'm noticing that the mock subjects are classified in ID order: 10, 11, 12…. A real response from Panoptes won't be ordered by subject ID, so we should simulate that in the tests too, just in case the queue is changing the subject order.

We can add tests for a particular subject queue order once the sequential subjects feature is added to Panoptes. I don't think this matters as of yet.

eatyourgreens:
>We can add tests for a particular subject queue order once the sequential subjects feature is added to Panoptes. I don't think this matters as of yet.

Sorry, I meant that we should mock a response with a random subject order, just to check that the map isn't ordering items by ID. Real responses from Panoptes aren't ordered by ID, like they are in these tests.

eatyourgreens:
Removing the previous subject on subject advance still fails for me. I'm not sure why.

```sh
718 passing (3s)
  7 pending
  1 failing

  1) Model > SubjectStore
       Actions > advance
         with a full queue
           should make the next subject in the queue active:
     AssertionError: expected { Object (id, already_seen, ...) } to be undefined
      at Context.undefined (src/store/SubjectStore.spec.js:62:73)
```

PR #1046, opened 2019-07-25
[MST Upgrade] FeedbackStore updates

srallen:
Package: lib-classifier

Describe your changes:
I missed some updates that the FeedbackStore needed on first pass, so these are the rest. I've added `isValidReference` where appropriate and updated tests. Purposely left in an `only` to show tests are passing.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/store/FeedbackStore.spec.js
```diff
@@ -1,150 +1,196 @@
...
     })
   })
 
-  describe('reset', function () {
+  describe('Actions > reset', function () {
+    let feedback, feedbackStub
     beforeEach(function () {
+      feedbackStub = FeedbackFactory.build({ isActive: true, rules: rulesStub, showModal: true })
+      feedback = FeedbackStore.create(feedbackStub)
       feedback.subjects = {
```
eatyourgreens: Should `feedback.subjects` be a mock subjectStore?

packages/lib-classifier/src/store/FeedbackStore.spec.js
```diff
@@ -1,150 +1,196 @@
...
       helpers.generateRules.resetHistory()
     })
 
     it('should set active state', function () {
-      expect(helpers.isFeedbackActive).to.have.been.calledOnceWith(project, subject, workflow)
+      expect(feedback.isActive).to.be.false()
+      feedback.createRules(subject)
+      expect(helpers.isFeedbackActive).to.have.been.calledOnce()
+      // Normally we would use sinon's withArgs call to test the spy for specific arguments
+      // But this is complicated with the fact that the arguments actually used in the store are MST references
```
eatyourgreens: Can we use the MST references eg. `feedback.projects.active`? Seems clearer than using `.args[0][0]` to get the first argument of the first call.

srallen: Does this mean to do this instead?
`expect(feedback.projects.active.toJSON()).to.deep.equal(project)`
We wouldn't be testing the spy anymore though doing this? We would be testing that the store is setting the resource map and active reference to our mocked resource. Unless I'm misunderstanding what the recommendation is here.

eatyourgreens: I was thinking `.withArgs(feedback.projects.active…)` to check for the MST references.

srallen: Ah yes. I tried that but now I'm not remembering why that didn't work. I'll try it again.

eatyourgreens: It's not a blocker, and I'm not massively familiar with how feedback works internally but that just came to mind when I was reading the code.

packages/lib-classifier/src/store/FeedbackStore.spec.js
```diff
@@ -1,150 +1,196 @@
...
+      // so they will not equal the stub resource object
+      // We convert the arguments to standard json for the expectations below
+      // Expect for subject
+      expect(helpers.isFeedbackActive.args[0][0].toJSON()).deep.equal(project)
+      expect(helpers.isFeedbackActive.args[0][1]).deep.equal(subject)
+      expect(helpers.isFeedbackActive.args[0][2].toJSON()).deep.equal(workflow)
+      expect(feedback.isActive).to.equal(helpers.isFeedbackActive.returnValues[0])
```
eatyourgreens: Maybe code this to `true` or `false`, depending on what we expect from the mock. I'm assuming it's supposed to set `isActive` to true.

If `isActive` is false here, that could be because it was false at the beginning of the test.

srallen: Before this test didn't actually test that `isFeedbackActive` was setting `self.isActive` to its return value. The store was instantiated with `isActive` as true, the stubbed function was just returning `feedback.isActive`, so it didn't make sense based on the `it` statement since the store and stubbed function was always true and referring to itself with no explicit and testable change. 

I've updated this so that the store is instantiated with `isActive` as false and tests for it, the stubbed function specifically returns `true`, and test that the function's return value is now set to `isActive`. Now this is testing what the `it` statement says it is testing. 

I can add another expectation that `isActive` is true, but I'll leave this one as is because we want to be sure that `isActive` is set to the return value since that's what the code is doing.

eatyourgreens: OK, so if I follow correctly, then if the code is broken so that `isFeedbackActive()` returns false for these rules, this test will pass even though the store is broken?

srallen: I would say that in that scenario that it is `isFeedbackActive` is what is broken and its own tests would ideally fail, not that the store is broken. The code sets `isActive` to whatever the return value of `isFeedbackActive`, which could be the same as what `isActive` is already set to. It doesn't check prior to setting it that it is not equal to the current setting. Maybe that should change in a future update and then this test would cover that too?

For the purposes of the test though, I think adding another expectation here that `isActive` is true is a good idea to make it explicit that the store is being updated with the return value of `isFeedbackActive` that we expect, but I don't think it's a good idea to replace the return value expectation.

packages/lib-classifier/src/store/FeedbackStore.spec.js
srallen: I can't remember what I was doing to make that not work, but I have it working now so I'll push an update since it is more readable.

PR #1049, opened 2019-07-29
Add a production build for local testing

eatyourgreens:
`docker-compose up` from the top-level `front-end-monorepo` directory should start a local server, running the following services:
- `test`: runs `yarn test` once on startup.
- `fe-project`: serves the project production app on http://localhost:3000.
- `fe-content`: serves the project content pages app on http://localhost:3001.

If the production image doesn't exist, it should be built by `docker-compose up`.
`docker-compose down` to stop the server.

Run the tests for the two NextJS apps with:
`docker-compose run --rm fe-project test`
`docker-compose run --rm fe-content test`

Run the test service to run tests across the whole repo:
`docker-compose run --rm test`

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
Jenkins does not like the multi-stage build in the Dockerfile, but I can't see why. I might take the Jenkins changes out of this PR.

srallen:
As discussed in the call, the jenkins updates will be put into their own WIP PR

eatyourgreens:
The Dockerfile changes here will break deploys from master, because the Jenkins docker plugin can't parse `FROM node:10 AS bootstrap`. https://github.com/jenkinsci/docker-workflow-plugin/pull/162

eatyourgreens:
I've removed the named stages from the Dockerfile and added extra build steps to the bootstrap script instead. Bootstrapping now builds the project and content pages apps after the React components and classifier.

eatyourgreens:
https://github.com/zooniverse/front-end-monorepo/pull/1058 runs without errors on Jenkins, after that change.

srallen:
@eatyourgreens I only had tried out the new docker setup locally. How were you testing with Jenkins?

eatyourgreens:
I’ve got a separate PR that builds this Dockerfile on Jenkins. The Jenkinsfile also has instructions that Roger wrote for testing changes on Jenkins.

srallen:
I needed to prune the old images to get this working again, so issue was on my end.

PR #1053, opened 2019-07-29
[MST Upgrade] Update subject viewer store

srallen:
Package: lib-classifier

I missed some updates the `SubjectViewerStore` needed when I first looked it over. This adds `isValidReference` and a name to the autorun observer. I added one simple unit test for the function of `resetSubject` since that is what is called in the observer, but this store is still missing a lot of tests and that is out of scope for the MST upgrade.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/SubjectViewerStore.spec.js
+++ b/packages/lib-classifier/src/store/SubjectViewerStore.spec.js
@@ -1,23 +1,48 @@
+import asyncStates from '@zooniverse/async-states'
 import SubjectViewerStore from './SubjectViewerStore'
 
-let subjectViewerStore
-
-describe('Model > SubjectViewerStore', function () {
-  before(function () {
-    subjectViewerStore = SubjectViewerStore.create()
-  })
-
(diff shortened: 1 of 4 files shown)


packages/lib-classifier/src/store/SubjectViewerStore.spec.js
```diff
@@ -1,23 +1,48 @@
...
 import SubjectViewerStore from './SubjectViewerStore'
 
-let subjectViewerStore
-
-describe('Model > SubjectViewerStore', function () {
-  before(function () {
-    subjectViewerStore = SubjectViewerStore.create()
-  })
-
+describe.only('Model > SubjectViewerStore', function () {
```
mcbouslog: prob intentional part of this process of merging mini-PRs into main related PR, but just noting `.only`

srallen: Yep, each had the `only` purposely left in to demonstrate that the tests for the store were passing and the branch they merge into is for testing that they all pass.

PR #1057, opened 2019-08-01
WIP Refactor app-project internals 🚧

rogerhutchings:
This is a holding PR - feature branches will be merged into this one, and then this branch will be merged into `master`.

I've come across some issues while building some of the new project widgets. These are mostly to do with low-level next.js setup and configuration, and I want to make sure these all get handled before building out a lot of extra pages.

- [ ] 404 handling for projects that don't exist
- [ ] standardised `getInitialProps` handling across `_app.js` and page components
- [ ]  switch to express for better dynamic route handling
- [ ] move `_app` and `_document` to the `src` directory so they can be split up, documented and tested (similar to what we're doing with the other pages)
- [ ] fixing up client-side routing
- [ ] remove `.gitkeep`


rogerhutchings:
On second thought and some exploratory work, I'm going to back away from this - we don't need it right now, and can circle back to most of these.

PR #1062, opened 2019-08-02
Allow feedback rules with falsy values

### LINKED ISSUE #1061: Feedback breaks if a rule has ID 0
https://github.com/zooniverse/front-end-monorepo/blob/0e3ae9002a215d465d23e79edd379e4ebb82b262/packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.js#L8-L10
Feedback rules break here if `value` is falsy. Space Warps triggered this by creating subject rules with ID 0.

This was fixed in https://github.com/zooniverse/Panoptes-Front-End/pull/5445 but the fix needs to be copied to the new classifier.

eatyourgreens:
Package:
lib-classifier

Closes #1061.

Copies over the changes to feedback from https://github.com/zooniverse/Panoptes-Front-End/pull/5445 Fixes feedback for rules with an ID of 0 by converting rule values to strings before checking that they exist.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


mcbouslog:
I tested this locally without issue, but now testing on https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/classify it doesn't appear to be showing feedback. I'm going to revert this and take a closer look, apologies for the git noise.

PR #1064, opened 2019-08-06
[MST Upgrade] Update workflow step store

srallen:
Package: lib-classifier

Updates the `WorkflowStepStore` to use `isValidReference` and `types.safeReference`. Updates the tests to be consistent in setup as the others. I've set this to merge into the branch for #1060 because that one has to come first. Once #1060 merges, I'll rebase against #1003 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/WorkflowStepStore.spec.js
+++ b/packages/lib-classifier/src/store/WorkflowStepStore.spec.js
@@ -1,31 +1,51 @@
-import { types } from 'mobx-state-tree'
+import RootStore from './RootStore'
 
 import WorkflowStepStore from './WorkflowStepStore'
 import {
   ClassificationFactory,
   MultipleChoiceTaskFactory,
+  ProjectFactory,
   SingleChoiceTaskFactory,
+  SubjectFactory,
   WorkflowFactory
 } from '../../test/factories'
+import stubPanoptesJs from '../../test/stubPanoptesJs'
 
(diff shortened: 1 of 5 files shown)


packages/lib-classifier/src/store/WorkflowStepStore.spec.js
```diff
@@ -175,106 +183,61 @@ describe('Model > WorkflowStepStore', function () {
...
-        }
-      })
-      ROOT_STORE_INSTANCE.workflowSteps.selectStep('S2')
+      const project = ProjectFactory.build({}, { activeWorkflowId: workflow.id })
+      const panoptesClientStub = stubPanoptesJs({ workflows: workflow, subjects })
+      const rootStore = setupStores(panoptesClientStub, project, workflow)
+      rootStore.classifications.createClassification(subject, workflow, project)
+      rootStore.workflowSteps.selectStep('S2')
       // returns as falsey undefined rather than explicit false
       // this is because usually the workflow has hide_classification_summaries as undefined in the config
```
rogerhutchings: We could explicitly set this to false in the workflow store if undefined

srallen: Yes, I'd like to include that in a separate PR though ultimately with the goal of reverse the logic for how summaries are shown (enable them rather than disable them)

Issue #694, opened 2019-04-10
Staged branch deploys for PR reviews


srallen:
**Package** app-project, lib-classifier?

**Feature or Issue Description**
It would be useful for PR review especially for the design review process (noted in #666)

rogerhutchings:
I've got an idea for how this could work:

1. A PR gets created on GitHub
1. Jenkins picks up the PR, builds a staging docker image which starts up all the apps, and a small proxy for handling routing
1. Image gets deployed to a new kubernetes pod, which runs it all
1. Jenkins comments on the original PR with the staging branch address
1. When the PR is closed, the pod is trashed

Questions / Things to do:

- Write a dockerfile (which can be based on https://github.com/zooniverse/front-end-monorepo/blob/master/Dockerfile) which builds the libraries and apps, starts them up, and starts a proxy with some routing rules
- Do kubernetes pods have individual IPs which we could use instead of having to automate changes to the web ingress, or is there another way we could access them?
- Would totally automating the process be a security concern? On reading this back, I think it could certainly be open to abuse, so we could configure lita to do it, or write a GitHub bot to do it via comments

adammcmaster:
I think we can get this working roughly how you've described. Two things I think we'll need to look into:

1. I'm not sure how easy or hard it might be to get Jenkins to comment on a PR. 
2. I'm also not sure offhand how to get Jenkins to delete the pod when the PR is closed.

I suspect both of those will be possible though.

> Do kubernetes pods have individual IPs which we could use instead of having to automate changes to the web ingress, or is there another way we could access them?

I think we could just add a wildcard domain (similar to how we have `*.preview.zooniverse.org` now) and add that to the ingress, proxying to Nginx or something that can translate the hostname to the internal Kubernetes service name (which can just be a prefix followed by the branch name). Maybe the Kubernetes ingress can do that without even needing Nginx.

> Would totally automating the process be a security concern? On reading this back, I think it could certainly be open to abuse, so we could configure lita to do it, or write a GitHub bot to do it via comments

I think that'll be fine. Jenkins will only execute on PRs from people who have write access to the repo so that will prevent any random person from abusing it.

srallen:
Named stages aren't supported by the Jenkin's Docker plugin: https://github.com/jenkinsci/docker-workflow-plugin/pull/162

There's an open pull request to solve this: https://github.com/jenkinsci/docker-workflow-plugin/pull/180

[...]

PR #1065, opened 2019-08-06
Add ADR 17

srallen:
For #694.

I have two questions from writing this that I still think we should decide upon:
- Should we continue to use https://fe-project.preview.zooniverse.org as the destination sub-domain for the master branch staging deploys? As we add more apps, using `fe-project` doesn't really make sense and won't accurately reflect what's deployed. I believe currently both `fe-project` and `fe-content-pages` are built and deployed on merge to master.
  - Should individual apps be deployed in their own images and process?
- Should we have a staging webpack build for the libraries to not uglify the builds so we can more easily debug using the staging app?

### DIFF
--- a/docs/arch/adr-17.md
+++ b/docs/arch/adr-17.md
@@ -21,7 +21,10 @@ We're going to setup a staging deployment that matches production as closely as 
 - The Jenkins file will be updated to use the git tags to determine the location of the deployment
 - Cloudfront will be configured to load the correct microservice app depending on route:
     - Both the staging (https://frontend.preview.zooniverse.org) and production domains (www.zooniverse.org) will have cloudfront configurations that will match URL traffic against rules setup in Cloudfront.
-    - The rules match paths on the URL, i.e. `/about/team` maps to the correct app via DNS resolution, e.g. `about/team` maps to `fe-content-pages` production app that hits the k8 cluster via DNS. A `GET` request on URL `www.zooniverse.org/about/team` will route like this via Cloudfront: `Cloudfront -> k8 ingress (DNS) -> a fe-content-pages pod`
+    - The cloudfront rules match paths on the incoming URL, i.e. `/about/team` maps to a registered service via DNS, e.g.
+        + When a `GET` request for URL `www.zooniverse.org/about/team` hits cloudfront, it maps to the `fe-content-pages.zooniverse.org` service domain. 
+        + Cloudfront then proxies that request via DNS lookup to the Kubneretes (K8) ingress service
+        + The K8 ingress then looks up the registered service domain and forwards the request, in this case, to a `fe-content-pages` service pod to serve the request and respond to the client. 
     - Generally staging and production would have the same behaviour mappings in Cloudfront and staging will be a place to test these mapping out before setting up in production.
 
 A future enhancement will be added for branch deploys for manual reviews. This can possibly be accomplished by:
--- a/docs/arch/adr-17.md
+++ b/docs/arch/adr-17.md
@@ -1,4 +1,4 @@
-# ADR 16: Monorepo Tooling
+# ADR 17: Staging Deployment
 
 Created: August 6, 2019
 
--- a/docs/arch/adr-17.md
+++ b/docs/arch/adr-17.md
@@ -20,7 +20,7 @@ We're going to setup a staging deployment that matches production as closely as 
 - Production deployment will now be done manually triggered by lita command on slack and using a git tag for production
 - The Jenkins file will be updated to use the git tags to determine the location of the deployment
 - Cloudfront will be configured to load the correct microservice app depending on route:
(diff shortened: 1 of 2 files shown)


docs/arch/adr-17.md
```diff
@@ -0,0 +1,36 @@
...
+## Status
+
+Accepted
+
+## Consequences
+
+- The deployment of the front-end next.js apps will match how the back-end apps are managed so we will now be consistent in process
+- We will need to train developers on the new deployment process
+- We do not have a clear set of steps for setting up the as needed branch deployments at this time
+  - We need to do more research into technically how to accomplish the ad hoc branch deploymenets and tear downs. It is not known if Helm can be used for this.
```
camallen:
```suggestion
  - We need to do more research into technically how to accomplish the ad hoc branch deployments and tear downs. It is not known if Helm can be used for this.
```

adammcmaster:
> What's the reason to move it, clarity of the name and what the domain does? fe doesn't tell me much to be honest and what does project mean in this context considering our domain vocabulary?

"project" refers to the fact that it was set up as the staging environment for the project page app specifically.

frontent.preview.zooniverse.org might be a better name for it.

rogerhutchings:
From talking to Cam this morning:

- In practice, we're going to have two kinds of pull request: one that changes a single app (e.g. new widget on project home page), and one that affects multiple apps (e.g. update to the shared component library)
- For PRs on a single app, we'd like to manually deploy it so it can be tested in isolation. On merging to master, that gets deployed to staging. We'd then do integration testing before manually deploying to production.
- For PRs across multiple apps, we'd test it locally before merging to master. Once it's deployed to staging, we'd do integration testing before manually deploying to production.

Please edit if I've missed anything @camallen

camallen:
Yes that's the idea, avoid deploy the complexity when deploying all apps together combined with an apps feature branch. We can test the changes differently depending on what changes they are, app vs shared code (impacts all apps) and run integration tests (automate some behaviour scripts) via staging systems.

> For PRs on a single app, we'd like to manually deploy it so it can be tested in isolation. On merging to master, that gets deployed to staging. We'd then do integration testing before manually deploying to production.

I think if a PR is only impacting the behaviour of a single app ideally we feature branch deploy just this app (automatically if we can) to test in isolation. This would reduce the complexity of the deploy by only considering the local app state.

We can still test app interactions locally via docker and against other current apps states once a merge to master deploys the app to staging systems.

> For PRs across multiple apps, we'd test it locally before merging to master. Once it's deployed to staging, we'd do integration testing before manually deploying to production.

Test locally in docker, issue a PR, test feature branch deploy works as expected and after merge to master run the integration tests to check the apps work together.

camallen:
I think it’s important that folks dev’ing on the app know how this ‘glue’ works to bind the apps together on a consistent domain target.

Perhaps we can add a section about how cloudfront works as well? Some starting copy..

The Cloudfront configs that wrap the current front end (pfe / monorepo) domain deploy URLs and will split traffic to the correct deployed microservice apps depending on the rules setup in cloudfront.

Both the staging and production domains (www.zooniverse.org) will have cloudfront configurations that will match URL traffic against rules setup in cloudfront.

The rules match paths on the URL, i.e. /about/team maps to the correct app via DNS resolution, e.g. about/team maps to fe-content-pages production app that hits the k8 cluster via DNS.

A GET request on URL `www.zooniverse.org/about/team` will route like this via cloudfront
```
Cloudfront -> k8 ingress (DNS) -> a fe-content-pages pod
```

Generally staging and production would have the same behaviour mappings in cloudfront and staging will be a place to test these mapping out before setting up in production. I can envisage times that staging would more paths i.e. when we add apps.

srallen:
> What's the reason to move it, clarity of the name and what the domain does? fe doesn't tell me much to be honest and what does project mean in this context considering our domain vocabulary?

Yes for clarity. 

> Won't the fe-project solely be the classifier component and pages or will it include other pages (apps?) including the project page landing pages, etc?

Since all the apps are deployed together in one go, then if we have a single sub-domain name and cloudfront setup to route to the correct app, then the current sub-domain doesn't accurately reflect what is being loaded necessarily. I like @adammcmaster suggested sub-domain naming. 

> It seems they are via the jenkins & kubernetes files, am i missing something?

What I mean is that if a PR is only changing the project app, then should only the project-app be deployed upon merge? It looks like this was already discussed in the latter comments :) 

> Can we use source mapping for minified code to handle this instead? Dev tools should handle them by default if defined, https://developers.google.com/web/tools/chrome-devtools/javascript/source-maps

The production webpack configs appear to create them, but I should confirm that. I just know that the uglified javascript is not useful when debugging production. It may be just a matter of tweaking configuration to make sure source maps are functioning as expected.

srallen:
@camallen and @rogerhutchings I've updated the ADR text to include more details about the PR process will be and what Cloudfront's role is. Let me know if this is reading clearly.

PR #1070, opened 2019-08-08
Refactor feedback metadata-to-rules

mcbouslog:
Package: lib-classifier

If these changes make sense, we'll open a new PR (based on #1062), which will in turn take care of #1061.

I merged #1062 and it broke TESS' feedback. I think it's because before merge `metadata-to-rules` would return an array of feedback rule objects, but with the changes TESS' metadata feedback info returned something unexpected. Here is a related screenshot (note `falsy` output is before filtering on Boolean):
Screen Shot 2019-08-07 at 4 27 04 PM

I have a few staging projects with LCV feedback, all of which worked with the changes. However, I didn't realize TESS includes subject metadata for feedback with headers like `feedback_[1]_id`, note the `ruleIndex` is `[1]` as opposed to just `1` or `abc` and I think that may have caused the issue, but I can't pinpoint exactly how or why.

staging project that works on `master` and on `falsy-feedback-values`: http://localhost:8080/?project=1810&workflow=3318
staging project that works on `master` but NOT on `falsy-feedback-values` (similar to TESS): http://localhost:8080/?project=1810&workflow=3351

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.js
+++ b/packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.js
@@ -7,6 +7,9 @@ function metadataToRules (metadata = {}) {
     const stringValue = value.toString()
 
     if (prefix === '#feedback' && stringValue) {
+      if (isNaN(ruleIndex)) {
+        console.error('Subject metadata feedback ruleIndex is improperly formatted')
+      }
       const rule = result[ruleIndex] || {}
       rule[propKey] = value
       result[ruleIndex] = rule
--- a/packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.js
+++ b/packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.js
@@ -8,7 +8,7 @@ function metadataToRules (metadata = {}) {
 
     if (prefix === '#feedback' && stringValue) {
       if (isNaN(ruleIndex)) {
-        console.error('Subject metadata feedback ruleIndex is improperly formatted')
+        console.error(`Subject metadata feedback rule index ${ruleIndex} is improperly formatted. The feedback rule index should be an integer.`)
       }
       const rule = result[ruleIndex] || {}
       rule[propKey] = value
--- a/packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.spec.js
+++ b/packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.spec.js
@@ -5,10 +5,10 @@ describe('feedback: metadataToRules', function () {
   function mockSubjectWithRule (ruleID) {
     return {
       metadata: {
-        '#feedback_1_id': ruleID,
-        '#feedback_1_answer': '0',
-        '#feedback_1_failureMessage': 'Actually, this sound is from noise (background)',
-        '#feedback_1_successMessage': 'Correct!'
+        '#feedback_[1]_id': ruleID,
+        '#feedback_[1]_answer': '0',
+        '#feedback_[1]_failureMessage': 'Actually, this sound is from noise (background)',
+        '#feedback_[1]_successMessage': 'Correct!'
       }
     }
   }
--- a/packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.spec.js
+++ b/packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.spec.js
@@ -60,7 +60,7 @@ describe('feedback: metadataToRules', function () {
       metadataToRules(improperMetadataSubject.metadata)
 
       expect(logError).to.have.been.calledThrice()
-      expect(logError).to.have.been.calledWith('Subject metadata feedback ruleIndex is improperly formatted')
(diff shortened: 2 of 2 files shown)


packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.js
```diff
@@ -15,7 +15,8 @@ function metadataToRules (metadata = {}) {
     return result
   }, [])
 
-  return rules.filter(Boolean)
+  const rulesValues = Object.values(rules)
```
mcbouslog: I don't understand why this works, as I think `rules` is an array, not an object.

packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.spec.js
```diff
@@ -5,10 +5,10 @@ describe('feedback: metadataToRules', function () {
   function mockSubjectWithRule (ruleID) {
     return {
       metadata: {
-        '#feedback_1_id': ruleID,
-        '#feedback_1_answer': '0',
-        '#feedback_1_failureMessage': 'Actually, this sound is from noise (background)',
-        '#feedback_1_successMessage': 'Correct!'
+        '#feedback_[1]_id': ruleID,
```
mcbouslog: I should prob add a comment or warning to explicitly suggest NOT wrapping the `ruleIndex` in brackets. Maybe there should be two sets of similar tests, one set for `ruleIndex` wrapped in brackets one set for it not?

eatyourgreens: The instructions for feedback explicitly state that Nin `feedback_N` should be an integer. https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/features/feedback/README.md

eatyourgreens: Having a spec for the case where N isn’t an integer would be good. Maybe update the spec to test subjects that conform to the rules and also subjects that don’t conform.

srallen:
I've ask the PH: TESS team to take the brackets out for any new sims and sector 13 will be coming soon. I'd say, let's close this PR, reopen the original with the added spec to test for a header that doesn't conform to convention, and when sector 13 is up, do a manual test against the project in production to confirm function.

mcbouslog:
Sounds good on closing this, reopening original PR. I've set this PR to merge into the original `falsy-feedback-values` branch, does this PR "add spec to test for a header that doesn't conform to convention" as desired? Apologies I might not be interpreting that correctly.

srallen:
@mcbouslog yes, it looks like that spec case is covered. Thanks.

srallen:
Sector 13 is now up, so this can be tested against production TESS with the new sim data.

packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.js
srallen: If this is the rules from the store, it's an observable array... which is an object. 🙃 IF this truly needs to be a vanilla array, then you should convert it first, `Array.from(rules)`

srallen: Oh this comment was outdated, so doesn't apply anymore, but yeah that's something to remember about data types wrapped by observables.

mcbouslog:
That's correct @srallen , plan is to merge this into original `falsy-feedback-values` then open a new PR to merge `falsy-feedback-values` into `master`.

eatyourgreens:
I can take another look tomorrow, but this may need to restore the original tests, for feedback that follows the instructions, which were changed and removed.

packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.js
```diff
@@ -7,6 +7,9 @@ function metadataToRules (metadata = {}) {
     const stringValue = value.toString()
 
     if (prefix === '#feedback' && stringValue) {
+      if (isNaN(ruleIndex)) {
+        console.error('Subject metadata feedback ruleIndex is improperly formatted')
```
eatyourgreens:
```suggestion
        console.error(`Subject feedback error: ${ruleIndex} is not a number.`)
```

packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.js
```diff
@@ -7,6 +7,9 @@ function metadataToRules (metadata = {}) {
     const stringValue = value.toString()
 
     if (prefix === '#feedback' && stringValue) {
+      if (isNaN(ruleIndex)) {
+        console.error(`Subject metadata feedback rule index ${ruleIndex} is improperly formatted. The feedback rule index should be an integer.`)
```
eatyourgreens: Do these errors get logged anywhere other than the console? If feedback isn't set up properly, how do we find out?

mcbouslog: There are some [basic validations in the Project Builder](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/features/feedback/lab/helpers/default-validations.js) but nothing that checks the rule index.

PR #1071, opened 2019-08-08
Use dynamic routing

rogerhutchings:
Package: app-project

~~**Requires #1069 merging first**~~

Following on from #1069, this updates the project app to use client side routing. ~~Let's quietly ignore #1038~~

- Moves pages around to use the [new fragment routing system](https://github.com/zeit/next.js/#dynamic-routing)
- Simplify some components, including not passing down props where unnecessary (I swear this should be one of our contributing requirements, given the use of MST / Context)
- Fix tests
- Removing the `server.js` file, since we don't need it anymore! Stuff configured there has been moved to `next.config.js`
- Stub remaining pages for the project app

~~**This is a WIP!** Please feel free to review and whatnot, but there is a breaking change in here - namely routing to the project home page - that needs to be reverted before merge, or it will break production by trying to navigate the client to a page that isn't finished yet. I'm going to fix this by replacing it with a component that forces a refresh, which will pick up the correct page; but in the meantime, I'm leaving it in so that we can test the client side routing in dev.~~

## 🚧 DANGER ZONE 🚧

This will allow the user to navigate to the new project home page, **which is not finished yet**. It should not be deployed until the home page is also ready to go.

If we want to deploy this PR before then, we should change the home page links to use `a` instead of `Link`.


packages/app-project/src/components/ProjectHeader/ProjectHeaderContainer.js
```diff
@@ -9,22 +9,15 @@ import getNavLinks from './helpers/getNavLinks'
 function storeMapper (stores) {
   return {
     isLoggedIn: stores.store.user.isLoggedIn,
-    project: stores.store.project
+    projectName: stores.store.project['display_name']
```
eatyourgreens: Out of interest, why this syntax and not `stores.store.project.display_name`?

packages/app-project/src/components/ProjectHeader/components/Nav/components/NavLink/NavLink.js
```diff
@@ -1,46 +1,50 @@
...
+  &:hover {
     text-decoration: none;
-    border-color: white;
   }
-  ${props =>
-    props.isActive &&
-    `
-    border-color: white;
-  `}
+  &[href]:hover {
```
eatyourgreens: Should be styled for `:focus` too, or are we relying on default styles for focused links?

packages/app-project/src/components/ProjectHeader/components/Nav/components/NavLink/NavLink.spec.js
```diff
@@ -1,31 +1,38 @@
...
   it('should render without crashing', function () {
     expect(wrapper).to.be.ok()
   })
 
   it('should have the correct text', function () {
-    expect(wrapper.text()).to.equal(TEXT)
+    expect(wrapper.text()).to.equal(LINK.text)
+  })
+
+  it(`should have an href if it's not the current page`, function () {
```
eatyourgreens: Maybe put the test context in a describe block:
```js
describe('if not the current page', function () {
  it('should have an href', function () {…})
})
```

packages/app-project/src/components/ProjectHeader/components/Nav/components/NavLink/NavLink.spec.js
```diff
@@ -1,31 +1,38 @@
...
+    expect(wrapper.attr('href')).to.equal(LINK.as)
   })
 
-  it('should have the correct `href`', function () {
-    expect(wrapper.prop('href')).to.equal(HREF)
+  it(`should not have an href if it's the current page`, function () {
```
eatyourgreens: Same here about describing the test context.

packages/app-project/src/components/ProjectHeader/components/ProjectTitle/ProjectTitle.js
```diff
@@ -1,51 +1,51 @@
...
 import styled from 'styled-components'
 
+const StyledHeading = styled(Heading)`
+  text-shadow: 0 2px 2px rgba(0, 0, 0, 0.22);
+`
+
 const StyledAnchor = styled(Anchor)`
   border-bottom: 3px solid transparent;
-  &:focus,
   &:hover {
```
eatyourgreens: Why was the `:focus` styling removed?

packages/app-project/src/components/ProjectHeader/ProjectHeaderContainer.js
rogerhutchings: The linter enforces `camelCase`

packages/app-project/src/components/ProjectHeader/components/ProjectTitle/ProjectTitle.js
rogerhutchings: `Anchor` doesn't have any _additional_ `focus` styling ([src](https://github.com/grommet/grommet/blob/master/src/js/components/Anchor/StyledAnchor.js#L27)) - it inherits from `focusStyle` ([src](https://github.com/grommet/grommet/blob/master/src/js/utils/styles.js#L148)). I was minded to leave it without so we're at least handling focus consistently.

packages/app-project/src/components/ProjectHeader/components/Nav/components/NavLink/NavLink.js
rogerhutchings: That's my thought (see comment below). FWIW I'm not a fan of the big `box-shadow` Grommet uses by default, but equally I'm not sure if the `border-bottom` alone is enough of a visual signifier. Thoughts?

rogerhutchings:
What kind of problems, out of interest? There's been a major version bump in Next (which this PR takes advantage of) while you've been off, so it may require rebuilding your Docker images

packages/app-project/src/components/ProjectHeader/components/Nav/components/NavLink/NavLink.js
rogerhutchings: cc'ing @beckyrother - any thoughts on what focus styles should look like?

eatyourgreens: If this isn’t enough of a visual signifier for focus, then it isn’t enough for hover either. Personally I find the default focus outline  very jarring but that’s just my opinion.

eatyourgreens:
Intermittent 404s loading pages, and the Back button occasionally stopped working between Classify and Home.

packages/app-project/src/components/ProjectHeader/components/ProjectTitle/ProjectTitle.js
eatyourgreens: Good point about consistent focus styling. I don’t feel strongly enough about the Grommet style to change it.

packages/app-project/src/components/ProjectHeader/components/Nav/components/NavLink/NavLink.js
beckyrother: Could someone post screenshots as examples please? I'm also not a fan of the default focus outline but would like to see what Grommet's looks like before making a decision.

rogerhutchings: This is the default Grommet focus, which is what's being used atm

Screenshot 2019-08-16 18 01 22

beckyrother: I don't hate this – maybe if it were possible to add some padding around the text it might not feel so weird?

rogerhutchings: If you don't hate it, then let's leave for another day

packages/app-project/src/components/ProjectHeader/components/Nav/components/NavLink/NavLink.js
```diff
@@ -1,46 +1,52 @@
...
   )
-}
 
-NavLink.propTypes = {
-  href: string.isRequired,
-  router: shape({
-    asPath: string
-  }),
-  text: string.isRequired
+  if (router.pathname === href) {
```
eatyourgreens: Is this checking whether we're on the linked page? If so, maybe make that explicit with something like `const isCurrentPage = router.pathname === href`.

packages/app-project/src/components/ProjectHeader/components/Nav/components/NavLink/NavLink.spec.js
```diff
@@ -1,31 +1,49 @@
...
   it('should render without crashing', function () {
     expect(wrapper).to.be.ok()
   })
 
   it('should have the correct text', function () {
-    expect(wrapper.text()).to.equal(TEXT)
+    expect(wrapper.text()).to.equal(LINK.text)
+  })
+
+  describe('when on the current page', function () {
```
eatyourgreens:
```suggestion
  describe('when not on the current page', function () {
```

packages/app-project/src/components/ProjectHeader/components/Nav/components/NavLink/NavLink.spec.js
```diff
@@ -1,31 +1,49 @@
...
+    it(`should have an href`, function () {
+      expect(wrapper.attr('href')).to.equal(LINK.as)
+    })
   })
 
-  it('should have the correct `href`', function () {
-    expect(wrapper.prop('href')).to.equal(HREF)
+  describe('when not on the current page', function () {
```
eatyourgreens:
```suggestion
  describe('when on the current page', function () {
```

packages/app-project/src/components/ProjectHeader/components/ProjectTitle/ProjectTitle.js
```diff
@@ -1,51 +1,57 @@
...
   )
-}
-
-Title.propTypes = {
-  title: string
-}
 
-Title.defaultProps = {
-  title: ''
+  if (router.pathname === href) {
```
eatyourgreens: Same here for clarifying what this is checking.

packages/app-project/src/components/ProjectHeader/helpers/getNavLinks/getNavLinks.js
```diff
@@ -3,26 +3,31 @@ import counterpart from 'counterpart'
 function getNavLinks (isLoggedIn, baseUrl) {
   const links = [
     {
-      href: `${baseUrl}/about`,
+      as: `${baseUrl}/about`,
+      href: `/projects/[owner]/[project]/about`,
```
eatyourgreens: Maybe link these to PFE for now, until these pages exist? Otherwise, I'm seeing infinite reloading when browsing to these pages.

rogerhutchings: Yeah - so the idea for these was to have a stub for them to navigate to, which then triggers a reload so the route gets picked up by PFE. On the staging environment (if my memory isn't failing me) we have a slightly different CloudFront routing config, so they don't actually fall back.

I'll take a different tack - if a link props object has `as` and `href`, we use `Link`; if it's only `href`, I'll use a regular `a`.

eatyourgreens: Sounds good. I was thinking maybe use full URLs, with the protocol, for external links, and just paths for links that can be handled by the client.

rogerhutchings:
Okay - I've now changed it so links that _can_ be handled client-side do so with a `Link` component, and those that can't use a regular `a` element. The stub pages are still there, so in production they'll 404 (although you should never be able to reach them anyway); in development you get a warning message.

eatyourgreens:
Looks good to me.

PR #1073, opened 2019-08-08
Add error logging middleware

srallen:
Package: app-project

This is an experiment to add middleware to MST store actions to wrap them in a try/catch and then if they fail, to abort the action, and console log and log to sentry. 

The reason for this is that I would've expected to see some errors in the Sentry logs by now and there are none, which I don't believe. I'm wondering if the current setup isn't capturing errors in the MST stores. 

How can we best test this to actually trigger a log to go to Sentry?

If this works, I'd like to create a logger library so then it can be used in the MST stores for the classifier since I don't think those are getting to Sentry either.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


rogerhutchings:
cc #962

packages/app-project/stores/Store.js
```diff
@@ -1,4 +1,5 @@
-import { getEnv, types } from 'mobx-state-tree'
+import { addMiddleware, getEnv, types } from 'mobx-state-tree'
+import { logNodeError } from '../src/helpers/logger'
```
rogerhutchings: fwiw I think I named this function poorly

rogerhutchings:
I _think_ I'm okay with this. I think there are some unexpected side effects that we'll find by aborting the existing call, buuuut no worse than crashing the app which is what will probably happen anyway.

srallen:
@rogerhutchings Only if we have a plan on how to test this on staging. I'm still not sure how to go about that.

rogerhutchings:
Assuming that Sentry is set up on staging...

- Add a new page with a weird URL that can only be routed to from staging (https://frontend.preview.zooniverse.org/projects/test/project/foobar)
- This page has a button that triggers an error in MST
- We merge it, it gets deployed to staging
- Hit the magic error button
- See if we get a Sentry error

packages/app-project/stores/Store.js
eatyourgreens: I’ve renamed it `logToSentry` in #2044.

PR #1056, opened 2019-07-30
React vx bars

srallen:
Package: lib-classifier

Describe your changes:
This is a branch that is a comparison to #1040 but using vx instead of entirely our own React SVG components + d3. It's very similar because I did borrow conceptually from vx, but I very quickly added the bottom axis (took roughly ~10 mins) and swapped out components for vx components. 

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?
- [ ] Is the changelog updated?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?

PR #1074, opened 2019-08-08
[WIP] vx scatter plot viewer

srallen:
Package: lib-classifier

This branches off of #1056 to experiment with a scatter plot viewer that is responsive, shows axis in the same style as the current LCV, and has a configurable zoom (x only, y only, both, or none). If you take a look at the story, if you enable 'zooming' in the knobs, you can zoom in either by double click or mouse wheel. The performance seems pretty bad with the mouse wheel zoom, but ok with double click. These events don't exactly zoom in on the point, but I still need to work that out.

I added also zoom in, out, and reset buttons so you can check out how it might work wired up to the `ImageToolbar` component and these currently work regardless of the `zooming` state. You can change zoom direction in the knob. This takes either, `x`, `y`, `both`, or `none`.

Panning does not work yet. I'd like to make the axis and labels configurable: label via prop and the axis styling with the ticks facing inner or outer by prop as well. Vx supports outer facing tickets out of the box, but I had to use the low level vx `Axis` component to reorient the ticks to face in. This obviously still needs some style tweaking. 

This does not also replicate the annotation action of the original LCV. I haven't conceptually begun to think how to do this yet as I'm focused on creating a scatter plot that can be configurable for the variable star use case. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


rogerhutchings:
As an approach, this looks a lot better than using vanilla D3. Two things occur:

- I'd be tempted to break out some of the setup in `VXLightCurveViewer`
- With the other drawing tools, we agreed on providing a common API using RxJS to handle events. Have you thought about / tried this, or does vx not play well with passing in events manually?

srallen:
> With the other drawing tools, we agreed on providing a common API using RxJS to handle events. Have you thought about / tried this, or does vx not play well with passing in events manually?

Drawing tools for images are setup this way, but not the LCV x-range annotation tool. We could do the same for interactive plot annotation or something else. I just don't know what that is yet. I don't really want to necessarily reuse code between image drawing tools and interactive plot annotation tools since I don't think we can predict if and when divergence needs to happen right now since this is new territory for us.

rogerhutchings:
@srallen What's the plan for this PR, is this still a WIP experiment, shall we get it in a state for merge or shall we trash it?

srallen:
I'm actively writing tests for this to get into a state for actual review. I don't consider it an experiment anymore.

srallen:
With the merge conflicts and size of this, I'll be closing this and breaking it up into smaller PRs.

Issue #806, opened 2019-05-08
How to navigate to a specific workflow? Query param, or URL?


rogerhutchings:
Currently, PFE derives the workflow from the URL query parameter. We've been talking about moving to a regular URL, something like `/owner/project/classify/workflow/[workflow_id]` instead.

I prefer the latter - it's a static resource, which is what a URL is for.

We should to make a decision on this, since it has deployment implications for TESS.

cc @srallen @camallen @eatyourgreens

srallen:
I don't think we need to do this for TESS because they only need one active workflow and the new one will be set as the default workflow. We'll still need to support workflow selection if a user hits a `/classify` route without the workflow specified, unless we're deciding now to no longer support that? We've discussed also moving to a paradigm to prompt and ask a user what to load if the result is error, not found, etc, so we could do that if only the `/classify` route is navigated to, but again, I don't think this is necessary now unless we find there are issues during the production test. 

As an aside, moving to a regular URL means we can also cleanly support subject set selection this way too, `/classify/workflow/:workflow-id/subject-set/:subject-set-id`

rogerhutchings:
`/classify` could redirect to the default workflow?

eatyourgreens:
I like the idea of redirecting so that volunteers can bookmark a specific workflow. This probably ties into our discussion the other day about how to deal with completed workflows too. The URLs should still work but maybe the classifier changes to indicate that the workflow is complete.

srallen:
The redirect could trickle down the selection priorities we have: UPP (if signed in), default, random active. If none work, then prompt the user with a 404 and ask them what to do by showing the available options for the current project or prompt to try another project.

srallen:
CSSI CitSci has a use case to allow volunteers to be able to load a specific subject by url to classify, so I think we can plan on supporting this similarly: `/classify/workflow/:workflow-id/subject-set/:subject-set-id/subject/:subject-id`

rogerhutchings:
Following the discussion in https://github.com/zooniverse/front-end-monorepo/pull/1077#discussion_r313985190, we need to ADR this up first.

PR #1077, opened 2019-08-14
Add Hero widget

### LINKED ISSUE #49: Add project home page
project-home

### LINKED ISSUE #1018: Add project landing page hero component
FireShot Capture 017 - Zooniverse Preview Mode - InVision - projects invisionapp com

rogerhutchings:
Closes #1018
Closes #49 :tada:

- Set default screen size to 'small'
- Update React version to match other apps
- Add Hero component
  - Add background
  - Add about link and intro
  - Add workflow loader (with translated strings!)

## Mobile

Screenshot 2019-08-14 16 41 44

## Desktop

Screenshot 2019-08-14 16 41 34

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/app-project/src/screens/ProjectHomePage/components/Hero/HeroContainer.js
```diff
@@ -0,0 +1,75 @@
...
+      <Hero workflows={workflows} />
+    )
+  }
+}
+
+HeroContainer.propTypes = {
+}
+
+HeroContainer.defaultProps = {
+}
```
srallen: Defined default props and prop types are missing here.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Background/Background.js
```diff
@@ -0,0 +1,39 @@
...
+    min-height: inherit;
+    object-fit: contain;
+    width: 100%;
+  `}
+`
+
+function Background (props) {
+  const { className, backgroundSrc, screenSize } = props
+  return (
+    <Img
```
srallen: This needs an alt text or an `alt=''` for an image that shouldn't be read out to screen readers.

packages/app-project/src/helpers/GrommetWrapper/GrommetWrapperContainer.js
```diff
@@ -25,7 +25,7 @@ class GrommetWrapperContainer extends Component {
     const mergedThemes = this.mergeThemes()
 
     return (
-      <Grommet theme={mergedThemes}>
+      <Grommet userAgent='Mobile' theme={mergedThemes}>
```
srallen: What is this `userAgent` prop doing?

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Background/BackgroundContainer.js
```diff
@@ -0,0 +1,32 @@
...
+class BackgroundContainer extends Component {
+  render() {
+    return (
+      <Background backgroundSrc={this.props.backgroundSrc} />
+    )
+  }
+}
+
+BackgroundContainer.propTypes = {
+  backgroundSrc: string
```
srallen: If there's no default prop for this, it should be set to be required.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Introduction/Introduction.js
```diff
@@ -0,0 +1,45 @@
...
+        />
+      </Link>
+    </Box>
+  )
+}
+
+Introduction.propTypes = {
+  description: string,
+  link: object,
+  title: string,
```
srallen: These should have default props or set to be required.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Introduction/IntroductionContainer.js
```diff
@@ -0,0 +1,58 @@
...
+
+IntroductionContainer.propTypes = {
+  description: string,
+  router: shape({
+    query: shape({
+      owner: string,
+      project: string
+    })
+  }),
+  title: string
```
srallen: Same comment about default props.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/WorkflowSelector.js
```diff
@@ -0,0 +1,70 @@
...
+  )
+}
+
+WorkflowSelector.propTypes = {
+  workflows: shape({
+    data: arrayOf(shape({
+      id: string
+    }))
+  })
+}
```
srallen: Same comment about default props.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.js
```diff
@@ -0,0 +1,37 @@
...
+  )
+}
+
+WorkflowSelectButton.displayName = 'WorkflowSelectButton'
+
+WorkflowSelectButton.propTypes = {
+  workflow: shape({
+    default: bool,
+    displayName: string,
+    id: string
```
srallen: Same comment about default props

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.js
```diff
@@ -0,0 +1,37 @@
...
+
+import theme from './theme'
+
+function WorkflowSelectButton (props) {
+  const { workflow } = props
+  const router = useRouter()
+
+  const as = workflow.default
+    ? `${router.asPath}/classify`
+    : `${router.asPath}/classify/workflow/${workflow.id}`
```
srallen: Woohoo! This is going to make building out workflow selection easier I think having a project home page to actually use for manual testing.

rogerhutchings: I'm actually thinking that all workflows should be routed to `/classify/workflow/${workflow.id}`, and that `/classify` simply does a redirect for consistency. Thoughts / feelings?

packages/app-project/src/helpers/GrommetWrapper/GrommetWrapperContainer.js
rogerhutchings: Setting the default screen size to small, as per the Grommet docs. Without it, it defaults the `screenSize` in `ResponsiveContext` to `medium`, which would jam up on mobile devices with small screens. Also, it'll save an API request on the `RecentTalkSubjects` widget, as it won't get those subjects by default on mobile; it'll adjust once `ResponsiveContext` gets the actual screen size, and then make the request if required.

rogerhutchings:
On `propTypes`, MST's author [recommends using optional props](https://twitter.com/mweststrate/status/809126681646628864); it appears as though components get instantiated before the props can be injected. Otherwise, we set a bunch of default falsey defaults ¯\\_(ツ)_/¯

packages/app-project/src/helpers/GrommetWrapper/GrommetWrapperContainer.js
rogerhutchings: To be fair, we could just trust that user agent detection in Grommet is sensible and request / render based on that

rogerhutchings:
@srallen would you mind taking another look?

packages/app-project/src/helpers/GrommetWrapper/GrommetWrapperContainer.js
srallen: Ok, but this now has the opposite problem, where we see a flash of the small screen layout before switching to the desktop layout when on a desktop large size.

The small screen layout also is showing recent subjects in a carousel now, so saving that network request isn't necessary anymore.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Introduction/Introduction.js
```diff
@@ -0,0 +1,45 @@
...
+  const { description, link, title } = props
+  return (
+    <Box>
+      <Box margin={{ bottom: 'xsmall' }}>
+        <SpacedText weight='bold'>
+          {title}
+        </SpacedText>
+      </Box>
+      <Box margin={{ bottom: 'small' }}>
+        <Text size='xlarge' weight='bold'>
```
srallen: Instead of a `Box` and `Text`, why not just `Paragraph` as that's a block element with margin already and also has semantic meaning?

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/WorkflowSelector.js
```diff
@@ -0,0 +1,70 @@
...
+        {counterpart('WorkflowSelector.message')}
+      </Text>
+
+      {(workflows.loading === asyncStates.error) && (
+        <Box
+          align='center'
+          justify='center'
+          margin={{ top: 'small' }}
+        >
+          There was an error fetching the workflows :(
```
srallen: This needs to be in the translations json.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.js
srallen: Yeah, I'm fine with `/classify` redirecting to `/classify/workflow/:workflow_id` each time, but this should be in an ADR (if it's not already).

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.spec.js
```diff
@@ -0,0 +1,16 @@
...
+
+describe('Component > WorkflowSelectButton', function () {
+  before(function () {
+    wrapper = shallow(<WorkflowSelectButton />)
+  })
+
+  it('should render without crashing', function () {
+    expect(wrapper).to.be.ok()
+  })
+})
```
srallen: There should be tests for the URL that is constructed for the link.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/WorkflowSelector.js
```diff
@@ -0,0 +1,70 @@
...
+        </Box>
+      )}
+
+      {(workflows.loading === asyncStates.success) && (
+        <Box margin={{ top: 'small' }}>
+          {workflows.data.map(workflow =>
+            <WorkflowSelectButton key={workflow.id} workflow={workflow} />
```
srallen: Will the design variants for a couple and a lot of workflow button links come in separate PRs?

packages/app-project/src/screens/ProjectHomePage/components/Hero/helpers/fetchWorkflowsHelper/fetchWorkflowsHelper.js
```diff
@@ -0,0 +1,67 @@
...
+function fetchDisplayNames (language, activeWorkflows) {
+  return panoptes
+    .get('/translations', {
+      fields: 'strings,translated_id',
+      language,
+      translated_id: activeWorkflows.join(','),
+      translated_type: 'workflow'
+    })
+    .then(response => response.body.translations)
+    .then(createDisplayNamesMap)
```
srallen: Both of these requests will have to paginate when there's 20+ workflows (which there are projects that do this). I'm ok with this being updated along with the design for that in a separate PR.

packages/app-project/src/screens/ProjectHomePage/components/Hero/helpers/fetchWorkflowsHelper/fetchWorkflowsHelper.spec.js
```diff
@@ -0,0 +1,72 @@
...
+      })
+
+    const result = await fetchWorkflowsHelper('en', ['1', '2'], '2')
+    expect(result).to.deep.equal([
+      { completeness: 0.4, default: false, id: '1', displayName: 'Foo' },
+      { completeness: 0.7, default: true, id: '2', displayName: 'Bar' }
+    ])
+
+  })
+})
```
srallen: Maybe there should be a test for the default workflow result with the boolean and a test for the error handling?

packages/lib-grommet-theme/src/index.js
```diff
@@ -132,7 +135,14 @@ const theme = deepFreeze({
...
       large: `50px`,
-      xlarge: `90px`
+      xlarge: `90px`,
+
+      'xxsmall-neg': '-5px',
+      'xsmall-neg': `-10px`,
+      'small-neg': `-20px`,
+      'medium-neg': `-30px`,
+      'large-neg': `-50px`,
+      'xlarge-neg': `-90px`
```
srallen: I've published the grommet theme recently because it is being used with the text editor app Will is working on. This PR doesn't need to publish this update, but can you open an issue so we know it needs to be done.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Background/Background.js
```diff
@@ -0,0 +1,40 @@
+import { withResponsiveContext } from '@zooniverse/react-components'
+import { string } from 'prop-types'
+import React from 'react'
+import styled from 'styled-components'
+
+const Img = styled.img`
+  height: 100%;
+  min-height: 61.8vh;
```
srallen: We might want a max height (cc @beckyrother)? It looks a bit big on iPad Pro sizes: 

Screen Shot 2019-09-03 at 2 35 10 PM

packages/app-project/src/screens/ProjectHomePage/components/Hero/HeroContainer.spec.js
```diff
@@ -0,0 +1,23 @@
...
+  })
+
+  it('should render without crashing', function () {
+    expect(wrapper).to.be.ok()
+  })
+
+  it('should render the `Hero` component', function () {
+    expect(componentWrapper).to.have.lengthOf(1)
+  })
+})
```
srallen: Could we have some tests for the `componentDidMount` async request and the expected component states: success and error?

packages/app-project/src/screens/ProjectHomePage/components/Hero/Hero.spec.js
```diff
@@ -0,0 +1,31 @@
...
+
+let wrapper
+
+describe('Component > Hero', function () {
+  before(function () {
+    wrapper = shallow(<Hero screenSize='small' />)
+  })
+
+  it('should render without crashing', function () {
+    console.info(wrapper.debug())
```
srallen: Is this leftover from test writing?

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Background/Background.js
beckyrother: I didn't realize y'all were working on this page!!!!! Really excited, cried a little, this looks great. 

Yes, you're right that there should be a max height. Here's an example where the content is quite tall; it's the max height of the hero image and list of workflows before the workflow list starts to scroll (this should never happen but just in case). The height of the hero image is **763px**. https://projects.invisionapp.com/d/main#/console/12924056/347708839/inspect

There's also a mobile version: https://projects.invisionapp.com/d/main#/console/12924056/344288107/inspect

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.js
eatyourgreens: Sounds good to me but `/classify` should redirect to the stored workflow ID from project preferences for logged-in volunteers, rather than just redirecting to the default workflow.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Background/Background.js
eatyourgreens: Why is the minimum height ~62% of the viewport height? That seems big to me, particularly in portrait layouts. For images, I'd set `min-height` in pixels.

eatyourgreens: You could also use [`align-items: stretch`](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items) if you want the image height to match the height of the box next to it.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.js
rogerhutchings: https://github.com/zooniverse/front-end-monorepo/issues/806

srallen: Yeah, the same ordering of selection would exist: user stored UPP workflow id --> project set workflow id in UPP --> project default --> random active selection. Any one of those would result in the id that gets redirected to.

packages/app-project/src/helpers/GrommetWrapper/GrommetWrapperContainer.js
rogerhutchings: I might try and implement https://github.com/artsy/fresnel with Grommet's breakpoints then

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.js
rogerhutchings: also https://github.com/zooniverse/front-end-monorepo/pull/1107

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Background/Background.js
beckyrother: The most important part of this page is the workflow selection, which is in the module next to the hero image. Thus, the hero image should take up this amount of the screen. For portrait layouts, please use the mobile version attached above.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.js
rogerhutchings: Also - for the purposes of this PR, I propose leaving as is for now

eatyourgreens: I think you can get rid of the ternary here and always route to the workflow ID when linking to a specific workflow (assuming these links are the same as the 'Choose a workflow' links on the current home page.)

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.spec.js
rogerhutchings: Sure. I've changed the implementation to use `withRouter` instead of `useRouter`, as I was unable to mock the router context in Enzyme, either by importing in the provider from Next or providing a context object in the wrapper options ¯\_(ツ)_/¯

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Background/Background.js
rogerhutchings: okay, `max-height` is now 763px on desktop and height is derived from the workflow select box

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.spec.js
srallen: I think the latest version is supposed to support the context API. Did you try this method? https://airbnb.io/enzyme/docs/api/ShallowWrapper/getWrappingComponent.html

rogerhutchings: I did, still came out as null :(

packages/app-project/src/screens/ProjectHomePage/components/Hero/HeroContainer.spec.js
rogerhutchings: Sure - are there any examples of testing async component methods kicking around?

srallen: I don't think we any other components directly making async HTTP calls in their component lifecycle methods, but I think we can take a similar approach as what we do in the store tests and leverage mocha's describe blocks to make sure the setup is correct for each scenario:

1. Stub the actual network request with nock
2. Setup three describe blocks, one for each state of the request: loading, success, and error
3. For the loading scenario, you want to run the expectation synchronously, so instantiate the wrapper instance in the same `it` statement function and run the expectation to check the state
4. For the success and error scenarios, we want to be sure the async call is finished before calling the expectation, so you'll want to instantiate the wrapper in a before or beforeEach function. Then the expectations check the component state is what we expect.

Since I've done so much with testing lately, I'd be happy together a rough example of what I'm describe in a branch off of this one if you'd like to have a fuller explanation of what I'm thinking.

rogerhutchings: That'd be super helpful, actually, I'm just about to get stuck into PRN stuff. I think this is the last thing to fix in this PR aside from the layout stuff I was going to tackle with @artsy/fresnel - mind if I do that in a separate PR?

rogerhutchings: https://github.com/zooniverse/front-end-monorepo/pull/1116

packages/lib-grommet-theme/src/index.js
rogerhutchings: https://github.com/zooniverse/front-end-monorepo/issues/1111

rogerhutchings:
okay, i've added the final set of tests - thanks to @srallen and @eatyourgreens for the help

packages/app-project/src/screens/ProjectHomePage/components/Hero/HeroContainer.spec.js
rogerhutchings: the approach i settled on was waiting for the target method to resolve by spying on it, then testing the child component props for what i was expecting

eatyourgreens: I've also read that you can return a promise from a `before` or `beforeEach` block, in which case mocha will wait on running the tests until those promises resolve.

rogerhutchings:
With #1107 merging, this is going to need to change (specifically the default `/classify` route behaviour), but that can go in a separate PR.

beckyrother:
@rogerhutchings the > is quite large, any way to adjust so it's the same height as the 'learn more' text? Other than that design looks great

packages/app-project/src/screens/ProjectHomePage/components/Hero/HeroContainer.spec.js
```diff
@@ -0,0 +1,166 @@
...
+    let scope
+    let wrapper
+    let componentWrapper
+    let fetchWorkflowsSpy
+
+    before(function () {
+      scope = nock('https://panoptes-staging.zooniverse.org/api')
+        .get('/translations')
+        .query(true)
+        .reply(200, {
```
srallen: Non-blocking, but if returning an error, just for consistency in mocking, this probably shouldn't be a 200 status.

srallen:
@beckyrother this will be ready for a design review when this merges and deploys to our staging instance (branch deployments are still WIP)

beckyrother:
Sorry. Please tag me when it's ready.

srallen:
@beckyrother no worries! I thought you might want to check it out as an actual app and not just screenshots.

srallen:
I've opened https://github.com/zooniverse/front-end-monorepo/issues/1133 to discuss the `Media` component use.

packages/app-project/src/screens/ProjectHomePage/components/Hero/HeroContainer.spec.js
rogerhutchings: Only the workflows endpoint is returning an error in this test, but both requests need to be successful in order to show the content (as currently written)

PR #1078, opened 2019-08-14
[WIP] Add drawing tools point, line

mcbouslog:
Package: lib-classifier

Describe your changes:
- refactors SingleImageViewer, InteractionLayerContainer for drawing
- changes InteractionLayer to pointer events (previously mouse events)
- adds DrawingContainer for current workflow step drawing marks (active and inactive)
- adds drawing helpers
- adds drawing tools Point and Line (and a Root component)

See inline comments and questions.

Actively working on tests, refactoring; opening as WIP for feedback on general approach...

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/helpers/withCoordsFromStream.js
```diff
@@ -0,0 +1,115 @@
...
+      this.props.storeMark(coordinates)
+    }
+
+    render () {
+      const { ...props } = this.props
+      const { activeCoordinates } = this.state
+
+      return (
+        <WrappedComponent
+          coordinates={activeCoordinates}
```
mcbouslog: `activeCoordinates` is an array of simplified event objects that consist of properties `x`, `y`, and the event type (i.e. `pointerdown`, `pointermove`, `pointerup`). 

The idea is that each drawing tool has unique functionality and knows when to conclude drawing, so let's send the minimum amount of event data and a callback to finish drawing to the drawing tool component, then the drawing tool component will do what it needs to do with the simplified events and let parent components/containers know when that particular drawing tool mark has been completed.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/tools/Root.js
```diff
@@ -0,0 +1,31 @@
...
+  const mainStyle = {
+    color: tool.color,
+    fill: 'transparent',
+    pointerEvents: 'none',
+    stroke: tool.color,
+    strokeWidth: active ? SELECTED_STROKE_WIDTH : STROKE_WIDTH
+  }
+
+  return (
+    <g {...mainStyle} tabIndex='-1'>
```
mcbouslog: I think delete mark functionality can start from a handler/callback here and go back to the DrawingContainer

eatyourgreens: Maybe add `focusable` here too to allow keyboard interaction in browsers that follow the original SVG spec and ignore `tabindex`. I'm thinking Edge but my knowledge of SVG is patchy and could well be wrong. 

Links may be the only way to reliably and easily add keyboard interaction across browsers, since SVG doesn't have buttons. From memory, that's how keyboard interaction is handled in the SVG account stats ribbon on the Zooniverse home page.

eatyourgreens:
Pedantic point, I know, but this goes beyond refactoring since it changes components from using mouse events to using pointer events.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/helpers/withCoordsFromStream.js
```diff
@@ -0,0 +1,115 @@
...
+      }
+    }
+
+    startDrawing () {
+      if (this.props.active) {
+        const { eventStream } = this.props
+        this.firstSubscription = eventStream.subscribe(event => {
+          if (event.type === 'pointerdown') {
+            this.secondSubscription = eventStream.subscribe(event => {
+              const svgCoordinateEvent = this.convertEvent(event)
```
rogerhutchings: You could use `map` here directly on the event stream - https://rxjs-dev.firebaseapp.com/api/operators/map

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/helpers/withCoordsFromStream.js
```diff
@@ -0,0 +1,115 @@
...
+        this.secondSubscription.unsubscribe()
+      }
+    }
+
+    startDrawing () {
+      if (this.props.active) {
+        const { eventStream } = this.props
+        this.firstSubscription = eventStream.subscribe(event => {
+          if (event.type === 'pointerdown') {
+            this.secondSubscription = eventStream.subscribe(event => {
```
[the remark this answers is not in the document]
rogerhutchings: I think something like this? https://medium.com/@paynoattn/3-common-mistakes-i-see-people-use-in-rx-and-the-observable-pattern-ba55fee3d031#ad2b

Rather than creating new streams, using `switchMap` on the existing stream to get what we want

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/DrawingContainer.js
```diff
@@ -0,0 +1,120 @@
...
+    newMarks.set(storeMark.id, storeMark)
+
+    this.setState({ activeMark: newMark, marks: newMarks })
+  }
+
+  render () {
+    const { activeMark, marks } = this.state
+    if (!activeMark) return null
+
+    const MarkComponent = getDrawingTool(activeMark.tool.type)
```
rogerhutchings: There's no error handling here in case we can't get the correct tool type

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/tools/Point.js
```diff
@@ -0,0 +1,75 @@
...
+  large: 20,
+  small: 10
+}
+const CROSSHAIR_SPACE = 0.2
+const CROSSHAIR_WIDTH = 1
+
+const Point = ({ active, coordinates, finishDrawing, scale, tool }) => {
+  if (coordinates.length === 0) return null
+
+  const currentCoordinates = coordinates[(coordinates.length - 1)]
```
rogerhutchings: `Array.prototype.pop()`?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/helpers/withCoordsFromStream.js
rogerhutchings: Does this mean that the drawing tool itself doesn't receive a stream, but a reference to an array held in state as a prop?

If so, it might be an idea to simply give the drawing tool its simplified stream (again, `switchMap` is likely your friend), as `setState` calls are batched, I can imagine lots of re-renders happening, and the drawing tool code itself could be simplified by using a stream.

rogerhutchings:
I've left some comments - I'm sorry, I really wanted to give this more time before I went on holiday - but I think it'd be really useful to have a README or similar with a reasonably high-level description of your approach as well to help piece it together

mcbouslog:
No worries @rogerhutchings! Lots of great comments, def agree on README. We'll keep at it, I'm sure there'll be more opportunity for help when you get back. Enjoy the holiday!

mcbouslog:
Replacing with smaller PR

PR #1080, opened 2019-08-16
[MST Upgrade] Additional store fixes

srallen:
Package: lib-classifier

This fixes a few issues found when removing all of the test runner scopes and tries to run the whole classifier test suite:
- The `FieldGuide` store had a major issue with its store setup that resulted in an infinite loop in attempting to request for the field guide. I wasn't able to determine a specific cause, but the solution I tried that worked was refactoring the store setup to use `applySnapshot` to setup the active project dependency and to only test the `FieldGuideStore`. The result of this is that we're no longer testing that `afterAttach` works because `afterAttach` in the MST lifecycle only fires if the store is attached to a parent store. 
- The failing tests for `SubjectStore` that @eatyourgreens first mentioned in some past PRs. The strange thing about this one is that if you did scope the tests to run to only the `SubjectStore` tests or to the specific test, they would pass. They only failed if you weren't scoping the test run at all. I made a couple of tweaks which involved moving one test into its own describe block because the setup of the previous one was causing the subsequent one to fail since they were sharing the same `rootStore` instance (why this didn't fail when scoping the tests to that describe block is a mystery to me). The other fix was just changing how many times the while loop runs.
- I've added some additional `if (processbrowser)` checks around console logging. 
- I've added a valid active reference check in the `WorkflowStore` that I missed first go around. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -77,7 +89,7 @@ describe('Model > SubjectStore', function () {
         })
 
         it('should request more subjects', function () {
-          while (rootStore.subjects.resources.size > MINIMUM_QUEUE_SIZE) {
+          while (rootStore.subjects.resources.size >= MINIMUM_QUEUE_SIZE) {
```
eatyourgreens: This should break the tests by triggering two requests to refill the queue, instead of one. One request from inside this while loop, then another when the queue advances on the next line. 

The while loop here should reduce the queue down to the max size, if it’s larger, or do nothing if the queue is already at or below the limit. The next line is then guaranteed to advance the queue and trigger a refill.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -39,22 +39,34 @@ describe('Model > SubjectStore', function () {
...
         initialSize = rootStore.subjects.resources.size
         rootStore.subjects.advance()
       })
 
       it('should reduce the queue size by one', function () {
         expect(rootStore.subjects.resources.size).to.equal(initialSize - 1)
       })
+    })
+
+    describe('setting an active subject', function () {
```
eatyourgreens: This is a bit of a hack, since it duplicates the previous describe block. Can you see why the tests fail if we have more than one expectation inside a describe block? 

For instance, this block of tests will fail if we add an expectation that advancing the queue should reduce the queue size by one. 

I also found that with two expectations inside a describe block, the second expectation always fails even if I swapped the order of the two.

eatyourgreens: The description here doesn’t really describe the test, which is to call advance on a full queue of subjects (same as the previous describe block.) Setting the active subject is an expectation of what should happen.

packages/lib-classifier/src/store/FieldGuideStore.spec.js
```diff
@@ -35,62 +36,53 @@ const fieldGuideWithoutIcon = FieldGuideFactory.build({
...
-  function fetchFieldGuide (rootStore) {
-    sinon.stub(rootStore.fieldGuide, 'fetchFieldGuide')
-    rootStore.projects.setResource(project)
-    return rootStore.projects.setActive(project.id)
-      .then(() => {
-        rootStore.fieldGuide.fetchFieldGuide.restore()
-        return rootStore.fieldGuide.fetchFieldGuide()
-      })
-  }
+function setupStores(clientStub) {
```
eatyourgreens: This needs a comment to explain why the mocks for this store are set up differently from the other stores.

packages/lib-classifier/src/store/FieldGuideStore.spec.js
```diff
@@ -35,62 +36,53 @@ const fieldGuideWithoutIcon = FieldGuideFactory.build({
...
+  return store
+}
 
-  function setupStores (clientStub) {
-    const store = RootStore.create({
-      classifications: {},
-      dataVisAnnotating: {},
-      drawing: {},
-      feedback: {},
-      subjects: {},
```
eatyourgreens: Should the original code have attached a projects store here? Is that the bug that broke the tests?

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: If I run the subject queue tests in isolation, they fail because of the extra advance call added here:
```
14 passing (1s)
  1 failing

  1) Model > SubjectStore
       Actions > advance
         with less than three subjects in the queue
           when the initial response has ten subjects
             should request more subjects:
     expected populateQueue to have been called exactly twice, but it was called thrice
    populateQueue() => [Promise] {  } at populateQueue (/usr/src/packages/lib-classifier/src/store/SubjectStore.js:55:16)
    populateQueue() => [Promise] {  } at Object.populateQueue (/usr/src/packages/lib-classifier/src/store/SubjectStore.js:110:14)
    populateQueue() => [Promise] {  } at Object.populateQueue (/usr/src/packages/lib-classifier/src/store/SubjectStore.js:110:14)
  AssertionError: expected populateQueue to have been called exactly twice, but it was called thrice
      populateQueue() => [Promise] {  } at populateQueue (src/store/SubjectStore.js:55:16)
      populateQueue() => [Promise] {  } at Object.populateQueue (src/store/SubjectStore.js:110:14)
      populateQueue() => [Promise] {  } at Object.populateQueue (src/store/SubjectStore.js:110:14)
      at Context.calledTwice (src/store/SubjectStore.spec.js:97:65)
```

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens:
```suggestion
    describe('with a full queue', function () {
```

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens:
```suggestion
          while (rootStore.subjects.resources.size > MINIMUM_QUEUE_SIZE) {
```

eatyourgreens: Logging the queue size inside the while loop gives different results if the tests are run in isolation or as part of the full test suite:

With `.only()` advancing the queue reduces the size by one, as expected:
```
9
8
7
6
5
4
3
Fetching more subjects
2
Fetching more subjects
          1) should request more subjects
```

Without `.only()` the queue has to be advanced twice before the subject changes:
```
9
9
8
8
7
7
6
6
5
5
4
4
3
3
Fetching more subjects
2
```

eatyourgreens: On further investigation, the workflow store tests define `rootStore` as a global variable, which is then interfering with the store state in other tests. Removing that global will probably fix the subject store tests.
https://github.com/zooniverse/front-end-monorepo/blob/3e09df2017e02b0ee1359c2a73e65ed2513fe442/packages/lib-classifier/src/store/WorkflowStore.spec.js#L10-L17

packages/lib-classifier/src/store/FieldGuideStore.spec.js
srallen: No, a `ProjectStore` was created on the `RootStore.create` call here. It just wasn't getting set to an empty object as a snapshot like the others because it needed to be set with an actual project for the field guide tests. I still don't understand what changed or broke between the update and the final merging of all the updates.

packages/lib-classifier/src/store/SubjectStore.spec.js
srallen: Good catch. I'll take a closer look.

eatyourgreens: I spent a couple of hours looking at this today but didn’t get very far. All I figured was that tests fail if you call `advance` more than once, and then only if run in conjunction with the other store tests. 

The workaround seems to be only call advance once after setting up the mocks.

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -39,22 +39,34 @@ describe('Model > SubjectStore', function () {
...
+      let rootStore
+      let previousSubject
+
+      before(function () {
+        rootStore = setupStores()
+      })
+
+      beforeEach(function () {
+        previousSubject = rootStore.subjects.resources.values().next().value.id
```
eatyourgreens: Calling `.next()` advances an iterator. I wonder if that is causing problems since the `advance()` method also advances the `values()` Iterator?

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: #1083 fixes one set of breaking tests for the subject store. I've tried removing/renaming `rootStore` and `setupStores` in the PR, in case they were conflicting with other stores, but advancing the queue still fails if it's called more than once.

eatyourgreens: I think I've found the bug. If I remove the `process.browser` check from the feedback middleware, then here's what this test logs.
```
ADVANCING
9
Aborting subject advance and showing feedback
9
ADVANCING
8
Aborting subject advance and showing feedback
8
ADVANCING
7
Aborting subject advance and showing feedback
7
ADVANCING
6
Aborting subject advance and showing feedback
6
ADVANCING
5
Aborting subject advance and showing feedback
5
ADVANCING
4
Aborting subject advance and showing feedback
4
ADVANCING
3
finished loop
Aborting subject advance and showing feedback
--------------
```

eatyourgreens: The feedback store is aborting the subject advance action on every other call.

eatyourgreens: Feedback is active and has 2 rules set in the mock store. I'm going to guess that the Factory classes are singletons, and that an earlier test enabled feedback on the mock project and workflow?

packages/lib-classifier/src/store/FieldGuideStore.spec.js
eatyourgreens: These fail when I run them with `.only` to isolate them so I think they've always been broken. We just haven't noticed because Travis isn't running the tests for any of these PRs.

srallen: They were working in the update PR for the FieldGuideStore, and I run tests locally before pushing anything up, so I don't know what changed, but between all the updates, scoping, and rebasing, I think it's not really worth the time to exactly pinpoint it.

eatyourgreens: Something to consider for the MST update would be adding the target branches to the Travis config, so that CI runs for those branches as well as master.

Maybe also change the docker image name in the compose files, so that this update builds and uses a different image from master. That might save some time with building the repo again in order to test these branches.

srallen:
Closing to reopen a couple of separate PRs for the remaining changes.

PR #1083, opened 2019-08-20
Add specs for subject queue advance

eatyourgreens:
Package:
lib-classifier

This adds a couple of missing specs for `subjectStore.advance()`. Advancing the queue should:
- make the next subject in the queue active.
- remove the current active subject.
- reduce the size of the queue by one.
- change the active subject.

I've also removed `values().next()` in favour of using `subjectStore.active` to get the active subject, but tests are still failing and I can't see why.

`rootStore` is removed fro the subject store tests, and `setupStores` is renamed, to avoid name collisions with other store tests.

Tests will pass if the subject store tests are run in isolation.

EDIT: tests pass if `rootStore.subjects.advance()` is only called once so I've edited some of the tests to do that.

UPDATE: Feedback helpers were stubbed to enable feedback. Restoring the helpers has (hopefully) fixed tests that failed because feedback aborted the subject queue advance.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/FeedbackStore.spec.js
+++ b/packages/lib-classifier/src/store/FeedbackStore.spec.js
@@ -58,6 +58,11 @@ describe('Model > FeedbackStore', function () {
       reducer: sinon.stub().callsFake(rule => rule)
     }
   })
+
+  after(function () {
+    helpers.isFeedbackActive.restore()
+    helpers.generateRules.restore()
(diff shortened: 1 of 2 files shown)


eatyourgreens:
So, tests are failing because feedback is cancelling the advance action on every other call.

PR #1085, opened 2019-08-20
[MST Upgrade] Update WorkflowStore and tests

srallen:
Package: lib-classifier

Fixes some remaining items that got missed in a previous pass:
- Use `isValidReference` when checking the active project before using it for the default workflow.
- Remove the `rootStore` global variable in the tests.

Some space changes crept when I copied and pasted from my old PR. I'll fix that with the linter in the branch it is merging into all in one go.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/WorkflowStore.spec.js
+++ b/packages/lib-classifier/src/store/WorkflowStore.spec.js
@@ -7,8 +7,6 @@ import {
 } from '../../test/factories'
 import stubPanoptesJs from '../../test/stubPanoptesJs'
 
-let rootStore
-
 const workflow = WorkflowFactory.build({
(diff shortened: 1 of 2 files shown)


packages/lib-classifier/src/store/WorkflowStore.spec.js
```diff
@@ -7,8 +7,6 @@ import {
 } from '../../test/factories'
 import stubPanoptesJs from '../../test/stubPanoptesJs'
 
-let rootStore
-
 const workflow = WorkflowFactory.build({
```
eatyourgreens: Not a blocking comment for this PR, but we should be careful in future to scope variables like this one to the tests that use them. This should be declared inside the callback function that uses it.

Issue #874, opened 2019-05-23
Feedback storing task ID and empty array to classification metadata


mcbouslog:
**Package**
lib-classifier

**Feature or Issue Description**
Feedback is storing task ID and an empty array to classification metadata even if feedback is not relevant to the classification.

![Screen Shot 2019-05-23 at 11 40 22 AM](https://user-images.githubusercontent.com/12118751/58270717-14b87780-7d50-11e9-998c-d637af3d6627.png)

In the FeedbackStore [`createClassificationObserver`](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/FeedbackStore.js#L42) is updating feedback regardless of if feedback is active or not. Which then [sets a task rule to an empty array](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/FeedbackStore.js#L115).

Possible solution - adding a check for `self.isActive` before (or within) the `onAction` in the `createClassificationObserver`.

Even with solution ^, `metadata.feedback = {}` is always in metadata due to the [Classification model](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/Classification.js#L7). Is that ok or something to also fix?

Also, I want to triple-check how this may affect the TESS-MAST pipeline (@zwolf).

mcbouslog:
Mentioning @eatyourgreens, @srallen

srallen:
I don't think this has any affect on the Caesar pipeline. The pipeline is using subject metadata to determine decisions on reducer types to use, not classification metadata.

srallen:
> adding a check for self.isActive before (or within) the onAction in the createClassificationObserver.

This seems reasonable to me since `feedback: { T1: [] }` does sort of make it look like feedback was meant to be shown and something wrong happened. I'm fine with `feedback: {}` in the metadata as a default value.

eatyourgreens:
Why is this an issue? Is it breaking something? There's a solution listed but I can't see the problem that's being solved.

srallen:
I think there was a concern yesterday that it might cause one, but I don't believe it is breaking anything. However, I would be for making the suggested update so that an empty array isn't added if feedback wasn't actually shown.

eatyourgreens:
Looking back at the code, feedback should only be added to the classification if it's active and has rules (I think.) It might need an additional check that `rules` isn't empty.
[...]

PR #1095, opened 2019-08-22
Feedback bug fixes

srallen:
Package: lib-classifier

For #874 (maybe closes?)

Describe your changes:
- Updates the conditional to check that the feedback rules sizes is greater than 0 before updating the classification metadata with feedback information. 
- Updates the feedback store `update` action to only update the feedback rules if feedback is active
- Adds specs for both

The feedback store tests could use refactoring, however, I'm trying to minimize the merge conflict that this will cause with the MST updates (and subsequent test updates that caused). So this is really only scoped to test for these bug fixes.

There may still be a scenario where `isActive` is true, but there are no feedback rules because the subject is a real subject and has no feedback rule subject metadata, so `isActive` shouldn't be true, but is. I've seen this on first load of a real subject on TESS, so it isn't a bug with the reset action. It is likely a bug with one of the helpers that determine whether or not to set `isActive` to true in the store. However, we haven't reliably and consistently replicated this yet although it has been observed. See slack discussion on it.

This PR, though, will at least prevent adding a `T1: []` feedback metadata to the classification metadata when the subject is real and `isActive` is false. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/FeedbackStore.js
+++ b/packages/lib-classifier/src/store/FeedbackStore.js
@@ -90,7 +90,6 @@ const FeedbackStore = types
       const workflow = getRoot(self).workflows.active
 
       self.isActive = helpers.isFeedbackActive(project, subject, workflow)
-
       if (self.isActive) {
         self.rules = helpers.generateRules(subject, workflow)
       }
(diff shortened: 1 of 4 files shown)


packages/lib-classifier/src/store/FeedbackStore.js
```diff
@@ -106,13 +105,15 @@ const FeedbackStore = types
...
-      const { task, value } = annotation
-      const taskRules = self.rules.get(task) || []
-      const updatedTaskRules = taskRules.map(rule => {
-        const ruleReducer = strategies[rule.strategy].reducer
-        return ruleReducer(rule, value)
-      })
-      self.rules.set(task, updatedTaskRules)
+      if (self.isActive) {
+        const { task, value } = annotation
+        const taskRules = self.rules.get(task) || []
```
eatyourgreens: Maybe the task rules should default to `undefined` instead of `[]`? I'm fairly certain that trying to set an undefined key on an object doesn't create a new key.

PR #1002, opened 2019-07-09
Upgrade MobX and MST

srallen:
Package: lib-classifier

Describe your changes:
The base WIP MobX/MST upgrade branch. Fixes for this minor version update will be in separate PRs and this one will merge into master when it's ready. These are not the latest versions, but the versions when breaking changes happen. 

#1003 should be reviewed first, then the PRs updating the individual stores. 

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?
- [ ] Is the changelog updated?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?

PR #1096, opened 2019-08-23
Add subjects.append(newSubjects)

eatyourgreens:
Package:
lib-classifier

Adds an `append` action to the subject store, plus a couple of specs.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/SubjectStore.js
+++ b/packages/lib-classifier/src/store/SubjectStore.js
@@ -111,6 +111,17 @@ const SubjectStore = types
       }
     }
 
+    function append (newSubjects) {
+      newSubjects.forEach(subject => {
+        self.resources.put(subject)
+      })
+
+      const validSubjectReference = isValidReference(() => self.active)
+      if (!validSubjectReference) {
+        self.advance()
+      }
+    }
+
     function * populateQueue () {
       const root = getRoot(self)
       const client = root.client.panoptes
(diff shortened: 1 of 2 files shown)


packages/lib-classifier/src/store/SubjectStore.js
```diff
@@ -111,6 +111,17 @@ const SubjectStore = types
       }
     }
 
+    function append (newSubjects) {
+      newSubjects.forEach(subject => {
```
eatyourgreens: This fails if I change it to
```js
newSubjects.forEach(self.resources.put)
```
and I can't see why. `self.resources.merge(newSubjects)` also throws errors, so I've left the loop in.

eatyourgreens:
I've based this on the MobX upgrade in #1002 to avoid merge conflicts with the store updates there.

packages/lib-classifier/src/store/SubjectStore.js
srallen: `merge` expects an array of entries, so `[['1', { location: 'foo.jpg' }], ['2', { location: 'bar.jpg' }]]` or an object. I'm not sure why calling put directly would fail either.

eatyourgreens: Maybe `self.resources` isn't a map at runtime, so the `put` method doesn't exist yet? I'm not sure either.

eatyourgreens: I had another look at this and `newSubjects.forEach(self.resources.put)` fails because `self.resources` isn't a valid MST node.

PR #1097, opened 2019-08-23
Talk widget: Add an image carousel

eatyourgreens:
Add a carousel view for recent Talk subjects at small screen sizes, plus spec.

Package:
app-project

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsCarousel.js
+++ b/packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsCarousel.js
@@ -11,6 +11,7 @@ function RecentSubjectsCarousel (props) {
   const width = 700
   return (
     <Carousel
+      className={className}
       controls='arrows'
       fill
     >
--- a/packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsCarousel.js
+++ b/packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsCarousel.js
@@ -1,0 +1,39 @@
+import counterpart from 'counterpart'
+import { arrayOf, shape, string } from 'prop-types'
+import React from 'react'
+import { Carousel } from 'grommet'
+import SubjectThumbnail from './components/SubjectThumbnail'
+
+// TODO: Use the subject viewers from the classifier
+function RecentSubjectsCarousel (props) {
+  const { className, href, subjects } = props
+  const height = 500
+  const width = 700
+  return (
+    <Carousel
+      controls='arrows'
+      fill
+    >
+      {subjects.map( subject => (
+        <SubjectThumbnail
+          key={subject.id}
+          height={height}
+          href={href}
+          subject={subject}
+          width={width}
+        />
+      ))}
+    </Carousel>
+  )
+}
+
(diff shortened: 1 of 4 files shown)


eatyourgreens:
Here's what it looks like on Galaxy Zoo. The only interaction with the image is to change to the next image left/right. There's no support for tapping a link through to another page.
Screenshot showing the image carousel on the Galaxy Zoo home page at small screen sizes.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsCarousel.js
```diff
@@ -0,0 +1,39 @@
...
+import SubjectThumbnail from './components/SubjectThumbnail'
+
+// TODO: Use the subject viewers from the classifier
+function RecentSubjectsCarousel (props) {
+  const { className, href, subjects } = props
+  const height = 500
+  const width = 700
+  return (
+    <Carousel
+      controls='arrows'
```
eatyourgreens: `controls='arrows'` isn't supported in a public release of Grommet yet. See https://github.com/grommet/grommet/pull/3176

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsCarousel.js
```diff
@@ -0,0 +1,39 @@
+import counterpart from 'counterpart'
+import { arrayOf, shape, string } from 'prop-types'
+import React from 'react'
+import { Carousel } from 'grommet'
+import SubjectThumbnail from './components/SubjectThumbnail'
+
+// TODO: Use the subject viewers from the classifier
+function RecentSubjectsCarousel (props) {
+  const { className, href, subjects } = props
```
srallen: `className` isn't being used here. It should maybe be applied to the `Carousel` since that's the parent?

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsCarousel.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+RecentSubjectsCarousel.propTypes = {
+  subjects: arrayOf(shape({
+    id: string
+  }))
+}
+
+RecentSubjectsCarousel.defaultProps = {
+}
```
srallen: Should we have default props here? Or set some to be required?

eatyourgreens:
@srallen do you think it's ok on phones that the subject doesn't link through to its Talk page? That's my only reservation about the carousel.

packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjectsCarousel.js
eatyourgreens: `subjects` should be required, I think, and `className` should default to undefined. `href` I'm not sure about because the links don't actually work in the Grommet carousel.

srallen:
@eatyourgreens yes, I think that's ok. The call to action button is close by and trying to click a link on a carousel can be annoying and hard to do (I meant to click a link, but hit the button to rotate the image or vice versa).

PR #1099, opened 2019-08-28
Allow feedback rules with falsy values 

### LINKED ISSUE #1061: Feedback breaks if a rule has ID 0

eatyourgreens:
Package:
lib-classifier

Closes #1061.

Copies over the changes to feedback from zooniverse/Panoptes-Front-End#5445
Allows feedback with a rule ID of 0 (or other rule keys with falsy values) by converting rule values to strings before checking that they exist.

Also adds a spec and a check for subject feedback where the rule index is not a number.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/feedback/helpers/generate-rules.spec.js
+++ b/packages/lib-classifier/src/store/feedback/helpers/generate-rules.spec.js
@@ -81,4 +81,42 @@ describe('feedback: generateRules', function () {
     }
     testSubjectAndWorkflow(subject, workflow)
   })
+
+  describe('with no matching rules', function () {
+    describe('when the workflow rule ID is truthy', function () {
+      const workflow = {
+        tasks: {
+          T0: mockTaskWithRule('1')
+        }
+      }
+
+      it('should return an empty object for a falsy subject rule ID', function () {
+        const subject = mockSubjectWithRule(0)
+        expect(generateRules(subject, workflow)).to.be.empty()
+      })
+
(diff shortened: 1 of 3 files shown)


mcbouslog:
Noting that currently I'm frequently seeing the bug where the feedback store `isActive = true` on load when it shouldn't be (real subject, not sim with feedback) on zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/classify, and want to check again after this merges. If this doesn't fix it, can create new issue.

srallen:
@mcbouslog could you open an issue just for `isActive` being true incorrectly? I have my doubts that this PR has fixed that...

eatyourgreens:
If you could clarify how this PR might have affected `isActive`, that would be great. I’m not seeing it myself.

srallen:
I don't think it has.... I think we're just seeing a separate bug from close scrutiny of the feedback functionality.

eatyourgreens:
#1096 should make it easier to test expected behaviour after adding new subjects and advancing the queue.

PR #1102, opened 2019-09-02
WIP Start stats page

rogerhutchings:
**Requires #1071 merging first**

- Add the stats page
- Add the `ByTheNumbers` component
- Move `AnimatedNumber` out of the `Stat` component so it can show other data
- Compose `AnimatedNumber` with `Stat` for existing numerical stats

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
I've based this against #1071 to make the diff easier to review.

packages/app-project/src/screens/StatsPage/StatsPage.spec.js
```diff
@@ -0,0 +1,21 @@
...
+  })
+
+  it('should render without crashing', function () {
+    expect(wrapper).to.be.ok()
+  })
+
+  it('should render the ByTheNumbers component', function () {
+    expect(wrapper.find(ByTheNumbers)).to.have.lengthOf(1)
+  })
+})
```
eatyourgreens: It should render a heading with the project title too.

packages/app-project/src/screens/StatsPage/components/ByTheNumbers/ByTheNumbers.js
```diff
@@ -0,0 +1,62 @@
...
+  classifications: number.isRequired,
+  completedSubjects: number.isRequired,
+  launchDate: string,
+  volunteers: number.isRequired
+}
+
+ByTheNumbers.defaultProps = {
+  classifications: 0,
+  completedSubjects: 0,
+  launchDate: null,
```
eatyourgreens: `null` isn't a string. Should this be an empty string by default?

packages/app-project/src/screens/StatsPage/components/ByTheNumbers/ByTheNumbers.spec.js
```diff
@@ -0,0 +1,20 @@
...
+      <ByTheNumbers
+        projectName='A test project'
+      />
+    )
+  })
+
+  it('should render without crashing', function () {
+    expect(wrapper).to.be.ok()
+  })
+})
```
eatyourgreens: This should at least test the required props.

packages/app-project/src/screens/StatsPage/components/ByTheNumbers/ByTheNumbersContainer.js
```diff
@@ -0,0 +1,72 @@
...
+
+    // TODO: Use luxon's `.setLocale()` method for doing date translations. This
+    // requires `full-icu` to be included, and I'm going to hold off on that
+    // until https://github.com/nodejs/node/issues/19214 is merged so we have
+    // access to Docker images including it; the docs call it a faff and we
+    // don't have UI translations atm anyway.
+    // See https://moment.github.io/luxon/docs/manual/formatting.html
+    if (launchDate) {
+      return DateTime.fromISO(launchDate)
+        .toLocaleString(DateTime.DATE_FULL)
```
eatyourgreens: The first argument passed to `toLocalString()` should be the locale.

If we're localising dates, then numbers should be localised with eg. `toLocaleString('en')` too (if they aren't already.)

packages/app-project/src/screens/StatsPage/components/ByTheNumbers/ByTheNumbers.js
rogerhutchings: Doesn't really matter, it's not required, so `null` is a permissible value. It can be an empty string if you prefer

packages/app-project/src/screens/StatsPage/components/ByTheNumbers/ByTheNumbersContainer.js
rogerhutchings: nope, this is a luxon datetime object, so it gets set with `setLocale`

good point on localising numbers, but we'd need to hold off for the same reason (no built-in intl support in node yet)

packages/app-project/src/screens/StatsPage/components/ByTheNumbers/ByTheNumbers.js
eatyourgreens: I was confused because the type is set as a string (I think.) `null` is an object.

packages/app-project/src/screens/StatsPage/components/ByTheNumbers/ByTheNumbersContainer.js
eatyourgreens: Is that using the browser locale by default? That won't be the same as the locale being used by the page.

rogerhutchings: it is at the moment, but we don't have a language setting in the store yet, so that's going to have to be hooked up in a future PR

eatyourgreens: True. Also that store is a per-user session store?

rogerhutchings: The UI store seems to be the logical place to put it - where the value gets populated from is totally up for grabs (cookie? query param? UPP? etc)

eatyourgreens: I was just wondering if the store is a single variable in memory on the server, or if we have a store for each browser session that's currently making requests. Same problem as storing the session theme, I guess.

eatyourgreens: This has been on my mind while considering how auth might work on the server.

eatyourgreens: I'm using `.toLocaleDateString()` in Node without any problems on #1128 so you might not need luxon here.

eatyourgreens:
This was based off #1071 so needs rebasing to account for the changes on that PR.

srallen:
@rogerhutchings given the merge conflict and the WIP status, is it ok to close this for now?

rogerhutchings:
sure

PR #1107, opened 2019-09-04
ADR-18 - Workflow routing

rogerhutchings:
Following chat in https://github.com/zooniverse/front-end-monorepo/pull/1077#discussion_r313985190, this adds an ADR for how we handle workflow routing. This was done in a bit of a rush, so comments welcome

cc #806


rogerhutchings:
So with the caveat that there's going to be an unavoidable full-page refresh in navigating between Talk and Classify, we're going to need to start storing some MST data on the browser. I'll add an edit to reflect this, but with PHT being a single workflow project it wouldn't need to be done at the same time as the PR that implements this change - sound okay?

srallen:
I agree, that's a separate issue. One where we would leverage stashing a snapshot of the store in the browser's local storage or indexedDB and load it back up when users navigate between Talk and the classify page.

eatyourgreens:
The store could be used to get the current workflow for all classifiers, with user prefs to persist settings across login sessions. That would greatly simplify the logic.

docs/arch/adr-18.md
```diff
@@ -0,0 +1,47 @@
+# ADR 18: Workflow routing
+
+Created: September 4, 2019
+
+## Context
+
+On PFE, there are a couple of different ways workflows are routed:
+
+- Navigating to `/classify`, the user sees either the default workflow for that project, or the workflow stored in project preferences for logged-in volunteers.
```
srallen: or random active workflow

docs/arch/adr-18.md
```diff
@@ -0,0 +1,47 @@
...
+- Navigating to `/classify?workflow=[workflow_id]` shows the workflow determined by the query parameter.
```
srallen: If enabled for the project or if the user has the correct role: owner, collaborator, tester, or admin

docs/arch/adr-18.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Status
+
+Proposed
+
+## Consequences
+
+- `/classify` will still have different effects depending on the user, but solely in terms of redirects rather than content.
+- We'll have to start storing at least portions of the MST state object in the browser, in order to allow for users to navigate back and forth between Talk and the classifier transparently.
+- Workflow access restriction can be managed at the workflow component file level.
```
srallen: We need to support auth server side so we can request for the user roles so we can determine if a user has permission to load a workflow in certain cases, i.e. project owners, collabs, testers and admins are allowed to load inactive workflows for testing purposes.

rogerhutchings: It'd make for a nicer UX, but I don't think it's essential. The classifier load is already deferred until the client bundle has loaded, so it's not a huge change to check the user status and its permissions for that project before allowing it to load.

If we _did_ have it, however, it would be a perfect fit for using the correct HTTP status code with a [custom `_error.js`](https://github.com/zeit/next.js/#custom-error-handling).

docs/arch/adr-18.md
```diff
@@ -0,0 +1,47 @@
...
+
+### `/classify` behaviour
+
+When navigating to `/classify`, the page should redirect the user to the correct workflow URL in the following priority:
+
+1. Workflow ID set in user project preferences by the user
+1. Workflow ID set in user project preferences by the project
+1. Project default workflow
+1. Random active selection
+
```
eatyourgreens: The page should redirect to the current workflow stored in the MST store.

That workflow should be stored on user login, and should be set according to the priorities listed here. I'm not sure if initialising a volunteer's workflow store falls under this ADR, but it's good to have the logic for workflow selection laid out like this.

srallen: Yeah, definitely. We can have that described in a separate ADR about workflow selection and priority as I think this is specifically about routing.

shaunanoordin:
🎏 Just to flag a quirk about the _WildCam Lab_ Assignments:
- WildCam Lab Assignments have URLs that look like... `https://www.zooniverse.org/projects/zooniverse/wildcam-gorongosa/classify?workflow=8074&classroom=1`
- The `classroom=true` flag exists to override the standard "can a user select this workflow?" rules.
  - For example, WildCam Gorongosa (the project) doesn't actually allow workflow workflows to be selected by the user. (Presumably because if it did, it might end up auto-listing a few hundred potential workflows/classrooms on the WildCam Gorongosa Zooniverse.org project page.)
  - Going to `https://www.zooniverse.org/projects/zooniverse/wildcam-gorongosa/classify?workflow=8074`, without the classroom flag, would instead _silently_ load the default workflow 338 instead. (The URL still shows 8074, which is confusing AF.)

Whatever the shape of the new workflow routing looks like, WildCam Labs can be modified to adopt it. However, this WildCam Labs scenario does show an edge case where an _"I know what I'm doing, I really want to select this workflow, for reals"_ flag might be needed.

docs/arch/adr-18.md
```diff
@@ -0,0 +1,47 @@
...
+The new URL that we redirect the user to should replace the entry for `/classify` in the browser history.
```
shaunanoordin: Query: how descriptive will the new URL be?

For example, let's say we have Project 'A Space Odyssey' with Workflow 2001 (default).

If I type in `zooniverse.org/projects/kubrick/a-space-odyssey/classify`, then (assuming no project preferences) I'd be redirected to/the browser history will rewrite itself to `zooniverse.org/projects/kubrick/a-space-odyssey/classify/2001` - correct?

I just wanted to check that it's this straightforward, because given the above statement...

> ... we currently have a requirement for CSSI CitSci to be able to route to specific subjects ...
> ...
> We adopt the following URL structure:
> `/projects/[owner]/[project]/classify/workflow/:workflow-id/subject-set/:subject-set-id/subject/:subject-id`

... I wasn't entirely sure if `/classify` would instead redirect me to `zooniverse.org/projects/kubrick/a-space-odyssey/classify/2001/subject-set/3000/subject/44444`, with the URL changing every time I'm served a new Subject. (Because that'd make bookmarking my current workflow a bit tricky.)

rogerhutchings: I'm assuming (but @srallen should confirm) that per-subject URLs won't be enabled for every project. Otherwise, yes, workflow bookmarking would be of less use - I guess you could pick up from that subject, and you'd get a banner saying you've already classified it and asking you to move on.

Another thing I've just thought of is where on the URL chain we handle redirects. We could cascade them: `classify` could redirect the user to `workflow`, which could _then_ redirect to `subject_set` etc. It would keep the code simple, but result in multiple redirects which would be a sub-par user experience. 

Alternatively, `classify` could bump you straight off to `/classify/2001/subject-set/3000/subject/44444` - more complex, but only one redirect.

This is an implementation issue that's outside the ADR, but worth thinking about

shaunanoordin:
Query: does this ADR plan to address what happens when a user types in an invalid workflow choice?

What I'm reading:
- this ADR explains what happens what to do when a user visits the Classifier _with no specific workflow set,_ e.g. `zooniverse.org/projects/marvel/avengers/classify`
- i.e. they should be redirected to/the browser history rewritten to _a specific workflow_ (usually the default, or a workflow stored in the proj pref): e.g. `zooniverse.org/projects/marvel/avengers/classify/workflow/101`

My assumption:
- since we're talking about how we're allowing users to bookmark by workflow, we're _also_ saying that a user can just _type in a specific workflow into the Classifier,_ e.g. `zooniverse.org/projects/marvel/avengers/classify/workflow/103`, and the Classifier will select workflow 103. (Assuming it's a valid workflow.)

My question:
-  what are our actions when the user specifies an invalid workflow? e.g. if `zooniverse.org/projects/marvel/avengers/classify/workflow/404` doesn't exist?
- will the user be sent to the default workflow?
  - (As stated in [comments above](https://github.com/zooniverse/front-end-monorepo/pull/1107#issuecomment-529482251) , this comes up as a bit of a problem if the default workflow is quietly selected instead of the specified workflow.)
- or will the user see an error message _"Sorry, that workflow does not exist and/or you're not allowed to see that workflow, pls go to the project home page and try again"_ ?
  - I'm very much in preference for this latter option.

docs/arch/adr-18.md
```diff
@@ -0,0 +1,47 @@
...
+
+This would be facilitated by the [dynamic routing feature in Next.js 9](https://github.com/zeit/next.js/#dynamic-routing) (see [#1071](https://github.com/zooniverse/front-end-monorepo/pull/1071)).
+
+Workflow links from the home page would route directly to their workflow-specific URLs.
+
+### `/classify` behaviour
+
+When navigating to `/classify`, the page should redirect the user to the correct workflow URL in the following priority:
+
+1. Workflow ID set in user project preferences by the user
```
shaunanoordin: This is probably a paranoia question, but: do we have plans for what happens if the Workflow ID saved in the project preferences becomes invalid? e.g. that WF gets switched off by the project owner.

e.g. last week I was working on WF 666, but today the project team shut it off, because they have a new default workflow 777. (And because the old workflow kept killing all the exorcists hired to debug it.)

- Will the _"when navigating to /classify, decide where to redirect to"_ logic be the one responsible for determining that WF 666 in the user's project prefs is no longer valid, and then moving on to the next priority step?
- or will the user be greeted with a general _"sorry, invalid workflow"_ error page, preferably one with a _"if you still have trouble, pls click this to reset your user preferences"_ button?

Sorry for the slight paranoia, but I was recalling how we had reports of a user who was unable to access the Classify page of a specific project _on one single Zooniverse account,_ but was able to classify just fine anonymously or on a second Zooniverse account. We spent a very long while trying to figure out if the auto-redirect-based-on-user-project-preferences logic was the problem in that user's case.

PS: small typo, all priorities are step `1.`

rogerhutchings:
Yeah, I don't want any magic happening either. We should treat it like any other web resource; if you don't have access to a workflow due to permissions, you get a 401 / 403 error (depending on login status) for example. We can do some nice stuff with the error page depending on the error code though.

docs/arch/adr-18.md
srallen: The initial implementation will only be for the citsci.org use case. They will generate a link in the UI of their app that the volunteer can click and load a specific subject to classify on. I don't think we need to keep other projects from being able to load specific subject sets and subjects to classify on, but we won't be advertising or building a UI for it on our site (maybe eventually, but this won't be part of CSSI). For this reason, I don't think we should generally be redirecting all projects to specific subject classification URLs as we don't know what the UI and potential side effects will be. I think the only redirect we support should be for `/classify/workflows/:workflow-id`

As for what the URL behavior should be for the citsci.org case as the volunteer is classifying? I think we can address that (with another additional ADR if necessary) when we start to build out that functionality.

srallen:
I'm supportive of an error message if you do not have permission to load that workflow with UI asking what you'd like to do. I think noting that is within scope of this ADR.

docs/arch/adr-18.md
srallen: This is a multi-part question and also not entirely in scope of this ADR, but within the scope of one about workflow selection. I believe the behavior of what the UPP store does will remain the same: we don't know why the project inactivated the workflow, just that they did, so all we can do is clear it out of the UPP so that the app doesn't attempt to load this workflow every time they visit that project. 

In that scenario, what the UI does is a different question and the same as what to do when the workflow is invalid generally. I think we should show an error and ask them what they want to do. This will be the change in behavior from PFE that this ADR should note.

srallen:
Re the classroom experience:

> The classroom=true flag exists to override the standard "can a user select this workflow?" rules.

We will likely have to continue doing something like this, but I'd like to consider that when we fully begin implementing workflow selection strategies and note in an ADR for that specifically.

eatyourgreens:
A lot of the conversation here is about how a workflow should be selected when the classify page first loads, if one isn't specified in the URL. At the moment, the classifier component handles selecting and loading a workflow. Are we proposing that the classifier component should be able to redirect the URL of the parent page, which requires it to have some knowledge of its parent and the window location? Or are we proposing that the responsibility for workflow selection should be moved out of the classifier component, so that the parent loads a workflow then passes it into the classifier?

rogerhutchings:
Good point. I'm strongly in favour of passing in a `workflow_id` prop. If we're trying to create a reusable classifier, then we shouldn't be adding in a load of features that are only used in PFE.

eatyourgreens:
The model I've been leaning towards with PFE is something like:
```jsx
<WorkflowSelector>
  <Classifier/>
</WorkflowSelector>
```

or `withWorkflow(Classifier)`.

See https://github.com/zooniverse/Panoptes-Front-End/pull/5318 for an example.

srallen:
I don't think workflow selection strategy should be in a component at all, classifier, HOC, or a parent. I think it should be a set of functions called by the workflow store. We could support a workflow id prop too like we do for projects.

docs/arch/adr-18.md
rogerhutchings: Noted in 9e140f4

rogerhutchings:
@srallen and I also chatted briefly about spinning up a small API to handle workflow selection.

rogerhutchings:
I think everything's been covered that needs to be in the ADR, with a future ADR to cover exactly how we do workflow selection. Is this now good to merge?

rogerhutchings:
closes #806

docs/arch/adr-18.md
```diff
@@ -0,0 +1,51 @@
...
+
+The new URL that we redirect the user to should replace the entry for `/classify` in the browser history.
+
+### Error handling
+
+When a user is unable to view a workflow at a given URL, we should provide an error page at that URL. Next.js allows you to provide a response code in `getInitialProps`, which we can use to tailor the message we show on a custom error component, so for non-existent workflows, `404`; incorrect permissions get a `401`/`403` (depending on login status) etc.
+
+## Status
+
+Proposed
```
rogerhutchings:
```suggestion
Accepted
```

PR #1108, opened 2019-09-05
Add a valid reference check to subjects.append()

eatyourgreens:
Check self.resources is valid before using it.

Package:
lib-classifier

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/SubjectStore.js
+++ b/packages/lib-classifier/src/store/SubjectStore.js
@@ -112,13 +112,14 @@ const SubjectStore = types
     }
 
     function append (newSubjects) {
-      newSubjects.forEach(subject => {
-        self.resources.put(subject)
-      })
+      const validResources = isValidReference(() => self.resources)
+      if (validResources) {
+        newSubjects.forEach(self.resources.put)
 
-      const validSubjectReference = isValidReference(() => self.active)
-      if (!validSubjectReference) {
-        self.advance()
+        const validSubjectReference = isValidReference(() => self.active)
+        if (!validSubjectReference) {
+          self.advance()
+        }
       }
     }
 


packages/lib-classifier/src/store/SubjectStore.js
```diff
@@ -112,13 +112,14 @@ const SubjectStore = types
     }
 
     function append (newSubjects) {
-      newSubjects.forEach(subject => {
-        self.resources.put(subject)
-      })
+      const validResources = isValidReference(() => self.resources)
```
srallen: `isValidReference` is for checking `types.safeReference` or `types.reference` types (see: https://github.com/mobxjs/mobx-state-tree#reference-validation-isvalidreference-tryreference-oninvalidated-hook-and-typessafereference). `resources` are `types.map` (which has a built in `types.optional` with a default), so this isn't necessary and might not work? 

Instead, I think you can:

```
if (self.resources && self.resources.size > 0) {
  // do stuff with self.resources
}
```

eatyourgreens: Looking at this again, I think `self.resources.put(subject)` was failing the `isStateTreeNode` check here so maybe `subject` is invalid? I'm not sure. https://github.com/mobxjs/mobx-state-tree/blob/9121976541e5392a5364c3cdf3d046f7f0165b8d/packages/mobx-state-tree/src/types/complex-types/map.ts#L172

eatyourgreens: To examine the error, run `newSubjects.forEach(self.resources.put)` outside the `if` block.

eatyourgreens: Checking the map size is greater than 0 will stop us adding subjects to an empty queue.

srallen: Doh, you're right. `if (self.resources)` should be enough to check for existence then.

eatyourgreens: Hopefully, but the error is thrown inside `self.resources.put` so there may need to be an additional check that it’s a valid tree node.

eatyourgreens:
@srallen The requested changes throw the error that this PR is fixing:
```shell
backend.js:1 Error: [mobx-state-tree] Value undefined is no MST Node
    at fail$1 (mobx-state-tree.module.js:2904)
    at getStateTreeNode (mobx-state-tree.module.js:2741)
    at MSTMap.put (mobx-state-tree.module.js:3621)
    at Array.forEach (<anonymous>)
    at Object.append (SubjectStore.js:160)
    at executeAction$$1 (mobx.module.js:658)
    at res (mobx.module.js:649)
    at runNextMiddleware (mobx-state-tree.module.js:2417)
    at next (mobx-state-tree.module.js:2430)
    at handler (mobx-state-tree.module.js:2269)
```

The line that's erroring is this one:
https://github.com/mobxjs/mobx-state-tree/blob/9121976541e5392a5364c3cdf3d046f7f0165b8d/packages/mobx-state-tree/src/types/complex-types/map.ts#L187

and the code that triggers the error is:
```js
function append (newSubjects) {
      if (self.resources) {
        newSubjects.forEach(self.resources.put)

        const validSubjectReference = isValidReference(() => self.active)
        if (!validSubjectReference) {
          self.advance()
        }
      }
    }
```

So `self.resources` exists but `this` is undefined inside `put`?

packages/lib-classifier/src/store/SubjectStore.js
eatyourgreens: From those docs,
>In addition to the options possible for a plain reference type, the optional options parameter object also accepts a parameter named acceptsUndefined, which is set to true by default, so it is suitable for model properties. When used inside collections (arrays/maps) it is recommended to set this option to false so it can't take undefined as value, which is usually the desired in those cases.

So I think arrays and maps should be updated as per the example there eg. 
```js
resources: types.map(types.safeReference(Subject, { acceptsUndefined: false }))
```
I don't think that will fix the error encountered here but it seems like good practice.

rogerhutchings:
`this` defaults to the `window` object inside `forEach`, and I'm guessing that the result of `getStateTreeNode(window)` is `undefined`.

`newSubjects.forEach(newSubject => self.resources.put(newSubject))` uses an arrow function, so you should have the right context then

eatyourgreens:
So why does adding to the resources map work in this PR, where `self.resources.put` is passed as the callback?

eatyourgreens:
I think the CI changes may have got lost in a rebase.

PR #1109, opened 2019-09-06
Add a multi-stage Docker build for production apps

eatyourgreens:
Package:
lib-classifier
app-project
app-content-pages

- Split the docker build into `bootstrap`, which installs the monorepo and builds dependencies, and `production-apps`, which builds the NextJS apps.
- Remove the NextJS builds from the bootstrap script.
- update the docker-compose setup to use the appropriate named target for each local docker image. `front-end-monorepo_dev` builds from the `bootstrap` stage. `front-end-monorepo_prod` builds from the `production-apps` stage.
- rename the test service to yarn, since it runs yarn commands, and remove local mounted volumes when testing the production set-up. If local files change, you'll need to run `docker-compose build` to test the changes in a new production image.
- Update the Jenkins build to build `zooniverse/front-end-monorepo` with the bootstrapped monorepo, then build and start the NextJS apps in their respective images.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM node:10
+FROM node:10 AS bootstrap
 
 RUN mkdir -p /usr/src
 WORKDIR /usr/src/
 
-COPY ./ .
+COPY ./ /usr/src
 RUN chown -R node:node .
 
 USER node
@@ -11,3 +11,14 @@ USER node
 RUN yarn install
 
 RUN yarn bootstrap
+
+FROM node:10 AS production-apps
+
+RUN mkdir -p /usr/src
+WORKDIR /usr/src/
+
+COPY --from=bootstrap /usr/src /usr/src
+
+RUN yarn workspace @zooniverse/fe-project build
+
+RUN yarn workspace @zooniverse/fe-content-pages build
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -1,13 +1,24 @@
...
 RUN yarn install
 
 RUN yarn bootstrap
+
+FROM node:10 AS production-apps
+
+RUN mkdir -p /usr/src
+WORKDIR /usr/src/
+
+COPY --from=bootstrap /usr/src /usr/src
```
eatyourgreens: Ideally, this step should only copy over the files that are needed to run the production apps, not the entire source tree, tests etc.

eatyourgreens:
Speed, really, but I think I need feedback from people who have used to this to find out if it’s useful.

eatyourgreens:
Bumping this because it's almost a month old and should fix the duplicate production builds on Travis CI.

rogerhutchings:
Just for reference, a first build with `--no-cache` took a brutal 23 minutes 😲 but it'll do that on master too. subsequently starting the project app was a bit snappier

PR #1110, opened 2019-09-06
Change default from null to types.model

srallen:
Package: lib-classifier

Describe your changes:
Changes the switch default from `null` to `types.model({})` so that the classifier doesn't crash with Kitteh Zoo anymore. Its drawing task uses an ellipse which we don't have a model for yet. 

We might want to change line 16 in the workflow step store to do the same?

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/tasks/DrawingTask.js
+++ b/packages/lib-classifier/src/store/tasks/DrawingTask.js
@@ -15,7 +15,7 @@ const Drawing = types.model('Drawing', {
         case 'point':
           return Point
         default:
-          return null
+          return types.model({})
       }
     }
   })),


shaunanoordin:
## Additional Notes

package: `lib-classifer`

This PR fixes an issue where the "I Fancy Cats" project crashes on load, because given its workflow's tasks of `{ T1, T2, init }`, (note how the first task is listed last) `T2` is a drawing task with a _not-yet-supported ellipse tool,_ and processing that Task _also_ causes the `init` Task to fail to process.

(This was confusing since it looked like the first task was crashing the app.)

This fix allows the default I Fancy Cats project to run on `lib-classifer` on `http://localhost:8080/?project=335` in a limited manner:
- classifier loads correctly now, instead of crashing
- first task (`init`), a simple question, works fine
- second task (`T1`), a simple question, works fine
- third task (`T2`), a drawing task with the unsupported `ellipse tool`, _crashses the app_ into a white screen.

IIRC, the previous behaviour is that the unsupported third task used to simply show a Task that did nothing, but still allowed the Classification to be submitted.

For now this fix works fine since the monorepo is still in development, 👌 but eventually a more robust _"what to do with Tasks or Tools that aren't yet supported"_ solution might help us handle  projects that are oddly configured.

PR #1114, opened 2019-09-10
Test feedback on subject change

eatyourgreens:
Add an onNewSubject action to the feedback store and add tests when feedback is and isn't active. Rearrange the feedback store tests so that global stubs are only set for the action tests.

Package:
lib-classifier

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ node_js:
 branches:
   only:
     - master
+    - mobx-upgrade
 
 cache: yarn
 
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Working on this, I noticed that the workflow task rules are reduced for each subject that loads, even though they produce the same rules each time. That could be improved by running `getFeedbackFromTasks` once for the workflow and caching the result. https://github.com/zooniverse/front-end-monorepo/blob/8005598971e9123023c2337df3e4822e7df849ba/packages/lib-classifier/src/store/feedback/helpers/get-feedback-from-tasks.js#L3-L15

The feedback model could also be improved by making `isActive` a computed value, rather than setting it on subject change.

eatyourgreens:
I've had a go at simplifying `getFeedbackFromTasks` in #1115, assuming that we don't mind if it breaks in IE.

srallen:
>  assuming that we don't mind if it breaks in IE.

Nope, we don't.

PR #1115, opened 2019-09-10
Refactor getFeedbackFromTasks

eatyourgreens:
Refactor getFeedbackFromTasks for legibility.
Remove lodash in favour of Object.entries(tasks) to iterate over workflow tasks and create a feedback rules object.

Package:
lib-classifier

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/lib-classifier/src/store/feedback/helpers/get-feedback-from-tasks.js
```diff
@@ -1,17 +1,18 @@
...
-    if (_.get(task, 'feedback.enabled', false) &&
-      _.get(task, 'feedback.rules', []).length > 0) {
-      return _.assign({}, result, {
-        [taskId]: task.feedback.rules
-      })
-    } else {
-      return result
+function getFeedbackFromTasks (tasks = {}) {
+  const result = {}
+  Object.entries(tasks).forEach(function getTaskRules ([taskKey, task]) {
```
eatyourgreens: If we want to support IE11, then this should be changed to use `Object.keys()`.

srallen: We're not supporting IE11 in the rebuild: https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-4.md

srallen: And fwiw, Microsoft is dropping legacy support for it in 2020 and will make Edge available for Win 7 users to download.

packages/lib-classifier/src/store/feedback/helpers/get-feedback-from-tasks.js
```diff
@@ -1,17 +1,18 @@
...
+    const { feedback } = task
+    const taskHasFeedback =
+      feedback &&
+      feedback.enabled &&
+      feedback.rules.length > 0
```
srallen: Maybe `feedback.rules && feedback.rules.length > 0`?

eatyourgreens: I've updated to add a check for `rules` and a default value of `{ enabled: false, rules: [] }` for the feedback object.

PR #1116, opened 2019-09-10
Make Hero widget responsive with SSR

rogerhutchings:
- Install @artsy/fresnel
- Move grommet theme merger to its own helper
- Add fresnel helper components
- Add media queries to Hero widget
- Update tests

Now there's no flashes of wrong layout :tada:

The way this works is by rendering the HTML for _each_ component tree, and providing a media query to choose which one to show in the browser. That meant updating tests to check for multiple
instances of a component, which may look a little weird.

This technique could also be used on the project header, site header, or anywhere else where we're rendering different component trees for different sizes.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


packages/app-project/src/helpers/theme/README.md
```diff
@@ -0,0 +1,5 @@
+# Theme
+
+Creates a Grommet theme by performing a deep merge of the Grommet base theme with the Zooniverse custom Grommet theme.
+
+This was originally done in the `GrommetWrapper` theme, but was ,oved here so it can be reused with `@artsy/fresnel` to extract the breakpoints for responsive sizing.
```
eatyourgreens:
```suggestion
This was originally done in the `GrommetWrapper` theme, but was moved here so it can be reused with `@artsy/fresnel` to extract the breakpoints for responsive sizing.
```

packages/app-project/src/screens/ProjectHomePage/components/Hero/Hero.spec.js
```diff
@@ -18,14 +18,14 @@ describe('Component > Hero', function () {
   })
 
   it('should render the `Background` component', function () {
-    expect(wrapper.find(Background)).to.have.lengthOf(1)
+    expect(wrapper.find(Background)).to.have.lengthOf(2)
```
eatyourgreens: This would be better if it could test the two layouts separately, and expect each layout to have a single background, introduction and workflow selector.

eatyourgreens: Maybe find the `Media` components by their `at` props and run tests on each for background, introduction and workflow selector? Something like tests for linked media here:
https://github.com/zooniverse/front-end-monorepo/blob/757a9c48efd75e7ea11af147304b751b603882e7/packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/components/SubjectThumbnail/SubjectThumbnail.spec.js#L30-L35

Issue #1112, opened 2019-09-09
Possible additional bugs with feedback on TESS


srallen:
**Package** lib-classifier

**Feature or Issue Description**
 One user reported recently getting a sim but not getting the feedback message: https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/talk/2110/1111008

We also still have an issue where `isActive` in the feedback state is remaining or initializing as true for real subjects:

![Screen Shot 2019-09-09 at 9 26 25 AM](https://user-images.githubusercontent.com/5016731/64539386-02a11b80-d2e4-11e9-8615-91dcd05e4309.png)

When the feedback state is in this incorrect state, we still get inaccurate metadata sent with the classification:

![Screen Shot 2019-09-09 at 9 28 55 AM](https://user-images.githubusercontent.com/5016731/64539558-501d8880-d2e4-11e9-983e-3b1c1e0b2bfa.png)

Which will cause unnecessary caesar extractions.

It is unclear if these are related.

srallen:
I saw `[mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[FeedbackStore Subject Observer autorun]' TypeError: Cannot read property 'toString' of null` on the branch with the mobx upgrade, which prevented the feedback from displaying and did not add the classification metadata correctly. I'm not sure if this is happenig on master yet, but this might be the bug we're looking for.

eatyourgreens:
That might be thrown by `generateRules` if feedback is active for a subject that doesn't have feedback.

srallen:
Yes, I'm going to spend a bit of time testing this on master to confirm the issues are there and not on the mobx upgrade branch (which I happen to be testing for final review).

srallen:
There is a subject metadata value that is sometimes `null` and the `metadata-to-rules` calls `toString` on the metadata values. `null.toString()` will throw too, so that needs to be checked for too.

srallen:
I've confirmed that's where the above MST error is coming from, but I'm going to write checks for both.

PR #1117, opened 2019-09-10
Check id existence before calling toString

srallen:
Package: lib-classifier

Related to #1112. Checks for the existence of the subject rule id and workflow rule id before calling `toString` on them. 

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/feedback/helpers/generate-rules.js
+++ b/packages/lib-classifier/src/store/feedback/helpers/generate-rules.js
@@ -24,11 +24,15 @@ function generateRules (subject, workflow) {
   _.forEach(workflowRules, (rules, taskId) => {
     const taskRules = _.reduce(rules, (result, workflowRule) => {
       const matchingSubjectRule = _.find(subjectRules, (subjectRule) => {
-        if (subjectRule.id && workflowRule.id) {
+        if ((subjectRule.id !== null && subjectRule.id !== undefined) &&
+          (workflowRule.id !== null && workflowRule.id !== undefined)) {
           return subjectRule.id.toString() === workflowRule.id.toString()
+        } else {
+          if (process.browser) {
+            console.warn('Feedback: Subject rule id or workflow rule id is null or undefined')
+          }
+          return false
         }
-
-        return null
       })
 
       if (matchingSubjectRule) {
--- a/packages/lib-classifier/src/store/feedback/helpers/generate-rules.js
+++ b/packages/lib-classifier/src/store/feedback/helpers/generate-rules.js
@@ -23,8 +23,13 @@ function generateRules (subject, workflow) {
 
   _.forEach(workflowRules, (rules, taskId) => {
     const taskRules = _.reduce(rules, (result, workflowRule) => {
-      const matchingSubjectRule = _.find(subjectRules, subjectRule =>
-        subjectRule.id.toString() === workflowRule.id.toString())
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Beware of rules with ID 0, like the ones Space Warps are using. I think those are fixed for subjects but I’m not 100% sure for workflows.

eatyourgreens:
I think rule IDs are required according to the feedback Readme in PFE, so I’m not sure about silently ignoring rules that don’t have an ID.

eatyourgreens:
Here’s that Readme. https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/features/feedback/README.md

srallen:
Sure, it's in the readme, but so far, several of the bugs we've had with feedback has been from projects not following those guidelines and we don't reject subject uploads with incorrect setup, so I'm not sure how we stop it. How do you propose we inform the projects their feedback setup is malformed? 

This change will prevent a TypeError to be thrown, which I think can be related to some of the odd states we've been observing, so I think we should do this. The other option would be to throw an error, but we don't have our sentry logging setup done, so this wouldn't be caught in a useful way for us to be aware of it yet.

eatyourgreens:
When we fixed feedback for bad indexes, we added a console warning that the index is invalid. I’d suggest doing the same here.

packages/lib-classifier/src/store/feedback/helpers/generate-rules.js
```diff
@@ -23,8 +23,13 @@ function generateRules (subject, workflow) {
...
   _.forEach(workflowRules, (rules, taskId) => {
     const taskRules = _.reduce(rules, (result, workflowRule) => {
-      const matchingSubjectRule = _.find(subjectRules, subjectRule =>
-        subjectRule.id.toString() === workflowRule.id.toString())
+      const matchingSubjectRule = _.find(subjectRules, (subjectRule) => {
+        if (subjectRule.id && workflowRule.id) {
+          return subjectRule.id.toString() === workflowRule.id.toString()
+        }
+
+        return null
```
eatyourgreens: I think the callback for `.find` should return a boolean, if it’s the same as the native JS `Array.find` method. So maybe return `false` rather than relying on `null` being falsy.

packages/lib-classifier/src/store/feedback/helpers/generate-rules.js
```diff
@@ -23,8 +23,13 @@ function generateRules (subject, workflow) {
 
   _.forEach(workflowRules, (rules, taskId) => {
     const taskRules = _.reduce(rules, (result, workflowRule) => {
-      const matchingSubjectRule = _.find(subjectRules, subjectRule =>
-        subjectRule.id.toString() === workflowRule.id.toString())
+      const matchingSubjectRule = _.find(subjectRules, (subjectRule) => {
+        if (subjectRule.id && workflowRule.id) {
```
eatyourgreens: This will fail if either rule ID exists but is 0.

packages/lib-classifier/src/store/feedback/helpers/generate-rules.js
srallen: It doesn't behave the same as the native `find`: https://lodash.com/docs/2.4.2#find

srallen: Wait, I had to read that a couple of times and I am wrong. Ignore me.

PR #1121, opened 2019-09-11
Make feedback.isActive a computed value

eatyourgreens:
Make `feedback.isActive` a view that wraps the `isFeedbackActive` helper. Update tests to stub `isFeedbackActive` when they mock active or inactive feedback.

Adds a `feedback.isValid` view, rather than validating feedback in component code.

Package:
lib-classifier

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/store/FeedbackStore.js
+++ b/packages/lib-classifier/src/store/FeedbackStore.js
@@ -24,7 +24,10 @@ const FeedbackStore = types
     },
     get isActive () {
       const { projects = {}, subjects = {}, workflows = {} } = getRoot(self)
-      return helpers.isFeedbackActive(projects.active, subjects.active, workflows.active)
+      const validProject = isValidReference(() => projects.active) && projects.active
+      const validWorkflow = isValidReference(() => workflows.active) && workflows.active
+      const validSubject = isValidReference(() => subjects.active) && subjects.active
+      return helpers.isFeedbackActive(validProject, validSubject, validWorkflow)
     },
     // This is a workaround for https://github.com/zooniverse/front-end-monorepo/issues/1112
     // (feedback incorrectly being active when there are no rules)
@@ -100,10 +103,9 @@ const FeedbackStore = types
     }
 
     function createRules (subject) {
-      const validProjectReference = isValidReference(() => getRoot(self).projects.active)
       const validWorkflowReference = isValidReference(() => getRoot(self).workflows.active)
 
-      if (validProjectReference && validWorkflowReference && subject) {
+      if (validWorkflowReference && subject) {
         const workflow = getRoot(self).workflows.active
 
         if (self.isActive) {
--- a/packages/lib-classifier/src/store/FeedbackStore.js
+++ b/packages/lib-classifier/src/store/FeedbackStore.js
@@ -26,6 +26,9 @@ const FeedbackStore = types
       const { projects = {}, subjects = {}, workflows = {} } = getRoot(self)
       return helpers.isFeedbackActive(projects.active, subjects.active, workflows.active)
     },
+    get isValid () {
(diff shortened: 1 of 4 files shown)


eatyourgreens:
This has big conflicts with #1002 so maybe wait for that to merge first.

eatyourgreens:
The merge conflicts here are big, so I'm going to redo this from the latest master.

eatyourgreens:
This is updated. I just need to figure out why some tests are failing.

packages/lib-classifier/src/store/FeedbackStore.js
```diff
@@ -23,6 +22,13 @@ const FeedbackStore = types
       return flatten(Array.from(self.rules.values()))
         .some(rule => rule.hideSubjectViewer)
     },
+    get isActive () {
+      const { projects = {}, subjects = {}, workflows = {} } = getRoot(self)
+      return helpers.isFeedbackActive(projects.active, subjects.active, workflows.active)
+    },
+    get isValid () {
```
rogerhutchings: As discussed in person:

- I'm not in love with the naming, but I understand why now
- More importantly, if this is a workaround for a bug, we need a comment to that effect and a link to an issue

eatyourgreens: No problem. This is a workaround for #1112 (feedback incorrectly being active when there are no rules.)

rogerhutchings:
```suggestion
    // This is a workaround for https://github.com/zooniverse/front-end-monorepo/issues/1112
    // (feedback incorrectly being active when there are no rules)
    get isValid () {
```

packages/lib-classifier/src/store/FeedbackStore.js
```diff
@@ -23,6 +22,15 @@ const FeedbackStore = types
       return flatten(Array.from(self.rules.values()))
         .some(rule => rule.hideSubjectViewer)
     },
+    get isActive () {
+      const { projects = {}, subjects = {}, workflows = {} } = getRoot(self)
+      return helpers.isFeedbackActive(projects.active, subjects.active, workflows.active)
```
srallen: What happens if these references are not valid? Should `isValidReference` be used first before calling `isFeedbackActive`?

packages/lib-classifier/src/store/FeedbackStore.js
```diff
@@ -96,9 +104,7 @@ const FeedbackStore = types
       const validWorkflowReference = isValidReference(() => getRoot(self).workflows.active)
 
       if (validProjectReference && validWorkflowReference && subject) {
```
srallen: Checking for a valid project reference here is now cruft.

packages/lib-classifier/src/store/FeedbackStore.js
eatyourgreens: The function returns false if they aren't valid, I think. I may be misunderstanding what a valid reference is.

srallen: Several of the errors I fixed in the MST upgrade had to do with attempting to use active references when they weren't valid. For safety reasons, I think it'd be best to check that they're valid first, especially if we're not sure on the behavior here.

eatyourgreens: Done.

`isFeedbackActive()` returns false if any of its arguments are falsy but I wasn't sure if that's sufficient.

packages/lib-classifier/src/store/FeedbackStore.js
eatyourgreens: With this PR, a subject can't have `isActive` true and an empty rules array. In theory (and according to the tests.)

PR #1131, opened 2019-09-18
Add recently classified subjects to the Classify page

### LINKED ISSUE #37: "Your recent classifications" widget
screenshot 2018-06-07 16 50 03

eatyourgreens:
Add components and a basic story for the recent subjects/classifications component on the Classify page. I'm going to leave sharing on social media for a later PR.

Screenshot of the recent classifications box from Galaxy Zoo, showing 3 subjects with controls to link to Talk, favourite the subject or add it to a collection.

TODO:
 - [x] Wire favourites up to the Panoptes API.
 - [x] Add a layout for small screens.

Package:
app-project

Closes #37.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?
- [x] Can you `rm -rf node_modules/ && yarn bootstrap` and app works as expected?
- [x] Can you run a [production build](https://github.com/zooniverse/front-end-monorepo#getting-started) of the app?


eatyourgreens:
Video subjects and portrait images are broken (see the storybook examples.) That seems to be a bug in the Media component, which also affects the Recent Subjects component on the project home page (related to #1147.)

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
```diff
@@ -0,0 +1,83 @@
...
+        />
+      </Anchor>
+      <TalkLink
+        href={href}
+      />
+      <FavouritesButton
+        checked={recent.favorite}
+        onClick={toggleFavorite}
+      />
+      <CollectionsButton
```
eatyourgreens: These buttons should be disabled if you aren’t logged in.

rogerhutchings:
There's an ugly transition going on when a new subject is added:

recent-classifications

You can probably get away with it since it wouldn't normally be in the viewport like this.

rogerhutchings:
There should probably be a message saying you haven't classified anything, rather than just a blank box

Screenshot 2019-09-23 11 06 05

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
```diff
@@ -0,0 +1,83 @@
...
+        onClick={addToCollections}
+      />
+    </Box>
+  )
+}
+
+SubjectPreview.propTypes = {
+  className: string,
+  height: number,
+  recent: shape({
```
rogerhutchings: this should probably just be called `subject` - `recent` isn't very descriptive

eatyourgreens:
Yeah, the placeholder is sized to the height and width that are passed in here:
https://github.com/zooniverse/front-end-monorepo/blob/82eae1e2d5236cae70ab606cabc6852022539af2/packages/lib-react-components/src/Media/components/ThumbnailImage/ThumbnailImage.js#L75
and I'm cropping the thumbnails to fit into a square ~350px on a side.

I'm open to suggestions as to how to do the image sizing better and still look good for portrait and landscape images.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/components/FavouritesButton/FavouritesButton.js
```diff
@@ -0,0 +1,44 @@
...
+
+export default function FavouritesButton (props) {
+  const { className, checked, disabled, onClick } = props
+  const label = checked ? 'FavouritesButton.remove' : 'FavouritesButton.add'
+  const filled = checked ? 'true' : undefined
+  return (
+    <MetaToolsButton
+      aria-checked={checked}
+      disabled={disabled}
+      icon={<Favourite className={className} color='dark-5' filled={filled} size='1em' />}
```
rogerhutchings: You don't need `className` here; if you log it, it's empty (both checked and unchecked), and there aren't any warnings thrown by it not being there.

If you're starting with a grommet component, [it appears to be handled for you](https://github.com/grommet/grommet-icons/blob/master/src/js/StyledIcon.js#L40)

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
```diff
@@ -0,0 +1,83 @@
...
+import CollectionsModal from '../../../../components/CollectionsModal'
+
+import en from './locales/en'
+
+counterpart.registerTranslations('en', en)
+
+function SubjectPreview ({ height, recent, slug, width }) {
+  const subjectURLs = recent.locations.map(location => Object.values(location)[0])
+  const subjectURL = subjectURLs[0]
+  const [ checked, setChecked ] = useState(recent.favorite)
```
rogerhutchings: Do we have any examples of tests with hooks so far? I was trying to write one for `useRouter` the other day and couldn't get it working.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
```diff
@@ -0,0 +1,83 @@
+import counterpart from 'counterpart'
+import { array, bool, number, shape, string } from 'prop-types'
+import React, { useState } from 'react'
+import styled from 'styled-components'
+import { Anchor, Box } from 'grommet'
+import { Media } from '@zooniverse/react-components'
+import { CollectionsButton, FavouritesButton, TalkLink } from './components'
```
rogerhutchings: British or American spelling of "favourite"? I don't mind which, as long as it's consistent

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/components/MetaToolsButton/MetaToolsButton.js
```diff
@@ -0,0 +1,59 @@
...
+    line-height: ${(props) => props.theme.paragraph.small.height};
+  }
+`
+
+function MetaToolsButton (props) {
+  const { className, disabled, icon, onClick, text } = props
+
+  return (
+    <StyledPlainButton
+      className={className}
```
rogerhutchings: no need for `className` again

rogerhutchings:
> I'm open to suggestions as to how to do the image sizing better and still look good for portrait and landscape images.

Sounds like a @beckyrother question

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
eatyourgreens: Good point. It represents a `recents` resource from Panoptes, rather than a subject, but I'm not sure if that's important.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/components/FavouritesButton/FavouritesButton.js
eatyourgreens: I'll try this but I think you are supposed to pass `className` down when you use `styled()`. `Favourite` here is a `styled(HeartIcon)`.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
eatyourgreens: It's `Favourite` in the classifier components. 🤷‍♂

rogerhutchings: I think `favorite` is used elsewhere in this PR? (might be from the API, mind)

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
eatyourgreens: Not yet.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
eatyourgreens: Yes, Panoptes uses `favorite`.

eatyourgreens:
I ran the linter and the resulting diff was massive. Maybe lint the project app on master first?

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
eatyourgreens: I can't see a `favorite` flag on recents returned from Panoptes, so I've changed this to use `favourite`. We can revisit that if the API does start indicating which subjects are in your favourites.

eatyourgreens: Subjects do have `subject.favorite` though. 🤔 I might have been too hasty with that change.

eatyourgreens:
The thumbnail sizing is a code problem, not a design problem. The design calls for them to be cropped to 270px wide, 220px high but the Media component doesn't support cropping to a specific window.

Here's an example with images of very different aspect ratios. There are other examples in the storybook.

Screenshot showing subject thumbnails in a jagged row, with very different heights.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
eatyourgreens: Done, but we now have `subject.subjectId` which I'm not wild about either.

eatyourgreens:
All linted now, although `app-project` needs someone to go through it by hand to fix manual errors.

I renamed recents to subjects but I'm still not wild about having `subject.subjectId` so I might fix that too.

>There should probably be a message saying you haven't classified anything, rather than just a blank box

Yeah. The grid also looks horrible with only one or two items, since grid columns expand to wrap content. I'll add that case to the storybook, and see if there's a way to set a minimum size for empty tracks in the grid.
Screenshot showing oversized previews for recent subjects when there are only one or two instead of three.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
eatyourgreens: Here, I think a test would click `FavouritesButton` and expected its `checked` attribute to change from `false` to `true`. See the counter example here:
https://reactjs.org/docs/hooks-faq.html#how-to-test-components-that-use-hooks

>From React’s point of view, a component using Hooks is just a regular component. If your testing solution doesn’t rely on React internals, testing components with Hooks shouldn’t be different from how you normally test components.

eatyourgreens: I've added those tests, so we now have examples of tests that use hooks.

Sounds like enzyme only partially supports hooks right now: https://blog.carbonfive.com/2019/08/05/shallow-testing-hooks-with-enzyme/

From reading up a bit around that, it may be a React issue with shallow rendering not updating the component via `useEffect`. The workaround might be to use `mount` to test anything other than `useState`.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/components/SubjectPreview/SubjectPreview.js
```diff
@@ -23,7 +23,8 @@ function SubjectPreview ({ height, isLoggedIn, subject, slug, width }) {
 
   function toggleFavourite () {
     subject.toggleFavourite()
-    setIsFavourite(subject.favorite)
+    console.log('isFavourite', isFavourite)
```
eatyourgreens: D'oh!

eatyourgreens:
I might have fixed the problem of cropping the thumbnails to a uniform size, by wrapping them in a fixed size box.
Screenshot showing subject thumbnails cropped to a uniform size for Planet Hunters TESS.

eatyourgreens:
The placeholder is ignoring the fixed size box that I added. Maybe it should be sized to 100% rather than fixed height and width?

srallen:
@beckyrother would like to write up a Talk post before this gets released, so FYI, do not merge until we get an ok from her

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/RecentSubjects.js
```diff
@@ -0,0 +1,59 @@
+import counterpart from 'counterpart'
+import { Grid, Paragraph } from 'grommet'
+import { array, bool, string } from 'prop-types'
+import React from 'react'
+
+import en from './locales/en'
+import ContentBox from '../../../../shared/components/ContentBox'
+import SubjectPreview from '../../../../shared/components/SubjectPreview'
```
eatyourgreens: #1165 will allow these paths to be written as `@shared/components/…`

eatyourgreens:
There's a `Media` component in the project app which does a completely different job than the Zooniverse `Media` component. I'm using both in this PR, so I've aliased the fresnel component to `ResponsiveContent`, since that seems to describe what it does.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/RecentSubjects.spec.js
```diff
@@ -0,0 +1,16 @@
...
+let wrapper
+
+describe('Component > RecentSubjects', function () {
+  before(function () {
+    wrapper = shallow(<RecentSubjects />)
+  })
+
+  it('should render without crashing', function () {
+    expect(wrapper).to.be.ok()
+  })
```
eatyourgreens: At the risk of making this PR even longer, the `RecentSubjects` component needs tests.

rogerhutchings: Separate PR?

eatyourgreens:
Dammit, that's because I changed the width from `270px` to `100%` on large viewports.

beckyrother:
Here's a good place to start tagging me in things when there's design questions. 

The width of thumbnail images shouldn't change and there should always be 3 items in that section. I'll make a placeholder image to use if there's one or two recents. If there haven't been any, I thought there was a design for that already with some text messaging?

rogerhutchings:
You're going to hate this:

Screenshot 2019-09-26 17 01 36

> The width of thumbnail images shouldn't change and there should always be 3 items in that section.

So it should be a carousel when there's not enough room for 3?

beckyrother:
I do hate this! But I hate it less than stretching out one image to fit the whole box. How about they scale down until the break and then it's a carousel with just one image?

eatyourgreens:
The buttons and links below each thumbnail won't be clickable in a carousel. At least, the Grommet carousel doesn't allow any interaction with its child elements.

beckyrother:
Yikes, let's not do a carousel then. How dumb! In that case just show 1 recent for small screens

eatyourgreens:
> In that case just show 1 recent for small screens

Way ahead of you there:
https://github.com/zooniverse/front-end-monorepo/pull/1131/files#diff-f3784c6b59b153ae98217ca5734e9088R45

beckyrother:
I'm still going to make a placeholder image for when there's either 1 or 2 recents. Could there be a couple different ones that rotate?

eatyourgreens:
What do the placeholders do for the link to the subject's Talk page, and the favourites and collections buttons?

NB. there should already be a RecentSubjects story in the storybook for one recent subject.

beckyrother:
I don't have access to Storybook :( 

The links should be visible but disabled if there isn't a subject

eatyourgreens:
Links can't be disabled but we could have the text 'Discuss on Talk' beneath each placeholder, without it linking to anything.

eatyourgreens:
I've updated the grid so that it should always have 3 columns in wide viewports. The `width: 100%` is still causing problems as the placeholder width, even with `overflow:hidden` on the parent.

beckyrother:
In that case can links be hidden? That's going to be frustrating to see buttons and they work sometimes and not other times.

eatyourgreens:
`fresnel` has completely broken the storybook. Oops. 😳

eatyourgreens:
>In that case can links be hidden? That's going to be frustrating to see buttons and they work sometimes and not other times.

Sure. At the moment those other two grid columns are completely empty.
Screenshot showing the Recent Classifications box when there is only one recent subject, ranged left.

beckyrother:
placeholder 1

On second thought let's just use this one placeholder

beckyrother:
This section also needs a link to the recent classifications page in the same style as the 'view more stats' link on the project home page

beckyrother:
Sam's just pointed out that we currently use the US spelling of 'Favorites'. I'm fine either way as long as we're consistent across the site.

eatyourgreens:
Placeholder images have been added when subjects aren't present.

eatyourgreens:
Here's the storybook for only one subject in Firefox.
Screenshot showing Recent Classifications with one subject and two placeholders.

beckyrother:
Is it possible to crop the placeholder images to be the same height as the real subject image?

eatyourgreens:
Sure.
Screenshot showing placeholders cropped to the same dimensions as the subject.

eatyourgreens:
@rogerhutchings I think this is ready now.

beckyrother:
Please post a screenshot so I can review this.

eatyourgreens:
I think I posted one on Friday.

beckyrother:
Ah there it is, github hid it. There's no way to show the buttons underneath as disabled rather than hiding them completely?

eatyourgreens:
Sure, but you asked not to do that, particularly since the placeholder can't link to Talk.

>In that case can links be hidden? That's going to be frustrating to see buttons and they work sometimes and not other times.

beckyrother:
Ha, sorry, it's early. This lgtm, going to approve. Thanks for your help.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/RecentSubjects.js
```diff
@@ -0,0 +1,78 @@
...
+            <Box
+              align='center'
+              justify='center'
+              height={height}
+              key={i}
+              overflow="hidden"
+              width={'100%'}
+            >
+              <img alt="" role='presentation' src='/static/subject-placeholder.png' />
+            </Box>
```
srallen: Non-blocking: I would suggest pulling this into a placeholder component and also using it with the `Media` component (it supports a `placeholder` prop that you can define a React component or HTML), so we have consistency in placeholders being used. Right now this is shown if there aren't enough recents, but I don't see any reason to also not use this as the placeholder while the image is being processed by the thumbnail service.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/RecentSubjects.stories.js
```diff
@@ -0,0 +1,140 @@
...
+  .add('one subject', () => (
+    <Grommet theme={zooTheme}>
+      <RecentSubjectsContainer
+        isLoggedIn={boolean('User logged in', true)}
+        recents={RECENTS.slice(0,1)}
+        projectName='NestCams'
+        slug='spotteron/nestcams'
+      />
+    </Grommet>
+  ))
```
srallen: All of the stories need viewports set or else if you select the narrow viewport one, the other stories stay in the `iphone5` preview. The others should have `, { viewport: { defaultViewport: 'responsive' }}` to reset it.

eatyourgreens:
The favourites button etc. are implemented slightly differently here, but I moved them into the top-level `shared` directory of the project app so that they could be easily moved into a library. The classifier would have to be updated to use the new components, but I think the changes are small.

eatyourgreens:
Do you know how to get the theme working with the stories? The theme toggle in the storybook doesn't seem to do anything.

srallen:
Check out what we do in the react components library stories: https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-react-components/src/Modal/Modal.stories.js

We just add a story that is 'dark theme'. Just need to pass in a theme to the `Grommet` with `dark: true`

eatyourgreens:
Ok. I've been using the dark/light theme toggle at the top of the screen, which usually works. Sounds like you could also add a knob to your stories that toggles the theme between dark and light. I've got one that switches the user between logged in and logged out.

srallen:
> I've been using the dark/light theme toggle at the top of the screen, which usually works.

This just toggles the background color of the story area, not the whole theme. 

> Sounds like you could also add a knob to your stories that toggles the theme between dark and light. I've got one that switches the user between logged in and logged out.

Sure, but that'd be more complex and more code.

srallen:
I suppose a knob would mean any of the stories can be previewed as light or dark.

eatyourgreens:
Exactly! One of my recent PRs has an example of a dynamic Boolean prop in the storybook. I think it might be the subject preview story here.

srallen:
Sounds like a good idea. Let's open an issue (or two) to track some of these follow up items?

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/RecentSubjects.js
eatyourgreens: The Media component has issues with image sizing but I can give this a go.

eatyourgreens:
@srallen I've added a theme toggle to all stories, which works for Recent Subjects but the Subject Preview component is not using the background colour from the theme. Is that expected?
Screenshot of Recent Classifications in dark mode in the storybook.
Screenshot of Subject Preview in dark mode in the storybook, showing that the background is still white.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/RecentSubjects.js
eatyourgreens: Sorry, are you suggesting that this image also be used as a placeholder in the `SubjectPreview` component, or that this `` element be replaced with a `Media` component?

eatyourgreens: Also noting here so that it isn't lost: `RecentSubjects` still isn't tested.

eatyourgreens: The `Media` placeholder has the same sizing bug that was fixed in #1160 for video and audio. You can see it here, where the placeholder is rendered at 700x700px (the maximum dimensions of the thumbnail), rather than the dimensions of the supplied image.

eatyourgreens:
>Instead of duplicating the collections button, favorites button, and associated icons, why not move it into the shared react components library?

`SubjectPreview` is in the top-level `shared` directory, and should be completely self-contained (but that code could always use a second eye on it.) I think the only thing holding it back from being packaged as reusable module is that the collections modal is bundled inside the subject preview. Consuming components can't specify the action that happens when 'Add to collections' is pressed, which is ok here but limits use of `SubjectPreview` outside the project app.

srallen:
> SubjectPreview is in the top-level shared directory, and should be completely self-contained (but that code could always use a second eye on it.) I think the only thing holding it back from being packaged as reusable module is that the collections modal is bundled inside the subject preview. Consuming components can't specify the action that happens when 'Add to collections' is pressed, which is ok here but limits use of SubjectPreview outside the project app.

I'm not suggesting the `SubjectPreview` be moved into the shared react components library, but the underlying metatools buttons and icons that are identical to what's being used in the classifier library. If there's an identified need to reuse the whole `SubjectPreview` composite component, then we can address that at that time.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/RecentSubjects.spec.js
srallen: I created #1190 to track this.

eatyourgreens:
>If there's an identified need to reuse the whole SubjectPreview composite component, then we can address that at that time.

Previews for subjects in Recents and Collections seem like they should use this component.

packages/app-project/src/screens/ClassifyPage/components/RecentSubjects/RecentSubjects.js
srallen: Tracked in #1195

PR #1139, opened 2019-09-19
Add docker testing notes

camallen:
Package: `app-project`

Add some notes on how to use docker containers to interactively test apps, i.e. from a bash console in the container with all the required test infrastructure installed.

I found this flow useful:
1.  dip into the code
0. make some changes
0. test my changes using `.only` mocha syntax
0. re-run the tests with `yarn test`
0. iterate until happy

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/README.md
+++ b/packages/app-project/README.md
@@ -47,7 +47,22 @@ yarn start
 See [Testing](#testing) for more details.
 
 #### Docker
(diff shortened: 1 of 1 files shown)


eatyourgreens:
👍 I didn't know you could change the entrypoint like that.

Issue #1106, opened 2019-09-03
Drawing task - delete marking


mcbouslog:
**Package**
lib-classifier

**Feature or Issue Description**
Allow markings from a drawing task in the current workflow step to be deletable. 

**To Dos**
...placeholder for project management, details TBD...

Issue #1104, opened 2019-09-03
Drawing task - read-only render for previous markings


mcbouslog:
**Package**
lib-classifier

**Feature or Issue Description**
Have a component, the markings renderer, that manages the rendering of previous annotation marks as read only. It will hide them if hide previous marks is toggled.

See [ADR 15](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-15.md)

**To Dos**
...placeholder for project management, details TBD...

Issue #1103, opened 2019-09-03
Drawing task - save markings to annotation


mcbouslog:
**Package**
lib-classifier

**Feature or Issue Description**
On workflow step advance (Next) or Done (or Done & Talk) save markings from drawing/interaction state to annotation

see [ADR 15](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-15.md)

**To Dos**
...placeholder for project management, details TBD...

simensta:
Using the [createViewModel()](https://github.com/mobxjs/mobx-utils#createviewmodel) functionality might be useful for this feature.

eatyourgreens:
Closed by #1339.

PR #1140, opened 2019-09-19
Add DrawingContainer

mcbouslog:
Package:
- lib-classifier
- helpful staging workflow: http://localhost:8080/?project=1876&workflow=3344

Staging project and workflow:
- http://localhost:8080/?project=1876&workflow=3344

Describe your changes:
- changes InteractionLayer event handlers from onMouse to onPointer
- refactors drawing store to more explicit naming
- renders DrawingContainer in InteractionLayerContainer if drawing task in active workflow step
- adds DrawingContainer for current and active workflow step markings
- adds Point and Line (and Root) drawing tool components

Notes:
- these changes are a portion of #1078 
- subsequent to these changes, adding the coordinate stream (as in #1078) and tool stores, saving markings to annotation (#1103), read-only render for previous markings (#1104) and delete marking (#1106) can all be worked on concurrently
- there is some temporary code to randomly add a point or line on `pointerup` to allow the items noted above to be worked on concurrently

Why WIP:
- [x] DrawingContainer tests
- [x] Root tests
- [x] Line tests

# Review Checklist

## General

## Apps

- [x] Does it work on mobile?
- [x] Can you `rm -rf node_modules/ && yarn bootstrap` and app works as expected?
- [x] Can you run a [production build](https://github.com/zooniverse/front-end-monorepo#getting-started) of the app?


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/DrawingContainer.js
```diff
@@ -0,0 +1,142 @@
...
+    }
+    if (tool.type === 'point') {
+      const coordinates = { x: (Math.floor(Math.random() * 400)), y: (Math.floor(Math.random() * 400)) }
+      this.finishMark(coordinates)
+    }
+  }
+
+  setActiveMark () {
+    const { activeDrawingTool, activeStepTasks } = this.props
+    const [activeDrawingTask] = activeStepTasks.filter(task => task.type === 'drawing')
```
srallen: I think we discussed having the drawing store filter this on a view getter function? Totally ok if that's going to come in a later PR.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/tools/Root.spec.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import { expect } from 'chai'
+import Root from './Root'
```
srallen: Non-blocking and more of an FYI. I think with the new module spec (mjs) in javascript that if the export name and the filename do not match that it'll error. So, we should try to match the file names to the export name whenever possible. Maybe rename the file to `DrawingToolRoot`?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/DrawingContainer.js
mcbouslog: Yes and yes, those changes are in PR #1164, [here's a link specifically to those views in the store](https://github.com/zooniverse/front-end-monorepo/pull/1164/files#diff-e76319fbfa25691959816b6c45adbf06R11).

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/DrawingContainer.js
```diff
@@ -0,0 +1,158 @@
...
+
+    this.setState({ activeMark: newMark, marks: newMarks })
+  }
+
+  render () {
+    const { activeMark, marks } = this.state
+    if (!activeMark) return null
+
+    const MarkComponent = getDrawingTool(activeMark.tool.type)
+    // TODO: add error handling if don't get requested tool type
```
simensta: @mcbouslog, should we address this to do in a separate PR? If so, I can create issue for https://github.com/zooniverse/front-end-monorepo/projects/11.

mcbouslog: I think punt for now? I'm not sure exactly what behavior we'd like, i.e. console warning prob at minimum, but do we also want something in UI?

srallen:
@simensta are those to dos noted in an issue somewhere so we don't lose track of them?

eatyourgreens:
Check out the daily classification chart for focusable SVG.

PR #1145, opened 2019-09-20
Test feedback with invalid or undefined arguments

eatyourgreens:
Package:
lib-classifier

Adds some tests to the feedback helper functions that call them with null, undefined or false (invalid) projects, workflows and subjects.

This is more for completeness than anything else, to rule out another potential cause of #1112.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?


eatyourgreens:
CI is failing because null or undefined values break workflow feedback.

Issue #1152, opened 2019-09-20
Design feedback: Project Home Page


beckyrother:
- [x] When a project does not have an active workflow, show a [disabled button](https://projects.invisionapp.com/d/main#/projects/prototypes/12923997/assets/77727637) rather than nothing (#1150)

- [x] All buttons need hover states (#1188, #1204)

- [x] Add spacing between multiple workflow buttons (#1170)

- [x] Hero component always min-height 545px. Classify + buttons should align to the bottom, while the title and description should align to the top. (#1171)

- [x] Recent subjects 270x200, crop images from the center at the longest side. This is how the current Recents widget handles images, and should be continued: For both consistency across projects and in consideration of smaller screen sizes. 

- [x] Dark theme toggle feels awkward. Can all the module boxes line up with the right side of the Recents module, so there's a consistent edge on the right side, even if it's different to the edge on the left side, see [The Atlantic](https://www.theatlantic.com/entertainment/archive/2019/09/blink-182-nine-album-review-wistful-and-catchy/598420/) for a cool iteration of this. (#1186)

- [x] What did we decide about the progress bars on the individual workflow buttons? https://zooniverse.slack.com/archives/C0Z7YNTSL/p1554151344033200 Is I think where we landed? Should still figure out something for times when a hover isn't possible. Happy to talk about this more. (#1151, #1204)

***Small Screens / mobile***

- [x] Hero component should always be 230px tall. The PH: TESS project is very tall and on some browsers and devices only the image will appear before scrolling. (#1155)

- [x] Recents – move pagination dots to underneath subject image; they are not currently visible on top of the image.

eatyourgreens:
Recents are using a [Grommet Carousel](https://v2.grommet.io/carousel) on small screens, which made them easy to set up but doesn't give us much control over how the controls are displayed. I have the pagination turned off in the code, but that option isn't supported by the version of Grommet that we're using.

srallen:
Grommet version just needs to be bumped to disable the pagination on small screens.

beckyrother:
If the pagination is disabled will it be clear that it's a carousel?

srallen:
@beckyrother it's now visible on https://frontend.preview.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/

The arrow buttons are still available with the Grommet carousel on small screen sizes, though we need to adjust the color of the icons on the light theme version (they're too dark gray, should probably be white with an opacity underneath?):

[...]

PR #1153, opened 2019-09-20
Adjust preview thumbnail size

eatyourgreens:
Set the maximum subject thumbnail size to 270px width, 200px height.

See #1152.

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjects.js
+++ b/packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/RecentSubjects.js
@@ -7,8 +7,8 @@ import SubjectThumbnail from './components/SubjectThumbnail'
 // TODO: Use the subject viewers from the classifier
 function RecentSubjects (props) {
   const { className, href, subjects } = props
-  const height = 350
-  const width = 400
+  const height = 200
+  const width = 270
   return (
     <Grid
       className={className}
--- a/packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/components/SubjectThumbnail/SubjectThumbnail.js
+++ b/packages/app-project/src/screens/ProjectHomePage/components/ZooniverseTalk/components/RecentSubjects/components/SubjectThumbnail/SubjectThumbnail.js
@@ -39,9 +39,9 @@ function SubjectThumbnail ({ height, href, width, subject }) {
       >
         <Media
           alt={`subject ${subject.id}`}
-          height={400}
+          height={700}
           src={subjectURL}
-          width={400}
+          width={700}
         />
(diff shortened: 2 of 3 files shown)


eatyourgreens:
This might introduce large gaps between the preview images on large viewports. See the screenshots in this discussion: https://github.com/zooniverse/front-end-monorepo/pull/1017#discussion_r320628438

eatyourgreens:
Here's PH-TESS on a reasonably large viewport, where the grid column width is ~ 350px but the image width is 270px.

The Recent Subjects box on a large viewport, where each preview has a large, white gap to the right.

beckyrother:
Honestly this whole page looks rough at that wide of a viewport. Can we do like we do on the non-logged-in home page and set a max width for the page? Here's the widest I've designed for: https://projects.invisionapp.com/d/main#/console/12924056/344288108/preview

eatyourgreens:
Screenshot showing heavily pixelated previews of 19th Century manuscript pages.
Transcription projects look particularly bad at 270px by 200px, because the thumbnail is resized down to 200px wide, then stretched out to 270px.

beckyrother:
Sorry I don't understand what that means? It starts square at 200x200 and is resized back up to 270?

eatyourgreens:
I think it looks much better if the thumbnails are sized to a 400x400px square, then cropped inside a 270x200px window with `overflow: hidden`.
Screenshot showing previews for 3 manuscript pages, where the images are resized down rather than up in the browser.

eatyourgreens:
>Sorry I don't understand what that means? It starts square at 200x200 and is resized back up to 270?

Yep, if you give the thumbnail server dimensions of 270x200, it will give you back a thumbnail that fits inside that box. In this case, that means it will reduce the height down to 200px.

eatyourgreens:
@beckyrother you can see the pixelation that causes in the preview images on the[ Worlds of Wonder home page](https://www.zooniverse.org/projects/lbeiermann/worlds-of-wonder).

beckyrother:
Got it, thanks! Agreed that the 400x400 option looks much better.

rogerhutchings:
On smaller screen sizes, the preview image comes out at 400 x 400: 

Screenshot 2019-09-26 10 33 22

eatyourgreens:
Good point. Maybe the thumbnail size should be 700x700.

rogerhutchings:
It might be a good place to use a `srcSet`?

eatyourgreens:
Maybe, but that would be a change to the shared Media component. I've already got some fixes for sizing in that component in #1160.

eatyourgreens:
@rogerhutchings thumbnails should be fixed on tablet-sized screens now.

eatyourgreens:
Over in #1131, I'm experimenting with using CSS values, not numbers, for `height` and `width`.

srallen:
I would be supportive of adding `srcSet` to the `Media` component. :+1:

PR #1155, opened 2019-09-22
Fix height of Hero image to 230px on small screens

rogerhutchings:
As per https://github.com/zooniverse/front-end-monorepo/issues/1152,
the height of the Hero image should be fixed to 230px on small screens.

This commit:

- Removes Grommet ResponsiveContext, which causes layout jumping
- Adds a CSS media query based on the Grommet theme breakpoints
- Sets the height of the image to 230px


packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Background/Background.js
```diff
@@ -1,39 +1,47 @@
...
-    height: auto;
+    height: 230px;
     min-height: inherit;
-    object-fit: contain;
+    object-fit: cover;
+    width: 100%;
+
+  @media (min-width: ${props => props.breakpoint}px) {
+    height: 100%;
+    object-fit: cover;
```
eatyourgreens: This is already declared so I don't think you need to repeat it.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Background/Background.js
```diff
@@ -1,39 +1,47 @@
...
+    object-position: 0 50%;
     width: 100%;
-  `}
+  }
```
eatyourgreens: Same here.

rogerhutchings:
> Should I consider removing withResponsiveContext from the work I'm doing on the Classify page?

I reckon so. Grommet's ResponsiveContext just isn't suitable for SSR (yet). There's also the stuff using fresnel if you need to render different component trees.

PR #1161, opened 2019-09-24
Make generateHostUrl server-agnostic

rogerhutchings:
The function generating the host URL (used in OG tags etc) is breaking when run on the client. This makes it server/client agnostic by checking for the `req` object, and falling back to `location` when `req` is unavailable.

### DIFF
--- a/packages/app-project/pages/_app.js
+++ b/packages/app-project/pages/_app.js
@@ -122,7 +122,11 @@ function getSlugFromUrl (relativeUrl) {
 }
 
 function generateHostUrl (context) {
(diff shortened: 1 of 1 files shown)


rogerhutchings:
`_app.js` is currently untested - we'd need to move the whole thing out to the `src` folder and write tests for it the same way we do for the actual page components.

eatyourgreens:
You could move this function into a helper library and test it there. Same with `getSlugFromUrl`. They're both declared outside the `MyApp` class.

PR #1162, opened 2019-09-24
Add no workflows fallback to Hero

### LINKED ISSUE #1150: Add fallback for no workflows in Hero widget
At the moment, the Hero widget doesn't show anything in the workflow selection area if there are no active workflows.

Options include:

- disabled button
- text message
- (any other suggestions?)

Related:

#1077

rogerhutchings:
Adds a disabled button with a `nothing to classify` message to the Hero component when there are no active workflows. 

Screenshot 2019-09-24 12 09 48

cc #1152
Closes #1150

### DIFF
--- a/packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/WorkflowSelector.js
+++ b/packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/WorkflowSelector.js
@@ -37,18 +37,25 @@ function WorkflowSelector (props) {
 
       {(workflows.loading === asyncStates.success) && (
         <Box margin={{ top: 'small' }}>
-          {workflows.data.map(workflow =>
+          {(workflows.data.length > 0) && workflows.data.map(workflow =>
             <WorkflowSelectButton key={workflow.id} workflow={workflow} />
+          )}
+          {(workflows.data.length === 0) && (
+            <Box background='accent-4' pad='xsmall'>
+              <Text size='small' textAlign='center'>
+                {counterpart('WorkflowSelector.noWorkflows')}
+              </Text>
+            </Box>
           )}
         </Box>
       )}
 
-      {(!asyncStates.values.includes(workflows.loading)) && (
+      {(![asyncStates.success, asyncStates.error].includes(workflows.loading)) && (
         <Box align='center' justify='center' margin={{ top: 'small' }}>
           <Box height='xxsmall' width='xxsmall'>
             <Bars
               fill={loaderColor}
-              height='100%'
+              height='80%'
               viewBox='0 0 135 140'
               width='100%'
             />
--- a/packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.js
+++ b/packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.js
@@ -8,17 +8,16 @@ import React from 'react'
 import theme from './theme'
 
 function WorkflowSelectButton (props) {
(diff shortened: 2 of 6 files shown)


eatyourgreens:
Can you replace the button with a yellow banner instead, that says `Nothing to classify`? It seems odd to me to have a disabled button that's never intended to be clicked, and the button label doesn't make much sense in terms of the button action.

If this button were not disabled, what would it do?

rogerhutchings:
What are you picturing? The same thing visually, just as a `div` instead of a `button`?

eatyourgreens:
Yeah, that would be great. 

If it must be a disabled button, I think it should have the same label as it has when it's enabled, but some other change in colour or style to indicate that it's disabled.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/WorkflowSelector.js
```diff
@@ -37,18 +37,26 @@ function WorkflowSelector (props) {
 
       {(workflows.loading === asyncStates.success) && (
         <Box margin={{ top: 'small' }}>
-          {workflows.data.map(workflow =>
+          {(workflows.data.length > 0) && workflows.data.map(workflow =>
             <WorkflowSelectButton key={workflow.id} workflow={workflow} />
           )}
+          {(workflows.data.length === 0) && (
+            <WorkflowSelectButton
```
eatyourgreens: Can this be a styled Box? A button that's never intended to be clicked seems like the wrong element to use here.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/components/WorkflowSelectButton/WorkflowSelectButton.js
```diff
@@ -8,29 +8,34 @@ import React from 'react'
...
-  return (
-    <Link as={as} href={href} passHref>
-      <Button label={workflow.displayName} />
-    </Link>
-  )
+  const { router, workflow, disabled, ...rest } = props
+
+  if (disabled) {
+    return (
+      <Button disabled={true} label={workflow.displayName} {...rest} />
```
eatyourgreens: Sorry, I think I meant can this be a styled Box instead of a button?

rogerhutchings: If we're using a Box, it probably makes more sense to put it in the `WorkflowSelector` component

eatyourgreens: That's true, we might want to disable a workflow button for some other reason.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/Background/Background.js
```diff
@@ -1,39 +1,45 @@
...
 
 const Img = styled.img`
-  height: 100%;
-  object-fit: cover;
-  object-position: 0 50%;
-  width: 100%;
-
-  ${props => props.screenSize === 'small' && `
-    height: auto;
+    height: 230px;
```
eatyourgreens: Has this crept in from another PR?

rogerhutchings: ugh, looks like it

eatyourgreens: Version changes from the grommet update have appeared in this PR too.

rogerhutchings: Okay, I've fixed the history

PR #1168, opened 2019-09-25
Add LCV annotation filter conditional for width greater than zero

mcbouslog:
Package:
- lib-classifier

Describe your changes:
- adds conditional for width > 0 to existing filter on brush annotations before saving brush annotations as classification annotations

Notes:
- I think this could be done in a few different ways, alternatively:
  - checking width on const declaration (line 210), if less than zero then handle then, would keep annotation (as opposed to adding in filter which eliminates entire related annotation)
  - separate filter, arguably easier to read, possible performance downside (and 2 filters on same thing (`raw`) would be strange)
  - I'm sure others
- I didn't see related tests, which I didn't think there were, but might be right in front of me, if so I'll add tests asap
- intentionally as `>` and not `>=`, as I can't think of instance where width would be 0, but of course open to suggestion

# Review Checklist

## General

## Apps

- [ ] Does it work on mobile?

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
@@ -204,10 +204,10 @@ class LightCurveViewer extends Component {
     if (!this.isCurrentTaskValidForAnnotation()) return // Sanity check
 
     const annotations = this.annotationBrushes
-      .filter((raw) => (raw.minX !== undefined && raw.maxX !== undefined && (raw.maxX - raw.minX) > 0))
+      .filter((raw) => (raw.minX !== undefined && raw.maxX !== undefined && raw.minX !== null && raw.maxX !== null))
       .map((raw) => {
         const x = (raw.minX + raw.maxX) / 2
-        const width = (raw.maxX - raw.minX)
+        const width = Math.abs(raw.maxX - raw.minX)
         const toolType = props.currentTask.tools[props.toolIndex].type
         return { x, width, tool: props.toolIndex, zoomLevelOnCreation: raw.zoomLevelOnCreation, toolType }
       })
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
@@ -204,7 +204,7 @@ class LightCurveViewer extends Component {
     if (!this.isCurrentTaskValidForAnnotation()) return // Sanity check
 
     const annotations = this.annotationBrushes
-      .filter((raw) => (raw.minX !== undefined && raw.maxX !== undefined))
+      .filter((raw) => (raw.minX !== undefined && raw.maxX !== undefined && (raw.maxX - raw.minX) > 0))
       .map((raw) => {
         const x = (raw.minX + raw.maxX) / 2
         const width = (raw.maxX - raw.minX)


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
```diff
@@ -204,7 +204,7 @@ class LightCurveViewer extends Component {
     if (!this.isCurrentTaskValidForAnnotation()) return // Sanity check
 
     const annotations = this.annotationBrushes
-      .filter((raw) => (raw.minX !== undefined && raw.maxX !== undefined))
+      .filter((raw) => (raw.minX !== undefined && raw.maxX !== undefined && (raw.maxX - raw.minX) > 0))
```
srallen: I think this is the wrong place to check for this. Other kinds of data might have valid negative x values and negative values would be valid for the x center that we record in the annotation. I think we want the final value of the width to be greater than zero, so I think we can achieve that by getting its absolute value. 

Hypothetically if negative values are valid for the data, let's say minX is -8 and maxX is -6, center x would be -7 and the width as it is currently calculated would be -2 which doesn't make sense, but the absolute value of 2 does.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
```diff
@@ -204,7 +204,7 @@ class LightCurveViewer extends Component {
...
       .map((raw) => {
         const x = (raw.minX + raw.maxX) / 2
         const width = (raw.maxX - raw.minX)
```
srallen: Maybe use `Math.abs()`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/abs

srallen:
As for testing, this is an area of the code base where we're lacking unfortunately. For now, let's focus on getting a fix in that we understand and can manually test. Adding tests for the d3 functions and views is a bit of a chore and might impede getting a quick fix out.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
mcbouslog: I'm very unsure of this, but as is, if minX is -8 and maxX is -6 would width be 2 (-6 - -8)? The part I'm not sure about is how raw.minX/raw.maxX (i.e. d3?) would interpret mins and maxs in negative value data territory.

mcbouslog: Regardless, I think the checks for `null` in the filter, and adding`Math.abs` in width declaration makes more sense (recent commit). Thanks!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
mcbouslog: Are we ok with a width of zero?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
srallen: Sorry yes, my logic is backwards with the min and max (I think I typed out too quickly!), but we can get a negative number if max is -8 and min is -6, so I think taking the absolute value is the right direction. Thanks!

I do think we have some possible issues with d3, but we're short circuiting it by adding a check before adding the annotation to the classification. Investigating what d3 is doing here is a follow up.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
srallen: I _think_ the zeros are coming from a scenario where `maxX` and `minX` are both null, so filtering out null values should prevent this.

srallen: It would be very hard if not impossible for someone to make a mark where subtracting min from max equals zero (max and min are the same value?), so it's more likely the values were null somehow to start with.

PR #1172, opened 2019-09-26
Enable client side routing between home and classify pages

rogerhutchings:
~~🚧 This PR is labelled `WIP` to prevent accidental merging.  🚧~~

This enables client-side routing between the new homepage and the classify page. It should be merged just prior to the homepage going live.

cc #1071


rogerhutchings:
Making this WIP again pending design sign off

PR #1186, opened 2019-09-30
Page-level layout fixes

### LINKED ISSUE #1180: Give content below the fold a max-width
@beckyrother would like a max-width of `1600px` for content below the fold on each page, i.e. everything but the Hero on the home page, and everything but the classifier on the classify page.

rogerhutchings:
- Sets the `max-width` of content below the fold on the classify page to `xxlarge`

Closes #1180 
cc #1152

### Notes

- `xxlarge` is set in the theme to `1536px` ([link](https://github.com/grommet/grommet/blob/master/src/js/themes/base.js#L264))
- This change uses the new `min`/`max` feature in the `Box` width prop: https://v2.grommet.io/box#width


rogerhutchings:
There's a layout fix needed in #1152 as well, so I'm going to roll it into this PR.

beckyrother:
Please post a screenshot or link to staging?

rogerhutchings:
small (500px):

s

medium (1400px):

m

large (2000px)

xl
:

beckyrother:
Thank you Roger! This looks great. One thing I just noticed (very sorry) is the button width on the hero should be a fixed 350px on desktop. As it gets wider, it looks less and less like a button.

rogerhutchings:
@beckyrother There's a predefined `medium` width in our theme that works out as 384px, so I've used that.

FireShot Capture 026 - I Fancy Cats - Zooniverse - People-powered research - localhost

packages/app-project/src/screens/ProjectHomePage/ProjectHomePage.js
```diff
@@ -10,33 +10,49 @@ import ProjectStatistics from '../../shared/components/ProjectStatistics'
...
   return (
-    <>
+    <Box align='center'>
       <ProjectAnnouncement />
       <Hero />
-      <Grid gap='medium' margin='medium'>
-        <Grid columns={responsiveColumns} gap='small'>
-          <ZooniverseTalk />
+
+      <Media at='default'>
```
eatyourgreens: Heads up that I found, while working on the classify page, that this inserts an empty div into the DOM, which can then break the layout in the storybook. `fresnel` styles these empty divs with `display: none` in the SSr'ed page, but that doesn't happen in the storybook.

rogerhutchings: Thanks - I'm okay with it here, since it's a page-level component, but I'd like to solve this for the storybook at some point .

eatyourgreens: Yeah. I think it's only a big problem inside grids, where those extra elements get rendered as empty columns or rows. I can see why `fresnel` have done this, but it's a bit of a hack when you only want one of those `Media` components to render, and the other to output `null`.

beckyrother:
> @beckyrother There's a predefined `medium` width in our theme that works out as 384px, so I've used that. If the width of the column goes below 384px (which can happen, as it's defined as a percentage), the button fills the width.
> 
> FireShot Capture 026 - I Fancy Cats - Zooniverse - People-powered research - localhost

Thanks! Please left-align that button to align with the 'classify' label

packages/app-project/src/screens/ProjectHomePage/ProjectHomePage.js
```diff
@@ -10,33 +10,49 @@ import ProjectStatistics from '@shared/components/ProjectStatistics'
 import ZooniverseTalk from './components/ZooniverseTalk'
 import ThemeModeToggle from '@components/ThemeModeToggle'
 import ProjectAnnouncement from './components/ProjectAnnouncement'
+import { Media } from '../../shared/components/Media'
```
srallen: Non-blocking: I have to second that the naming of this is confusing. We should rename this one to be reflective of it being for responsive layout.

PR #1189, opened 2019-10-01
Make lower footer responsive

rogerhutchings:
The lower footer links (privacy policy, jobs etc.) are always displayed horizontally, which is too wide for smaller screens (e.g. iPhone SE). This makes them render vertically on `small` screen sizes.

Weirdly, I found that trying to do it with `styled-components` (by creating a `StyledBox`) broke the spacing between child items :(

Wide:

FireShot Capture 028 - Storybook - localhost

Narrow:

FireShot Capture 030 - Storybook - localhost


rogerhutchings:
No problem @beckyrother, but I'll bump that one into a separate issue (#1191)

Issue #1151, opened 2019-09-20
Add completeness bar to workflow selection buttons


rogerhutchings:
The design for the workflow select button in the Hero widget includes a completeness bar which needs to be added:

<img width="383" alt="Screenshot 2019-09-20 13 56 54" src="https://user-images.githubusercontent.com/2725841/65328568-85608b00-dbae-11e9-939e-30d8551667c3.png">

Related: #1077 

And https://zooniverse.slack.com/archives/C0Z7YNTSL/p1554151344033200

beckyrother:
(Working on this now)

beckyrother:
Think I was trying to be too clever the first time. How does this feel? 
![workflow-button-3 2](https://user-images.githubusercontent.com/1392130/66015798-9e641880-e499-11e9-9e8a-4f5085a2a0b9.gif)

For reference, here's the first one we were thinking about (pretend the button text is black like it should be):
![workflow-button 1](https://user-images.githubusercontent.com/1392130/66015819-b045bb80-e499-11e9-8eba-b86b2eed799c.gif)

eatyourgreens:
Lose the blue tooltip, and I think this is much better (and something we can implement.)

eatyourgreens:
#1204 adds a basic background gradient, which is muted somewhat on hover/focus. Each link is announced in VoiceOver as something like 'Link: Workflow name 60% complete'.

beckyrother:
If we need to remove the blue tooltip, let's add a secondary line to the button to indicate the percent complete and give a reason that the button is a progress bar. Design tk.

Issue #887, opened 2019-05-29
Standardize buttons


rogerhutchings:
Our designs specify to use a lightened border as the hover state for buttons. However, we're applying this inconsistently, and there are buttons with resizing borders still around. This needs to be set in the theme, and the existing buttons checked for compliance

srallen:
We could better standardize the buttons by leveraging the newer theme API for buttons from Grommet: https://v2.grommet.io/button
https://v2.grommet.io/button#theme

See the description of default, primary, and secondary.

PR #1199, opened 2019-10-01
Add GoldButton to react components, update text color

### LINKED ISSUE #1198: Dark theme gold button hover text color
**Package** lib-classify, app-project

**Describe the bug**

The hover state of the gold button on dark theme is hard to read:

Screen Shot 2019-10-01 at 1 59 11 PM

This was feedback we received from the initial implementation of the button using dark theme from PFE

**To Reproduce**

**Expected behavior**

Text color should be black like currently PFE:

Screen Shot 2019-10-01 at 1 56 40 PM

@beckyrother will be updating the Invision styles.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

srallen:

Package: lib-react-components, lib-classifier, app-project

Closes #1198

Describe your changes:

Abstracts the shared theme into a `GoldButton` component into the react component library rather than have to update themes in two places. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
Has anyone got better ideas for the name of this button? Names should describe what a component does, not what it looks like.

srallen:
The button component is identical to the Grommet `Button`, with just the added custom theming as noted in the updated readme in this PR. `GoldButton` seems aptly named to me. 

The goal here was to eliminate the custom theme duplication we now had between an app and a library so that I wouldn't need to update the same thing in two different places.

srallen:
I forgot to include the screenshots from earlier. The `NextButton` and `WorkflowSelectButton` both use `GoldButton`. Now on hover, they have the black text:

Screen Shot 2019-10-01 at 3 44 57 PM

Screen Shot 2019-10-01 at 3 45 17 PM

eatyourgreens:
The latest designs in #1151 change the colour of this button so that it isn't always gold, and will need to be incorporated here.

eatyourgreens:
The style guide refers to this as a primary button, so maybe `PrimaryButton` is a better name?

srallen:
> The style guide refers to this as a primary button, so maybe PrimaryButton is a better name?

We already have components named for how they're styled in the react components library. The button with teal as its primary color is also listed as a primary button in the style guide. 

> The latest designs in #1151 change the colour of this button so that it isn't always gold, and will need to be incorporated here.

Yes, I see the updated styles, but I don't think that has to be incorporated here in this PR. I see utility in having this added to the react component library and the workflow selector button can be refactored for its now unique styling for #1151 after this. The gold style button is used elsewhere including the collections page design.

rogerhutchings:
Given that this is the primary button, why not update our grommet theme to set the `button.primary.color` property, and use `extend` for any other styles required? We can target the primary variant by using ``${props => prop.primary && `...`}``

eatyourgreens:
I've got a PR incoming to add workflow progress, so please hold off on this.

eatyourgreens:
Can this wait until #1204 is merged?

srallen:
Yes, I'll experiment with @rogerhutchings suggestion to move the styles directly into the theme rather than extension.

beckyrother:
Thank you @rogerhutchings. I'll clarify a bit but some of these questions need to be really sorted as we make a style guide. Can this be something we do while I'm in Oxford, or is that too far away? It would be nice to talk in person.

- Primary buttons can be gold or teal, depending on the context. 
- In dark mode, most buttons are outlined
- Secondary buttons are outlined

srallen:
> Could this also be achieved by using `<Button primary color='gold'>..</Button>` or `<Button primary color='teal'>..</Button>`
> Are there any changes in here that couldn't be achieved by updating the theme for Buttons?

I attempted to incorporate the variants into the main theme itself, but it didn't really work because of some underlying assumptions Grommet is making. If you define the `color` prop on the button, Grommet assumes that that is the only color you want to use and will default to it regardless of `dark` being true (see: https://github.com/grommet/grommet/blob/758b1b5619a93e3928dee4aeadc5830587250996/src/js/components/Button/StyledButton.js#L28-L35)

So I attempted of using a custom prop, `buttonColor`, to set the intended color for the primary button theme, but couldn't get certain theme properties to work with a function, so we could still support light and dark variants as needed (rather than a string or object of strings for light/dark variants). I attempted to move all of it into a custom `extend` function, but found I was added _a lot_ of ternaries since the gold and teal variants aren't _exactly_ the same when they're light vs dark. (For example: gold uses black text color for light and dark theme. Teal uses white for light and teal for dark)

After this, I decided to just revert to the original idea, to add a `GoldButton` variant with its own custom theme extension to the react components library, and leave the teal brand color as the primary default in the main theme.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButton.js
```diff
@@ -9,21 +8,16 @@ import en from './locales/en'
...
-)
-
 function NextButton (props) {
   const { autoFocus, disabled, onClick } = props
+  const label = counterpart('NextButton.next')
   return (
-    <Button
-      a11yTitle={counterpart('NextButton.next')}
+    <GoldButton
+      a11yTitle={label}
```
eatyourgreens: This shouldn't duplicate the button label. It looks like you can remove it, since the button has a visible label.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButton.spec.js
```diff
@@ -41,12 +41,12 @@ describe('NextButton', function () {
...
 
     it('should not be disabled if props.disabled is false', function () {
-      expect(wrapper.find('WithTheme(Button)').props().disabled).to.be.false()
+      expect(wrapper.props().disabled).to.be.false()
     })
 
     it('should be disabled if props.disabled is true', function () {
       wrapper.setProps({ disabled: true })
-      expect(wrapper.find('WithTheme(Button)').props().disabled).to.be.true()
+      expect(wrapper.props().disabled).to.be.true()
```
eatyourgreens: I'm not really sure what this is testing, since `props.disabled` is set to `true` on the previous line.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButton.spec.js
```diff
@@ -41,12 +41,12 @@ describe('NextButton', function () {
     })
 
     it('should not be disabled if props.disabled is false', function () {
-      expect(wrapper.find('WithTheme(Button)').props().disabled).to.be.false()
+      expect(wrapper.props().disabled).to.be.false()
```
eatyourgreens: This doesn't seem to be testing anything either.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButton.spec.js
```diff
@@ -57,12 +57,12 @@ describe('NextButton', function () {
     })
 
     it('should not be auto-focused if props.autoFocus is false', function () {
-      expect(wrapper.find('WithTheme(Button)').props().autoFocus).to.be.false()
+      expect(wrapper.props().autoFocus).to.be.false()
```
eatyourgreens: Same comments here about testing the values of props. I'm not really sure what that's testing.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButton.spec.js
```diff
@@ -57,12 +57,12 @@ describe('NextButton', function () {
...
     })
 
     it('should be auto-focused if props.autoFocus is true', function () {
       wrapper.setProps({ autoFocus: true })
-      expect(wrapper.find('WithTheme(Button)').props().autoFocus).to.be.true()
+      expect(wrapper.props().autoFocus).to.be.true()
```
eatyourgreens: Same here about testing a prop that was set on the previous line.

rogerhutchings:
Non-blocking: there's one other approach that might make sense, and that's to make this a `PrimaryButton` rather than `GoldButton`. It then becomes style-agnostic, and could accept a `color` prop

srallen:
@rogerhutchings is that the same suggestion as before? It's a good idea, but as I found there are some assumptions that Grommet is making that make this difficult to do. If the `color` prop is defined, then we can't vary what is applied in the light vs dark theme because Grommet assumes you just want the button to only be that color now. I can bypass that by using a unique prop, say `buttonColor`, however the `primary.color`, `color`, and `border` properties in theme also throw errors when you try to use a function to use the component props. The function gets passed into the Grommet's utility normalize function that then throws because of assumptions being made there. Unfortunately I don't have a good suggestion to make this more flexible, unless we decide to bypass themes altogether and just use styled-components directly to style these.

@eatyourgreens can I address those comments separately in another PR? I'd like to keep this PR scoped to just adding a new reusable component rather than refactor the `NextButton` and its tests.

eatyourgreens:
The duplicated label needs removing before this is mergedbut the comments on the tests were just questions because I don't really follow what those tests are doing.

srallen:
Ok I've removed `a11yTitle` from the `NextButton`

rogerhutchings:
Not quite. You could set up a button component for each button style, one teal, one gold; whether those are themed grommet Buttons or styled components, your call. You'd then choose which one to use based on the `color` prop to PrimaryButton, and pass the rest of the props to the child Button component.

So we'd only use our PrimaryButton for primary button components, which we know would be styled correctly, and we wouldn't have to be getting a component from a different place just because it's a different colour

srallen:
@rogerhutchings Ah, ok, thanks for the clarification. That could be a subsequent future enhancement and we could incorporate a fix for #887

eatyourgreens:
Re. naming, my only concern (from previous experience), is that a future design change alters the colour of these buttons and we have a `GoldButton` component for buttons that aren't gold.

PR #1212, opened 2019-10-07
Add a task registry

### LINKED ISSUE #1141: Text task
**Package** lib-classifier

The text task needs to be added w/ the buttons for inserting the metadata values like deletion

**To Dos**

- [x] Simple text task using a textarea element
- [x] Metadata buttons to insert metadata values

eatyourgreens:
Add a registry to hold references to task components and models.
Remove named imports from the task area, workflow stores and classification stores.
 Load in task components and models dynamically from the registry.

~~This is based off #1232 at the moment.~~

Package:
lib-classifier

Closes #1141.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
```diff
@@ -9,6 +9,8 @@ import getTaskComponent from './helpers/getTaskComponent'
 import TaskHelp from './components/TaskHelp'
 import TaskNavButtons from './components/TaskNavButtons'
 
+import externalTasks from '../../../../../../plugins/tasks'
```
eatyourgreens: This could be aliased to `@plugins` for convenience.

packages/lib-classifier/src/plugins/tasks/index.js
```diff
@@ -0,0 +1,5 @@
+import TextTask from './TextTask'
+
+export default {
+  text: TextTask
+}
```
eatyourgreens: Maybe this should be a Map?

beckyrother:
Is this the same as the text modifiers, and if so, please post a screenshot of the text input with the modifier buttons?

srallen:
@beckyrother this is an initial setup for the data side of things, no design has been applied yet.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextAnnotation.js
```diff
@@ -0,0 +1,8 @@
+import { types } from 'mobx-state-tree'
+
+const TextAnnotation = types.model('TextAnnotation', {
+  task: types.identifier,
+  value: types.maybeNull(types.string)
```
srallen: This should default to be a an empty string since this is what is used by the classifier to add a default annotation when a value isn't filled in and the task is not required. An empty string was used for the default (https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/text/index.jsx#L148) and I don't think there's any reason to change this.

So, `types.optional(types.string, '')` instead.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextTask.js
```diff
@@ -0,0 +1,12 @@
+import { types } from 'mobx-state-tree'
+
+const TextTask = types.model('TextTask', {
+  help: types.optional(types.string, ''),
+  instruction: types.string,
+  required: types.maybe(types.boolean), // Should this be an optional type with the default to true?
```
srallen: There definitely should be a default here. Why would the default be true and not false? The project builder doesn't default to required being true.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextAnnotation.js
eatyourgreens: Good point.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextTask.js
eatyourgreens: I think I copied this comment over from the single choice task model by mistake.

eatyourgreens:
I’m not wild about the naming but I used ‘core’ for tasks that are embedded into the main classifier code. I had tried extracting the question tasks, as an experiment, but that broke the classifier. 

I think it would be safer to build out new tasks and move older code over when we’re sure that we won’t break anything.

eatyourgreens:
I've added a simple registry, with tests, to hold references to the task components and models. I've also renamed all the `core…` stuff to `legacy…` instead.

srallen:
>  I had tried extracting the question tasks, as an experiment, but that broke the classifier.

Out of curiosity, any idea what broke?

eatyourgreens:
Dependencies that were imported from relative paths, I think. I’m sure they can be moved, but it isn’t the quick cut-and-paste that I tried.

eatyourgreens:
This isn't styled yet but this is what it looks like with three transcription tasks on the screen:
Screenshot of a prototype project with three text tasks in one step.

srallen:
@eatyourgreens would you be willing to write up an ADR on this proposed architectural change? Like I'm said, I'm supportive of the refactor proposed here in general, but I'm concerned with having both the legacy and this new plugin architecture in the code at the same time without a documented explanation somewhere.

srallen:
Also, I noticed the code to convert combo tasks to steps, and is there a reason this is being added to this PR? Again, seems like a separate refactor that's not necessary for adding the text task.

eatyourgreens:
The text task is usually used inside combo tasks. I don't think you can add text transcription without  converting those old combo tasks to workflow steps.

eatyourgreens:
>We so far have been building the new classifier to just deprecate the combo task and have no support for it.

That's why I've replaced the combo with workflow steps here.

srallen:
I think adding backwards compatibility is a discussion we should have at some point, but separately from adding the text task. Right now the decision is no support for it. Backward compatibility is a kind of support for it. Can this wait and let's have an issue opened up for discussion for it?

eatyourgreens:
I'd rather just get it out of the way. It's a very small change (maybe ten lines of code at most?), and I've written it now.

eatyourgreens:
I'm happy to write an ADR, but we need an idea of what shape new tasks should take before we can write that. The idea of this PR is to get some feedback on what shape task design should take.

eatyourgreens:
This is all set up now to add wrap text selections in tags and should also be updating the classifications store. Do we want to update the store on every key press, or update annotations once typing is finished (maybe when focus leaves the textarea?)

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextTask.js
```diff
@@ -0,0 +1,27 @@
...
+    if (currentAnnotations && currentAnnotations.size > 0) {
+      return currentAnnotations.get(self.taskKey)
+    }
+    return undefined
+  }
+}))
+.actions(self => ({
+  updateAnnotation(value) {
+    const { addAnnotation } = getRoot(self).classifications
+    addAnnotation(value, self)
```
eatyourgreens: I find it very confusing that we update an annotation by calling `classifications.addAnnotation(value, task)`.

packages/lib-classifier/src/store/Classification.js
```diff
@@ -1,6 +1,7 @@
-import { types, getType } from 'mobx-state-tree'
+import { types, getRoot, getType } from 'mobx-state-tree'
```
eatyourgreens: I don't think `getRoot` is used anywhere.

packages/lib-classifier/src/store/Classification.js
```diff
@@ -36,15 +37,12 @@ const ClassificationMetadata = types.model('ClassificationMetadata', {
 
 const Classification = types
   .model('Classification', {
-    annotations: types.map(types.union({
-      dispatcher: (snapshot) => {
-        const snapshotType = getType(snapshot)
```
eatyourgreens: `getType` can only be called on a tree node, not a snapshot, so I've had to remove this.

packages/lib-classifier/src/plugins/tasks/Registry.js
```diff
@@ -0,0 +1,29 @@
+class Registry {
+  constructor () {
+    this.register = {}
+  }
+
+  get (key) {
+    return this.register[key] || {}
```
rogerhutchings: what's the benefit to returning an empty object here?

srallen:
Can the plugin architecture be attempted again for the single choice task in a separate PR? I would feel way more comfortable with this if we had an idea of what the migration path would look like and the dev time involved. Like I said before, I'm for the general idea, but I'm uncomfortable with not knowing what's involved with refactoring the other tasks into plugins. Once we know, we can write up an ADR and add issues for people to pick up when they can for the rest of the tasks.

eatyourgreens:
My worry is that adding the Text task to the classifier as is means adding more bad code, which we have to go back and rewrite later. #1228 might help with that. 

I’m happier with a development path that adds new features without breaking existing code (particularly without breaking PH-TESS.) My suggestion would be to flag existing task loading code as deprecated (mostly the big lists of imports and if statements in the stores), use dynamic loading for new tasks, then move deprecated code over as and when time is available.

The question tasks do sound like good subjects, since nothing’s using them yet. The data visualisation task makes me more nervous.

packages/lib-classifier/src/plugins/tasks/Registry.js
eatyourgreens: `registry.get(type).TaskModel` will be undefined rather than crashing the classifier.

eatyourgreens:
I should add that I took a lightweight approach here so that it doesn’t hold up development.

srallen:
The current code isn't great, but we know it works and it is mostly tested. I'd rather not be in a situation like PFE's code base where we have refactors half completed that never get finished and creates a more complex code base. Ultimately the deadlines we have for the grants matter more at the moment. I agree that whatever ends up being done that the data vis task should be last to be touched. If we can have a proof of concept for an existing task, a documented path forward, and a time estimate, then I'll be supportive of the move.

eatyourgreens:
No problem breaking the Text task out into its own PR. I’ll probably base that off of #1228, since the Text task code makes the task model responsible for managing its own annotation. 

I’m assuming only one annotation per task at the moment but I’ve opened an issue about that since it isn’t true for transcription projects eg. many lines on a page sharing the same text input task.

srallen:
Thank for being open to separating these out. Initial glance at #1228 looks good so far. About the annotations, I think you're right that it'll have to be an array. I'll comment in the issue for that.

eatyourgreens:
One thing I could do is write up the current process for adding a new task, and notes on how it could be improved.

eatyourgreens:
@srallen @rogerhutchings I've updated this to migrate all existing tasks. That turned out to be quicker than only moving the question tasks.

packages/lib-classifier/src/store/ClassificationStore.js
```diff
@@ -108,23 +108,12 @@ const ClassificationStore = types
...
-      return taskTypes[taskType] || undefined
-    }
-
     function createDefaultAnnotation (task) {
       const validClassificationReference = isValidReference(() => self.active)
 
       if (validClassificationReference) {
         const classification = self.active
-        const annotationModel = getAnnotationType(task.type)
+        const annotationModel = taskRegistry.get(task.type).AnnotationModel
```
eatyourgreens: #1228 removes the need for this.

packages/lib-classifier/src/plugins/tasks/Registry.js
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  add (key, value) {
+    if (this.register[key]) {
+      throw new Error(`Registry error: key ${key} is already registered.`)
```
srallen: This is a bit of a wild idea since I've never used this myself in production code, but how about using a set which enforces uniqueness? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Support across browsers looks pretty good.

packages/lib-classifier/src/plugins/tasks/Registry.js
```diff
@@ -0,0 +1,29 @@
+class Registry {
+  constructor () {
+    this.register = {}
```
srallen: Maybe this can be a set rather than an object? See my comment below about uniqueness.

eatyourgreens: Sounds good to me. I’ll do some reading. When I wrote this class, I thought I was basically creating a Map that errors if you try to set an existing key.

eatyourgreens:
Agh. Conflicts in a PR that moves all the task files. I’ll take a look at that tomorrow. 

Good point about being tied to MST. Ideally we could replace the stores without changing any of the views. I don’t have a good answer for that. How we organise the file system might depend mainly on whether we want to move tasks out into their own library.

eatyourgreens:
I've rebased this to master and removed the text task, so it should be more-or-less independent of #1232 now.

srallen:
I'm going to hold off rebasing this until #1232 merges.

eatyourgreens:
I’ve rebased this to latest master.

eatyourgreens:
I've updated this to include the text task and the fix for the task navigation buttons.

packages/lib-classifier/src/plugins/tasks/index.js
```diff
@@ -0,0 +1,23 @@
...
+  taskRegistry.add('text', TextTask)
+} catch (e) {
+  console.log(e.message)
+}
+
+const taskModels = taskRegistry.values('TaskModel')
+const annotationModels = taskRegistry.values('AnnotationModel')
+export default taskRegistry
+export { taskModels }
+export { annotationModels }
```
srallen: Non-blocking, but this can be written as `export { taskModels, annotationModels }` in one line.

rogerhutchings: Or even `export { taskRegistry as default, taskModels, annotationModels }`

eatyourgreens:
>I still see an empty DrawingTask folder in the original Tasks/component/ file structure.

Here's that directory on this branch.
https://github.com/zooniverse/front-end-monorepo/tree/d33ec1a3b4157682e111455ffc3589fa268007a3/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components

packages/lib-classifier/src/plugins/tasks/index.js
eatyourgreens: I think elsewhere we always export the default on a separate line.

eatyourgreens:
Good points about documentation. I've started a readme with basic instructions for adding a new task.

eatyourgreens:
Here's a first pass at [instructions for adding a new task](https://github.com/zooniverse/front-end-monorepo/blob/0695fe45158eddc49514cc95aab6aa04d69d18b8/packages/lib-classifier/src/plugins/tasks/readme.md).

eatyourgreens:
@rogerhutchings is this ok to merge?

rogerhutchings:
👍

PR #1219, opened 2019-10-10
Create steps from combo tasks

eatyourgreens:
Update `convertWorkflowToUseSteps` to create steps from `task.tasks` in combo tasks.
Update `setTasks` to log unsupported tasks and continue, rather than failing at the first unsupported task in a workflow.
Add a combo task to the tests when a workflow has no defined steps.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/WorkflowStepStore.js
+++ b/packages/lib-classifier/src/store/WorkflowStepStore.js
@@ -177,7 +177,7 @@ const WorkflowStepStore = types
       const taskKeys = Object.keys(workflow.tasks)
       const { first_task } = workflow
 
-      function comboToStepTasks (task) {
+      function getStepTasksFromCombo (task) {
         task.tasks.forEach(function (taskKey) {
           taskKeys.splice(taskKeys.indexOf(taskKey), 1)
         })
@@ -192,7 +192,7 @@ const WorkflowStepStore = types
 
         if (workflow.tasks[first_task].type === 'combo') {
           const combo = workflow.tasks[first_task]
-          firstStep.taskKeys = comboToStepTasks(combo)
+          firstStep.taskKeys = getStepTasksFromCombo(combo)
         }
 
         taskKeys.splice(taskKeys.indexOf(first_task), 1)
@@ -204,7 +204,7 @@ const WorkflowStepStore = types
         if (task.type !== 'shortcut') {
           let stepTasks = [taskKey]
           if (task.type === 'combo') {
-            stepTasks = comboToStepTasks(task)
+            stepTasks = getStepTasksFromCombo(task)
           }
           self.steps.put({
             stepKey: `S${index + 1}`,
--- a/packages/lib-classifier/src/store/WorkflowStepStore.js
+++ b/packages/lib-classifier/src/store/WorkflowStepStore.js
@@ -164,29 +164,51 @@ const WorkflowStepStore = types
         // put is a MST method, not native to ES Map
         // the key is inferred from the identifier type of the target model
         const taskToStore = Object.assign({}, workflow.tasks[taskKey], { taskKey })
-        self.tasks.put(taskToStore)
+        try {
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Testing the text task in #1212 against real projects, I found that they use multiple text tasks in combination. I made some quick changes to the tasks -> steps conversion function, which create a new workflow step from the tasks listed in a combo.
Screenshot of a Notes from Nature project with two text tasks in one step.

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -164,29 +164,51 @@ const WorkflowStepStore = types
         // put is a MST method, not native to ES Map
         // the key is inferred from the identifier type of the target model
         const taskToStore = Object.assign({}, workflow.tasks[taskKey], { taskKey })
-        self.tasks.put(taskToStore)
+        try {
+          self.tasks.put(taskToStore)
+        }
+        catch (e) {
+          console.log(`${taskKey} ${taskToStore.type} is not a supported task type`)
```
rogerhutchings: This would be better logged to sentry - packages/app-project/src/helpers/logger/logReactError.js

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -164,29 +164,51 @@ const WorkflowStepStore = types
...
+        }
       })
     }
 
     function convertWorkflowToUseSteps (workflow) {
       const taskKeys = Object.keys(workflow.tasks)
+      const { first_task } = workflow
+
+      function comboToStepTasks (task) {
```
rogerhutchings: - This could be a pure function
- Having `taskKey` and `taskKeys` refer to different things in different scopes is a little confusing

packages/lib-classifier/src/store/WorkflowStepStore.js
eatyourgreens: Can that be used in the classifier?

packages/lib-classifier/src/store/WorkflowStepStore.js
eatyourgreens: `task.tasks` (returned from the API) is an array of task keys. I'm not sure what I can do to make the naming clearer.

The tests are written for individual methods, so I've kept the changes here inside the `convertWorkflowToUseSteps` method.

packages/lib-classifier/src/store/WorkflowStepStore.js
eatyourgreens: How does that error logger work without throwing an exception that causes the store to fail at the first unsupported task? The code here is supposed to store all supported tasks from the workflow, while skipping over unsupported tasks like the combo task (which is instead converted to a workflow step.)

packages/lib-classifier/src/store/WorkflowStepStore.js
eatyourgreens: Looking at this again, the helper function is removing those combo tasks from the `taskKeys` array, so it can't run outside the scope of `convertWorkflowToUseSteps`.

packages/lib-classifier/src/store/WorkflowStepStore.js
rogerhutchings: Oh, that's right; that's annoying. At some point I'd like to be able to pass in a logger function to the classifier so it can be handled by the consuming app.

Basically, I'd want sentry to log an error if we have a project running on the classifier that it can't support, but that's outside this scope.

eatyourgreens: That's come up with feedback too, where we'd like to report if there's a mistake in the subject metadata that breaks feedback, without completely crashing the classifier.

PR #1228, opened 2019-10-14
Add a default annotation to tasks

eatyourgreens:
This PR delegates responsibility for creating new annotations to individual tasks, with the aim of simplifying the classifications store and making it easier to add new tasks.

 - Add a default annotation to each task model.
 - Add a generic createAnnotation action to all tasks.
 - Remove getAnnotationType and createDefaultAnnotation from the classifications store.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/tasks/DataVisAnnotationTask.js
+++ b/packages/lib-classifier/src/store/tasks/DataVisAnnotationTask.js
@@ -1,8 +1,9 @@
 import { types } from 'mobx-state-tree'
 import Task from './Task'
 import { Graph2dRangeXTool } from './dataVisTools'
+import { DataVisAnnotation } from '../annotations'
 
-const DataVisAnnotation = types.model('DataVisAnnotation', {
+const DataVisTask = types.model('DataVisTask', {
   help: types.optional(types.string, ''),
   instruction: types.maybe(types.string),
   tools: types.array(types.union({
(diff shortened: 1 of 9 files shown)


eatyourgreens:
Tests might be failing because mock tasks need updating with the new actions and properties.

eatyourgreens:
Tests are passing and I can answer questions in the dev classifier so I think this is ready for a review.

packages/lib-classifier/src/store/tasks/Task.js
```diff
@@ -3,5 +3,11 @@ import { types } from 'mobx-state-tree'
 const Task = types.model('Task', {
   taskKey: types.identifier
 })
+.actions(self => ({
+  createAnnotation () {
```
eatyourgreens: Maybe this should be `createAnnotation(value)` where `value` is optional.

packages/lib-classifier/src/store/tasks/DataVisAnnotationTask.js
```diff
@@ -1,8 +1,9 @@
 import { types } from 'mobx-state-tree'
 import Task from './Task'
 import { Graph2dRangeXTool } from './dataVisTools'
+import { DataVisAnnotation } from '../annotations'
 
-const DataVisAnnotation = types.model('DataVisAnnotation', {
+const DataVisTask = types.model('DataVisTask', {
```
srallen: Since we've been talking about naming, how about `DataVisTaskModel` here?

eatyourgreens: Good point.

eatyourgreens:
In my head, a classification (or drawing) gives an annotation to a task. The task makes changes to annotation, then hands it back to the source. That’s where the actions came from. I hadn’t thought about annotations being responsible for doing work. I was thinking of them as parcels of data that get handed around by the other models.

srallen:
Interesting! I had conceptually a different model in my head. Tasks are read only representations of what we receive with the workflow resource telling the classifier what to render and annotations are the mutable data that is passed around. But I realized when I was trying to follow this code around that the annotations don't know how to update themselves without tasks, so, it gets a bit confusing!

eatyourgreens:
Annotations are coupled to tasks, though, so an annotation model could import a task model that gives it a context in which to update itself. A multiple choice answer is never going to be updated by a drawing task, for example. 

I think the model could work both ways, so maybe comes down to which route is easiest to implement.

eatyourgreens:
I've added some tests for the Task model.

PR #1232, opened 2019-10-15
Add a text input task

### LINKED ISSUE #1141: Text task

eatyourgreens:
Adds the text task from #1212 without any of the changes to component loading. The new task code is manually added to each component or store that depends upon it.

I've kept the file structure from #1212 for convenience, but that can be changed.

EDIT: functionally, this works but it needs styling. 

Package:
lib-classifier

Closes #1141.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
@@ -45,7 +45,7 @@ class Tasks extends React.Component {
       // there has to be a better way
       // but works for now
       return (
-        <Box as='form' justify='between' fill>
+        <Box as='form' gap="small" justify='between' fill>
           {tasks.map((task) => {
             const TaskComponent = getTaskComponent(task.type)
             if (TaskComponent) {
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
@@ -3,7 +3,6 @@ import { Box, Paragraph } from 'grommet'
 import { inject, observer } from 'mobx-react'
 import PropTypes from 'prop-types'
 import React from 'react'
-import { ThemeProvider } from 'styled-components'
 
 import getTaskComponent from './helpers/getTaskComponent'
 import TaskHelp from './components/TaskHelp'
@@ -46,24 +45,22 @@ class Tasks extends React.Component {
       // there has to be a better way
       // but works for now
       return (
-        <ThemeProvider theme={{ mode: this.props.theme }}>
-          <Box as='form' justify='between' fill>
-            {tasks.map((task) => {
-              const TaskComponent = getTaskComponent(task.type)
-              if (TaskComponent) {
-                return (
-                  <Box key={task.taskKey} basis='auto'>
-                    <TaskComponent disabled={!ready} task={task} {...this.props} />
-                  </Box>
-                )
-              }
+        <Box as='form' justify='between' fill>
+          {tasks.map((task) => {
+            const TaskComponent = getTaskComponent(task.type)
+            if (TaskComponent) {
+              return (
+                <Box key={task.taskKey} basis='auto'>
+                  <TaskComponent disabled={!ready} task={task} {...this.props} />
+                </Box>
+              )
+            }
 
-              return (<Paragraph>Task component could not be rendered.</Paragraph>)
-            })}
-            {isThereTaskHelp && <TaskHelp tasks={tasks} />}
-            <TaskNavButtons disabled={!ready} />
-          </Box>
-        </ThemeProvider>
+            return (<Paragraph>Task component could not be rendered.</Paragraph>)
+          })}
+          {isThereTaskHelp && <TaskHelp tasks={tasks} />}
+          <TaskNavButtons disabled={!ready} />
+        </Box>
       )
     }
 
@@ -80,16 +77,14 @@ Tasks.propTypes = {
   isThereTaskHelp: PropTypes.bool,
   loadingState: PropTypes.oneOf(asyncStates.values),
   ready: PropTypes.bool,
-  tasks: PropTypes.arrayOf(PropTypes.object),
-  theme: PropTypes.string
+  tasks: PropTypes.arrayOf(PropTypes.object)
 }
 
 Tasks.defaultProps = {
   isThereTaskHelp: false,
   loadingState: asyncStates.initialized,
   ready: false,
-  tasks: [],
-  theme: 'light'
+  tasks: []
 }
 
 @inject(storeMapper)
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.stories.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.stories.js
@@ -4,7 +4,7 @@ import { storiesOf } from '@storybook/react'
 import zooTheme from '@zooniverse/grommet-theme'
 import React from 'react'
 import sinon from 'sinon'
-import { Grommet } from 'grommet'
+import { Box, Grommet } from 'grommet'
 import { Provider } from "mobx-react"
 
 import { Tasks } from './Tasks'
@@ -24,8 +24,38 @@ const mockStore = {
   }
 }
 
+function MockTask({ dark, isThereTaskHelp, subjectReadyState, step, store, tasks, zooTheme}) {
+  const background = dark ?
+    zooTheme.global.colors['dark-1'] :
+    zooTheme.global.colors['light-1']
+  return (
+    <Provider classifierStore={store}>
+      <Grommet theme={Object.assign({}, zooTheme, { dark })}>
+        <Box
+          background={background}
+          pad='1em'
+          width='380px'
+        >
+          <Tasks
+            isThereTaskHelp={isThereTaskHelp}
+            loadingState={asyncStates.success}
+            step={step}
+            subjectReadyState={subjectReadyState}
+            tasks={tasks}
+          />
+        </Box>
+      </Grommet>
+    </Provider>
+  )
+}
+
 storiesOf('Tasks', module)
 .addDecorator(withKnobs)
+.addParameters({
+  viewport: {
+    defaultViewport: 'responsive'
+  }
+})
 .add('loading', function () {
   return (
     <Provider classifierStore={mockStore}>
@@ -51,6 +81,7 @@ storiesOf('Tasks', module)
   }]
   const dark = boolean('Dark theme', false)
   const subjectReadyState = select('Subject loading', asyncStates, asyncStates.success)
+  const isThereTaskHelp = boolean('Enable task help', true)
   const store = Object.assign({}, mockStore, {
     workflows: {
       loadingState: asyncStates.success
@@ -63,17 +94,16 @@ storiesOf('Tasks', module)
     }
   })
   return (
-    <Provider classifierStore={store}>
-      <Grommet theme={Object.assign({}, zooTheme, { dark })}>
-        <Tasks
-          isThereTaskHelp={true}
-          loadingState={asyncStates.success}
-          step={step}
-          subjectReadyState={subjectReadyState}
-          tasks={tasks}
-        />
-      </Grommet>
-    </Provider>
+    <MockTask
+      dark={dark}
+      isThereTaskHelp={isThereTaskHelp}
+      loadingState={asyncStates.success}
+      step={step}
+      store={store}
+      subjectReadyState={subjectReadyState}
+      tasks={tasks}
+      zooTheme={zooTheme}
+    />
   )
 })
 .add('multiple tasks', function () {
@@ -102,6 +132,7 @@ storiesOf('Tasks', module)
   ]
   const dark = boolean('Dark theme', false)
   const subjectReadyState = select('Subject loading', asyncStates, asyncStates.success)
+  const isThereTaskHelp = boolean('Enable task help', true)
   const store = Object.assign({}, mockStore, {
     workflows: {
       loadingState: asyncStates.success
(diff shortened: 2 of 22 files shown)


packages/lib-classifier/src/plugins/tasks/TextTask/models/TextTask.js
```diff
@@ -0,0 +1,28 @@
...
+  help: types.optional(types.string, ''),
+  instruction: types.string,
+  required: types.optional(types.boolean, false),
+  taskKey: types.identifier,
+  text_tags: types.array(types.string),
+  type: types.literal('text')
+})
+  .views(self => ({
+    get annotation () {
+      const { currentAnnotations } = getRoot(self).classifications
```
eatyourgreens: I don’t like the coupling with the classifications model here. The annotation could come from a classification or a drawing mark.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextTask.js
```diff
@@ -0,0 +1,28 @@
...
+      if (currentAnnotations && currentAnnotations.size > 0) {
+        return currentAnnotations.get(self.taskKey)
+      }
+      return TextAnnotation.create({ task: self.taskKey })
+    }
+  }))
+  .actions(self => ({
+    updateAnnotation (value) {
+      const { addAnnotation } = getRoot(self).classifications
```
eatyourgreens: Same comment about coupling the task to the classifications model.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.stories.js
```diff
@@ -128,4 +128,47 @@ storiesOf('Tasks', module)
...
+  return (
+    <Provider classifierStore={store}>
+      <Grommet theme={Object.assign({}, zooTheme, { dark })}>
+        <Tasks
+          isThereTaskHelp={true}
+          loadingState={asyncStates.success}
+          step={step}
+          subjectReadyState={subjectReadyState}
+          tasks={tasks}
+          theme={dark ? 'dark' : 'light'}
```
srallen: I don't think it should be necessary to set the theme as a prop here since it's set in the Grommet component which should inject it into context.

packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/TextTask.js
```diff
@@ -0,0 +1,75 @@
+import { PlainButton } from '@zooniverse/react-components'
+import { Box, FormField, TextArea } from 'grommet'
+import React from 'react'
+
+function TextTask (props) {
+  const { autoFocus, disabled, task } = props
+  const defaultValue = task.annotation ? task.annotation.value : ''
```
srallen: Can we have a default value set in props and/or isn't there a default set in the store model?

packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/TextTask.js
```diff
@@ -0,0 +1,75 @@
...
+      {task.text_tags.map(tag => (
+        <PlainButton
+          key={tag}
+          disabled={disabled}
+          justify='start'
+          onClick={setTagSelection}
+          text={tag}
+          value={tag}
+        />
+      ))}
```
srallen: Maybe wrap this in a `Box` with the direction set to row for display of multiple buttons if enabled?

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextAnnotation.js
```diff
@@ -0,0 +1,8 @@
+import { types } from 'mobx-state-tree'
+
+const TextAnnotation = types.model('TextAnnotation', {
+  task: types.identifier,
```
srallen: There's a general `Annotation` model that sets this and the other task types are composed with general model. I still see value in using composition to make sure we don't repeat ourselves. What do you think of continuing this with the plugin architecture? Since it's just this one part of the model, I don't have a strong opinion to continue this, but just want to get thoughts on it.

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -3,26 +3,39 @@ import { addDisposer, getRoot, isValidReference, onAction, types } from 'mobx-st
...
 import Step from './Step'
 import { DrawingTask, DataVisAnnotationTask, MultipleChoiceTask, SingleChoiceTask } from './tasks'
+import { TextModel } from '@plugins/tasks/TextTask'
+
+function taskDispatcher (snapshot) {
+  if (snapshot.type === 'drawing') return DrawingTask
+  if (snapshot.type === 'multiple') return MultipleChoiceTask
+  if (snapshot.type === 'single') return SingleChoiceTask
+  if (snapshot.type === 'dataVisAnnotation') return DataVisAnnotationTask
+  if (snapshot.type === 'text') return TextModel
```
srallen: Just a minor quibble on naming here, but maybe `TextTaskModel` just for clarity?

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.stories.js
eatyourgreens: I was wondering if this could be removed.

packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/TextTask.js
eatyourgreens: This is initially undefined because the annotation isn’t created until the first call to `addAnnotation()`. Using the model makes much more sense. #1228 tries to make tasks responsible for their default annotations.

packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/TextTask.js
eatyourgreens: Thanks! I didn’t know how to style this.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextAnnotation.js
eatyourgreens: Composition is definitely the way to go. If the task code moves, I think the shared code has to go with it.

I’ve started adding common views and actions to the base task model in a couple of other PRs.

packages/lib-classifier/src/store/WorkflowStepStore.js
eatyourgreens: Yeah, the naming is not consistent at all, particularly where models and views have the same name.

eatyourgreens:
@srallen can you see why the font size and text styling is wrong inside the textarea? I think I’m missing something obvious.

eatyourgreens:
Styling still needs work but here's what it looks like.
Screenshot of a test workflow with three text tasks and some text entered to show the font size and weight.

packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/TextTask.js
eatyourgreens: I added `task.defaultAnnotation` as a view in the model and return that if the stored annotation is undefined.

packages/lib-classifier/src/store/WorkflowStepStore.js
eatyourgreens: The worst example might be `DataVisAnnotation` and `DataVisAnnotationTask`.

eatyourgreens:
Updated styling. I've overridden the default font weight for all form inputs, so this will also change the 'Add to collections' inputs and the login/register inputs.
Screenshot showing a test workflow with improved text styling and spacing for the text inputs.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
```diff
@@ -46,24 +45,22 @@ class Tasks extends React.Component {
...
-            {tasks.map((task) => {
-              const TaskComponent = getTaskComponent(task.type)
-              if (TaskComponent) {
-                return (
-                  <Box key={task.taskKey} basis='auto'>
-                    <TaskComponent disabled={!ready} task={task} {...this.props} />
-                  </Box>
-                )
-              }
+        <Box as='form' gap="small" justify='between' fill>
```
eatyourgreens: This doesn't need to be a `form` tag.

packages/lib-classifier/src/store/Classification.js
```diff
@@ -36,15 +37,12 @@ const ClassificationMetadata = types.model('ClassificationMetadata', {
 
 const Classification = types
   .model('Classification', {
-    annotations: types.map(types.union({
-      dispatcher: (snapshot) => {
-        const snapshotType = getType(snapshot)
```
eatyourgreens: `getType` can't be run on a snapshot, only a tree node. This dispatcher works because we pass in a tree node as the snapshot. `getType` is the default dispatcher, so I've removed this function because it doesn't really do anything.

beckyrother:
Is this ready for design review?

srallen:
@beckyrother yes I think so

beckyrother:
@srallen is this on Storybook or should I go from @eatyourgreens's screenshot: https://github.com/zooniverse/front-end-monorepo/pull/1232#issuecomment-543116158

srallen:
@beckyrother screenshots. The storybook deploy is only for master at the moment.

packages/lib-classifier/src/store/WorkflowStepStore.js
srallen: Naming is hard

srallen:
There's a test failing now after rebasing this, so I'm looking into it.

srallen:
The `TextTask` model included some refactors still that will be introduced in #1212, so I've changed the the task to be a composed model like the other tasks are so that this is simply adding the text task, although, still with the plugin folder structure.

srallen:
@beckyrother here's the modifiers with left alignment:

Screen Shot 2019-10-21 at 12 02 47 PM

I noticed a couple other things in your invision design that aren't implemented. Did you want the "Text Modifiers" header above the modifiers and well as the bolding for the question text?

srallen:
This PR also doesn't implement the error styling. Is it ok if that's in a separate PR?

eatyourgreens:
How can a classification task have an error state? The text can't be wrong, and we disable the Done/Next buttons for required tasks that are incomplete.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextTask.js
eatyourgreens: Moved to #1239.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextTask.js
eatyourgreens: Moved to #1239.

eatyourgreens:
I've added the 'text modifiers' label to all the buttons.

srallen:
I'll open an issue about the error state design and we can ask about it there. It can be considered out of scope for this PR.

beckyrother:
An error state could be useful for times when only numbers or only letters should be used – dates, for example. We may not use it very often, but it's worth having just in case. It's considered poor UX to just not allow people to type in characters without telling them why; an error state will solve for that.

Ok to move to a separate PR.

eatyourgreens:
We don't support specifying a format (date, number etc.) for the text task in the project builder so that couldn't be implemented at the moment anyway.

beckyrother:
Then this can be added at a later date

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.stories.js
```diff
@@ -123,7 +122,50 @@ storiesOf('Tasks', module)
...
+      activeStepTasks: tasks,
+      isThereANextStep: () => false,
+      isThereAPreviousStep: () => false,
+      isThereTaskHelp: true
+    }
+  })
+  return (
+    <Provider classifierStore={store}>
+      <Grommet theme={Object.assign({}, zooTheme, { dark })}>
+        <Tasks
```
srallen: Non-blocking comment, but maybe we can wrap this with a `Box` that has a max-width set to the max-width allowed in the classifier design so these display closer to how they'll actually look in context.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.stories.js
```diff
@@ -123,7 +122,50 @@ storiesOf('Tasks', module)
...
+          isThereTaskHelp={true}
```
srallen: Non-blocking, maybe make this toggleable using a knob?

packages/lib-grommet-theme/src/index.js
```diff
@@ -186,7 +186,10 @@ const theme = deepFreeze({
           unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
         }
       `
-    }
+    },
+    input: {
+      weight: 'inherit'
```
srallen: Just noticed this here... Would you update its changelog and note this change? It can be listed as unpublished for now until we're read to publish another update.

srallen:
@beckyrother Here's the latest version screenshot:

Screen Shot 2019-10-23 at 11 22 36 AM

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.stories.js
eatyourgreens: Good point. The repeated JSX code in each story could be pulled out into its own helper component too.

packages/lib-grommet-theme/src/index.js
eatyourgreens: My bad, I forgot that the theme is published. Will unpublished changes be picked up when we deploy master? I can't remember if the classifier installs dependencies from npm or symlinks them from the local file system.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.stories.js
eatyourgreens: Where is the max-width for the task area in the code?

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.stories.js
eatyourgreens: I've added a bunch of improvements to the story styling, including this. Slightly outside the scope of this PR but all the Tasks stories should be looking better now.

packages/lib-grommet-theme/src/index.js
srallen: I think it's using symlinks if the version number in the package.json matches the version number in the theme's package.json, which still should be true. We can confirm on merge with staging.

srallen:
@beckyrother is this ok to merge? Thanks!

beckyrother:
@srallen Looks good, thanks!

PR #1235, opened 2019-10-16
Make project announcement dismissible

### LINKED ISSUE #1176: Make project announcement component dismissible
The project announcement component was added in #1177 as a static component. The design shows that it should be dismissible, and the original issue (#594) states it should be dismissible for the session.

What's the best way to implement this? My current thinking is that on dismissal, the UI store is updated and saves that preference to [`sessionStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage). When the client bundle loads, it checks `sessionStorage` the dismissal preference and updates the UI. This means that we can't SSR it in a dismissed state, but that doesn't bother me too much.

If there are any better implementation ideas, shout out

rogerhutchings:
Closes #1176.

## Gif

recent-classifications

## Notes

- Tests are currently failing as I couldn't get them working for storing a session cookie. Is there any obvious fix jumping out to a reviewer?
- There is a clear argument for not having this particular value in the global state, since it doesn't affect anything else. Instead, we could pass down the cookie from the `req` object to the banner container, and derive its state from that and the project announcement text. I'm fine with this, but I'd like a second opinion before I change things.
- If not, it might be worth abstracting out the cookie logic to its own store; thoughts?
- An animation for dismissal would probably look nicer, but will save it for a future PR.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/src/components/ProjectAnnouncement/ProjectAnnouncement.js
+++ b/packages/app-project/src/components/ProjectAnnouncement/ProjectAnnouncement.js
@@ -1,12 +1,20 @@
 import { Box, Paragraph } from 'grommet'
-import { Markdownz } from '@zooniverse/react-components'
-import { string } from 'prop-types'
+import { CloseButton, Markdownz } from '@zooniverse/react-components'
+import { string, func } from 'prop-types'
 import React from 'react'
 import styled from 'styled-components'
+
+import { Media } from '@shared/components/Media'
 
 const StyledParagraph = styled(Paragraph)`
   font-weight: bold;
   max-width: 60em;
+`
+
+// Hides the close button from view / the accessibility tree, so the
+// announcement text can be centred to the viewport
+const Balancer = styled(CloseButton)`
+  visibility: hidden;
 `
 
 const components = {
@@ -18,18 +26,44 @@ const components = {
 }
 
 function ProjectAnnouncement (props) {
-  const { announcement } = props
+  const { announcement, closeFn } = props
+
   return (
-    <Box align='center' background='neutral-4' fill='horizontal' pad='small'>
-      <Markdownz components={components}>
-        {announcement}
-      </Markdownz>
-    </Box>
+    <>
+      <Media at='default'>
+        <Box align='center' background='neutral-4' fill='horizontal' pad='small'>
+          <Box align='center' direction='row' gap='small' justify='between'>
+            <Markdownz components={components}>
+              {announcement}
+            </Markdownz>
+            <CloseButton closeFn={closeFn} />
+          </Box>
+        </Box>
+      </Media>
+
+      <Media greaterThan='default'>
+        <Box align='center' background='neutral-4' fill='horizontal' pad='small'>
+          <Box align='center' direction='row' gap='small' justify='between' width='xxlarge'>
+            <Balancer />
+            <Box align='center' fill='horizontal'>
+              <Markdownz components={components}>
+                {announcement}
+              </Markdownz>
+            </Box>
+            <CloseButton closeFn={closeFn} />
+          </Box>
+        </Box>
+      </Media>
+    </>
   )
 }
 
+
+
+
 ProjectAnnouncement.propTypes = {
-  announcement: string.isRequired
+  announcement: string.isRequired,
+  closeFn: func.isRequired
 }
 
 export default ProjectAnnouncement
--- a/packages/app-project/src/components/ProjectAnnouncement/ProjectAnnouncement.js
+++ b/packages/app-project/src/components/ProjectAnnouncement/ProjectAnnouncement.js
@@ -1,65 +1,23 @@
-import { Box, Paragraph } from 'grommet'
-import { CloseButton, Markdownz } from '@zooniverse/react-components'
 import { string, func } from 'prop-types'
 import React from 'react'
-import styled from 'styled-components'
 
 import { Media } from '@shared/components/Media'
-
-const StyledParagraph = styled(Paragraph)`
-  font-weight: bold;
-  max-width: 60em;
-`
-
-// Hides the close button from view / the accessibility tree, so the
-// announcement text can be centred to the viewport
-const Balancer = styled(CloseButton)`
-  visibility: hidden;
-`
-
-const components = {
-  p: (nodeProps) => (
-    <StyledParagraph color='black' margin='none'>
-      {nodeProps.children}
-    </StyledParagraph>
-  )
-}
+import NarrowProjectAnnouncement from './components/NarrowProjectAnnouncement'
+import WideProjectAnnouncement from './components/WideProjectAnnouncement'
 
 function ProjectAnnouncement (props) {
-  const { announcement, closeFn } = props
-
   return (
     <>
       <Media at='default'>
-        <Box align='center' background='neutral-4' fill='horizontal' pad='small'>
-          <Box align='center' direction='row' gap='small' justify='between'>
-            <Markdownz components={components}>
-              {announcement}
-            </Markdownz>
-            <CloseButton closeFn={closeFn} />
-          </Box>
-        </Box>
+        <NarrowProjectAnnouncement {...props} />
       </Media>
 
       <Media greaterThan='default'>
-        <Box align='center' background='neutral-4' fill='horizontal' pad='small'>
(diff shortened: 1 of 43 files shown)


beckyrother:
@rogerhutchings does the X have a hover state? Hard to tell in the gif (thank you for that btw!)

packages/app-project/stores/UI.spec.js
```diff
@@ -1,93 +1,175 @@
...
+    it('should not set the cookie on instantiation', function () {
+      expect(setAnnouncementBannerCookieSpy).to.not.have.been.called()
+    })
+
+    it('should not update the cookie on instantiation if there is already one', function () {
+      document.cookie = cookie.serialize('dismissedAnnouncementBanner', 1234567890, {
+        path: `/projects/${PROJECT.slug}`
+      })
+      store = UI.create()
+      expect(setAnnouncementBannerCookieSpy).to.not.have.been.called()
```
srallen: Looking back over this, this isn't a useful expectation. What this should be testing is not that the specific action isn't called, because it won't ever in this test setup, but rather than on patch, it doesn't update the cookie if the existing parsed one is equal to the one in the store. 

It should be noted that the comparable mode test would need to be updated to make sense as well. That can be done separately.

packages/app-project/stores/UI.spec.js
```diff
@@ -1,93 +1,175 @@
...
+
+    afterEach(function () {
+      setAnnouncementBannerCookieSpy.restore()
+    })
+
+    it('should not set the cookie on instantiation', function () {
+      expect(setAnnouncementBannerCookieSpy).to.not.have.been.called()
+    })
+
+    it('should not update the cookie on instantiation if there is already one', function () {
```
srallen: This should be `'should not update the cookie on patch if there is already one equivalent to what is stored'`

srallen:
@beckyrother it's using the `CloseButton` which is available in the storybook! https://zooniverse.github.io/front-end-monorepo/@zooniverse/react-components/index.html?path=/story/closebutton--light-theme-default

packages/app-project/stores/UI.js
```diff
@@ -1,38 +1,96 @@
...
+    return {
+      dismissedAnnouncementBanner,
+      mode
+    }
+  })
+
   .actions(self => ({
     afterCreate () {
       self.createModeObserver()
+      self.createDismissedAnnouncementBannerObserver()
```
srallen: You might want to consider adding this in `afterAttach` instead of `afterCreate` because this patch depends on the project store existing, but this will make it a bit more trickier to test, so this is non-blocking.

packages/app-project/stores/UI.js
```diff
@@ -1,38 +1,96 @@
...
+    },
+
+    createDismissedAnnouncementBannerObserver() {
+      const modeDisposer = autorun(() => {
+        onPatch(self, (patch) => {
+          const { path } = patch
+          if (path === '/dismissedAnnouncementBanner') {
+            self.setAnnouncementBannerCookie()
+          }
         })
```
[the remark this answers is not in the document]
rogerhutchings: Not seeing that in the docs

packages/app-project/stores/UI.js
```diff
@@ -1,38 +1,96 @@
...
+      const announcementHash = stringHash(announcement)
+      self.dismissedAnnouncementBanner = announcementHash
+    },
+
+    setAnnouncementBannerCookie() {
       // process.browser doesn't exist in the jsdom test environment
       if (process.browser || process.env.BABEL_ENV === 'test') {
         const parsedCookie = cookie.parse(document.cookie) || {}
+        if (self.dismissedAnnouncementBanner !== parsedCookie.dismissedAnnouncementBanner) {
+          const isProduction = process.env.NODE_ENV === 'production'
```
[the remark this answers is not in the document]
rogerhutchings: It's used to correctly set the cookie domain to `zooniverse.org`, so `NODE_ENV`?

rogerhutchings:
I worked out what was happening - `jsdom` has an internal URL, and cookies were being scoped to a different path. I fixed this by making the jsdom object a global, so it can be reconfigured for this set of tests, and restoring it when finished.

srallen:
Tests are passing locally. I'll try rerunning Travis.

packages/app-project/stores/UI.js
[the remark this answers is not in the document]
srallen: https://mobx.js.org/refguide/autorun.html#options

packages/app-project/stores/UI.js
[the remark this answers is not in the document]
srallen: I get confused about when we use one versus the other.

packages/app-project/src/components/ProjectAnnouncement/ProjectAnnouncement.js
```diff
@@ -18,18 +26,44 @@ const components = {
...
+            <Box align='center' fill='horizontal'>
+              <Markdownz components={components}>
+                {announcement}
+              </Markdownz>
+            </Box>
+            <CloseButton closeFn={closeFn} />
+          </Box>
+        </Box>
+      </Media>
+    </>
```
srallen: This seems a bit complicated. It seems like this could be simplified using flex box, but I'd have to play around with it to figure out what to suggest.

rogerhutchings: I know. The aim was to get the text centred to the viewport, with an off-centre close button, that doesn't use absolute positioning (no weird edge cases where they collide), and that keeps its proportions when text is resized. One for another day, probably

rogerhutchings: I know. The aim was to get the text centred to the viewport, with an off-centre close button, that doesn't use absolute positioning (no weird edge cases where they collide), and that keeps its proportions when text is resized. One for another day, probably

srallen: Centering stuff is harder than it should be 👍

packages/app-project/stores/UI.js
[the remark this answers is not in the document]
rogerhutchings: https://github.com/zooniverse/front-end-monorepo/pull/1235/commits/3e787a9e48fb4aabad385bebaa1c36bb54819dfa#diff-9c3e748a74d94414d552499dd2decd23R67

packages/app-project/stores/UI.js
rogerhutchings: I moved it to `afterAttach` and everything was fine... ¯\_(ツ)_/¯

packages/app-project/src/components/ProjectAnnouncement/components/NarrowProjectAnnouncement/NarrowProjectAnnouncement.spec.js
```diff
@@ -0,0 +1,28 @@
...
+  it('should render without crashing', function () {
+    expect(wrapper).to.be.ok()
+  })
+
+  it('should render a paragraph with the announcement text', function () {
+    const paragraphWrapper = wrapper.find('p')
+    expect(paragraphWrapper.text()).to.contain(ANNOUNCEMENT)
+  })
+
+  xit('should be the full width of its container', function () {
```
srallen: This is still set to be pending.

packages/app-project/src/components/ProjectAnnouncement/components/WideProjectAnnouncement/WideProjectAnnouncement.spec.js
```diff
@@ -0,0 +1,28 @@
...
+  it('should render without crashing', function () {
+    expect(wrapper).to.be.ok()
+  })
+
+  it('should render a paragraph with the announcement text', function () {
+    const paragraphWrapper = wrapper.find('p')
+    expect(paragraphWrapper.text()).to.contain(ANNOUNCEMENT)
+  })
+
+  xit('should be the full width of its container', function () {
```
srallen: Still set to be pending.

rogerhutchings:
Tests have been restored and pass - thanks @srallen

Issue #1174, opened 2019-09-26
Add Point model


mcbouslog:
Point store - consumes coordinate stream, stores coordinates for Point component, finishes mark when appropriate based on coordinate stream

PR #1249, opened 2019-10-25
Refactor drawing store, add point and line stores

### LINKED ISSUE #1175: Add Line model
Line store - consumes coordinate stream, stores coordinates for Line component, finishes mark when appropriate based on coordinate stream

mcbouslog:
Package: lib-classifier

staging workflows:
- both point & line: http://localhost:8080/?project=1876&workflow=3344
- point only: http://localhost:8080/?project=1876&workflow=3363
- line only: http://localhost:8080/?project=1876&workflow=3362

Describe your changes:
- adds `coordinateStream` to DrawingStore, which streams events on `pointerdown`, with coordinates converted for svg
- adds PointStore and LineStore for creating related marking with coordinates for related component, including when to stop drawing

Closes #1175 and #1174 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/store/DrawingStore.js
```diff
@@ -1,56 +1,175 @@
...
+  return markStores[toolType] || null
+}
 
 const DrawingStore = types
   .model('DrawingStore', {
-    activeDrawingTool: types.optional(types.number, 0)
+    activeDrawingToolIndex: types.optional(types.number, 0),
+    activeMark: types.safeReference(types.union({
+      dispatcher: (snapshot) => {
+        const snapshotType = getType(snapshot)
```
eatyourgreens: Something I learnt while working on #1212: `getType` can only be called on a tree node, not on a snapshot. Here it will return `LineStore` for a line or `PointStore` for a point so the rest of the function becomes `return snapshotType` which can then be shortened further to `dispatcher: getType`. Again, that only works if you're passing around valid tree nodes, not JSON snapshots of models.

packages/lib-classifier/src/store/DrawingStore.js
```diff
@@ -1,56 +1,175 @@
...
+        if (snapshotType.name === 'LineStore') return LineStore
+        if (snapshotType.name === 'PointStore') return PointStore
+        return undefined
+      }
+    }, LineStore, PointStore)),
+    marks: types.map(types.union({
+      dispatcher: (snapshot) => {
+        const snapshotType = getType(snapshot)
```
eatyourgreens: Same comments here about the dispatcher function.

packages/lib-classifier/src/store/markings/LineStore.js
```diff
@@ -0,0 +1,97 @@
+import { addDisposer, getRoot, isValidReference, types } from 'mobx-state-tree'
+import { autorun } from 'mobx'
+import { filter, skip } from 'rxjs/operators'
+
+import Mark from './Mark'
+
+const Line = types
+  .model('LineStore', {
```
eatyourgreens: The naming of these models confuses me. Does this model represent an individual line, or a store to hold lines?

packages/lib-classifier/src/store/markings/PointStore.js
```diff
@@ -0,0 +1,84 @@
+import { addDisposer, getRoot, isValidReference, types } from 'mobx-state-tree'
+import { autorun } from 'mobx'
+import { filter } from 'rxjs/operators'
+
+import Mark from './Mark'
+
+const Point = types
+  .model('PointStore', {
```
eatyourgreens: Same comment about naming here. I'm not sure if this model represents a point or a store to hold points.

packages/lib-classifier/src/store/markings/LineStore.js
eatyourgreens: In that case, I'd suggest following the other model names (eg. `Classification` vs. `ClassificationStore`) and name this `Line` or `LineMark`.

packages/lib-classifier/src/store/DrawingStore.js
```diff
@@ -1,56 +1,175 @@
-import { addDisposer, getRoot, onAction, types } from 'mobx-state-tree'
+import cuid from 'cuid'
+import { addDisposer, getRoot, getType, isValidReference, onAction, types } from 'mobx-state-tree'
 import { autorun } from 'mobx'
 import { Subject } from 'rxjs'
+import { filter, map, skipUntil } from 'rxjs/operators'
+
+import { LineStore, PointStore } from './markings'
```
eatyourgreens: This can probably use
```js
import { * as Marks } from './markings'
```
then enumerate the values of `Marks` when you need them. That way, you won't need to update this line with new marking models.

packages/lib-classifier/src/store/DrawingStore.js
```diff
@@ -1,56 +1,175 @@
...
+        toolIndex: self.activeDrawingToolIndex
+      })
+
+      self.marks.put(newMark)
+      self.activeMark = tempId
     }
 
-    function setActiveDrawingTool (toolIndex) {
-      self.activeDrawingTool = toolIndex
+    function setEventStream (stream) {
```
srallen: Does anything call this?

packages/lib-classifier/src/store/markings/LineStore.js
```diff
@@ -0,0 +1,97 @@
...
+    coordinatesArray: types.maybeNull(
+      types.array(types.model({
```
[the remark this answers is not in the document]
mcbouslog: Sounds good regarding `maybeNull`, I'm updating accordingly.

The thought process behind a `coordinatesArray` and `coordinates` view was so that all markings would store coordinates similarly (`coordinatesArray`), but have a unique view. I also found it kept the actions more straightforward; as opposed to manipulating an object (like `{ x1: number, y1: number, x2: number, y2: number }` which would be different for each marking) in the action that sets the most recent event coordinates, this keeps the set coordinates action to just pushing the latest coordinates. But I'm not confident this is the optimal approach, just initial attempt.

I'm not sure how editing a marking will fit into this structure.

packages/lib-classifier/src/store/markings/LineStore.js
```diff
@@ -0,0 +1,97 @@
...
+          self.setCoordinatesSubscription()
+          self.setStopSubscription()
+        }
+      }, { name: 'LineStore DrawingTask Observer autorun' })
+      addDisposer(self, drawingTaskDisposer)
+    }
+
+    function setCoordinates (event) {
+      if (!self.coordinatesArray) {
+        self.coordinatesArray = [event, event]
```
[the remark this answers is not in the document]
mcbouslog: It should not include the entire event object, just coordinates, I'll update accordingly.

packages/lib-classifier/src/store/DrawingStore.js
mcbouslog: I use it for testing purposes, but there's prob a better approach.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/DrawingContainer.js
```diff
@@ -1,143 +1,42 @@
...
 
-          const MarkingComponent = getDrawingTool(marking.tool.type)
           return (
             <MarkingComponent
               key={id}
-              coordinates={marking.coordinates}
-              tool={marking.tool}
+              active={mark === activeMark}
+              coordinates={mark.coordinates}
+              tool={tool}
```
eatyourgreens: You can probably store the tool on the Mark model, then reference `mark.tool` here.

packages/lib-classifier/src/store/DrawingStore.js
```diff
@@ -1,56 +1,175 @@
...
-      })
+      }, { name: 'DrawingStore WorkflowStep Observer autorun' })
 
       addDisposer(self, workflowStepsDisposer)
     }
 
-    function reset () {
-      self.activeDrawingTool = 0
+    function createMark () {
+      const MarkStore = getMarkStore(self.activeDrawingTask.tools[self.activeDrawingToolIndex].type)
```
eatyourgreens: Maybe the drawing task should have an active tool view, so that you can reference `drawingTask.activeTool` here? It feels a bit weird that tools are stored on the drawing task, but the active tool index is stored in this store.

packages/lib-classifier/src/store/DrawingStore.js
```diff
@@ -1,56 +1,175 @@
...
+      // TODO add error/null handling ^
+
+      const tempId = cuid()
+
+      const newMark = MarkStore.create({
```
eatyourgreens: Maybe `MarkModel` rather than `MarkStore` here, since you're creating a new mark, not a store that holds marks.

eatyourgreens:
Re. the merge conflicts, #1212 moved the task views and models to `@plugins/tasks`. The drawing task code itself wasn't changed, but module imports were changed.

There are also now some aliases to avoid long relative paths when importing modules. `@plugins` for `src/plugins` and `@store` for `src/store`.

packages/lib-classifier/src/store/markings/LineStore.js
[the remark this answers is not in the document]
mcbouslog: Attempted to refactor to set coordinates directly on model with `applyPatch` in `setCoordinates` action, console logged and confirmed patches were firing and everything looked to be functioning as expected, but in DOM the point/line component wasn't visually updating on pointer moves. It'd show point/line on initial pointer down, then if dragged mouse around nothing, but on pointer up the point or line would show as expected. Didn't dig too much into in light of upcoming drawing reassessment.

packages/lib-classifier/src/store/DrawingStore.js
```diff
@@ -1,56 +1,156 @@
...
+  line: Line,
+  point: Point
+}
+
+function getMarkModel (toolType) {
+  return markModels[toolType] || null
+}
+
 const DrawingStore = types
   .model('DrawingStore', {
```
eatyourgreens: I'm not 100% following why there's a `DrawingStore` model and a `DrawingTask` model. Does this model the drawing itself ie. the marks created by all drawing tasks from the current workflow?

packages/lib-classifier/src/store/DrawingStore.js
```diff
@@ -1,56 +1,156 @@
...
-    activeDrawingTool: types.optional(types.number, 0)
+    activeDrawingToolIndex: types.optional(types.number, 0),
```
eatyourgreens: The active tool index seems like it should more naturally be the property of a drawing task.

packages/lib-classifier/src/store/DrawingStore.js
```diff
@@ -1,56 +1,156 @@
...
+    activeMark: types.safeReference(types.union({
+      dispatcher: getType
+    }, Line, Point)),
+    marks: types.map(types.union({
```
eatyourgreens: This is a repeat of the types declaration from the previous line. They could be rolled up into a single piece of code.

packages/lib-classifier/src/store/DrawingStore.js
```diff
@@ -1,56 +1,156 @@
...
+        onAction(getRoot(self), (call) => {
           if (call.name === 'selectStep') self.reset()
         })
-      })
+      }, { name: 'DrawingStore WorkflowStep Observer autorun' })
 
       addDisposer(self, workflowStepsDisposer)
     }
 
+    function createMark () {
```
eatyourgreens: I think creating a new mark should be the responsibility of the active tool eg. `activeTool.createMark()`. Otherwise, I think the code gets into mess trying to figure out what form a new Mark should take. Each tool 'knows' what form its particular marks should take.

eatyourgreens:
Playing around with [the test workflow](http://localhost:8080/?project=1876&workflow=3344) there's a problem with the line tool: drawing doesn't finish when I release the pointer. I have to click a second time in order to finish drawing a line.

packages/lib-classifier/src/store/markings/Line.js
```diff
@@ -0,0 +1,100 @@
+import { addDisposer, getRoot, isValidReference, types } from 'mobx-state-tree'
+import { autorun } from 'mobx'
+import { filter, skip, skipUntil } from 'rxjs/operators'
+
+import Mark from './Mark'
+
+const LineModel = types
+  .model('LineModel', {
```
eatyourgreens: These models should reflect the drawing tools currently in use, so for the line tool, a mark snapshot should look something like:
```js
{
  "tool": 3,
  "frame": 0,
  "x1": 309.7053527832031,
  "y1": 326.549072265625,
  "x2": 764.7003173828125,
  "y2": 236.323974609375
}
```

packages/lib-classifier/src/store/markings/Point.js
```diff
@@ -0,0 +1,87 @@
+import { addDisposer, getRoot, isValidReference, types } from 'mobx-state-tree'
+import { autorun } from 'mobx'
+import { filter, skipUntil } from 'rxjs/operators'
+
+import Mark from './Mark'
+
+const PointModel = types
+  .model('PointModel', {
```
eatyourgreens: The Point model should look like this:
```js
{
  "tool": 0,
  "frame": 0,
  "x": 498.4888610839844,
  "y": 374.8988342285156
}
```

packages/lib-classifier/src/store/markings/Mark.js
```diff
@@ -0,0 +1,8 @@
+import { types } from 'mobx-state-tree'
+
+const Mark = types.model('Mark', {
+  id: types.identifier,
+  toolIndex: types.optional(types.number, 0)
```
eatyourgreens: `toolIndex` should be `tool` and `frame` needs to be added.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/tools/Point.js
```diff
@@ -15,7 +15,7 @@ const SELECTED_RADIUS = {
 const CROSSHAIR_SPACE = 0.2
 const CROSSHAIR_WIDTH = 1
 
-const Point = ({ active, coordinates, finishDrawing, scale, tool }) => {
+const Point = ({ active, coordinates, scale, tool }) => {
```
eatyourgreens:
```suggestion
function Point ({ active, coordinates, scale, tool }) {
```

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/tools/Line.js
```diff
@@ -2,11 +2,8 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import DrawingToolRoot from './DrawingToolRoot'
 
-const Line = ({ active, coordinates, finishDrawing, tool }) => {
+const Line = ({ active, coordinates, tool }) => {
```
eatyourgreens:
```suggestion
function Line ({ active, coordinates, tool }) {
```

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/DrawingContainer.js
```diff
@@ -1,143 +1,42 @@
...
-          key={activeMark.id}
-          active
-          finishDrawing={this.finishMark}
-          tool={activeMark.tool}
-        />
-        {marks.size > 0 && Array.from(marks, ([id, marking]) => {
-          if (!marking) {
-            return null
-          }
+        {Array.from(marks, ([id, mark]) => {
```
eatyourgreens: I could be wrong, but shouldn't the `DrawingContainer` component only be rendering marks from previous tasks ie. marks which can't be modified or deleted?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -37,29 +61,27 @@ class InteractionLayerContainer extends Component {
...
-
     return (
       <>
         <InteractionLayer
           onPointerDown={this.onPointerDown}
           onPointerMove={this.onPointerMove}
           onPointerUp={this.onPointerUp}
         />
-        {isDrawingInActiveSteps && <DrawingContainer />}
+        {this.props.isDrawingInActiveWorkflowStep && <DrawingContainer />}
```
eatyourgreens: If the next step in the workflow isn't a drawing task, all my marks from this step will vanish when I press Next.

eatyourgreens: I think the conditional here should be on the drawing canvas AKA interaction layer. Any marks that have been drawn should always be rendered, but we don't always want to enable drawing.

`InteractionLayer` and `DrawingContainer` should be renamed to describe what they actually do.

packages/lib-classifier/src/store/markings/LineStore.js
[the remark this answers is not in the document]
eatyourgreens: I don't think you need to be updating the store until drawing is complete. At least, I don't think we're interested in observing mark changes while you're still drawing. Otherwise, I think that just leads us back into the performance issues that PFE has when the annotation is updated on every change to a mark.

I'd suggest rendering the SVG shapes from components and paths stored in component state, then updating the store with the final values when the pointer lifts.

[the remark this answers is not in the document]
srallen: Storing in local component state definitely was the original plan, but it shifted to the store due to the attempt to integrate with the rx.js stream and the stream was initialized in the store... I'm not sure if there was any other complications that led to us shifting to using the MST store.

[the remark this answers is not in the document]
eatyourgreens: If I were to add RxJS back in, I think this time I'd start by subscribing to the stream in the event handlers in the drawing component, then rendering directly from the streamed coordinates without involving the store.

packages/lib-classifier/src/store/markings/LineStore.js
```diff
@@ -0,0 +1,97 @@
...
+  .model('LineStore', {
+    coordinatesArray: types.maybeNull(
+      types.array(types.model({
+        x: types.number,
+        y: types.number
+      }))
+    )
+  })
+  .views(self => ({
+    get coordinates () {
```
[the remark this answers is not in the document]
eatyourgreens: I'm going to change these models in #1266.

packages/lib-classifier/src/store/markings/PointStore.js
```diff
@@ -0,0 +1,84 @@
...
+    coordinatesArray: types.maybeNull(
```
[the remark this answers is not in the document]
eatyourgreens: #1266.

packages/lib-classifier/src/store/markings/PointStore.js
```diff
@@ -0,0 +1,84 @@
...
+      }
+    }
+
+    function stop () {
+      self.coordinatesSubscription.unsubscribe()
+      self.stopSubscription.unsubscribe()
+      getRoot(self).drawing.createMark()
+    }
+
+    function setStopSubscription () {
```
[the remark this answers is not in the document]
eatyourgreens: Maybe `mark.onComplete()` or `mark.finish()`?

packages/lib-classifier/src/store/DrawingStore.js
eatyourgreens: Resolved in #1266.

packages/lib-classifier/src/store/DrawingStore.js
eatyourgreens: Resolved in #1266.

packages/lib-classifier/src/store/DrawingStore.js
eatyourgreens: Resolved in #1266.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/tools/Point.js
eatyourgreens: Resolved in #1266.

packages/lib-classifier/src/store/markings/LineStore.js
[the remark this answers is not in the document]
eatyourgreens: I've now run into the same problem, on #1266, where updating a mark does not trigger a new render of the `DrawingContainer` component.

eatyourgreens:
I've squashed this into #1266, so I'll close it.

PR #1251, opened 2019-10-28
Actually call to create a default annotation on next or on done

srallen:

Package: lib-classifier

Describe your changes:
Right now, staging 400 errors because the default annotation isn't actually being created when you click next or done. This was missed by me in review of #1228. I've fixed this by removing the now cruft `createDefaultAnnotation` and calling `addAnnotation` directly. I've added some tests to check that default annotations are actually created for the TaskNav container. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.spec.js
@@ -159,7 +159,7 @@ describe('TaskNavButtonsContainer', function () {
     })
   })
 
-  describe.only('#onSubmit', function () {
+  describe('#onSubmit', function () {
     let wrapper
     let completeClassificationSpy
     let addAnnotationSpy
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.spec.js
@@ -7,27 +7,31 @@ import TaskNavButtonsContainer from './TaskNavButtonsContainer'
 import ClassificationStore from '../../../../../../../../store/ClassificationStore'
 import SingleChoiceTask from '../../../../../../../../store/tasks/SingleChoiceTask'
 import MultipleChoiceTask from '../../../../../../../../store/tasks/MultipleChoiceTask'
+import Step from '../../../../../../../../store/Step'
 import { SubjectFactory, WorkflowFactory, ProjectFactory } from '../../../../../../../../../test/factories'
 
 const steps = observable.map([
-  ['S0', { taskKeys: ['T0'] }],
-  ['S1', { taskKeys: ['T1'] }]
+  Step.create({ stepKey: 'S0', taskKeys: ['T0'] }),
+  Step.create({ stepKey: 'S1', taskKeys: ['T1'] })
 ])
 
-const tasks = [
-  SingleChoiceTask.create({
-    answers: [{ label: 'yes' }, { label: 'no' }],
-    question: 'Is there a cat?',
-    required: false,
-    taskKey: 'init',
-    type: 'single'
-  }), MultipleChoiceTask.create({
-    answers: [{ label: 'napping' }, { label: 'standing' }, { label: 'playing' }],
-    question: 'What is/are the cat(s) doing?',
-    required: false,
-    taskKey: 'T1',
-    type: 'multiple'
-  })
+const singleChoiceTask = SingleChoiceTask.create({
+  answers: [{ label: 'yes' }, { label: 'no' }],
+  question: 'Is there a cat?',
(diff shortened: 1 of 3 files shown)


eatyourgreens:
I think you can call `task.createAnnotation()` to create the missing annotation.

srallen:
> I think you can call task.createAnnotation() to create the missing annotation.

My tests start failing because they do not get added to the classification annotation if I call `task.createAnnotation()` directly in the TaskNavContainer. Calling `addAnnotation` ensures it's being created and added to the annotations. I'm not sure if I'm doing something wrong with this suggestion but the current solution is working.

eatyourgreens:
You’re right. Tasks create annotations but aren’t responsible for storing them, because they might be creating them for drawing subtasks.

If the annotation value is optional in `addAnnotation`, it might be worth changing the signature of that action to `addAnnotation(task, value)` so that it’s easier to call with just the task.

eatyourgreens:
`task.updateAnnotation(value)` is what I was trying to think of. It will create an annotation on the classification if it doesn't exist. https://github.com/zooniverse/front-end-monorepo/blob/733d7155a8764b530c962eb7e2717147ff830517/packages/lib-classifier/src/store/tasks/Task.js#L22-L25

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.spec.js
```diff
@@ -4,28 +4,36 @@ import { observable } from 'mobx'
 import { expect } from 'chai'
 import sinon from 'sinon'
 import TaskNavButtonsContainer from './TaskNavButtonsContainer'
+import ClassificationStore from '../../../../../../../../store/ClassificationStore'
```
eatyourgreens: Are we still waiting to merge the PR that aliases these to `@store/ClassificationStore` etc?

packages/lib-classifier/src/store/ClassificationStore.js
```diff
@@ -116,7 +116,9 @@ const ClassificationStore = types
           const annotation = classification.annotations.get(task.taskKey) || task.createAnnotation()
           // new annotations must be added to this store before we can modify them
           classification.annotations.put(annotation)
-          annotation.value = annotationValue
+          if (annotationValue) {
```
eatyourgreens: Outside the scope of this PR, but maybe `annotation.update(annotationValue)` and then each annotation is responsible for validating the type and value of `annotationValue`.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.spec.js
```diff
@@ -214,24 +188,29 @@ describe('TaskNavButtonsContainer', function () {
...
     })
 
-    it('should prevent the event default', function () {
+    it('should create a default annotation for each task if there is not an annotation for that task', function () {
       wrapper.instance().onSubmit({ preventDefault: preventDefaultSpy })
-      expect(preventDefaultSpy).to.have.been.calledOnce()
+      const classification = classificationStore.active.toJSON()
+
+      tasks.forEach((task) => {
+        expect(addAnnotationSpy.withArgs(null, task)).to.have.been.calledOnce()
```
eatyourgreens: Outside the scope of this PR, but a better test here would be to expect that the annotation for this task, stored on the classification, is the default annotation.

eatyourgreens: Nevermind, that's the next check. This expectation can be deleted. We don't care how the default annotation is implemented, only that it's created. We want to test behaviour, not implementation.

eatyourgreens:
Testing this, I've discovered that tests all pass even if  the `steps` prop is an empty array. I'm not sure what's going on there, but `should call props.selectStep when there is a previous step` passes even without any steps.

eatyourgreens:
@srallen the bugfix looks good to me. I've updated it to use `task.updateAnnotation()` and added a workflow step store to the tests.

eatyourgreens:
PFE creates the default annotation when a volunteer begins a task. https://github.com/zooniverse/Panoptes-Front-End/blob/6d0656ddfda20bd83d2e7d8bfd63294b5bb8b1a5/app/classifier/task-nav.jsx#L23-L40

Maybe the classifier should do that too, rather than creating the annotation when the task is complete?

eatyourgreens:
The fix seems to work, but the tests aren't generating the default annotation, yet still pass.

srallen:
> Maybe the classifier should do that too, rather than creating the annotation when the task is complete?

There was a reason why I didn't do this, but it's not immediately coming to mind.

eatyourgreens:
I started down the route of mocking a `WorkflowStepStore` for this, since the task nav buttons depend heavily on it, but gave up. It seemed like it would be a good idea to mock the initial step store state in order to generate initial props for some of the tests, rather than mock props which might not reflect a real store.

srallen:
I think refactoring the step portion of these tests is out of scope.

Issue #1143, opened 2019-09-19
[VSLCV] Scatter Plot Performance


srallen:
**Package** lib-classifier

**Feature or Issue Description**
Issue tracking the build out of the scatter plot portion of the variable star light curve viewer. 

https://projects.invisionapp.com/d/main?origin=v7?origin=v7#/projects/prototypes/17294981

**To Dos**

- [x] axes
  - [x] 2D
  - [x] customizable labels
  - [x] inner facing
  - [x] outer facing
- [x] zoom
  - [x] both directions
  - [x] x only
  - [x] y only
  - [x] constrained by max and min scale
  - [x] constrained by data boundaries
- [x] pan
  - [x] both directions
  - [x] x only
  - [x] y only
  - [x] constrained by max and min scale
  - [x] constrained by data boundaries
- [x] responsive
- [x] support for multiple light curves
  - [x] point color set by spectra
- [ ] ~functional in light and dark theme~ moved into own issue
- [ ] ~keyboard controls~ moved into own issue

srallen:
Mouse wheel zoom performance test results:

Now here's a weird one; I was going to disagree with you on this, until I found out what triggers the poor performance.

- when using a MacBook Pro's trackpad, on macOS + Chrome78, if the dev console is open, then zooming via mouse scroll is extremely jaggy, jumping inconsistently between multiple zoom levels per scroll.
- when using a MacBook Pro's trackpad, on macOS + Firefox 70, zooming via mouse scroll isn't very smooth, but better than Chrome with the dev console open.
- on macOS 10 + Safari, zooming via mouse scroll is pretty smooth, which is an affront to everything I understand about browsers.
- when using an external USB mouse, zooming via mouse scroll is mostly smooth on all tested conditions and browsers.
  - (Chrome with the dev console open was slightly slower, but not as janked out as with touchpad + open dev console)

Overall thoughts summarised as an emoji: 😕

srallen:
From #1256:

> UI Issue, Minor: when click-dragging to pan, it's possible to select the text of the x-axis and y-axis. (And also any text that's outside of the ScatterPlotViewer itself.) It seems that the pan action doesn't prevent the default "click-drag to select text" browser action.

srallen:
Gonna close this and will reopen if we get reports of issues from the beta.

PR #1256, opened 2019-11-01
[VSLCV] Zooming and Panning ScatterPlotViewer

srallen:

Package: lib-classifier

Describe your changes:
This adds pan/zoom functionality to the ScatterPlotViewer. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
If it’s useful, there’s a `withKeyZoom` helper in the classifier that adds the keyboard controls too.

srallen:
There's a handful of panning related tests for the `ZoomingScatterPlot` still failing. I'll have to fix those later. 

The basic idea is that we have a `VXZoom` component that implements vx's `Zoom`. The `VXZoom` component defines handlers for all of the expected DOM event handlers, but uses the default vx zoom constrains. The default constrain doesn't work for the scatter plot, so we have a `ZoomingScatterPlot` that sets its own set of zooming constraint functions, passes it as a prop to `VXZoom` as well as the `ScatterPlot` as the child component.

srallen:
I'd like to implement the keyboard controls separately. There's a lot going on here already.

srallen:
@shaunanoordin do you have time to review this since you'll be somewhat familiar with the general function here. I borrowed a bit from the original LCV.

This can be manually tested in the storybook with the pan/zoom story. There's also a bunch of automated tests for the pan/zoom function both in general with `VXZoom` and specific to the scatter plot since there are unique constrains it should use.

As an aside, I ended up going with a standard clip-path. The lighthouse testing didn't show much difference between just using that and also styles to not display the circles outside of the clip-path.

shaunanoordin:
## PR Review (Quick ver)

This PR adds Pan and Zoom functionality to the Scatter Plot-type Subject Viewer component.

- package: `lib-classifier`
- This component can be tested in Storybook (`yarn storybook`) under [Subject Viewers > ScatterPlotViewer > pan and zoom](http://localhost:6006/?path=/story/subject-viewers-scatterplotviewer--pan-and-zoom)
- Primary functionality:
  - with pan enabled, click-dragging on the subject viewer should move the view on the subject in the x- and y-directions.
    - (To test later: limiting the drag to the x- or y-axis only)
  - with zoom enabled, scrolling on the subject viewer should zoom in/out 
  - the zoom in, zoom out, and reset (zoom + pan) buttons on the Subject Viewer should work as advertised.
  - panning should not allow the view to go past the bounds of the scatterplot data. (e.g. if scatterplot goes from x=[0, 30], do not allow the pan to go so far left that it reaches x=-12550821 where the Far Lands lie.)
  - zoom levels should respect zoom limits specified on the component.
- If I understand correctly, keyboard functionality is not added as part of this PR. (I'll double check my assumption in the code later.)

Tested on macOS 10 + Chrome 78 👌 

### Status

Review ongoing.

Primary functionality on Storybook looks mostly good, though I'm noting there's _possibly_ one issue:
- scrolling to zoom (on a MacBook touchpad) also causes the page (or rather, the component viewport) to scroll up/down.

I'll check more on that later, since I'm not sure if the scrolling issue is specific to Storybook's component-only view, or also appears on a full project view. For now, I'm reading through the code; brb.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+
+  onMouseLeave() {
+    if (!this.zoom.isDragging && !this.props.panning) return
+    this.zoom.dragEnd()
+  }
+
+  onWheel(event) {
+    // performance of this is pretty bad
```
shaunanoordin: > performance of this is pretty bad

Now here's a weird one; I was going to disagree with you on this, until I found out what triggers the poor performance.
- when using _a MacBook Pro's trackpad,_ on macOS + Chrome78, if the _dev console is open,_ then zooming via mouse scroll is extremely jaggy, jumping inconsistently between multiple zoom levels per scroll.
- when using _a MacBook Pro's trackpad,_ on macOS + Firefox 70, zooming via mouse scroll isn't very smooth, but better than Chrome with the dev console open.
- on macOS 10 + Safari, zooming via mouse scroll is pretty smooth, which is an affront to everything I understand about browsers.
- when using _an external USB mouse,_ zooming via mouse scroll is mostly smooth on all tested conditions and browsers.
  - (Chrome with the dev console open was slightly slower, but not as janked out as with touchpad + open dev console)

Overall thoughts summarised as an emoji: 😕

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
```diff
@@ -0,0 +1,160 @@
...
+    if (this.props.zooming) {
+      const zoomDirection = (-event.deltaY > 0) ? 'in' : 'out'
+      this.zoomToPoint(event, zoomDirection)
+    } else {
+      event.preventDefault()
```
shaunanoordin: Query: I'm not sure what's the purpose of `event.preventDefault()` here? But it may not matter because...

Quirk: ...it seems that it's never triggered? If `!props.zooming` (i.e. by disabling the 'zooming' property in the Storybook knob for ScatterPlotViewer), it seems that `onWheel()` is never called.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
```diff
@@ -0,0 +1,160 @@
...
+  onMouseLeave() {
+    if (!this.zoom.isDragging && !this.props.panning) return
+    this.zoom.dragEnd()
+  }
+
+  onWheel(event) {
+    // performance of this is pretty bad
+    if (this.props.zooming) {
+      const zoomDirection = (-event.deltaY > 0) ? 'in' : 'out'
+      this.zoomToPoint(event, zoomDirection)
```
shaunanoordin: On the subject of _"zooming via mouse scroll also causes the page to scroll",_ I was going to suggest that adding an `event.preventDefault()` at the end of this `if` block might solve the problem.

However, this is only true for Firefox and Safari - for _Chrome,_ `event.preventDefault()` doesn't prevent the page from scrolling. I'll give a check to see if `event.nativeEvent.doVoodoo(); return false` might be a better bet.

shaunanoordin: OH! So it's a known problem in Chrome. Attempting to create...

```
if (this.props.zooming) {
  const zoomDirection = (-event.deltaY > 0) ? 'in' : 'out'
  this.zoomToPoint(event, zoomDirection)
  event.preventDefault()
} else { ...
```

...in Chrome causes the following error:

```
[Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/6662647093133312
```

Goddanggit, sorry, my solution won't work then.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
```diff
@@ -0,0 +1,160 @@
...
+        scaleXMin={zoomConfiguration.minZoom}
+        scaleXMax={zoomConfiguration.maxZoom}
+        scaleYMin={zoomConfiguration.minZoom}
+        scaleYMax={zoomConfiguration.maxZoom}
+        passive
+        width={parentWidth}
+      >
+        {zoom => {
+          this.zoom = zoom
+          {/* console.log('zoom', zoom.transformMatrix) */}
```
shaunanoordin: Vestigial console.log here.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
shaunanoordin: OK, I see where the quirk comes in: in `ScatterPlotViewer.js`, if that component's `zooming` property is set to false, it renders `<ScatterPlot>` instead of `<ZoomingScatterPlot>`, meaning the `<VXZoom>` rect isn't rendered at all.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer/components/ZoomingScatterPlot/ZoomingScatterPlot.js
```diff
@@ -0,0 +1,245 @@
...
+        translateY: prevTransformMatrix.translateY
+      })
+    }
+
+    if (isXScaleMin || isYScaleMin) {
+      return this.zoomReset()
+    }
+
+    if (isXAxisOutOfBounds || isYAxisOutOfBounds) {
+      return prevTransformMatrix
```
shaunanoordin: This implementation of the _"prevent user from panning beyond the bounds of the Subject's data"_ logic is legit, and is in fact too legit, too legit to quit.

Also, bonus 👍 for accounting for the wacky logic for x-axis-only constraints, y-axis-only constraints, and both-axis constraints.

One quirk to note is that using "revert to a previous transform" logic - as opposed to the more complex "calculate a new transform that is as far as the user wants to go but still within the bounds of the data" - will cause huge pan values to cancel out, leaving the user in the same place.

e.g. the valid data range is x=[0, 10], and currently x=2. If the user attempts to pan left by -5 the view will remain at x=2 instead of going "as far left as possible" to x=0.

That said, at the moment, this is only _a theoretical problem,_ as the only way I could "cancel my pan action", so to speak, was to purposely attempt to make a huge pan jump by quickly click-dragging a large distance - _and nobody sensible will do that._

This _may_ only matter if we introduce 'fixed' jumps later though, e.g keyboard left button jumps left by a fixed -5, but the view is at x=2, so pressing left key seems to "stop panning" well before the user sees the left-most edge of the Subject data.

(Clarification: this is not a blocker, but just a flag for possible future gotchas.)

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
eatyourgreens: Should you be canceling bubbling with `stopPropagation`? I don't know if wheel events bubble. `preventDefault` will prevent the default behaviour for the current DOM element eg. an `<a href>` changes the window location in response to click events. I _think_ you want to prevent the default behaviour on the window itself. I'm not sure.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
srallen: I've copied your comment into https://github.com/zooniverse/front-end-monorepo/issues/1143#issuecomment-558698182

I think we'll have to make a final decision about supporting mouse wheel at some point before finishing this. I'm not 100% on supporting it yet.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
srallen: Yes, that's true with how I've implemented it with the scatter plot, but in theory the `VXZoom` component can be potentially integrated elsewhere for other subject viewers.... This might be cruft, but I'm covering my bases here I think.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
srallen: The default event handler provided by the VX Zoom component calls `preventDefault` and I referenced these when writing handlers for our use case: 

https://github.com/hshoff/vx/blob/7a64bed6f5cfc66df4982a8688f1c25770669ea9/packages/vx-zoom/src/Zoom.tsx#L240-L246

According to stack overflow, I should also be returning false to prevent this? https://stackoverflow.com/a/10313183/5372931

This article is interested (And also might have an explanation for some of the observed poor performance): https://www.bennadel.com/blog/3544-trapping-the-wheel-event-may-prevent-chrome-browser-bug-in-which-the-scroll-wheel-stops-working-in-overflow-container.htm

In the comments, there's a CSS property I could try, but it isn't supported in Safari unfortunately: https://caniuse.com/#search=overscroll-behavior

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer/components/ZoomingScatterPlot/ZoomingScatterPlot.js
srallen: I'm not sure the end user will know that they aren't actually fully panned to the allowed range, so as long as the data is available in view since I'm using a constant to add a buffer similar to the LCV, I think it is ok. 

Of course, when we actually test this out with volunteers, my assumption may be proved wrong and I think you're correct that this might have to be adjusted when keyboard control is implemented too.

srallen:
> I'm curious to see how annotation layers will be plugged in. (My previous solution for the LCV was to tie the annotations layer and the zooming controls far tighter than is healthy for modular code maintenance.)

Me too! Since the VSLCV does not require an annotations layer, I've put off investigating how this could work. I hope I haven't accidentally painted myself into a corner with this.

srallen:
>  I assume this is intentional (since in most cases, there's no need to pan if the Subject data isn't zoomed in on... unless there's some sort of Subject data that's very wide...?),

Correct, it doesn't make sense to pan without zoom. It is separately configurable since it's possible that that is desired behavior, but not for the scatter plot. I've noted this in the readme.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
srallen: Hm, the only way I've gotten it to work is to define the overflow styles on the document body when mouse enters and leaves.

eatyourgreens: Interesting. In PFE we prevent the window from scrolling by calling `event.stopPropagation()` to kill the event before it bubbles up to the window.

`return false` from a handler will cancel the event in older browsers, but I'm not sure if any browsers are using that event model any more. It's very old.

srallen: From what I've read, stopping propagation will work with other events that trigger scroll, like keyboard or touch, but not wheel. Chrome also decided to be different than all the others yet again because reasons, so injecting some CSS on certain events was the only way I found it to work consistently cross browser.

PR #1266, opened 2019-11-11
Add initial drawing tools

### LINKED ISSUE #1174: Add Point model

### LINKED ISSUE #1175: Add Line model

eatyourgreens:
Package:
lib-classifier

Line and point drawing tools without rxjs. This PR sets up models for point and line tools, and their corresponding marks. New marks are created when the drawing canvas AKA `InteractionLayer` is clicked/tapped and finalised when the pointer lifts.

The model is:
- a DrawingTask, which contains a set of `tools` and an `activeTool`.
- Line and Point tools, each of which has a `marks` map and a method to create a new mark.
- Line and Point marks, each of which has the properties of the rendered SVG shape and a `toolComponent` property which returns the corresponding React component.

The UI is:
- an `InteractionLayer`, which creates new marks on pointerdown.
- a collection of `MarkComponents`, children of the `InteractionLayer`, which swallow pointer events but don't do anything with them at the moment.

TODO: (possibly in future PRs)
- [x] split mark creation up into `start` and `finish` handlers, so that we can create marks with pointer gestures eg. a new line starts at one point, ends at another.

Closes #1174. 
Closes #1175. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/store/DrawingStore.js
```diff
@@ -76,49 +50,21 @@ const DrawingStore = types
...
-    function createMark () {
-      const MarkModel = getMarkModel(self.activeDrawingTask.tools[self.activeDrawingToolIndex].type)
+    function createMark (event) {
+      const activeTool = self.activeDrawingTask.tools[self.activeDrawingToolIndex]
 
       const tempId = cuid()
 
-      const newMark = MarkModel.create({
+      const newMark = activeTool.createMark({
         id: tempId,
```
srallen: I don't think the unique id is necessary anymore if a reference to it is not being used in the stores.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/tools/Point.js
```diff
@@ -31,8 +33,25 @@ const Point = ({ active, coordinates, scale, tool }) => {
...
+    e.preventDefault()
+    e.stopPropagation()
+    setActive(true)
+  }
+
+  function deselect (e) {
+    console.log('deselect')
+    e.preventDefault()
+    e.stopPropagation()
+    setActive(false)
```
srallen: Would active state be connected with whether or not the delete button shows? If so, then on pointer up, this mark should still be active.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -28,36 +24,51 @@ class InteractionLayerContainer extends Component {
...
+  
   onPointerMove (event) {
-    this.props.addToStream(event)
+    const { activeMark } = this.state
+    activeMark && activeMark.setCoordinates(this.convertEvent(event))
   }
 
   onPointerUp (event) {
-    this.props.addToStream(event)
+    this.setState({ activeMark: null })
```
srallen: Related to the comment I've made about the display of the delete button. What does 'active' mean? That it is selected and can be edited? This is true when you first create the mark with the PFE interaction.

packages/lib-classifier/src/store/DrawingStore.js
eatyourgreens: I think `marks` is still a map, which needs an identifier? This will probably come in useful once we're selecting and editing marks.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/tools/Point.js
eatyourgreens: I copied `active` over from the original PR but I haven't 100% figured out what it means yet.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
eatyourgreens: Good point. If we follow PFE, there should always be one mark that's focussed and editable.

packages/lib-classifier/src/store/DrawingStore.js
srallen: Oh I see, yes.

srallen: Then, we'll need a reference for selecting one out of the map? Removing the `safeReference` type may be premature?

eatyourgreens: It should still be there.

eatyourgreens: Oh wait. I took the active mark out of the store. It's in component state now.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
eatyourgreens: So, I deleted this line and now the initial drawing gesture never completes. A point continues to follow the pointer even after pointerup. 🤔

packages/lib-classifier/src/store/DrawingStore.js
eatyourgreens: This ID is used as a key when rendering marks, so I'm going to leave it in.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/tools/Point.js
eatyourgreens: Also noting that this event handler never runs. I'm not sure what's going on there.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
```diff
@@ -4,16 +4,32 @@ import sinon from 'sinon'
...
+      createMark: sinon.stub().callsFake(() => mockMark)
+    }
+  }
+  const mockSVGEvent = {
+    matrixTransform: sinon.stub().callsFake(() => ({
+      x: 100,
+      y: 200
+    }))
+  }
+  const mockSVG = {
```
eatyourgreens: Is there a better way to create an SVG element for these tests?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
```diff
@@ -27,18 +43,11 @@ describe('Component > InteractionLayer', function () {
...
-
-  it('should bind the onPointerMove prop', function () {
-    wrapper.simulate('pointermove')
-    expect(onPointerMove).to.have.been.called()
-  })
-
-  it('should bind the onPointerUp prop', function () {
-    wrapper.simulate('pointerup')
-    expect(onPointerUp).to.have.been.called()
+  it('should create a mark on pointer down', function () {
```
eatyourgreens: This still needs tests for pointer move and pointer up. Maybe test that the mock mark is correctly updated by those events?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.spec.js
```diff
@@ -2,49 +2,21 @@ import { shallow } from 'enzyme'
...
+        addToStream={() => {}}
+        storeSVG={() => {}}
+      />)
   })
 
   it('should render an InteractionLayer', function () {
-    const wrapper = shallow(<InteractionLayerContainer.wrappedComponent />)
+    const wrapper = shallow(
+      <InteractionLayerContainer.wrappedComponent
+        addToStream={() => {}}
```
eatyourgreens: These props should be updated to reflect changes to the component.

eatyourgreens: Also needs tests for the case where we've got existing marks to render.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/drawingTools/Point.js
```diff
@@ -1,12 +1,29 @@
...
+import { Point as PointMark } from '../markings'
 
 const Point = types.model('Point', {
   color: types.optional(types.string, ''),
   label: types.optional(types.string, ''),
+  marks: types.map(PointMark),
   max: types.maybe(types.union(types.string, types.number), ''),
   min: types.maybe(types.union(types.string, types.number), ''),
   size: types.maybe(types.enumeration(['large', 'small'])),
   type: types.literal('point')
```
eatyourgreens: There's a bunch of tool properties that are shared between the point and line tools, which could be moved into a common Tool model.

yarn.lock
```diff
@@ -14891,7 +14891,7 @@ rxjs@^5.5.2:
   dependencies:
     symbol-observable "1.0.1"
 
-rxjs@^6.4.0, rxjs@~6.5.2:
+rxjs@^6.4.0:
```
eatyourgreens: Is rxjs used somewhere else in the repo?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -3,63 +3,58 @@ import PropTypes from 'prop-types'
...
-        <InteractionLayer
-          onPointerDown={this.onPointerDown}
-          onPointerMove={this.onPointerMove}
-          onPointerUp={this.onPointerUp}
-        />
-        {isDrawingInActiveSteps && <DrawingContainer />}
-      </>
+      <InteractionLayer
+        activeDrawingTask={activeDrawingTask}
+        activeDrawingToolIndex={activeDrawingToolIndex}
```
eatyourgreens: `activeDrawingToolIndex` has been replaced by `activeDrawingTask.activeToolIndex`.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -3,63 +3,58 @@ import PropTypes from 'prop-types'
...
+        createMark={createMark}
```
eatyourgreens: I think I've removed `createMark` in favour of `tool.createMark`.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.spec.js
```diff
@@ -2,49 +2,21 @@ import { shallow } from 'enzyme'
...
-  }
-]
-
 describe('Component > InteractionLayerContainer', function () {
   it('should render without crashing', function () {
-    shallow(<InteractionLayerContainer.wrappedComponent />)
+    shallow(
+      <InteractionLayerContainer.wrappedComponent
+        addToStream={() => {}}
+        storeSVG={() => {}}
```
eatyourgreens: These props have been removed and should be replaced with the new props for the container component.

eatyourgreens:
I can break the commit history up into smaller chunks, to make this easier to review. Otherwise, it is pretty much done.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/Drawing/tools/Point.js
eatyourgreens: I've moved selecting marks into #1308.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
eatyourgreens: `activeMark` here is a mark that is in the process of being created but the drawing tools also use `active` to switch styling ie. an active mark is one that can be edited or deleted.

yarn.lock
srallen: Just discovered this neat command: https://yarnpkg.com/lang/en/docs/cli/why/

eslint and storybook uses it:

Screen Shot 2019-11-19 at 10 18 27 AM

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/DrawingTask.js
```diff
@@ -29,30 +29,17 @@ const StyledText = styled(Text)`
...
-function storeMapper (stores) {
-  const { activeDrawingTool, setActiveDrawingTool } = stores.classifierStore.drawing
-  return {
-    activeDrawingTool,
-    setActiveDrawingTool
-  }
-}
-
-@inject(storeMapper)
 @observer
```
srallen: I don't think this component needs to be an observer anymore. I tried removing it and looks like everything is still working as expected.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -1,24 +1,94 @@
...
+        width='100%'
+        height='100%'
+        fill='transparent'
+      />
+      {activeDrawingTask &&
+        activeDrawingTask.tools.map( tool => {
+          const marksArray = Array.from(tool.marks.values())
+          return marksArray.map(mark => {
+
+            const MarkingComponent = observer(mark.toolComponent)
```
srallen: Why is this being wrapped with an `observer`? Later for the editing functionality?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
srallen: Could this also test that the point tool mark's size has gone up by one?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -3,63 +3,39 @@ import PropTypes from 'prop-types'
...
-
-  onPointerUp (event) {
-    this.props.addToStream(event)
-  }
 
   render () {
-    const { activeStepTasks } = this.props
-    const isDrawingInActiveSteps = activeStepTasks && activeStepTasks.some(task => task.type === 'drawing')
-
+    const { activeDrawingTask, activeDrawingToolIndex, createMark, marks, svg } = this.props
```
srallen: the `activeDrawingToolIndex, createMark, marks` props aren't being used.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.spec.js
srallen: Is this update going to happen in this PR or in a follow up?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.spec.js
srallen: Are the suggested tests going to be updated with this PR or in later PRs?

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/DrawingTask.js
```diff
@@ -62,16 +49,17 @@ class DrawingTask extends React.Component {
         </StyledText>
 
         {task.tools.map((tool, index) => {
```
srallen: It's possible that a drawing task can be created, but with no tools defined, so there needs to be an existential check on the line above before we call map.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
```diff
@@ -0,0 +1,56 @@
+import PropTypes from 'prop-types'
+import React, { useState } from 'react'
+
+const STROKE_WIDTH = 1.5
+const SELECTED_STROKE_WIDTH = 2.5
+
+function DrawingToolRoot ({ children, tool }) {
```
srallen: Non-blocking suggestion: Maybe this can be written as a HOC instead and decorate the marks?

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/markings/Point.js
```diff
@@ -0,0 +1,31 @@
...
+import Mark from './Mark'
+
+const PointModel = types
+  .model('PointModel', {
+    x: types.optional(types.number, 0),
+    y: types.optional(types.number, 0)
+  })
+  .views(self => ({
+    get toolComponent () {
+      return PointComponent
```
srallen: Nice to see this works.

eatyourgreens:
I think a few of the suggested changes are taken care of in #1308.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
eatyourgreens: I'll give that a go.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
eatyourgreens: I think I've rewritten all of this in #1308, so that you can select marks as well as create them.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.spec.js
eatyourgreens: It should be done here, but whichever approach gets this out of the way fastest, really.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/drawingTools/Point.js
eatyourgreens: Redoing the drawing tool models probably rates its own PR based off either this or #1308.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -1,24 +1,94 @@
...
+import React, { useState } from 'react'
+
+function InteractionLayer ({ activeDrawingTask, svg }) {
+  const [ activeMark, setActiveMark ] = useState(null)
+  
+  function convertEvent (event) {
+    const type = event.type
+
+    const clientX = event.clientX
+    const clientY = event.clientY
```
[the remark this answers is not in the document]
eatyourgreens: I think I've changed that in #1308.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -1,24 +1,94 @@
...
+    const svgEventOffset = getEventOffset(clientX, clientY)
+
+    const svgCoordinateEvent = {
+      pointerId: event.pointerId,
```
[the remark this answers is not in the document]
eatyourgreens: Ditto.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -1,24 +1,94 @@
...
+    >
+      <rect
+        id='InteractionLayer'
+        width='100%'
+        height='100%'
+        fill='transparent'
+      />
+      {activeDrawingTask &&
+        activeDrawingTask.tools.map( tool => {
+          const marksArray = Array.from(tool.marks.values())
```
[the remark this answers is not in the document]
eatyourgreens: `tools.marks` defaults to an empty map.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -1,24 +1,94 @@
...
+          return marksArray.map(mark => {
```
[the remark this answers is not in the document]
eatyourgreens: `array.map()` doesn't iterate if `array` is empty.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -3,63 +3,39 @@ import PropTypes from 'prop-types'
...
-InteractionLayerContainer.wrappedComponent.propTypes = {
-  activeStepTasks: PropTypes.arrayOf(
-    PropTypes.shape({
-      type: PropTypes.string
-    })
-  ),
-  addToStream: PropTypes.func.isRequired
+InteractionLayerContainer.propTypes = {
+  isDrawingInActiveWorkflowStep: PropTypes.bool,
+  svg: PropTypes.object
```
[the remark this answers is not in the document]
eatyourgreens: No idea, I'm afraid. This PR is a straight rewrite of #1249 without rxjs but I tried not to touch too much of that code.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/DrawingTask.js
```diff
@@ -62,16 +49,17 @@ class DrawingTask extends React.Component {
...
           return (
             <TaskInput
               checked={checked}
+              disabled={tool.isComplete}
               index={index}
               key={`${task.taskKey}_${index}`}
               label={tool.label}
               labelIcon={<InputIcon icon={<ToolIcon type={tool.type} />} tool={tool} />}
-              labelStatus={<InputStatus tool={tool} />}
+              labelStatus={<InputStatus count={tool.marks.size} tool={tool} />}
```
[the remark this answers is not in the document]
eatyourgreens: `tool.marks` defaults to an empty map.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
```diff
@@ -0,0 +1,56 @@
...
+  )
+}
+
+DrawingToolRoot.propTypes = {
+  active: PropTypes.bool,
+  children: PropTypes.node.isRequired,
+  tool: PropTypes.shape({
+    color: PropTypes.string
+  })
+}
```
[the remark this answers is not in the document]
eatyourgreens: I think this is all handled in #1308.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/Line.js
```diff
@@ -2,22 +2,18 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import DrawingToolRoot from './DrawingToolRoot'
 
-const Line = ({ active, coordinates, finishDrawing, tool }) => {
-  if (!coordinates) return null
-  if (active && coordinates.type === 'pointerup') {
-    finishDrawing(coordinates)
-  }
+function Line ({ active, mark, tool }) {
```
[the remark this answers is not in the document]
eatyourgreens: It's a placeholder for selecting and moving a mark in #1308.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/Line.js
```diff
@@ -2,22 +2,18 @@ import PropTypes from 'prop-types'
...
+  const { x1, y1, x2, y2 } = mark
 
   return (
     <DrawingToolRoot active tool={tool}>
```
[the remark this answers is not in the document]
eatyourgreens: #1308 removes the drawing tool root from these mark components.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/Line.js
```diff
@@ -2,22 +2,18 @@ import PropTypes from 'prop-types'
...
-      <line {...coordinates} />
+      <line x1={x1} y1={y1} x2={x2} y2={y2} />
     </DrawingToolRoot>
   )
 }
 
 Line.propTypes = {
   active: PropTypes.bool,
```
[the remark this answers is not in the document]
eatyourgreens: See #1308.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/Point.js
```diff
@@ -29,14 +29,9 @@ const Point = ({ active, coordinates, finishDrawing, scale, tool }) => {
     radius = RADIUS[size] / averageScale
   }
 
-  if (!coordinates) return null
-  if (active && coordinates.type === 'pointerup') {
-    finishDrawing(coordinates)
-  }
-
   return (
     <DrawingToolRoot active tool={tool}>
```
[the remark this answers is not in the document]
eatyourgreens: See #1308.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/drawingTools/Line.js
```diff
@@ -1,11 +1,29 @@
 import { types } from 'mobx-state-tree'
+import { Line as LineMark } from '../markings'
 
 const Line = types.model('Line', {
   color: types.optional(types.string, ''),
   label: types.optional(types.string, ''),
+  marks: types.map(LineMark),
   max: types.maybe(types.union(types.string, types.number), ''),
   min: types.maybe(types.union(types.string, types.number), ''),
```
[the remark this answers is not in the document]
eatyourgreens: I copied over the types from #1249. I think the drawing tools need their own PR that sets up an abstract tool model which is extended by actual tools.

Looking at the diff, is the incorrect type on master?

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/drawingTools/Point.js
```diff
@@ -1,12 +1,29 @@
 import { types } from 'mobx-state-tree'
+import { Point as PointMark } from '../markings'
 
 const Point = types.model('Point', {
   color: types.optional(types.string, ''),
   label: types.optional(types.string, ''),
+  marks: types.map(PointMark),
   max: types.maybe(types.union(types.string, types.number), ''),
   min: types.maybe(types.union(types.string, types.number), ''),
   size: types.maybe(types.enumeration(['large', 'small'])),
```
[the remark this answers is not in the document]
eatyourgreens: Default should be 'large', which I see is actually set in the Point code.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/DrawingTask.js
eatyourgreens: I found that the task area didn't update when you created and deleted marks, but I can have another look in #1308.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
eatyourgreens: So that marks move on the screen when you drag them and change the coordinates in their model.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
eatyourgreens: I'm going to add a `task.marks` view back to the `DrawingTask` model in #1308 so that marks disappear when you delete them.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/DrawingTask.js
eatyourgreens: `task.tools` defaults to an empty array.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
eatyourgreens: See #1308.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/DrawingTask.js
eatyourgreens: When I remove this, I see the point tool saying '0 drawn' after I've made two marks.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
eatyourgreens: That gives a type error: `TypeError: svg.createSVGPoint is not a function`, which I'm also getting in the storybook for #1308.

eatyourgreens: D'oh. `document.createElement` creates a `HTMLElement` so, yeah, that won't have the SVG interfaces. It'll probably work if I add the appropriate namespace for `svg`.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
eatyourgreens: >Could this also test that the point tool mark's size has gone up by one?

The tests are using a generic `MockMark` but could, I think, be changed to use an actual mark model. The Point model doesn't store any info about the marker's size, though.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
eatyourgreens: `document.createElementNS` with the SVG namespace doesn't work either.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
eatyourgreens: See #1308.

srallen:
Ok I'll look over #1308 today.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
```diff
@@ -3,17 +3,44 @@ import React from 'react'
...
+    setCoordinates: sinon.stub()
+  }
+  const mockDrawingTask = {
+    activeToolIndex: 0,
+    activeTool: {
+      createMark: sinon.stub().callsFake(() => mockMark)
+    },
+    tools: [
+      {
+        marks: new Map([]),
```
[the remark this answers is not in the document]
eatyourgreens: I opted to mock the drawing task properly instead, and create a new model.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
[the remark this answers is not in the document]
eatyourgreens: Looking back through the code, `svg` is a reference to a DOM node that's taken from a ref forwarded down from the parent image viewer? So it's null if I render `<SingleImageViewer />` without a ref? The code is kind of confusing.

eatyourgreens:
Re. the marks appearing small, the pixel sizes are fixed constants in the code so they will appear smaller on larger images. There's a note in [the drawing tools project](https://github.com/zooniverse/front-end-monorepo/projects/11) that scaling is still to be implemented.

eatyourgreens:
Updating the `InteractionLayer` tests has broken tests downstream in #1308 but I'll deal with that another day.

shaunanoordin:
⚠️ Broken functionality: in PR #1338 I approved a change that **breaks the app when there's a Question Task.**

Loading a Project with a Question-type Task such as [I Fancy Cats](http://localhost:8080/?env=staging&project=335) will cause the app to crash, with the dev console error:

```
Uncaught TypeError: Cannot read property 'activeTool' of undefined
    at storeMapper (InteractionLayerContainer.js:50)
```

I'm spinning off a PR to fix this, since I'm pretty sure the problem is just that the InteractionLayer assumes `activeDrawingTask` is always defined.

@eatyourgreens Before I make my fix, I wanted to confirm: the InteractionLayer should exist _regardless of the current Task,_ correct? I assume that eventually, the "Drag to Pan View" function that's available to all Workflows/Tasks will also be handled by this InteractionLayer.

shaunanoordin:
> @eatyourgreens Before I make my fix, I wanted to confirm: the InteractionLayer should exist regardless of the current Task, correct?

Wait, scratch that assumption. I re-read the code:

```
class InteractionLayerContainer extends Component {
  render () {
    const { activeDrawingTask, activeTool, disabled, svg } = this.props
    return activeDrawingTask ? 
    <InteractionLayer
      activeDrawingTask={activeDrawingTask}
      activeTool={activeTool}
      disabled={disabled}
      svg={svg}
    /> : 
    null
  }
}
```

This implies that the `InteractionLayer` doesn't even appear if there's no activeDrawingTask, so InteractionLayer is dedicated for Drawing Tasks ONLY. I assume Drag-to-Pan functionality is another problem for another component for another time.

Anyway, this makes fixing the bug WAY easier 👌

eatyourgreens:
I forgot the new `activeTool` prop in the tests too. I've just pushed a fix for that.

shaunanoordin:
Oh no, my changes from #1341 are breaking the tests in `InteractionLayerContainer.spec.js`. I forgot to factor in that `InteractionLayerContainer.js` now only renders the child Interaction Layer if `activeDrawingTask` AND `activeTool` are rendered:

```
describe('with an active drawing task', function () {
  it('should render an InteractionLayer', function () {
    const wrapper = shallow(<InteractionLayerContainer.wrappedComponent activeDrawingTask />)
    // activeDrawingTask is empty here, if I'm correct.
    expect(wrapper.find('InteractionLayer')).to.have.lengthOf(1)
  })
})
```

I'll fix this.

shaunanoordin:
Wait, I just saw c67129c60af6291f9cb424cb27d70f279ada4d1b - jim++

eatyourgreens:
I've already fixed that.

eatyourgreens:
Nice to have coveralls back.

eatyourgreens:
Basic drawing functionality is all done now. I think any other drawing features can come in separate PRs.

eatyourgreens:
Coveralls does not like the lack of tests for drawing components.

shaunanoordin:
At the time of writing - and commit 680ad23 - the `drawing-remove-rxjs` PR looks to be in a very stable place, with only a few missing parts, and some quirks.

Features:
- Workflows can have Drawing Tasks with "Point" (click to mark) and "Line" (click-and-drag to mark) tools
- Drawing marks are scaled according to the size of the Subject Image to remain significantly visible to the user.
- Drawing marks are maintained as user traverses between Tasks. (Previous annotations are visible but inert on the Subject Viewer as user progresses through to the next Tasks.)
- Drawing annotations are submitted with Classifications

Missing:
- Documentation
- Tests?
- Mouse Event handlers, which are required for Safari, since that browser doesn't know how to handle Pointer Events. (Proposed Solution, AFAIK, is to use Polyfills instead.)
- prevent marks from being drawn out of bounds. (PR #1334) 

Quirks:
- Drawing annotations submitted with Classifications are in a format different from PFE's
- ❓ if you create a **Line** Drawing Tool that has a Max limit of (say) 3, then the 3rd line you try draw behaves very poorly.
  - You can still draw the line if you move the cursor veeery slowly... (i.e. allow the line's end point circle to always catch up with yoru cursor)
  - ...but if you move the mouse too fast, the drawing action breaks and the actively drawn line exists in limbo (i.e. the InteractionLayer doesn't listen for pointer movement, or pointer up events.)
  - I suspect it's because _as soon as_ the InteractionLayer detects that number_of_annotations_for_current_tool === max_allowed_for_current_tool, then it _automatically "switches off" any interactions on the InteractionLayer_ (i.e. setting pointer-events: none), regardless of whether there's an active tool being used to draw something.

Tested on `http://localhost:8080/?env=staging&project=1895` on macOS 10 + Chrome (Incognito) and Firefox 70

Formal review incoming

eatyourgreens:
>I suspect it's because as soon as the InteractionLayer detects that number_of_annotations_for_current_tool === max_allowed_for_current_tool, then it automatically "switches off" any interactions on the InteractionLayer (i.e. setting pointer-events: none), regardless of whether there's an active tool being used to draw something.

That's a good point. I think the `disabled` flag is set from the size of the active marks map. Is it worth fixing here or in a separate issue?

shaunanoordin:
> That's a good point. I think the disabled flag is set from the size of the active marks map. Is it worth fixing here or in a separate issue?

Separate issue, I think - this PR's looking good as is.

eatyourgreens:
Maybe the answer is not to put a new mark into the active marks map until the initial gesture is finished and validated. I'm fine with trying that in a separate PR.

Issue #19, opened 2018-04-23
Subject Renderer component


rogerhutchings:
(via the ZTM dicussion)

This component should:

- not update parent app state
- only manage local state like pan/zoom settings

PR #1268, opened 2019-11-12
Add ADR-20 about the transcription task

srallen:

Package: ADR

Describe your changes:
Adding a proposal ADR describing the transcription task. We need to make sure we clearly understand what we're doing here. I've included a link to the user stories @beckyrother put together. I've noted in the consequences that the transcription line mark has a different user interaction than the standard line mark. 

Another question is whether or not we can use the HTML datalist element for the input/dropdown for the previous transcriptions. If not, I might advise that we just have the text task input and a separate dropdown for previous transcriptions rather than some kind of custom combo function component like the current PFE dropdown. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/docs/arch/adr-20.md
+++ b/docs/arch/adr-20.md
@@ -4,7 +4,7 @@ Created: November 12, 2019
 
 ## Context
 
-ILMS funded us to experiment with transcription tools. The project [Anti-Slavery Manuscripts](https://www.antislaverymanuscripts.org/classify) has a workflow where volunteers transcribe lines of text first by marking the line, then adding the transcription in a sub-task. This task is slightly unique in that:
+IMLS funded us to experiment with transcription tools. The project [Anti-Slavery Manuscripts](https://www.antislaverymanuscripts.org/classify) has a workflow where volunteers transcribe lines of text first by marking the line, then adding the transcription in a sub-task. This task is slightly unique in that:
 
 - Incomplete classifications can be submitted
 - Previous transcriptions from caesar are loaded and presented as an option in the sub-task
@@ -15,12 +15,16 @@ ILMS funded us to experiment with transcription tools. The project [Anti-Slavery
 
 We will be porting the ASM functionality to the main classifier as a new transcription task. The task will be composite of:
 
-- A drawing line task that is created by two pointer down and up events to mark the points of the line.
-  - the starting point and ending point will have visual indicators.
+- A drawing line task for transcription that is created by two pointer down and up events to mark the points of the line.
+  - the starting point and ending point will have visual indicators communicating the direction of creation.
+  - the line color will indicate the current status of completion
+  - completed lines cannot be edited
   - Note that ASM leveraged a hacked polygon tool to do this, but we should have a line tool variant specifically for this instead. Downstream aggregation can leverage polygon type aggregations if it fits.
 - A sub-task will display once the transcription line mark is made
-  - a datalist input will display showing the previous caesar aggregations
-  - sub-task will not be a modal
+  - a text input will display suggestions from previous caesar aggregations.
+    - The suggestions can be selected and inserted as an editable value in the text input
+    - Grommet's [TextInput](https://storybook.grommet.io/?path=/story/textinput--suggestions) with the suggestions prop will be utilized so that it can be styleable with our theme.
+  - sub-task will not be a modal, but a movable div linked to the currently selected line mark
 
 [More detailed user stores are in this google doc](https://docs.google.com/document/d/16abI-wkRlEXsWgACfFQVqwO76aEopohIjQiRfNQKWiw/edit)
 
@@ -32,5 +36,12 @@ Proposed
 
 - The transcription line mark is a different user interaction than the standard line mark. The standard line is a dragging interaction rather than two clicks.
 - The workflow lab editor will need a task option for "transcription task" which will auto-configure it with the drawing task with sub-task described in the decision
-- Can the [HTML datalist element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist) be used? If not, then we might want to consider a text input and a separate dropdown of available Caesar aggregations. 
+
+Open questions:
+
+- How will cross-writing, where the writer might've rotated the page to begin writing again, be handled?
+- How will a line of text left to right be distinguished from one rotated 90 degrees and written again?
+- How will left-to-right lines of text be distinguished from right-to-left?
+
+We are planning on having visual indication via the style of the SVG dots at the end of the line to indicate directionality, but this is not recorded in the annotation. Adding rotation and line direction as drawn in the annotation could be used for downstream aggregation to alleviate these questions. We will check with @CKrawczyk and update this proposal when an annotation format is agreed upon for this. We can deviate in the annotation model from a standard line drawing plus text task sub-task since this will be a new 'transcription' task in the models. 
 
--- a/docs/arch/adr-20.md
+++ b/docs/arch/adr-20.md
@@ -1,0 +1,36 @@
+# ADR 20: Transcription Task
+
+Created: November 12, 2019
+
+## Context
+
+ILMS funded us to experiment with transcription tools. The project [Anti-Slavery Manuscripts](https://www.antislaverymanuscripts.org/classify) has a workflow where volunteers transcribe lines of text first by marking the line, then adding the transcription in a sub-task. This task is slightly unique in that:
+
+- Incomplete classifications can be submitted
+- Previous transcriptions from caesar are loaded and presented as an option in the sub-task
+  - volunteers can select the previous transcription, edit it, and submit a new transcription
+  - lines with previous transcriptions or retired lines are displayed visually by color
+
+## Decision
+
+We will be porting the ASM functionality to the main classifier as a new transcription task. The task will be composite of:
+
+- A drawing line task that is created by two pointer down and up events to mark the points of the line.
+  - the starting point and ending point will have visual indicators.
+  - Note that ASM leveraged a hacked polygon tool to do this, but we should have a line tool variant specifically for this instead. Downstream aggregation can leverage polygon type aggregations if it fits.
+- A sub-task will display once the transcription line mark is made
+  - a datalist input will display showing the previous caesar aggregations
+  - sub-task will not be a modal
+
+[More detailed user stores are in this google doc](https://docs.google.com/document/d/16abI-wkRlEXsWgACfFQVqwO76aEopohIjQiRfNQKWiw/edit)
+
+## Status
+
+Proposed
+
+## Consequences
+
+- The transcription line mark is a different user interaction than the standard line mark. The standard line is a dragging interaction rather than two clicks.
+- The workflow lab editor will need a task option for "transcription task" which will auto-configure it with the drawing task with sub-task described in the decision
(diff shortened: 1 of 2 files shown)


srallen:
Here's an example of how the datalist could work: https://stackblitz.com/edit/react-html-datalist-autocomplete

srallen:
ASM is using an input and buttons for previous transcriptions: https://github.com/zooniverse/anti-slavery-manuscripts/blob/2289958ef0b12fd528d8783c410c6a943270790d/src/components/SelectedAnnotation.jsx#L227-L240

Is the advantage of this method for styles? Do we really need to style this if we can get a lot of functionality out of the native HTML datalist?

srallen:
@shaunanoordin or @wgranger do you have any information about the approach you went with for ASM specifically for the text input and display of the previous transcriptions as buttons?

beckyrother:
@srallen what should be reviewed?

srallen:
@beckyrother I need a sanity check on the summary description of what the transcription task is in this ADR.

srallen:
Note: Grommet's TextInput can be configured to have a suggestion list and the input is editable: https://storybook.grommet.io/?path=/story/textinput--suggestions

docs/arch/adr-20.md
```diff
@@ -0,0 +1,36 @@
+# ADR 20: Transcription Task
+
+Created: November 12, 2019
+
+## Context
+
+ILMS funded us to experiment with transcription tools. The project [Anti-Slavery Manuscripts](https://www.antislaverymanuscripts.org/classify) has a workflow where volunteers transcribe lines of text first by marking the line, then adding the transcription in a sub-task. This task is slightly unique in that:
```
wgranger: This should read "IMLS" funded us...

docs/arch/adr-20.md
```diff
@@ -0,0 +1,36 @@
...
+  - volunteers can select the previous transcription, edit it, and submit a new transcription
+  - lines with previous transcriptions or retired lines are displayed visually by color
+
+## Decision
+
+We will be porting the ASM functionality to the main classifier as a new transcription task. The task will be composite of:
+
+- A drawing line task that is created by two pointer down and up events to mark the points of the line.
+  - the starting point and ending point will have visual indicators.
+  - Note that ASM leveraged a hacked polygon tool to do this, but we should have a line tool variant specifically for this instead. Downstream aggregation can leverage polygon type aggregations if it fits.
```
wgranger: We had to abide by the polygon tool because that was the only tool supported through the aggregator at the time (2017). A line tool aggregator should now exist.

wgranger:
Left a couple comments. As for displaying previous transcriptions as buttons:

We would display previous images/frames of subject as previous and next buttons of a filmstrip viewer. The image buttons would not contain superimposed transcription lines. Very important to remember to use a thumbnail image for these buttons, because loading the page with 7 full size images is much more demanding than one full size image and six thumbnails.

docs/arch/adr-20.md
wgranger: https://github.com/zooniverse/aggregation-for-caesar

srallen:
> We would display previous images/frames of subject as previous and next buttons of a filmstrip viewer. The image buttons would not contain superimposed transcription lines. Very important to remember to use a thumbnail image for these buttons, because loading the page with 7 full size images is much more demanding than one full size image and six thumbnails.

ADR #19 in #1260 is about the pages filmstrip viewer which is similar to the ASM one. We're planning on doing an initial build to meet the requirements for transcription use, but it'll expand to eventually be the default viewer for multi-image subjects. Thanks for confirm that using the thumbnail service is the right direction for this. cc @simensta about this comment.

srallen:
> However, I'll note that the 2-click method of polygon-as-line tool was much easier to use than having to drag, which is the case for the regular line tool, so I'd be in favor of continuing to use the adapted polygon tool for collaborative transcription if it's under question.

Yes, we'll continue to plan on using a 2-click, but as a sub-classed line tool rather than a sub-classed polygon. This is more for code clarity. The volunteer shouldn't see a difference in function. 

> do we need to mention the process of 'greying out' annotations as it relates to this tool, or is that part of a separate issue?

We do have a separate issue for implementing caesar aggregation: #1271. We should include a clear description of how that will function as a part of this, however. I'll update the summary to include that.

srallen:
> How will you handle cross-writing, where the writer might've rotated the page to begin writing again? How will you distinguish a line of text left to right from one rotated 90 degrees and written again.

> Any concern for right to left transcriptions?

Aren't these downstream aggregation concerns? Are you suggesting we include some kind of data in the raw annotations to help with these? Since this is going to be a separate 'transcription task' that project builders will have to select, we could in theory change the annotations to better fit if needed and diverge from a drawing task + sub-task details JSON structure. I hesitate to modify the annotation structure and what is included, though, unless @CKrawczyk is asking for it. We can ask him about it when he's back in the office.

eatyourgreens:
Right-to-left versus left-to-right is no big deal when you're measuring galaxies and giraffes, but can be a problem when you're aggregating text (from what I remember of Coleman's talk about analysing text transcriptions.) Do we have an example of a line annotation from ASM?

srallen:
One of the feature requests is to visually indicate with the dot parts of the mark start and finish of the line direction. We could include directionality in the annotation if it is needed I think.

eatyourgreens:
Sounds good. If the transcription tool ends up being a new tool, we want to avoid having to generate models from ambiguous snapshots.

snblickhan:
Just chiming in to note that aggregation is indeed direction-specific, so having visual start/end indicators would be a major step toward ensuring support for RTL text transcription projects.

srallen:
Adding a WIP to make sure this doesn't accidentally merge before getting final approval from @snblickhan

srallen:
I took a look at the classification JSON structure that ASM submits and it's a standard polygon with a sub-task:

``` json
{
 "task": "T1",
  "value": [
      "details": [
         { "value": "" }
      ],
      "frame": 0,
      "points": [
        { x: 1, y: 1 },
        { x: 100, y: 100 } 
      ]
  ]
}
```

If we modify this for the composite transcription task, maybe the JSON includes a page metadata that records the the current page rotation? Also the direction in which the points were created either `'ltr'` or `'rtl'`? We can update the ADR once we decide. 

@snblickhan ASM has a second task asking if all the lines are transcribed or not? Should this be standard with the transcription task, too? I don't see this in the user stories doc I've linked to and I just want to confirm that.

snblickhan:
> I took a look at the classification JSON structure that ASM submits and it's a standard polygon with a sub-task:
> 
> ```json
> {
>  "task": "T1",
>   "value": [{
>       "details": [
>          { "value": "" }
>       ],
>       "frame": 0,
>       "points": [
>         { "x": 1, "y": 1 },
>         { "x": 100, "y": 100 } 
>       ]
>   }]
> }
> ```
> 
> If we modify this for the composite transcription task, maybe the JSON includes a page metadata that records the the current page rotation? Also the direction in which the points were created either `'ltr'` or `'rtl'`? We can update the ADR once we decide.

I'd definitely want to involve @CKrawczyk in that conversation. There's definitely a way to tell using the slope of the line, but the parameters will likely be different for various documents. (e.g. if main text written RTL is 0 and main text written LTR is 180, one author's writing in the margins may be at 90 degrees, one at 270, etc. -- possibly even on the same page).
> 
> @snblickhan ASM has a second task asking if all the lines are transcribed or not? Should this be standard with the transcription task, too? I don't see this in the user stories doc I've linked to and I just want to confirm that.

Instead of asking if all the lines are transcribed, we'll want to ask if all the lines are greyed out. E.g. 3 users may transcribe the entire document, but that doesn't mean they'll have reached consensus, so we want to make sure we're asking if consensus has been reached, which is indicated by grey lines.

eatyourgreens:
You can get the direction and angle of the line from the order of the points array. I'm pretty sure that's how the aggregation code figures it out, which is why it's important that volunteers create the first point at the beginning of the line and last point at the end.

Direction may also be upward/downward as well as left-to-right or right-to-left.

docs/arch/adr-20.md
```diff
@@ -0,0 +1,47 @@
...
+
+- A drawing line task for transcription that is created by two pointer down and up events to mark the points of the line.
+  - the starting point and ending point will have visual indicators communicating the direction of creation.
+  - the line color will indicate the current status of completion
+  - completed lines cannot be edited
+  - Note that ASM leveraged a hacked polygon tool to do this, but we should have a line tool variant specifically for this instead. Downstream aggregation can leverage polygon type aggregations if it fits.
+- A sub-task will display once the transcription line mark is made
+  - a text input will display suggestions from previous caesar aggregations.
+    - The suggestions can be selected and inserted as an editable value in the text input
+    - Grommet's [TextInput](https://storybook.grommet.io/?path=/story/textinput--suggestions) with the suggestions prop will be utilized so that it can be styleable with our theme.
```
eatyourgreens: The text task uses `TextArea` so that long lines wrap inside the available area. The lines might be too long to comfortably display inside a `TextInput`, which is designed for short strings like names, passwords etc.

docs/arch/adr-20.md
```diff
@@ -0,0 +1,47 @@
...
+  - sub-task will not be a modal, but a movable div linked to the currently selected line mark
```
eatyourgreens: The sub-task should be modal to the extent that creating new marks is blocked intil the current line is complete.

eatyourgreens: The survey task opens its Identify modal inside the task area. Can we copy that behaviour for the transcription text task?

If the sub-task must be a popup, then it must not obscure the region of the page being transcribed and it must be moveable from the keyboard.

docs/arch/adr-20.md
```diff
@@ -0,0 +1,47 @@
...
+Proposed
+
+## Consequences
+
+- The transcription line mark is a different user interaction than the standard line mark. The standard line is a dragging interaction rather than two clicks.
+- The workflow lab editor will need a task option for "transcription task" which will auto-configure it with the drawing task with sub-task described in the decision
+
+Open questions:
+
+- How will cross-writing, where the writer might've rotated the page to begin writing again, be handled?
```
eatyourgreens: The beginning and end points will be above and below each other.

docs/arch/adr-20.md
```diff
@@ -0,0 +1,47 @@
...
+- How will a line of text left to right be distinguished from one rotated 90 degrees and written again?
+- How will left-to-right lines of text be distinguished from right-to-left?
```
eatyourgreens: Again, from the direction of the beginning and end points of the text.

srallen: Yes, visually, however we haven't decided how to record that in the annotation. I'll be merging this and opening an RFC about the annotation structure for the transcription task and this ADR will be updated once that RFC is resolved.

docs/arch/adr-20.md
srallen: The design of the sub-task is a movable pop-up. I'll be opening an issue to track this. If you'd like to suggest an alternative design, it can be discussed there and this ADR can be updated as needed.

docs/arch/adr-20.md
srallen: We can create stories to check the behavior.

srallen:
Merging with proposal status and I'll open follow up issues for dev work and/or RFCs.

Issue #1052, opened 2019-07-29
Drawing tasks - potential design issues


mcbouslog:
Drawing tools preview on staging: https://frontend.preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground/classify?env=staging

This issue is a placeholder for formal documentation and discussion, please edit this initial comment however see fit.

**Package**
lib-classifier

**Feature or Issue Description**
drawing tasks, including task inputs, markings, anything related

**Potential Issues**
- drawing tool input icon background has poor contrast in light theme, see [these related comments](https://github.com/zooniverse/front-end-monorepo/pull/913#pullrequestreview-265687164)
- review each tool stroke width, design
- review marking's delete button
- color options

beckyrother:
[Here's what](https://projects.invisionapp.com/d/main#/console/12923997/376615153/build) I was thinking for the button background. Click the first "Line" button to see what it would look like selected. 

All of the colors in the [DSM](https://projects.invisionapp.com/dsm/zooniverse/primary-brand/folder/colors/5c68fab7fb26bb001855f29) aside from the royal blue have accessibility ratings of AA or higher, so we can leave that one out if it's not needed. 

Thoughts?

srallen:
The cursor icons currently used by PFE:
`crosshair` cursor when a drawing tool is selected and hovering over the interactive layer
`pointer` cursor on hover of the delete button
`grab` cursor on hover of the mark
`move` cursor on hover of the drag handle
`grabbing` cursor while dragging the mark
`not-allowed` cursor for the whole area if the max number of marks is already drawn.

eatyourgreens:
Tool choices are disabled when you hit the maximum number of marks for a given tool, but there's no visible styling for disabled tools in the task area. Try it here on master: http://localhost:8080/?project=908&workflow=3370. The subject shows the `not-allowed` cursor so this might not be a big deal.

eatyourgreens:
There should be an animation of some sort when a mark is deleted.

beckyrother:
Yes. Using Penguin Watch as an example, the mark should quickly fade down and shrink in size as shown there, but not move down to the right.

eatyourgreens:
https://github.com/zooniverse/front-end-monorepo/issues/1400

mcbouslog:
> The cursor icons currently used by PFE:
> `crosshair` cursor when a drawing tool is selected and hovering over the interactive layer
> `pointer` cursor on hover of the delete button
[...]

PR #1308, opened 2019-11-15
Drawing: select, move and delete marks

### LINKED ISSUE #305: Task validations
**Package** lib-classifier

Current PFE has validations to check for completeness for each task type, i.e. https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/single/index.jsx#L77-L79

Its purpose according to PFE's readme:

> Given a task and an annotation, this determines whether or not the classifier will allow the user to move on to the next task.

It's checked in the task nav: https://github.com/zooniverse/Panoptes-Front-End/blob/21cf42485929a62938112d5e2d4bca1a6702b00e/app/classifier/task-nav.jsx#L75-L79

These were static methods on the components. Instead, we could add these to the annotation type model using an onAction MST listener?

**To Dos**

- [ ] Example

### LINKED ISSUE #1105: Drawing task - draggable
**Package**
lib-classifier

Allow markings from a drawing task in the current workflow step to be draggable.

**To Dos**
...placeholder for project management, details TBD...

### LINKED ISSUE #1106: Drawing task - delete marking

eatyourgreens:
Package:
lib-classifier

- Differentiates between creating a mark (initial click and drag) and selecting a mark (click an existing mark and, optionally, drag.) 
 - Adds `mark.isValid`, which is checked on pointer up. Invalid marks are deleted.
 - starts an API for passing position changes to the Mark model, including `mark.move({ x, y })`, where `{ x, y }` is the relative change in the mark position.
 - Allows marks to be selected and dragged with a pointer.
 - Adds the delete button to marks.
 - Adds drag handles to the line tool, and a `draggable` decorator for SVG components.
 - updates the canvas AKA `InteractionLayer` when marks are moved, resized or deleted.

Closes #1106.
Closes #1105. 
Closes #305. 

Line and Point tool workflow for testing: http://localhost:8080/?project=1876&workflow=3344

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
```diff
@@ -1,46 +1,107 @@
...
     color: tool && tool.color ? tool.color : 'green',
     fill: 'transparent',
     stroke: tool && tool.color ? tool.color : 'green',
-    strokeWidth: active ? SELECTED_STROKE_WIDTH : STROKE_WIDTH
+  }
+
+  function convertEvent (event) {
+    const type = event.type
+
+    const svgEventOffset = getEventOffset(event.clientX, event.clientY)
```
eatyourgreens: Client coords change when the window is scrolled. Use page coordinates or lock the window scroll position while drawing.

eatyourgreens: I think this will only be a problem if the page scrolls while you're in the middle of a drawing interaction.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
```diff
@@ -1,49 +1,57 @@
...
   }
 
-  function onPointerDown (event) {
-    console.log(`${tool.type} clicked`)
-    event.stopPropagation()
-    event.preventDefault()
-    setActive(true)
+  function onKeyDown (event) {
+    switch (event.key) {
+      case 'Backspace': {
```
srallen: Just as a follow up to the conversation on Slack from last week. I can delete a mark using either space, enter, or backspace _only_ when the delete button is focused. I think I'm fine with this. My concern was the backspace behavior generally when on the site while on the classifier page using Firefox (and other browsers with this behavior). @beckyrother is this ok with you?

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/components/draggable/draggable.stories.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+storiesOf('draggable', module)
+  .addDecorator(withKnobs)
+  .addParameters({
+    viewport: {
+      defaultViewport: 'responsive'
+    }
+  })
+  .add('basic', function () {
```
srallen: This story doesn't seem to work for me. I can't drag the blue circle.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
eatyourgreens: It should really respond to Backspace when the mark itself has focus, but I've just realised I haven't copied that over from PFE either.

Overriding `window.onKeyDown` is common in key-binding libraries but I'd always advise against using it.

NB. you can also activate the button with the spacebar, which is standard behaviour but overrides the spacebar as PageDown.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/components/draggable/draggable.stories.js
eatyourgreens: If you look in the console, you'll see there's an error because the `<svg>` tag in the story isn't an `SVGElement` in the DOM. I have no idea why the browser isn't rendering it properly. Namespaces maybe?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
```diff
@@ -61,4 +63,17 @@ describe('Component > InteractionLayer', function () {
...
+  })
+
+  it('should drag the new mark on pointer down + move', function () {
+    const fakeEvent = {
+      type: 'pointer'
+    }
+    wrapper.simulate('pointerdown', fakeEvent)
+    wrapper.simulate('pointermove', fakeEvent)
+    expect(mockMark.initialDrag).to.have.been.calledOnce()
+  })
```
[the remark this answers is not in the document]
eatyourgreens: The drawing API inside the `InteractionLayer` has changed a lot and, I think, needs a bunch of tests. I'm also still inclined to rename it to `DrawingCanvas`.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
eatyourgreens: Having said that, we should probably get rid of the default WebKit focus outline. I guess PFE removes it in the CSS somewhere?
Screenshot of drawing marks showing a big, blue focus outline around the selected mark.

eatyourgreens:
@srallen I've added mark validation to this PR but we should be careful to keep it focussed, otherwise it's just going to grow as we remember features of PFE that aren't added yet.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/Line.js
```diff
@@ -1,14 +1,24 @@
...
+    mark.setCoordinates(coords)
+  }
+
   return (
-    <DrawingToolRoot active tool={tool}>
+    <g>
       <line x1={x1} y1={y1} x2={x2} y2={y2} />
-    </DrawingToolRoot>
+      <line x1={x1} y1={y1} x2={x2} y2={y2} strokeWidth={GRAB_STROKE_WIDTH} strokeOpacity="0" />
+      <DragHandle svg={svg} x={x1} y={y1} dragMove={(e, d) => onHandleDrag({ x1: x1 + d.x, y1: y1 + d.y, x2, y2 })} />
```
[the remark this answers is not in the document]
eatyourgreens: These should be focusable too.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/components/draggable/draggable.stories.js
eatyourgreens: Or maybe it's a broken ref. I'm seeing that the `svg` ref is null.

eatyourgreens:
Tests are passing mostly because a lot of this new behaviour isn't tested.

eatyourgreens:
Re. I Fancy Cats, I could also generate interesting errors by clicking on the subject during a question task. I've fixed it by only rendering the drawing canvas during drawing tasks but we're still missing a component that takes any drawing annotations from `classification.annotations` and renders them as a static SVG layer.

We're also not adding marks from the canvas to the current task annotation, yet.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingMarks/DrawingMarks.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import { observer } from 'mobx-react'
+import DrawingToolRoot from '@plugins/tasks/DrawingTask/components/tools/DrawingToolRoot'
+import DeleteButton from '@plugins/tasks/DrawingTask/components/tools/DeleteButton'
+
+function DrawingMarks ({ activeMark, onDelete, onDeselectMark, onPointerDown, onPointerUp, onSelectMark, svg, tool }) {
```
eatyourgreens: I don't think `onPointerDown` and `onPointerUp` are used any more.

eatyourgreens: `svg` is being used as a prop in so many places that it would make sense to move it into context.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/components/draggable/draggable.js
```diff
@@ -0,0 +1,110 @@
...
+      const { setPointerCapture } = this.wrappedComponent.current
+      const { x, y, pointerId } = this.convertEvent(event)
+      this.setState({ coords: { x, y }, dragging: true })
+      this.props.dragStart({ x, y, pointerId })
+      setPointerCapture && this.wrappedComponent.current.setPointerCapture(pointerId)
+    }
+
+    dragMove (event) {
+      const { coords, dragging } = this.state
+      if (dragging) {
```
eatyourgreens: Check the `pointerId` somewhere here and don't drag if it doesn't match the pointer that started the drag.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/components/draggable/draggable.js
```diff
@@ -0,0 +1,110 @@
...
+      releasePointerCapture && this.wrappedComponent.current.releasePointerCapture(pointerId)
+    }
+
+    render () {
+      const { children, dragStart, dragMove, dragEnd, ...rest } = this.props
+      return (
+        <g
+          onPointerDown={this.dragStart}
+          onPointerMove={this.dragMove}
+          onPointerUp={this.dragEnd}
```
eatyourgreens: Add `onMouseDown|Move|Up` for older browsers (mostly Safari.)

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -28,36 +28,35 @@ function InteractionLayer ({ activeDrawingTask, svg }) {
...
+      const { activeTool } = activeDrawingTask
+      activeTool.deleteMark(activeMark)
     }
-    setActiveMark(null)
   }
 
   return (
     <g
-      onPointerDown={onPointerDown}
+      onPointerDown={createMark}
```
srallen: Having descriptive names of what the handler is doing is fine with me, but I'd like some consistency. If we're using more descriptive names for the handlers, then I'd like `onPointerMove` and `onPointerUp` renamed as well, perhaps to `dragMark` and `validateMark` perhaps?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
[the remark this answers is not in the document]
srallen: I think the original idea for `InteractionLayer` is that it would be capturing all events of interactivity, not just for drawing, but if this is deviating from that idea, I'd be ok with a name change. We'll have to note that in an update to the ADR.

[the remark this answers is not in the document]
srallen: Will the tests be updated in a follow up PR?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingMarks/DrawingMarks.js
srallen: Yep, `onPointerDown` and `onPointerUp` aren't being used.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingMarks/DrawingMarks.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import { observer } from 'mobx-react'
```
srallen: This isn't being used.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingMarks/DrawingMarks.js
srallen: This component needs tests.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/components/draggable/draggable.stories.js
srallen: I see `Uncaught TypeError: Cannot read property 'createSVGPoint' of null` thrown in the console now.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
```diff
@@ -1,56 +1,100 @@
...
+    isActive,
+    label,
+    mark,
+    onDelete,
+    onDeselect,
+    onSelect,
+    svg,
+    tool
+  }, ref) => {
+  console.log('rendering', label)
```
srallen: Leftover console log?

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
```diff
@@ -1,56 +1,100 @@
...
-  function onPointerMove (event) {
-    console.log(`${tool.type} moved`)
-    event.stopPropagation()
-    event.preventDefault()
+  function select (event) {
+    if (!active) {
+      setActive(true)
+      onSelect(mark)
+      console.log('selecting', label)
+      console.log(document.activeElement)
```
srallen: Leftover console logs from testing?

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
srallen: Checking in with @beckyrother again on the delete button behavior (see my comment above).

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
```diff
@@ -1,56 +1,100 @@
...
-const STROKE_WIDTH = 1.5
-const SELECTED_STROKE_WIDTH = 2.5
+const STROKE_WIDTH = 2
+const SELECTED_STROKE_WIDTH = 3
 
-function DrawingToolRoot ({ children, tool }) {
-  const [ active, setActive ] = useState(true)
+const StyledGroup = styled('g')`
+  :focus {
+    outline: 1px solid black;
```
srallen: The focus color used in the theme is `#addde0`. How does it look using this? Perhaps you can use the `withTheme` decorator instead to get the theme colors?

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/components/draggable/draggable.stories.js
srallen: Maybe remove the story for now and add it later once the API is finalized?

srallen:
I've noted the cursor styles in https://github.com/zooniverse/front-end-monorepo/issues/1052#issuecomment-558782416 for future implementation.

eatyourgreens:
>The pre-existing mark that is not focused has to be selected to focus it first, then you can start dragging them, whereas on PFE you can immediately start dragging. It's unclear if this change of behavior is intended.

This is a bug and I'm baffled by it. I see a bunch of re-renders in the console logs on mark select. Could those be causing the problem?

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
eatyourgreens: This is a debugging line.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
eatyourgreens: Debugging why dragging doesn't start when you select a mark. The commit is marked WIP.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingMarks/DrawingMarks.js
eatyourgreens: Yep.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
eatyourgreens: I think the original names for the mark creation methods in PFE were `initStart`, `initDrag` and `initEnd`. We could use those?

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/components/draggable/draggable.stories.js
eatyourgreens: >I see Uncaught TypeError: Cannot read property 'createSVGPoint' of null thrown in the console now.

Yes, the SVG isn't an `SVGElement` so doesn't implement the SVG interfaces, including `createSVGPoint`. Wrong element namespace (HTML instead of SVG) maybe?

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/tools/DrawingToolRoot.js
eatyourgreens: Also debugging the mark selection bug.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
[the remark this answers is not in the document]
eatyourgreens: I _think_ my comment was written before I wrote these new tests. I'm not sure. I guess finishing mark creation (pointer up) is still missing tests?

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/components/draggable/draggable.stories.js
eatyourgreens: Oh ignore that. `svg` is null so the ref is being created by the story.

eatyourgreens:
I've removed the WIP debugging commit from this branch.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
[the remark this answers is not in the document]
eatyourgreens: >I think the original idea for InteractionLayer is that it would be capturing all events of interactivity, not just for drawing, but if this is deviating from that idea, I'd be ok with a name change. We'll have to note that in an update to the ADR.

In PFE, each task carries a reference to the React component that should be used to render annotations for that task eg. [`DrawingTask.SVGRenderer`](https://github.com/zooniverse/Panoptes-Front-End/blob/3d842ffabc009857ed86b720382439455c025fb8/app/classifier/tasks/drawing/index.cjsx#L13)

We can copy that idea across to the task models, so that the interaction layer is available outside of the drawing task.

That said, this should be spec'ed somewhere. I've been working to the idea that this component is responding specifically to drawing task events. If that's not correct, we should catch it while it can still be changed.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingMarks/DrawingMarks.js
eatyourgreens: So do the drawing marks.

eatyourgreens:
That mark selection bug is fixed.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/markings/Mark.js
```diff
@@ -5,5 +5,10 @@ const Mark = types.model('Mark', {
   frame: types.optional(types.number, 0),
   toolIndex: types.optional(types.number, 0)
 })
+  .views(self => ({
+    get isValid () {
+      return true
```
rogerhutchings: Should a mark be valid by default? If you forget to write a new `isValid` method when writing a new tool, you could create invalid marks without meaning to.

rogerhutchings: If this does change, it's probably worth adding a console message to say the default `isValid` method has kicked in

eatyourgreens: You'd notice when testing, I'd hope, because valid marks won't be deleted from the drawing canvas.

eatyourgreens: Points are an example of a type of mark that doesn't have a validation.

eatyourgreens: `isValid` is optional, not required.

rogerhutchings: If you forget to override `isValid` when making a new tool, invalid marks won't be deleted either, right? Might be an edge case, and you'd hope it'd get caught in review, but it might not.

Also, for marks that end up out of bounds, `isValid` might be a sensible way to handle them. You could have an `isOutOfBounds` view that checks whether any of the coords constituting a mark is outside of the drawing area, and `isValid` calls that as part of its check.

eatyourgreens: >If you forget to override isValid when making a new tool, invalid marks won't be deleted either, right? Might be an edge case, and you'd hope it'd get caught in review, but it might not.

Exactly, and you'll notice that when they don't vanish from the canvas in testing.

Out-of-bounds is out of scope for this PR, but these models don't have a concept of a drawing area. See also the calculation of the delete button position. Valid position on the drawing canvas should probably be calculated in the view, which does have the canvas bounds.

eatyourgreens: Anyway, if you look through the original code in PFE, `isValid` is optional for marks.

rogerhutchings: 👍

rogerhutchings:
Not sure if this has been mentioned already (I have a vague memory of it somewhere, but just in case) - there are no cursor changes. e.g. `crosshair` over the marking area, `grab`/`grabbing` while dragging etc

eatyourgreens:
There's an issue for styling but I can't remember the number. It's lost somewhere back in the comments on this PR.

eatyourgreens:
I think the `grabbing` cursor is the only one I can't figure out because I'm not sure how to change styles with `useState` and that style doesn't apply to all `draggable` objects, only some.

rogerhutchings:
Nice one. I've just tested in the browser again, and noticed a few things:

- When creating a line and drawing the second point, the cursor flickers rapidly between `pointer` and `move`.
- If you move the cursor back along the line you've just created, the line isn't shortened. It needs a steady hand to do for the line, but I can imagine this happening for a rectangle a lot more easily
- With the point, clicking to create and holding down to drag is quite jerky. Once it's created, dragging is a lot smoother

rogerhutchings:
> I'm not sure how to change styles with useState

You could pass `dragging` / whatever state value into a styled element as a prop, and then you can do something like `${props => props.dragging && 'cursor: grab;'}`

eatyourgreens:
I can shorten a line by dragging back along the line during mark creation (Firefox 70.)

I think the cursor will depend on where the pointer is. Hover styles won't apply unless it's exactly over the element but `pointerMove` and `pointerUp` are handled even if the pointer moves out of the initial element. See the lines in the `draggable` code that capture and release the pointer.

rogerhutchings:
Also firefox 70, not letting go from initial click:

line

eatyourgreens:
Weird, I can't reproduce that at all.

rogerhutchings:
It does shorten if the cursor goes outside the mark, but as soon as it's back on it stops again.

eatyourgreens:
Event being swallowed by the line?

rogerhutchings:
That's what I'm thinking, yeah

eatyourgreens:
Fixed!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingToolMarks/DrawingToolMarks.js
```diff
@@ -0,0 +1,81 @@
...
+  onDeselectMark: PropTypes.func,
+  onSelectMark: PropTypes.func,
+  svg: PropTypes.instanceOf(Element).isRequired,
+  tool: PropTypes.object.isRequired
+}
+
+DrawingToolMarks.defaultProps = {
+  activeMarkId: '',
+  onDelete: () => true,
+  onDeselectMark: (mark) => true,
```
rogerhutchings: No need for `mark`

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingToolMarks/DrawingToolMarks.js
```diff
@@ -0,0 +1,81 @@
...
+  onSelectMark: (mark) => true
```
rogerhutchings: No need for `mark`

rogerhutchings:
Woops, should have done that as one review. Anyway, those are non-blocking, and I'm happy with this. There's still @srallen's requested changes, and there's a note to yourself about supporting `onKeyDown` / `onKeyUp` for older browsers

eatyourgreens:
I think the requested changes are mostly done. Just the mouse fallback when pointers aren’t supported, and a bunch of new code that isn’t tested.

eatyourgreens:
I'm inclined to merge this and handle any outstanding issues in smaller issues/PRs. Those should be easier to track and review.

eatyourgreens:
All the requested changes from the last review have been made (remove invalid marks on creation, add cursor styles and fix the selection bug.) I'm going to squash and merge.

PR #1332, opened 2019-11-28
Drawing: Add mouse event handlers

eatyourgreens:
Add mouse handlers as fallbacks when pointer events aren't supported. Map each mouse event handler to the equivalent pointer event.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
@@ -8,7 +8,7 @@ const StyledRect = styled('rect')`
   cursor: ${props => props.disabled ? 'not-allowed' : 'crosshair'};
 `
 
-function InteractionLayer ({ activeDrawingTask, activeTool, disabled, height, scale, svg, width }) {
+function InteractionLayer ({ activeDrawingTask, activeTool, disabled, height, scale, svg, usePointer, width }) {
   const [ activeMark, setActiveMark ] = useState(null)
   const [ creating, setCreating ] = useState(false)
 
@@ -60,17 +60,30 @@ function InteractionLayer ({ activeDrawingTask, activeTool, disabled, height, sc
     }
   }
 
+  const pointerHandlers = {
+    onPointerMove,
+    onPointerUp
+  }
+
+  const mouseHandlers = {
+    onMouseMove: onPointerMove,
+    onMouseUp: onPointerUp
+  }
+
+  const rootProps = usePointer ? pointerHandlers : mouseHandlers
+
+  const rectProps = usePointer ? { onPointerDown } : { onMouseDown: onPointerDown }
+
   return (
     <g
-      onPointerMove={onPointerMove}
-      onPointerUp={onPointerUp}
+      {...rootProps}
     >
       <StyledRect
         disabled={disabled}
         width={width}
         height={height}
         fill='transparent'
-        onPointerDown={onPointerDown}
+        {...rectProps}
       />
       {activeDrawingTask &&
         activeDrawingTask.tools.map(tool => {
@@ -98,12 +111,14 @@ InteractionLayer.propTypes = {
   disabled: PropTypes.bool,
   scale: PropTypes.number,
   svg: PropTypes.instanceOf(Element).isRequired,
+  usePointer: PropTypes.bool,
   width: PropTypes.number.isRequired
 }
 
 InteractionLayer.defaultProps = {
   disabled: false,
-  scale: 1
+  scale: 1,
+  usePointer: !!window.PointerEvent
 }
 
 export default InteractionLayer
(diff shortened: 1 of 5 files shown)


eatyourgreens:
Mouse events are missing an equivalent for `setPointerCapture` and `releasePointerCapture`. Other than that, this seems to work in Safari.

shaunanoordin:
Code changes look good and simple 👍 But give me one second to test for that "one tap = two events" bug that we encountered in [PFE PR 5411](https://github.com/zooniverse/Panoptes-Front-End/pull/5411#issuecomment-503624654) - I'm 90% sure it won't be a problem here because we're only accounting for _Pointer, and Mouse Events,_ and not _Pointer, Mouse, and Touch Events._ (IIRC it was Pointer + Touch that caused issues.)

shaunanoordin:
😢 Blargh! My tests aren't going very well

## Quick Review

This PR adds fallbacks for Mouse Events for the DrawingTools, i.e. for when Pointer Events aren't available. Unfortunately, I'm encountering some issues that makes the Mouse fallback more of a problem than a safety net.

Tested on `http://localhost:8080/?env=staging&project=1844` (note: preferring the Japanese ghost image because it didn't make the marks tiny) on macOS 10 + Chrome 78, using a 🐭 mouse:

Screen Shot 2019-11-28 at 17 34 09

- Clicking on the Subject to add a drawing Mark creates **two** Marks instead of one.
- In the image above, all I did was click on the Subject image, then moved my cursor away.
- Observe the events:
  - `InteractionLayer.onPointerDown()` triggers twice, indicating that the mouse click triggered both a Pointer Event and a Mouse Event.
  - `InteractionLayer.onPointerUp()` and `draggable.dragEnd()` each trigger twice, indicating the same.

(Ignore the "Drag Move" console logs; there was some confusion in my part where I didn't realise `draggable.dragMove()` listens for _all pointer/mouse movement_ on the draggable object, and it's the inner `if (dragging && event.pointerId === pointerId)` condition that checks if the draggable object is in fact being dragged.)

(Very minor nitpick: why is it `draggable.dragStart()` instead of `draggable.onDragStart()` ?)

### Status

It's a bit of a breaking issue, I'm afraid - I'm pretty sure there's a very simple solution to this, like setting a CSS or something. (If the solution IS that simple yet obscure, we should defo document it.)

I'm going to re-examine the solution implemented in PFE PR 5411.

ASIDE: I've also tested 👆 touchscreen functionality - with an iPhone X emulated on Chrome - and noted that ol' _"I can't perform click-drag actions, so I need CSS cursor: pinch-zoom or something"_ bug, but **that is a separate issue from this PR,** as far as I'm concerned.

rogerhutchings:
Also - @srallen suggested using a polyfill (such as https://github.com/jquery/PEP) instead of coding around the lack of support: https://zooniverse.slack.com/archives/C06LHUC6R/p1574361136129000?thread_ts=1574355865.114400&cid=C06LHUC6R

shaunanoordin:
I was going to ask _"but do we really need a Mouse Event fallback?"_ but then I realised who it's for: Safari. 😢 Boo, hiss!

srallen:
My preference would be to try out the polyfill option first and if it works, use that since we'll only need to support a fallback for another ~10 months. Safari major versions are on an annual release schedule. Once the next major version of Safari is released, we can remove the polyfill.

eatyourgreens:
I've updated this to use mouse events or pointer events but not both.

PEP sounds useful but I'm wary of overriding mouse events globally when we still have code that uses `onMouseDown` etc.

eatyourgreens:
This branch works for me in Safari 12.1.2 and also latest Chrome and Firefox.

eatyourgreens:
> why is it draggable.dragStart() instead of draggable.onDragStart() ?

[`onDragStart`](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/ondragstart) is already taken (and not supported on mobile.)

srallen:
> PEP sounds useful but I'm wary of overriding mouse events globally when we still have code that uses onMouseDown etc.

I don't see in the docs or code where PEP is doing this globally. It only fires in browsers that don't support the pointer events API and only listens on elements with the `touch-action` attribute added. Could you point out where you're reading this?

eatyourgreens:
@srallen Apologies. My understanding of PEP was that it globally hijacks mouse events and replaces them with pointer events, in browsers that don't natively support pointer events.

We weren't able to get `touch-action: none` working in PFE (see https://github.com/zooniverse/Panoptes-Front-End/pull/5411) but please do feel free to open a PR adding PEP and `touch-action` to the classifier.

In the meantime, this PR should get the drawing components working in Safari 12, without side effects in other components that listen to mouse events.

srallen:
I've opened #1368 using PEP for comparison.

eatyourgreens:
Superseded by #1368 which I'm hoping also polyfills pointer capture, which this PR does not. Without pointer capture, I don't think you'll be able to draw on an iPad.

eatyourgreens:
I'll hold off on deleting this branch until #1368 is merged and tested on staging.

PR #1339, opened 2019-12-02
Drawing: Copy drawn marks to drawing annotations

eatyourgreens:
- When a drawing task is complete, move its marks from the drawing tools to the task annotation.
- When a drawing task starts, copy any existing marks from the task annotation to the task's drawing tools.
- Pass previous drawing annotations into the interaction layer component and render their values as inactive marks.

Adds `task.start()` and `task.complete()` to the task model, which run when a task starts and finishes respectively.

TODO:
- [x] show marks from previous tasks (#1104.)
- [x] submit all drawn marks with the completed classification.
- [x] clear the drawing tasks on subject change.

Closes #1104.

Test URL: http://localhost:8080/?project=908&workflow=3370

Package:
lib-classifier

Closes #1103.

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

Issue #1334, opened 2019-11-29
Drawing: add canvas bounds


eatyourgreens:
**Package**
lib-classifier

**Is your feature request related to a problem? Please describe.**
The boundaries of the drawing canvas are used in a few places in the current drawing code.
  - marks are [deleted if outside the canvas](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/drawing-tools/delete-if-out-of-bounds.coffee).
 - the mark delete button position is moved [if out-of-bounds](https://github.com/zooniverse/Panoptes-Front-End/blob/3d842ffabc009857ed86b720382439455c025fb8/app/classifier/drawing-tools/line.cjsx#L47-L50).

**Describe the solution you'd like**
There's a couple of different ways this could be approached.
- Add the canvas boundaries to the Mark model.
- Check coordinates against the canvas boundaries in the view components before rendering marks.

**Describe alternatives you've considered**

**Additional context**

PR #1340, opened 2019-12-02
Drawing: scale marks

eatyourgreens:
Calculate the subject image scale when the subject image loads. Pass the scale down as a prop to marks. Scale rendered mark sizes and delete button position.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
@@ -9,9 +9,11 @@ const StyledRect = styled('rect')`
   pointer-events: ${props => props.disabled ? 'none' : 'all'};
 `
 
-function InteractionLayer ({ activeDrawingTask, activeTool, disabled, svg }) {
+function InteractionLayer ({ activeDrawingTask, activeTool, disabled, height, svg, width }) {
   const [ activeMark, setActiveMark ] = useState(null)
   const [ creating, setCreating ] = useState(false)
+  const [ scale, setScale ] = useState(1)
+  const interactionLayer = React.createRef()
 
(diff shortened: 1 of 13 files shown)


eatyourgreens:
Now that #1339 has merged, this should also scale any previously drawn marks.

shaunanoordin:
> Now that #1339 has merged, this should also scale any previously drawn marks.

I'm not seeing this yet, unfortunately. Aren't the previous marks rendered in `<InteractionLayerContainer />`, whereas the `scale` is calculated only in (and for) `<InteractionLayer>` ?

Tested on http://localhost:8080/?env=staging&project=1895 on commit a6c82d2efe55083636db7305f840a4e9491ad42c:

Screen Shot 2019-12-03 at 15 48 10

_On Task 1, I draw a few Points which are scaled up._

Screen Shot 2019-12-03 at 15 48 20
_When moving to Task 2, the Points drawn in Task 1 use the default scale of 1.0_

shaunanoordin:
Thought: should the `scale` be stored as `SubjectViewer.subjectScale`, so it'll be visible to the parent Container? Putting it in the SubjectViewer might also be handy if we need to refer to the zoom values, e.g. if we want to maintain a consistent mark size independent of zoom levels.

eatyourgreens:
>I'm not seeing this yet, unfortunately. Aren't the previous marks rendered in <InteractionLayerContainer />, whereas the scale is calculated only in (and for) <InteractionLayer> ?

That's because I hadn't written the code yet. I've just committed changes which move the scale calculation up to the parent subject viewer.

eatyourgreens:
An SVG context might be useful to store the common props that are being passed around to every SVG component. #1334 is going to need client bounding rectangles for each drawn mark, as well as the subject image, so that would be a good place to think about that.

PR #1357, opened 2019-12-05
Drawing: check for out-of-bounds marks

### LINKED ISSUE #1334: Drawing: add canvas bounds

eatyourgreens:
Delete marks when they are completely outside the SVG drawing canvas by comparing the bounding client rectangles of the mark element and the parent SVG element. Based on PFE's [`isInBounds`](https://github.com/zooniverse/Panoptes-Front-End/blob/abab2b0115231dc34fd824f70f46a0c372b3043e/app/lib/is-in-bounds.coffee#L1-L8) function.

Package:
lib-classifier

Closes #1334.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
@@ -15,18 +15,33 @@ describe('Component > InteractionLayer', function () {
     initialPosition: sinon.stub(),
     setCoordinates: sinon.stub()
   }
-  const mockSVGEvent = {
+  const mockSVGPoint = {
+    x: 100,
+    y: 200,
     matrixTransform: sinon.stub().callsFake(() => ({
       x: 100,
       y: 200
     }))
   }
-  const mockSVG = {
-    createSVGPoint: sinon.stub().callsFake(() => mockSVGEvent),
-    getScreenCTM: sinon.stub().callsFake(() => ({
-      inverse: sinon.stub()
-    }))
+  const mockScreenCTM = {
+    a: 1,
+    b: 1,
+    c: 1,
+    d: 1,
+    e: 1,
+    f: 1,
+    inverse: () => ({
+      a: 1,
+      b: 1,
+      c: 1,
+      d: 1,
+      e: 1,
+      f: 1
+    })
   }
+  const svg = document.createElementNS("http://www.w3.org/2000/svg","svg")
+  svg.createSVGPoint = () => mockSVGPoint
+  svg.getScreenCTM = () => mockScreenCTM
 
   beforeEach(function () {
     const mockDrawingTask = DrawingTask.TaskModel.create({
@@ -55,7 +70,7 @@ describe('Component > InteractionLayer', function () {
         activeDrawingTask={mockDrawingTask}
         activeTool={activeTool}
         height={400}
-        svg={mockSVG}
+        svg={svg}
         width={600}
       />)
   })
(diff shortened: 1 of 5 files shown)


eatyourgreens:
>The "is it out of bounds?" check only runs after the Mark deselected. The check doesn't run again if, e.g. the visible area of the SVG changes due to the window size changing, and the previously in-bounds Mark becomes out-of-bounds.

I don't think this can happen. Mark coordinates should be independent of viewport size. If they aren't, then classification is broken.

We should double-check whether PFE checks against the `image` element's bounding rectangle, or against the parent SVG element, since those two rectangles are different.

eatyourgreens:
WIP because I wasn’t sure about the code but if it works then maybe it’s worth merging?

eatyourgreens:
TODO: check what happens if you create a point then drag it out of the SVG during `initDrag`. I suspect it won’t be deleted with this code since a mark doesn’t become draggable until after it’s created.

eatyourgreens:
Answering my own question: you can't drag a mark off the canvas during `initDrag` because the area outside the interaction layer doesn't listen to pointer events. The new mark stops exactly at the edge of the image.

eatyourgreens:
Yeah, the bounds are set by the SVG element because we don’t have a ref to the image. I’d like to fix that too, but haven’t figured out how yet.

srallen:
Oh just to be clear, my only requested change is just for specs for this.

eatyourgreens:
If I can refactor this to use event targets instead of refs for the marks then I think that will make it easy to test. Otherwise, I’m going to have to figure out how to work with refs in functional components in enzyme.

eatyourgreens:
Initial specs for the `DrawingToolMarks` are in #1372, by the way, where I changed it to pass in an array of marks to render.

eatyourgreens:
Testing the bounds check has run into enzyme (or JSDom's) problem rendering SVG elements properly. I think this same error came up on another PR. `TypeError: svg.createSVGPoint is not a function`.

I think the `svg` prop is being rendered in the tests as HTML, not SVG.

eatyourgreens:
@srallen I've added tests. I switched the component from using refs to using event targets, so that the tests can pass a mock event target when dragging ends.

eatyourgreens:
Hmm, you can drag a line out of the viewer if you use its drag handles. That will miss this check, which runs when you drag a mark as a whole.

eatyourgreens:
If the calculation was done, per shape, in SVG coordinates rather than screen pixels then it would just be a check of SVG `{ x, y }` positions against the image rectangle defined by `{ 0, 0, naturalWidth, naturalHeight }`.

That's not a difficult calculation, but would vary from one shape to another. It could be implemented by writing a method for each shape, something like `mark.overlaps(rectangle)`, which loops through the points for that model and checks if any lie inside `rectangle`.

The current implementation is the one used in PFE but it's kind of crude because it compares the client rectangles of two DOM elements to check if they overlap. For example, the corner of the bounding rectangle of a diagonal line can overlap the SVG element even if the line points themselves do not lie anywhere inside that element.

Issue #500, opened 2019-02-22
[RFC] Drawing tools mark models


srallen:
**Package** lib-classifier

**Feature or Issue Description**
The MobX stores will have models defined for these expected drawing marks, however, we need to check in with @CKrawczyk since the current tools are inconsistent or not always in an optimal format for aggregation. We should discuss this and document in an ADR any changes for these tools. 

**To Dos**

- [ ] Example

CKrawczyk:
# Overview of the current drawing tools.

## Definitions

**RHC**: A right handed coordinate system, this is defined as a system where **positive** angles rotate an object from the **+x axis** to the **+y axis** with **`angle=0` along the +x axis**

**LHC**: A left handed coordinate system, this is defined as a system where **positive** angles rotate an object from the **+x axis** to the **-y axis** with **`angle=0` along the +x axis**

**Domain**: The range of values a number can take `[` or `]` is inclusive, `(` or `)` is exclusive.

**Upper origin**: The point `0, 0` is in the upper left of the plot

**Lower origin**: The point `0, 0` is in the lower left of the plot

## Notes about coordinate systems

The SVG coordinate system used by the the current front-end is a **RHC** and has an **upper origin**, this results in positive angles rotating clockwise.

Most plotting software (R, Python, Matlab) are **RHC** with a **lower origin**, this results in positive angles rotating counter-clockwise.

Both cases above have **RHC** but a positive angle rotates in opposite directions, this is why the visible direction of rotation should **not** be the only thing used to figure out what direction the angle should go.  The position of the origin matters too.

## The current drawing tools
For this list I am only concerned about tools with a specified shape (i.e. no free-hand tools or polygons)

### Circle
- `x`: Center `x` position
- `y`: Center `y` position
- `r`: Radius

### Column
- `x`: Left most line of the column
- `width`: Full width of the column

### Ellipse
- `x`: Center `x` position
- `y`: Center `y` position
- `rx`: Radius along the x-axis (before rotation)
- `ry`: Radius along the y-axis (before rotation)
- `angle`: Degrees, **LHC**, domain `(-180, 180]`

### fullWidthLine
- `y`: `y` position of the line

### fullHeightLine
- `x`: `x` position of the line

### line
- `x1`: Start point `x` position
- `y1`: Start point `y` position
- `x2`: End point `x` position
- `y2`: End point `y` position

### point
- `x`: `x` position of the point
- `y`: `y` position of the point

### rectangle
[...]

PR #1360, opened 2019-12-06
Drawing: add the ellipse tool

### LINKED ISSUE #1291: Ellipse drawing tool
> an oval of any size and axis ratio; this tool can be rotated.

There is also the AnchoredEllipse tool:

> Creates an ellipse in the center of the subject during the first click, and does not allow it to be dragged

It's currently unclear if anyone's using the anchored ellipse, so we can think about deprecating it. We could also replicate it by making it a configurable option on the main ellipse tool.

## Related code in PFE

- [ellipse.cjsx](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/drawing-tools/ellipse.cjsx)
- [anchored-ellipse.jsx](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/drawing-tools/anchored-ellipse.jsx)

## Related Issues

- #500

eatyourgreens:
- Add view components, tool model and mark model for the PFE ellipse drawing tool.
- Rotation angle is switched from negative to positive.

- Makes `(x, y, angle)` common, optional properties for all marks, defining the position of the mark and rotation about that position as per #500.

- Adds `mark.getDistance(x1, y1, x2, y2)` and `mark.getAngle(x1, y1, x2, y2)` helpers to calculate the distance and angle between two points.

- updates the `dragMove` prop for draggable components to return the `{x, y}` coordinates of the move event, as well as the difference in `x` and `y`.

An experiment to see how much work is involved in converting a drawing tool across from PFE. The original ellipse coffeescript was converted to JS with [decaffeinate](https://decaffeinate-project.org/repl/#?useCS2=true&useJSModules=true&loose=false&evaluate=true&stage=full&code=). Static methods were mapped to tool actions.

~~`getDeleteButtonPosition` was mapped straight across from the coffeescript, but seems to be working differently here.~~ Fixed by defining position and rotation in the drawing tool root.

Package:
lib-classifier

Closes #1291.  

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
Looking at [I Fancy Cats](https://master.pfe-preview.zooniverse.org/projects/brooke/i-fancy-cats/classify), the delete button for ellipses behaves slightly strangely there too so 🤷‍♂.

eatyourgreens:
If we want to add line direction to the transcription work, then this PR will need to be merged to add drawing support for angles.

eatyourgreens:
`toolIndex` was introduced in @mcbouslog's original work. I think because `tool` refers to a tool model? I don't remember either. It's trivial to change it.

The discussion about angles and coordinates is ongoing in #500. We should try to get some consensus there.

 #500 seems the natural place for discussion of the mark ID too, since it discusses what goes into a submitted annotation.

`mark.annotations` is a place holder for subtask annotations, until subtasks are built. Again, we're discussing in #500 how that gets represented in the final submitted classification.

I think you're right that we need a decision on the questions in #500 too. I don't think Caesar is particularly fussy about what we call things, only that we name them consistently. Here, I'm enforcing that consistency by putting those properties into the base Mark model.

srallen:
I've opened #1390 as the proposed ADR from #500 RFC. Let's wait for review at least from @CKrawczyk before updating the associated issue for the ellipse tool to review against for some of these open questions.

eatyourgreens:
There's a couple of changes in here that will be needed for other tools: the changes to `dragMove` and the addition of pointer capture when creating a mark. They can be cherry-picked out if necessary.

srallen:
One of the proposed items in #1390 is to remove default values because they create bias. Noting here that this impacts the ellipse because we have a default value for a minimum radius. Changing this can wait until the changes in #1390 are agreed upon.

eatyourgreens:
There's a minimum radius in the mark validation but the default that biasses classifications is the ratio of the minor to major axes, `DEFAULT_SQUASH`, which is set to 0.5.

srallen:
@CKrawczyk can you confirm that the default radius size does not have a biasing effect on the ellipse?

CKrawczyk:
Do you mean the minimum radius?  If so, that does not bias the results.  If a research team is interested in features that small they should adjust the scaling on the uploaded images rather than expect volunteers to draw small shapes.

The default axis ratio is the only bias we have seen with the ellipse tool.

eatyourgreens:
I've removed DEFAULT_SQUASH completely, and set `ry` to a tiny value so that the initial shape is essentially a line. Let's see how that goes. If we want configurable values for the initial y radius then that can go into a separate PR since it involves changing the tool model and PFE.

eatyourgreens:
Changing `x` and `y` to `x_center` and `y_center` throws a bunch of MST errors, so needs a bit of thought. `mark.x` and `mark.y` are writable values, so can't simply be change to a read-only computed value.

We could change `x` and `y` to `x_center` and `y_center` throughout the code eg. where a mark's coordinates are updated after dragging. This will involve duplicating some code across models just to change the variable names, because points use `{ x, y }` but ellipses use `{ x_center, y_center }`. That's a clumsy approach. It may introduce bugs later down the line and I'd prefer to avoid it. If we are pushing for that approach, then the consequences should be spelled out in the model ADR.

That latter option is a big enough change to the drawing code that it should maybe go into a separate PR, so it can be reviewed in isolation.

A more minor point is that we lint for snake case, preferring camel case in the JS code so the linter will ask for `xCenter` and `yCenter`.

eatyourgreens:
Here, I've gone for the option of duplicating code across models with different variable names so, for example `mark.initialPosition({ x, y })` sets `mark.x` and `mark.y` for a Point but `mark.x_center` and `mark.y_center` for an Ellipse.

eatyourgreens:
The mark models still need tests too.

srallen:
@CKrawczyk Thanks for the confirmation. I'm trying to be perhaps overly thorough in making sure we capture all the changes needed.

@eatyourgreens good to know what the practical application is of the suggested changes are. I'll have a look and think about if we still need to revise.

eatyourgreens:
I’ll put up a quick PR tomorrow that changes the Point model to use `x_center` and `y_center` so that we can see how it looks.

packages/lib-classifier/src/plugins/drawingTools/components/Ellipse/Ellipse.js
```diff
@@ -0,0 +1,51 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import DragHandle from '../DragHandle'
+
+const GUIDE_DASH = [4, 4]
+const GUIDE_WIDTH = 1
+
+function Ellipse ({ active, children, mark, onFinish, scale, svg, tool }) {
```
srallen: Should this be using the new SVG Context?

srallen: Also `tool` isn't being used and can be removed.

packages/lib-classifier/src/plugins/drawingTools/components/Ellipse/Ellipse.js
```diff
@@ -0,0 +1,51 @@
...
+          <DragHandle dragMove={onYHandleDrag} x={0} y={-1 * ry} scale={scale} svg={svg} />
+        </g>
+      }
+    </g>
+  )
+}
+
+Ellipse.propTypes = {
+  active: PropTypes.bool,
+  tool: PropTypes.object
```
srallen: There are several props missing from the prop types here.

packages/lib-classifier/src/plugins/drawingTools/components/Point/Point.js
```diff
@@ -22,7 +22,7 @@ function Point ({ active, children, mark, scale }) {
   const radius = RADIUS[size] / scale
 
   return (
-    <g transform={`translate(${mark.x}, ${mark.y})`}>
```
srallen: Why is this removed?

packages/lib-classifier/src/plugins/drawingTools/models/marks/Point/Point.js
```diff
@@ -23,8 +23,8 @@ const PointModel = types
       const theta = (DELETE_BUTTON_ANGLE) * (Math.PI / 180)
       const dx = (20 / scale) * Math.cos(theta)
       const dy = -1 * (20 / scale) * Math.sin(theta)
-      const x = self.x + dx
-      const y = self.y + dy
+      const x = dx
+      const y = dy
```
srallen: Should there be changes to Point in a PR for Ellipse?

packages/lib-classifier/src/plugins/drawingTools/components/DrawingToolRoot/DrawingToolRoot.js
```diff
@@ -55,6 +56,10 @@ const DrawingToolRoot = forwardRef(function DrawingToolRoot({
     onDeselect(mark)
   }
 
+  let transform = ''
+  transform = (mark.x && mark.y) ? `${transform} translate(${mark.x}, ${mark.y})` : transform
+  transform = mark.angle ? `${transform} rotate(${mark.angle})` : transform
```
srallen: There should be a spec for this

packages/lib-classifier/src/plugins/drawingTools/components/Ellipse/Ellipse.js
srallen: `tool` isn't being used here. `svg` should switch to use context.

packages/lib-classifier/src/plugins/drawingTools/components/Ellipse/Ellipse.js
srallen: There are a few props missing from the type definitions here. `tool` can be removed.

packages/lib-classifier/src/plugins/drawingTools/components/Point/Point.js
```diff
@@ -22,7 +22,7 @@ function Point ({ active, children, mark, scale }) {
...
+    <g>
```
srallen: Why was this removed?

packages/lib-classifier/src/plugins/drawingTools/models/marks/Ellipse/Ellipse.js
```diff
@@ -0,0 +1,95 @@
...
+const MINIMUM_RADIUS = 5
+
+const EllipseModel = types
+  .model('EllipseModel', {
+    x_center: types.maybe(types.number),
+    y_center: types.maybe(types.number),
+    rx: types.optional(types.number, 0),
+    ry: types.optional(types.number, 0),
+    angle: types.optional(types.number, 0)
+  })
```
srallen: Let's add `toolType: types.literal('ellipse')` here

packages/lib-classifier/src/plugins/drawingTools/models/marks/Ellipse/Ellipse.js
```diff
@@ -0,0 +1,95 @@
...
+      return self.y_center
+    }
+  }))
+  .actions(self => {
+
+    function initialDrag ({ x, y }) {
+      const rx = self.getDistance(self.x, self.y, x, y)
+      const angle = self.getAngle(self.x, self.y, x, y)
+      self.rx = rx
+      self.ry = rx * 0.0001
```
srallen: Maybe a TODO comment here about switching this to configurable value from the lab?

packages/lib-classifier/src/plugins/drawingTools/components/Ellipse/Ellipse.js
eatyourgreens: I think `svg` can go completely.

packages/lib-classifier/src/plugins/drawingTools/components/Point/Point.js
eatyourgreens: It's been a while since I wrote this but it should have moved to the drawing tool root.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Point/Point.js
eatyourgreens: Yep.

packages/lib-classifier/src/plugins/drawingTools/components/DrawingToolRoot/DrawingToolRoot.js
eatyourgreens: Good point.

packages/lib-classifier/src/plugins/drawingTools/components/Ellipse/Ellipse.js
eatyourgreens: `svg` should be deleted, I think.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Ellipse/Ellipse.js
eatyourgreens: Let's add it as a configurable value if volunteers or project owners ask for it.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Point/Point.js
eatyourgreens: Looking back through the commit history, this is also because transforms are handled by the drawing tool root. The delete button position is now relative to `0, 0`.

packages/lib-classifier/src/plugins/drawingTools/components/DrawingToolRoot/DrawingToolRoot.js
eatyourgreens: Turns out the drawing tool root doesn't have any real tests at all, yet.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Ellipse/Ellipse.js
srallen: This is not consistent with the decision in ADR 24. If we're going to do something different, i.e. really small values for defaults, then we should document that as the decision instead when, why, etc. I'd prefer we stay consistent with ADR 24 unless we have good reason to do something different.

eatyourgreens: It can be set to 0. The rendering is a bit weird for a zero-radius shape, but you can still draw ellipses.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Ellipse/Ellipse.js
eatyourgreens: Out of interest, why `toolType` and not `type`?

packages/lib-classifier/src/plugins/drawingTools/components/DrawingToolRoot/DrawingToolRoot.js
eatyourgreens: #1410 adds specs for `DrawingToolRoot`.

eatyourgreens: I'll cherry pick those onto this branch and add tests for `transform`

packages/lib-classifier/src/plugins/drawingTools/models/marks/Ellipse/Ellipse.js
eatyourgreens: This should be added to all marks in a separate PR, since it changes how snapshots are handled for all marks.

packages/lib-classifier/src/plugins/drawingTools/components/Point/Point.js
eatyourgreens: It wasn't. It's moved to `DrawingToolRoot`.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Ellipse/Ellipse.js
srallen: To clearly distinguish it from the task type. We've had enough trouble keeping straight that it was only the task models that had types defined on them and not the annotation and mark models, so the more specific naming, the better, imo.

srallen: A separate PR is ok. Can you open an issue to track it?

packages/lib-classifier/src/plugins/drawingTools/models/marks/Ellipse/Ellipse.js
srallen: Ok, I'd like a TODO comment here and an updated issue so that we know this is something to come back to. Defaulting to either a very small number or zero has some UX implications that I think Becky should comment on, but ultimately, the decision is to make this configurable, so there needs to be planned work at some point for updating the lab unless an argument is made to update the proposed ADR with an alternative proposal. I asked about really small defaults in the discussion, but the message is strongly none if possible and if they have to exist, make the project owners actively involved in setting them so they understand the effect.

eatyourgreens:
This PR has changes that affect all marks. I'm going to pull them out into their own PRs so that they don't block transcription.
- add pointer capture to newly drawn marks.
- add distance and angle calculation to all marks.
- add position and rotation to the mark root.

I'll leave them in this PR since the Ellipse tool will break without those changes.

eatyourgreens:
This isn't required for the IMLS deadline so is pretty low priority.

srallen:
@eatyourgreens since this has a few merge conflicts and is lower priority, is it ok to close this for now and we can circle back to it later? I do think this is close, but I'm not sure if it'd be easier to do a fresh branch or to attempt a rebase.

eatyourgreens:
The merge conflicts were minor so I've fixed them.

eatyourgreens:
Shaun's work on the Rectangle tool reminded me that this PR still has changes to `dragMove` that would be useful on `master`. I can pull those out into their own PR. I think it's a one-line change.

srallen:
That would be helpful, but not really a priority.

Issue #1306, opened 2019-11-14
Implement server-side authentication


rogerhutchings:
We need to handle auth on the server. Specifically, picking up user project preferences for determining workflow selection etc.

cc #365 #1132 

Time estimate: 6 weeks

rogerhutchings:
Okay, made some progress on this today with @camallen, to the point where I got a cookie passed along with the page request :tada:

## Steps to reproduce

- Spin up a local copy of the Panoptes API, set up a new user, and get a `client_id`.

- Update `config/initializers/session_store.rb b/config/initializers/session_store.rb` with the following:

    ```rb
    Rails.application.config.session_store :cookie_store,key: '_Local_Panoptes_session', domain: '.zooniverse.org'
    ```
  _(This step is optional, but helps you identify the cookie passed along with the request as being from your local copy of Panoptes.)_

- Update the `default` block in `config/cors_config.yml`:

    ```yaml
    default: &default
      headers: :any
      request_methods: [delete, get, post, options, put, head]
      expose: [ETag, X-CSRF-Param, X-CSRF-Token]
      max_age: 300
      allows:
        - {
            origins: !ruby/regexp '/^https?:\/\/(127\.0\.0\.1|localhost|[a-z0-9-]+\.local|10\.[0-9]+\.[0-9]+\.[0-9]+|192\.[0-9]+\.[0-9]+\.[0-9]+|[a-z0-9-]+\.zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$/',
            resource: "/users*",
            credentials: true,
          }
        - {
            origins: !ruby/regexp '/^https?:\/\/(127\.0\.0\.1|localhost|[a-z0-9-]+\.local|10\.[0-9]+\.[0-9]+\.[0-9]+|192\.[0-9]+\.[0-9]+\.[0-9]+|[a-z0-9-]+\.zooniverse\.org|[a-z0-9-]+\.(pfe-)?preview\.zooniverse\.org)(:\d+)?$/',
            resource: "/oauth/*",
            credentials: true,
          }
        - { origins: "*", resource: "*" }
    ```

- Start your local copy of Panoptes with `docker-compose up` (you will need to change the `docker-compose.yml` to change the host port)

- Start your local copy of `app-project`

[...]

Issue #302, opened 2018-11-26
Create ADR for server-side auth


rogerhutchings:
**Package**

lib-auth

**Feature or Issue Description**

We've created a sweet auth library that only works in the browser. Given that we're moving towards SSR, we should create a strategy for handling auth on the server as well.

While researching the issue, I came across [next-auth](https://github.com/iaincollins/next-auth/), which I managed to get working out of the box with OAuth2 with minimal pain. This also has the benefit of hiding the `accessToken` from the client.

However, to actually get authenticated requests up and running, there's additional wiring needed:

- It uses [Passport.js](http://www.passportjs.org/) under the hood, which is an express middleware. To make authenticated requests, we'd need to proxy the API to a route on the app, pass that route to Panoptes.js, and make requests through that.
- Passport encourages the use of a session database
- We should probably create custom strategies for Passport.js to simplify config.

This will probably require us to abstract things out slightly differently to the way we have it now:

- The user stores get moved into their own package, so they can be used server and client side.
- The interface components get moved into their own package.
- We write a small wrapper for stores and interface so they can be used easily in a client-side only app

This is probably more work than we can fit in the time available before we down tools on TESS, so we should create an ADR to document this next section of work.

srallen:
Instead of a separate lib for the views, they could be added to the existing react-components lib.

rogerhutchings:
We'll also need to get the User Project Preferences on the server as well, since it's used in workflow selection, which is done in the project app rather than the classifier: #1132

PR #1361, opened 2019-12-06
Add ADR 21: Auth Client

rogerhutchings:
Adds an ADR for discussion on how to build a new auth client that will work for the rebuild.

cc #302 
cc #1306


eatyourgreens:
`unsubscribeEmail` is used by the unsubscribe form but that's only relevant when/if we move the unsubscribe page over. https://github.com/zooniverse/Panoptes-Front-End/blob/21cf42485929a62938112d5e2d4bca1a6702b00e/app/pages/unsubscribe.jsx#L40-L47

docs/arch/adr-21.md
```diff
@@ -0,0 +1,35 @@
...
+- The only way to configure `panoptes-javascript-client` is _directly_ via env vars, or by passing in query parameters to the URL. We can't get env vars on the client, so that's out, and query parameters become unwieldy very quickly.
+
+This hasn't been an issue when working with Single Page Apps, since env vars are baked in at transpilation time by the build tools. At the moment, we just avoid the problem by having the staging build of the project use the production API.
+
+However, for the Next.js-based apps we're building right now, we need a different approach, and being able to configure the auth client from a single source of truth is required. Ideally, this would be source-agnostic: the client should be able to be configured from a config file, [env vars](https://12factor.net/config), or whatever you want, but that's up to the consuming app to decide.
+
+## Decision
+
+Rewrite the existing auth client as a separate package. The new client will follow a pattern that's well established in the wider ecosystem: the package exposes a function which accepts a config object as an argument and returns a configured client.
```
eatyourgreens: Just to clarify my understanding here: The factory takes a config object and returns a singleton, right, which can then go into context or a store?

rogerhutchings: That's the idea, with the (obvious) caveat that as it's non-serialisable it'll have to be recreated on the client side

docs/arch/adr-21.md
```diff
@@ -0,0 +1,35 @@
...
+
+The client is purely for authentication, and will expose the methods for registration, sign in, sign out, and getting bearer tokens.
+
+The [existing config](https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/config.js) will be turned into a separate package. Config settings can be imported wholesale from that for convenience.
+
+For use in the rebuild apps, we would create a `ConfigStore` which we populate with the relevant config settings. The `ConfigStore` snapshot is then used to rehydrate the client, so we get a single source of truth on both client and server.
```
eatyourgreens: If we handle auth on the server, do we need auth on the client too? Could we just pass the Authorisation header to the client so that it can add it to requests.

docs/arch/adr-21.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Status
+
+Proposed
+
+## Consequences
+
+- I'm not planning to add in query param support to the client. This can be added in to the consuming app if required.
```
eatyourgreens: 👍 from me. I don't think the client should be telling consuming apps how to write URLs.

docs/arch/adr-21.md
eatyourgreens: I'm wondering if we even need an auth client in the browser. We could proxy login/out and registration through server-side auth, maybe? Then all the browser needs to know is whether or not to add an Auth header to requests.

docs/arch/adr-21.md
rogerhutchings: We _can_, you can modify the response headers in Next pretty easily - @camallen, would there be any security implications to doing that?

eatyourgreens: Should be okay over HTTPS, I think.

rogerhutchings: That's what I assumed, but I wanted to check

docs/arch/adr-21.md
rogerhutchings: We'd need to handle token refreshing though.

docs/arch/adr-21.md
```diff
@@ -0,0 +1,35 @@
...
+- There are a few methods, such as `unsubscribeEmail`, which I don't think are being used. These can be added in later if required.
```
srallen: I would suggest that this line should say we should drop this from the auth client as out of scope. I think this is the responsibility of the app that contains this form.

docs/arch/adr-21.md
```diff
@@ -0,0 +1,35 @@
...
+
+However, for the Next.js-based apps we're building right now, we need a different approach, and being able to configure the auth client from a single source of truth is required. Ideally, this would be source-agnostic: the client should be able to be configured from a config file, [env vars](https://12factor.net/config), or whatever you want, but that's up to the consuming app to decide.
+
+## Decision
+
+Rewrite the existing auth client as a separate package. The new client will follow a pattern that's well established in the wider ecosystem: the package exposes a function which accepts a config object as an argument and returns a configured client.
+
+The client is purely for authentication, and will expose the methods for registration, sign in, sign out, and getting bearer tokens.
+
+The [existing config](https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/config.js) will be turned into a separate package. Config settings can be imported wholesale from that for convenience.
```
srallen: Sounds good to me.

eatyourgreens:
`unsubscribeEmail` might be in the auth client because it makes a credentialed request, so be careful about removing. 

Architecturally, I don’t think it’s any different than having specific client handlers for the login and registration forms. We could make those the responsibility of the client app too, but that might e painful to app developers who want to easily authenticate Zooniverse accounts.

docs/arch/adr-21.md
srallen: Just one note, the panoptes.js client has a config, that we could move into this separate package. The config file there was based on the one from PJC, so we should make sure this new config package works across the board.

Issue #1314, opened 2019-11-20
[RFC] Transcription task annotation JSON structure


srallen:
**Package** lib-classifier

The new transcription task is a composite of new variants of existing tasks: the line drawing tool of the drawing task and the text task with autosuggestions from caesar if they exist. In addition, ADR 20 (#1268) identified the potential need to track in the annotation directionality and page rotation to aid in aggregation. We need to decide how these should be annotated. 

ASM uses a modified polygon drawing tool capped at two points and a standard text task annotation in the sub-task. An example of this annotation JSON:

``` json
{
 "task": "T1",
  "value": [{
      "details": [
         { "value": "the text of the line being transcribed" }
      ],
      "frame": 0,
      "points": [
        { "x": 1, "y": 1 },
        { "x": 100, "y": 100 } 
      ]
  }]
}
```

Here's a proposal of what the new transcription task annotation structure could look like:

``` json
{
 "task": "T1",
  "value": [{
      "details": [
         { "value": "the text of the line being transcribed" }
      ],
      "frame": 0,
      "x1": 194.80206298828125,
      "y1": 356.3935241699219,
      "x2": 339.3945617675781,
      "y2": 356.3935241699219,
      "direction": 0,
      "page_rotation": 0
  }]
}
```

`direction`: a number in degrees to indicate writing direction, with the only possible values being 0 equal to left-to-right, 90 equal to top-to-bottom, 180 equal to right-to-left, and 270 equal to bottom-to-top. I think we should have a consistent definition of what each of these mean, so a document that is written in RTL would likely have most of its transcribed lines annotated with a 180 for direction for instance. This would derived after calculating the slope of the line. 

`page_rotation`: this can be derived from the state of the image meta-tools. This will be a number in degrees of 0, 90, 180, 270.

It seems to me that the direction calculation could happen either client side or downstream in the aggregation code, so it's not clear to me how necessary it is for this to happen in the client. However, page rotation is a volunteer configuration of how they're currently viewing the subject, and we can definitely add that in.

eatyourgreens:
`classification.annotations` is a map by task key (I think) so that will need to be changed in order to record more than one line. We can either generate a new instance of the task model for each line (one task per annotation) or have one task that has many annotations (stored as an array maybe?).

Either way, I think the classification model will need some rewriting.

srallen:
We have an issue already tracking that: https://github.com/zooniverse/front-end-monorepo/issues/1226

[...]

Issue #1239, opened 2019-10-17
RFC: drawing sub-task


eatyourgreens:
**Package**
lib-classifier

**Is your feature request related to a problem? Please describe.**
https://github.com/zooniverse/front-end-monorepo/blob/1edd001edb31550a023f51b1aa15bbe70e675b3b/packages/lib-classifier/src/store/tasks/Task.js#L7-L26
Task models get their annotations from the classifications store and call `classifications.addAnnotation(value, task)` on update. It should be possible to change both of those depending on the task context (workflow task or drawing subtask.)

**Describe the solution you'd like**
Tasks should be agnostic about where their annotations come from. I think the task view component could have `annotation` and `onComplete` (or `onChange`?) as props. The task area would pass in the annotation from the classification, and `classifications.addAnnotation` as the callback. Drawing subtasks would pass in the annotation from the drawing mark, and a callback that updates the drawing tool annotation.

There might be better solutions which involve passing context to the task model instead, so I'm opening this issue for discussion. Is there a way to replace `getRoot(self).classifications` with something more flexible in the Task model code?

eatyourgreens:
Something to consider: move the annotation code (adding annotations, removing annotations etc.) out of the `classifications` store and into an `annotations` store in its own right. Use that `annotations` model as the model for both `classification.annotations` and `mark.details` (or whatever we call drawing subtask annotations.)

srallen:
For the UI, the subtask will no longer be anchored to the drawing mark anymore and should be a movable div area so that volunteers can see areas of the image as needed. This is based on feedback primarily from humanities/transcription projects.

srallen:
@eatyourgreens if we implemented the idea proposed in this issue, do you have a sense for how long it would take?

wgranger:
For floating modal boxes I used https://github.com/bokuweb/react-rnd. Rnd meaning, resize and drag, because the the ASM boxes were initially resizable.

There are a lot of different options for creating a floating modal, but I chose the option above because it seemed the most straightforward at the time. Overall, I'd say it worked pretty well, although there are a lot of important props you'll have to be mindful of should you choose the Rnd route.

eatyourgreens:
Cheers! Seems like this could be done quickly so I'd put it down as a week.

srallen:
I'm going to open a separate issue about the UI side of this and copy over the comments. The original discussion here was about the data side of how to store the sub-task annotations.

eatyourgreens:
[...]

Issue #123, opened 2018-08-24
Implement workflow steps


srallen:
**Package** lib-classifier

**Feature or Issue Description**
Our new concept of workflow steps is described here and what it would solve: https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-5.md

To this end, A MST store for workflow steps should be created.

Workflows currently are structured as so:
(The `first_task` can be the task key of T1, T2, etc. The lab auto creates an 'init' task which can be later deleted and replaced.)

```
{
  id: '1',
  first_task: 'init', 
  tasks: {
    init: {},
    T1: {},
    T2: {}
  }
}
```

What we would like to change this to:
Note the substantive change here is that the steps map to an array of task objects rather than a single object. The renaming to steps is for clarity, but in theory doesn't need to happen. Also note that the summary is a step type now. There could also be a `validation` type for transcription projects that validate/approve their transcriptions prior to submission.

```
{
  id: '1',
  first_step: 'S1',
  steps: {
    S1: [
      { type: 'multiple', value: [] }
    ],
    S2: [
      { type: 'survey', value: [] },
      { type: 'multiple', value: [] } // This is like the survey + shortcut combo task
    ],
    S3: [
      { type: 'summary', content: '' }
    ]
  }
}
```

**To Dos**

- [ ] Example

eatyourgreens:
I think tasks still need keys, so that task components can update annotations and so that we don't break services that consume classifications. I'm assuming the annotations array still looks something like this: `[{ task: 'T1', value: "something"}, {task: 'T2', value: []}, {task: 'T1', value: "something else."}]` where tasks are identified by keys and may be annotated more than once during a workflow.

srallen:
@eatyourgreens Good point. Here's a revise version of what the structure could look like:

```
{
  id: '1',
  first_step: 'S1',
  steps: {
    S1: {
      tasks: [{ task: 'T1', type: 'multiple' }],
      next: 'S2'
    },
    S2: {
      tasks: [{ task: 'T2', type: 'survey' }, { task: 'T3', type: 'multiple' }], // This is like the survey + shortcut combo task
      next: 'S3'
    },
    S3: {
      summary: { content: '' }
  }
}
```

[...]

PR #1366, opened 2019-12-09
Drawing: Add tasks to drawing tools

eatyourgreens:
Add tool.tasks to the Tool model. 
- Allow question and text tasks as drawing tool tasks. 
- Add `tool.hasCompletedTasks` to check for complete annotations on subtasks.
- Copy task tests from the Step model.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/models/tools/Tool/Tool.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/models/tools/Tool/Tool.js
@@ -1,5 +1,8 @@
 import { types } from 'mobx-state-tree'
 import { Mark } from '../../marks'
+import SingleChoiceTask  from '@plugins/tasks/SingleChoiceTask'
+import MultipleChoiceTask from '@plugins/tasks/MultipleChoiceTask'
+import TextTask from '@plugins/tasks/TextTask'
 
 const Tool = types.model('Tool', {
   color: types.optional(types.string, ''),
@@ -7,10 +10,19 @@ const Tool = types.model('Tool', {
   marks: types.map(Mark),
   max: types.optional(types.union(types.string, types.number), Infinity),
   min: types.optional(types.union(types.string, types.number), 0),
+  tasks: types.array(types.union(
+    SingleChoiceTask.TaskModel,
+    MultipleChoiceTask.TaskModel,
+    TextTask.TaskModel
+  ))
 })
   .views(self => ({
     get disabled () {
       return self.marks.size >= self.max
+    },
+
+    get hasCompletedTasks () {
+      return self.tasks.reduce((isStepComplete, task) => isStepComplete && task.isComplete, true)
     },
 
     get isComplete () {
--- a/packages/lib-classifier/src/plugins/drawingTools/models/tools/Tool/Tool.spec.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/models/tools/Tool/Tool.spec.js
@@ -1,4 +1,10 @@
 import Tool from './Tool'
+import {
+  MultipleChoiceTaskFactory,
+  SingleChoiceTaskFactory
+} from '@test/factories'
+import taskRegistry from '@plugins/tasks'
+import ClassificationStore from '@store/ClassificationStore'
 
 const toolData = {
   color: '#ff0000',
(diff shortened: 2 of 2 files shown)


eatyourgreens:
I’m pretty sure we’ve already discussed this in an RFC for the new classifier. I’ll look it up tomorrow.

eatyourgreens:
The reason being that Caesar has trouble with tool details.

eatyourgreens:
There’s no changes to the lab since this is only an internal representation of the tool tasks within the classifier’s models. The payload received from Panoptes won’t need to change.

srallen:
I don't see this proposed change as clearly discussed in any of the possibly related RFC issues: 
- Transcription sub-task: https://github.com/zooniverse/front-end-monorepo/issues/1314#issuecomment-561717018
- Drawing sub-task: https://github.com/zooniverse/front-end-monorepo/issues/1239
- Mark models: https://github.com/zooniverse/front-end-monorepo/issues/500

I've asked for clarification from @CKrawczyk but there are a few other items to discuss such as which kinds of tasks the sub-task should support as well. I really think we need to be sure we discuss this in issue first, then write an ADR if it's a change affecting architecture, which I believe this is, then write the code. It's difficult to ask for documentation after code is written and I think we should get better at practicing doing documentation first as a team.

eatyourgreens:
I’ll look it up tomorrow and see what we decided. It was basically that subtasks annotations should be represented in exactly the same way as classification annotations ie. subtasks should have task keys. They don’t in PFE.

eatyourgreens:
This isn’t a replacement for `details`. For each task snapshot in `details`, we will add a task key, create a Task and then put the task in the `tasks` array. It’s a purely internal representation of tool tasks using the existing classifier models.

srallen:
Ok, thanks for some clarification. I'm finding it hard to follow what is for what, which is why I keep asking for documentation prior to work beginning. Is there an issue that this is linked to?

eatyourgreens:
Just the issues about adding tasks and their annotations to drawing marks.

eatyourgreens:
Here you go. https://github.com/zooniverse/front-end-monorepo/issues/1239#issuecomment-561732576

eatyourgreens:
Reading back through the code, the logic for validating a tool here is completely wrong, which is my fault for copying it directly from the Step model. The spec for validating a tool with subtasks should be:
- a mark is complete when it has annotations for each required tool task.
- a tool is complete when all of its marks are complete.

eatyourgreens:
@srallen Here's where we discussed changing tool details: https://github.com/zooniverse/front-end-monorepo/issues/123#issuecomment-418108424

srallen:
A single comment in an old closed RFC issue focused on a different issue and solution, even if there is some relation, isn't a discussion. A decision on the sub-task wasn't made in ADR 5. 

I'd prefer we finalize the plan for sub-tasks in its relevant RFC, https://github.com/zooniverse/front-end-monorepo/issues/1239, and then commit the decision as an ADR, so it is easy to reference, follow, and remember for the whole team.

eatyourgreens:
Sounds good to me. I've been updating #1239 with comments about progress, and planned next steps, anyway.

This PR is just focussed on adding tasks to the state tree so that we can render them in the subject viewer. I do think there is a separate discussion that needs to be had fairly soon about how the annotations from those tasks will be serialised and posted to Panoptes/Caesar/any other service that wants to use them. #500 seems like the natural place to have that discussion. https://github.com/zooniverse/front-end-monorepo/issues/500#issuecomment-564448229

By the way, I'm fine with renaming `tool.tasks` to `tool.details` but I think it will be a little confusing since its type will be `types.array(Task)`. We'll probably also end up with code like:
`tool.details.map(task => <TaskComponent task={task} />` which also might not be clear to a first-time viewer who doesn't know that tasks are called details.

srallen:
I'm fine with a renaming when there's a good reason and we document that in an ADR, but it's serialized from Panoptes as `details` so how would applying a response as a snapshot work with this renaming? We can have a breaking change, but right now I don't think this will work without a mapping from one to another and then later a plan to update the lab to change `details` to `tasks`. If we're going to do a breaking change, we could also plan on flattening the task structure on the workflow side as well as on the annotation side similar to what's discussed in #500.

eatyourgreens:
`details` is invalid as a Snapshot because the Task model requires an identifier and details don’t have task keys. Same reason workflow tasks can’t be applied directly as a snapshot.

srallen:
I'm confused by your response. Workflow tasks do have unique identifiers, T0, T1, etc, so my question still remains. How does this work with a response from Panoptes?

Doing a quick check with a project that has a drawing tool with a sub-task created, here's what the response looks like from Panoptes with a sub-task defined in details:

Screen Shot 2019-12-11 at 1 15 42 PM

And how it's stored in the stores:

Screen Shot 2019-12-11 at 1 15 13 PM

tasks is empty.

Either we need to match the data structure coming from Panoptes or have a plan on how this is changing with documentation and likely some kind of helper function to make the conversion like we do with tasks to steps as a temporary interim step until the lab can be updated.

eatyourgreens:
Sorry, I meant  workflow tasks don't have a `taskKey` property, which is required by the Task model. There's a line in one of the classifier models that adds this property to new tasks before creating them.

eatyourgreens:
Looking at the workflow model, to see how workflow tasks were modelled, I think `tool.details` will be `types.array[types.frozen]` and then `tool.tasks` will be an array of Task models built from the details array.

I’ll read up, though, to see if there’s a way of applying a snapshot via a function that turns it into a valid Task snapshot.

eatyourgreens:
Oh, and tasks is supposed to be empty in this PR. It doesn’t add any code to create the task models.

srallen:
Before diving too much into that, I think the option of flattening the task structure should be considered. I'm in favor of doing this on the workflow model side like we're considering with the annotations.

eatyourgreens:
From a mark, I’d like to access subtasks as `mark.tool.tasks` but that could be a computed view, I suppose. See my earlier comment about calculating completeness for marks and their parent tools. 

My reservation about storing tasks elsewhere in the state tree would be that it introduces a dependency that makes tools and marks hard to mock and test. We’d have to mock a workflow as well, I think.

eatyourgreens:
References may help here.

eatyourgreens:
#1372 sets up tools as ancestors of marks in the tree, which kind of points where I’m going with this.

eatyourgreens:
Closing this in favour of a PR where I spec the behaviour first.

PR #1368, opened 2019-12-09
Add PEP for Safari 12 fallback

srallen:

Package: lib-classifier

Describe your changes:
This an alternative to #1332 and for comparison reasons. I've tested this in browser stack using Safari 12 and it works as expected. Other modern browsers do not fire the polyfill and it is scoped to just the `InteractionLayer` following the PEP API. 

The advantage of this is that there's less code we have to maintain and less guesswork on making sure we're correctly falling back when we need to. Around September 2020, we should be able to drop this because Apple releases major version for Safari annually. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/package.json
+++ b/packages/lib-classifier/package.json
@@ -31,6 +31,7 @@
     "hash.js": "~1.1.7",
     "lodash": "~4.17.11",
     "mobx-utils": "~5.0.2",
+    "pepjs": "~0.5.2",
     "react-pose": "~4.0.8",
     "react-resize-detector": "~4.2.0",
     "seven-ten": "~3.4.0",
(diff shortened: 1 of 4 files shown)


eatyourgreens:
Looks good to me. Just two questions. 

Should the interaction layer be styled with `touch-action: none`?

Has this been tested on a touch screen eg. mobile Safari?

srallen:
I don't have an actual iOS device that is running Safari 12, but it worked testing in browser stack. (I'll be opening another follow up PR to modify the webpack config so we can correctly load up localhost in browser stack for Safari testing)

The PEP.js docs say the `touch-event` is used by the script to know which element to target for polyfilling, so I don't think we need to style it.

eatyourgreens:
BrowserStack runs iOS on actual devices, so that sounds good to me.

We might need `touch-action: none` (or `touchAction='none' in the SVG attributes?) to override the default window scrolling on iOS/Android but that can be tested on staging.

PR #1369, opened 2019-12-10
Drawing: Add a transcription line tool

### LINKED ISSUE #1349: Transcription line drawing tool
**Package** lib-classifier

This is related to implementing the transcription task (#1313) (Also see [ADR 20](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-20.md). The standard line tool has been identified as not having exactly the UI nor drawing interaction that is desired for transcription task. This variant of the line drawing tool for transcription will instead:

- Be drawn by two clicking actions for the start point and the end point
-  Have a visual indication of which point is the starting point and which is the end

Associated [user stories](https://docs.google.com/document/d/16abI-wkRlEXsWgACfFQVqwO76aEopohIjQiRfNQKWiw/edit)

> As a user I want to click at the beginning of a line of text and see an open circle icon indicating the start of the line.

> As a user I want to see a line connecting the first circle I created to my pointer so I can keep track of which line I’m transcribing.

> As a user I want to click at the end of a line of text and see a closed circle indicating the end of the line.

> As a user I want to see the line connecting the start and end of the line of text I’m planning to transcribe.

eatyourgreens:
- Add a transcription line tool, based on the polygon tool from PFE.
- Add an `onFinish` handler to all marks, which fires when initial mark creation is complete and resets the interaction layer.
- Add `fill` and `radius` options to drag handles, so that we can override the default properties and differentiate between the start and end of a line.
- Add tests for the `DragHandle` component.

Try it out on staging project 908, workflow 3372.
http://localhost:8080/?project=908&workflow=3372

Package:
lib-classifier

Closes #1349.

Describe your changes:

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
The new tool type is currently set to `polygon` so that it can be easily tested with the project builder.

packages/lib-classifier/src/plugins/drawingTools/models/tools/TranscriptionLineTool/TranscriptionLineTool.js
```diff
@@ -0,0 +1,21 @@
+import { types } from 'mobx-state-tree'
+import Tool from '../Tool'
+import { TranscriptionLine } from '../../marks'
+
+const TranscriptionLineTool = types.model('TranscriptionLine', {
+  marks: types.map(TranscriptionLine),
+  type: types.literal('polygon')
```
eatyourgreens: This is set to `polygon` for testing and should be changed for the final tool.

eatyourgreens:
Now with tests that check the rendered line against the user stories.

eatyourgreens:
Adding some tests for the changes to the `DragHandle` component might help with the coverage here.

eatyourgreens:
Mmmm, merge conflicts because the `DragHandle` props changed.

eatyourgreens:
Should this tool be moved to `drawingTools/experimental`? I've just realised I've put it in with all the other public drawing tools.

srallen:
@eatyourgreens That's a good question. I think we should keep the transcription task and associated new tools behind an experimental tool feature flag. Perhaps add a view function to the project model to check if X tool is in the array in a separate PR? We haven't had to do this yet in the new code.

eatyourgreens:
@srallen @rogerhutchings suggested a `drawingTools/experimental` directory for experimental tools, which I think is a good idea too. We'd have to remember to move the code when a tool is made public.

srallen:
I like the suggestion to have a clear file structure for them and we can also include some readme information about why these are experimental.  :+1:

eatyourgreens:
I've moved the `TranscriptionLine` components and models to `drawingTools/experimental`.

packages/lib-classifier/src/plugins/drawingTools/models/tools/index.js
```diff
@@ -1,2 +1,4 @@
 export { default as LineTool } from './LineTool'
 export { default as PointTool } from './PointTool'
+export { default as Tool } from './Tool'
+export { default as TranscriptionLineTool } from '@plugins/drawingTools/experimental/models/tools/TranscriptionLineTool'
```
srallen: Same comment about the import here.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/helpers/getIcon.js
```diff
@@ -43,7 +43,7 @@ const icons = {
 }
 
 function getIcon (icon) {
-  return icons[icon] || null
+  return icons[icon] || (() => null)
```
srallen: I don't understand the purpose of this change.

packages/lib-classifier/src/plugins/drawingTools/experimental/models/tools/TranscriptionLineTool/TranscriptionLineTool.js
```diff
@@ -0,0 +1,21 @@
+import { types } from 'mobx-state-tree'
+import { Tool } from '@plugins/drawingTools/models/tools'
+import { TranscriptionLine } from '../../marks'
+
+const TranscriptionLineTool = types.model('TranscriptionLine', {
+  marks: types.map(TranscriptionLine),
+  type: types.literal('polygon')
```
srallen: I think we need to distinguish this from polygon, even with the similarities. `transcriptionLine`?

packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
```diff
@@ -0,0 +1,79 @@
...
+
+import { Mark } from '@plugins/drawingTools/models/marks'
+
+const TranscriptionLineModel = types
+  .model('TranscriptionLineModel', {
+    x1: types.maybe(types.number),
+    y1: types.maybe(types.number),
+    x2: types.maybe(types.number),
+    y2: types.maybe(types.number),
+    closed: types.optional(types.boolean, false)
```
srallen: Is there a reason to use `closed` instead of PFE's `_inProgress`? This reverses the logic, so if the logic does have a reason to be reversed, then I might prefer `finished` to be consistent with the other action and function names being used around the code since closed === finished.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/helpers/getIcon.js
eatyourgreens: The original code will crash the classifier if there's no icon, since `null` isn't a React component.

packages/lib-classifier/src/plugins/drawingTools/experimental/models/tools/TranscriptionLineTool/TranscriptionLineTool.js
eatyourgreens: Sounds good to me. This is only here because my test workflow uses `polygon` in the project builder.

packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
eatyourgreens: `finished` sounds better to me. `closed` was copied from the polygon code, and is used by ASM (https://github.com/zooniverse/front-end-monorepo/issues/1314#issuecomment-564084683) but I don't think it's needed either.

I think that makes sense for polygons too, because my understanding is a polyline can be finished *and* either open or closed (I could be wrong.)

packages/lib-classifier/src/plugins/drawingTools/components/index.js
```diff
@@ -4,3 +4,4 @@ export { default as DeleteButton } from './DeleteButton'
 export { default as DrawingToolRoot } from './DrawingToolRoot'
 export { default as Line } from './Line'
 export { default as Point } from './Point'
+export { default as TranscriptionLine } from '@plugins/drawingTools/experimental/components/TranscriptionLine'
```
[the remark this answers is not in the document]
eatyourgreens: I dislike looking up through the filesystem for dependencies, but that's mostly when the path is something like `../../../target`. My preference is to use aliases instead of `../` wherever possible but I have no strong opinion here.

packages/lib-classifier/src/plugins/drawingTools/components/Line/Line.spec.js
```diff
@@ -18,4 +31,56 @@ describe('Line tool', function () {
...
+
+    it('should move the start point on drag move', function () {
+      const wrapper = shallow(<Line
+        active
+        mark={mark}
+      />)
+      const dragMove = wrapper.find(DragHandle).find('[x=100]').prop('dragMove')
+      dragMove({}, { x: 10, y: 20 })
+      expect(mark.x1).to.equal(110)
+      expect(mark.y1).to.equal(220)
```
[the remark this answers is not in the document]
eatyourgreens: Good point.

eatyourgreens:
Stories are definitely lacking for the drawing tools in general. It's on my todo list as something to open in its own PR. Hopefully they're simple to set up, although the `draggable` story is still broken.

eatyourgreens:
>Then later in a follow up PR if we agree to filter those out, then the complete action on the DrawingTask model can filter out properties prefaced by an underscore? (Or alternatively they're filtered at onCompleteClassification.

Marks could have a common action, or view, that's called on complete to get the filtered properties. If we use underscore, then that can be defined on the Mark model. I might be overthinking this.

`task.complete()` shouldn't alter the mark, in my opinion, because we might return to this task via the Back button and render the marks for editing again. The drawing task annotation also contains references to the marks on the original tool, not actual marks, since a node can't exist at two places in the tree. I think `onCompleteClassification` is safer for final serialisation of the drawn marks.

packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
eatyourgreens: I've renamed it and moved it to volatile state.

One question that did occur to me: can finished state be set from a snapshot eg. loading in a previous annotation and displaying it?

packages/lib-classifier/src/plugins/drawingTools/components/index.js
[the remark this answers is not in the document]
eatyourgreens: I think what I really want here is something like
```js
import * as experimentalComponents from '@plugins/drawingTools/experimental/components'
export { ...experimentalComponents }
```

eatyourgreens:
@srallen I've added stories for finished, active and unfinished lines. There's no interactivity, because the `draggable` HOC is still broken in the storybook.

beckyrother:
@srallen how about this for the icon (svg attached in a zip) 
transcription
[transcription.svg.zip](https://github.com/zooniverse/front-end-monorepo/files/3989307/transcription.svg.zip)

packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
srallen: It looks like the answer is no: https://mobx-state-tree.js.org/concepts/volatiles#modelvolatile

> Volatile props will not show up in snapshots, and cannot be updated by applying snapshots

Volatile state otherwise seems like a good pick for this use case. We _might_ be able to use `onSnapshot` to call an action that updates the volatile state though: https://mobx-state-tree.js.org/API/#onsnapshot

eatyourgreens: Sorry, I meant should finished state be set from a snapshot in our rendering code? I vaguely remember there being a user story where we show you a finished annotation from someone else.

The transcription line stories have examples of calling `mark.finish()` to finish a mark. The only question would be, how does the rendering component know that a mark from an annotation snapshot is finished and cannot be edited at all.

Something to think about in the new year. Go and enjoy Christmas!

eatyourgreens: Should we be able to load an incomplete classification from Panoptes and render those annotations as marks? I believe ASM allowed this since transcription of a document can take more than one session.

I think that's ok here because lines rendered from a saved annotation will always be finished.

srallen: Yes, eventually. This should be in the user stories,.

packages/lib-classifier/src/plugins/drawingTools/components/index.js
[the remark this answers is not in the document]
srallen: Meant to write before hitting resolve: I don't have a strong preference. I'm still getting used to the aliased imports.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.js
```diff
@@ -0,0 +1,69 @@
+import PropTypes from 'prop-types'
+import React from 'react'
+import { DragHandle } from '@plugins/drawingTools/components'
+
+const FINISHER_RADIUS = 8
+const GRAB_STROKE_WIDTH = 6
+
+function TranscriptionLine ({ active, mark, onFinish, scale, svg }) {
```
srallen: Cruft svg prop needs to be removed.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.js
```diff
@@ -0,0 +1,69 @@
...
+          svg={svg}
+          x={x1}
+          y={y1}
+          dragMove={(e, d) => onHandleDrag({ x1: x1 + d.x, y1: y1 + d.y, x2, y2 })}
+        />}
+      {active &&
+        <DragHandle
+          radius={6}
+          scale={scale}
+          svg={svg}
```
srallen: svg prop here should be removed.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.js
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <g>
+      <line x1={x1} y1={y1} x2={x2} y2={y2} />
+
+      {active &&
+        <DragHandle
+          fill='transparent'
+          radius={6}
+          scale={scale}
+          svg={svg}
```
srallen: svg prop here should be removed.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.js
```diff
@@ -0,0 +1,69 @@
...
+    </g>
+  )
+}
+
+TranscriptionLine.propTypes = {
+  active: PropTypes.bool,
+  mark: PropTypes.object.isRequired,
+  onFinish: PropTypes.func,
+  scale: PropTypes.number,
+  svg: PropTypes.instanceOf(Element).isRequired
```
srallen: This can be removed from the proptype definitions.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
```diff
@@ -0,0 +1,96 @@
...
+import sinon from 'sinon'
+import { TranscriptionLineTool } from '@plugins/drawingTools/models/tools'
+import { DrawingToolRoot } from '@plugins/drawingTools/components'
+import TranscriptionLine from './TranscriptionLine'
+
+
+class DrawingStory extends Component {
+  constructor () {
+    super()
+    this.svg = React.createRef()
```
srallen: This can be removed.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  render () {
+    const { children, mark, tool } = this.props
+    const { svg } = this
```
srallen: This ref can be removed in favor of using context

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
```diff
@@ -0,0 +1,96 @@
...
+    return (
+      <svg viewBox='0 0 300 400' height={300} width={400} ref={this.svg}>
```
srallen: ref can be removed

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
```diff
@@ -0,0 +1,96 @@
...
+        <DrawingToolRoot
+          label='Transcription line'
+          mark={mark}
+          svg={svg}
+          tool={tool}
+        >
+          {React.cloneElement(children, { svg: svg.current })}
```
srallen: This isn't needed anymore.

packages/lib-classifier/src/plugins/drawingTools/experimental/models/tools/TranscriptionLineTool/TranscriptionLineTool.js
srallen:
```suggestion
  type: types.literal('transcriptionLine')
```

eatyourgreens:
I think this is ready now. I added the `tool` view, fixed the tool type and cleaned up leftover props and refs.

packages/lib-classifier/src/plugins/drawingTools/models/tools/index.js
eatyourgreens: Changing this to a relative path broke the export but I'm not sure why.

eatyourgreens:
TODO: add the tool icon.

eatyourgreens:
Transcription tools in the classifier, showing the SVG icon as a blob of colour instead of the letter A.
The SVG icon is rendering as an amorphous blob of colour, so I'm going to leave it out of this PR.

packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
```diff
@@ -0,0 +1,86 @@
...
+      const y = self.y1
+      return { x, y }
+    },
+
+    get isValid () {
+      return true
+    },
+
+    get tool () {
+      return getParentOfType(self, TranscriptionLineTool)
```
srallen: Should this change to `getParent` too?

eatyourgreens:
Invalid task snapshots log an error in the console but I don’t think we’ve ever had error handling in the classifier task area.

packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
eatyourgreens: Depends whether you want strict typing for the parent tool. The other mark models use `getParentofType` to return a specific type of tool but the generic Mark uses `getParent`.

packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
```diff
@@ -0,0 +1,86 @@
...
+
+const TranscriptionLineModel = types
+  .model('TranscriptionLineModel', {
+    x1: types.maybe(types.number),
+    y1: types.maybe(types.number),
+    x2: types.maybe(types.number),
+    y2: types.maybe(types.number)
+  })
+  .volatile(self => ({
+    finished: false
```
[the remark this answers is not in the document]
eatyourgreens: I'm not sure if identifiers can be stored in volatile state. I think identifiers must be present in snapshots.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
```diff
@@ -0,0 +1,89 @@
...
+    )
+  })
+  .add('active', function () {
+    const tool = TranscriptionLineTool.create({
+      color: 'red',
+      type: 'transcriptionLine'
+    })
+    const mark = tool.createMark({
+      id: 'transcriptionLine1',
+      x1: 10,
```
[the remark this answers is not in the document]
eatyourgreens: I'm pretty sure `createMark()` adds the `toolType` automatically. That's how the other drawing tools are set up.

packages/lib-classifier/src/plugins/drawingTools/experimental/models/tools/TranscriptionLineTool/TranscriptionLineTool.js
```diff
@@ -0,0 +1,21 @@
...
+import { Tool } from '@plugins/drawingTools/models/tools'
+import { TranscriptionLine } from '@plugins/drawingTools/models/marks'
+
+const TranscriptionLineTool = types.model('TranscriptionLine', {
+  marks: types.map(TranscriptionLine),
+  type: types.literal('transcriptionLine')
+})
+  .actions(self => {
+    function createMark (mark) {
+      const newMark = TranscriptionLine.create(mark)
```
eatyourgreens:
```suggestion
      const newMark = TranscriptionLine.create(Object.assign({}, mark, { toolType: self.type }))
```

packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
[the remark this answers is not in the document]
srallen: It's a general question. I realize identifiers can't, but others we were considering for underscored could possibly? I'm not sue.

beckyrother:
@eatyourgreens Is there a link to that issue @srallen requested? Sounds like it'll be good to have a discussion about style here

packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
srallen: Hm, it seems odd to have the base model potentially have a different function that the child ones, but I don't have a strong opinion on it.

eatyourgreens:
@beckyrother just the comment and screenshot at the moment. https://github.com/zooniverse/front-end-monorepo/pull/1369#issuecomment-575513064

srallen:
The request for the issue to be opened is to make sure that it is known and tracked once this PR closes since the change won't be happening in this PR. I ask and delegate issue management because if I did it all myself along with the docs, milestone, project management, it would take up a significant amount of my time. My expectation is that part of code ownership and responsibility when opening a PR on our team is to acknowledge requests for issue creation and to do it. 

It's late on a Friday, so I understand it won't happen today. That's ok. Monday is fine.

eatyourgreens:
Ive been in a waiting room in A&E. I’ll deal with this some time next week.

srallen:
:( I hope everything is ok.

eatyourgreens:
The demo is back up at http://localhost:8080/?project=908&workflow=3372

beckyrother:
Thanks for opening that issue Sarah!

PR #1375, opened 2019-12-11
[VSLCV] Individual bar focus

### LINKED ISSUE #1243: [VSLCV] individual bar focus
**Package** lib-classifier

The bars in the BarChartViewer should have some kind of individual bar focus/hover state that shows the individual value. The way the stats bar chart does this by showing the number inside the bar width probably isn't going to work for this because we don't have an idea what the values will be. Any number with more than two or three digits won't fix inside the bar width comfortably. 

Probably some kind of tooltip and the library we're using, vx, has one: https://vx-demo.now.sh/static/docs/vx-tooltip.html

Unfortunately, they don't have a demo of it in the gallery.

Maybe @beckyrother has other ideas.

srallen:

Package: lib-classifier

Closes #1243

Describe your changes:
This adds tooltips using a react implementation of tippy.js. The library is opinionated with styles and they're necessary for correct positioning (uses CSS under the hood), so in order for that to work, I've added mocha and webpack configs to support css import in the component. As well as some `!important`s in some styled-component extensions. The styles are approved by @beckyrother in #1391 and this is ready for final dev review. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
@beckyrother how does the implementation of the design aside from the position issues look?

beckyrother:
Looks perfect aside from positioning!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXTooltip/VXTooltip.js
```diff
@@ -0,0 +1,99 @@
...
+    top: initialTop
+  } = props
+  const fontSize = theme.text.small.size
+  const horizontalPadding = theme.global.edgeSize.medium
+  const verticalPadding = theme.global.edgeSize.xsmall
+  const tooltipStyles = { background: 'transparent', borderRadius: 'none', boxShadow: 'none', padding: 0, zIndex: '1' }
+  const triangleHeight = 10
+  const triangleWidth = 20
+  const heightOfTooltip = getNumberValue(fontSize) + (getNumberValue(verticalPadding) * 2) + (triangleHeight * 1.5)
+  const widthOfTooltip = getNumberValue(fontSize) + (getNumberValue(horizontalPadding) * 2) + (triangleHeight * 1.5)
```
eatyourgreens: Why is the width calculated from the font size and triangle height?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXTooltip/VXTooltip.js
```diff
@@ -0,0 +1,99 @@
...
+import PropTypes from 'prop-types'
+import { withTheme } from 'styled-components'
+import { Tooltip } from '@vx/tooltip'
+import { withBoundingRects, withBoundingRectsProps } from '@vx/bounds'
+import { Box } from 'grommet'
+import { SpacedText } from '@zooniverse/react-components'
+import Triangle from '../../../../shared/Triangle'
+
+function getNumberValue (string) {
+  const splitString = string.split('px')
```
eatyourgreens: This will break if fonts are defined in ems, rems or points.

srallen: Yeah, it's not ideal. It's assuming that we're using the zoo theme, which is using px.

srallen:
Gonna look into https://popper.js.org/ as an alternative to deal with the inconsistent positioning.

srallen:
I'm going to checkout a branch from this one to experiment with popper.js

srallen:
I redid this in #1391 with a react implementation of tippy.js and it works well. The tooltips can be seen in the storybook using either keyboard tab focus or mouse enter events.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/BarChartViewer/components/Bars/Bars.js
```diff
@@ -0,0 +1,127 @@
...
+        <StyledTippy
+          arrow={true}
+          animation='scale'
+          content={<StyledTooltipContent fontFamily={font.family} label={value.toString()} />}
+          key={key}
+          placement='top'
+          trigger='mouseenter focus'
+        >
+          <svg
+            tabIndex='0'
```
eatyourgreens: Does `tabindex` on SVG work in Edge? If not, then this might need `focusable` too.

eatyourgreens: Actually, I'm not sure why this SVG is focusable.

srallen: I checked Edge and it's working as expected. I left this as tabbable for screen reader read out for the bar chart as a whole.

eatyourgreens: I'm confused because the child `Bar` is what should be read out, and that does have `focusable` set.

eatyourgreens: If this SVG should be read out, it'll need an appropriate role and a label. I think it will also mask any children from being read out.

srallen: Ok, I wasn't sure about which would take precedence or if they would work together (somehow?). It can be removed.

srallen:
> although Period filter 1: 0.4661477096 is slightly confusing without context. Is the number a time period of some sort?

I'm not sure! The sample data I have from the researcher didn't label it and it was something I was going to ask about on Thursday.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/BarChartViewer/components/Bars/Bars.js
srallen: Removing the tab index from the svg element breaks the tooltips from showing at all in Firefox :(

srallen: Ok, after reviewing the tippy.js-react docs, I remember now why I did this. It wasn't for voice over, but in fact a workaround for the fact that vx doesn't forward refs in the `Bar` component: 

> If you're using a library that doesn't forwardRef for you, and doesn't give access to the ref via innerRef or similar, you can use a wrapper <span> element as a workaround.

I used an svg element since we're within an svg. Is there a better element I can use like span but works in an svg?

srallen: The example showed a tab index being used on the workaround wrapper element: https://github.com/atomiks/tippy.js-react#component-children

srallen:
The Firefox bug might be with the underlying positioning library (popper) that tippy uses. I'll have to do some more investigation to confirm. If it is indeed the underlying library, we _might_ have to wait for the next version release for the fix (which according to their repos should be soon).

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/BarChartViewer/components/Bars/Bars.js
eatyourgreens: What's confusing here is that `tabIndex` is defined on the wrapper, but `focusable` and the accessible label are added its child. They should really all be declared together, not on different elements.

eatyourgreens: >Removing the tab index from the svg element breaks the tooltips from showing at all in Firefox :(

Did you remove it or move it to the `Bar`? It'll need to be present somewhere in order for the element to be focusable in the browser.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/BarChartViewer/components/Bars/Bars.js
```diff
@@ -0,0 +1,127 @@
...
+          >
+            <Bar
+              aria-label={alt}
+              data-label={label}
+              data-value={value}
+              fill={fill}
+              focusable
+              height={barHeight}
+              index={index}
+              role='list item'
```
eatyourgreens:
```suggestion
              role='listitem'
```
No whitespace in attribute values.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/BarChartViewer/components/Bars/Bars.js
srallen: Yes, I tried that first.

srallen: Since the documentation demonstrates this is the workaround for situations where forward refs aren't correctly supported, I'd like to leave this as is. I think it'd be appropriate to open an issue on the vx repository about the forwarding refs issue, but I don't think I can drop this implementation without dropping vx and/or tippy and I don't really want to do either of those things since they generally are working. Is having an issue to track this to come back to it ok for now?

srallen:
@eatyourgreens I don't know if I can fix the positioning bug in Firefox since all of that is being handled by the popper.js dependency, which in turn is a dependency of tippy.js. The code I've implemented does nothing for calculating position. I tried monkeying around with my installed node packages to try to single out exactly what's happening and I'm having a bit of trouble getting local changes to show up... I'm hesitant to spend a lot more time on this because there several positioning bugs marked as fixed in the popper repository slated for v2 and that is in release candidate status. There's parallel issues for tippy to update once v2 releases, so I'd like to try waiting for that since I think it's going to happen shortly. Would a tracking issue for this bug and monitoring of the release status of the dependencies be a suitable temporary action for this so I can continue progress?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/BarChartViewer/components/Bars/Bars.js
eatyourgreens: That's fine as long as you move `focusable` and the accessible labels up from the `Bar` component to this wrapper. The change I'm asking for is that focusable behaviour is currently declared on two different elements.

>What's confusing here is that tabIndex is defined on the wrapper, but focusable and the accessible label are added to its child. They should really all be declared together, not on different elements.

eatyourgreens:
I'm not sure what to do about the Firefox bug either. If it isn't fixed in v2, will this be reverted?

My other change request was to silence the annoying announcement of the axis labels in VoiceOver. I think `role=presentation` on each axis should do that.

srallen:
> If it isn't fixed in v2, will this be reverted?

I'll investigate all possible options including reverting, but if I can pinpoint the bug, possibly contribute a fix.

eatyourgreens:
For Firefox, a workaround would be to add a clear focus indicator to the active bar, so that I can see which one the number refers to.
Bar chart in Firefox, showing the tooltip in the wrong place.

srallen:
I've opened https://github.com/zooniverse/front-end-monorepo/issues/1436 to track the positioning bug in FIrefox. I believe the other items have been addressed, so @eatyourgreens this is ready for review again.

eatyourgreens:
Looks good! I’ll try it in VoiceOver tomorrow. One small thing. For the focusable bars, `aria-label` should have been moved too.

srallen:
I've noted it in https://github.com/zooniverse/front-end-monorepo/issues/1440

PR #1380, opened 2019-12-13
Add an auth client

rogerhutchings:
Adds a basic auth client as detailed in ADR 21 (https://github.com/zooniverse/front-end-monorepo/pull/1361).

- Add auth client
- Add unit tests
- Add functional tests in isolation
- Add functional tests for staging API

My goals for this were to:
- have a client that could accept a config
- ensure it works on both client and server side
- minimize the amount of magic going on

There aren't many docs yet, but the code is all annotated with JSDoc. I think there's a subtlety around how classes and class instances are annotated that I'm missing, but otherwise it should be pretty clear. Hinting works in the IDE as well, which is nice.

So, there's stuff still to do, obviously, but this is my first pass and I'd like to get feedback on it before adding more in. Specifically, if anyone has any experience in headless browser testing, that'd be super useful for adding browser tests!

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-auth/README.md
+++ b/packages/lib-auth/README.md
@@ -1,6 +1,6 @@
 # @zooniverse/auth
 
-An authentication library for the Panoptes API which provides tokens for making authenticated requests.
+An authentication library for the Panoptes API which provides tokens for making authenticated requests. It uses [`axios`](https://github.com/axios/axios) for making XHR requests under the hood.
 
 ## Example
 
--- a/packages/lib-auth/README.md
+++ b/packages/lib-auth/README.md
@@ -1,0 +1,57 @@
+# @zooniverse/auth
+
+An authentication library for the Panoptes API which provides tokens for making authenticated requests.
+
+## Example
+
+```js
+import createClient from '@zooniverse/auth
+
+const client = createClient({
+  clientAppID: 'someClientAppID',
+  hostUrl: 'http://enterprise.com'
+})
+
+async function login () {
+  const user = {
+    login: 'jean-luc@picard.com'
+    password: 'TeaEarlGreyHot'
+  }
+  return client.signIn(user)
+}
+
+// Use this to make an authenticated request, like getting the user object!
+const accessToken = await login()
+```
+
+## Testing
+
+Tests are kept in the `/test` directory, which allows us to keep all test-related code in one place, rather than having unit tests with the code and functional tests in another directory.
+
+### Unit tests
+
+The unit tests test individual methods, and are kept in the `/test/unit` directory.
+
+```sh
+yarn test:unit
+```
+
+### Functional tests
+
+The functional (or end-to-end) tests treat the client as a black box, and are kept in the `/test/functional` directory. There are two types of functional testing available:
+
+#### Isolation
+
+The test suite is run against the client in isolation, with requests to the API mocked by `nock`.
+
+```sh
+yarn test:functional:node
+```
+
+#### Staging
+
+Performs the same functional test suite as `isolation`, but against the live staging API.
+
+```sh
+yarn test:functional:node:staging
+```
--- a/packages/lib-auth/package.json
+++ b/packages/lib-auth/package.json
@@ -13,7 +13,7 @@
     "docs": "jsdoc src",
     "lint": "zoo-standard",
     "lint:fix": "zoo-standard --fix",
-    "test": "mocha ./test/unit/playground.js",
+    "test": "yarn test:unit",
     "test:ci": "BABEL_ENV=test yarn test:unit --reporter=min",
     "test:functional:node": "mocha ./test/functional/node/isolation/test.js",
     "test:functional:node:staging": "mocha ./test/functional/node/staging/test.js",
--- a/packages/lib-auth/package.json
+++ b/packages/lib-auth/package.json
@@ -10,7 +10,7 @@
     "yup": "~0.27.0"
   },
   "scripts": {
-    "docs": "jsdoc src",
+    "docs": "jsdoc2md src/*.js > docs/README.md",
     "lint": "zoo-standard",
     "lint:fix": "zoo-standard --fix",
     "test": "yarn test:unit",
@@ -25,6 +25,7 @@
     "chai": "~4.2.0",
     "dirty-chai": "~2.0.1",
     "jsdoc": "~3.6.3",
+    "jsdoc-to-markdown": "~5.0.3",
     "mocha": "~6.1.4",
     "nock": "~11.7.0"
   }
--- a/packages/lib-auth/package.json
+++ b/packages/lib-auth/package.json
@@ -14,6 +14,7 @@
     "lint": "zoo-standard",
     "lint:fix": "zoo-standard --fix",
     "test": "mocha ./test/unit/playground.js",
+    "test:ci": "BABEL_ENV=test yarn test:unit --reporter=min",
     "test:functional:node": "mocha ./test/functional/node/isolation/test.js",
     "test:functional:node:staging": "mocha ./test/functional/node/staging/test.js",
     "test:unit": "mocha \"./test/unit/**/*.spec.js\""
--- a/packages/lib-auth/package.json
+++ b/packages/lib-auth/package.json
@@ -1,0 +1,30 @@
+{
+  "name": "@zooniverse/auth",
+  "version": "1.0.0",
+  "main": "src/index.js",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "axios": "~0.19.0",
+    "cookie": "~0.4.0",
+    "detect-node": "~2.0.4",
+    "yup": "~0.27.0"
+  },
+  "scripts": {
+    "docs": "jsdoc src",
+    "lint": "zoo-standard",
+    "lint:fix": "zoo-standard --fix",
+    "test": "mocha ./test/unit/playground.js",
+    "test:functional:node": "mocha ./test/functional/node/isolation/test.js",
+    "test:functional:node:staging": "mocha ./test/functional/node/staging/test.js",
+    "test:unit": "mocha \"./test/unit/**/*.spec.js\""
+  },
+  "devDependencies": {
+    "@zooniverse/standard": "~1.0.0",
+    "axios-debug-log": "~0.6.2",
+    "chai": "~4.2.0",
+    "dirty-chai": "~2.0.1",
+    "jsdoc": "~3.6.3",
+    "mocha": "~6.1.4",
+    "nock": "~11.7.0"
+  }
+}
--- a/packages/lib-auth/src/client.js
+++ b/packages/lib-auth/src/client.js
@@ -300,7 +300,7 @@ class Client {
 
     // If there's no session, we can't log out.
     if (!currentSession) {
-      return null
+      return
     }
 
     const csrfToken = await this._getCSRFToken()
@@ -311,7 +311,7 @@ class Client {
     }
     await this._httpClient.delete('/users/sign_out', config)
     this._resetState()
-    return null
+    return
   }
 }
 
--- a/packages/lib-auth/src/client.js
+++ b/packages/lib-auth/src/client.js
@@ -98,10 +98,8 @@ class Client {
       return ''
     }
     const { cookieName } = this._config
-    // via https://plainjs.com/javascript/utilities/set-cookie-get-cookie-and-delete-cookie-5/
-    const regexString = '(^|;) ?' + cookieName + '=([^;]*)(;|$)'
-    const targetCookie = document.cookie.match(regexString)
-    return targetCookie ? targetCookie[2] : ''
+    const cookies = cookie.parse(document.cookie)
+    return cookies[cookieName] || ''
   }
 
   /**
--- a/packages/lib-auth/src/client.js
+++ b/packages/lib-auth/src/client.js
@@ -1,0 +1,320 @@
+const cookie = require('cookie')
+const isNode = require('detect-node')
+
+const createImmutable = require('./helpers/create-immutable')
+
+/**
+ * The main Client class, which provides methods to authenticate against the API and request access tokens.
+ *
+ * @class
+ */
+class Client {
+  /**
+   * Creates a new instance of the client
+   *
+   * @param {object} config - Configuration parameters for the client
+   * @param {object} httpClient - A configured HTTP client for making requests
+   */
+  constructor (config, httpClient) {
+    createImmutable(this, '_config', config)
+    createImmutable(this, '_httpClient', httpClient)
+    this._resetState()
+  }
+
+  /**
+   * Makes a request to exchange a JWT for a access token.
+   *
+   * @async
+   * @private
+   * @param {string} jwt - A JSON Web Token
+   * @returns {Promise} Resolves to the new access token.
+   */
+  async _exchangeJWTForToken (jwt) {
+    if (!jwt) {
+      return null
+    }
+
+    const data = {
+      'client_id': this._config.clientAppID,
+      'grant_type': 'password'
+    }
+
+    const config = {
+      headers: {
+        cookie: `${this._config.cookieName}=${jwt}`
+      },
+      withCredentials: true
+    }
+
+    const response = await this._httpClient.post('/oauth/token', data, config)
+    this._handleNewTokenData(response.data)
+    return this._state.accessToken
+  }
+
+  /**
+   * Gets an access token. Returns the current access token if it's still valid, otherwise attempts to try and refresh it before returning.
+   *
+   * @async
+   * @returns {Promise} Resolves to the access token, or an empty string if not available.
+   */
+  async getAccessToken() {
+    if (!this._state.accessToken) {
+      return ''
+    } else if (this._isAccessTokenValid()) {
+      return this._state.accessToken
+    } else {
+      return this._refreshTokenData()
+    }
+  }
+
+  /**
+   * Requests a cross-site request forgery (CSRF) token, which is used in the main authentication-related actions.
+   *
+   * @async
+   * @private
+   * @returns {Promise} Resolves to the CSRF token.
+   */
+  async _getCSRFToken () {
+    try {
+      const url = `/users/sign_in?now=${Date.now()}`
+      const config = { withCredentials: true }
+      const response = await this._httpClient.head(url, config)
+      const csrfToken = response.headers['x-csrf-token']
+      return csrfToken
+    } catch (error) {
+      console.error(error)
+      throw error
+    }
+  }
+
+  /**
+   * Tries to get the Panoptes session cookie from the browser's cookie jar.
+   *
+   * @private
+   * @returns {string} The JWT from the browser cookie, or an empty string
+   */
+  _getJWTFromBrowser () {
+    if (isNode) {
+      return ''
+    }
+    const { cookieName } = this._config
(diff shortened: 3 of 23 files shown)


packages/lib-auth/src/helpers/validate-config.js
```diff
@@ -0,0 +1,34 @@
...
+
+const clientAppID = yup.string()
+  .matches(/^[a-z0-9]+$/i)
+  .required()
+
+const cookieName = yup.string()
+  .default('_Panoptes_session')
+
+const hostUrl = yup.string()
+  .url()
```
rogerhutchings: `localhost` is currently rejected, but should be valid as well

packages/lib-auth/src/client.js
```diff
@@ -0,0 +1,320 @@
...
+        'project_email_communication': user.projectEmailCommunication,
+        'project_id': user.projectId
+      }
+    }
+    const registerResponse = await this._httpClient.post('/users', data)
+
+    // After the user is created, we need to sign in to actually get an access
+    // token. `panoptes-javascript-client` manages to do it without this step;
+    // there's no `set-cookie` header on the response that I can see, so I'm not
+    // actually sure how.
```
srallen: My guess it that it doesn't get a token until the first API request is made because every API request is preceded by this check for a token: https://github.com/zooniverse/panoptes-javascript-client/blob/085a94a9b0f028a47db68fff013541b467707a8d/lib/api-client.js#L10-L13

packages/lib-auth/test/functional/node/isolation/test.js
```diff
@@ -0,0 +1,118 @@
...
+    })
+
+    after(function () {
+      nock.cleanAll()
+    })
+
+    it('should let you register as a new user', async function () {
+      const newToken = await client.register(fixtures.newUser)
+      expect(newToken).to.equal(fixtures.mockAccessToken)
+    })
```
srallen: We should test a failure case too. What is the expected behavior if the server returns a 400 or 500 range response?

packages/lib-auth/test/functional/node/isolation/test.js
```diff
@@ -0,0 +1,118 @@
...
+
+    after(function () {
+      nock.cleanAll()
+    })
+
+    it('should let you sign in as an existing user', async function () {
+      const newToken = await client.signIn(fixtures.existingUser)
+      expect(newToken).to.equal(fixtures.mockAccessToken)
+    })
```
srallen: Same question about what the expectation is for 400s and 500s. Does axios do any handling for you on 400s, 500s, like superagent does? http://visionmedia.github.io/superagent/#error-handling

packages/lib-auth/src/client.js
```diff
@@ -0,0 +1,320 @@
...
+    const csrfToken = await this._getCSRFToken()
+    const config = {
+      headers: {
+        'x-csrf-token': csrfToken
+      }
+    }
+    await this._httpClient.delete('/users/sign_out', config)
+    this._resetState()
+    return null
+  }
```
srallen: Does `signOut` need to return anything. In the specs, nothing is done with the return value of `signOut`. If there's a case where something should be done with it, then there should be a spec for that.

packages/lib-auth/test/functional/node/staging/test.js
```diff
@@ -0,0 +1,35 @@
...
+  describe('sign out', function () {
+    it('should let you sign out', async function () {
+      await client.signOut()
+    })
+  })
+
+  describe('sign in', function () {
+    it('should let you sign in as an existing user', async function () {
+      await client.signIn(fixtures.newUser)
+    })
```
srallen: Are these missing the expectations?

packages/lib-auth/test/unit/helpers/create-immutable.spec.js
```diff
@@ -0,0 +1,13 @@
+const createImmutable = require('../../../src/helpers/create-immutable')
+
+describe('Unit Tests > helpers > createImmutable', function () {
+  it('should set an property on a given object that can\'t be changed', function () {
```
srallen: I think this should read `should not set a property`

packages/lib-auth/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "axios": "~0.19.0",
+    "cookie": "~0.4.0",
+    "detect-node": "~2.0.4",
+    "yup": "~0.27.0"
+  },
+  "scripts": {
+    "docs": "jsdoc src",
+    "lint": "zoo-standard",
+    "lint:fix": "zoo-standard --fix",
+    "test": "mocha ./test/unit/playground.js",
```
srallen: Is this cruft since `playground.js` is added to the gitignore file?

packages/lib-auth/src/helpers/validate-config.js
eatyourgreens: `localhost` should definitely not work.

packages/lib-auth/src/client.js
```diff
@@ -0,0 +1,320 @@
...
+  async _getCSRFToken () {
+    try {
+      const url = `/users/sign_in?now=${Date.now()}`
+      const config = { withCredentials: true }
+      const response = await this._httpClient.head(url, config)
+      const csrfToken = response.headers['x-csrf-token']
+      return csrfToken
+    } catch (error) {
+      console.error(error)
+      throw error
```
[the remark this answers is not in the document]
eatyourgreens: I assume we want to examine the error in the console but not continue with authentication? I think this block should return null, or an empty string, rather than throwing.

packages/lib-auth/src/helpers/validate-config.js
rogerhutchings: I'd modified my `cors-config` in my local instance of Panoptes - you're right, it shouldn't work.

packages/lib-auth/src/client.js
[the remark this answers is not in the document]
rogerhutchings: I chose to re-throw here to get a helpful stack trace. Returning something falsey is an option too, but you'd need to add a check in the method calling this one and throw an error there so it can be handled appropriately.

packages/lib-auth/src/client.js
```diff
@@ -0,0 +1,320 @@
...
+   * @returns {string} The JWT from the browser cookie, or an empty string
+   */
+  _getJWTFromBrowser () {
+    if (isNode) {
+      return ''
+    }
+    const { cookieName } = this._config
+    // via https://plainjs.com/javascript/utilities/set-cookie-get-cookie-and-delete-cookie-5/
+    const regexString = '(^|;) ?' + cookieName + '=([^;]*)(;|$)'
+    const targetCookie = document.cookie.match(regexString)
```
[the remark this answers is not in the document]
rogerhutchings: Ugh, I _think_ there was but that was a Christmas ago 🎄 It probably can.

packages/lib-auth/test/functional/node/isolation/test.js
rogerhutchings: Not by default, but you can add it with an interceptor: https://github.com/axios/axios#interceptors

packages/lib-auth/src/client.js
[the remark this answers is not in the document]
eatyourgreens: I'm still not following why an error has to be thrown.

packages/lib-auth/README.md
```diff
@@ -0,0 +1,57 @@
+# @zooniverse/auth
+
+An authentication library for the Panoptes API which provides tokens for making authenticated requests.
```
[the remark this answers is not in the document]
eatyourgreens: I don't think Node supports service workers anyway. Do we expect to be authenticating in the browser at all, or is this purely server-side auth?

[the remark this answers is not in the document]
eatyourgreens: D'oh, I should have read the description of this library before writing that.

packages/lib-auth/src/client.js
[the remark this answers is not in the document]
eatyourgreens: The docs say this method returns a Promise that resolves to a token, so I think it will need to return something, even on error.

packages/lib-auth/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "main": "src/index.js",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "axios": "~0.19.0",
+    "cookie": "~0.4.0",
+    "detect-node": "~2.0.4",
+    "yup": "~0.27.0"
+  },
+  "scripts": {
+    "docs": "jsdoc src",
```
eatyourgreens: Where do the docs get published? Should they be included with this PR?

packages/lib-auth/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "lint": "zoo-standard",
+    "lint:fix": "zoo-standard --fix",
+    "test": "mocha ./test/unit/playground.js",
+    "test:functional:node": "mocha ./test/functional/node/isolation/test.js",
+    "test:functional:node:staging": "mocha ./test/functional/node/staging/test.js",
+    "test:unit": "mocha \"./test/unit/**/*.spec.js\""
+  },
```
eatyourgreens: I think this needs a `test:ci` script for Travis.

packages/lib-auth/README.md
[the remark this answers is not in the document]
rogerhutchings: Done. If it comes up, however, you could pass in an http client with an Axios-like interface to the client constructor - https://github.com/zooniverse/front-end-monorepo/pull/1380/files#diff-cdb35fbed372a2361d0992b4fb0953ccR18

packages/lib-auth/src/client.js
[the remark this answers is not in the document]
rogerhutchings: done

packages/lib-auth/src/client.js
rogerhutchings: Probably not, actually

packages/lib-auth/package.json
rogerhutchings: It is, yes

packages/lib-auth/package.json
rogerhutchings: They're not currently published - they could be, but I didn't have time to work out where the best place to do that is / how to automate it when I was finishing up. Have you got any ideas?

packages/lib-auth/test/functional/node/isolation/test.js
eatyourgreens: Anyone know,  off the top of their head, what the response is when registration fails (user already exists)? We should have a test for that response.

packages/lib-auth/test/functional/node/isolation/test.js
eatyourgreens: Similar to my previous comment, there should be a test for signIn when the username or password are wrong.

packages/lib-auth/test/functional/node/staging/test.js
eatyourgreens: Looks like it. This test should check the returned token?

packages/lib-auth/package.json
eatyourgreens: How about `docs/` in the package root? Maybe we generate them manually, for the time being, until we figure out automating them. We could perhaps generate them when pushing to GitHub.

eatyourgreens:
Should CI be running the functional tests or the unit tests? From a quick look at the code, the unit tests don't seem to be testing much except that methods exist. Also note that unit tests like that are brittle and make the client hard to refactor, particularly where they expect private methods to exist.

eatyourgreens:
@srallen are tests for error cases the only outstanding change requests that you have?

rogerhutchings:
@eatyourgreens  My plan was to have a browser + node set of functional tests that could be run against staging and in isolation, and a set of unit tests. I ran out of time at the end, unfortunately, while I was looking into the tooling for automating the browser tests.

Anyway, if there's stuff in there that shouldn't be just trash it

eatyourgreens:
Cheers! I just wanted to make sure I maybe wasn't missing something before I made major changes.

packages/lib-auth/test/functional/node/isolation/test.js
camallen: Looks like 422 if the built user resource can't be saved, https://github.com/zooniverse/Panoptes/blob/99857e1554608434c7696a26ea725d09f9b06474/app/controllers/registrations_controller.rb#L75

It might 500 if a db uniq index has a duplicate value on save.

eatyourgreens: Cheers! I'll add tests for both.

packages/lib-auth/package.json
eatyourgreens: [`jsdoc2md`](https://www.npmjs.com/package/jsdoc-to-markdown) looks useful since `jsdoc` produces HTML by default.

eatyourgreens: Voilà, [a generated README](https://github.com/zooniverse/front-end-monorepo/tree/fb511c23363b174e89092b6624cbfc02a0a0fbc5/packages/lib-auth/docs). If I install `jsdoc2md` globally, then we can use this to replace the handwritten Readme files in the classifier.

eatyourgreens:
Looking through the docs, I don't see client methods to request password reset links or to unsubscribe an email address. Should the client support those too, or is that something a consuming app can do once it has a token from this client?

I should note that I only have a sketchy idea how Panoptes auth works.

packages/lib-auth/package.json
eatyourgreens: With an appropriate ADR, of course.

srallen:
> I don't see client methods to request password reset links or to unsubscribe an email address. Should the client support those too, or is that something a consuming app can do once it has a token from this client?

I think the idea is to not have those be in the client, but in the consuming app, which there would only be one that all other apps link to for account management purposes. The other methods truly are reusable by other apps, so they go here. If there's a reason this is a bad idea, then I'm happy to hear it.

eatyourgreens:
They might need to use the `withCredentials` flag on the request, which I think is why we have them in the `auth` library at the moment. I could be completely wrong.

srallen:
It should be easy enough to add in to the request as needed (fetch API: https://github.com/github/fetch#sending-cookies)

camallen:
> They might need to use the `withCredentials` flag on the request, which I think is why we have them in the `auth` library at the moment. I could be completely wrong.

Password reset and unsubscribe both (currently) do not require auth. I can see unsubscribe possibly using a session cookie in the future but it was removed to be open a while back.

packages/lib-auth/src/client.js
```diff
@@ -0,0 +1,318 @@
...
+      const csrfToken = response.headers['x-csrf-token']
+      return csrfToken
+    } catch (error) {
+      console.error(error)
+      throw error
+    }
+  }
+
+  /**
+   * Tries to get the Panoptes session cookie from the browser's cookie jar.
```
eatyourgreens, line 91: Panoptes sessions use HttpOnly cookies so I’m not sure how this works.

packages/lib-auth/src/client.js
```diff
@@ -0,0 +1,318 @@
...
+  _resetState () {
+    this._state = {
+      accessToken: '',
+      accessTokenCreatedAt: NaN,
+      accessTokenExpiresAt: NaN,
+      accessTokenExpiresIn: NaN,
+      refreshToken: '',
+      scope: ''
+    }
+  }
```
eatyourgreens, line 247: Is there a better way of storing the tokens than making them publicly available in `client._state`?

packages/lib-auth/src/client.js
```diff
@@ -0,0 +1,318 @@
...
+   *
+   * @private
+   * @returns {string} The JWT from the browser cookie, or an empty string
+   */
+  _getJWTFromBrowser () {
+    if (isNode) {
+      return ''
+    }
+    const { cookieName } = this._config
+    const cookies = cookie.parse(document.cookie)
```
eatyourgreens, line 101: `document.cookie` isn’t available for HttpOnly cookies.

packages/lib-auth/src/client.js
```diff
@@ -0,0 +1,318 @@
...
+
+  /**
+   * Makes a request to exchange a JWT for a access token.
+   *
+   * @async
+   * @private
+   * @param {string} jwt - A JSON Web Token
+   * @returns {Promise} Resolves to the new access token.
+   */
+  async _exchangeJWTForToken (jwt) {
```
eatyourgreens, line 32: I’m confused. I thought the JWT that we use for auth is the token that’s returned from this method. JWT here seems to refer to the Panoptes session cookie.

packages/lib-auth/src/client.js
srallen: I believe Rog last worked with @camallen to get a development version of this working locally with Panoptes, so I'm mentioning @camallen to help answer your question.

camallen: When i was working with Rog was getting the browser to send the cookie to the server side node app. We got that working similar to how we solved the CFE single sign on. 

I assume this code was meant to run server side, though the comments suggest browser 🤔 - indeed our cookies are secure and http only client side in the browser.

eatyourgreens: This gets the cookie from `document.cookie` so it must be run in a browser and, yeah, we deliberately hide our session cookies from `document.cookie` so that they aren't visible to XSS attacks.

srallen: My understanding is this client is supposed to work server side and client side. There may be separate pathways for those use cases, but I'm not sure.

srallen: Looking over the proposal (noted in the PR description: https://github.com/zooniverse/front-end-monorepo/pull/1361/files) it's not explicitly stated, but I recall that being the stated goal, a working client for server side and client side, so that it can function for the long term goal of working server side auth and be a replacement of the old auth client for the SPAs we have. Happy to revisit this if this goal is too broad because working server side auth is more important for the long term goals of the monorepo, imo.

eatyourgreens: Server-side auth would work by getting the session secret from the incoming request cookies, then trading the session for a token that can be used as an Auth header. NextJS has an auth example for this, which I've found really useful.
https://github.com/vercel/next.js/blob/7c203b80c98d5b2f5a2af41d29f8032b1242e9c9/examples/with-cookie-auth-fauna/pages/api/profile.js#L10-L19

The docs for the `resumeSession` method mention getting the session cookie from the request, but that code isn't in this PR.
https://github.com/zooniverse/front-end-monorepo/blob/a7ebdfa2adb875a7e862290857a2440e804bb148/packages/lib-auth/src/client.js#L253

Client-side auth is still puzzling me, because the implementation here won't work (and shouldn't work because Panoptes sessions should not be accessible from JS.) This client seems, to me, like it would only work if you re-authenticate with your username and password on each request. I feel like I'm missing something obvious.

eatyourgreens:
Babel 7 supports the [proposed syntax for private class members](https://github.com/tc39/proposal-class-fields#private-fields), so we could use that to hide the client state, and internal methods, from public access. I'm very nervous about the access token, and refresh token, being publicly available in `client._state`.

I'm also unsure about the security implications of `client.getAccessToken()`. Could a malicious script instantiate a client in the browser, then run that method to get your token from your (secure) session cookie?

PR #1372, opened 2019-12-11
Drawing: Use the state tree to get a mark's tool

eatyourgreens:
Use `getParentOfType` to get the tool that created a mark.
Remove redundant tool props from mark components.
Pass marks to annotations by reference.
Clear stale marks from a tool if the current annotation is empty.
Remove `workflowSteps.tasks`, which was only being used to look up the tool for a given mark.
Add tests for rendered marks in the subject viewer.

Package:
lib-classifier

Closes #1363.

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #1381, opened 2019-12-16
ADR 22: Drawing tools

### LINKED ISSUE #1350: Documenting the drawing tools API
The DrawingTask folder in plugins should have a readme that describes the tool and task API. [ADR 15](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-15.md) should be updated with notes on how the actual implementation works compared to the decision.

eatyourgreens:
A description of the drawing tools and their public interfaces.

Package:
lib-classifier

Closes #1350.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/docs/arch/README.md
+++ b/docs/arch/README.md
@@ -21,3 +21,4 @@
 - [ADR 19: multiFrame Viewer](adr-19.md)
 - [ADR 20: Transcription Task](adr-20.md)
 - [ADR 21: Auth Client](adr-21.md)
+- [ADR 22: Drawing tools](adr-22.md)
--- a/docs/arch/ad-22.md
+++ b/docs/arch/ad-22.md
@@ -1,8 +1,12 @@
 # ADR 21: Drawing tools
 
+## Context
+
+With new drawing tools being developed for the classifier, we need an API that's common to all drawing tools and marks, which can be easily extended by tool developers. This document lays out an overview of the drawing tool model and the public interfaces common to all tools and all marks.
+
 ## The drawing model
 
-A drawing task has drawing tools. Each tool creates marks. On task completion, a drawing annotation is created, which is an array of all drawn marks.
+A drawing task has drawing tools. Each tool creates marks. On task completion, a drawing annotation is created, which is an array of all drawn marks. Each mark has a corresponding React component which renders the SVG for that particular shape.
 
 ## Filesystem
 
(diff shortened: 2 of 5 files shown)


eatyourgreens:
NB. some of the described behaviour is in open PRs, so subject to change.

srallen:
So, for example in ADR 15, in the decision we say that ` The interactive layer will have local component state to keep track of which mark is selected for editing as well as handle the edit and delete events.`, but the reality this is partly happening in the MST stores, so I just want to be sure we update the original ADR to record what was actually done. 

Another statement:`Have a component, the markings renderer, that manages the rendering of previous annotation marks as read only. It will hide them if hide previous marks is toggled.` I don't think there's a separate markings renderer either.

eatyourgreens:
The `DrawingToolMarks` component is the renderer. See #1372 where it’s used to render marks from previous annotations. 

The current mark is `activeMark` in the `InteractionLayer` state. There’s nothing in the models that indicates which mark is active.

eatyourgreens:
Individual mark data is in the stores, though. I’ll update that tomorrow.

srallen:
Ok, so, the point is that there's clearly confusion of how what we conceptually described matches up to what was actually done, so I think a summary would be helpful here. 

The `DrawingTooRoot` isn't just a read-only render of previous marks. As you say, the individual marks are stored in the MST stores, when the initial attempt with rx.js, there was a map of marks in the local component store.

eatyourgreens:
>when the initial attempt with rx.js, there was a map of marks in the local component store.

Oh yeah, that was in the `DrawingContainer`, which was deleted early on. I think that implementation might have misinterpreted what we were asking for in the ADR.

eatyourgreens:
>This is great documentations on the models. I think this is better as a readme in the root of the /plugins/drawingTools folder since this isn't recording a decision.

This is recording the architectural decisions I made (unilaterally, admittedly) on where to store drawing tool code and how drawing tools and marks should be modelled. ADR-15 doesn't say anything about drawing tool architecture.

srallen:
Ok, so if this is recording a decision, could you put this in the same format as the other ADRs with a `Decision`, `Status`, and `Consequences` sections? Also, the overall title of this ADR doesn't really distinguish this from ADR 15 since that is entitled `Drawing Tools API`. Maybe `Drawing Tool Store Models`?

eatyourgreens:
Good point. This one describes an API but 15 doesn’t.

eatyourgreens:
Maybe I should mark 15 as deprecated and this as superseding it? Then go into more detail about how drawing is implemented in the subject viewer. http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions

srallen:
That's a great idea! Thanks for pulling that up. I like `superseded` with a link to the new ADR.

eatyourgreens:
I should have read that blog post before writing the other ADR too.

eatyourgreens:
I've updated the headings and added an example. I still need to expand the description to include rendering and editing marks.

I've marked the status as 'proposed' because this document includes changes that are still in open PRs, so subject to change.

PR #1388, opened 2019-12-18
Add subject image rotation

eatyourgreens:
Add a `rotate()` transform to SVG subject images.
Add a `rotation` angle and `rotate()` action to the SubjectViewer store.
Enable the rotation button in the image toolbar.

TODO:
- [x] conditionally disable rotation for subject viewers that don't use it.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/ImageToolbar.js
+++ b/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/ImageToolbar.js
@@ -36,7 +36,7 @@ class ImageToolbar extends Component {
           <MoveButton />
           <ZoomInButton />
           <ZoomOutButton />
-          <RotateButton disabled />
+          <RotateButton />
           <FullscreenButton disabled />
           <ResetButton />
         </Box>
--- a/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/RotateButton/RotateButtonContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/RotateButton/RotateButtonContainer.js
@@ -6,10 +6,13 @@ import RotateButton from './RotateButton'
 
 function storeMapper (stores) {
   const {
-    rotate
+    rotate,
+    rotationEnabled
   } = stores.classifierStore.subjectViewer
 
+  const disabled = !rotationEnabled
   return {
+    disabled,
     rotate
   }
 }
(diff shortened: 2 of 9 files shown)


shaunanoordin:
The rotation feature looks good in most cases, but there one notable UI bug and one minor DOM issue to note. Documentation incoming!

shaunanoordin:
`transform-origin` IS the answer! `transform-origin: 50% 50%` on an SVG element will ensure the rotation occurs based on the centre of the Subject image.

```
<svg>
  <g id="transform-layer" style="transform-origin: 50% 50%" transform="rotate(-90)">
    <image xlink:href="..." />
    <g id="interaction-layer">...</g>
  </g>
</svg>
```

shaunanoordin:
Ah! I'm noting that the sibling PR #1388 (Pan and Zoom) actually DOES reset pan/zoom on a new Subject load, so that might solve the issue of the rotation in this PR not resetting when a new image comes in.

eatyourgreens:
PFE passes the rotation origin to the `rotate()` transform
https://github.com/zooniverse/Panoptes-Front-End/blob/538b01a1488690873ce6d87b23fbb4c22c653933/app/components/pan-zoom.jsx#L267

Apparently that's how the [SVG transform](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform) works. CSS transforms may or may not work.

eatyourgreens:
Moving the transform from the `svg` element to a child breaks drawing at the moment. The drawing layer inverts the SVG transform to get the SVG coordinates of a mouse click here:
https://github.com/zooniverse/front-end-monorepo/blob/6e686bf20dbf50f05490b98ace7bb0133d788e4e/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js#L30-L36

I _think_ that could be changed by getting the transformation matrix from the drawing layer itself, since that would return all transformations that have been applied to that element? Or maybe we'd have a transformation layer and get the screenCTM from that layer? Either way, we'd either pass a bunch of refs around as props, or set up a context that passes the transformation down to children so that they can get the current matrix.

I think that might be out of scope of this PR.

eatyourgreens:
#1393 sets up an SVG context, so that we can pass the transformation context down separately from the parent SVG element.

shaunanoordin:
> #1393 sets up an SVG context, so that we can pass the transformation context down separately from the parent SVG element.

👍 Cool! I'll re-review this PR with a note that any SVG transformation magic will be taken care of in 1393. (IIRC, the functionality was pretty much on point the last time I took a look at the code, but I'll re-run my checks to confirm.)

> EDIT: also, why is a new SVGPoint named SVGEvent in that code snippet? I've only just noticed that.

Uuuhhh... I have no idea. It's definitely supposed to be [svgPoint](https://developer.mozilla.org/en-US/docs/Web/API/SVGPoint), but both `InteractionLayer.js` and `draggable.js` use `const svgEvent = svg.createSVGPoint()`

I see you replaced that in PR 1393 though, so that's good 👌

PR #1389, opened 2019-12-18
Subject image pan and zoom

### LINKED ISSUE #1208: Image subject viewer zoom and pan
**Package** lib-classifier

The image subject viewers need to be zoomable and pan-able. We could consider using d3 or vx to implement zoom if the image itself is an svg `image`. #1094 needs resolving first.

d3 zoom: https://github.com/d3/d3-zoom

vx (d3-like zoom plus declarative react components): 
https://github.com/hshoff/vx/tree/master/packages/vx-zoom
https://vx-demo.now.sh/zoom-i

**Additional context**

eatyourgreens:
Add pan and zoom handlers to the single image viewer.
Add a dynamic viewbox to the single image viewer.
Make subject images draggable and focusable and add a keyDown handler.
Add up and down key handlers to keyboard zoom.
Add zoom on wheel scroll.
Add continuous zoom on long press of the toolbar buttons.

Package:
lib-classifier

Closes #1208.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
A working demo of pan and zoom for SVG images. To test this, go to I Fancy Cats (or any project with image subjects), click the image toolbar to focus it then use the pan/zoom keys to pan and zoom (+/- for zoom, cursor keys for panning.)

eatyourgreens:
On drawing projects (eg. http://localhost:8080/?project=908&workflow=3370) you can use the keyboard controls directly from drawn points, which allows us to support the pan/zoom that's in use for Seabird Watch and Penguin Watch (pan + zoom from the keyboard while also drawing with a pointer or mouse.)

eatyourgreens:
This now has mouse wheel support (with tests) and continuous zoom on long press of the toolbar buttons.

packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/ZoomInButton/ZoomInButtonContainer.js
```diff
@@ -17,10 +17,31 @@ function storeMapper (stores) {
...
+    this.timer = ''
+  }
+
+  onPointerDown (event) {
     const { zoomIn } = this.props
+    this.timer = setInterval(zoomIn, 100)
+  }
+
+  onPointerUp (event) {
+    clearInterval(this.timer)
```
shaunanoordin: Yoiks, this is a bit problematic - if I click down on the Zoom In button, then move the mouse cursor out of the button BEFORE lifting the mouse click, then the zoom in action will trigger _ad infinitum._

If we're using the `intervals/timer` solution for the Zoom In Button (I'm not sure if this comment on Slack - _"I figured out how to get a Grommet Button to fire continuously on long press: wrap it in a container that listens to pointerdown"_ - meant that you already have a different solution in mind) - then the `clearInterval()` also needs to trigger at  `onPointerOut`, `onPointerLeave`, and/or `onPointerCancel`.

Also, if you want to be as paranoid as me, it's good to call `clearInterval(this.timer)` before `this.timer = setInterval(zoomIn, 100)` in `onPointerDown`, so it's impossible to create multiple active timers.

packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/ZoomOutButton/ZoomOutButtonContainer.js
```diff
@@ -19,16 +19,29 @@ function storeMapper (stores) {
...
 
-  onClick () {
-    this.props.zoomOut()
+  onPointerDown (event) {
+    const { zoomOut } = this.props
+    this.timer = setInterval(zoomOut, 100)
+  }
+
+  onPointerUp (event) {
+    clearInterval(this.timer)
```
shaunanoordin: My comments on the `onPointerUp` of the `ZoomInButtonContainer` also applies here.

It's actually possible to create a never-ending Zoom In timer AND a never-ending Zoom Out timer. The resulting conflict causes the Subject image to jitter back and forth between different sizes at epileptic speeds, which makes for a fun, wholesome evening of viewing entertainment for the family. (My Netflix is down, you see.)

packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/ZoomInButton/ZoomInButtonContainer.js
```diff
@@ -17,10 +17,31 @@ function storeMapper (stores) {
...
+  constructor () {
+    super()
+    this.onPointerDown = this.onPointerDown.bind(this)
+    this.onPointerUp = this.onPointerUp.bind(this)
+    this.timer = ''
+  }
+
+  onPointerDown (event) {
     const { zoomIn } = this.props
+    this.timer = setInterval(zoomIn, 100)
```
shaunanoordin: It's possible for really fast clicks on the Zoom In button to not register a _zoom in action._

This is because, if the `onPointerUp () { clearInterval(this.timer) }` occurs in less than 100ms, then the `zoomIn()` action isn't even triggered once.

One solution would be...

```
zoomIn()
this.timer = setInterval(zoomIn, 100)
```

(On my MacBook, tapping quickly on the trackpad can be done in less than 100ms, so no zoom in action occurs. Physically clicking down on the trackpad is usually done in more than 100ms, so in most cases, a zoom in action occurs.)

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
```diff
@@ -30,6 +63,81 @@ class SingleImageViewerContainer extends React.Component {
...
+  dragMove (event, difference) {
+    this.setState(prevState => {
+      const { viewBox } = Object.assign({}, prevState)
+      viewBox.x -= difference.x / 1.5
+      viewBox.y -= difference.y / 1.5
+      return { viewBox }
+    })
+  }
+
+  onPan (dx, dy) {
```
shaunanoordin: I'm making a note that the "pan view" and "zoom" actions function by modifying the `<svg>`'s `viewBox` attribute, instead of the `transform` attribute. This is fine in the context of getting this PR to work, but not standard with, say, the ["rotate view" PR](https://github.com/zooniverse/front-end-monorepo/pull/1388), which uses `transforms`.

Let me get to my full review, and I'll make a note whether it's worth discussing these different approaches.

One pertinent question though is, where did the `10` come from in `viewBox.x += dx * 10` ?

eatyourgreens:
My understanding is `transform` would move the image inside the SVG, breaking drawing point coordinates. 

Pointer capture will fix the pointer up handler.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
```diff
@@ -30,6 +63,81 @@ class SingleImageViewerContainer extends React.Component {
     })
   }
 
+  dragMove (event, difference) {
+    this.setState(prevState => {
+      const { viewBox } = Object.assign({}, prevState)
+      viewBox.x -= difference.x / 1.5
```
shaunanoordin: Where's the `1.5` coming from, by the way? I noticed that the drag-move action doesn't move the image as much as the mouse action.

For example, if I start click-dragging at the right-edge of an image and drag all the way left (100% the length of the SVG width), the Subject image only moves left a distance of 50%  of the width of the SVG.

shaunanoordin: Note: this was at natural zoom (Zoom In/Out wasn't clicked after Subject was loaded)

shaunanoordin: Also, I'm not going to investigate touch events too much - there's still a bit of wonkiness with touch-drag causing page scroll.

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/withKeyZoom.js
```diff
@@ -37,13 +37,25 @@ function withKeyZoom (WrappedComponent) {
           return true
         }
         case 'ArrowRight': {
```
shaunanoordin: Arrow navigation mostly works, with the following caveats:

- Up ⬆️ key moves image down ⬇️, and down ⬇️ key moves image up ⬆️... while the ⬅️ left key moves the image ⬅️ left, right ➡️ key moves the image ➡️ right
  -  Basically, there seems to be a mismatch between how the ↕️  vertical and ↔️  horizontal panning is mapped to keys.
- The pan via keys doesn't account for zoom, so a zoomed-out image will move at visually small increments of length, compared to a zoomed-in image. 
  - (I can't tell if zoom makes a difference with panning via mouse drag, though.)
- I'm not sure if the pan accounts for rotation either, so a rotated image might easily 
  - ...but this is something that should only be solved after the PR #1389 (Rotate View) is merged.

eatyourgreens:
The 1.5 is an arbitrary smoothing factor to stop jerky motion on pointer move. All the other numbers are made up too. 

I should check with PFE to see which direction the arrow keys should move the image in. 

This also needs sensible limits set for pan and zoom.

packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/ZoomInButton/ZoomInButtonContainer.js
eatyourgreens: That's a good point about starting multiple timers. I've added pointer capture to solve the problem of releasing the pointer outside the button.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
eatyourgreens: The `1.5` was an arbitrary smoothing factor so that we don't get jerky motion when dragging.

That's interesting about the page scroll. We've fixed that in PFE so it should not be re-introduced here.

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/withKeyZoom.js
eatyourgreens: I've updated the left-right pan so that it matches [Seabird Watch](https://www.zooniverse.org/projects/penguintom79/seabirdwatch/classify). You can test it out on http://localhost:8080/?project=478&env=production and compare with the live project.

eatyourgreens: I'm not sure if we should scale motion, since that would make it hard to move the image by small amounts at high magnifications. I don't think we scale motion in PFE.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
eatyourgreens: The `10` was an arbitrary choice. We can tune it against real subjects for live projects.

eatyourgreens:
>My (mild) concern is that this implementation is different from the sibling PR #1388, which deals with rotation, and uses transform=rotate(X) to change the view.

By the way, this is the implementation used in PFE, and various older Zooniverse projects before that. `translate` and `scale` are worth experimenting with, though.

PR #1390, opened 2019-12-18
Drawing annotation model ADR

srallen:

Package:

Related to #500 and #823

Describe your changes:
This is a proposal ADR for drawing tool mark annotation models. This does not include a proposal for sub-tasks as I think that deserves a separate ADR to keep this scoped. I've left the number off since we have a few ADR PRs open right now. That will be updated when it can be.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
@lcjohnso and @vrooje I've invited you to review this proposed change to drawing tool mark models from a researcher's perspective. Please let us know if you have any comments or concerns. Thanks!

docs/arch/adr-xx.md
```diff
@@ -0,0 +1,65 @@
...
+  - The tools that have defaults are ellipse, rotate rectangle, fan. 
+  - A proposed example for the ellipse may involve requiring the volunters to click twice to set the the two radii. ([comment](https://github.com/zooniverse/front-end-monorepo/issues/500#issuecomment-516788821))
+- The freehand drawing tools mark annotation will be a string of the SVG's path and it will be the responsibility of post-classification analysis to convert this to usable x,y points. We will include a sample script in the `DataDigging` repo for project owners to reference on how to do this. Aggregation in Caesar will have to be updated to do the conversion first.
+- An added `toolType` string name definition will be included. This would be `point`, `ellipse`, etc. This enables the aggregation for Caesar code to auto-configure which extractor to use without checking data types ([comment](https://github.com/zooniverse/front-end-monorepo/issues/823#issuecomment-493896524)).
+  - `tool` will change to `toolIndex` to clarify it is referring to the index of the input from the task area.
+  - The parent drawing task will have a `taskType` added as well.
```
srallen: Just a point of clarification here, tasks and drawing tools both have `type` on their models, so I'm proposing a more explicit naming here. This may be unnecessary perhaps. I'm open to discussion on it.

eatyourgreens: How would this differ from the existing types?

docs/arch/adr-xx.md
```diff
@@ -0,0 +1,65 @@
...
+## Decision
+
+The shape's mark annotation models should change for consistency and improved post-classification analysis in the following ways:
+
+- The annotation should use the mathematical standard of _RHC_ with a domain of `(-180, 180]` or `[0, 360)` for consistent angle calculation
+- The annotation model should use `angle` for naming
+- The annotation model should consistency use `x_center` and `y_center` for shapes where that is applicable
+  - The exceptions are point, line, fan. The fan tool is not symmetric. 
+  - Shape clustering in aggregation is always done with the center
+- All rotations should be defined about `x_center` and `y_center` point. If the rotation cannot be defined around the center point, then the point used that should be clearly recorded in the annotation as `x_rotation` and `y_rotation`
```
eatyourgreens: From a developer's point of view, I'd prefer not to have logic in the code along the lines of `if x elseif x_center elseif x_rotation`. Here, for example: https://github.com/zooniverse/front-end-monorepo/blob/df5b630209dc65e658865b2f75a2657df490241e/packages/lib-classifier/src/plugins/drawingTools/components/DrawingToolRoot/DrawingToolRoot.js#L59-L61

docs/arch/adr-xx.md
```diff
@@ -0,0 +1,65 @@
...
+- Default values should be removed wherever possible. This may result in some tools drawing UX changing. 
+  - The tools that have defaults are ellipse, rotate rectangle, fan. 
+  - A proposed example for the ellipse may involve requiring the volunters to click twice to set the the two radii. ([comment](https://github.com/zooniverse/front-end-monorepo/issues/500#issuecomment-516788821))
+- The freehand drawing tools mark annotation will be a string of the SVG's path and it will be the responsibility of post-classification analysis to convert this to usable x,y points. We will include a sample script in the `DataDigging` repo for project owners to reference on how to do this. Aggregation in Caesar will have to be updated to do the conversion first.
+- An added `toolType` string name definition will be included. This would be `point`, `ellipse`, etc. This enables the aggregation for Caesar code to auto-configure which extractor to use without checking data types ([comment](https://github.com/zooniverse/front-end-monorepo/issues/823#issuecomment-493896524)).
+  - `tool` will change to `toolIndex` to clarify it is referring to the index of the input from the task area.
```
eatyourgreens: From memory, `tool` is only ever used to look up the type of tool in `task.tools` so `toolType` would remove it completely.

docs/arch/adr-xx.md
```diff
@@ -0,0 +1,65 @@
...
+Some of the mark annotation models have a few other issues as well:
+
+- Some shapes have default values which an create bias. For example, the ellipse has a default axis ratio of 0.5 and many volunteers have left the default creating a bias ([comment](https://github.com/zooniverse/front-end-monorepo/issues/500#issuecomment-516788821))
+- The freehand drawing tool has peformance impact on the browser as the drawing is being created and with the job to create classification exports as well. This is because the current annotation consists of every single x, y point created
+
+## Decision
+
+The shape's mark annotation models should change for consistency and improved post-classification analysis in the following ways:
+
+- The annotation should use the mathematical standard of _RHC_ with a domain of `(-180, 180]` or `[0, 360)` for consistent angle calculation
```
eatyourgreens: This ADR should pick one of those two domains and recommend it.

docs/arch/adr-xx.md
```diff
@@ -0,0 +1,65 @@
...
+- The annotation model should use `angle` for naming
```
eatyourgreens: For naming rotation angles?

docs/arch/adr-xx.md
```diff
@@ -0,0 +1,65 @@
...
+- The annotation model should consistency use `x_center` and `y_center` for shapes where that is applicable
```
eatyourgreens: Same comment as below about avoiding conditional logic in the drawing tool root.

docs/arch/adr-xx.md
eatyourgreens: In #1372 `mark.tool` is a reference to the tool model that created this mark, so 👍 from me to this change.

docs/arch/adr-xx.md
```diff
@@ -0,0 +1,65 @@
...
+- An added `toolType` string name definition will be included. This would be `point`, `ellipse`, etc. This enables the aggregation for Caesar code to auto-configure which extractor to use without checking data types ([comment](https://github.com/zooniverse/front-end-monorepo/issues/823#issuecomment-493896524)).
+  - `tool` will change to `toolIndex` to clarify it is referring to the index of the input from the task area.
+  - The parent drawing task will have a `taskType` added as well.
+
+Each Github issue tracking the port of each drawing tool will be updated with the specific changes relevant to that tool's mark annotation model. Certain models may have additional properties for use within the front-end application, like a unique id for use in maps, that will be removed from the annotation before submitting with the classification. These additional properties are internal for the application and may confuse project owners if included.
```
eatyourgreens: If we flatten subtask annotations in the classification, then we'll need to link each annotation back to the mark that created it. I've proposed in #500 that we use `mark.id` to link marks to annotations (if we don't use `mark.annotations` to establish the relationship.)

docs/arch/adr-xx.md
srallen: I'm just proposing a rename of the existing `type` on both the task model and the tool model, though I didn't write that here clearly, but thinking about this again I've decided its unnecessary, so I'm going to remove this.

docs/arch/adr-xx.md
shaunanoordin: I'm personally leaning towards `(180, 180]` but that may just be my preferences being warped by JavaScript.

shaunanoordin:
The ADR generally looks good to me 👍 - I only have very little to add to this discussion, and _none_ of them _directly_ affect this PR.

My functionally minor but wordy tangents:

- once we sort out the shape of the annotations, I feel it'd be nice to spell out some explicit examples. (Probably per tool type. And probably plonk them  in the Labs page or something.)
  ```
  // Yo, yo, if you make a Triangle tool, then a single annotation mark would look like...
  annotations: [
    { toolIndex: 0,  // Check your Workflow to confirm that your Triangle tool is correctly at this index location
      x_center: 10.123, // Centre of triangle, any numeric value
      y_center: 22.456, // Centre of triangle, any numeric value
      angle: 37.1,  // Degrees, counter-clockwise, from -180 to 180
      ... etc ...
    },
  ]
  ```
- two reasons: 1. I don't think researchers/project owners will read this ADR, and 2. even when I was trying to figure what was the "correct shape" of any given annotation, I was confused; I had to submit Classifications on PFE to figure out the baseline.
- I made a note of the _"shape of PFE annotations look different from monorepo annotations"_ issue in a previous review - https://github.com/zooniverse/front-end-monorepo/pull/1339#pullrequestreview-325327427 - but I think this ADR already covers my main concerns.
  - this ADR already makes notes of the use of `toolIndex` vs `tool`, and that certain front-end-only values such as `annotation.id` need to be stripped before being submitted to the API 👍 
  - this PR already made a note that sub-tasks will be covered in a separate ADR, so I'll talk about the inconsistency of PFE using `details: []` and monorepo using `annotations: {}` in that separate PR.

docs/arch/adr-xx.md
CKrawczyk: It is possible to have multiple instances of the same `toolType` within a drawing task, so `toolIndex` will still be needed.

Example: there are two point tools, one red, the other blue.  These will both have the same `toolType` (point), but should *not* be aggregated together.  Penguin Watch and Seabird Watch are both set up like this.

eatyourgreens: Good point. In the code we have to identify a tool by type and colour, because tools don't have identifiers.
https://github.com/zooniverse/front-end-monorepo/blob/6e686bf20dbf50f05490b98ace7bb0133d788e4e/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js#L79-L80

eatyourgreens: One question this raised for me: a project owner can change the order of drawing tools, or even remove some, in the project builder. Is this a rare enough event that we can rely on the array order to identify each tool?

docs/arch/adr-xx.md
srallen: `x_center` or `x_rotation` will be replacing the non-specific `x` and same for `y`. I think this is easily avoided using the a computed view function on the model. I'll update the description here to include this suggestion.

docs/arch/adr-xx.md
srallen: We can discuss this in the follow up ADR about sub-tasks.

srallen: I'll rewrite this section to remove the example about id though.

srallen:
@shaunanoordin I agree we should have a specific example for each tool and I'll plan on updating each related issue once this ADR proposal has been agreed to and merged. I'll also open an issue for ADR 7 where we proposed including a json schema with each annotation.

srallen:
@CKrawczyk and @eatyourgreens I think I've addressed your comments and this is ready for review again.

docs/arch/adr-24.md
```diff
@@ -0,0 +1,64 @@
...
+  - A proposed example for the ellipse may involve requiring the volunteers to click twice to set the the two radii. ([comment](https://github.com/zooniverse/front-end-monorepo/issues/500#issuecomment-516788821))
+- The freehand drawing tools mark annotation will be a string of the SVG's path and it will be the responsibility of post-classification analysis to convert this to usable x,y points. We will include a sample script in the `DataDigging` repo for project owners to reference on how to do this. Aggregation in Caesar will have to be updated to do the conversion first.
+- `tool` will change to `toolIndex` to clarify it is referring to the index of the input from the task area. The `toolIndex` is useful to distinguish between multiple instances of the same tool tip in a drawing task.
+- Tasks and drawing tools will have a `type` attribute that map to an enumeration of the type of task or tool like `drawing` or `point`, `ellipse`, etc respectively. This enables the aggregation for Caesar code to auto-configure which extractor to use without checking data types ([comment](https://github.com/zooniverse/front-end-monorepo/issues/823#issuecomment-493896524)).
```
srallen: Just noting that the models already have `type` defined on them, but I wanted to include that here as having a purpose and this is new to the models compared to PFE.

docs/arch/adr-24.md
```diff
@@ -0,0 +1,64 @@
...
+Some of the mark annotation models have a few other issues as well:
+
+- Some shapes have default values which an create bias. For example, the ellipse has a default axis ratio of 0.5 and many volunteers have left the default creating a bias ([comment](https://github.com/zooniverse/front-end-monorepo/issues/500#issuecomment-516788821))
+- The freehand drawing tool has peformance impact on the browser as the drawing is being created and with the job to create classification exports as well. This is because the current annotation consists of every single x, y point created
+
+## Decision
+
+The shape's mark annotation models should change for consistency and improved post-classification analysis in the following ways:
+
+- The annotation should use the mathematical standard of _RHC_ with a domain of `(-180, 180]` for consistent angle calculation
```
eatyourgreens:
```suggestion
- The annotation should use the mathematical standard of _RHC_ with a domain of `[-180, 180]` for consistent angle calculation
```

srallen:
One of the suggested changes is removing default values wherever possible because of the bias those introduce. This could mean that we remove the minimum sizes for any shape as well. We (unintentionally?) did this with the brush annotation for the light curve viewer for TESS and some volunteers complained about it. I don't think these complaints override the need to remove bias however. Are there other impacts to removing minimum sizes to shapes that I might not be thinking of?

docs/arch/adr-24.md
```diff
@@ -0,0 +1,64 @@
...
+- The annotation model should use `angle` for naming rotation angles. This replaces usage of `rotation`.
+- The annotation model should replace `x` and `y` with `x_center` and `y_center` for shapes where that is applicable
+  - The exceptions are non-shapes like point, line, and transcription line tools, and non-symmetric shapes like fan.
```
eatyourgreens: I don't think polylines or freehand lines have a centre, or a centre of rotation.

docs/arch/adr-24.md
```diff
@@ -0,0 +1,64 @@
...
+  - Shape clustering in aggregation is always done with the center
+- All rotations should be defined about `x_center` and `y_center` point. If the rotation cannot be defined around the center point, then the point used should be clearly recorded in the annotation as `x_rotation` and `y_rotation`
```
eatyourgreens: Having `{ x, y }`, `{ x_center, y_center }` and `{ x_rotation, y_rotation }` still makes me uneasy in terms of how we'll fit that into the SVG rendering code.

docs/arch/adr-24.md
```diff
@@ -0,0 +1,64 @@
...
+  - Conditional logic in component code can be avoided by using a Mobx-state-tree's computed view functions to get either `x_center` or `x_rotation` mapped to `mark.x`.
```
eatyourgreens: Can you add code here to demonstrate how you envisage this working?

eatyourgreens:
https://github.com/zooniverse/Panoptes-Front-End/issues/5604 has examples of default values biassing classifications for the fan tool.

docs/arch/adr-24.md
srallen: This isn't an exhaustive list, just examples.

docs/arch/adr-24.md
srallen: @CKrawczyk you recommended `(-180, 180]` in your original comment in #500, but in a later comment suggested `[180, -180]` (see: https://github.com/zooniverse/front-end-monorepo/issues/500#issuecomment-563251422). Could you confirm if `-180` should be inclusive or exclusive?

docs/arch/adr-24.md
srallen: Line 43 suggesting we leverage MST's computed views addresses this concern. I'll add an example as asked.

docs/arch/adr-24.md
srallen: Sure

docs/arch/adr-xx.md
srallen: I think they could use workflow versions in later analysis to confirm which tool maps to which array index?

docs/arch/adr-24.md
CKrawczyk: Sorry for the confusion, taking a look at the `atan2` implementation in JS (same as python and C) it uses `[-180, 180]`, where the only way to get `-180` is for the "delta x" to be `-inf`.  See the examples section of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2 for the edge cases around -180 and 180.

srallen: Thanks for the confirmation.

docs/arch/adr-24.md
srallen: I've updated to include a code example. Does this address your concern? If not, could you be specific about what your concerns are?

docs/arch/adr-24.md
CKrawczyk: Looking at JS implementation of `Math.atan2` (same as that used in Python and C) the range is `[-180, 180]`, with the value of `-180` only being used with "delta x" is `-inf` (something that can't happen with our drawing tools).

See the examples section of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/atan2 for the edge case values around the `-180` `180` branch cut.

srallen:
I've included the suggestion to remove all default values wherever possible (see: https://github.com/zooniverse/front-end-monorepo/pull/1390/files#diff-91a231f67f4535e98da1eab296e0828fR83-R85). The current PFE PR to update the fan tool makes the spread configurable (see: https://github.com/zooniverse/Panoptes-Front-End/pull/5605). Should we allow biasing default values to be configurable? Would project builders actually configure these? My instinct is telling me most wouldn't bother and that we would still have a biasing problem. Perhaps if the default values were so small as to make the tools nearly unusable might work, but at that point, I think perhaps we should reconsider the interaction entirely and have the volunteer set these values with clicks. The trade off is that this will slow down drawing. I'd really like to have a researcher's perspective on this. @CKrawczyk, @vrooje, or @lcjohnso do you have thoughts on this?

docs/arch/adr-24.md
eatyourgreens: It does. I don't see why we have `{ x, y }`, { x_center, y_center }` and `{ x_rotation, y_rotation }` if they are mutually exclusive. It only makes the models clumsy and harder to maintain.

srallen: For downstream analysis. @CKrawczyk asked for explicit naming for what the point represents. I don't agree that it is clumsy as developers will have to think specifically about the shape of the tool (is it symmetrical for instance) when building a new one as well.

srallen: Also to be clear, we won't have a `{ x, y }` in any of the shapes anymore. They almost all will be replaced with `{ x_center, y_center }` with the exception of shapes that have a point of rotation elsewhere like the fan tool.

srallen: Line 39 mentions the replacement in the models.

eatyourgreens: I could very well be wrong but I read @CKrawczyk's comments as asking for `{ x_center, y_center }` in addition to `{ x, y }`, not in place of. Likewise, recording the centre of rotation _and_ the centre of the shape when they're different.

srallen: The original request in #500 included: 

> x and y should be consistently defined as the center of a shape (with exception of the fan tool as this is not symmetric). All shape clustering needs to be done with the center anyways, so this saves a set when analyzing the data.
> x and y could be renamed x_center and y_center to make this clear to the science teams

@CKrawczyk could you confirm that you would like `{ x, y }` to be renamed as `{ x_center, y_center }` where applicable? For the fan tool, which is the only non-symmetrical shape we have I think, do you want `{ x, y  }` to be replaced by `{ x_rotation, y_rotation }` and have the center be included separately, too? If so, would that be true for other non-symmetrical shapes?

eatyourgreens: I think the only shape that affects is the rectangle, where we'd change `{ x, y }` to record the centre. I think it's currently recording the top left corner, which isn't very useful. If we always use `{ x, y }` for clustering then it makes sense to me to name them consistently, however we name them. What I'd like to avoid is each shape having its own rules for naming the same thing (ie. the point used for clustering.)

srallen:
Another option would be to support configurability specifically for shapes for certain shapes that need it, have no default value, and if none are set, then just display an error in the classifier which would indeed make the project builder think about what these should be for their project. Out of the possible options, I think I'm liking this one best. We still may want to have min and max values for possible configurations.

CKrawczyk:
I think the issue of bias is more important than have a fast drawing task.  The volunteers can learn how to use the tools, and in many cases you can't remove the bias from the results.  You end up asking questions like "is the aspect ratio 0.5 because it was good enough, did the volunteer not know how to adjust the value, or did they adjust it and finally settle on 0.5?"

If defaults are made configurable in the project builder I think it is important that there is text explaining why it configurable, a warning of the bias it can cause, and the advice to *not* set the default to a typical value within the data set as volunteers would be less likely to adjust away from it.

docs/arch/adr-24.md
srallen: Actually, I confused myself about this. Task models have `type`, but annotations do not, so this will need to be added.

eatyourgreens:
>I think the issue of bias is more important than have a fast drawing task. The volunteers can learn how to use the tools, and in many cases you can't remove the bias from the results.

Point taken but the issue here is that the ellipse and fan tools crash the SVG renderer with `NaN` errors if we don't start with an initial guess at a shape.

eatyourgreens:
See the conversation on https://github.com/zooniverse/Panoptes-Front-End/issues/5604 for discussion of configuring the fan tool.

docs/arch/adr-xx.md
eatyourgreens: Yeah, I think workflow version plus array index should work as a tool identifier. Array indexes on their own as identifiers still makes me nervous.

docs/arch/adr-24.md
eatyourgreens: Annotation type could be useful if it saves a task lookup to get the type.

I thought marks had a type too, but I don't see it in [the README](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/plugins/drawingTools#mark-models).

This ADR is specifically for the drawing task and drawing annotation but it sounds like it is starting to recommend changes for other task and annotation models?

srallen: This was my bad and confusion on my part. We recorded the purpose of the `type` for use with the MST stores in past ADRs, but not that we wanted to include it in the annotations. I asked in an RFC and @CKrawczyk said it would be useful, but it didn't end up in the past ADR. I'm happy to write up a separate one on general annotation format changes.

srallen:
Ok, so as an interim solution, would configurable parameters in the project builder for the shapes impacted **without** a suggested default from us, an error if it is isn't set, and when it is set have clear language about the bias effect be good enough? Changing the UX of the drawing to require the volunteers to set these parameters each time when they drawing is possible, but I don't think we know how to do this (do we prompt them to input a value or use clicks or ...?) and I would want to ask @beckyrother to do some research into it before implementing it. I'm not sure we have time for this unfortunately.

docs/arch/adr-24.md
eatyourgreens: Sounds good to me. I could have sworn that marks, at least, have a type already.

docs/arch/adr-24.md
srallen: Because non-symmetrical shapes don't use the center of the shape as the point of rotation, I think @CKrawczyk is suggesting we use a naming convention that clearly distinguishes this. This does introduce some logic to manage in downstream, but it sounds like knowing absolutely what the point represents is more important?

CKrawczyk:
> Ok, so as an interim solution, would configurable parameters in the project builder for the shapes impacted **without** a suggested default from us, an error if it is isn't set, and when it is set have clear language about the bias effect be good enough?

This sounds like the best way forward at the moment.

docs/arch/adr-24.md
srallen: I did too! But I double checked again just now and tasks and tools do have types, but annotations and marks do not.

srallen: I did another pass on this statement to clarify even more what the proposed change is.

srallen:
@CKrawczyk I updated the text to describe the proposed change that will affect the lab in configuring the tools and the classifier to display an error. Requiring the project owner to input a value means they'll hopefully become aware of the biasing effect and can at least experiment before launching for a value that ideally has minimal impact. I included a statement in the consequences about future UX testing to hopefully eliminate this entirely someday.

srallen:
I'm going to go ahead and review since @lcjohnso is still out and leave this ADR as a proposal so we can refine as needed during application of the proposal.

lcjohnso:
My comments on this ADR, for completeness:

1) Repeating [my comment on ADR 25](https://github.com/zooniverse/front-end-monorepo/pull/1397#pullrequestreview-344119539): My primary (meta)concern: how breaking changes are handled and communicated to project teams. This topic is more appropriately handled in its own ADR, but including a new classifier version metadata flag should only be the beginning of how new data formatting consequences are passed along to teams. Fine to defer these questions to another ADR, but standards for how breaking changes are communicated to project teams should be made a priority.

2) Re: default values for ellipse, rotate rectangle, and fan tools -- I'm happy with where this ended up after the conversation above.  I was not originally in favor of removing defaults and making tools harder to configure in an effort to drive home the point that biases are an issue.  However, as long as there are clear error messages in case that tools are missing their originally-blank default values, and the scope of this change in behavior is limited to three drawing tools, then this current decision is acceptable.  But still necessary to acknowledge that unless there is a default-less way of placing these three drawing markers, the potential for bias will persist.

srallen:
@lcjohnso thanks for the feedback. I agree that we will need to have a strategy on communication soon. We do have [ADR 07](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-07.md) which is about how to programmatically communicate changes for classification task annotation models by way of JSON schema definitions. We don't have a holistic plan on communication, however. It'd be good to liaison with you and others on how best to do this.

The decision on the default values is not ideal, but I think an adequate compromise for the time being. If we change to zero default and have UX strategies for those tools, then that's when we can version bump the classifier again.

Issue #1315, opened 2019-11-20
Drawing sub-task UI


srallen:
**Package** lib-classifier

The current PFE implementation is like a tool-tip that contains the tasks to be completed fixed to the position of the drawn mark. This can be seen in the dev classifier using the ellipse: https://master.pfe-preview.zooniverse.org/dev/classifier

The fixed position is a hindrance to transcription projects, so in ASM, this changed to be a movable pop-up. 

It's not a modal since neither prevent interaction with the rest of the app, though, the ASM version does not allow you to make new marks until the sub-task pop-up is closed. It does allow you to use the image toolbar (i.e. rotate, zoom, etc). 

The new general classifier version UI, @beckyrother has advised that the sub-task should be based on the ASM version. @wgranger used this library for the UI: https://github.com/bokuweb/react-rnd

In the PR for ADR 20 (https://github.com/zooniverse/front-end-monorepo/pull/1268#discussion_r348332981), @eatyourgreens is asking if instead the sub-task appears in the task area similar to the survey task chosen option. @beckyrother, what are your thoughts?

The transcription task, #1313, will use a drawing sub-task of the same design (unless @beckyrother says otherwise), just preconfigured from the lab to have a text task that can be populated with auto-suggestions.

eatyourgreens:
If it is a popup, then it must be movable from the keyboard. If you do move it, it should probably remember the new position so that you don't have to move it again each time it opens.

Copying the survey task modal behaviour seems like less hassle for the volunteers. There's no risk that the transcription box will obscure the page, and the modal blocks the current task while still allowing you to zoom and pan the subject.

eatyourgreens:
Something I'm noticing with the [ASM text input](https://www.antislaverymanuscripts.org/classify) is that it always covers up the text that I'm trying to transcribe. Volunteers have complained about this on other transcription projects that use popup boxes so I don't want us to repeat that experience in the new transcription task.

That pushes me more towards having the text input appear to the right of the page, in or above the task area. @beckyrother do you have any thoughts about this?

beckyrother:
We chose to place the text box on top of the subject image so it would be near the text when a volunteer is transcribing. The ASM text box can be moved around to show the text below it; that would be my recommendation moving forward.

Here is an InVision prototype: https://projects.invisionapp.com/d/main#/console/12923997/396381419/preview

eatyourgreens:
[...]

PR #1397, opened 2020-01-06
Add ADR 25 about drawing sub-tasks

### LINKED ISSUE #500: [RFC] Drawing tools mark models

srallen:

Package: N/A

Closes #500. For #1315 and #1239 (#1376 closes it) and related to #1314.

Describe your changes:
This adds an ADR about the drawing sub-task annotation JSON structure and its UI.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
@beckyrother could you let me know if I've correctly captured the UI changes for the drawing sub-task in the ADR decision?

srallen:
@CKrawczyk and @eatyourgreens let me know if I've captured the discussion from various issues correctly.

docs/arch/adr-25.md
```diff
@@ -0,0 +1,157 @@
...
+
+To support movability and resizing, we will leverage [react-rnd](https://github.com/bokuweb/react-rnd) which is the same library ASM used. Grommet's `Layer` cannot be used since it is intended for actual modal or side panel use and cannot be arbitrarily positioned or moved.
+
+## Status
+
+Proposed
+
+## Consequences
+
+Flattening the annotations array for drawing task sub-tasks is conceptually consistent with the move to using workflow steps to flatten the combo task. It is however a breaking change and this change will have to be communicated to project builders. As with other classifications from the new classifier, this can be checked by the presence of `classifier_version: 2.0` in the classification metadata. In the future, we would also like to include a link to json schema for each annotation type as recommended in [ADR 07](adr-07.md).
```
eatyourgreens: Will this change involve changes to the classification API eg. can we still use `classification.addAnnotation(task, value)` from our task components?

Ideally, I think I'd like a model that extends the existing classifier model without breaking existing tasks like the PH-TESS data task.

eatyourgreens:
#1376 adds a model and API for drawing subtasks. Should that be mentioned here? I've [already commented on #1315](https://github.com/zooniverse/front-end-monorepo/issues/1315#issuecomment-566468662) assuming that this is the API that will be used to implement the views.

docs/arch/adr-25.md
```diff
@@ -0,0 +1,157 @@
+# ADR 25: Drawing Sub-task 
+
+Created: January 6, 2020
+
+## Context
+
+A drawing task's sub-task is designed to support volunteers answering additional questions for each drawing mark annotation. It allows single choice or multiple choice question task, text task, dropdown task, and slider task.
```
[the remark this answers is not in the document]
eatyourgreens: We should be careful about using 'drawing mark annotation' here and in other docs. A drawing task has a single drawing annotation. The annotation is composed of the marks for that task but individual marks don't have annotations.

docs/arch/adr-25.md
```diff
@@ -0,0 +1,157 @@
...
+      ]
+    }
+  ]}
+]}
+```
+
+This annotation structure has a few issues:
+
+- the annotations in the `details` array have no indication of the task types these originally are
+- the annotations in the `details` array does not have an unique identifier
```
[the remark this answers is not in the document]
srallen: I'm for more explicit naming, but I think downstream uses it and TESS launched with that in the annotation model... Certainly can change it, but I'd be for getting annotation json schemas done before this.

docs/arch/adr-25.md
srallen: I don't think so. We just need to make sure the `task` identifier matches up to what is in the `details` array. I think it would involve calling `classification.addAnnotation` twice. Once for the drawing task to update its annotation for `details` and once for the mark's sub-task annotation. I think the flattening has benefits for raw classification exports too, not just aggregation, but I could be wrong. Perhaps researcher input would benefit here too, so I'll invite the same as we invited on #1390.

srallen:
I'm happy to add comments about the model and API, but I think there are a couple of open questions here that we need consensus on which may impact what #1376 is doing.

As for the transcription task, I think it's reasonable to always require the sub-task there, but I don't think we can do this for all mark sub-tasks across the board. I don't think the transcription sub-task's implementation should impact the decisions here. We can update ADR 20 as needed for that.

docs/arch/adr-25.md
```diff
@@ -0,0 +1,157 @@
...
+    "taskType": "drawing",
+    "value": [
+      {
+        "frame": 0,
+        "tool": 0,
+        "toolType": "point",
+        "x": 452.18341064453125,
+        "y": 202.87478637695312,
+        "details": [
+          {"task": "T0.0.0"},
```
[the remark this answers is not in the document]
srallen: I've added a comment in the consequences section about potential benefits to the flatter structure outside of aggregation. I'd like more researcher input into this specific suggestion though.

docs/arch/adr-25.md
```diff
@@ -0,0 +1,157 @@
...
+          {"task": "T0.1.0"},
+          {"task": "T0.1.1"}
+        ]
+      }
+    ]
+  },
+  {
+    "task": "T0.0.0",
+    "taskType": "single",
+    "value": 0
```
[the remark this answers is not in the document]
srallen: I've updated the suggested convention for the identifier to include the annotation index.

srallen:
@lcjohnso and @vrooje could you provide input from a researcher perspective about the proposed change to flatten the annotation structure for drawing mark sub-tasks? Thanks!

docs/arch/adr-25.md
eatyourgreens: If we use array index as a mark identifier while classifying then this won't work since the mark array index changes when we add and delete marks. I was under the impression that the structure proposed here isn't written to the classification until the `completeClassification` action fires.

eatyourgreens: In fact, marks is a map, not an array, so this definitely needs more thought.

eatyourgreens: On example, of the top of my head: when I click on a mark to edit the sub tasks, how are the subtask annotations retrieved from the store?

I don't think the model proposed here works while classification is in progress.

docs/arch/adr-25.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Status
+
+Proposed
+
+## Consequences
+
+Flattening the annotations array for drawing task sub-tasks is conceptually consistent with the move to using workflow steps to flatten the combo task. It is however a breaking change and this change will have to be communicated to project builders. As with other classifications from the new classifier, this can be checked by the presence of `classifier_version: 2.0` in the classification metadata. In the future, we would also like to include a link to json schema for each annotation type as recommended in [ADR 07](adr-07.md).
+
+Flattening also has added benefits for Panoptes when generating classification exports. It can parse through a flattened array to convert machine readable strings to human readable strings for each task without having to check for values in a nested `details` array and then traverse it. In the raw classification export, this also benefits researchers that want to analyze the outputted CSV directly and prefer a flatter json structure.
```
CKrawczyk: I agree with this statement for the most part, but I want to point out that additional work would have to be done for the CSV to flatten into an expected shape.  As it is if it using the `TASK_KEY.ANNOTATION_INDEX.TOOL_INDEX.DETAILS_INDEX` key to flatten the annotation object the resulting CSV file would end up with one column for each sub-task answer given  and those columns would each have one non-null row (e.g. if a volunteer draws 50 points each with 2 sub-tasks there would be 100 new columns).  

If it was instead flattened on `TASK_KEY.TOOL_INDEX.DETAILS_INDEX` but still kept the original key for mapping back to the point it would be a more manageable shape (e.g. if a volunteer draws 50 points each with 2 sub-tasks there would be 2 new columns each with a list of 50 annotations).

docs/arch/adr-25.md
```diff
@@ -0,0 +1,154 @@
...
+    "value": [
+      {"option-3": 1},
+      {"option-4": 1},
+      {"option-5": 1}
+    ]
+  }
+]}
+```
+
+The sub-task identifiers follow a convention of `TASK_KEY.ANNOTATION_INDEX.TOOL_INDEX.DETAILS_INDEX`.
```
eatyourgreens: The naming is confusing here. There's only one annotation in the example, with two marks (I think.) Is `ANNOTATION_INDEX` the index of a particular mark inside the annotation? If so, I'd suggest `MARK_INDEX` instead.

eatyourgreens: Maybe change the order to task > tool > mark > detail as well? That reflects the hierarchy in the drawing model.

eatyourgreens: On reflection, this isn't a task identifier, since it points to a specific mark. I think Coleman's picked this up in his comment, but `annotation.task` should point to the task definition, so that analysis code knows how to parse `annotation.value`. In this case, the task definition is stored in Panoptes in `workflow.tasks[taskKey].tools[toolIndex].details`, rather than on each mark.

If all we need to analyse `annotation.value` is the task type, then this can be fixed by adding `annotation.type` to annotations. It depends whether any other info from the task is needed.

docs/arch/adr-25.md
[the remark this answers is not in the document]
eatyourgreens: The parent mark needs a separate identifier. `annotation.task` should point to the task that created this mark, which is stored in `workflow.tasks[taskKey].tools[toolIndex].details[detailIndex]`.

The mark identifier would be used, for example, when gathering all subtask annotations produced by a particular cluster of marks.

docs/arch/adr-25.md
CKrawczyk: I would be happy changing it to a true task identifier (e.g. `TASK_KEY.TOOL_INDEX.DETAILS_INDEX`) and having another field with the mark identifier `ANNOTATION_INDEX`.

eatyourgreens: If we're going to use this as the model while drawing is in progress, then the mark identifier will need to be a value that doesn't change as we add and remove marks.

If this model is only used to save completed classifications to Panoptes, then annotation index will work just fine.

My vote goes with the latter option, because a consequence of the former is breaking changes to the classifications store model, even for workflows that don't use the drawing task.

srallen: My vote goes for latter too because we can't write JS maps to Panoptes anyway, and you've noted that marks are maps in the MST models, so we will be doing some transformations at classification submission time regardless.

srallen:
Thanks for all the feedback. I'll try to get another pass at this today or tomorrow.

docs/arch/adr-25.md
eatyourgreens: Cool. In that case, I think #1376 can be reviewed independently of this ADR, because that PR doesn't touch on how classifications are generated from marks.

I think that will make working with subtasks cleaner too. We can annotate a subtask with `mark.addAnnotation(task, value)` while continuing to use `classification.addAnnotation(task, value)` to annotation classifications. (Noting that `task` in the classifier store API is an actual Task model, not a task key.)

srallen: I'll add a note about the suggestions here being the annotation model at submission of the classification.

srallen: Ok, I've made some changes that include clarifying that this is a proposal for the classification annotation JSON structure at submission and added `markIndex` to the sub-task annotation structure. I used `markIndex` instead of `annotationIndex` to be clear we're referring to an individual drawing task's `value` array rather than the parent `annotations` array. I sometimes conflate an individual's drawing task's `value` array which represents the marks with the parent `annotations` array in describing this. Is this clear or just more confusing?

docs/arch/adr-25.md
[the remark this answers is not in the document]
srallen: Ok, I've reverted the convention for the sub-task key to be reflective of the workflow task structure and added a separate `markIndex` for sub-task annotations.

docs/arch/adr-25.md
srallen: If we could have a specific example of when and why a researcher prefers a flatter structure that would be great. I just know we've been asked for that in general.

docs/arch/adr-25.md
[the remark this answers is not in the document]
eatyourgreens: It occurred to me that if we gave annotations identifiers, then `details` here could be an array of annotation IDs, each pointing to an annotation in `classification.annotations`.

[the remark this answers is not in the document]
eatyourgreens: I think we've discussed in the past that `details` wasn't a very good choice of name for the mark's task annotations array. We could change it to `annotations` for clarity.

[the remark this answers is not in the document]
srallen: @CKrawczyk or @lcjohnso do you have a preference? I'm not sure if `annotations` is clear enough to differentiate from the parent `annotations` key in the classification... but I agree `details` isn't very good either.

docs/arch/adr-25.md
srallen: This was discussed in another comment thread: https://github.com/zooniverse/front-end-monorepo/pull/1397#discussion_r364835591

docs/arch/adr-25.md
srallen: @CKrawczyk and @eatyourgreens is the most recent text satisfactory?

docs/arch/adr-25.md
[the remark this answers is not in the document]
eatyourgreens: The code uses `annotations` because both `classification` and `mark` use the `AnnotationStore` model. That may have been premature optimisation: the idea being that the subtask popup box should act exactly like the task area, only updating `mark` instead of `classification`.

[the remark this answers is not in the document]
CKrawczyk: As it is a sub-task would `sub_annotations` (or similar) work?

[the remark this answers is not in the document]
eatyourgreens: Off the top of my head, I think that depends on whether the task components can be updated to take either `annotations` or `sub_annotations` as a prop.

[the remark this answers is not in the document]
srallen: We're conflating in this discussion what the internal stores and components are doing with what is being submitted in the classification annotation. The Task components can take whatever we want them to take as a prop. The question I'm posing has to do with the naming at classification submission.

docs/arch/adr-25.md
```diff
@@ -0,0 +1,161 @@
...
+        ]
+      }
+    ]
+  },
+  {
+    "task": "T0.0.0",
+    "taskType": "single",
+    "markIndex": 0,
+    "value": 0
+  },
```
eatyourgreens: Maybe add this to the example to show how two marks from the same tool would be represented?
```json
{
    "task": "T0.0.0",
    "taskType": "single",
    "markIndex": 2,
    "value": 0
}
```

eatyourgreens:
Are we waiting on feedback from @vrooje before merging this?

srallen:
Nope, though I would prefer if this did get an explicit approval from a dev at the very least. My question from this comment thread hasn't been responded to [link](https://github.com/zooniverse/front-end-monorepo/pull/1397#discussion_r367640788). And whether or not `details` should be renamed to something else in the classification annotation isn't resolved yet. I haven't had opportunity to fully circle back to this yet, but it'd be nice if someone would respond to my question.

docs/arch/adr-25.md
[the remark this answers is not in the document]
eatyourgreens: #1432 allows us to pass any annotation into a Task component, regardless of where it comes from or what the source is called. At the moment, master is hard-coded to use `classification.annotations`. That's my fault.

eatyourgreens:
>If we could have a specific example of when and why a researcher prefers a flatter structure that would be great. I just know we've been asked for that in general.

https://github.com/zooniverse/front-end-monorepo/pull/1397#discussion_r365453415

Was this ever answered?

srallen:
> Was this ever answered?

No, outside of our own internal reasons, from what I remember, researchers that use R or spreadsheets to analyze the exports would like flatter structures. I could include a statement about that? @lcjohnso do you have any specific examples we can include in the ADR of researchers requesting flatter JSON?

lcjohnso:
@srallen Flat structures facilitate research teams being able to load data without json-based manipulation.  There are many teams who would benefit from the ability to read-in a CSV to Excel and start analyzing their results, as opposed to needing to first parse json.

docs/arch/adr-25.md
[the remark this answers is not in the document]
srallen: There isn't a clear resolution in this discussion on what to do, but we have issue #1443 now to track specifically the transformations needing to happen before submitting the annotations as a classification, so any further discussion can happen there and we can update the ADR as needed.

docs/arch/adr-25.md
srallen: I'll take no response as acceptance, but we have issue #1443 where this can be discussed further.

srallen:
Thanks @lcjohnso, I'll include that in the text of the ADR.

docs/arch/adr-25.md
srallen: I've included this. Thanks.

srallen:
Further discussion can happen in  #1443 and this ADR can be updated in future PRs as needed while implementation is happening.

PR #1399, opened 2020-01-08
Refactor banner to not block being able to delete TESS annotations

### LINKED ISSUE #932: Move already seen banners on top of subject
**Package** lib-classifier

It's impossible to remove the column data annotation with the banners:

Image from iOS

srallen:

Package: lib-classifier

Closes #932 

Describe your changes:
I've refactored the variant already-seen banners to display above the subject as block the rather than be absolutely positioned over the top portion of the subject. The latter was preventing annotations from being deleted while classifying on TESS if there was a reason the banner is showing. 

This incorporates the design feedback from #932.

I modified the animation from being a slide from top and fade to just a fade in. The reason being is that with the banner being a standard positioned block, it _will_ bump the subject down slightly as it renders. Usually when the page is loading, this shouldn't really be all that visible, but the slide + fade in is a bit more jarring now than a fade in only. To demonstrate...

Here's how this looks with the slide + fade:

already-seen-banner-animation-previous

Here's how this looks only as a fade in:

already-seen-animation-fade

Here's how this looks in context as you load the classify page on TESS:

already-seen-tess-better

And last, here's a screenshot showing that the annotation marks can now be deleted as needed since the button is no longer blocked:

Screen Shot 2020-01-07 at 3 57 36 PM

As an aside, simplifying the animation means we can use the built in Grommet animations and remove react-pose. If we find that we need a more complex animation that Grommet doesn't do or doesn't do well, then we can add it back.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
@beckyrother fade in and slide down is what the animation was previously. I can definitely revert it back if that's what you'd like. 

What if we added the same fade in animation to the subject viewer so they both fade in? In context, the state of the subject is known when the subject is loaded, so both are loading and rendering at the same time unlike the storybook example.

beckyrother:
Makes sense! Let's go with that. Thanks Sarah!

srallen:
@beckyrother here's how this looks with the image load in context:

already-seen-fade-in-with-subject

I'm holding off adding a loading animation to the TESS LCV because adding CSS to D3 plots is a bit more complicated and I'd rather do it to the new generic scatter plot that I hope to update the underlying TESS LCV to use later on if that is ok?

beckyrother:
Looks great!

packages/lib-classifier/src/components/Classifier/components/Banners/components/Banner/Banner.js
```diff
@@ -64,54 +52,51 @@ class Banner extends Component {
...
-            key={bannerText}
-            justify='between'
-            pad='small'
-            show={show}
+    return show && (
+      <Box
+        align='center'
+        animation='fadeIn'
+        background={{ color: background }}
+        className={className}
```
rogerhutchings: `classname` is probably no longer necessary, since it's using the standard `Box` component now.

PR #1402, opened 2020-01-10
Drawing: add x_center, y_center to Point marks

eatyourgreens:
Add x_center, y_center to the Point model.
Make mark.x, mark.y read-only aliases for mark.x_center and mark.y_center for all marks.

See #1390 for the proposal to add `x_center` and `y_center`.

Package:
lib-classifier/plugins/drawingTools

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/README.md
+++ b/packages/lib-classifier/src/plugins/drawingTools/README.md
@@ -54,11 +54,15 @@ The [base Mark model](https://github.com/zooniverse/front-end-monorepo/tree/mast
 - _isValid (boolean)_ Read only. True if any required validations pass for this mark (eg. minimum length for a line.)
 - _tool (Tool)_ Read only. A reference to the tool that created this mark.
 - _addAnnotation(task, value)_ Add `value` to the annotation for `task`, which should be a valid task for this mark.
+- _x (number)_ Read only. X position of the mark's centre, in SVG coordinates relative to the subject image. Override this if `mark.x` should refer to a different position eg. the mark's top left corner for rectangles.
+- _y (number)_ Read only. Y position of the mark's centre, in SVG coordinates relative to the subject image. Override this if `mark.y` should refer to a different position eg. the mark's top left corner for rectangles.
 
 All marks should extend the Mark model by implementing the following views and actions:
 - _coords (Object { x, y })_ Read only. Returns the `{ x, y }` coords for this mark.
 - _deleteButtonPosition(scale) (Object { x, y })_ Given the image scale, return the `{ x, y }` position for this mark's delete button.
 - _toolComponent (React.Component)_ Read only. Returns the React component used to render this mark.
+- _x_center_ (number)_ x position of the mark's centre, in SVG coordinates relative to the subject image.
+- _y_center_ (number)_ y position of the mark's centre, in SVG coordinates relative to the subject image.
 - _initialDrag({ x, y })_ Called on drag when first creating the mark. `{ x, y }` are the new position of the dragged pointer in the frame of the subject image.
 - _initialPosition({ x, y })_ Called on initial click/tap when creating the mark. `{ x, y }` are the position of the pointer in the frame of the subject image.
 - _move(difference)_ Called on drag when moving the mark. `difference` is the change in position since the last move: `{ x, y }`.
@@ -67,5 +71,5 @@ All marks should extend the Mark model by implementing the following views and a
 In addition, mark models should extend the base Mark model with any properties specific to the new shape. These mark properties will be passed to Panoptes as the annotation for this mark. Marks may specify the following properties, which have a special meaning when rendering marks.
 
 - _angle (number)_ Rotation angle of the mark in degrees, measure clockwise from the positive x-axis.
-- _x (number)_ x position of the mark's centre of rotation, in SVG coordinates relative to the subject image.
-- _y (number)_ y position of the mark's centre of rotation, in SVG coordinates relative to the subject image.
+- _x_rotation_ (number)_ x position of the mark's centre of rotation, in SVG coordinates relative to the subject image.
+- _y_rotation_ (number)_ y position of the mark's centre of rotation, in SVG coordinates relative to the subject image.
--- a/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
@@ -22,6 +22,14 @@ const BaseMark = types.model('BaseMark', {
 
     get tool () {
       return getParentOfType(self, Tool)
+    },
+
+    get x () {
+      return self.x_center
+    },
+
+    get y () {
+      return self.y_center
     }
   }))
 
--- a/packages/lib-classifier/src/plugins/drawingTools/models/marks/Point/Point.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/models/marks/Point/Point.js
@@ -7,8 +7,8 @@ import Mark from '../Mark'
 
 const PointModel = types
   .model('PointModel', {
-    x: types.optional(types.number, 0),
-    y: types.optional(types.number, 0)
+    x_center: types.optional(types.number, 0),
+    y_center: types.optional(types.number, 0)
   })
   .views(self => ({
     get coords () {
@@ -38,23 +38,23 @@ const PointModel = types
   }))
   .actions(self => {
     function initialDrag ({ x, y }) {
-      self.x = x
-      self.y = y
+      self.x_center = x
+      self.y_center = y
     }
 
     function initialPosition ({ x, y }) {
-      self.x = x
-      self.y = y
+      self.x_center = x
+      self.y_center = y
     }
 
     function move ({ x, y }) {
-      self.x += x
-      self.y += y
+      self.x_center += x
+      self.y_center += y
     }
 
     function setCoordinates ({ x, y }) {
-      self.x = x
-      self.y = y
+      self.x_center = x
+      self.y_center = y
     }
 
     return {


shaunanoordin:
Package: `lib-classifier`

Code changes look legit; I'll run a quick test shortly to confirm `x_center-y_center` holds up to functional checks.

- I'm thinking maybe `src/plugins/drawingTools/README.md` be updated? Any time a `get`ter is set for a class, it's useful to write it down in big honking notes in case somebody in the future uses `myCustomMark.x = 100; console.log(myCustomMark.x)` and keeps wondering why it returns `undefined`. (Disclaimer: I need to check that getter-vs-property conflicts resolve this way.) 
- Since `Line` uses `x1, x2, y1, y2`, it doesn't need `center_x` and `center_y`... which presumably means if anyone was silly enough to access `line.x` it'll just go "narp, no idea what you're asking for, here's an undefined".

shaunanoordin:
Minor update: (irrelevant to this PR, but fun trivia for the kids and the whole family!)

getter-vs-property conflicts DO resolve _in favour of the getter,_ as I thought.

```
class Mark {
  constructor () {
    this._x = 'LOL';
  }

  get x () {
    return this._x;
  }
}

let mark = new Mark();
mark.x = '1000';
console.log(mark.x);

// Console prints 'LOL', not '1000'
```

eatyourgreens:
You can't set a computed value. If you try `mark.x = 10` MobX State Tree will throw an error, since `mark.x` and `mark.y` are read-only.

Good point about updating the Readme. I'll do that.

eatyourgreens:
In that case, we are at an impasse. `mark.x` and `mark.y` can’t be computed values for some marks but not for others.

eatyourgreens:
>Also to be clear, we won't have a { x, y } in any of the shapes anymore. They almost all will be replaced with { x_center, y_center } with the exception of shapes that have a point of rotation elsewhere like the fan tool.
https://github.com/zooniverse/front-end-monorepo/pull/1390#discussion_r363356309

I took this to mean that `{ x, y }` are replaced by `{ x_center, y_center }` where applicable. It seems very dangerous, to me, to have point marks writing to values that are defined as read-only in all other models. Unless you're suggesting we have completely separate positioning code for the Point model?

srallen:
I would refer to the final language in the ADR as the proposed final decision rather than comments. Points aren't shapes, so the recommendation of this change in the proposal is not applicable to this particular tool mark. However, since it is a proposal, it is open to change, so please outline what should changed compared to ADR 24, but ADR 24 should be the standard in which the drawing tool PRs should be compared against. If we start making PRs immediately that are inconsistent with it and we don't update it, then it defeats the purpose of the ADR documentation.

eatyourgreens:
I had already made this clear, both in the discussion for ADR 24 and the discussion for the Ellipse tool, yet the ADR was merged anyway.

srallen:
Yes I'm aware of the concerns that you raised, but I thought I had addressed them in the ADR and they were satisfactory to you? If I'm mistaken, then apologize.

This is an evolving process for documentation. I did make a decision call to merge after it received approval with the noted caveat that it may be updated as we proceed with implementation in an attempt to try to move forward. It's not the specific code that I have an issue with; it's that I want to be sure that we keep the ADR up-to-date with what we're actually doing because it's much harder after the fact to do documentation. So what I'd like to see is either in this PR or in a PR parallel to it is a proposed change to the decision in the ADR. As we're going forward, what I also hope we do is review code against ADRs, too.

eatyourgreens:
OK, that makes sense. This PR was written to inform ADR-24, so that we had an example implementation to refer to.

I'd also like to see ADR's written alongside working code, or after something is built in order to record the decisions that went into the build.

At the moment we have the problem that ADR 24 makes decisions that don't quite work when implemented. It calls for a slightly different API for each type of mark model, whereas in earlier ADR's we decided that we would build a common API for drawing. We can have one or the other, but not both.

srallen:
I would disagree that we should draft ADRs as code is being written. I see writing the first draft of a proposal ADR as an important first step to make sure we collect all of the recommendations into a single point of reference to be sure we're incorporating everything we're thinking about. ADR 24 has a proposal status which indicates to me that it is still WIP. This specific point is more general than the drawing tool changes recommended here, so I'm happy to continue a meta conversation about the ADR process in an RFC issue about it.

srallen:
> At the moment we have the problem that ADR 24 makes decisions that don't quite work when implemented. It calls for a slightly different API for each type of mark model, whereas in earlier ADR's we decided that we would build a common API for drawing. We can have one or the other, but not both.

The recommendation in ADR 24 is specifically about the annotation model that is submitted at classification and shapes have different needs than non-shapes. I'm not quite understanding what the conflict is here since the API is internal methods for the in-progress marking? When I have opportunity, I will dig into it more to try to understand better what the potential issue you are raising is.

eatyourgreens:
The specific issue that's solved in this PR is that `{ x, y }` are computed values (read-only) for all marks except Points, where they are observable properties (read-write). So points can assign to `mark.x` and `mark.y` (and must, in order to be moved around) but other marks will error with that same code.

So, Point has to have separate code from the other mark models. It can't share the same drawing API.

eatyourgreens:
That said, I'm less fussed about it after reading [Goodbye, Clean Code](https://overreacted.io/goodbye-clean-code/).

srallen:
That is a good article. Thanks for sharing. I do think we will have situations where some marks' models cannot or we choose to not have it absolutely conform with the base model for a given reason. This will be true for the freehand tools where we're recommending using the SVG path as its annotation. And it may be true for other non-shapes and/or non-symmetrical shapes we may want to support in the future. That being said, I do not think this reality contradicts the stated goal of a common API. The common API still generally still exists as all marks will still be a composition of the base models. If something can't be in the base model anymore, we can refactor as needed. Additionally, there may be different needs between the internal API model and what is submitted as classification.

Now back to the specific Point tool. It's in a gray area for me as technically a point could be thought of as a "center", but it's not a shape since it has no width or height. As long as we stay flexible for future mark needs, I'm ok with Point using `x_center` and `y_center`, but ADR 24 still needs to be updated because it lists Point as an exception to this change.

eatyourgreens:
#1427 is an opportunity to find out what needs to be written by a new developer vs. what we can provide via the base Tool and Mark models. It's also a chance to see how we communicate changes in the base models. The latest merges to master have broken it, I'm guessing because of changes to the Mark model.

eatyourgreens:
>Now back to the specific Point tool. It's in a gray area for me as technically a point could be thought of as a "center", but it's not a shape since it has no width or height. As long as we stay flexible for future mark needs, I'm ok with Point using x_center and y_center, but ADR 24 still needs to be updated because it lists Point as an exception to this change.

From a purely pragmatic point of view, I'd find out whether it's better for aggregation rules to always cluster on `{ x_center, y_center }` or whether clustering rules should look for `{ x_center, y_center }` and fall back to `{ x, y }` if those aren't present.

My interpretation of the ADR was that clustering should always use `{ x_center, y_center }` and fail if those aren't present but I could be wrong.

srallen:
@CKrawczyk Could you provide input for the questions @eatyourgreens has in the previous comment? I believe the aggregation code can or will do what is needed based on the request that @CKrawczyk made. 

As I understand it, the request is for clarifying the annotation model for researchers that for symmetrical shapes, x and y should be the center of the shape, and the name in the annotation should be clearly labeled as such, therefore the requested change for this renaming in the annotation was *only* for shape marks. Hence my worry that using `x_center` and `y_center` for Point will be confusing because it's not really the center of anything. If this change needed to happen to all mark annotations, I think @CKrawczyk would have asked for that. 

Our code needs to be flexible and adaptable for variation in the mark annotation model and I think ultimately the needs of researchers come first if we can do it. Abstraction, optimization, etc comes second. Of course this isn't an absolute, and some optimization will be prioritized for the freehand tools when they get ported.

CKrawczyk:
> As I understand it, the request is for clarifying the annotation model for researchers that for symmetrical shapes, x and y should be the center of the shape, and the name in the annotation should be clearly labeled as such, therefore the requested change for this renaming in the annotation was only for shape marks.

This is exactly the case.  For a point `x` and `y` are fine because they are unambiguous about what they are referring to.

For some insight into how the aggregation handles this, I use a [look up table](https://github.com/zooniverse/aggregation-for-caesar/blob/master/panoptes_aggregation/shape_tools.py) to define what data is used to cluster each shape.  These parameters are directly passed into the clustering code without any transforms applied, so if the center is not provided it can't be used for clustering (without writing specific clustering code for each of the shapes independently).  If the center is provided then all the shapes can use exactly the same reducer without any code specific to a shape.

tl;dr each shape can have different labels for the parameter names, just as long as there are ones for the center when a shape has a center.

eatyourgreens:
Ah, right. I understand now. So, whenever we change a shape in the code here, that piece of Python code also needs to be updated?

CKrawczyk:
Yep, any extractor needs to be able to pull out the parameters it needs, so the python code needs to know what they are called.  Unless some object is passed along with the classifications saying "here are this shapes parameters" these values need to be hard-coded into aggregation.

camallen:
FWIW publishing classification / annotation payload schemas is one way to ensure loose coupling between systems. 

This has been handled in other serializer systems before, protobufs, avro, thrift etc  https://docs.confluent.io/current/schema-registry/index.html

Not saying we should do this exact thing but we could look at publishing schemas for use in the creation services and the consuming services

eatyourgreens:
I think there's an open issue somewhere about publishing schemas for the drawing tools. It might be on PFE.

eatyourgreens:
I've definitely had conversations with Roger about how to publish URLs with annotations, where the URL resolves to a schema for the version of the task/tool that was used to create that annotation.

srallen:
It's in fact documented as a proposed ADR: https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-07.md

Now that we've cleared that all up regarding what recommended changes apply to shapes only, I'd like to close this. The docs can be updated in a separate PR. Does that sound good?

PR #1408, opened 2020-01-13
TranscriptionLine tool clean up

srallen:

Package: lib-classifier

This is just a bit of clean up removing cruft and fixing some of the naming. I've added `toolType` to the mark to make this consistent with the decisions in ADR 24.

I added a try/catch around `getParentOfType` which gets the stories working again.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
@@ -7,25 +7,18 @@ import { TranscriptionLineTool } from '@plugins/drawingTools/models/tools'
 import { DrawingToolRoot } from '@plugins/drawingTools/components'
 import TranscriptionLine from './TranscriptionLine'
 
-
 class DrawingStory extends Component {
-  constructor () {
-    super()
-    this.svg = React.createRef()
-  }
-
   render () {
     const { children, mark, tool } = this.props
-    const { svg } = this
+
     return (
-      <svg viewBox='0 0 300 400' height={300} width={400} ref={this.svg}>
+      <svg viewBox='0 0 300 400' height={300} width={400}>
         <DrawingToolRoot
           label='Transcription line'
           mark={mark}
-          svg={svg}
           tool={tool}
         >
-          {React.cloneElement(children, { svg: svg.current })}
+          {React.cloneElement(children)}
         </DrawingToolRoot>
       </svg>
     )
@@ -41,10 +34,11 @@ storiesOf('Drawing tools | Transcription Line', module)
   .add('complete', function () {
     const tool = TranscriptionLineTool.create({
       color: 'red',
-      type: 'polygon'
+      type: 'transcriptionLine'
     })
     const mark = tool.createMark({
       id: 'transcriptionLine1',
+      toolType: 'transcriptionLine',
       x1: 10,
       y1: 20,
       x2: 205,
@@ -60,10 +54,11 @@ storiesOf('Drawing tools | Transcription Line', module)
   .add('active', function () {
     const tool = TranscriptionLineTool.create({
       color: 'red',
-      type: 'polygon'
+      type: 'transcriptionLine'
     })
     const mark = tool.createMark({
       id: 'transcriptionLine1',
+      toolType: 'transcriptionLine',      
       x1: 10,
       y1: 20,
       x2: 205,
@@ -79,10 +74,11 @@ storiesOf('Drawing tools | Transcription Line', module)
   .add('unfinished', function () {
     const tool = TranscriptionLineTool.create({
       color: 'red',
-      type: 'polygon'
+      type: 'transcriptionLine'
     })
     const mark = tool.createMark({
       id: 'transcriptionLine1',
+      toolType: 'transcriptionLine',      
       x1: 10,
       y1: 20,
       x2: 205,
--- a/packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
@@ -5,6 +5,7 @@ import { Mark } from '@plugins/drawingTools/models/marks'
 
 const TranscriptionLineModel = types
   .model('TranscriptionLineModel', {
+    toolType: types.literal('transcriptionLine'),
     x1: types.maybe(types.number),
     y1: types.maybe(types.number),
     x2: types.maybe(types.number),
--- a/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
@@ -21,7 +21,11 @@ const BaseMark = types.model('BaseMark', {
     },
 
     get tool () {
-      return getParentOfType(self, Tool)
+      try {
+        return getParentOfType(self, Tool)
+      } catch (e) {
+        console.log(`no parent of ${Tool.type} available`)
+      }
     }
   }))
 
--- a/packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.js
@@ -5,7 +5,7 @@ import { DragHandle } from '@plugins/drawingTools/components'
 const FINISHER_RADIUS = 8
 const GRAB_STROKE_WIDTH = 6
 
-function TranscriptionLine ({ active, mark, onFinish, scale, svg }) {
+function TranscriptionLine ({ active, mark, onFinish, scale }) {
   const { x1, y1, x2, y2, finished } = mark
   const finisherRadius = FINISHER_RADIUS / scale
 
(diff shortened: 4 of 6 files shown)


eatyourgreens:
I've been using
```js
const someTool = someTool.create(snapshot)
const someMark = someTool.createMark()
```
to create marks with parents in tests and stories. I can't remember if I added that to the drawing tool readme, but it may work.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
```diff
@@ -7,25 +7,18 @@ import { TranscriptionLineTool } from '@plugins/drawingTools/models/tools'
...
-    this.svg = React.createRef()
-  }
-
   render () {
     const { children, mark, tool } = this.props
-    const { svg } = this
+
     return (
-      <svg viewBox='0 0 300 400' height={300} width={400} ref={this.svg}>
+      <svg viewBox='0 0 300 400' height={300} width={400}>
```
eatyourgreens: I think this needs to be wrapped in `SVGContext.Provider`, now that `svg` is passed in context and not as a ref.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
```diff
@@ -21,7 +21,11 @@ const BaseMark = types.model('BaseMark', {
     },
 
     get tool () {
-      return getParentOfType(self, Tool)
+      try {
+        return getParentOfType(self, Tool)
+      } catch (e) {
+        console.log(`no parent of ${Tool.type} available`)
+      }
```
eatyourgreens: Should we enforce that marks must be created via `tool.createMark()`? I can't think of a situation where a mark would exist without a parent tool.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
srallen: The mark doesn't use the SVG Context, so is that necessary?

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
srallen: Not in context of an app, but it does in the story!

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
eatyourgreens: Anything that's draggable does use the SVG context.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
eatyourgreens: Surely the story should reflect how we're creating marks in the app? That's how I've been creating marks for tests.

eatyourgreens: I put the example of how to do this in the ADR, which probably isn't the place anyone would look for an example.  : /
https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-22.md#the-drawing-tools-api

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
eatyourgreens: In this case, `DrawingToolRoot` and the drag handles are the context consumers.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
eatyourgreens: By the way, this will attempt to find a parent of type `Tool`, but the transcription line's  parent tool is a `TranscriptionLineTool`. Each mark model has to implement this view for its specific tool type.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
srallen: The story is importing the undecorated `DrawingToolRoot`, so it's not decorated by `draggable` and itself isn't a consumer of the SVG context nor is the mark. I may be missing something though, so pointing to a specific line would be helpful.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
srallen: So I'm not sure how to do that for the story without adding a Provider for MST stores so that the tree transversal function works. The ADR recommendation doesn't really apply here?

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
eatyourgreens: That's odd. I wonder why it does that? Either way, the mark's drag handles are draggable so watch out for context errors from those.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
eatyourgreens: This should work to create a line that you can pass into the React component.
```js
const transcriptionTool = TranscriptionLineTool.create({ type: 'transcriptionLine', color: 'red' })
const transcriptionLine - transcriptionTool.createMark({ id, x1, y1, x2, y2 })
```
`tool.createMark(snapshot)` should be described in the tools Readme, but I'm not sure how useful that is. PR's to improve the documentation are always welcome!

Maybe we need some examples like this? Examples would be useful for the classifier models too.

eatyourgreens: I'm going to write some of the missing stories for the other drawing tools and see if my advice actually works.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
eatyourgreens: This is actually importing the decorated `DrawingToolRoot` from `drawingTools/components`.

eatyourgreens: #1410 updates `DrawingToolRoot` to export the undecorated component for testing etc.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
```diff
@@ -7,25 +7,18 @@ import { TranscriptionLineTool } from '@plugins/drawingTools/models/tools'
...
         <DrawingToolRoot
           label='Transcription line'
           mark={mark}
-          svg={svg}
           tool={tool}
         >
-          {React.cloneElement(children, { svg: svg.current })}
+          {React.cloneElement(children)}
```
eatyourgreens: Why is this cloning the children without changing them?

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
eatyourgreens: See #1413 for working Point and Line stories. I think this PR has forgotten to add a `tools` view to the `TranscriptionLine` model. By default, `mark.tool` will look for a parent Tool model, which will fail here because the tool model types don't match.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.stories.js
eatyourgreens: Trying out other marks in #1413, dragging throws `TypeError: svg is not defined` but nothing crashes so I think the default empty context is ok here.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
eatyourgreens: #1414 will fix this by changing the fallback view to return the mark's grandparent, rather than search for a parent node of a specific type.

packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/TranscriptionLine/TranscriptionLine.js
```diff
@@ -5,6 +5,7 @@ import { Mark } from '@plugins/drawingTools/models/marks'
 
 const TranscriptionLineModel = types
   .model('TranscriptionLineModel', {
+    toolType: types.literal('transcriptionLine'),
```
eatyourgreens: `toolType` must be added to the drawing code that creates new marks. Otherwise, clicking on the drawing canvas to start a new transcription line will throw an error in MST.

`toolType` should also be added to the tests for drawing a new mark, and also to the drawing tools Readme.

These are the reasons that I would prefer if `toolType` was added to all drawing tools in a separate PR.

srallen:
I'm closing this. I thought this would be helpful, but it doesn't seem like it is. I'll just add a review comment to the original PR that the cruft svg props and the stories need updating though.

PR #1409, opened 2020-01-14
Remove useState from tasks

### LINKED ISSUE #1407: Task Area rendering with previous Classification
## Classification & UI Bug

Package: `lib-classifier`

On a project featuring a workflow with a single simple yes/no question, the Task Area of the Classifier UI _visually retains_ the yes/no answer of the previous Classification.

Tested on macOS 10 + Chrome 79

### Details

In this example, I'm using the Classifier to view a sample project (http://localhost:8080/?env=staging&project=1898) with only one yes/no question.

Screen Shot 2020-01-13 at 17 33 03

First, I receive the 1st Subject and select the "No" answer. I then click Submit.

Screen Shot 2020-01-13 at 17 33 22

The 1st Classification is submitted and the 2nd Subject is loaded. Viewing the Network console, we note that the Classification is submitted with the correct values.

In the UI though, we see the first problem: the 'no' answer has already been selected, despite me not taking any further action after clicking the Submit button.

**Sub-Issue 1: UI doesn't reset between Subjects**

- Expected: when loading a new Subject, the task area's questions/answers should be reset. (i.e. neither 'yes' nor 'no' should be selected as an answer on the UI.)
- Actual: when loading a new Subject, the previous answer in the task area persists.

Without touching the yes/no answers, I now click the Submit button again.

Screen Shot 2020-01-13 at 17 33 45

The 2nd Classification is submitted and the 3rd Subject is loaded. Viewing the Network console, we note that the Classification is submitted with **incorrect values.** Specifically, the UI says 'no' was selected (expected annotation value: 1), but the Classification says the user didn't choose any answer (actual annotation value: null)

**Sub-Issue 2: when UI doesn't reset, Classification sent doesn't reflect answers shown in UI**

- Expected: when the UI shows a yes/no answer being selected, that answer should be reflected in the Classification submitted to the API
- Actual: if the user doesn't interact with the UI, despite the UI showing a yes/no answer being selected, the Classification submitted to the API is blank

### Status

⚠️ This is a serious UI bug that can affect scientific classifications.

eatyourgreens:
- Remove `useState` from tasks, so that the current task value is always passed in as a prop via the task annotation.
- Observe tasks for changes.
- Rewrite tests to directly test annotation values.

Package:
lib-classifier

Closes #1407.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
@@ -48,7 +48,7 @@ class Tasks extends React.Component {
       return (
         <Box as='form' gap='small' justify='between' fill>
           {tasks.map((task) => {
-            const { TaskComponent } = taskRegistry.get(task.type)
+            const TaskComponent = observer(taskRegistry.get(task.type).TaskComponent)
             if (TaskComponent) {
               return (
                 <Box key={task.taskKey} basis='auto'>
--- a/packages/lib-classifier/src/plugins/tasks/MultipleChoiceTask/components/MultipleChoiceTask.js
+++ b/packages/lib-classifier/src/plugins/tasks/MultipleChoiceTask/components/MultipleChoiceTask.js
@@ -2,7 +2,7 @@ import { Markdownz } from '@zooniverse/react-components'
 import { Box, Text } from 'grommet'
 import { observable } from 'mobx'
 import { inject, observer, PropTypes as MobXPropTypes } from 'mobx-react'
-import React, { useState } from 'react'
+import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import zooTheme from '@zooniverse/grommet-theme'
@@ -33,7 +33,7 @@ function MultipleChoiceTask (props) {
     task
   } = props
   const { annotation } = task
-  const [ value, setValue ] = useState(annotation.value)
+  const { value } = annotation
 
   function onChange (index, event) {
     const newValue = value ? value.slice(0) : []
@@ -43,7 +43,6 @@ function MultipleChoiceTask (props) {
       const indexInValue = newValue.indexOf(index)
       newValue.splice(indexInValue, 1)
     }
-    setValue(newValue)
     task.updateAnnotation(newValue)
   }
 
--- a/packages/lib-classifier/src/plugins/tasks/MultipleChoiceTask/components/MultipleChoiceTask.spec.js
+++ b/packages/lib-classifier/src/plugins/tasks/MultipleChoiceTask/components/MultipleChoiceTask.spec.js
@@ -13,7 +13,9 @@ const task = {
   required: false,
   taskKey: 'T1',
   type: 'multiple',
-  updateAnnotation: sinon.stub()
+  updateAnnotation: sinon.stub().callsFake(value => {
+    task.annotation.value = value
+  })
 }
 
 describe('MultipleChoiceTask', function () {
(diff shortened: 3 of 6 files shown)


srallen:
@shaunanoordin drawing tasks are different from single choice or multiple choice tasks in that they default to having the first tool selected.

Original PFE code:
https://github.com/zooniverse/Panoptes-Front-End/blob/59f8e1caace0fb88b0e799265a5a5d7145d21847/app/classifier/tasks/drawing/index.cjsx#L46-L48

In the UI component:
https://github.com/zooniverse/Panoptes-Front-End/blob/59f8e1caace0fb88b0e799265a5a5d7145d21847/app/classifier/tasks/drawing/index.cjsx#L102

New classifier:
https://github.com/zooniverse/front-end-monorepo/blob/52acb067c4d18d677f22ebd2caca02847db33a8c/packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingTask.js#L8-L9

In the UI component:
https://github.com/zooniverse/front-end-monorepo/blob/52acb067c4d18d677f22ebd2caca02847db33a8c/packages/lib-classifier/src/plugins/tasks/DrawingTask/components/DrawingTask.js#L52 

We should document this somewhere to point out that there is a selected input by default for the drawing task on mount.

eatyourgreens:
The original report focussed on the task area, so I didn’t look at the subject viewer here. Did I miss something there?

The workflow steps store resets everything when a subject loads. It’s undocumented, sorry, but I’d start looking there.

eatyourgreens:
I’d also assume that the crash you’re seeing is MST crashing because you’ve tried to modify a point mark that’s no longer present in the tree. I’ve found that MST generates good error messages in the console.

shaunanoordin:
> That being said, I didn't fully understand your report when I first read it, so there's an issue with marks remaining? I'll have a look too to help pin point this.

> The original report focussed on the task area, so I didn’t look at the subject viewer here. Did I miss something there?

@srallen @eatyourgreens Sorry, this was my slip up - I wasn't thorough enough with my initial report, and I should have tested for 'workflows with 1x Drawing Tasks'. 🤦‍♂ I'll update #1407 with my investigation on that.

> I’d also assume that the crash you’re seeing is MST crashing because you’ve tried to modify a point mark that’s no longer present in the tree. I’ve found that MST generates good error messages in the console.

@eatyourgreens That's right!

Error message: `Uncaught Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree. (Object type: 'Point', Path upon death: '/workflowSteps/steps/S0/tasks/0/tools/0/marks/ck5e276p100013h5w9w6dnkqy', Subpath: 'id', Action: ''). Either detach nodes first, or don't use objects after removing / replacing them in the tree.`

Side note: on workflows with '1x Drawing Task', when a new Subject loads, the _Task Area_ correctly shows "0 marks drawn", but the previous Classification's marks persist on the _Subject Viewer._ So you've taken care of everything that needs to be done in the Task Area. 👌

shaunanoordin:
I've updated the initial report with details on the 'workflows with 1x Drawing Task' persistent marks problem: https://github.com/zooniverse/front-end-monorepo/issues/1407#issuecomment-574252865

srallen:
@shaunanoordin I see the issue you're seeing. Marks are remaining between subjects. I think this might be a similar but separate issue, so it may be best to document it in its own issue and keep this scoped to fixing the task area?

shaunanoordin:
👍 Roger that, @srallen. I'll spin off a new issue from #1407 dedicated specifically to persistent drawing marks.

I _think_ I have a lead - it seems that the DrawingTask model's `reset()` (which _does_ clear all the marks from the store) is called _after_ the InteractionLayer component already renders all the visible drawing marks.

For now though, I'm going to give this PR a 👍 so we can close #1407.

PR #1422, opened 2020-01-16
Drawing: Add x, y, angle to drawing tool root

eatyourgreens:
Make (x, y, angle) common, optional properties for all marks, defining the position of the mark and rotation about that position.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
This has been pulled out of #1360 so that it can be reviewed separately.

eatyourgreens:
There's also overlap with #1410 because this adds tests to `DrawingToolRoot`.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
```diff
@@ -7,6 +7,7 @@ import AnnotationsStore from '@store/AnnotationsStore'
 
 const BaseMark = types.model('BaseMark', {
   id: types.identifier,
+  angle: types.maybe(types.number),
```
srallen: I think the `BaseMark` model should only have properties that are actually common to all marks. This causes the point mark annotation to be submitted with `angle: undefined` in the classification. I think this could potentially cause confusion in the downstream analysis (someone will ask why points have undefined angles).

eatyourgreens:
I totally agree. I was hoping undefined properties aren’t serialised but apparently not. 

I’ve got an idea for testing this using `types.compose` to add an angle so I’ll try that tomorrow and see what happens.

eatyourgreens:
The tests can be fixed by adding the angle to a generic mark object, so I'll just do that.
```js
mark.angle = -45
```

eatyourgreens:
This PR includes the changes from #1410 so I think that PR doesn't need review now.

Issue #1226, opened 2019-10-14
Support more than one annotation per task


eatyourgreens:
**Package**
lib-classifier

**Is your feature request related to a problem? Please describe.**
Workflows allow tasks to be repeated more than once (eg. [Roll the Credits](https://www.zooniverse.org/projects/sroosa/roll-the-credits)), storing more than one annotation for each task.

Annotations are mapped in the classification using the task key as an identifier, which allows only one annotation per task per classification.
https://github.com/zooniverse/front-end-monorepo/blob/b3581665747c059d382f2df00c9fc7fdae12bb70/packages/lib-classifier/src/store/annotations/Annotation.js#L3-L5

**Describe the solution you'd like**
`classification.annotations` could be stored as an array, or each annotation could be assigned its own identifier.

eatyourgreens:
Drawing tools tasks which share a common text input task might run into this problem too.

eatyourgreens:
One solution for drawing tools would be to give each mark its own annotations map (or array.) I think we should be guided here by whichever format works best for tools that consume classifications, like Caesar.

srallen:
Making `classifications.annotation` an array sounds good. Then `addAnnotation` could use `lastIndexOf` to get the correct annotation for the current task key if it already exists? I think I initially used the map thinking ahead for the annotation persistence feature (if the annotation already exists, just update it the existing one) but didn't realize this would effectively make task reuse impossible since keys would need to be unique in a map. 

I was trying to think out how we can get it to work with the map, but it would mean touching all the code for the annotation models, so it'd be way more involved and risky as well as probably necessitating a conversion for down stream use, so I don't think we should try to go this route.

eatyourgreens:
Sounds good to me. `lastIndexOf` is what I've used in PFE for exactly the same problem.

Off the top of my head, a map would require annotation identifiers, then tracking those identifiers (maybe through the task model?) so that we know which annotation you are currently editing. I'm not 100% sure how the last bit would work.

The downside of the array, I think, is that it might be harder to move backwards and forwards through it, particularly on a workflow like Notes from Nature where there are multiple tasks per step, and we want to preserve those annotations.

Branching question workflows might be tricky if you go back, then go forward choosing a new branch for your next annotation. Nothing insurmountable though, and we know that we already have this working with arrays in PFE.

eatyourgreens:
>The primary goal of identifiers is not validation, but reconciliation and reference resolving.
[...]

PR #1434, opened 2020-01-19
Decouple task models from classifications

### LINKED ISSUE #1239: RFC: drawing sub-task

### LINKED ISSUE #1439: Task area stories are broken in the classifier storybook
**Package**
lib-classifier

**Describe the bug**
```
[mobx-state-tree] Error while converting `{
  "stepKey":"S1",
  "taskKeys":["init"],
  "tasks":{
      "T1":{"taskKey":"init","required":false,"answers":[{"label":"yes"},{"label":"no"}],"help":"Choose an answer from the choices given, then press Done.","question":"Is there a cat?","type":"single"}
  }
}` to `Step`
```

**To Reproduce**
Open the classifier storybook and navigate to the Tasks section.

eatyourgreens:
Includes #1432 but I thought this was big enough to deserve its own PR.

- Removes hard-coded references to the classification store from task models. Instead, `task.setAnnotation(annotation)` is used to set a reference to the task annotation, which could have come from a classification or a drawing mark with subtasks. Required tasks without annotations are automatically incomplete.

- Adds unique IDs to annotations (`annotation.id`) so that each annotation can be referenced from a task while it is being worked on.

- Tests are updated to remove mock classification stores. Instead, mock tasks are passed references to mock annotations via `task.setAnnotation()`.

- Task area component tests that rely on mock tasks with annotations are updated to mock a classifications store and pass references to active classification annotations into the mock tasks.

- Task area stories are fixed by mocking a classifications store and a workflow steps store before rendering the task area.

- Drawing task specs are updated to add a test for subtasks where we store single choice question answers on three mock points.

Package:
lib-classifier

Closes #1239 .
Fixes #1439.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
This model fails when you try to simulate real subtasks by adding annotations to multiple points which share the same task. 😞 
```
Error: [mobx-state-tree] Cannot resolve a reference to type 'SingleChoiceAnnotation' with
id: 'T3.0.0' unambigously, there are multiple candidates: /task/tools/0/marks/point1/annotations/T3.0.0,
/task/tools/0/marks/point2/annotations/T3.0.0
```

So the annotation identifier has to be unique in the tree, not just its path?

eatyourgreens:
I've updated the drawing tools Readme here to show how we might add text to multiple lines of text on the page. I'd appreciate feedback on that API.

eatyourgreens:
>Adds unique IDs to annotations (annotation.id) so that each annotation can be referenced from a task while it is being worked on.

This is a breaking change to the Annotation model, so needs testing against PH-TESS.

This PR (big as it is) could also add tests to the light curve viewer to check that annotation works as expected.

srallen:
@shaunanoordin are you available to review this? Thanks!

shaunanoordin:
Yikes, I should have started with this one instead of focusing on the drawing task reviews today - sorry! OK, this is now on my radar and I'll get a review done first thing next week. First things first, I'm going to update this branch.

packages/lib-classifier/src/plugins/tasks/models/Annotation.js
```diff
@@ -1,7 +1,8 @@
 import { types } from 'mobx-state-tree'
 
 const Annotation = types.model('Annotation', {
-  task: types.identifier,
+  id: types.identifier,
+  task: types.string,
```
shaunanoordin: Note to self: confirm that the changes from` task:identifier` -> `id.identifier, task:string` are consistent

packages/lib-classifier/src/plugins/tasks/models/Task.js
```diff
@@ -1,32 +1,33 @@
...
   required: types.maybe(types.boolean),
   type: types.literal('default')
 })
   .views(self => ({
-    get annotation () {
-      const { classifications } = getRoot(self)
-      const { annotation } = classifications || {}
-      const currentAnnotation = annotation ? annotation(self) : self.defaultAnnotation
-      return currentAnnotation
-    },
```
shaunanoordin: Note to self: since this PR removes the `.annotations` view from the general Task model, it would mean that the _Task model_ is no longer responsible for _initialising its own default annotation._

(Unless I'm missing some hidden initialisation with `annotation: types.maybe(types.safeReference(Annotation))`)

Which might mean that it's _theoretically possible_ for `myTask.annotation` to return null or undefined (or something equally wonky) if the _new_ "initialising a default annotation" step goes awry.

So: the responsibility for "initialising" a default annotation value has been moved to `src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js`, so check this bit to see if it's possible to screw this up.

- Can the annotation initialiser (`const annotation = addAnnotation(task) ; task.setAnnotation(annotation)`) be called _0 times?_ (Thoughts: mmm... probably not. Would there ever be a scenario where you want to have an instance of a Task _model..._ but not want to render the Task _component?_ )

- Can the annotation initialiser be called _multiple times,_ causing multiple default annotations? (Thoughts: uh... what can cause the `Tasks` component to re-render? Multi-Task, multi-Step workflows tested fine in early tests, but what if you go back/forth between Tasks?)

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
```diff
@@ -51,8 +51,11 @@ class Tasks extends React.Component {
       return (
         <Box as='form' gap='small' justify='between' fill>
           {tasks.map((task) => {
-            addAnnotation(task)
-            const annotation = classification.annotation(task)
+            // classifications.addAnnotation(task, value) retrieves any existing task annotation from the store
+            // or creates a new one if one doesn't exist.
+            // The name is a bit confusing.
+            const annotation = addAnnotation(task)
+            task.setAnnotation(annotation)
```
shaunanoordin: Note to self: remember, what's essentially happening here is that the _"initialise default Annotations for this Task"_ (or _"initialise a default instance of the Annotation model for this instance of a Task model"_ ) action has been moved out of the Task _model_ itself and into the Task _component._

This is fine in theory, but keep an eye out for that tricksy React render() and pay attention to when it's called.

packages/lib-classifier/src/store/AnnotationsStore.js
```diff
@@ -7,17 +7,24 @@ const AnnotationsStore = types
...
   .views(self => ({
     annotation (task) {
-      return self.annotations.get(task.taskKey) || task.createAnnotation()
+      let taskAnnotation = task.createAnnotation()
+      self.annotations.forEach(annotation => {
+        if (annotation.task === task.taskKey) {
+          taskAnnotation = annotation
+        }
+      })
+      return taskAnnotation
```
shaunanoordin: [minor][curiosity question] This block in theory should be functionally the same as the line it's replacing - get the annotation object for the specific task, and if there's no annotation object, create one.

Out of curiosity, why use `annotations.forEach(...)` instead of the simpler `annotations.get(taskKey)`? Wait, does the old line of code, `self.annotations.get(task_key_that_does_not_exist)` return a weird MobX `EmptyObjectOrWhatever` instead of `undefined` or `null`, causing the `||` operator to never fork to `task.createAnnotation()` ?

shaunanoordin:
Initial thoughts:

- Code looks good, though it did take me a while to grok the changes.
  - One major change is basically _where_ the default annotations for a Task (model) are initialised, I think - instead of having it done in the Task model itself, it's done externally by the Tasks component.
- Initial testing looks good, in the sense that I'm able to submit a bunch of Classifications that look legit, across a number of differently configured Workflows.
  - Workflow with a single Yes-No question: 👌 
    Screen Shot 2020-02-03 at 18 34 25
  - Workflow with a single Drawing Task, with multiple Drawing Tools (Line and Rectangle only, Point is non-functional due to a small bug addressed in a separate PR): 👌 
    Screen Shot 2020-02-03 at 18 33 51
  - Workflow with a Yes-No Task, a Drawing Step (multiple tools), and a Text Task: 👌 
    Screen Shot 2020-02-03 at 18 37 32
    - Note: one-way only, I didn't press the Back button to retry answers.

Initial thoughts look good, though I should run the Classifier through a few more tests to see if I can break the "initialise default annotations" action.

packages/lib-classifier/src/plugins/tasks/models/Task.js
eatyourgreens: `addAnnotation(task, value)` is really badly named. It doesn't add a new annotation for `task` if one already exists: it replaces the existing one. In fact, the classification can only have one annotation per task. There's an issue tracking this: #1226.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
eatyourgreens: `task.defaultAnnotation` still exists, I think. The thing to remember is where that annotation is stored. Originally, all annotations were stored in `classification.annotations` but that's not true now that we have subtasks. A Task component might be passed an annotation from `mark.annotations` instead.

packages/lib-classifier/src/store/AnnotationsStore.js
eatyourgreens: `map.get(identifier)` can only be used with identifiers. If you try `annotations.get(taskKey)` MobX State Tree will error because `taskKey` isn't an annotation ID.

Task keys aren't identifiers any more because each task model only exists once in the tree. However, that task may have created many annotations eg. a text sub task which annotates 50 lines of text will have 50 annotations, all with the same task key.

I spent a lot of time reading and re-reading [Identifiers and References](https://mobx-state-tree.js.org/concepts/references) trying to get my head around this. An identifier must always resolve to a single, unique node in the tree, as far as I can tell.

This block is not quite the same as the original. It loops through the annotation collection and returns the last annotation for `task.taskKey`, if there are multiple annotations for that task.

eatyourgreens: >Task keys aren't identifiers any more because each task model only exists once in the tree. 

I could change the tree so that each task can have multiple instances, one for each time you perform a task eg. one task model per line of text. However, I think that would require dismantling the classifier and rebuilding a large chunk of it. Only one task per task key is an assumption that's built quite deeply into the classifier model at the moment.

packages/lib-classifier/src/plugins/tasks/models/Task.js
eatyourgreens: If you don't specifically pass an annotation to `myTask`, with `myTask.setAnnotation`, then `myTask.annotation` will be undefined and the task won't be able to do any work.

packages/lib-classifier/src/plugins/tasks/models/Annotation.js
shaunanoordin: 👌

packages/lib-classifier/src/plugins/tasks/models/Task.js
shaunanoordin: Thanks for the clarification 👌 The fact that `addAnnotation()` _overwrites_ existing annotations makes me feel better about the "multiple added annotations" (now a non-)issue, but having only one annotation per task is its own thing.

Anyway, I tried tracing when `ClassificationStore.addAnnotation` gets triggered when navigating on different kind of Workflows and everything _looks_ legit, though I'm still scratching my head as to the _'how'_ things work in some cases. (e.g. _"Wait, how is ClassificationStore.addAnnotation() called with a non-undefined 2nd argument/annotationValue? I definitely see it happening when I complete the current active Task and submit/move to the next task, but I don't see where in the code has ClassificationStore.addAnnotation(task, someNonEmptyValue)")_

Not worth discussing my thoughts at the moment though, as I still need to digest, and this PR still needs to get merged soon.

packages/lib-classifier/src/store/AnnotationsStore.js
shaunanoordin: 👌 Yoiks! This was more involved than I expected. Thanks for highlighting the issue with Identifiers and Resources for MST.

PR #1435, opened 2020-01-19
Remove source maps from production

eatyourgreens:
Remove source maps from the production config for lib-react-components and lib-classifier. This should speed up build times and may improve download times.

Package:
lib-classifier
lib-react-components

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/webpack.dist.js
+++ b/packages/lib-classifier/webpack.dist.js
@@ -8,7 +8,6 @@ const EnvironmentWebpackPlugin = new webpack.EnvironmentPlugin({
 })
 
 module.exports = {
-  devtool: 'source-map',
   entry: './src/components/Classifier/index.js',
   externals: [
     '@zooniverse/grommet-theme',
--- a/packages/lib-react-components/webpack.dist.js
+++ b/packages/lib-react-components/webpack.dist.js
@@ -3,7 +3,6 @@ const path = require('path')
 const PeerDepsExternalsPlugin = require('peer-deps-externals-webpack-plugin')
 
 module.exports = {
-  devtool: 'source-map',
   entry: './src/index.js',
   mode: 'production',
   module: {


eatyourgreens:
I'm seeing faster builds, locally, but the Lighthouse performance score is still terrible and I'm seeing 9s JS execution times before the classifier is ready in Chrome.

srallen:
Wouldn't removing source maps make it harder to debug?

eatyourgreens:
As far as I can tell, they aren't in use in production. I don't see the browser making a request for them on the live PH-TESS project.

srallen:
I don't get the same scores as you. Yes, there's room for improvement, but I think more analysis needs to be done.

Screen Shot 2020-01-20 at 10 47 50 AM

eatyourgreens:
Here's what I get for PH-TESS.
Lighthouse audit of the PH-TESS Classify page, with a performance score of 2.

eatyourgreens:
>I think the intention was to wire them up for error logging to sentry, which hasn't ever been finished / isn't working for some reason.

`main.map.js` is a little over 3MB for the classifier, so that's a large extra download to add if you want Sentry to use source maps in the browser.

srallen:
If the goal is to improve performance, then I don't think removing source maps are going to help since I do think the end goal is to use them. The difference in scores between you and I are stark, so I think more research is needed here.

eatyourgreens:
The difference is probably hardware.

srallen:
The large source map size could be a symptom of something else.

eatyourgreens:
Source maps are large by definition. Performance advice is generally to use them in staging/development but never on a production site.

srallen:
> Source maps are large by definition. Performance advice is generally to use them in staging/development but never on a production site.

That's not really true anymore and they only get loaded when you open up devtools. [Should I Use Source Maps in Production?](https://css-tricks.com/should-i-use-source-maps-in-production/) 

There is however a bunch of webpack options we aren't using or aren't using effectively. We may want to try out the `nosources-source-map` option for production.
https://webpack.js.org/configuration/devtool/#production

There are many other things like this that we aren't doing, so that's why I suggest more research into it.

eatyourgreens:
That’s true. We should at least find out if `main.map.js` is being published to the server with the classifier. 

This PR doesn’t change browser performance but it makes building the classifier and ZRC much faster.

eatyourgreens:
From memory, classifier errors can’t be debugged in dev tools from the Classify page of the NextJS app. I could be wrong about that, but it would mean the source map isn’t being used. I’ll try an experimental local build that generates a stack trace and see what happens.

srallen:
Yes, there are some configuration issues which might be why it's not working with Sentry either. It's been on my back burner for a while to get error reporting to Sentry working.

So back to this PR, I don't think speeding up build times is really an issue that needs resolving ahead of actual app performance and error logging. The build happens on merge to master and the actual production deployment is fast when the production release tag is updated.

eatyourgreens:
Running a production build, with the dev tools open, the classifier source map is not being loaded. I can open an issue for that.
Classifier stack trace for Galaxy Zoo, showing that errors are not mapped back to the classifier source code.

eatyourgreens:
Sure. The source maps issue is #1438. Dynamic imports is a completely separate thing though. My bad for mentioning download times here.

eatyourgreens:
Actually, I would recommend removing source maps until we've fixed #1438. The fix for that can add them back in. So maybe keep this open.

srallen:
I think the problem is in next.js, not the classifier, with regard to source map loading. Removing the source maps from the classifier I think will make it harder to debug the problem.

eatyourgreens:
Good point. I want to upgrade NextJS to 9.2, but maybe wait until IMLS is done first.

Issue #1443, opened 2020-01-22
Drawing: saving and retrieving classifications to/from Panoptes


eatyourgreens:
**Package**
lib-classifier

As per [ADR-25](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-25.md#annotation-json-structure-1) we will need to transform classifications into a format that is useful to Panoptes and Caesar before making the POST request.

The transformation will be:

  1. `classification.annotations` is converted from a map to an array.
  2. For each drawing task in the `annotations` array, we loop through each `mark` in `annotation.value`. For each  `mark`:
    1. For each `annotation` in `mark.annotations`:
       1. Add `markIndex` to the annotation.
       2. Add the annotation to `classification.annotations`.
       3. Add an array item to `mark details` consisting of the annotation task key: `{ task }`.
   2. Delete `mark.annotations`.
   3. Delete any private (underscored) mark properties.

On retrieving an incomplete classification from Panoptes, we will need to reverse the above transformation in order to construct a Classification model, with Annotations, from the API response JSON.

In the short term, this transformation can be handled by the `classifications.completeClassification` action. Longer term, should we consider building an adapter  that sits between the classifier and Panoptes, and can convert classifications to and from API responses?

Time estimate: 1 week

srallen:
> classification.annotations is converted from a map to an array.

MST has a built-in feature if you use `toJS` that will remove the observables and convert maps to objects. There is a parameter to disable that and when you do, those data types stay maps. We disable the conversion to objects because maps can be converted to objects *or* to an array of pairs. The latter is sometimes preferable if you need to maintain an ordered structure. For classification annotations, we call `toJS` with the option to convert to an object disabled since the order is important:

https://github.com/zooniverse/front-end-monorepo/blob/29463a6f52e2f321fb6200aca118242ca5519099/packages/lib-classifier/src/store/ClassificationStore.js#L177

Then we manually convert it to an array of the map's values

https://github.com/zooniverse/front-end-monorepo/blob/29463a6f52e2f321fb6200aca118242ca5519099/packages/lib-classifier/src/store/ClassificationStore.js#L179

The utility function: https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/utils/convertMapToArray.js

So point number one in the list is already done.

> Add an array item to mark details consisting of the annotation task key and the mark index: { task, markIndex }

[...]

PR #1403, opened 2020-01-10
Tasks: add task.reset() and step.reset()

eatyourgreens:
Add `step.reset()` and `task.reset()` to reset steps and tasks for a new subject. Call `step.reset()` for each workflow step on subject ready.

I've also added notes about the components and models to the README for tasks.

Package:
lib-classifier

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #1450, opened 2020-01-27
Upgrade MobX State Tree to 3.15

eatyourgreens:
Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/package.json
+++ b/packages/lib-classifier/package.json
@@ -95,7 +95,7 @@
     "mobx": "~5.9.0",
     "mobx-devtools-mst": "~0.9.21",
     "mobx-react": "~6.1.4",
-    "mobx-state-tree": "~3.11.0",
+    "mobx-state-tree": "~3.15.0",
     "mocha": "~6.1.4",
     "mst-middlewares": "~3.14.0",
     "nock": "~10.0.6",
--- a/packages/lib-classifier/src/store/AnnotationsStore.js
+++ b/packages/lib-classifier/src/store/AnnotationsStore.js
@@ -11,12 +11,12 @@ const AnnotationsStore = types
     }
   }))
   .actions(self => {
-    function addAnnotation (task, annotationValue) {
+    function addAnnotation (task, value) {
       const annotation = self.annotation(task)
       // new annotations must be added to this store before we can modify them
       self.annotations.put(annotation)
-      if (annotationValue !== undefined) {
-        annotation.value = annotationValue
+      if (value !== undefined) {
+        annotation.update(value)
       }
     }
 
--- a/packages/lib-classifier/src/store/Classification.js
+++ b/packages/lib-classifier/src/store/Classification.js
@@ -34,6 +34,13 @@ const ClassificationMetadata = types.model('ClassificationMetadata', {
   }),
   workflowVersion: types.string
 })
+  .actions(self => ({
+    update (newMetadata) {
+      Object.keys(newMetadata).forEach(key => {
+        self[key] = newMetadata[key]
+      })
+    }
+  }))
 
 const Classification = types
   .model('Classification', {
--- a/packages/lib-classifier/src/store/ClassificationStore.js
+++ b/packages/lib-classifier/src/store/ClassificationStore.js
@@ -76,6 +76,7 @@ const ClassificationStore = types
 
       const tempID = cuid()
       const newClassification = Classification.create({
+        annotations: {},
         id: tempID, // Generate an id just for serialization in MST. Should be dropped before POST...
         links: {
           project: project.id,
--- a/packages/lib-classifier/src/store/ClassificationStore.js
+++ b/packages/lib-classifier/src/store/ClassificationStore.js
@@ -140,7 +140,7 @@ const ClassificationStore = types
 
       if (validClassificationReference && validSubjectReference) {
         const classification = self.active
-        const subjectDimensions = toJS(getRoot(self).subjectViewer.dimensions)
+        const subjectDimensions = toJS(getRoot(self).subjectViewer.dimensions, { recurseEverything: true })
 
         const metadata = {
           finishedAt: (new Date()).toISOString(),
@@ -154,7 +154,7 @@ const ClassificationStore = types
 
         const feedback = getRoot(self).feedback
         if (feedback.isValid) {
-          metadata.feedback = toJS(feedback.rules)
+          metadata.feedback = toJS(feedback.rules, { recurseEverything: true })
         }
 
         // TODO store intervention metadata if we have a user...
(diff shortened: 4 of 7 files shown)


eatyourgreens:
This was a failed attempt to get snapshot processors working for #1443 but I think I figured out why `completeClassification` broke 3.15 and put in a fix. I also made `ClassificationMetadata` responsible for its own updates while trying to debug broken classifications.

eatyourgreens:
It’s trying to set the value of a dead annotation?

eatyourgreens:
`annotation.update()` in #1432 was supposed to avoid errors like this:
https://github.com/zooniverse/front-end-monorepo/blob/6e5aa3f96a7dff7dec4dbe197350911c6f4a10a9/packages/lib-classifier/src/plugins/tasks/models/Annotation.js#L13-L15

Sounds like I missed some actions that set `annotation.value` directly. 🤔

eatyourgreens:
After updating `addAnnotation` to use `annotation.update()` I'm seeing the same error on the PH-TESS test project.
```
Uncaught Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree. (Object type: 'DataVisAnnotation', Path upon death: '/classifications/resources/ck6207gev00013h609kbrd1zl/annotations/T1', Subpath: '', Action: '/classifications/resources/ck6207gev00013h609kbrd1zl/annotations/T1.update()'). Either detach nodes first, or don't use objects after removing / replacing them in the tree.
```

I think it's trying to add that annotation to the classification that was just submitted, not the new one.

eatyourgreens:
I've done a bit of logging and tracing with the PH-TESS test project. What I'm seeing is that the second classification is retrieving the annotation from the first classification when it calls `self.annotations.get('T0')`. Could this be another issue caused by us using identifiers that are not unique in the tree? I'm not sure how to debug further.

srallen:
@eatyourgreens is this upgrade necessary for any of the features being worked on right now? If not, then could we close this and open an issue to track this and the issue that came up? We will need to upgrade at some point, but maybe not right now since there is a complication.

eatyourgreens:
3.15 fixes some bugs in snapshot processing and in observable maps. I suspect the maps change is what is causing problems here.

eatyourgreens:
I’m slightly concerned that stale annotations are persisting after a new subject loads.

srallen:
Yes, I understand the update may fix some issues in MST, but that doesn't answer my question, so I'll try to be more exact: 
1) Is this necessary for #1475?
2) should this be fixed and reviewed before #1475 ? 
3) is this a blocker for anything else?
4) if not, then does this need to happen now?

srallen:
I'm asking because the last time we got blocked on a minor version bump for MST, it was a fairly massive overhaul of nearly all the stores, so I'm trying to evaluate how critical this is.

eatyourgreens:
The processors in #1475 should really be using 3.12 (or 3.13?) or greater but they seem to be working on 3.11 so 🤷🏻‍♂️.

That bug with stale annotations should be double checked on master to make sure old annotations are properly cleared before classification begins on a new subject.

eatyourgreens:
In terms of priority, #1434 is absolutely top priority (we can't transcribe without it), then #1475 (so that we can save transcriptions) then this, (possibly behind anything else that may come up in the meantime.)

I am very worried about those stale annotations, though.

eatyourgreens:
Fixed it!

`classification.annotations` is never explicitly cleared when creating a new  classification. I believe this is a bug that is present on master too.

I'll push the fix. It's one extra line of code.

srallen:
Great catch! I should be able to review this tomorrow.

eatyourgreens:
I caught it because #1475 was complaining that the initial classification snapshot, on creation, didn't have `annotations` defined.

eatyourgreens:
Has the stale annotations bug affected PH-TESS at all? I get very worried when I see buggy behaviour that could affect submitted classifications. Particularly here, since this is part of the code that looks like it is untested on master.

srallen:
Ok, so since we know this is on master, we should determine when the bug was introduced and not couple the fix to an upgrade which introduces new variables here.

srallen:
Ok, based on discussion on slack and some testing on master, it's unlcear if this is actually an issue or not on master.

srallen:
The annotations being undefined doesn't make sense because it's set to `types.map` on the classification model and that type automatically sets a types.optional with a default.

srallen:
Unfortunately this is still having an issue with stale annotations. I would guess the source is the decoupling of the annotations model from classifications?

Seeing on a drawing task with my test project http://localhost:8080/?project=1844 with the second subject:
```
Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[observer(observed)]' Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree. (Object type: 'DrawingAnnotation', Path upon death: '/classifications/resources/ck69p236c00043h60dl1datxq/annotations/ck69p1zsa00013h60h4zaziud', Subpath: 'task', Action: '/workflowSteps.resetSteps()'). Either detach nodes first, or don't use objects after removing / replacing them in the tree.
```

I've been doing a lot of manual testing comparing to master, this branch, and so test branches I've been creating and I've only replicated these kind of MST errors on this branch. If there is an issue on master, then I haven't replicated it. Following the trace, the error is originating by the `tool.marks.clear()` of the `DrawingTask`'s `reset` action. The classifications store used to be solely responsible for reseting itself and subsequently annotations, but now that the annotations model can have different parent nodes, I _think_ what's happening is that it's not entirely being reset. If that is related to what's going on here, I would expect to see this on master, though.

srallen:
Possibly related to https://github.com/zooniverse/front-end-monorepo/pull/1403 refactor but until this can be reliably replicated on master, it's still uncertain. Maybe the observer should be for classification submission, not on subject ready?

eatyourgreens:
I think classification submission renders the new subject with the marks from the previous subject, because that classification hasn’t been reset yet.

srallen:
The ClassificationsStore does get reset with new subjects, but newly separate AnnotationsStore does not. What I think needs to happen is that the AnnotationsStore needs to observe its parent and reset itself if the parent resets itself. Because this might be an issue on master, this should be done in a separate PR. I agree the upgrade can happen later. Let's close this for now? In a later PR, I'd recommend upgrading the mobx dependency in parallel because sometimes there are underlying bugfixes in the core mobx library and MST relies on the observable functionality from that.

srallen:
I setup most of the original observers between the stores, so happy to take that on.

eatyourgreens:
I forgot to copy the reset over when I separated annotations out. That’s on me.

eatyourgreens:
Regarding versions, should we aim to match versions with the project app?
https://github.com/zooniverse/front-end-monorepo/blob/6bbd8c1ae2670cd4342c4c0971ce47c960a3d887/packages/app-project/package.json#L41-L44

PR #1457, opened 2020-01-30
Add mark specific callback action for onPointerUp

### LINKED ISSUE #1454: Drawing: new marks can't be completed
**Package**
lib-classifier

**Describe the bug**
Pointer capture by the drawing canvas (#1420) means that the `onFinish` handler (#1369) is never called for new marks. Points and Lines can be started but not finished.

srallen:

Package: lib-classifier

Closes #1454 

Describe your changes:
I've added to the mark models a specific `onPointerUp` action that then can call a callback if applicable. In this case, it is to call `onFinish` from the `InteractionLayer` for points and lines, but not the transcription line. I'm not sure where to add specs for this since the mark models don't have them it seems. 

With the dev classifier (`yarn dev` in `/pacakges/lib-classifier`)
The point and line can be tested here: http://localhost:8080/?project=1233

The transcription line can be tested here: http://localhost:8080/?project=1764&workflow=3392

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/models/marks/Line/Line.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/models/marks/Line/Line.js
@@ -73,10 +73,15 @@ const LineModel = types
       self.y2 = y2
     }
 
+    function onPointerUp(event, callback) {
+      callback(event)
+    }
+
     return {
       initialDrag,
       initialPosition,
       move,
+      onPointerUp,
       setCoordinates
     }
   })
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
@@ -68,9 +68,14 @@ function InteractionLayer ({ activeDrawingTask, activeTool, disabled, height, ma
     target.releasePointerCapture(pointerId)
   }
 
+  function onPointerUp (event) {
+    activeMark.onPointerUp(event, onFinish)
+  }
+
   return (
     <g
       onPointerMove={onPointerMove}
+      onPointerUp={onPointerUp}
       touch-action='none'
     >
       <StyledRect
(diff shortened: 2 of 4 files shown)


srallen:
oh, @shaunanoordin, does this need to be added to the rectangle mark model?

eatyourgreens:
The problem was caused by pointer capture so I've opened #1465 to revert that commit.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Line/Line.js
```diff
@@ -73,10 +73,15 @@ const LineModel = types
       self.y2 = y2
     }
 
+    function onPointerUp(event, callback) {
```
eatyourgreens: This can be added to the generic Mark model, if it's the same for all Marks.

srallen: It's not though. `TranscriptionLine` doesn't call the callback and I'd prefer to have each mark have its own ability to be flexible. Once we migrate all the drawing tools, then we can consider abstraction?

srallen:
I'm a little confused about why we should revert #1465 as that implements a functionality that we'll need for the drawing tools. This PR adds a means to support both the requirements for the transcription line, the ability to draw partial shapes from #1465, and adds a needed flexibility to the mark models. Is there a plan to implement the functionality supported by #1465 in another way?

> The hook is missing for the Rectangle tool, but adding it to the general Mark model (instead of all the specific drawing mark models) should be a better solution.

I disagree. I'd rather have the flexibility for potential differentiation between marks and how they handle certain events before making an abstraction. We haven't moved over all the drawing tools yet. 

Happy to make the addition to rectangle and rename the action though. :+1:

eatyourgreens:
This PR keeps the event handlers that were broken by #1420 but adds a new event listener to work around those broken handlers. #1474 fixes the broken handlers.

I think what we want is #1474 plus a new hook that fires when initial mark creation is complete. At the moment we have `initialPosition` (called when we place a mark) and `initialDrag` (called during the first dragging gesture.) We'd like a third hook that fires when mark creation is finished.

srallen:
I've reviewed and approved #1474. Feel free to borrow whatever ideas from this that are useful.

PR #1475, opened 2020-02-03
Drawing: Add classification snapshot processors

eatyourgreens:
Package:
lib-classifier

Towards #1443.

Use snapshot processors to serialise classifications as per [ADR 25](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-25.md#annotation-json-structure-1).

Add a new `toSnapshot` view to annotations and classifications, which returns a snapshot where all references have been resolved to full snapshots.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
@@ -1,3 +1,4 @@
+import { getSnapshot } from 'mobx-state-tree'
 import Mark from './Mark'
 import { Tool } from '@plugins/drawingTools/models/tools'
 
@@ -5,11 +6,16 @@ describe('Models > Drawing Task > Mark', function () {
   let mark
 
   before(function () {
-    mark = Mark.create({ id: 'test', toolType: 'default' })
+    mark = Mark.create({ toolType: 'default' })
   })
 
   it('should exist', function () {
     expect(mark).to.be.ok()
+  })
+
+  it('should have an id', function () {
+    expect(mark.id).to.exist()
+    expect(mark.id).to.be.a('string')
   })
 
   it('should have a toolIndex', function () {
@@ -190,6 +196,22 @@ describe('Models > Drawing Task > Mark', function () {
           expect(drawingTool.isComplete).to.be.true()
         })
       })
+    })
+  })
+
+  describe('snapshots', function () {
+    let snapshot
+
+    before(function () {
+      snapshot = getSnapshot(mark)
+    })
+
+    it('should not have an ID', function () {
+      expect(snapshot.id).to.be.undefined()
+    })
+
+    it('should have an annotations array', function () {
+      expect(snapshot.annotations).to.be.a('array')
     })
   })
 })
--- a/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
@@ -66,7 +66,7 @@ describe('Models > Drawing Task > Mark', function () {
   })
 
   it('should be able to store annotations', function () {
-    expect(mark.annotations).to.be.ok()
+    expect(mark.annotations).to.be.a('map')
   })
 
   describe('getDistance', function () {
--- a/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
@@ -5,6 +5,41 @@ import { Tool } from '@plugins/drawingTools/models/tools'
 describe('Models > Drawing Task > Mark', function () {
   let mark
 
+  function mockMark (options) {
+    const defaultOptions = {
+      required: false
+    }
+    const { required } = options || defaultOptions
+    const toolData = {
+      color: '#ff0000',
+      label: 'Point',
+      max: '10',
+      min: 1,
+      type: 'default'
+    }
+    const details = [
+      {
+        type: 'multiple',
+        question: 'which fruit?',
+        answers: ['apples', 'oranges', 'pears'],
+        required
+      },
+      {
+        type: 'single',
+        question: 'how many?',
+        answers: ['one', 'two', 'three'],
+        required
+      }
+    ]
+    const drawingTool = Tool.create(Object.assign({}, toolData, { details }))
+    const multipleTaskSnapshot = Object.assign({}, drawingTool.details[0], { taskKey: 'multiple' })
+    const singleTaskSnapshot = Object.assign({}, drawingTool.details[1], { taskKey: 'single' })
+    const multipleTask = drawingTool.createTask(multipleTaskSnapshot)
+    const singleTask = drawingTool.createTask(singleTaskSnapshot)
+    const mark = drawingTool.createMark({ id: 'mockMark' })
+    return { drawingTool, mark, multipleTask, singleTask }
+  }
+
   before(function () {
     mark = Mark.create({ toolType: 'default' })
   })
@@ -60,38 +95,12 @@ describe('Models > Drawing Task > Mark', function () {
 
   describe('with subtasks', function () {
     let mark
-    const toolData = {
-      color: '#ff0000',
-      label: 'Point',
-      max: '10',
-      min: 1,
-      type: 'default'
-    }
 
     describe('with incomplete, optional tasks', function () {
       let drawingTool
 
       before(function () {
-        const details = [
-          {
-            type: 'multiple',
-            question: 'which fruit?',
-            answers: ['apples', 'oranges', 'pears'],
-            required: false
-          },
-          {
-            type: 'single',
-            question: 'how many?',
-            answers: ['one', 'two', 'three'],
-            required: false
-          }
-        ]
-        drawingTool = Tool.create(Object.assign({}, toolData, { details }))
-        const multipleTaskSnapshot = Object.assign({}, drawingTool.details[0], { taskKey: 'multiple' })
-        const singleTaskSnapshot = Object.assign({}, drawingTool.details[1], { taskKey: 'single' })
-        drawingTool.createTask(multipleTaskSnapshot)
-        drawingTool.createTask(singleTaskSnapshot)
-        mark = drawingTool.createMark({ id: 'mockMark' })
+        ({ drawingTool, mark } = mockMark())
       })
 
       it('should be complete', function () {
@@ -107,26 +116,7 @@ describe('Models > Drawing Task > Mark', function () {
       let drawingTool
 
       before(function () {
-        const details = [
-          {
-            type: 'multiple',
-            question: 'which fruit?',
-            answers: ['apples', 'oranges', 'pears'],
-            required: false
-          },
-          {
-            type: 'single',
-            question: 'how many?',
-            answers: ['one', 'two', 'three'],
-            required: true
-          }
-        ]
-        drawingTool = Tool.create(Object.assign({}, toolData, { details }))
-        const multipleTaskSnapshot = Object.assign({}, drawingTool.details[0], { taskKey: 'multiple' })
-        const singleTaskSnapshot = Object.assign({}, drawingTool.details[1], { taskKey: 'single' })
-        drawingTool.createTask(multipleTaskSnapshot)
-        drawingTool.createTask(singleTaskSnapshot)
-        mark = drawingTool.createMark({ id: 'mockMark' })
+        ({ drawingTool, mark } = mockMark({ required: true }))
       })
 
       it('should be incomplete', function () {
@@ -145,26 +135,7 @@ describe('Models > Drawing Task > Mark', function () {
       let singleTask
 
       before(function () {
-        const details = [
-          {
-            type: 'multiple',
-            question: 'which fruit?',
-            answers: ['apples', 'oranges', 'pears'],
-            required: true
-          },
-          {
-            type: 'single',
-            question: 'how many?',
-            answers: ['one', 'two', 'three'],
-            required: true
-          }
-        ]
-        drawingTool = Tool.create(Object.assign({}, toolData, { details }))
-        const multipleTaskSnapshot = Object.assign({}, drawingTool.details[0], { taskKey: 'multiple' })
-        const singleTaskSnapshot = Object.assign({}, drawingTool.details[1], { taskKey: 'single' })
-        multipleTask = drawingTool.createTask(multipleTaskSnapshot)
-        singleTask = drawingTool.createTask(singleTaskSnapshot)
-        mark = drawingTool.createMark({ id: 'mockMark' })
+        ({ drawingTool, mark, multipleTask, singleTask } = mockMark({ required: true }))
       })
 
       it('should be incomplete', function () {
@@ -200,9 +171,14 @@ describe('Models > Drawing Task > Mark', function () {
   })
 
   describe('snapshots', function () {
+    let multipleChoiceAnnotation
+    let singleChoiceAnnotation
     let snapshot
 
     before(function () {
+      const { mark, multipleTask, singleTask } = mockMark()
+      singleChoiceAnnotation = mark.addAnnotation(singleTask, 1)
+      multipleChoiceAnnotation = mark.addAnnotation(multipleTask, [0, 2])
       snapshot = getSnapshot(mark)
     })
 
@@ -211,7 +187,10 @@ describe('Models > Drawing Task > Mark', function () {
     })
 
     it('should have an annotations array', function () {
-      expect(snapshot.annotations).to.be.a('array')
+      expect(snapshot.annotations).to.deep.equal([
+        singleChoiceAnnotation.toSnapshot(),
+        multipleChoiceAnnotation.toSnapshot()
+      ])
     })
   })
 })
--- a/packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
+++ b/packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
@@ -21,38 +21,42 @@ describe('Model > DrawingAnnotation', function () {
   })
 
   describe('annotation snapshots', function () {
+
+    function buildMockAnnotation ({ taskSnapshot }) {
+      const drawingTask = DrawingTask.create(taskSnapshot)
+      const annotation = drawingTask.createAnnotation()
+      // TODO: build a more realistic model tree here.
+      types.model('MockStore', {
+        drawingTask: DrawingTask,
+        annotation: DrawingAnnotation
+      })
+      .create({
+        drawingTask,
+        annotation
+      })
+      return { annotation, drawingTask }
+    }
+
     describe('without subtasks', function () {
       let snapshot
 
       before(function () {
-        const drawingTask = DrawingTask.create({
-          instruction: 'draw things!',
-          taskKey: 'T0',
-          tools: [
-            {
-              type: 'point'
-            }
-          ],
-          type: 'drawing'
+        const { annotation, drawingTask } = buildMockAnnotation({
+          taskSnapshot: {
+            instruction: 'draw things!',
+            taskKey: 'T0',
+            tools: [
+              {
+                type: 'point'
+              }
+            ],
+            type: 'drawing'
+          }
         })
         const pointTool = drawingTask.tools[0]
-        const point = pointTool.createMark({
+        pointTool.createMark({
           x: 50,
           y: 100
-        })
-        const annotation = DrawingAnnotation.create({
-          id: 'drawing1',
-          task: 'T0',
-          taskType: 'drawing',
-          value: [ point ]
-        })
-        const mockStore = types.model('MockStore', {
-          drawingTask: DrawingTask,
-          annotation: DrawingAnnotation
-        })
-        .create({
-          drawingTask,
-          annotation
         })
         snapshot = annotation.toSnapshot()
       })
@@ -95,29 +99,22 @@ describe('Model > DrawingAnnotation', function () {
       let snapshot
 
       before(function () {
-        const drawingTask = DrawingTask.create({
-          instruction: 'draw things!',
-          taskKey: 'T0',
-          tools: [
-            {
-              type: 'point',
-              details: [{
-                type: 'single',
-                answers: ['yes', 'no'],
-                question: 'yes or no?'
-              }]
-            }
-          ],
-          type: 'drawing'
-        })
-        const annotation = DrawingAnnotation.create(drawingAnnotationSnapshot)
-        const mockStore = types.model('MockStore', {
-          drawingTask: DrawingTask,
-          annotation: DrawingAnnotation
-        })
-        .create({
-          drawingTask,
-          annotation
+        const { annotation, drawingTask } = buildMockAnnotation({
+          taskSnapshot: {
+            instruction: 'draw things!',
+            taskKey: 'T0',
+            tools: [
+              {
+                type: 'point',
+                details: [{
+                  type: 'single',
+                  question: 'Yes or no?',
+                  answers: [ 'yes', 'no' ]
+                }]
+              }
+            ],
+            type: 'drawing'
+          }
         })
         const pointTool = drawingTask.tools[0]
         const point1 = pointTool.createMark({
--- a/packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
+++ b/packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
@@ -54,7 +54,7 @@ describe('Model > DrawingAnnotation', function () {
           drawingTask,
           annotation
         })
-        snapshot = annotation.toSnapshot
+        snapshot = annotation.toSnapshot()
       })
 
       it('should contain exactly one annotation', function () {
@@ -127,7 +127,7 @@ describe('Model > DrawingAnnotation', function () {
         const questionTask = pointTool.tasks[0]
         point1.addAnnotation(questionTask, 0)
         point2.addAnnotation(questionTask, 1)
-        snapshot = annotation.toSnapshot
+        snapshot = annotation.toSnapshot()
       })
 
       it('should contain the task annotation plus one annotation for each subtask', function () {
--- a/packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
+++ b/packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
@@ -1,19 +1,187 @@
+import { types } from 'mobx-state-tree'
+import DrawingTask from './DrawingTask'
 import DrawingAnnotation from './DrawingAnnotation'
 import { Point } from '@plugins/drawingTools/models/marks'
 
-const point = Point.create({ id: 'mockAnnotation', frame: 0, toolIndex: 0, toolType: 'point', x: 100, y: 150 })
-
-const drawingAnnotationSnapshot = {
-  id: 'drawing1',
-  task: 'T0',
-  taskType: 'drawing',
-  value: [ point.id ]
-}
-
 describe('Model > DrawingAnnotation', function () {
+  const point = Point.create({ id: 'mockAnnotation', frame: 0, toolIndex: 0, toolType: 'point', x: 100, y: 150 })
+
+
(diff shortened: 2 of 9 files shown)


eatyourgreens:
Flattening `classification.annotations` to a single array isn't implemented yet but should be possible by returning an array of annotations (not a single annotation) from `drawTaskAnnotation.toSnapshot`.

One thought that occurs to me re. the flattened annotation structure recommended in ADR 25 is that the Panoptes classifications export looks up annotation task keys in `workflow.tasks`, in order to get task questions etc. Will the subtask task keys break the classification exports, since those tasks don't exist in `workflow.tasks`?

eatyourgreens:
#1434 adds the ability to mock classifications with subtask annotations for tests, which could then be used to test this PR once it's merged.

eatyourgreens:
Yeah, the logs are in there to double check that the processors work as advertised. 

Turns out MST generates a lot of snapshots. Every time a model changes as far as I can tell. 

I was hoping someone could help me understand why the task nav buttons have broken. Those tests are completely rewritten in #1431. I’m hoping that will fix them here too. 

Anyway, does using snapshot processors look like a good approach?

eatyourgreens:
Or #1434, even, for the task nav tests.

eatyourgreens:
Tests should hopefully pass now that #1434 is merged in.

packages/lib-classifier/src/store/Classification.js
```diff
@@ -46,6 +47,43 @@ const Classification = types
     }),
     metadata: types.maybe(ClassificationMetadata)
   })
+  .views(self => ({
+    get toSnapshot () {
```
srallen: Is there any chance this will have an invalid cached version of a classification since this is using a getter? We might not want or need this to be pre-computed.

packages/lib-classifier/src/store/Classification.js
```diff
@@ -46,6 +47,43 @@ const Classification = types
...
+    return newSnapshot
+  })
+  .postProcessSnapshot(snapshot => {
+    console.log('Classification postprocessor')
+    const newSnapshot = Object.assign({}, snapshot)
+    // remove temporary classification IDs
+    // TODO: leave the ID if it came from Panoptes
+    delete newSnapshot.id
+    // convert annotations to an array
+    newSnapshot.annotations = Object.values(snapshot.annotations)
```
srallen: Why not use the `convertMapToArray` helper already written and tested?

packages/lib-classifier/src/store/Classification.js
eatyourgreens: i don't know enough about views to say. I thought they were computed each time the model changed?

packages/lib-classifier/src/store/Classification.js
eatyourgreens: Does that work on plain objects? I'm not sure if the snapshot of an observable map is a JS Map.

eatyourgreens:
>how exactly is the snapshot serialized and how does it function compare to toJS?

My understanding is that MST snapshots are immutable and give a snapshot of the model state after any transactions are complete (ignoring both volatile state and views.) You can pass a snapshot to `model.create(snapshot)` but that's not true of a JS object created by `toJS`. I don't even know if those are immutable, or guaranteed to respect MST transactions. Do they clone the original object or copy by reference?

The pre-processors are lifecycle hooks but I haven't looked into limiting when they run. It does seem, to me, like we really want a Panoptes snapshot, or Caesar snapshot, which is only generated when exported to those systems.

[`types.snapshotProcessor`](https://mobx-state-tree.js.org/API/#snapshotprocessor) in the most recent versions of MST looks interesting, but we'd probably need to upgrade from 3.11. 3.11 has them, but they don't seem fully developed yet.

eatyourgreens:
Actually, the use of `toJS`, rather than snapshots, on master might possibly explain why we were seeing references to stale annotations persist after a classification was complete.

packages/lib-classifier/src/store/Classification.js
eatyourgreens: I just tried it and it throws `TypeError: map.values is not a function`.

eatyourgreens:
>Will the subtask task keys (T0.0.0 etc.) break the classification exports, since those tasks don't exist in workflow.tasks?

The answer to this is apparently yes, the flattened array will break classification exports, so that needs to be added to the consequences of ADR 25.

srallen:
> Actually, the use of toJS, rather than snapshots, on master might possibly explain why we were seeing references to stale annotations persist after a classification was complete.

Do you have a more specific explanation of how this would even be possible? It's not even clear to me we have stale annotations happening on master.

eatyourgreens:
>Do you have a more specific explanation of how this would even be possible? It's not even clear to me we have stale annotations happening on master.

I'm trying to remember what I was thinking but it was very, very late at night and I'm still sick so slightly woozy. I think my line of thought was that `toJS` might copy by reference, in which case stale references could still be hanging around in RAM. I'm not sure.

packages/lib-classifier/src/store/Classification.js
srallen: Ok, so this sort of answers my question and is one of the concerns I have about using snapshots. We have no choice in how the observable map is converted into javascript with the snapshot, whereas `toJS` gives us that choice. The snapshot converts the observable map to just an object. We had a long discussion about how to manage the conversion, which resulted in ADR 10: https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-10.md

srallen: Is there a plan for stripping underscored values. For instance, there's an id submitted with drawing annotations that would have no use downstream, so this should be underscored and removed:

Screen Shot 2020-02-05 at 3 13 04 PM

packages/lib-classifier/src/store/Classification.js
```diff
@@ -46,6 +47,38 @@ const Classification = types
...
+    }
+  }))
+  .preProcessSnapshot(snapshot => {
+    const newSnapshot = Object.assign({}, snapshot)
+    // generate classification IDs, if not present
+    newSnapshot.id = snapshot.id || cuid()
+    // convert any annotations arrays to a map
+    const annotationsMap = {}
+    if (snapshot.annotations && Array.isArray(snapshot.annotations)) {
+      snapshot.annotations.forEach(annotation => annotationsMap[annotation.task] = annotation)
```
srallen: The naming of `annotationsMap` is a little misleading because it's not actually being converted into a map here, just a object. Since order of the annotations is important, how about converting into an array of pairs:

`const arrayOfPairs = [['T1', { taskType: 'single', value: 0 }], ['T2', { taskType: 'text', value: 'hello' }]]` 

then use `new Map(arrayOfPairs)`?

packages/lib-classifier/src/store/Classification.js
srallen: My understanding is that it will be recomputed if one of the state values that the derivations uses changes, but I find this to be risky in this case. We only really want the final state of the classification to be retrieved here. I think all we have to do is remove `get` and this will no longer be cacheable. My comment applies for all the models using this pattern.

eatyourgreens: Good point. 🤔

packages/lib-classifier/src/store/Classification.js
eatyourgreens: IDs are stripped by the post-processors. See classifications and marks for examples. I'll add that to annotations too.

You can double-check yourself, but `getSnapshot` should preserve the order of maps. I'll add tests to verify that annotation order is not changed.

packages/lib-classifier/src/store/Classification.js
eatyourgreens: Worth trying but this is creating the new snapshot that we then pass into `classification.create(snapshot)` during `createClassification()`.

packages/lib-classifier/src/store/Classification.js
eatyourgreens: I'm not seeing those IDs now that I've fixed the annotation post-processor.
A completed classification with annotation IDs stripped from drawing and question task annotations.

packages/lib-classifier/src/store/Classification.js
eatyourgreens: I think MST accepts an array of pairs as a snapshot for a new map, but I'll double check.

I could add tests for this as well. It was low priority for me because we aren't actually loading in an incomplete classification from Panoptes yet.

eatyourgreens:
By the way, the lifecycle hooks are listed here in the new docs: https://mobx-state-tree.js.org/overview/hooks

packages/lib-classifier/src/store/Classification.js
eatyourgreens: Transforming the input snapshot to an array of pairs gives this error 😞 
```
at path "/annotations" snapshot 
`[["T1",{"task":"T1","taskType":"single","value":1}],
["T0",{"task":"T0","taskType":"text","value":"Hello"}]]` 
is not assignable to type: 
`map<string, (DataVisAnnotation | DrawingAnnotation | 
MultipleChoiceAnnotation | SingleChoiceAnnotation | 
TextAnnotation)>` (Value is not a plain object)
```

I could have sworn that the MST docs say you can create a map from an array of `[key, value]` pairs.

eatyourgreens: Using a map
```
Map {
  'T1' => { task: 'T1', taskType: 'single', value: 1 },
  'T0' => { task: 'T0', taskType: 'text', value: 'Hello' }
}
```
also throws during `classification.create()`.

```
at path "/annotations" snapshot `{}` is not assignable to type: 
`map<string, (DataVisAnnotation | DrawingAnnotation | MultipleChoiceAnnotation | 
SingleChoiceAnnotation | TextAnnotation)>` (Value is not a plain object),
```

eatyourgreens: Oooh, annotations are mapped by `annotation.id` now, not `annotation.task`. This is going to be interesting, since we strip IDs when we save to Panoptes.

eatyourgreens: This hasn't been a problem for us to date because we only ever create classifications with an empty annotations collection.

eatyourgreens: I've added a couple of mock Panoptes annotations to the classification tests, along with a test to check that the array order is preserved after `classification.create(snapshot)`.

eatyourgreens:
Re. the lifecycle hooks, are you suggesting we should detach or destroy the classification before saving it to Panoptes?

srallen:
> Re. the lifecycle hooks, are you suggesting we should detach or destroy the classification before saving it to Panoptes?

Yeah, in `ClassificationsStore` something like `completeClassification` -> `detach(childClassificationModel)` -> `beforeDetach` lifecycle which serializes to prepare for post -> `submitClassification` --> adds to queue. It might not work, but I _think_ it's the right MST lifecycle method if we only want this to happen once (specifically for post). `afterAttach` would be the one to use for pre-processing just once on get incomplete. 

There are the snapshot helper methods, i.e. `getSnapshot`, etc, that we can still use here. 

This is a pretty different approach, though, so what we can do is finish reviewing this as is, and have a separate comparison branch if it works?

eatyourgreens:
It's an interesting idea but I agree we should experiment on a different branch.

Performance-wise, I _think_ we're ok because processors can only take in a plain object, generate a new plain object from that then return the new object. So there's not a lot of slow processing that can happen unless the model snapshots happen to be big or complex for some reason. Mostly I'm using the post-processor to strip IDs from models. I guess the transformation might slow down if there are a lot of annotations, maybe? I could add a test that generates 100 – 200 marks, say, each with one annotation attached. That might be reasonable to simulate a transcription subject.

I think all the complicated transformations here happen during `classification.toSnapshot()`, and we have total control over when that runs.

Re. the lifecycle hooks, for model creation the order will be something like:
`snapshot > preProcessSnapshot(snapshot) > model.create(processedSnapshot) > afterAttach()`
so `afterAttach` might run too late to be useful.

You're right, though, that postProcessors run every time state changes, so on every action. The reason being that MST caches a new snapshot of the current state every time that a model changes.

eatyourgreens:
CI failed because the Docker image couldn't be built. I've restarted the Travis build.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
```diff
@@ -192,5 +198,21 @@ describe('Models > Drawing Task > Mark', function () {
...
+    })
+
+    it('should not have an ID', function () {
+      expect(snapshot.id).to.be.undefined()
+    })
+
+    it('should have an annotations array', function () {
+      expect(snapshot.annotations).to.be.a('array')
+    })
+  })
```
srallen: There should be tests for `applySnapshot` as well, that an id is generated and the annotations array is converted to a map?

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
```diff
@@ -192,5 +198,21 @@ describe('Models > Drawing Task > Mark', function () {
...
+      snapshot = getSnapshot(mark)
+    })
+
+    it('should not have an ID', function () {
+      expect(snapshot.id).to.be.undefined()
+    })
+
+    it('should have an annotations array', function () {
+      expect(snapshot.annotations).to.be.a('array')
+    })
```
srallen: Could we test that the annotations are ordered as expected in the outputted snapshot? It makes me nervous that we can't choose how the snapshot serializes maps.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
```diff
@@ -1,19 +1,187 @@
...
+          x: 50,
+          y: 100
+        })
+        const annotation = DrawingAnnotation.create({
+          id: 'drawing1',
+          task: 'T0',
+          taskType: 'drawing',
+          value: [ point ]
+        })
+        const mockStore = types.model('MockStore', {
```
srallen: We should be careful with mocked parent stores. I ran into a lot of trouble with mocks like these in tests back when I updating the classifier to MST 3.11.0 a few months ago. I'd prefer an actual integration test with the real parents.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
```diff
@@ -1,19 +1,187 @@
...
+                type: 'single',
+                answers: ['yes', 'no'],
+                question: 'yes or no?'
+              }]
+            }
+          ],
+          type: 'drawing'
+        })
+        const annotation = DrawingAnnotation.create(drawingAnnotationSnapshot)
+        const mockStore = types.model('MockStore', {
```
srallen: Same comment about mocking parent stores.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
```diff
@@ -1,19 +1,187 @@
...
+        })
+        .create({
+          drawingTask,
+          annotation
+        })
+        snapshot = annotation.toSnapshot()
+      })
+
+      it('should contain exactly one annotation', function () {
+        expect(snapshot).to.have.lengthOf(1)
```
srallen: I'd like the expectation here to check the type since snapshots are doing type conversions. Strings can also have length. I think with chai, these can be chained, so `expect(snapshot).to.be.an('array').that.has.lengthOf(1)`

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
```diff
@@ -1,19 +1,187 @@
...
+          y: 200
+        })
+        const questionTask = pointTool.tasks[0]
+        point1.addAnnotation(questionTask, 0)
+        point2.addAnnotation(questionTask, 1)
+        snapshot = annotation.toSnapshot()
+      })
+
+      it('should contain the task annotation plus one annotation for each subtask', function () {
+        expect(snapshot).to.have.lengthOf(3)
```
srallen: Same comment here about explicitly checking for an array type too.

packages/lib-classifier/src/plugins/tasks/models/Annotation.js
```diff
@@ -1,13 +1,30 @@
...
+import cuid from 'cuid'
+import { getSnapshot, types } from 'mobx-state-tree'
 
 const Annotation = types.model('Annotation', {
   id: types.identifier,
   task: types.string,
   taskType: types.string
 })
+  .preProcessSnapshot(snapshot => {
+    const newSnapshot = Object.assign({}, snapshot)
```
srallen: Just checking, is this necessary because the serialization is immutable? Maybe to make this clear, the variable should be named `mutableSnapshot`?

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
eatyourgreens: I thought I'd added a test for that. Did I only test classifications?

packages/lib-classifier/src/plugins/tasks/models/Annotation.js
eatyourgreens: Yeah, you can't change `snapshot`. You have to create a completely new object.

>Note 2: The preProcessSnapshot transformation must be pure; it should not modify its original input argument!**_

https://mobx-state-tree.js.org/overview/hooks

eatyourgreens:
Working on #1491, @shaunanoordin has just discovered that the existing `toJS()` conversion doesn't include any entries in `mark.annotations` for any drawn marks. We can add an annotation to a mark from a subtask popup, but it won't appear in the final classification.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
eatyourgreens: The preprocessor should already be tested with `Mark.create(snapshot)` but I'll double check.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
eatyourgreens: The output order of Mark subtasks isn’t important in the final analysis. That’s why I tested classifications but didn’t test these.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
eatyourgreens: I don’t think the drawing annotation does any type conversion itself. ADR 25 says it should flatten the annotations array but I think that’s all that needs to be tested here. 

Other components convert maps to arrays but they have their own tests.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
```diff
@@ -1,19 +1,187 @@
...
+          value: 0,
+          markIndex: 0
+        }
+        const point2Answer = {
+          task: 'T0.0.0',
+          taskType: 'single',
+          value: 1,
+          markIndex: 1
+        }
+        expect(snapshot).to.deep.equal([ annotation, point1Answer, point2Answer ])
```
eatyourgreens: Here’s where we test that subtasks annotations are output from marks as expected.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
eatyourgreens: A very good rule to follow when testing is only write as much code as you need to get the job done. I’m very strongly against adding unnecessary complication to the tests. 

Here, a drawing annotation is a collection of references to a drawing task, so the test mocks a task and an annotation. There’s no dependence on any other models so they aren’t included.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
eatyourgreens: I've updated this to mock and test a mark with annotations. That said, `mark.annotations` is removed when we serialise the classification (see ADR 25) so what this is testing is that we shouldn't lose any annotations, I guess.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
eatyourgreens: `mark.id` already has a test but `mark.annotations` was only tested for existence. I've updated that to test for type.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
eatyourgreens: Just to be clear, this is testing the drawing annotation in isolation. An annotation on its own will throw invalid reference errors, so we test with an annotation and its task.

Integration tests could mock a `ClassificationStore` and a workflow (`WorkflowStepsStore`?), run through annotations for some mock task then complete the classification. That's outside the scope of the annotation tests but `TaskNavButtons` has code that does this: https://github.com/zooniverse/front-end-monorepo/blob/f4f704fea6f24009270d6df92e0887b2e882067f/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.spec.js#L21-L67

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
eatyourgreens: Drawing annotations don't do any type conversion. The snapshot is a flattened array of annotations, per ADR 25. I think we're getting confused here with classification snapshots, or maybe mark snapshots?

eatyourgreens: The set of tests after this do test the snapshot in detail, and will fail if it sin't an array.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
eatyourgreens: Again, this is testing an annotation in isolation.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
eatyourgreens: That's checked later in the code:
`expect(snapshot).to.deep.equal([ annotation, point1Answer, point2Answer ])`.

eatyourgreens:
I've updated the tests to be more strict about type-checking. Mocking a full classification and workflow, then testing the classifications that it creates, is outside the scope of the tests here, but a good idea and worth considering.

In that case, I'd like a test like that runs through annotating and classifying a subject on a mock of the PH-TESS workflow, as well as a test that runs through a mock transcription workflow. I'm not sure if we have any tests at the moment to alert us if the TESS classification workflow breaks (beyond testing the project manually on staging.) I'm not sure where the best place would be to add tests like those.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.spec.js
srallen: My comment has to do with type conversions and we should make sure the type conversion is doing what we expect it to independently in keeping with the points raised in ADR 10. MST isn't really giving us a choice in how the map conversion happens using snapshots so I wanted extra safety checks.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
srallen: My original comment here doesn't have to do with type conversion, so I'm not sure what to say about that.

I really don't think we should use mocks for stores at all. I completely disagree with that approach based on the experience I had upgrading to MST 3.11.0 and the massive test rewrite I had to do to figure out what was going wrong (the mock stores didn't reflect the reality of how stores were being used), the comments on how to do tests from the core maintainers of MST, and general reading on when to use mocks in testing and when not to. I've been moving away from doing this whenever I write store tests.

I won't block this PR over this, but I really think don't think we should be testing like this.

eatyourgreens: I'm not really sure how to test references any other way. The referenced model has to exist in the same tree somewhere. In this case that's the marks that make up the annotation.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
srallen: Ok, we disagree then.

srallen:
> In that case, I'd like a test like that runs through annotating and classifying a subject on a mock of the PH-TESS workflow, as well as a test that runs through a mock transcription workflow. I'm not sure if we have any tests at the moment to alert us if the TESS classification workflow breaks (beyond testing the project manually on staging.) I'm not sure where the best place would be to add tests like those.

The annotating for PH: TESS is coupled to the d3 subject viewer and I found writing tests for that to be pretty difficult because of d3's chaining API. I decided a better use of my time was to focus on building out the new scatter plot viewer and eventually port TESS to use that and continue to manually check TESS functionality. Feel free to have a look at it though.

eatyourgreens:
I'd prefer to wait on merging this until the `DrawingAnnotation` tests are fixed. I have tried writing the tests without the mock store but the annotation snapshot fails when it tries to resolve the annotation value here:
https://github.com/zooniverse/front-end-monorepo/blob/204590bf9085063f78520e4d9afb3b8729e5dca3/packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.js#L14-L15

eatyourgreens:
One possible solution would be that we store marks in two places: once on the drawing tools and copies of the marks (to avoid reference conflicts) on the annotation value. The tools edit their marks. When editing is finished, copies of the marks are stored on the annotation.

eatyourgreens:
>The annotating for PH: TESS is coupled to the d3 subject viewer and I found writing tests for that to be pretty difficult because of d3's chaining API. I decided a better use of my time was to focus on building out the new scatter plot viewer and eventually port TESS to use that and continue to manually check TESS functionality. Feel free to have a look at it though.

I was thinking mock up an instance of the TESS task model, add some mock annotations to it then test the resulting classification is as expected. Something like mocking a drawing task in the following code, but starting from a classification model and a workflow model. Create instances of those that mock PH-TESS, then run through the steps that the classifier would run through to create and annotate a classification. Test the resulting model state, maybe by generating a classification snapshot to submit to Panoptes:
https://github.com/zooniverse/front-end-monorepo/blob/851a5468d0f6d8dfe87db4edc8a6cede52f84496/packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingTask.spec.js#L98-L132

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.spec.js
eatyourgreens: >I really think don't think we should be testing like this.

That's fine but if you could share what you've been reading, and what you've learnt then the rest of us can learn from it too. At the very least, some guidance on how to test the models would be useful. Here, I have no idea how to test a view that resolves identifiers elsewhere in the tree. All I've got to go on is fairly sketchy advice from the MST forums.
https://spectrum.chat/mobx-state-tree/general/getsnapshot-with-reference-field~43a6e2d2-9655-4571-b599-9cac4ee9c94d
https://spectrum.chat/mobx-state-tree/general/getsnapshot-with-resolved-references~ba70fa0a-9c45-4cee-a2e0-5213d01132fc

srallen: I've shared this before, and this is a good place to start: https://github.com/mobxjs/mobx-state-tree/issues/1404 and specifically this comment: https://github.com/mobxjs/mobx-state-tree/issues/1404#issuecomment-544453565

PR #1491, opened 2020-02-06
Drawing SubTasks - Popup UI component and dedicated store

### LINKED ISSUE #1315: Drawing sub-task UI

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Fixes #1315

The aim of this PR is to add SubTasks for Drawing Tasks. e.g. for drawing tasks that says "use this Rectangle tool to mark an animal in this photo, then type the name of that animal in this sub task"

This PR consists of two elements:

- [ ] adding a resizable, movable popup to the UI that allows users to interact with each sub-task.
- [ ] modifying the stores so we can track the state of the popup and figuring out _which active drawing mark_ is being handled by the popup UI. 

Should look like:

Screen Shot 2020-02-06 at 18 56 09

### Status

Currently looks like:

Screen Shot 2020-02-06 at 18 51 50
_Technically, isn't an elephant just a cat with tusks and a trunk and large ears and a massive size?_

Very much a WIP


eatyourgreens:
Looking good! You should be able to get the tasks from `activeTool.tasks` which, I think, is available from the drawing task.

packages/lib-classifier/package.json
```diff
@@ -50,18 +51,11 @@
...
-    "@babel/plugin-proposal-optional-chaining": "~7.8.3",
-    "@babel/polyfill": "~7.7.0",
-    "@babel/preset-env": "~7.7.1",
-    "@babel/preset-react": "~7.7.0",
-    "@babel/register": "~7.7.0",
     "@babel/core": "~7.8.4",
     "@babel/plugin-proposal-class-properties": "~7.8.3",
     "@babel/plugin-proposal-decorators": "~7.8.3",
+    "@babel/plugin-proposal-object-rest-spread": "~7.6.2",
+    "@babel/plugin-proposal-optional-chaining": "~7.8.3",
```
srallen: I appear to have made a mistake in my rebase when updating dependencies earlier in the week. Could you submit a fix for this in a separate PR?

packages/lib-classifier/src/components/Classifier/components/SubTaskPopup/SubTaskPopup.js
```diff
@@ -0,0 +1,93 @@
...
+import SingleChoiceTask from '@plugins/tasks/SingleChoiceTask'
+
+import styled from 'styled-components'  // TODO: check what's the best way to style this component
+import zooTheme from '@zooniverse/grommet-theme'
+
+// EXPERIMENTAL
+// ----------------
+import taskRegistry from '@plugins/tasks'
+const StyledBox = styled(Box)`
+  border: 2px solid ${zooTheme.global.colors.brand}
```
srallen: Box has a border prop, so this shouldn't be necessary? https://v2.grommet.io/box#border

packages/lib-classifier/package.json
srallen: I opened https://github.com/zooniverse/front-end-monorepo/pull/1497 to fix this.

packages/lib-classifier/src/components/Classifier/components/SubTaskPopup/SubTaskPopup.js
```diff
@@ -0,0 +1,105 @@
...
+`
+// ----------------
+
+import en from './locales/en'
+
+counterpart.registerTranslations('en', en)
+
+function storeMapper (stores) {
+  
+  const { activeMark } = stores.classifierStore.subTaskPopup
```
eatyourgreens: I think `activeMark` is better stored on the drawing task model.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -9,8 +9,8 @@ const StyledRect = styled('rect')`
   cursor: ${props => props.disabled ? 'not-allowed' : 'crosshair'};
 `
 
-function InteractionLayer ({ activeDrawingTask, activeTool, disabled, height, marks, move, scale, width }) {
-  const [ activeMark, setActiveMark ] = useState(null)
+function InteractionLayer ({ activeDrawingTask, activeTool, disabled, height, marks, move, setActiveMark, scale, width }) {
+  const [ activeMark, setActiveMark_state ] = useState(null)
```
eatyourgreens: I'd remove `setActiveMark` from local state and move it to the drawing task model. Then call `setActiveMark(mark)`, or `activeDrawingTask.setActiveMark(mark)`, when selecting the active mark.

Similarly move the active mark to `drawingTask.activeMark` and get it from the store.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -87,9 +89,15 @@ function InteractionLayer ({ activeDrawingTask, activeTool, disabled, height, ma
...
-          onDelete={() => setActiveMark(null)}
+          onDelete={() => {
+            setActiveMark_state(null)
+            setActiveMark(null)
+          }}
           onFinish={onFinish}
-          onSelectMark={mark => setActiveMark(mark)}
+          onSelectMark={mark => {
+            setActiveMark_state(mark)
+            setActiveMark(mark)
```
eatyourgreens: Get rid of local state here too and set the active mark in the store, then reference it from the active drawing task.

packages/lib-classifier/src/components/Classifier/components/SubTaskPopup/SubTaskPopup.js
eatyourgreens: I've moved this to `activeDrawingTask.activeMark` in #1505.

packages/lib-classifier/src/components/Classifier/components/SubTaskPopup/SubTaskPopup.js
```diff
@@ -0,0 +1,131 @@
...
+  const { addAnnotation } = stores.classifierStore.classifications
```
eatyourgreens:
```suggestion
  const { addAnnotation } = activeMark
```

shaunanoordin:
Replaced by #1517, but I'm going to make a note of the feedback from this PR for the next.

- `activeDrawingTask.activeMark` has been taken care of, since #1517 is based after #1505 is merged 👍 
- Styles will be redone altogether, as I need to get a better grasp of the best way to use zooTheme and, if appropriate, the CSS helpers. (See more in 1517)

Issue #1482, opened 2020-02-05
RFC: should we have separate staging and production Docker images?


eatyourgreens:
**Package**
app-project
app-content-pages

In staging and production, the NextJS apps are both built from `zooniverse/front-end-monorepo:latest`. Here's the app-project Dockerfile as an example (content pages is pretty much exactly the same.)
https://github.com/zooniverse/front-end-monorepo/blob/c364e6ad457b5166af18e9789e201481569c6592/packages/app-project/Dockerfile#L1-L10

Our build process is:
- *staging*: Pull down the latest master commit, install and bootstrap the monorepo dependencies then push the resulting image to `zooniverse/front-end-monorepo:latest`. Build the app images from that base image.
- *production*: Pull down the `production-release` tag, install and bootstrap the monorepo dependencies then push the resulting image to `zooniverse/front-end-monorepo:latest`. Build the app images from that base image.

My question is: should we continue doing this or would it be safer to build separate base images for production and staging? I guess I'm worried about inadvertently pulling down an image with the latest HEAD of master when what I might really want is the latest production release.

eatyourgreens:
I guess the current process is fine as long as we always rebuild the base image before using it?

camallen:
I get why the boostrap image exists for the `yarn bootstrap` https://github.com/lerna/lerna/tree/master/commands/bootstrap

However I'm confused as to why the `production-apps` image contains both the builds for the `fe-project` and the `fe-content-pages` apps. These artefacts will be in both deployed app images and they really should be split out. 

It seems folks planned to use the combined `production-apps` image for running all services from the root (port 3000 and 3001), however they can use isolated docker builds as well. Ideally this combined images could be used for testing via travis but we don't use that way. Are we using this combined image any other way?

I think we're overcomplicating the root image build process here, we can rebuild the individual packages (using the bootstrap image as base) in both compose for dev and jenkins for deployment. 

Once we've split out the builds, we add docker image `ENV` values for switching to staging / production builds using via compose and jenkins cmds to ensure we're building the versions of the apps we specifically want. 

As it stands any merge to master builds one image for both staging and production deploys. We Can switch to using the tag release system we're using in almost all other repositories to isolate between master -> staging deploys and `production-release` -> production deploys. 

This isolation will allow us to build independent deploy env images and allow master work to progress without impacting staging deploys (can pick a point between release tag and master head to release features).

eatyourgreens:
[...]

PR #1494, opened 2020-02-07
Add build and default ENV vars for app dockerfiles

camallen:
Package: app-project & content-pages

Allow the builds to inject ENV vars from build env (jenkins or docker-compose) that can be used to prebuild the bundle with the correct env data values. This ensures our ENV vars for sentry etc can be injected at build / runtime. 

This also adds a APP_ENV for sentry env reporting to allow easier filtering of errors (production, staging, development etc).

This PR adds another way to run the project-app via local docker development. I think we can either look to reuse the base bootstrap image** as we currently do OR we look at scrapping the base bootstrap image and rebuild each time (needs some benchmarking to decide if there are benefits to using bootstrap).

** Longer term we should cycle on how we use docker & compose for local dev and testing. I think we want to build local bootstrap images with the deps installed (like it's doing now) vs pulling the latest image from Docker Hub. Perhaps we'll have to setup local dev Dockerfiles that differ slightly from deployed Dockerfiles. Out of scope for this PR though.

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
I'm slightly confused. The local dev setup doesn't pull images from DockerHub.

packages/app-project/docker-compose.yml
```diff
@@ -2,31 +2,41 @@ version: '3.4'
...
-    command: ["dev"]
+      context: .
+      args:
+        ASSET_PREFIX: https://fe-project.zooniverse.org
+        SENTRY_DSN: https://2a50683835694829b4bc3cccc9adcc1b@sentry.io/1492691
+    command: [yarn, "dev"]
+    environment:
+      - PORT=3000
+      - ASSET_PREFIX= http://localhost:3000
+      - SENTRY_DSN=https://2a50683835694829b4bc3cccc9adcc1b@sentry.io/1492691
```
camallen: This doesn't have to be set for local dev builds if the SENTRY DSN is passed into the build via `args` - I left it here to show it can be modified at runtime, especially using `yarn dev`

camallen:
> I'm slightly confused. The local dev setup doesn't pull images from DockerHub.

Yep - this is switching to use per package image builds (as the jenkins file does). The current development setup builds different images to what the jenkins CI tooling does. The current root dir docker-compose.yml builds all images from the production-apps build step and not the per package docker file. 

My current idea is to have a docker-compose.yml root file that builds the base image and then folks cd to the app directory to run the specific apps via `docker-compose build ..` and that will re-use the locally built bootstrap image for deps (this would be via a different Dockerfile.dev which uses `FROM locally_boostrapped_image:local`)

Or we keep the current setup but we have to understand that it's not really a reproduction of our CI builds. 

This setup i've got here (at least for project-app) would get the CI env vars injecting into the built images for our staging & production deployments and thus it should fix the error reporting issue we've got now.

eatyourgreens:
This PR breaks the local dev workflow with docker-compose. I'd really like to see it broken into two PRs. One to fix the undefined Sentry environment in the builds, and another to propose changes to how we develop locally with docker-compose.

>The current root dir docker-compose.yml builds all packages into the one image for reuse via the production-apps build step. It does not use the not the per package docker / compose files for running apps.

This is intentional. If I've made changes across multiple packages, I want to be able to build and test the bootstrapped monorepo against builds of the NextJS apps locally. This should be in the Readme, but `docker-compose build` and `docker-compose up` should bootstrap my changes and allow me to browse them locally in both the project app _and_ the content pages app. For example, I might have made changes to the theme, or shared components library, that need to be checked against both apps. Or I might need to check a new build of the classifier in the project app. See https://github.com/zooniverse/front-end-monorepo/pull/1496#issuecomment-584026371 for an example of this in use.

This PR, as far as I can tell, only allows us to check changes to `packages/app-project` locally.

>My current idea is to have a docker-compose.yml root file that builds the base image and then folks cd to the app directory to run the specific apps via docker-compose build .. and that will re-use the locally built bootstrap image for deps (this would be via a different Dockerfile.dev which uses FROM locally_boostrapped_image:local)

My understanding of [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) is that they're intended to avoid having to maintain separate Dockerfiles, which is why we're using them for local dev here.

packages/app-project/Dockerfile
```diff
@@ -1,7 +1,20 @@
 FROM zooniverse/front-end-monorepo:latest
```
eatyourgreens: This image is rebuilt on each production and staging build (see #1482), so can't be relied upon for local dev. Local dev should use a base image that's built from your local copy of the monorepo.

packages/app-project/docker-compose.yml
```diff
@@ -2,31 +2,41 @@ version: '3.4'
...
+    image: front-end-monorepo_project-app:local
     build:
-      context: ../../
-      target: bootstrap
-    entrypoint:
-      - "yarn"
-      - "workspace"
-      - "@zooniverse/fe-project"
-    command: ["dev"]
+      context: .
```
eatyourgreens: Same comment as above. The `app-project` Dockerfile works for a production deploy, but can't be relied upon for a local dev build.

packages/app-project/docker-compose.yml
```diff
@@ -2,31 +2,41 @@ version: '3.4'
...
+      - ASSET_PREFIX= http://localhost:3000
+      - SENTRY_DSN=https://2a50683835694829b4bc3cccc9adcc1b@sentry.io/1492691
+      - NODE_ENV=development
+      - PANOPTES_ENV=staging
+      - NEXT_TELEMETRY_DISABLED=1
+      - APP_ENV=development
     ports:
       - "3000:3000"
     volumes:
+      - project_app_node_modules_cache:/usr/src/packages/app-project/node_modules
```
eatyourgreens: `node_modules` should be installed to the root of the monorepo, not to each package individually.

eatyourgreens:
To make things more fun, the content pages app gets its environment from `packages/app-content-pages/.env`, if present. The Contentful API secrets are in there for local dev.

packages/app-project/Dockerfile
camallen: Yep - longer term for the dev / testing workflow we should look at how we can use a different Dockerfile to use locally built images vs relying on this boostrap image. 

This is used in the deployment builds so it's going to rely on that bootstrap image from the jenkins CI run. We can certainly revisit how this works but short term i'd like to leave this as is.

Jenkinsfile
```diff
@@ -42,11 +42,6 @@ pipeline {
         }
 
         stage('Build app Docker images') {
-          environment {
```
camallen: I'm injecting these via the Dockerfile ENV vars which are available in build, as such they don't need to be defined in the CI.

packages/app-project/docker-compose.yml
camallen: Agree - we can revist how these work for local dev.

FWIW the `FROM zooniverse/front-end-monorepo:latest` is the built image from the `bootstrap` stage in the base image. It runs `install` and `bootstrap` which install the app dependencies for resuse - https://github.com/lerna/lerna/tree/master/commands/bootstrap

packages/app-project/docker-compose.yml
camallen: thanks - i've modified the cache path. We can remove the cache if it doesn't prove useful for quick yarn install to add / update packages vs and a full rebuild

packages/app-project/docker-compose.yml
eatyourgreens: The bootstrap runs [`bootstrap.sh` here](https://github.com/zooniverse/front-end-monorepo/blob/master/bin/bootstrap.sh) too, so I think that image is also building the libraries.

Jenkinsfile
```diff
@@ -61,7 +56,8 @@ pipeline {
                   script {
                     def dockerRepoName = 'zooniverse/fe-content-pages'
                     def dockerImageName = "${dockerRepoName}:${GIT_COMMIT}"
-                    def newImage = docker.build(dockerImageName)
+                    def buildArgs = "--build-arg ASSET_PREFIX SENTRY_DSN ."
```
eatyourgreens: Should this be `"--build-arg ASSET_PREFIX --build-arg SENTRY_DSN ./"` I've been reading the docs and advice for build args but I'm not sure what the syntax is.

Jenkinsfile
```diff
@@ -81,7 +77,8 @@ pipeline {
                   script {
                     def dockerRepoName = 'zooniverse/fe-project'
                     def dockerImageName = "${dockerRepoName}:${GIT_COMMIT}"
-                    def newImage = docker.build(dockerImageName)
+                    def buildArgs = "--build-arg ASSET_PREFIX SENTRY_DSN ."
```
eatyourgreens: Same question here about build args syntax.

packages/app-project/docker-compose.yml
```diff
@@ -1,32 +1,42 @@
...
-      - "yarn"
-      - "workspace"
-      - "@zooniverse/fe-project"
-    command: ["storybook"]
+    image: front-end-monorepo_project-app:local
+    command: [yarn, "storybook"]
+    environment:
+      - NODE_ENV=development
+      - PANOPTES_ENV=staging
+      - NEXT_TELEMETRY_DISABLED=1
```
eatyourgreens: The storybook doesn't use NextJS, or Panoptes, so I think these aren't needed.

camallen: Good catch - will remove and update.

Jenkinsfile
camallen: yep - correct. they have to be specified each time.

Jenkinsfile
camallen: yep - they need to be specified for each env var. thanks i'll update

packages/app-project/docker-compose.yml
camallen: Thanks for pointing that bootstrap.sh out. 

I'm trying to come to grips with the various build tools we're using in the FEM, lerna and yarn and how these are working together for reuse across the apps.

packages/app-content-pages/docker-compose.yml
```diff
@@ -1,19 +1,28 @@
...
-      target: bootstrap
-    entrypoint:
-      - "yarn"
-      - "workspace"
-      - "@zooniverse/fe-content-pages"
-    command: ["dev"]
+      context: .
+      args:
+        ASSET_PREFIX: https://fe-content-pages.zooniverse.org
+        SENTRY_DSN: https://1f0126a750244108be76957b989081e8@sentry.io/1492498
```
camallen: I don't think these need to be in here, shall i remove them?

packages/app-project/docker-compose.yml
```diff
@@ -1,32 +1,40 @@
...
-      target: bootstrap
-    entrypoint:
-      - "yarn"
-      - "workspace"
-      - "@zooniverse/fe-project"
-    command: ["dev"]
+      context: .
+      args:
+        ASSET_PREFIX: https://fe-project.zooniverse.org
+        SENTRY_DSN: https://2a50683835694829b4bc3cccc9adcc1b@sentry.io/1492691
```
camallen: I don't think these need to be in here, shall i remove them?

packages/app-content-pages/Dockerfile
```diff
@@ -1,9 +1,21 @@
 FROM zooniverse/front-end-monorepo:latest
```
eatyourgreens: If I'm testing the NextJS app locally, is there a way that I can check which versions of the shared libraries are installed with this image?

For instance, can I use this base image for local development if I'm working on a PR that also adds new components to the zooniverse react components library? I'm not 100% clear whether this uses the local symlinks for installed dependencies, or whether it forces me to use the libs that were built during the most recent Jenkins build and pushed to `zooniverse/front-end-monorepo:latest`.

packages/app-project/Dockerfile
eatyourgreens: OK. I've got the same questions here as I had with the content pages app. If I'm making changes to the shared libs (eg. classifier, zooniverse react components or zoo theme) as part of a PR, will those changes be reflected in a local docker-compose build?

packages/app-project/docker-compose.yml
eatyourgreens: I think we're all still trying to come to grips with the build tools.

eatyourgreens:
Storybook builds seem to be broken on this branch.
```
storybook_1    | WARN Broken build, fix the error above.
storybook_1    | WARN
storybook_1    |  You may need to refresh the browser.
storybook_1    | 
storybook_1    | 
storybook_1    | WARN force closed preview build
```
I'll continue to investigate.

packages/app-project/Dockerfile
camallen: So these changes are for deployment / CI pipeline not for local dev work, as the only system using these package dockerfiles are the CI tools.

Testing changes / builds locally will still take place via the root docker-compose directives and the root dockerfile: 
- https://github.com/zooniverse/front-end-monorepo/blob/2234b9cae3f0fc967d430e1016b8636564b8a083/docker-compose.yml#L15
- https://github.com/zooniverse/front-end-monorepo/blob/2234b9cae3f0fc967d430e1016b8636564b8a083/Dockerfile#L22

> If I'm making changes to the shared libs (eg. classifier, zooniverse react components or zoo theme) as part of a PR, will those changes be reflected in a local docker-compose build?

If you use each packages docker-compose files they will **not** include your changes as the base image it's building off is the CI image not the locally built images. 

Instead for dev work / testing you can use the root docker-compose / dockerfile to rebuild the base image and each app package (as linked above which includes the bootstrap step for building react components and lib classifier).

packages/app-content-pages/Dockerfile
camallen: The easiest way to dev / test local builds is to use the project root dockerfile and compose
as it will rebuild a local image with the local deps and the project and content-pages app. 

Local deps will be based on the local bootstrap image and will including the bootstrapped local react components and lib classifier code as well.

eatyourgreens: The top-level Dockerfile bootstraps the monorepo and runs production builds of the NextJS apps (`yarn start`). For local dev work, I want to use a dev build of the NextJS app ie. `yarn dev` so that I get hot reloading for code changes.

eatyourgreens:
The storybook is broken but I'm not sure if that's been introduced by this pR or if it's broken on master.

I'm concerned about losing the ability to run local dev builds of the project and content app with `docker-compose`. That alone would be enough for me to reject this PR but I'd prefer to hear opinions from other people who work on the monorepo.

eatyourgreens:
Running the content pages app with `docker-compose up`, I get an error when I browse http://localhost.zooniverse.org:3000/about/team even though I have a `packages/app-content-pages/.env` set up with the Contentful API keys.

>There was an error fetching the Team data. Please refresh and try again.

packages/app-content-pages/Dockerfile
camallen: Right - so in the root compose file we currently call `yarn start`, we could switch that to use 'yarn dev' and inject the local package code into the running image for hot code reloading etc. Would that help with local testing etc?

FWIW the per package docker-compose files now use `yarn dev` and inject the local code to the container for hot code loading etc.
- https://github.com/zooniverse/front-end-monorepo/blob/27c3dfc186810f3fd112b4ca75352fef41350a79/packages/app-project/docker-compose.yml#L8
- https://github.com/zooniverse/front-end-monorepo/blob/27c3dfc186810f3fd112b4ca75352fef41350a79/packages/app-content-pages/docker-compose.yml#L22

ok, i think there is some confusion about my intentions with this PR. I'm trying to ensure our CI tools inject the correct build ENV data for the resulting production images.

I think the issue here is that our CI tool builds a bootstrap base image from the root Docker file (the rest of the multi stage docker file data is ignored). During CI we restrict concurrent builds to ensure the `zooniverse/front-end-monorepo:latest` image reflects our previous CI build step and then build both the project-app and content-pages package apps from their package docker files, these are isolated builds that we use in production. 

This CI process means that our root compose and dockerfiles aren't reproducing the production builds as they stand and most likely shouldn't be trying too either.

eatyourgreens: >Local deps will be based on the local bootstrap image and will including the bootstrapped local react components and lib classifier code as well.

If we're using local builds and dependencies, why is the build also pulling an image of the full monorepo down from dockerhub?

To test this out, I added the sentence "This is a local build" to the classifier code and rebuilt with `yarn build`.

Running `yarn dev` in `packages/app-project`, here's what I see for http://localhost.zooniverse.org:3000/projects/nora-dot-eisner/planet-hunters-tess/classify?env=production "This is a local build" shows, as expected.
The Planet Hunters TESS classifier with the words "This is a local build." above the subject viewer.

Shutting that down, then building and starting the project app with `docker-compose`, here's what I see. "This is a local build" doesn't show because I'm using the most recent copy of the classifier from the Jenkins build. This makes local dev unpredictable, depending on whether the most recent build was a staging or production build.
The Planet Hunters TESS classifier looking exactly as it does on the live web site.

eatyourgreens:
I think this PR might be trying to do too much for a single PR. Would it be possible to fix the Jenkins build environment and look at the changes to docker-compose separately?

packages/app-content-pages/Dockerfile
camallen: Again - this is a difference of CI / production builds using the freshly minted `zooniverse/front-end-monorepo:latest` image which doesn't have your code. We need this isolation for production builds in the per package Dockerfile. 

We can support your use case for local development work. My current train of thought is to provide that via the root dockerfile / compose files vs the per package ones. I chose the per package compose setup to prioritise testing of our production image builds not local dev work, we can change this.

camallen:
> think this PR might be trying to do too much for a single PR. Would it be possible to fix the Jenkins build environment and look at the changes to docker-compose separately?

We can, however most of the changes will come over as the main way to test those production build changes work as expected is via per package compose file changes as well.

packages/app-content-pages/Dockerfile
eatyourgreens: >Right - so in the root compose file we currently call yarn start, we could switch that to use 'yarn dev' and inject the local package code into the running image for hot code reloading etc. Would that help with local testing etc?

Not really, because we do also want that local production build of everything as a final check of any code changes. We have encountered bugs in the past that didn't show up when running `yarn dev` but did stop the project and content apps working when run with `yarn build && yarn start`.

Our current workflow is to run dev servers && apps in the local package directories, while changing code, then run a production build at the top level as a final check when we're happy with our feature branch.

camallen: ok then. We have to decide how to test the production image changes for CI / deployment. Currently each package's Dockerfile is not used apart from CI builds and deployments. They are not used for local testing anywhere else.

This uses the root dockerfile not the package one, https://github.com/zooniverse/front-end-monorepo/blob/2234b9cae3f0fc967d430e1016b8636564b8a083/docker-compose.yml#L15

The per package docker-compose also does not use the package dockerfile, it uses the root dockerfile https://github.com/zooniverse/front-end-monorepo/blob/2234b9cae3f0fc967d430e1016b8636564b8a083/packages/app-project/docker-compose.yml#L7

This is not good, we really should be ensuring our deployment dockerfiles are tested and working as we expect. Hence my changes in the package docker-compose files to support this. 

We need to ensure we support both ensuring our production images work as expected (and are tested) and local development workflows and testing. I think we need two root compose files:
- One for production builds and testing nextjs app via `yarn start` on each package
- One for local code development workflows via `yarn dev` on each package

camallen: Alternatively we can use a different per package testing setup (dev dockerfile / different compose setup) to ensure the local changes across apps are included in the setup. This would look similar to the current setup. I can switch the current compose setup for production testing explicitly. I'm happy to do this.

However we need to ensure we test both cases and ensure our deployment Dockerfiles are correctly setup, working as expected and we can test them locally.

eatyourgreens: To be honest, I would add the NextJS app builds back into the `bootstap.sh` script, then use one bootstrap image for everything. I think that's the simplest solution.

camallen: it still won't test the production dockerfile changes that i'm proposing in this PR, we need a way to test them.

camallen:
> Storybook builds seem to be broken on this branch.
> 
> ```
> storybook_1    | WARN Broken build, fix the error above.
> storybook_1    | WARN
> storybook_1    |  You may need to refresh the browser.
> storybook_1    | 
> storybook_1    | 
> storybook_1    | WARN force closed preview build
> ```
> 
> I'll continue to investigate.

This is what i just got on master using `docker-compose build` on root and `docker-compose up` from the app-project package, seems like it's missing the 'babel-loader' package.
```
storybook_1  |     + 317 hidden modules
storybook_1  | 
storybook_1  | ERROR in multi /usr/src/node_modules/@storybook/core/dist/server/common/polyfills.js /usr/src/node_modules/@storybook/core/dist/server/preview/globals.js ./.storybook/config.js /usr/src/node_modules/webpack-hot-middleware/client.js?reload=true&quiet=true
storybook_1  | Module not found: Error: Can't resolve 'babel-loader' in '/usr/src/packages/app-project'
storybook_1  |  @ multi /usr/src/node_modules/@storybook/core/dist/server/common/polyfills.js /usr/src/node_modules/@storybook/core/dist/server/preview/globals.js ./.storybook/config.js /usr/src/node_modules/webpack-hot-middleware/client.js?reload=true&quiet=true main[2]
storybook_1  | Child HtmlWebpackCompiler:
storybook_1  |                           Asset     Size               Chunks  Chunk Names
storybook_1  |     __child-HtmlWebpackPlugin_0  6.5 KiB  HtmlWebpackPlugin_0  HtmlWebpackPlugin_0
storybook_1  |     Entrypoint HtmlWebpackPlugin_0 = __child-HtmlWebpackPlugin_0
storybook_1  |     [../../node_modules/@storybook/core/node_modules/html-webpack-plugin/lib/loader.js!../../node_modules/@storybook/core/dist/server/templates/index.ejs] /usr/src/node_modules/@storybook/core/node_modules/html-webpack-plugin/lib/loader.js!/usr/src/node_modules/@storybook/core/dist/server/templates/index.ejs 2.1 KiB {HtmlWebpackPlugin_0} [built]
storybook_1  | 
storybook_1  | WARN Broken build, fix the error above.
storybook_1  | WARN You may need to refresh the browser.
```

eatyourgreens:
@camallen any idea why the content pages app has stopped working locally for me? https://github.com/zooniverse/front-end-monorepo/pull/1494#issuecomment-587471949

packages/app-content-pages/Dockerfile
eatyourgreens: If the base image builds the production JS files for the Next JS apps, we can get rid of the individual app images. Will we still need the per-app Dockerfiles to run "yarn start" when the base image is deployed to kubernetes?

camallen:
> @camallen any idea why the content pages app has stopped working locally for me? [#1494 (comment)](https://github.com/zooniverse/front-end-monorepo/pull/1494#issuecomment-587471949)

Most likely as the .env file isn't copied into the container via the compose setup and the base image from CI builds doesn't have those secrets in it. Adding an entry to inject that file via compose would be in here, https://github.com/zooniverse/front-end-monorepo/blob/2234b9cae3f0fc967d430e1016b8636564b8a083/packages/app-content-pages/docker-compose.yml#L17

eatyourgreens:
Ah right. I was confused because it does pick up the environment when I run `docker-compose build` and `docker-compose up` from master. I couldn't immediately see what had changed.
A local copy of the team page with content loaded from Contentful.

eatyourgreens:
Oh wait, you linked to code on master, which does work. I'm confused again.

packages/app-content-pages/Dockerfile
camallen: doing so will increase the bloat of each docker image (i.e. the project-app will have the code  for the content-pages app, etc).

> Will we still need the per-app Dockerfiles to run "yarn start" when the base image is deployed to kubernetes?
Not necessarily, we can use args via K8s to run the particular app code, https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#define-a-command-and-arguments-when-you-create-a-pod

I still don't think that's a great solution, i'm working on something that might alleviate the issues with dev / deployment workflows.

eatyourgreens: >doing so will increase the bloat of each docker image (i.e. the project-app will have the code for the content-pages app, etc).

That's not a big deal, to be honest, when you consider that each app image still has all the source code for the libraries etc.

camallen:
> Oh wait, you linked to code on master, which does work. I'm confused again.

apologies - posted the wrong link, i've updated.

packages/app-content-pages/Dockerfile
camallen: Something like this might help with allowing normally app testing as you've outlined and provide a way to test the production docker images / builds.
https://github.com/zooniverse/front-end-monorepo/pull/1522

eatyourgreens: 👍 I'll take a look.

Building a single base image would also have the advantage that we can tag it with a commit ID. The more I think about that idea, the more I like it.

srallen:
I agree with @camallen that we need a way to test the deployment builds. I would be supportive of having separate compose files for dev work. For dev workflows, I think we should minimize combining dev work that is impacting multiple libraries and apps. That to me is a sign that you need to break up your work into smaller units and submit smaller PRs that introduce a change in a single library.

packages/app-content-pages/Dockerfile
camallen: It might simplify the setup, we will still need a root and per package compose file for booting the apps locally for dev / testing workflows.

the single dockerfile works till we have different env vars etc, at that point we could re-split the root dockerfile up to something like we've got now.

srallen:
Whatever changes do happen here, do be sure to update the docs in the repository readme about deployment.

eatyourgreens:
That's a good point about docs. I think the project app [instructions for running a local production build](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/app-project#running-in-production) may have changed. Probably the content app instructions too.

packages/app-content-pages/Dockerfile
camallen: Ok - based on the latest changes from #1522 are you happy with your workflow for testing etc with the way this is now setup?

packages/app-content-pages/docker-compose-prod.yml
```diff
@@ -0,0 +1,24 @@
+version: '3.2'
```
eatyourgreens: Should this be '3.4'? 3.2 is a pretty old version of the compose file spec.

README.md
```diff
@@ -52,7 +52,7 @@ cd front-end-monorepo
 docker-compose build
 ```
 
-`docker-compose up` runs production builds of the project app at http://localhost:3000 and the content pages app at http://localhost:3001
+`docker-compose up` runs local builds of the project app at http://localhost:3000 and the content pages app at http://localhost:3001
```
eatyourgreens:
```suggestion
`docker-compose up` runs local production builds of the project app at http://localhost:3000 and the content pages app at http://localhost:3001
```

I think we need to be clear here that we're running production builds and not dev builds.

eatyourgreens:
Latest changes seem to have fixed #1499 for the content pages app.

👍

packages/app-content-pages/docker-compose-prod.yml
eatyourgreens: The build works, so I guess this isn't using any features from 3.4.

PR #1495, opened 2020-02-07
Reset action for wherever annotations are stored

srallen:

Package: lib-classifier

Describe your changes:
I've done a bit of refactoring based on the conversation in #1450. I wasn't able to reliably replicate the annotation dead node issue on master, but it's possible it's happening in two separate branches which indicates the issue might be on master. Regardless, I think this is a refactor that should happen as a result of previous refactors. Several models were missing reset actions. WIP because it still needs tests.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingTask.js
```diff
@@ -8,7 +8,7 @@ const toolModels = Object.values(tools)
 
 const Drawing = types.model('Drawing', {
   activeToolIndex: types.optional(types.number, 0),
-  annotation: types.maybe(types.safeReference(DrawingAnnotation)),
+  annotation: types.safeReference(DrawingAnnotation),
```
eatyourgreens: Does `types.safeReference` allow undefined values? The task annotation will be undefined until one is explicitly passed to the task from the classification.

packages/lib-classifier/src/store/AnnotationsStore.js
```diff
@@ -17,6 +17,20 @@ const AnnotationsStore = types
...
   .actions(self => {
+    function afterAttach() {
+      createClassificationObserver()
+    }
+
+    function createClassificationObserver() {
+      const classificationsDisposer = autorun(() => {
+        const classifications = getParent(self)
+        onAction(classifications, (call) => {
+          if (call.name === 'reset') self.reset()
```
eatyourgreens: I think `onAction` is expected to call either `next()` or `abort()` after interrupting an action.

eatyourgreens:
There’s no test to check that the active resource is undefined (or a valid reference?) after resetting a store. I’ll look at that on Monday.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingTask.js
eatyourgreens: Note to myself: I have been using `type.safeReference` completely wrong throughout the classifier.

packages/lib-classifier/src/store/AnnotationsStore.js
srallen: It's not interrupting an action. `onAction` just provides a means for adding a callback for actions: https://mobx-state-tree.js.org/API/#onaction

I think you're thinking of the handler for custom middleware: https://mobx-state-tree.js.org/concepts/middleware

srallen:
I would prefer tackling fixing the tests because I had a hard time following how the tree is working now and fixing the tests would improve my understanding.

packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingTask.js
srallen: Yes, it does. Wrapping it in `types.maybe` was redundant.

eatyourgreens: I've been using `safeReference` completely wrong, so I'll open a PR to fix that.

packages/lib-classifier/src/store/AnnotationsStore.js
eatyourgreens: I am! My bad.

So this calls `classifications.active.annotations.reset()` before `classifications.active` is cleared. Why doesn't clearing `classifications.active` clear out the annotations. It would be good to write tests for that before adding any new code.

packages/lib-classifier/src/plugins/drawingTools/models/tools/Tool/Tool.js
```diff
@@ -50,10 +50,18 @@ const Tool = types.model('Tool', {
       self.marks.delete(mark.id)
     }
 
+    function reset () {
+      self.marks.forEach((mark) => {
+        mark.reset()
```
eatyourgreens: Won't `marks.clear()` on the next line automatically clear out `mark.annotations` for each mark?

packages/lib-classifier/src/store/AnnotationsStore.js
```diff
@@ -17,6 +17,20 @@ const AnnotationsStore = types
     }
   }))
   .actions(self => {
+    function afterAttach() {
+      createClassificationObserver()
+    }
+
+    function createClassificationObserver() {
+      const classificationsDisposer = autorun(() => {
+        const classifications = getParent(self)
```
eatyourgreens: The name here is confusing because marks are also instances of `AnnotationsStore`.

`getParent(self)` returns the map for a member of a map, so this will return `classifications.resources` for a classification and `tool.marks` for a mark.

srallen:
> Can we have tests to check whether clearing classifications.resources empties classifications.active.annotations?

Yes, of course. That's why this is still labeled as WIP.

packages/lib-classifier/src/store/AnnotationsStore.js
srallen: `classification.active` is an unique reference to a resource stored in the the map of classification resources. When the reference is invalidated by setting it to undefined, then what MST does is remove the resource it references from the map. I don't believe it'll taken any action related to any other node in the tree, even a child node, and I think that's for safety purposes. You don't want cascading side effects that you're unaware of I think. So what we have happening now is we have an orphaned Annotations node and thus the errors.

eatyourgreens: Ah right, so when we invalidate a node, we should invalidate its children first? This is definitely worth having a test for.

By the way, [3.14.1](https://github.com/mobxjs/mobx-state-tree/blob/master/changelog.md#3141) added `acceptsUndefined` to maps. I've been wondering if that's why we saw behaviour change in 3.15.

packages/lib-classifier/src/store/AnnotationsStore.js
srallen: That was a mistake on my part in reading the code. This observer can be more generalized then.

packages/lib-classifier/src/store/AnnotationsStore.js
srallen: Good to know about the change. Definitely think that's plausible when we attempted the update.

packages/lib-classifier/src/plugins/drawingTools/models/tools/Tool/Tool.js
srallen: No I don't think so for the same reason that I gave for the ClassificationsStore, but this line isn't necessary since you've pointed out that the Mark model is a composition with the AnnotationsStore, so the observer can be modified to observe either parent regardless and reset itself if the parent resets itself.

srallen:
I've added tests and including some refactors. The most notable is to specifically add back in `self.active = undefined` back into the default reset action for the basic `ResourceStore` model. What's weird is that it successfully would set the active reference to undefined in the base model on reset when the resources map are cleared, but didn't with the `ClassificationsStore`. Testing `reset` in isolation with the `AnnotationsStore`.... I have no explanation for this, but I added this back in for safety and am testing for it.  

As an aside, you will see a TODO comment in the Tool model specs for the sub-task. There's a typo there and it's weird the tests are passing with that typo. Since it's out of scope of this, I didn't fix it, but we should in a follow up PR.

packages/lib-classifier/src/store/AnnotationsStore.spec.js
```diff
@@ -66,11 +76,58 @@ describe('Model > AnnotationsStore', function () {
...
+        max: '10',
+        min: 1,
+        type: 'default'
+      })
+      tool.createMark({ id: '1', annotations: { [singleChoiceAnnotation.id]: singleChoiceAnnotation } })
+      const mark = tool.marks.get('1')
+      const resetSpy = sinon.spy(mark, 'reset')
+      expect(mark.annotations.size).to.equal(1)
+      tool.reset()
+      expect(resetSpy).to.have.been.calledOnce()
```
eatyourgreens: `expect(tool.marks).to.be.empty()`?

packages/lib-classifier/src/plugins/drawingTools/models/tools/Tool/Tool.spec.js
```diff
@@ -84,6 +84,7 @@ describe('Model > DrawingTools > Tool', function () {
       const textTaskSnapshot = Object.assign({}, tool.details[2], {taskKey: 'text'})
       const multipleTask = tool.createTask(multipleTaskSnapshot)
       const singleTask = tool.createTask(singleTaskSnapshot)
+      // TODO come back to this. Why is this passing?
```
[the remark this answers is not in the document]
eatyourgreens: This is passing because we set `textTask = tool.createTask(singleTaskSnapshot)` then expect  `textTask` to equal itself. The tests should be checking the task against the snapshots.

At the moment, none of these expectations can ever fail. 😳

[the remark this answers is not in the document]
eatyourgreens: Fixed in #1504.

srallen:
Yeah those `onAction` warnings are annoying and there's not a way to turn it off in dev. The docs also don't provide an explanation about why that's dangerous either.

PR #1506, opened 2020-02-12
fix content pages pages when API data is unavailable

### LINKED ISSUE #1499: Local production builds of the content pages app give 500 errors
**Package**
app-content-pages

**Describe the bug**
Building and running the content pages app with `docker-compose build` and `docker-compose up` from the root of the monorepo gives 500 errors. The actual error is this (for the publications page):
```
TypeError: publicationsData.map is not a function
```
The team page gives a similar error, but `teamData` not `publicationsData`.

Running the app with docker-compose from `packages/app-content-pages` is fine. I think the problem might be that the Contentful environment is undefined for the top-level Docker build.

**To Reproduce**
Run `docker-compose build` and `docker-compose up` (or `docker-compose up --build`) from the root of the monorepo. Browse to http://localhost.zooniverse.org:3001/about/publications once the NextJS apps start up.

**Expected behavior**
You should be able to browse the publications page at
http://localhost.zooniverse.org:3001/about/publications
and the team page at
http://localhost.zooniverse.org:3001/about/team

camallen:
Package: app-content-pages

Closes #1499 

If the contentful API call errors (missing secret client setup) then the `teamData|publicationsData` are left undefined. This PR fixes this issue by adding a check for undefined values and replaces them with an empty array on props and adds default values to the API requests

For bonus points (and learnings), I added tests on the default props API calls to contentful API for team and publications data.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-content-pages/src/screens/Publications/PublicationsContainer.js
+++ b/packages/app-content-pages/src/screens/Publications/PublicationsContainer.js
@@ -10,8 +10,7 @@ import Publications from './Publications'
 counterpart.registerTranslations('en', en)
 
 function PublicationsContainer(props) {
-  const error = props.error
-  const publicationsData = props.publicationsData || []
+  const { error, publicationsData } = props
   const [activeFilter, setActiveFilter] = useState(null)
 
   const filters = createFilters(publicationsData, activeFilter, setActiveFilter)
@@ -46,6 +45,10 @@ PublicationsContainer.getInitialProps = async ({ req }) => {
 PublicationsContainer.propTypes = {
   error: string,
   publicationsData: array,
+}
+
+PublicationsContainer.defaultProps = {
+  publicationsData: []
 }
 
 export default PublicationsContainer
--- a/packages/app-content-pages/src/screens/Publications/PublicationsContainer.js
+++ b/packages/app-content-pages/src/screens/Publications/PublicationsContainer.js
@@ -10,7 +10,8 @@ import Publications from './Publications'
 counterpart.registerTranslations('en', en)
 
 function PublicationsContainer(props) {
-  const { error, publicationsData } = props
+  const error = props.error
+  const publicationsData = props.publicationsData || []
   const [activeFilter, setActiveFilter] = useState(null)
 
   const filters = createFilters(publicationsData, activeFilter, setActiveFilter)
@@ -30,9 +31,12 @@ function PublicationsContainer(props) {
 PublicationsContainer.getInitialProps = async ({ req }) => {
   const host = getHost(req)
   let error
-  const publicationsData = await request.get(host + '/api/publications')
-    .then(res => res.body)
-    .catch(err => error = err.message)
+  let publicationsData = []
+  try {
+    publicationsData = (await request.get(host + '/api/publications')).body
+  } catch (err) {
+    error = err.message
+  }
   return {
     error,
     publicationsData
--- a/packages/app-content-pages/src/screens/Team/TeamContainer.js
+++ b/packages/app-content-pages/src/screens/Team/TeamContainer.js
@@ -10,8 +10,7 @@ import Team from './Team'
 counterpart.registerTranslations('en', en)
 
 function TeamContainer (props) {
-  const error = props.error
-  const teamData = props.teamData || [];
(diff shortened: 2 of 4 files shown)


packages/app-content-pages/src/screens/Team/TeamContainer.js
```diff
@@ -30,8 +31,9 @@ function TeamContainer (props) {
 TeamContainer.getInitialProps = async ({ req }) => {
   const host = getHost(req)
   let error
-  const teamData = await request.get(host + '/api/team')
-    .then(res => res.body)
+  let teamData
```
eatyourgreens:
```suggestion
  let teamData = []
```
It's good practice to set a default here.

packages/app-content-pages/src/screens/Team/TeamContainer.js
```diff
@@ -30,8 +31,9 @@ function TeamContainer (props) {
...
+  await request.get(host + '/api/team')
+    .then(res => teamData = res.body)
```
eatyourgreens: Not a blocking comment but the linter should be warning about returning assignment here. Good practice is to return an actual value or wrap your function so that it returns nothing.

packages/app-content-pages/src/screens/Team/TeamContainer.js
camallen: thanks - OOI what about the `let error` above, I assume as that will be undefined instead of used in the mapping functions it's ok to not have a default?

eatyourgreens: Depends how consuming code is supposed to detect errors, I guess. It looks like you return both `teamData` and `error`. Will anything bad happen if both are defined at the same time?

camallen: no - it handles the missing data (original error). I've added mappable default for teamData

packages/app-content-pages/src/screens/Team/TeamContainer.js
camallen: intruging - only linter errors I got from `yarn lint` were about return values from fat arrow functions.

camallen: nope - i modified the code to remove that....master linter errors are there `Arrow function should not return assignment`

eatyourgreens: That's the one. Here you are returning `teamData = res.body`.

camallen:
> Looks good. I think the publications page will need the same changes.

Done - I think this is ready for another review

packages/app-content-pages/src/screens/Publications/PublicationsContainer.js
```diff
@@ -10,7 +10,8 @@ import Publications from './Publications'
 counterpart.registerTranslations('en', en)
 
 function PublicationsContainer(props) {
-  const { error, publicationsData } = props
+  const error = props.error
+  const publicationsData = props.publicationsData || []
```
eatyourgreens: I somehow missed that this is a React component. You can set default values for props for the component in `PublicationsContainer.defaultProps`

packages/app-content-pages/src/screens/Team/TeamContainer.js
```diff
@@ -10,7 +10,8 @@ import Team from './Team'
 counterpart.registerTranslations('en', en)
 
 function TeamContainer (props) {
-  const { error, teamData } = props
+  const error = props.error
+  const teamData = props.teamData || [];
```
eatyourgreens: Same comment here about setting defaults in `TeamContainer.defaultProps`.

packages/app-content-pages/src/screens/Publications/PublicationsContainer.js
camallen: Perfect :) will do

packages/app-content-pages/src/screens/Team/TeamContainer.js
camallen: Done - i think this is ready now.

PR #1510, opened 2020-02-12
Fix 1507 - allow MultipleChoiceTask component to observe changes again

### LINKED ISSUE #1507: Classifier: multiple choice answers don't reflect annotation state
**Package**
lib-classifier

**Describe the bug**
Checking answers for the multiple choice task updates the annotation value but the checked answers aren't shown in the browser. Here the annotation value is `[ 1, 2 ]` but it appears be `[]`.
A screenshot of the multiple choice task, showing no answers apparently selected although two actually are.

shaunanoordin:
## PR Overview

package: `lib-classifier`
Fixes #1507 

This PR fixes an issue where, when a user clicks on an Answer label, the label isn't highlighted (marking it as selected/active) BUT the answers are correctly recorded in the Classifications.

Can be tested on any workflow with a single multiple-choice-answer task, e.g. http://localhost:8080/?env=staging&project=1895

### Details

The issue stems from (what appears to be) a conflict with the _store observers_ applied to the MultiChoiceTask component. Specifically, when the MultiChoiceTask component is exported with the `withTheme` decorator, the component stops listening for changes in the annotations store.

This means that when a user clicks on an Answer (i.e. action on a component), the annotations store will _correctly_ update (i.e. the selected answer is added to the classification to be submitted), BUT the component in turn does _not_ register that any changes were made, hence does _not_ re-render the component with the answer label highlighted.

### Status

This is a quick UI fix _only for the MultipleChoiceTask,_ and it's ready for review. 👌 

A deeper look into the decorators/observer problem should be done as a follow up.

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/MultipleChoiceTask/components/MultipleChoiceTask.js
+++ b/packages/lib-classifier/src/plugins/tasks/MultipleChoiceTask/components/MultipleChoiceTask.js
@@ -2,7 +2,7 @@ import { Markdownz, pxToRem } from '@zooniverse/react-components'
 import { Box, Text } from 'grommet'
 import React from 'react'
 import PropTypes from 'prop-types'
-import styled, { css, withTheme } from 'styled-components'
+import styled, { css } from 'styled-components'
 
 import TaskInput from '../../components/TaskInput'
 
@@ -105,5 +105,5 @@ MultipleChoiceTask.propTypes = {
   theme: PropTypes.object
 }
 
-export default withTheme(MultipleChoiceTask)
+export default MultipleChoiceTask
 export { MultipleChoiceTask }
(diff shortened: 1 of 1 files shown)


packages/lib-classifier/src/plugins/tasks/MultipleChoiceTask/components/MultipleChoiceTask.js
```diff
@@ -2,7 +2,7 @@ import { Markdownz, pxToRem } from '@zooniverse/react-components'
 import { Box, Text } from 'grommet'
 import React from 'react'
 import PropTypes from 'prop-types'
-import styled, { css, withTheme } from 'styled-components'
+import styled, { css } from 'styled-components'
 
 import TaskInput from '../../components/TaskInput'
```
srallen:
```suggestion
import TaskInput from '../../components/TaskInput'
import zooTheme from '@zooniverse/grommet-theme'`
```

shaunanoordin:
## PR Update

Changes have been made:

- CSS `background` has been reverted to use `zooTheme`
- `props.theme` has been removed

👌

srallen:
I made couple more tiny changes: removing the now unused css helper and moving the `pxToRem` function out of the styled-component template.

PR #1515, opened 2020-02-13
Tmp revert for migration

adammcmaster:
This reverts commits since the last production release (#1475, #1506, but NOT #1514) so that we can move the current production build to Azure.


adammcmaster:
This PR itself can of course be reverted as soon as we've deployed to Azure.

PR #1517, opened 2020-02-13
Drawing SubTasks - Popup UI component (v2)

### LINKED ISSUE #1315: Drawing sub-task UI

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Fixes #1315
Replaces #1491 

The aim of this PR is to add SubTasks for Drawing Tasks. e.g. for drawing tasks that says "use this Rectangle tool to mark an animal in this photo, then type the name of that animal in this sub task"

Requirements:

- [x] Popup UI component should appear and displays subtasks when appropriate. (i.e. when a drawing task has sub-tasks, AND when a mark for that drawing task is selected)
  - [ ] Popup UI component shouldn't behave like a UX jackass, e.g. appearing right in front of whatever the user is trying to mark or draw on.
  - [ ] Popup UI component should appear as designed. (See below)
- [x] when submitting a Classification, all "sub-annotations" made for sub-tasks should be submitted.

Should look like:

Screen Shot 2020-02-06 at 18 56 09

### Details

**The Past**

- Replaces 1491 due to changes in how `activeMark` is now part of `activeDrawingTask` in the stores. (Previously, 1491 created its own dedicated sub-task store)

**Future Challenges**

- ⚠️ I **very strongly** suspect I'll need to **rebase or otherwise create v3** once #1475 (Classification snapshot preprocessor) gets merged, since this PR changes `ClassificationStore.completeClassification()` by adding...
  ```
  classificationToSubmit.annotations.forEach(annotation => {
    annotation.value && annotation.value.forEach(annotationValue => {
      if (annotationValue.annotations) {
        annotationValue.annotations = convertMapToArray(annotationValue.annotations)
      }
    })
  })
  ```
  ...so that each sub-task's annotations are properly submitted as classifications, whereas PR 1475 changes the whole submit-my-Classification process altogether.
- ⚠️ This PR _may require_ the "drawingMark.onFinishInitialDrawing()" or similar to prevent the unwanted UI behaviour of appearing WHILE the user is still drawing.
  - I have ideas, but they might be terrible and/or require a separate PR.

### Status

Very much a WIP.


eatyourgreens:
#1475 merged, then reverted so we could deploy, but should be back soon.

Once it's in, we should try to get some classifications flowing through to Panoptes so that we can check this issue: https://github.com/zooniverse/Panoptes/issues/3252

packages/lib-classifier/src/components/Classifier/components/SubTaskPopup/SubTaskPopup.js
```diff
@@ -0,0 +1,119 @@
...
+
+counterpart.registerTranslations('en', en)
+
+function storeMapper (stores) {
+  const { activeStepTasks } = stores.classifierStore.workflowSteps
+  const [activeDrawingTask] = activeStepTasks.filter(task => task.type === 'drawing')
+  const activeMark = (activeDrawingTask && activeDrawingTask.activeMark)
+    ? activeDrawingTask.activeMark
+    : undefined
+  const { addAnnotation } = stores.classifierStore.classifications
```
eatyourgreens:
```suggestion
  const { addAnnotation } = activeMark
```
You'll want to add these annotations to the active mark, not to the classification.

shaunanoordin:
Ooh, thanks Jim - I'll rebase as soon as 1475 is back! 👌

eatyourgreens:
It’s merged back in. You can take out the changes to classifications here and try merging that in instead.

packages/lib-classifier/src/components/Classifier/components/SubTaskPopup/SubTaskPopup.js
shaunanoordin: Whoops! Thanks for spotting this, I forgot to clean this up from the previous PR since `props.addAnnotation` is no longer used; I pull `props.activeMark.addAnnotation` straight during the render()

shaunanoordin:
Removed changes to ClassificationsStore, and merging from master...

shaunanoordin:
I've just re-applied master to the PR, and I'm now seeing annotations for drawing subtasks being submitted to /classifications in a rather format.

Which, if I'm reading the discussions correctly, means the changes to sub-task annotations as outlined in #3252 / [ADR-25](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-25.md) **is now live and working as expected.** 👍 

Screen Shot 2020-02-13 at 17 55 50

Screen Shot 2020-02-13 at 17 56 10

shaunanoordin:
Note to self: `activeMark.x_center` and `activeMark.y_center` can't be used to calculate the the UI position of the popup, because 1. the x_center and y_center are relative to the size of the Subject image, which would be pretty squished, and 2. don't factor for zoom and pan or rotation.

Instead: introduce "popupComponent.props.parentHtmlNode" and "store.activeMarkHtmlNode" as ways to get bounding boxes.

shaunanoordin:
Note to self: `DrawingTask.setSubTaskVisibility()` perfectly ensures that the popup UI doesn't block users when they're drawing a mark. However, it may overlap with `InteractionLayer.setCreating()` which in practice may be doing the exact same thing. Check if these two things should be folded.

shaunanoordin:
Note to self: still need to figure out how to get a Ref on the Active Mark (the HTML DOM node bit, not the store model). `<SubTaskPopup />` has been repositioned to catch the SVG Context, so we can get the bounding box of the `<svg>` element, but the active mark is still buried too deeply in the hierarchy for me to get a grasp on where I can attach the ref.

TODO: re-check hierarchy
```
InteractionLayerContainer
  Interaction Layer  // setActiveMark() is caught here
    DrawingToolMarks
      mapOf(Mark) // Component refs need to be defined here...
        Mark  // ...or here
```

TODO: `Mark` has forwardRef() but I'm not sure if the `ref` here is being used? Or probably re-read the [Forwarding Refs docs](https://reactjs.org/docs/forwarding-refs.html)

shaunanoordin:
## PR Update

This PR now has full functionality, but it's still not ready for a merge due to several factors.

First off, this is where we're at now:

Screen Shot 2020-02-18 at 19 09 30

### Still To Do / Problems

Mundane problems that need my attention:

- Tests need to be written 😢 
- SubTaskPopup needs to respect dark/light theme (hmm... different PR maybe, because this is getting huge as is?)

⚠️ UI design items that may need @beckyrother's attention:

- The Close (X) button is just hanging out in the corner there being cool and aloof, because unlike the [Invision design](https://projects.invisionapp.com/d/main#/console/12923997/396381420/preview), the sub-task's _text_ (e.g. "Transcribe Line") doesn't appear next to it.
  - Becky looks to the camera and asks, _"Wait, why?"_
  - Camera pans to Shaun, who says, _"The sub-task popup is being made to support any number of Tasks. As a result, the sub-task popup is basically a draggable, resizable version of the right-hand-side Task Area, and that means each Task plays inside its own sub-box."_
    Screen Shot 2020-02-18 at 19 21 25
- I'm not sure if "Save" is the best way to describe that last button, because _all your answers to SubTasks are 'instantly recorded',_ regardless of whether or not you click the Save button.
  - Becky spits her coffee directly into the camera. _"Wait, what??"_
  - Shaun dodges the spit take, only to crash into a 2m tall Lego sculpture set up earlier in this episode. _"It's not my fault! The sub-tasks work exactly like Tasks in the right-panel Task Area, so actions like clicking on an answer are recorded immediately!"_
- Question: what changes can be made to get the UX in a place that's sensible to the users, and the UI still close enough to the original design?
  - Creating new Task components _specifically_ for the sub-task popups that 1. look better in the popup and 2. don't save any changes unless the 'Save' button is clicked is possible, but probably not a good idea
  - Keeping the Close (X) button and removing the Save button would be my default suggestion.
  - Note to self: if we're keeping the Close (X) button, create a custom SVG for the _circled-X_ since the default grommet-icons doesn't have one, and no other component in the lib-classifier uses that icon yet, AFAIK.
  - I'll hold off many any specific changes until we can confirm the ideal way forward, because with good UI design you-
    - Camera switches suddenly to the front door, where Grant bursts in and says his catchphrase, _"-you can't take it for Granted!"_ Cue laughter, credits, end episode.

Advanced problems that need my attention and some ibuprofen:

- SubTaskPopup can be resized. Cool.
  - But this means it's possible for Tasks with multiple-subtasks to get bunch really really horribly if the user resizes it to a size that's too small. (i.e. anything smaller than the default)
    Screen Shot 2020-02-18 at 19 19 12
  - Currently implemented solution is the judicious use of overflow: auto to make the popup still _sorta_ workable?
  - Idea: disable [enableResize](https://github.com/bokuweb/react-rnd) of React-Rnd? 
  - Counterpoint: but then users can't resize a 'large popup' to fit into their screen.
  - Counter-counterpoint: Shrinking a popup that has way too many sub-tasks won't make it easier to read, you doofus.
- The initial position of the subtask popup uses very basic calculations right now. It... works?
  - ⚠️ 💀 Calculating a **good** initial position of the subtask popup however is going to be an absurd chicken-and-egg problem because _you need to know the size of the popup (factoring in Task sizes, etc) before you can calculate where to display it but you can't calculate the size until you display the popup in the DOM_ 
    - Idea: display, then pop a split-second timer to reposition.
    - Counterpoint: that's inelegant and you'd need to monitor onPropChange to see when `(currentSubTaskVisibility === true && previousSubTaskVisibility === false) || (currentActiveMark !== previousActiveMark)`
    - Counter-counterpoint: that was not a counterpoint, that was a complaint.
- z-indexes may become a nightmare.
  - I mean why is this even happening:
    Screen Shot 2020-02-18 at 19 32 34
  - Counterpoint: fffuuuuu
  - Counter-counterpoint: cccccccckkkkk
  - Note to self: The disabled "add to favourites" button appears on top of the popup, but not the clickable "Subject Info" button. Field guide is probably the same too. Task area labels have position: relative. Why this happens eludes me, but it's worth investigating.
  - Actual idea: wait, why not just set zIndex to 100000000000 to see if that solves the problem, then adapt a more permanent fix from there?

### Status

WIP

srallen:
For the close button styles, we have a reusable component for that: https://zooniverse.github.io/front-end-monorepo/@zooniverse/react-components/index.html?path=/story/closebutton--light-theme-default

shaunanoordin:
Wunderbar, thanks Sarah!

shaunanoordin:
Reminder to self: [window.getComputedStyle](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) is a way of figuring out the actual applied/computed styles on visible DOM elements.

shaunanoordin:
⚠️ A new challenger approaches! 

While the current incarnation of the SubTask popup works well when the sub-tasks are single/multi choice answers (i.e. buttons that can be clicked), there's an issue with the text input which prevents it from being selected/focused on. Any attempt to CLICK on the text input will initiate the React-Rnd popup's drag action, instead of giving the text input the focus. (You can still tab to it, though.)

Solution 1:

> Will:
> Oh. You’re trying to focus on a textarea inside a RND component, but the drag handler is always activating?
> I’d pay attention to this prop https://github.com/bokuweb/react-rnd#draghandleclassname-string
> Which is a pretty common gotcha in ReactRND. Try to only attach a draghandler to a component you want to be draggable (instead of the entire box and all children). I think that’s how I handled it in ASM

Will gets 👍 karma points for that

Solution 2 might work with the newer version of Rnd-React: https://github.com/bokuweb/react-rnd#cancel-string

The `<Rnd cancel>` is the flipside of `Rnd handle`; `cancel` tells you which elements should IGNORE the drag action, `handle` tells you which elements are the ONLY ones which should respect the drag action.

...I _think._ Gotta try this.

shaunanoordin:
Solution 2 is the easier implementation:

```
<Rnd
  minWidth={100}
  minHeight={100}
  default={defaultPosition}
  cancel=".subtaskpopup-element-that-ignores-drag-actions"
>
  <Box pad="xsmall" fill>
    {tasks.map(task => {
      ...
      if (annotation && TaskComponent) {
        return (
          <Box
            className="subtaskpopup-element-that-ignores-drag-actions"
          >
            <TaskComponent
              ...
            />
          </Box>
        )
      }
    })}
  </Box>
</Rnd>
```

packages/lib-classifier/package.json
```diff
@@ -1,4 +1,4 @@
-{
+  {
```
eatyourgreens: Accidental indentation change here?

shaunanoordin: Whoops! I don't know hot this got there, thanks for spotting it.

beckyrother:
Thank you Shaun! Notes:

> The Close (X) button is just hanging out in the corner there being cool and aloof, because unlike the Invision design, the sub-task's text (e.g. "Transcribe Line") doesn't appear next to it.

Is it possible to move the X to the right of the tasks, so that it's hanging out on the side rather than creating additional white space at the top? Otherwise this isn't a dealbreaker.


> I'm not sure if "Save" is the best way to describe that last button, because all your answers to SubTasks are 'instantly recorded', regardless of whether or not you click the Save button.

Whether or not the Save button is actually saving responses, the word 'save' helps volunteers feel safe (get it) in that their responses _have_ been recorded. It's a placebo button, essentially. Please continue to auto-save.


> Question: what changes can be made to get the UX in a place that's sensible to the users, and the UI still close enough to the original design?

Original design intent: the 'Save' button records the classification and closes the modal, while the X discards the classification and closes the modal. Is this possible?

You can download the close button as an svg in InVision by using the Inspect mode, clicking on the x, and clicking the download button on the right:
Classifier_2018_Inspect_Mode_-_InVision




>SubTaskPopup can be resized. Cool.

Original design intent was resizing only the text input. Please let me know earlier next time if something like this comes up, always happy to clarify.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingToolMarks/DrawingToolMarks.js
```diff
@@ -12,6 +12,15 @@ function DrawingToolMarks ({ activeMarkId, marks, onDelete, onDeselectMark, onFi
     const MarkingComponent = observer(mark.toolComponent)
     const ObservedDeleteButton = observer(DeleteButton)
     const isActive = mark.id === activeMarkId
+    const ref = React.createRef()
+    
+    function onFinishWithRef (event) {
+      onFinish(event, ref)
```
eatyourgreens: It might be safer to pass `ref.current` here. Otherwise, you’re passing a reference to one of this component’s internal variables.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingToolMarks/DrawingToolMarks.js
```diff
@@ -12,6 +12,15 @@ function DrawingToolMarks ({ activeMarkId, marks, onDelete, onDeselectMark, onFi
...
+    }
+    
+    function onSelectMarkWithRef (mark) {
+      onSelectMark(mark, ref)
```
eatyourgreens: Same comment about passing `ref.current` here.

shaunanoordin: 👍

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingToolMarks/DrawingToolMarks.js
shaunanoordin: 👍 Good point, I've addressed that in the most recent commit.

shaunanoordin:
@beckyrother Thanks for the feedback 👌 

- I'll keep the (placebo) Save button as-is. Since all changes auto-save, this means both the X-close button and Save button on the SubTaskPopup continue to be functionally identical. Which means that...

> Original design intent: the 'Save' button records the classification and closes the modal, while the X discards the classification and closes the modal. Is this possible?

- ...unfortunately, the X-close button on _the SubTaskPopup_ doesn't discard changes made to the sub-task annotation. (Deleting the drawn rectangle/line/point Mark does the job, though.) Sorry, but the only implementation that does that I can think of right now (keep a snapshot of the sub-task annotations on popup open, then use that to revert) would take more thought & time to implement, as it'd also need to factor other ways that a popup could 'close' - e.g. when a user selects another Mark or attempts to create another one.

- I got the 'x' button already, thanks - now to see about moving its position to be more aesthetically pleasing.

shaunanoordin:
## PR Update

Oh my GODS I finally figured out the correct way to do themes and tests! That's it, those are the last few pieces needed for this PR to be functional and _compleat._

- Themes: light and dark.
  Screen Shot 2020-02-25 at 15 05 39
  Screen Shot 2020-02-25 at 15 05 46
- Tests: they're there.

### Status

I'm removing WIP from this PR since everything here is now viable for review. I'm thinking of tinkering on a few minor things, but these might be a separate PR.

- Moving that close button to be more visually attractive
- Addressing a minor UI issue where the popup box doesn't reposition itself if, while it's open, you select (click on) another existing Mark.
- Addressing a notable UI issue where we can't calculate the optimum position of the popup when it appears.
- Addressing a notable UI issue where the Z-index of other components interfere with the popup.

EDIT: need to update the branch, BRB.

packages/lib-classifier/package.json
```diff
@@ -35,6 +35,7 @@
     "mobx-utils": "~5.0.2",
     "pepjs": "~0.5.2",
     "react-resize-detector": "~4.2.0",
+    "react-rnd": "~10.1.5",
```
eatyourgreens: Is this PR missing the changes to the Yarn lock file for this?

eatyourgreens:
Works first time with the ASM staging workflow (3392)! Excellent work!
A staging subject from Anti-Slavery Manuscripts, with the subtask open in order to transcribe a line.

I'm seeing some z-index problems with the tool selector on the right. My immediate inclination would be to merge this PR then maybe iterate to figure out why it isn't quite styling properly. What do you think?

eatyourgreens:
With more than one line marked on the page, I can select a different line while the current line is being edited. At that point, it's no longer clear which mark the transcription form is associated with. I'm assuming the original mark. Here, for example, I opened the form when I drew the upper line, but then selected the lower line after the form opened. It now looks like I'm entering text for the second line. I think mark selection should be disabled while the task form is open. Maybe disable the whole interaction layer while the popup is open?
A page showing two lines (one selected) and a popup transcription box.

eatyourgreens:
Submitted classifications match the format from ADR-25 :+1:.

eatyourgreens:
SERIOOUS BUG: this is a little hard to reproduce but:
- mark two lines, adding different text for each.
- open one.
- select the other.
- without closing the popup, keep selecting and deselecting lines in the background.

If you do that ~3 times, the text value for both lines becomes the same.

After discussion with @shaunanoordin, keying the popup by `mark.id` might fix this.

shaunanoordin:
> I'm seeing some z-index problems with the tool selector on the right. My immediate inclination would be to merge this PR then maybe iterate to figure out why it isn't quite styling properly. What do you think?

I agree, I think the UI improvements should be its own PR. There are three things that PR needs to handle:

1. figure out the wonky z-indexes that's causing certain parts to layer above the popup.

> With more than one line marked on the page, I can select a different line while the current line is being edited. At that point, it's no longer clear which mark the transcription form is associated with. I'm assuming the original mark.

2. When another mark is selected, the popup should reposition itself to that mark.

(EDIT: following our discussion, one thing to do is to add a `key={activemark.id}` to the popup so that it forces a re-render after the selected mark changes)

(EDIT EDIT: due to the bug you discovered, this will need to be accelerated into this PR, not another)

3. the default position of the popup could be better.

eatyourgreens:
Here's an annotation I created, where I'd entered 'Somewhere or other' for line 1 and 'Hello world' for line 2, then selected and deselected lines in the background until the subtask broke. It looks like the bug is affecting the stored values for each mark.
Screenshot of a test annotation where both lines have accidentally been assigned the same text value.

eatyourgreens:
@shaunanoordin I'm afraid I'm still seeing that bug after pulling down the latest commit.

I'm also noticing that the text input isn't focused when the popup opens, so I can't start typing without clicking on the form first. Can it use autofocus?

shaunanoordin:
> I'm afraid I'm still seeing that bug after pulling down the latest commit.

The issue is specific to the Text Task for the moment; I'm still trying to trace what's triggering the data overwrites.

> I'm also noticing that the text input isn't focused when the popup opens, so I can't start typing without clicking on the form first. Can it use autofocus?

Good idea! SubTaskPopup now autofocuses on the first Sub Task on the list. 👍

eatyourgreens:
Sorry, one other comment: if I navigate the drawn marks with the tab key, then focus doesn't return to the active mark when the popup closes. Also, it's annoying that the popup automatically opens each time I tab to a new mark. Can the interaction be tab + Enter for keyboard users?

packages/lib-classifier/src/components/Classifier/components/SubTaskPopup/SubTaskPopup.js
```diff
@@ -0,0 +1,185 @@
...
+              // or creates a new one if one doesn't exist.
+              // The name is a bit confusing.
+              const annotation = activeMark.addAnnotation(task)
+              task.setAnnotation(annotation)
+              const TaskComponent = observer(taskRegistry.get(task.type).TaskComponent)
+              
+              if (annotation && TaskComponent) {
+                return (
+                  <TaskBox
+                    key={task.taskKey}
```
eatyourgreens: I think this should be keyed by `annotation.id`.

shaunanoordin: I've changed this to `annotation.id`, but note that this _masks_ the problem of Text Tasks overwriting one another. Now, it's become the "merry go round" model of the problem.

eatyourgreens:
I've pushed a fix but it needs double-checking because I don't 100% understand why it works.

packages/lib-classifier/src/components/Classifier/components/SubTaskPopup/SubTaskPopup.js
eatyourgreens: It's something to do with there only ever being one task in the store, but it's handling the annotations for more than one drawn line. When we change lines, the stored annotation is briefly out of sync with the task component, I think.

beckyrother:
> With more than one line marked on the page, I can select a different line while the current line is being edited. At that point, it's no longer clear which mark the transcription form is associated with. I'm assuming the original mark. Here, for example, I opened the form when I drew the upper line, but then selected the lower line after the form opened. It now looks like I'm entering text for the second line. I think mark selection should be disabled while the task form is open. Maybe disable the whole interaction layer while the popup is open?

Yes, you shouldn't be able to interact with anything outside the popup. If you want to edit a different line, you'll need to close the current popup by either clicking save or the x (or outside the box, @shaunanoordin is that possible), and then clicking a different line. 

If you'd like to make it really clear which line you're currently editing, consider making that line's stroke thicker or lowering the opacity of the other lines.

shaunanoordin:
> I've pushed a fix but it needs double-checking because I don't 100% understand why it works.

I've done some thorough tests on workflows with 1 Text task, workflows with 3 Text tasks, and workflows with 1 Drawing task + 1 Text sub-task. The solution works, though the _"for text subtasks, when switching between active marks without closing the popup, onMount() sees annotation X but onUnmount() sees annotation Y"_ issue is still there... though I'm starting to see that it's because the TextTask component is looking at `props.task.annotation` instead of `props.annotation`. (It comes back to what you said - the Task remains the same but the Annotations _value_ change, but since the TextTask only looks at the task prop, it doesn't realise the necessary for a re-render.)

shaunanoordin:
> Yes, you shouldn't be able to interact with anything outside the popup. If you want to edit a different line, you'll need to close the current popup by either clicking save or the x (or outside the box, @shaunanoordin is that possible), and then clicking a different line.

This would solve a lot of the UI interaction issues like a sledgehammer, so I'm game to looking into this. Thoughts: a transparent Grommet modal, though I need to check if that blocks tab-navigation to other elements.

I still think understanding the cause of the weirdness is worth looking into, because this might come back to haunt us later.

eatyourgreens:
One more thought about focus management: on close, focus should shift from the save (or close) button back to the original mark. I'm not 100% sure where focus is going at the moment.

shaunanoordin:
@beckyrother Sorry, but what are the changes that you requested? I'm looking at https://github.com/zooniverse/front-end-monorepo/pull/1517#pullrequestreview-365071517 but I don't see any details to what needs changing. I'm not sure if I'm missing something on GitHub.

beckyrother:
@shaunanoordin sorry! They were in a comment above:

Yes, you shouldn't be able to interact with anything outside the popup. If you want to edit a different line, you'll need to close the current popup by either clicking save or the x (or outside the box, @shaunanoordin is that possible), and then clicking a different line.

If you'd like to make it really clear which line you're currently editing, consider making that line's stroke thicker or lowering the opacity of the other lines.

shaunanoordin:
## PR Update

- The SubTaskPopup is now wrapped in an invisible modal layer, so users cannot click on or navigate to anything other than the popup, while the pop up exists.
  - @beckyrother at the moment, as requested, I've made sure that if someone clicks "outside" the space of the popup, nothing happens. A user has to click on the Save button, or click on the X close button, or press the **Esc** key on the keyboard.
  - However, I personally feel that the "clicking outside the space of the popup" action should _also_ close the popup, as it feels more natural. (Because the 'blocking' modal layer is invisible, there are no indicators otherwise telling users _why_ everything else is unclickable. And it has to be invisible too, because otherwise a greyed-out 'blocking' modal layer will make it difficult to see the Subject image you're transcribing. 🤷‍♂ )
  - I'll leave this as is, but when you get a chance to test it whenever this branch gets deployed to preview, do let me know. The code to toggle between "clicking outside the popup = nothing happens" vs "clicking outside the popup = close the popup" is easily changed via `<Layer onClickOutside={this.close.bind(this)} />`

Also, thanks to the introduction of the modal layer...

- zIndex layering is no longer an issue 👍 
- we no longer need to calculate the parent SVG's visual positioning, since the popup is now anchored to the modal, which in turn is anchored to the top-left of the viewport.
- NOTE TO SELF: possibly move <SubTaskPopup /> up a few levels now it no longer has to be within the SVG context provider.

shaunanoordin:
Oh! Also, we already have **increased stroke thickness** for **actively selected** subjects, it's just that the difference is currently a factor of 1.5

Screen Shot 2020-02-27 at 19 17 57
_Currently, an un-selected Mark has stroke-width 2, while a selected Mark has stroke-width 3._

Screen Shot 2020-02-27 at 19 18 19
_This is what it'll look like if I change it to unselected: 2, selected: 4_

Screen Shot 2020-02-27 at 19 18 49
_This is what it'll look like if I change it to unselected: 2, selected: 5_

Screen Shot 2020-02-27 at 19 21 20
_Transcription lines don't change stroke width, but change colours instead._

@beckyrother I'll change the Selected Mark's stroke-width to 4, unless you advise otherwise.

beckyrother:
Hi Shaun, sorry for the confusion but yes, clicking outside the popup should close the popup.

Thicker strokes look great! Thanks for demoing :)

shaunanoordin:
> Hi Shaun, sorry for the confusion but yes, clicking outside the popup should close the popup.

👍 Sweet, thanks for confirming Becky - that's back in. 😄 

## PR Update

OK, I'm wrapping up development for this feature so I'll be re-requesting reviews so this becomes merge-able.

What's in this PR:

- If a drawing tool mark has subtasks, then when you finish creating a new one, a popup will open.
- If you click on an existing drawing tool mark with subtasks, a popup will open.
- The subtask popup functions the same as the right-hand Task Area.
  - All changes to the sub-task values are **saved automatically.** (This is important to note because there's a 'Save' button in that UI that only closes the popup. _Not_ clicking the save button to close the popup doesn't revert any changes made.)
- When the subtask popup is open, interaction with other elements are blocked.
  - There's an invisible modal layer blocking interaction.
  - Clicking outside the popup, or clicking the save button, or clicking the close (X) button, or pressing the Esc key, will close the popup.
- The selected/active drawing tool mark is now more visually distinguishable from other marks.

There are a few factors that should be addressed, but I'll document them in a separate issue for a handover. Here's a quick list:

- When selecting an existing mark, popup shouldn't appear immediately (i.e. don't hook it to onFocus.)
- Consider keyboard navigation (currently, popups that appear onFocus screws with tab nav) 
- The usage of modals (Grommet Layers) might need to be re-examined due to lack of scrollbars (yes, this can matter in certain situations)
- The initial popup position can be improved.

Most of the things above are me fussing a lot about the UI/UX experience, so whoever's taking over dev will have final say if those items make sense or if I'm being too paranoid. Other things I need to document: (this is mostly a reminder for myself)

- Solution for zIndex, if we drop the modal model.
- This needs more testing with multi-frame viewer. (I'm particularly concerned about this if the modal model is removed.)

beckyrother:
>When selecting an existing mark, popup shouldn't appear immediately (i.e. don't hook it to onFocus.)

Could you explain this a bit more? I'm not sure I understand the reasoning.

shaunanoordin:
> > When selecting an existing mark, popup shouldn't appear immediately (i.e. don't hook it to onFocus.)

> Could you explain this a bit more? I'm not sure I understand the reasoning.

@beckyrother Can do!

front-end-monorerepo-issue-1517-mouse

Currently, with this PR, the mouse behaviour follows Fig 1. As soon as your mouse **clicks down** on a drawing tool mark, the popup appears.

This is fine if you just wanted to select the mark and open the popup, but there are _two_ possible actions intended to come from a mouse click. 1. the user is clicking to select a mark and open a popup, or 2. the user is clicking down on the mouse, and _holding down the mouse button_ to initiate a _drag-to-move_ action for the drawing tool mark.

The problem with the behaviour in Fig 1 is that it doesn't cater for the drag-to-move action. As soon as you click down on the mouse, in the first millisecond before you can do anything, BAM! You have a popup in your face.

Fig 2 should be the recommended way of handling popups - i.e. if the user clicks down on the mouse button, assume they're initiating drag-to-move. (Even if they're moving 0 pixels.) It's only when _the mouse button is released_ should the subtask popup appear.

Or more easily:

```
onMouseDown => initiate drag-to-move action.
onMouseMove => if mouse button is held, execute drag-to-move action.
onMouseUp => complete drag-to-move action, and open the subtask popup.
```

beckyrother:
Makes sense! Thank you for the illustration, really helped! And I agree to move forward with that plan 👍

srallen:
@shaunanoordin Is this WIP or just waiting for final review?

shaunanoordin:
@srallen This PR is ready and waiting for a final review. 👌 

All features and functionality that was intended has been built in, and any known follow up items has been documented in #1546

shaunanoordin:
👍 I've updated the base `yarn.lock` file to include `react-rnd` by running the bootstrap script again. I'm now waiting for the tests to finish, and then we should be good to go!

PR #1526, opened 2020-02-18
Color and glyph picker utility helpers

srallen:

Package: lib-classifier

Describe your changes:
*Note* #1524 should be reviewed first. When it merges, I'll change the base for this to master.

This adds some utility helper functions that determine which color and glyph symbol to use with scatter plots. I'll be adding the use of these helpers to the `Controls` component in a subsequent PR. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/helpers/getDataSeriesColor/getDataSeriesColor.spec.js
```diff
@@ -0,0 +1,118 @@
...
+      before(function () {
+        logError = sinon.stub(console, 'error')
+      })
+
+      after(function () {
+        console.error.restore()
+      })
+
+      it('should error to the console and return an empty string', function () {
+        dataSeriesWithInvalidCustomColor.data.forEach((series) => {
```
wgranger: Can you help me understand how an invalid color could be submitted and why we would want to return a blank string (`''`) instead of the default `light-4` here? Is this mainly just to cover every potential edge case?

srallen: We're allowing researchers to specify a color in their subject JSON because color might have meaning in the research, like with the variable star use case, where red and blue represent specific spectra.

I've also chosen to support specifying colors to use by way of a theme object, like our zooniverse Grommet theme, but it can possibly be changed out for some other theme object that uses the same structure and conventions. It's possible this can also have an error in it. So instead of a fallback color in these edge cases, the data points render without a fill color and the console has the error message. Seeing the data points without a fill color should hopefully give the researcher or the dev a clue that something is wrong.

wgranger: This makes sense. Good call on not displaying the error so the researcher can respond accordingly.

PR #1532, opened 2020-02-19
[WIP] Fix content apps content api

### LINKED ISSUE #1499: Local production builds of the content pages app give 500 errors

camallen:
Package: `app-content-pages`

Closes #1499

The Content API has app proxies at `/api/team` in the nextjs app. The way these are set are from the `ASSET_PREFIX` env var or from the hosted domain. In the root docker-compose setup our app is running on a different host port from the next JS app so we need to ensure the `ASSET_PREFIX` env var points to the correct content proxy API host, i.e. the running nextjs app host:port.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,26 +1,26 @@
 version: '3.7'
 
 services:
-  yarn:
-    image: front-end-monorepo_dev:latest
-    build:
-      context: ./
-      target: bootstrap
-    entrypoint:
-      - "yarn"
-    command: ["test:ci"]
-  fe-project:
-    image: front-end-monorepo_prod:latest
-    build:
-      context: ./
-      target: production-apps
-    entrypoint:
-      - "yarn"
-      - "workspace"
-      - "@zooniverse/fe-project"
-    command: ["start"]
-    ports:
-      - "3000:3000"
+  # yarn:
+  #   image: front-end-monorepo_dev:latest
+  #   build:
+  #     context: ./
+  #     target: bootstrap
+  #   entrypoint:
+  #     - "yarn"
+  #   command: ["test:ci"]
+  # fe-project:
+  #   image: front-end-monorepo_prod:latest
+  #   build:
+  #     context: ./
+  #     target: production-apps
+  #   entrypoint:
+  #     - "yarn"
+  #     - "workspace"
+  #     - "@zooniverse/fe-project"
+  #   command: ["start"]
+  #   ports:
+  #     - "3000:3000"
   fe-content:
     image: front-end-monorepo_prod:latest
     build:
@@ -31,6 +31,8 @@ services:
       - "workspace"
       - "@zooniverse/fe-content-pages"
     command: ["start"]
+    environment:
+      - ASSET_PREFIX=http://localhost:3000
     ports:
       - "3001:3000"
 
--- a/packages/app-content-pages/src/screens/Team/TeamContainer.js
+++ b/packages/app-content-pages/src/screens/Team/TeamContainer.js
@@ -32,8 +32,12 @@ TeamContainer.getInitialProps = async ({ req }) => {
   let error
   let teamData = []
   try {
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -1,26 +1,26 @@
...
-      context: ./
-      target: production-apps
-    entrypoint:
-      - "yarn"
-      - "workspace"
-      - "@zooniverse/fe-project"
-    command: ["start"]
-    ports:
-      - "3000:3000"
+  # yarn:
```
camallen: This is very much a wip for quick testing, i'll reinstate these services once folks have tested locally and are happy with the solution

packages/app-content-pages/src/screens/Team/TeamContainer.js
```diff
@@ -32,8 +32,12 @@ TeamContainer.getInitialProps = async ({ req }) => {
   let error
   let teamData = []
   try {
+    console.log('FETCHING THE TEAM API DATA')
```
camallen: These will go - leaving here for logging the known errors. Perhaps we should better handle the unavailable API proxy error and thus the error reporting on the home page.

eatyourgreens: Does the original error get shown or is it replaced by a generic “something went wrong” message. Maybe show the specific error message on the page, after the generic error?

camallen: The only error that's shown is the one translated one by https://github.com/zooniverse/front-end-monorepo/blob/60d70e1c83bd991461e134a21499a50aa6895ab8/packages/app-content-pages/src/screens/Team/TeamContainer.js#L21

eatyourgreens: I'd add the error message from the request error, if you want a more useful message displayed in the browser.

docker-compose.yml
```diff
@@ -31,6 +31,8 @@ services:
       - "workspace"
       - "@zooniverse/fe-content-pages"
     command: ["start"]
+    environment:
+      - ASSET_PREFIX=http://localhost:3000
```
eatyourgreens: I'm testing out #1563 at the moment and I'm testing on `http://localhost.zooniverse.org:3000` and `http://localhost.zooniverse.org:3001` so that the Panoptes API works. Does the hostname make a difference here?

camallen: It probably should match to avoid cross origin issues and best practice would be to use `local.zooniverse.org:3001`. In theory this would work as the this prefix is valid as the content pages app is running on `http://localhost:3001` on the host machine.

I'm probably going to close this PR as i don't have time to work on it, it does however solve the issue of why the content pages app doesn't work from root docker-compose setup.

eatyourgreens: Yeah, I added this line to my set-up for testing #1563 and it worked just fine. I don't see any project icons on the publications page but I think that's a separate issue.

PR #1537, opened 2020-02-21
Fix 1536 - allow Classifier to submit Classifications without Crashing

### LINKED ISSUE #1536: Classifier crashes on Classification Submit, after PR 1493
## Major Functionality Bug

Package: `lib-classifier`

When a user submits a Classification on the Classifier, the whole page crashes to white with the following error message:

```
Uncaught TypeError: Cannot read property 'removeEventListener' of null
    at onUnmount (SVGPanZoom.js:66)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
    at safelyCallDestroy (react-dom.development.js:17203)
    at commitUnmount (react-dom.development.js:17540)
    at unmountHostComponents (react-dom.development.js:17873)
    at commitDeletion (react-dom.development.js:17904)
    at commitAllHostEffects (react-dom.development.js:18685)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
```

This error occurs as of PR #1493 (commit 9df22c02d69eaa0b067bc3b74bfbdbf166562c75), and we can safely narrow it down to the `onMount` and `onUnmount` actions in `SVGPanZoom`.

### Status

Major bug, but I think I can put a quick fix in. If that fails, I'll revert #1493 and re-approach the issue next week (Mon 24 Feb)

shaunanoordin:
## PR Overview

Fixes #1536 
Package: `lib-classifier`

This PR fixes issue 1536, where the Classifier would crash every time a Classification is submitted. This problem seems to stem from the `SVGPanZoom` component attempting to remove an event listener for an `undefined` DOM element on unmount, which in turn was caused by the use of `createRef()` instead of `useRef()`

### Dev Notes

Here are the tech deets:

- when _mounting the component,_ `scrollContainer.current` is correctly defined as a HTML node.
- when _unmounting_ however, `scrollContainer.current` is `undefined`
- Turns out, this is because `createRef()` doesn't play well with _functional components..._ because it apparently supposed to be called only once (i.e. `constructor () { this.myRef = createRef() }` ), and with functional components the initial value of `createRef()` is lost between renders/updates?
- Anyway, `useRef()` keeps a consistent ref across all the lifecycle stages of a _functional component._

Here's a better explanation: https://stackoverflow.com/questions/54620698/whats-the-difference-between-useref-and-createref

> The difference is that createRef will always create a new ref. In a class-based component, you would typically put the ref in an instance property during construction (e.g. this.input = createRef()). You don't have this option in a function component. useRef takes care of returning the same ref each time as on the initial rendering.

### Status

Ready for review

EDIT: WAIT, no, not ready for review. I think I spotted another small problem. Give me one sec to document...

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
@@ -23,19 +23,23 @@ function SVGPanZoom ({
 
   const [ zoom, setZoom ] = useState(1)
   const [ viewBox, setViewBox ] = useState(defaultViewBox)
+  
+  function preventDefault (e) {
+    e.preventDefault()
+  }
 
   function onMount () {
     setOnDrag(onDrag)
     setOnPan(onPan)
     setOnZoom(onZoom)
-    scrollContainer.current.addEventListener('wheel', e => e.preventDefault())
+    scrollContainer.current.addEventListener('wheel', preventDefault)
   }
 
   function onUnmount () {
     setOnDrag(() => true)
     setOnPan(() => true)
     setOnZoom(() => true)
-    scrollContainer.current.removeEventListener('wheel', e => e.preventDefault())
+    scrollContainer.current.removeEventListener('wheel', preventDefault)
   }
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


shaunanoordin:
## PR Update

If I'm reading this correctly, these two things don't work, I think? But I need someone to sanity check me on this.

```
function onMount () {
  ...
  scrollContainer.current.addEventListener('wheel', e => e.preventDefault())  // This creates ONE instance of a event-preventer-function
}

function onUnmount () {
  ...
  scrollContainer.current.removeEventListener('wheel', e => e.preventDefault())  // This attempts to remove ANOTHER instance of an event-preventer-function, doesn't it?
}
```

So the mount creates Instance 1, but unmount attempts to remove Instance 2, so Instance 1 stays put in all cases. The solution to this would be to turn `e => e.preventDefault()` into a named function, `function preventDefault (e) { e.preventDefault() }` and passing that to both addEventListener and removeEventListener.

I need to double check this because it's Friday evening and I'm on a bit of an adrenaline rush in preparation for board games night. BRB!

shaunanoordin:
Alright, the additional eventListener fix has been added. This is now ready for review! 👍

PR #1539, opened 2020-02-24
Transcription: show transcribed lines

eatyourgreens:
Show transcribed lines on top of a subject page: red for incomplete lines, grey for completed lines.

A screenshot of Anti-Slavery Manuscripts staging workflow 3085 showing lines drawn on top of a collaborative subject.

Package:
lib-classifier

Depends on #1517 and #1520 so merge those first. Maybe also #1540 to generalise the `TranscriptionLine` component so that it can be shared between views.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.spec.js
```diff
@@ -0,0 +1,31 @@
...
+    it('should be clickable', function () {})
+  })
+
+  describe('completed lines',function () {
+    it('should be grey', function () {})
+    it('should be labelled with the consensus text', function () {})
+    it('should show the label on hover', function () {})
+    it('should be focusable', {})
+    it('should show the label on focus', function ())
+    it('should not be clickable')
```
srallen: The focus/hover state of the completed lines is a help tooltip text:

> As a user I want to hover over others’ previous markings to see a hover state explaining that these markings are other users’ markings.

The consensus text is something that is from the click event to open up the sub-task that is no longer editable (or something similar)

> As a user I want to be able to read completed (greyed-out) transcriptions. 

Is there a reason to not do it like this?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.stories.js
```diff
@@ -0,0 +1,21 @@
...
+const config = {
+  notes: {
+    markdown: readme
+  }
+}
+
+storiesOf('TranscribedLines', module)
+
+  .add('default', () => (
+    <Grommet theme={zooTheme}>
```
srallen: Wrapping this with the `Grommet` component is not necessary for tests unless you are using the `withTheme` or `withThemeContext` HOCs.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.spec.js
srallen: From the user stories, the hover/focus state is for a tooltip:

> As a user I want to hover over others’ previous markings to see a hover state explaining that these markings are other users’ markings.

Then the `consensus_text` is viewed with an onclick event to show the subtask that is disabled from editing (or something similar). 

> As a user I want to be able to read completed (greyed-out) transcriptions. 

Is there a reason you're not scaffolding this to behave this way?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.stories.js
eatyourgreens: This is the default produced by the `plop` template.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.spec.js
eatyourgreens: ASM has code that I think shows the consensus text on focus/hover. That’s where these specs have come from.

eatyourgreens:
Clicking one of these lines should create a Transcription Line, which is then added to the active task annotation. They aren’t Transcription Lines themselves, however.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.spec.js
eatyourgreens: The other reason for the labelling here was to add descriptive alt text for the line.

eatyourgreens: Here's what I'm aiming for, from the ASM production collaborative workflow on this branch.
Screenshot of the browser DOM inspector, showing each SVG line described by the consensus text for that line.

Help text for each line can, of course, be added via `aria-describedby` or similar.

eatyourgreens:
#1540 generalises the `TranscriptionLine` component so that I think it could be used here. Clicking on an incomplete line would still draw a new mark, which is then added to the active task annotation. From conversations with @shaunanoordin, I believe that's how ASM handles editing previous transcriptions.

eatyourgreens:
Updated to use the `TranscriptionLine` component from #1540.
Screenshot showing grey and red marked lines on an ASM staging subject from workflow 3085.

eatyourgreens:
Here's the staging workflow (3392) showing the green and blue colours on top of an ASM page.
Screenshot of the dev classifier showing an ASM page with 3 blue lines and one green (active) line.

eatyourgreens:
NB. the staging projects (3085 and 3392) don't support `consensus_text` in reductions at the moment. Those can be tested with ASM on production at the moment. http://localhost:8080/?project=4973&workflow=5339&env=production

Retired lines (single line of consensus text):
Alert box for a retired line, showing the consensus line "Dear Mr. Garrison".

Partially transcribed lines (3 choices of possible text):
Alert box for a partially transcribed line, showing 3 choices for a date (with different years.)

eatyourgreens:
If Caesar has the start and end of the line reversed, then I'm assuming I should be able to move the start and end points of my annotation so as to correct that.

eatyourgreens:
Here's a good example of a complicated page with cross-writing, from ASM production ([subject 13968127](https://www.zooniverse.org/projects/bostonpubliclibrary/anti-slavery-manuscripts/talk/subjects/13968127).)
A complex page, with many cross-written lines marked in red.

beckyrother:
Transcription line colors:
Green: #06FE76
Blue: #235DFF
Red: #FF3C25
Grey: #8C8C8C

These are from https://projects.invisionapp.com/dsm/zooniverse/primary-brand/folder/colors/5c68fab7fb26bb001855f29a

eatyourgreens:
Thanks! I've updated the colours in the code.

Annotated page with updated line colours.

eatyourgreens:
Here are a couple of examples from ASM production (workflow 5339) with the new colours. The second has examples of lines where the start and end points are reversed, too.

A fully transcribed page from workflow 5339.
A partially completed page from workflow 5339.

eatyourgreens:
#1545 fixes the appearance of the open circles at the start of each line.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.spec.js
eatyourgreens: Simple tooltips are done but I think it would be more accessible to also show the current consensus when lines have focus.
A focused line, showing a tooltip clarifying that the line is complete.

beckyrother:
Thanks! @shaunanoordin has (I think?) been working on making those lines thicker. In the most recent screenshot here they're quite thin and I worry about them getting lost in the case of a very busy image. Will that update be included here as well?

eatyourgreens:
@beckyrother I forgot to correct the lines for the scaling of the subject image. Here's what they look like with that correction. The circles should now be 10px on the screen, not 10px on the subject image.
Screenshot of a completed page showing grey lines indicating the completed transcriptions.

srallen:
@eatyourgreens, based on a read through of the diff, is it correct that the remaining things to do for this are to write the specs for the component? The other items left in TODO comments and tracked in issues can be done in follow up PRs. Is there anything else missing that makes this WIP that should be done in this PR?

eatyourgreens:
Creating new marks doesn't really work but that could be quite complicated, after talking to @shaunanoordin about how it was done in ASM. This still feels like a rough sketch, to me, of how transcribed lines might be rendered.

I've added a story, using one of the subjects from ASM production, so that you can try it out in the browser without hacking the model code to work with existing ASM subjects and workflows.

eatyourgreens:
Still very rough, but the story now supports clicking a red line to create a new mark (green and blue lines.)

Storybook screenshot showing a page with red (incomplete) and grey (complete) lines. Two of the red lines have been used to create new transcription lines (green and blue.)

eatyourgreens:
There's still an open accessibility question, which is that colour (red, green, blue or grey) is being used to convey the state of a line. That prevents us from meeting the level A requirements for [WCAG 2 Success Criterion 1.4.1](https://webaim.org/articles/contrast/#sc141).

srallen:
> Still very rough, but the story now supports clicking a red line to create a new mark (green and blue lines.)

Red represents a finished line, so shouldn't create a new mark.

> There's still an open accessibility question, which is that colour (red, green, blue or grey) is being used to convey the state of a line. That prevents us from meeting the level A requirements for WCAG 2 Success Criterion 1.4.1.

I think that's what the tooltips are for, but I'll follow up as necessary. 

I think for this branch, we should just focus on adding the specs and any refinements or refactors happen in new PRs. I'm planning on picking this up and adding the `transcription` task model later this week. Thanks.

srallen:
Ok, I'll follow up on this now.

srallen:
I'm seeing a MST lifecycle race condition happen on this branch. Currently the lifecycle is:
- Create root store
- child nodes fire after attach
- For `Subject`, this fires a call to create `TranscriptionReductions` and a call to request caesar reductions happens after attachment
- For `SubjectStore`, this fires a call to request the subject queue at the same time because this particular workflow's linked subjects only has one subject, so another request for the subject queue fires immediately
- If the second request for the subject queue and appending to the store finishes before the request to caesar does, then `transcriptionReductions` gets set to undefined because these don't exist on the subject resource and it's the same subject
- The request to caesar fails (workflow 3392 has a config issue) with the MST error:
``` 
Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree. (Object type: 'TranscriptionReductions', Path upon death: '/subjects/resources/73330/transcriptionReductions', Subpath: 'error', Action: '/subjects/resources/73330/transcriptionReductions.fetchCaesarReductions()'). Either detach nodes first, or don't use objects after removing / replacing them in the tree.
```

I'm thinking about a few possible changes:
- Use optional type for transcription reductions, rather the maybe type. Always have a child `TranscriptionReduction` node attached similar to what we do on the `RootStore`.
- Update `TranscriptionReductions` to request caesar reductions when the subject id changes.

eatyourgreens:
If the workflow only has one subject, don't you run into a situation where subject 73330 is constantly being fetched and overwritten in the subjects map, since we map by subject ID?

eatyourgreens:
I think any pending API requests should be tracked and killed in `beforeDestroy`, if possible. That sounds like it would be best practice.

srallen:
Yeah that'd be a good idea too. It's rare set of criteria to trigger this race condition, but one we should address.

srallen:
I took a look at the linked PR, and flow cancellation hasn't been implemented: https://github.com/mobxjs/mobx-state-tree/pull/691

We could use the [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController), but this isn't spec, however, it's well supported across browser (because Google). The spec is in progress for observables: https://github.com/tc39/proposal-observable/issues/201

I think I have a solution that goes back closer to the root of the problem, which is requesting for the queue again and getting the same subject. I'll have a commit up in a moment.

srallen:
I've added a check to see if the subject already exists in the store's resources map. If not, then put it. We do something like this in the generic Resources store, but only if the request is for a single resource. This just adapts the same check for the queue populating.

eatyourgreens:
A real Panoptes project will always return 10 unique subjects from the API, even if it's the same 10 subjects when the available subject pool is empty. I wouldn't be massively worried about test subject sets that only have 1 subject. Maybe clone one of the ASM staging sets to test this?

I forgot to take WIP off the commit that I was using to test out ideas. Should that be done before pushing anything else to this branch?

srallen:
What I fixed would still be a problem if its the same 10 subjects in the response if we don't check for their existence already in the resources map. The queue immediately requesting for more subjects may or may not be an issue depending on how we want that to function, but in this case, it revealed the issue of how we were putting the subjects in the resources queue.

srallen:
The commits will be squashed on merge, so I don't really care about rewriting commit history. It was WIP when you made the commit, so that's an accurate representation of history, imo.

srallen:
What I'm working on now is tidying up the story to work and filling out the tests.

srallen:
I got the story working. It looks a bit funky because the mocked reduction data is not for the subject image:

Screen Shot 2020-03-18 at 4 47 40 PM

I'm gonna call this good for now, since it's mostly for proof of concept at this point, but we'll want to circle back to this story to have it make more sense.

srallen:
Gotta fix broken tests too. I did some renaming of a few variables since incorporating the `transcription` task.

eatyourgreens:
There's a GraphQL query in the story which will request ASM reductions for that specific page. Here's a screenshot, from earlier in the thread.

Storybook screenshot showing a page with red (incomplete) and grey (complete) lines. Two of the red lines have been used to create new transcription lines (green and blue.)

srallen:
Ok, I redid the story (thanks!) so it makes sense now, fixed broken tests, and got about half done writing tests for `TranscribedLines`. Should be able to finish tomorrow and then start on a story for displaying the previous transcriptions text.

srallen:
Must've left and only scope in, oops.

srallen:
I think this is ready for review. I ended up in the end doing a little bit of refactoring when I rebased to get the `transcription` task working with this, but it was mostly for my own readability and comprehension. I did a bit of clean up with the props in the `InteractionLayerContainer` for instance.

@shaunanoordin do you have time early next week to review this? The functionality can be tested with the story: http://localhost:6006/?path=/story/transcribedlines--default

The remaining functionality for this will come in separate PRs.

srallen:
> There are a number of empty xit tests in TranscribedLines.specs.js, but I assume this is because the functionality can't be built in yet.

Yes, this is correct. The specs will be built out per feature as we finish this up.

> While I understand that the "interact with lines" feature (i.e. clicking on a red line will transform it into a green line) is just a placeholder, I'm scratching my head about a few interaction scenarios.

How the colors are applied to the lines will be refactored when we add in the function to create a new annotation from a previous transcription. 

Just for FYI: 
We have a workflow on the staging ASM that is configured with the new transcription task, transcription line, and has caesar configurations set up. The issue is that there are no reductions and the data couldn't just be copied over from the original staging workflow that does have test classifications and reductions. So once this merges and a few more of the UI issues associated with this are addressed, we can ask team members to make classifications to populate reductions.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.spec.js
```diff
@@ -0,0 +1,107 @@
...
+        expect(consensusLineWrapper.props()['aria-label']).to.equal(completeLines[index].consensusText)
+      })
+    })
+
+    xit('should show the label on hover', function () {
+      lines.forEach((component, index) => {
+        const consensusLineWrapper = wrapper.find({ ['aria-describedby']: `complete-${index}`})
+        consensusLineWrapper.simulate('hover')
+      })
+    })
```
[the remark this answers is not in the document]
srallen: Not yet. This test isn't actually passing yet as it has no expectation and the default behavior for a test without an expectation is to pass it, so I marked it as pending so it was clear what the state of this is. I'm not sure how to test this without snapshots because the visibility of the tooltip is being handled by css.

PR #1540, opened 2020-02-25
Transcription line updates

eatyourgreens:
Set colours for line state: active, transcribed, completed or default.
Add start and end circles.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.js
@@ -2,10 +2,18 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import { DragHandle } from '@plugins/drawingTools/components'
 
-const FINISHER_RADIUS = 8
+const FINISHER_RADIUS = 6
 const GRAB_STROKE_WIDTH = 6
+const COLOURS = {
+  active: '#5cb85c',
+  default: '#00ced1',
+  transcribed: '#c33',
+  complete: '#979797'
+}
 
-function TranscriptionLine ({ active, mark, onFinish, scale }) {
+function TranscriptionLine ({ active, mark, onFinish, scale, state }) {
+  state = active ? 'active' : state
+  const colour = COLOURS[state]
   const { x1, y1, x2, y2, finished } = mark
   const finisherRadius = FINISHER_RADIUS / scale
 
@@ -19,10 +27,16 @@ function TranscriptionLine ({ active, mark, onFinish, scale }) {
   }
 
   return (
-    <g>
+    <g
+      color={colour}
+      fill={colour}
+      stroke={colour}
+      strokeWidth={1}
+    >
       <line x1={x1} y1={y1} x2={x2} y2={y2} />
+      <line x1={x1} y1={y1} x2={x2} y2={y2} strokeWidth={GRAB_STROKE_WIDTH / scale} strokeOpacity='0' />
 
-      {active &&
+      {active ?
         <DragHandle
           fill='transparent'
           radius={6}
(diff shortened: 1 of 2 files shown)


shaunanoordin:
I'm currently trying to test this on Anti-Slavery Manuscripts (staging) - `http://localhost:8080/?env=staging&project=1764&workflow=3085` - but I'm getting weird errors because monorepo's calls to the `/queued` endpoint specifies workflow ID, but not `subject_set_id`. Uh, I might be missing something here, gimme a sec to double check.

Which project/URL did you test this on, btw?

eatyourgreens:
Try workflow 3372, which has a drawing task with a couple of transcription lines.
http://localhost:8080/?project=908&workflow=3372

#1520 fixes the error where a workflow can't be loaded from Panoptes if it uses grouped subject selection.

eatyourgreens:
There's also a couple of new stories to show the colours for lines that have been transcribed and lines that have been completed/retired. `yarn storybook` from `packages/lib-classifier` to start up a local copy of the classifier storybook.

shaunanoordin:
Ah, solid! 👍 Sorry, I made the mistake of jumping straight into testing for _Caesar-enabled projects_ as soon as I saw the word "Transcribed Lines", when I should have just tested for the basic Transcription line.

Colours look good!

eatyourgreens:
I'm wondering if there's a case where someone would really want to use two transcription line tools in a drawing task. I'm probably overthinking this.

shaunanoordin:
> I'm wondering if there's a case where someone would really want to use two transcription line tools in a drawing task. I'm probably overthinking this.

I was actually thinking about this too, but my conclusion was that we can cross that bridge when it users ask for it.

I already have a few solutions bookmarked if/when that arises:

- All Transcription Lines respect their tool's user-defined colour, but...
  - give active lines a solid white outline or a glowing effect (colour animation), or
  - pass the tool's user-defined colour through a state-based transformer, e.g.
```
function getColour (col, state)
  switch (state) {
    case 'active':
      return brighten(col)
    case 'transcribed':
      return darken(col)
    ...
  }
}
```

eatyourgreens:
>I was actually thinking about this too, but my conclusion was that we can cross that bridge when it users ask for it.

That's a sensible philosophy.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/TranscriptionLine.js
```diff
@@ -2,10 +2,18 @@ import PropTypes from 'prop-types'
...
+  active: '#5cb85c',
+  default: '#00ced1',
+  transcribed: '#c33',
+  complete: '#979797'
+}
 
-function TranscriptionLine ({ active, mark, onFinish, scale }) {
+function TranscriptionLine ({ active, mark, onFinish, scale, state }) {
+  state = active ? 'active' : state
+  const colour = COLOURS[state]
```
shaunanoordin: Note that if state is somehow an invalid value, the colour of the T-Line will default to the tool's user-designated colour.

beckyrother:
>I'm wondering if there's a case where someone would really want to use two transcription line tools in a drawing task. I'm probably overthinking this.
>I was actually thinking about this too, but my conclusion was that we can cross that bridge when it users ask for it.

I'd suggest splitting that into two tasks within the transcription workflow – no need to add so much complication for us or for volunteers

eatyourgreens:
We can work on this more in #1539 if need be.

srallen:
The hard coded colors are intended for use within a `transcription` task type, so that the `transcriptionLine` tool would function as all other drawing tools in this scenario.

beckyrother:
Thanks Shaun. As I mentioned above, there shouldn't be multiple transcription types in the same task, so the English/Spanish differentiation will not be an issue. 

How would it look to add a thin white stroke around the lines to increase background contrast?

beckyrother:
Sorry, didn't realize this had been closed already. Was hoping for a final design review before it was merged.

eatyourgreens:
@beckyrother it's now part of #1539 so you can still review there.

eatyourgreens:
@beckyrother here's a link to the ASM staging project, where you can also try out the transcription tool.
https://frontend.preview.zooniverse.org/projects/wgranger-test/anti-slavery-testing/classify?env=staging&workflow=3392

PR #1541, opened 2020-02-25
[VSLCV] Y-axis inversion control

### LINKED ISSUE #1425: [VSLCV] Axis inversion for composite viewer
**Package** lib-classifier

The composite subject viewer needs a button that triggers the inversion of the y-axis. This will make the calculations and pass down the updated data via props to the scatter plot. 

Time estimate: 1 week

**Additional context**

Design: https://projects.invisionapp.com/d/main#/console/17294981/360635448/inspect

srallen:

Package: lib-classifier

Closes #1425

Describe your changes:
This adds the functionality for the y-axis inversion for the `VariableStarViewer`. Also finishing touches on the styles for the button and for the y-axis on the smaller raw JSON data scatter plot. The axis now take a `numTicks` prop. The phased plot, which doesn't display points yet, will also have its y-axis inverted when the button is hit. 

I've added some specs for the ordering of the axes now. There was an error previously with the y-axis already being inverted that I hadn't caught. this is because I had the range parameters setup incorrectly for the d3 linear scale. The section on this page about "A flipped range" explains it a little better than the standard d3 docs, imo: https://observablehq.com/@d3/d3-scalelinear

This is ready for design and code review. ~WIP because I need to update the readme docs.~

Here's a gif demonstrating the functionality:

flip-example

# Review Checklist

## General

## Components
- [x] Is the component accessible? 
  - [x] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [x] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
@beckyrother good question. The "Flip" text gets the underline, but the icon doesn't have a focus/hover state change. What would you like it to do?

beckyrother:
Let's do the same as the red/blue buttons: 
image

I'm hesitant on the flip underline – since the other tools don't have an underline when you interact with them, for consistency I don't think it should either. Thoughts?

srallen:
> I'm hesitant on the flip underline – since the other tools don't have an underline when you interact with them, for consistency I don't think it should either. Thoughts?

It's an auto-styling being done by Grommet because I've set the button to be `plain` meaning it doesn't apply the border or background color like standard buttons, however, the `plain` styling assumes you want the button to be styled more like an anchor link, which uses an underline for hover/focus. This can be easily modified to use styles like the checkboxes. 👍

srallen:
@beckyrother I've updated the focus and hover styles. Does this look ok?

Screen Shot 2020-02-28 at 4 20 53 PM

beckyrother:
Looks great!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer/helpers/utils.spec.js
```diff
@@ -113,4 +133,197 @@ describe('ScatterPlotViewer > helpers > utils', function () {
...
+          expect(point.y).to.be.a('number')
+        })
+      })
+    })
+
+    describe('when there is multiple series of data', function () {
+      it('should return the same data JSON object input as the function parameter', function () {
+        const inputData = lightCurveMockData.variableStar.data
+        const points = getDataPoints(inputData)
+        expect(points).to.deep.equal(inputData)
```
wgranger: What is the purpose of this function (`getDataPoints`)? It seems a bit odd to feed data into a function, and expect the return to be the exact same as the data that was sent in.

srallen: The `ScatterPlotViewer` supports two JSON formats: one type that only can be for a single series of data aka TESS's JSON format and one type that supports either single or multiple data series (The goal is to eventually migrate the TESS subject viewer to use the more general `ScatterPlotViewer` since it is tested and the original d3 LCV mostly is not). The utility function here converts the first type (TESS-like single data series) to be like the latter type (can be multiple data series), but if the data from the subject JSON file is already in that structural format, then just return it as is. 

I'll plan on adding some info about this into the scatter plot readme to clarify what's going on here in a follow up PR. Thanks for the question.

beckyrother:
Oh lord sorry, don't know how to use github. this looks great, @wgranger please disregard

PR #1544, opened 2020-02-27
Fix 1543 - inactive drawing marks must not be draggable

### LINKED ISSUE #1543: Inactive Drawing Marks can still be drag-moved
## UI Bug

Package: `lib-classifer`

Expected: If you make drawing annotation marks in Step X of the workflow, then in step X+1 those drawing marks should be visible but not _interactive._ i.e. You should not be able to edit or delete marks made in a previous Workflow Step.

Actual: drawing marks from a previous step CAN be _moved/repositioned_ by the drag action.

Screen Shot 2020-02-26 at 17 39 08
_Two rectangles drawn in Step 1._

Screen Shot 2020-02-26 at 17 39 24
_The red rectangle is moved in Step 2._

### Status

A fix is already being pushed, this is just the documentation.

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Fixes #1543 

This PR ensures that when a drawing mark isn't active (i.e. it was made in a previous workflow step), it can't be moved.

### Status

Ready for review

EDIT: Just saw @eatyourgreens's comment in https://github.com/zooniverse/front-end-monorepo/issues/1543#issuecomment-591973928

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingToolMarks/DrawingToolMarks.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingToolMarks/DrawingToolMarks.js
@@ -48,9 +48,9 @@ function DrawingToolMarks ({ activeMarkId, marks, onDelete, onDeselectMark, onFi
         key={mark.id}
         isActive={isActive}
         coords={mark.coords}
-        dragStart={selectMark}
-        dragMove={moveMark}
-        dragEnd={deselectMark}
+        dragStart={isActive && selectMark}
+        dragMove={isActive && moveMark}
+        dragEnd={isActive && deselectMark}
         label={`Mark ${index}`}
         mark={mark}
         onDelete={deleteMark}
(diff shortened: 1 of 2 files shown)


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingToolMarks/DrawingToolMarks.js
```diff
@@ -29,6 +29,7 @@ function DrawingToolMarks ({ activeMarkId, marks, onDelete, onDeselectMark, onFi
     }
 
     function moveMark (event, difference) {
+      if (!isActive) return;
```
eatyourgreens: Looks good. Maybe this should use a callback prop, `onMoveMark`, like the other handlers? Then we can control draggable behaviour from the parent, and turn it off completely if necessary.

shaunanoordin:
## PR Update

This PR now implements an `onMove` handler for DrawingToolMarks, which in turn means that the actual _action of moving a mark_ (mark.move()) is handled by the parent `InteractionLayer`

Please let this comment actually go through. I've been trying to update this thread for the past half and hour to no avail, for the GitGods have forsaken me.

eatyourgreens:
Changes look good. I don't know if CI is running at the moment.

eatyourgreens:
Tests are passing locally, so I think this is good, once Travis does its thing.

eatyourgreens:
The WIP check is stopping this from merging. 😡

shaunanoordin:
WTF, WIP?

shaunanoordin:
Wait let me try changing the title of this PR to see if that bumps it

shaunanoordin:
Solved!

eatyourgreens:
Ingenious!

PR #1550, opened 2020-03-02
Classifier: Add the move cursor to draggable subjects

### LINKED ISSUE #1489: SingleImageViewer pan cursor
**Package** lib-classifier

From the review of #1488 

> Overall, this PR looks good. Only suggestion I would make is to maybe have the cursor change to cursor: move with hover when pan is enabled? However, that's certainly not blocking as it's purely a UI suggestion.

eatyourgreens:
Show the move cursor when subject images can be panned by dragging.

Package:
lib-classifier

Closes #1489.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
@@ -4,6 +4,7 @@ import { storiesOf } from '@storybook/react'
 import zooTheme from '@zooniverse/grommet-theme'
 import { Box, Grommet } from 'grommet'
 import { Provider } from 'mobx-react'
+import SubjectViewerStore from '@store/SubjectViewerStore'
 import SingleImageViewer, { SingleImageViewerContainer } from './SingleImageViewerContainer'
 import ZoomInButton from '../../../ImageToolbar/components/ZoomInButton/ZoomInButton'
 import ZoomOutButton from '../../../ImageToolbar/components/ZoomOutButton/ZoomOutButton'
@@ -23,16 +24,6 @@ const subject = {
   ]
 }
 
-let zoomCallback
-
-function onZoom(type) {
-  zoomCallback(type)
-}
-
-function setZoomCallback(callback) {
-  zoomCallback = callback
-}
-
 const mockStore = {
   classifications: {
     active: {
@@ -42,10 +33,9 @@ const mockStore = {
   drawing: {
     addToStream: sinon.stub()
   },
-  subjectViewer: {
-    enableRotation: () => null,
-    setOnZoom: setZoomCallback
-  },
+  subjectViewer: SubjectViewerStore.create({
+    move: true,
+  }),
   workflowSteps: {
     activeStepTasks: []
   }
@@ -68,27 +58,25 @@ const darkThemeConfig = Object.assign({}, config, { backgrounds: backgrounds.dar
 storiesOf('Subject Viewers | SingleImageViewer', module)
   .add('light theme', () => {
     return (
-      <Grommet theme={zooTheme}>
+      <ViewerContext theme={zooTheme}>
         <Box height='medium' width='large'>
-          <SingleImageViewerContainer
-            enableInteractionLayer={false}
+          <SingleImageViewer
             subject={subject}
           />
         </Box>
-      </Grommet>
+      </ViewerContext>
     )
   }, config)
   .add('dark theme', () => {
     const darkZooTheme = Object.assign({}, zooTheme, { dark: true })
     return (
-      <Grommet theme={darkZooTheme}>
+      <ViewerContext theme={darkZooTheme}>
         <Box height='medium' width='large'>
-          <SingleImageViewerContainer
-            enableInteractionLayer={false}
+          <SingleImageViewer
             subject={subject}
           />
         </Box>
-      </Grommet>
+      </ViewerContext>
     )
   }, darkThemeConfig)
   .add('pan and zoom', () => {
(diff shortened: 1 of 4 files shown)


eatyourgreens:
Yeah. The connection is indirect, in that `move` disables the interaction layer styles. I’ve not been seeing the move cursor when drawing is enabled.

eatyourgreens:
I've added the `move` flag as a viewer prop. The viewer now renders an SVG `<image />` if `move` is false. While drawing with the mouse/touchpad, the subject can still be panned and zoomed from the keyboard (which matches existing behaviour in PFE and PH-TESS.)

I also rewrote the tests to use async `beforeEach`, so that I could add a test for changing the `move` prop from false to true.

The multi-image viewer might need to be updated to add that change too.

eatyourgreens:
I've fixed the stories so that the light/dark theme stories don't throw an error. The pan/zoom story should work from the keyboard now, when the SVG has focus.

eatyourgreens:
The default focus indicator is hard to make out in Firefox but here is the storybook zoomed and panned with the keyboard.
Screenshot of the storybook showing the single image viewer zoomed in.

Zooming/panning with the mouse is broken but I can't see why.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
```diff
@@ -23,16 +24,6 @@ const subject = {
...
-let zoomCallback
-
-function onZoom(type) {
-  zoomCallback(type)
-}
-
-function setZoomCallback(callback) {
-  zoomCallback = callback
-}
-
```
srallen: Removing these breaks the zoom buttons in the third story with pan and zoom. They get covered up by the image itself because there isn't a layout setup to make sure that doesn't happen. The stories could be updated to better reflect the image toolbar usage. Maybe actually import it and connect it if we're going to mock the store.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
```diff
@@ -42,10 +33,9 @@ const mockStore = {
   drawing: {
     addToStream: sinon.stub()
   },
-  subjectViewer: {
-    enableRotation: () => null,
-    setOnZoom: setZoomCallback
```
srallen: Same comment about removing the zoom callback.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
```diff
@@ -42,10 +33,9 @@ const mockStore = {
...
-  },
+  subjectViewer: SubjectViewerStore.create({
+    move: true,
```
srallen: I don't think all the stories should be demoing the same pan/zoom state? It makes it seem like the subject viewer is always in this mode.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
eatyourgreens: I think the button layout is broken on master. Mark commented on it recently.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
eatyourgreens: I don’t think they are. `move` is false for the first two.

eatyourgreens: I’m confused by this code too. `move` should not be true for all three stories.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
eatyourgreens: I've fixed the layout by moving the buttons above the subject viewer. I've also imported the connected buttons so that they use the `subjectViewer` model (and work continuously on pointer down.)

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
eatyourgreens: I think setting `move: true` might be right. `move` is always true in the actual subject viewer, unless we're using the mouse to draw on, not drag, the image.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
eatyourgreens: I've also added keyboard shortcuts to those buttons.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
eatyourgreens: I've expanded the tools in the story to include the toggle between annotating and panning the subject. With the default `subjectViewer` model, annotation is enabled but PFE has movement enabled by default unless the current task is a drawing task.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.stories.js
srallen: Looks good. Thanks!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
```diff
@@ -194,4 +204,4 @@ SingleImageViewerContainer.defaultProps = {
 class DecoratedSingleImageViewerContainer extends SingleImageViewerContainer { }
 
 export default DecoratedSingleImageViewerContainer
-export { SingleImageViewerContainer }
+export { DraggableImage, SingleImageViewerContainer }
```
srallen: I think this might be helpful for the VSLCV. :+1:

Issue #1511, opened 2020-02-12
Possible incorrect usage of observer + inject decorators for task components


srallen:
**Package** lib-classifier

**Describe the bug**
The console is currently warning twice about this:
`Mobx observer: You are trying to use 'observer' on a component that already has 'inject'. Please apply 'observer' before applying 'inject'`

Can be observed here: http://localhost:8080/?project=1862

The stack trace points to this line: 

https://github.com/zooniverse/front-end-monorepo/blob/52a73f9500d61f8771aeec458eb11e557effb4db/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js#L59

This warning doesn't pop up for all task types and doesn't seem to immediately cause issue, however, the attempt in #1496 to add the `withTheme` decorator to the single choice task and multiple choice task caused issues. With the single choice task, some integration tests broke. With multiple choice task, issue #1507 happened. The fix was to revert the addition of the `withTheme` HOC. 

We need to investigate a better way to wrap task components with the observer HOC to make sure they're observing for changes from the store observables and then re-render. 

First, we could consider migrating some components away from `inject` since we are on mobx-react v6 with the classifier. Second, we should have integration tests for all of the task types.

eatyourgreens:
`DataVisAnnotation` is the only task that uses `inject`.
https://github.com/zooniverse/front-end-monorepo/blob/7aceaca920df06d831e2de8e653d8d3291b2e5a2/packages/lib-classifier/src/plugins/tasks/DataVisAnnotationTask/components/DataVisAnnotationTask.js#L36-L38

It has completely different props from the other tasks, which are passed a task and an annotation.

eatyourgreens:
The problem with the `withTheme` decorator might be related to this issue (I'm not sure): https://github.com/mobxjs/mobx-react/issues/471

PR #1551, opened 2020-03-02
Classifier: Pass the theme prop to task components

eatyourgreens:
Add the `withTheme` decorator to the task area, then pass the theme down to individual tasks as a prop.

See #1507 and #1511 for context. `observer` can't be applied to a component that's been decorated with `withTheme`, so we have to be careful how the theme is passed down to observed tasks.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
@@ -3,6 +3,7 @@ import { Box, Paragraph } from 'grommet'
 import { inject, observer } from 'mobx-react'
 import PropTypes from 'prop-types'
 import React from 'react'
+import { withTheme } from 'styled-components'
 
 import TaskHelp from './components/TaskHelp'
 import TaskNavButtons from './components/TaskNavButtons'
@@ -40,7 +41,7 @@ class Tasks extends React.Component {
   }
 
   [asyncStates.success] () {
-    const { addAnnotation, classification, isThereTaskHelp, subjectReadyState, step, tasks } = this.props
+    const { addAnnotation, classification, isThereTaskHelp, subjectReadyState, step, tasks, theme } = this.props
     const ready = subjectReadyState === asyncStates.success
     if (classification && tasks.length > 0) {
       // setting the wrapping box of the task component to a basis of 246px feels hacky,
@@ -61,10 +62,11 @@ class Tasks extends React.Component {
               return (
                 <Box key={annotation.id} basis='auto'>
                   <TaskComponent
+                    {...this.props}
                     disabled={!ready}
                     annotation={annotation}
                     task={task}
-                    {...this.props}
+                    theme={theme}
                   />
                 </Box>
               )
@@ -102,6 +104,7 @@ Tasks.defaultProps = {
 }
 
 @inject(storeMapper)
+@withTheme
 @observer
 class DecoratedTasks extends Tasks {}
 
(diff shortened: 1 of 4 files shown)


packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
```diff
@@ -101,6 +103,7 @@ Tasks.defaultProps = {
   tasks: []
 }
 
+@withTheme
```
srallen: Quick suggested change here, from [the docs](https://github.com/mobxjs/mobx-react#provider-and-inject) on `inject` and `observer`:

> When using both `@inject` and `@observer`, make sure to apply them in the correct order: observer should be the inner decorator, inject the outer. There might be additional decorators in between.

So, this should be:
```
@inject(storeMapper)
@withTheme
@observer
```

eatyourgreens: Sounds good. I copied the order here from another component, so it might need to be fixed somewhere else.

eatyourgreens: The Field Guide was applying the decorators in this order, so I've fixed that here.

packages/lib-classifier/src/plugins/tasks/MultipleChoiceTask/components/MultipleChoiceTask.js
```diff
@@ -30,7 +29,8 @@ function MultipleChoiceTask (props) {
     annotation,
     className,
     disabled,
-    task
+    task,
+    theme
```
srallen: A proptype and default props should be added for this.

packages/lib-classifier/src/plugins/tasks/MultipleChoiceTask/components/MultipleChoiceTask.js
```diff
@@ -5,12 +5,11 @@ import PropTypes from 'prop-types'
 import styled from 'styled-components'
```
srallen:
```suggestion
import styled, { css } from 'styled-components'
```

packages/lib-classifier/src/plugins/tasks/MultipleChoiceTask/components/MultipleChoiceTask.js
```diff
@@ -5,12 +5,11 @@ import PropTypes from 'prop-types'
...
 
 import TaskInput from '../../components/TaskInput'
-import zooTheme from '@zooniverse/grommet-theme'
 
 const maxWidth = pxToRem(60)
 const StyledBox = styled(Box)`
   img:only-child, svg:only-child {
-    background: ${zooTheme.global.colors.brand};
+    ${props => props.theme && css`background: ${props.theme.global.colors.brand};`}
```
srallen: TypeError here with css being undefined.

packages/lib-classifier/src/plugins/tasks/SingleChoiceTask/components/SingleChoiceTask.js
```diff
@@ -4,14 +4,11 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import styled, { css, withTheme } from 'styled-components'
```
srallen:
```suggestion
import styled, { css } from 'styled-components'
```

eatyourgreens:
Thanks @srallen those imports and prop definitions should be fixed now.

PR #1552, opened 2020-03-02
Add frame scroll handler

mcbouslog:
Package: lib-classifier

Describe your changes:
- adds refs, scroll handlers so active frame is visible if list of frames has overflow
- misc. styling cleanup

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/FrameCarousel.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/FrameCarousel.js
@@ -49,15 +49,24 @@ export function NextLabel () {
   )
 }
 
-export const StyledInput = styled.input`
+export const StyledFrame = styled.label`
   ${props => props.theme && css`
-    &:checked + img {
-      outline: ${props.theme.global.colors['neutral-4']} solid;
+    input:checked + img {
+      border: ${props.theme.global.colors['neutral-4']} solid;
+    }
+    input:focus + img {
+      outline: 2px solid ${tint(0.5, props.theme.global.colors.brand)};
+    }
+
+    input:hover + img {
+      outline: 2px solid ${tint(0.5, props.theme.global.colors.brand)};
     }
   `}
-  clip: rect(0 0 0 0);
-  overflow: hidden;
-  position: absolute;
+  input {
+    opacity: 0.01;
+    position: absolute;
+  }
+
 `
 
 export const StyledImage = styled.img`
@@ -97,21 +106,20 @@ class FrameCarousel extends React.Component {
       const url = location[mimeType]
       const activeFrame = frame === index
       return (
-        <label
+        <StyledFrame
           key={`${url}-${index}`}
         >
-          <StyledInput
+          <input
             checked={activeFrame}
             name='frame'
             onChange={() => onFrameChange(index)}
-            src={url}
             type='radio'
           />
           <StyledImage
             alt={counterpart('MultiFrameViewer.FrameCarousel.thumbnailAltText')}
             src={url}
           />
-        </label>
+        </StyledFrame>
       )
     })
 
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/FrameCarousel.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/FrameCarousel.js
@@ -71,20 +71,18 @@ export const StyledFrame = styled.label`
 `
 
 export const StyledImage = styled.img`
-  float: center;
-  height: 3em;
-  margin: 0.5em;
+  height: 40px;
   object-fit: cover;
   padding: 0;
-  width: 3em;
+  width: 40px;
 `
 
 class FrameCarousel extends React.Component {
   constructor () {
     super()
+    this.handleScroll = this.handleScroll.bind(this)
     this.handlePrevious = this.handlePrevious.bind(this)
     this.handleNext = this.handleNext.bind(this)
-    this.handleScroll = this.handleScroll.bind(this)
     this.activeLabel = React.createRef()
     this.frameList = React.createRef()
   }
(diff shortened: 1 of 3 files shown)


srallen:
@mcbouslog I'm gonna have a closer look at this to get a better idea of what might be going on. Any further insight since we last talked? Totally ok if not.

mcbouslog:
I force-pushed some changes (cleaned up commits, fixed tests, apologies if causes issues), but I have not been able to determine why the document scrolls when going from the 5th to 6th frame in the storybook.

I tried console logging `onScroll` from the `ul` element, `onFocus` from the `input`, and I think it has to do with the `onFocus` (like you mentioned), but `preventDefault` and `stopPropagation` did not help with initial efforts.

srallen:
@mcbouslog Have a look at [my debugging branch ](https://github.com/zooniverse/front-end-monorepo/tree/frame-scroll-debug). There were a few things going on:

- Some wonky CSS causing the offset top calculation to not be reliable. I removed the margins and em units so the math was reliable. Also added some missing `position: relative` CSS and removed invalid CSS (`float: center`)
- Then I made the scroll handler a callback for the frame change. 

I think the scroll is working as intended now... There still needs to be some clean up from my branch, so feel free to cherry-pick from it. The styles will need fixing, but let's do that in a separate PR and get the working scroll with specs in this one. Then we can apply styles and make sure they don't break the scroll.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/FrameCarousel.js
```diff
@@ -49,73 +49,108 @@ export function NextLabel () {
...
+    this.frameList = React.createRef()
+  }
+
+  componentDidMount () {
+    this.handleScroll()
+  }
+
+  handleScroll () {
+    const labelIsAboveContainerTop = this.activeLabel.current?.offsetTop <= this.frameList.current?.scrollTop
+    const labelIsBelowContainerBottom = this.activeLabel.current?.offsetTop >= (this.frameList.current?.scrollTop + this.frameList.current?.clientHeight)
```
mcbouslog: Added `?` for testing, I think ok per:
Screen Shot 2020-03-04 at 2 39 18 PM

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/FrameCarousel.js
```diff
@@ -49,73 +49,108 @@ export function NextLabel () {
...
+
+    if (labelIsAboveContainerTop || labelIsBelowContainerBottom) {
+      const newContainerTopPosition = this.activeLabel.current.offsetTop - this.activeLabel.current.offsetHeight
+      this.frameList.current.scrollTop = newContainerTopPosition
+    }
+  }
+
+  handleFrameChange (frameIndex) {
+    const { onFrameChange } = this.props
+    onFrameChange(frameIndex)
```
mcbouslog: I've omitted the callback, and I think it still works because of the CSS changes you noted? I think because the focus is properly updating, but I'm not entirely sure.

mcbouslog: I have the changes (adding callback to SubjectViewerStore, MultiFrameContainer, and to handleFrameChange in FrameCarousel) commented out locally, can quickly add and commit if I'm mistaken.

mcbouslog: With the callback, I was seeing it fall behind if I went to the bottom, then looped around with arrow down/right to top, then back to bottom with keys.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/FrameCarousel.js
srallen: It's stage 4, so should be fine to use!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/FrameCarousel.js
srallen: Quite possible it's not necessary with correct focus handling. I'll have a look.

mcbouslog: I think if I add the `handleScroll` to `handleFrameChange` it causes the strange mouse behavior you note, but I'm trying to gather more details and specifics to confirm and explain, I'll update with more info.

mcbouslog: Which actually makes sense (sorta) as it comes back to what's getting focus, as in when clicking or using arrow keys I'm moving focus, but with buttons I'm not changing the focus.

mcbouslog:
I think clicking frames, keyboard navigation on frames and prev/next buttons update scroll as expected now, but could definitely use a double-check.

eatyourgreens:
Have you considered using [`Element.scrollIntoView()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView)? We use it on Talk to [scroll long markdown posts](https://github.com/zooniverse/markdownz/blob/93718c40e8aa9ec40e7813e55a50699cd6dd8586/src/components/markdown.jsx#L64-L67). Scripting might not be necessary here since default browser behaviour is to scroll the focussed element into view anyway.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/FrameCarousel.js
srallen: Yep. This is now working as it should with the lifecycle method and adding the scroll handler did cause the incorrect behavior with the click interaction. Thanks.

srallen:
@eatyourgreens yep, we talked about that, but as you've already noted, we don't need to use it since focus state is on the radio buttons or next/previous button already.

PR #1553, opened 2020-03-03
Classifier: Tasks: Test all registered task models

eatyourgreens:
Import the task registry into the classifier Tasks tests, then run the tests once for each registered task type.

#1511 noted that we were only testing the single choice answer task, so this PR expands the tests to cover all available tasks.

This PR also updates the Tasks component tests to test against the root store, with a mock project, workflow and subject.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.spec.js
```diff
@@ -1,133 +1,170 @@
...
+        }
+      }, {
+        client: {
+          panoptes: {
+            get: sinon.stub().callsFake(() => Promise.resolve({ body: {}}))
+          }
+        }
+      })
+      classification = rootStore.classifications.active
+      addAnnotation = rootStore.classifications.addAnnotation
```
eatyourgreens: If we try to use destructuring here
```js
({ addAnnotation } = rootStore.classifications)
```
then the interpreter, without any semicolons to guide it, will try to execute `classifications.active` as a function and throw an error.
```js
classification = rootStore.classifications.active({ addAnnotation } = rootStore.classifications)
```

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.spec.js
```diff
@@ -1,133 +1,170 @@
...
-            loadingState={asyncStates.success}
-            subjectReadyState={asyncStates.success}
-            step={step}
-            tasks={tasks}
+            tasks={step.tasks}
           />
         )
-        taskWrapper = wrapper.find('SingleChoiceTask')
+        // Is there a better way to do this?
+        expect(wrapper.find(TaskComponent.displayName)).to.have.lengthOf(1)
```
eatyourgreens:
```js
wrapper.find(TaskComponent)
```
returns 0 nodes, so I've fallen back on searching by display name. I can't see why searching for the component itself would fail.

srallen: FWIW, my WIP Grommet upgrade branch is updating Enzyme use to take advantage of the new options available for injecting context. I'm not sure if there have been any improvements in the minor version update for this.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.spec.js
srallen: I don't really have a strong opinion about linter rules other than consistency and also not spending any time on deciding what those rules should be... 

Reading [the docs from standard](https://standardjs.com/rules.html#semicolons), I think they'd suggest to do something like:
`;({ addAnnotation } = rootStore.classifications)`

But as I said, I don't actually care what the rule is all that much, just that we stay consistent. There's a variant of standard that uses semi-colons if you feel strongly about it: https://github.com/standard/semistandard

eatyourgreens: No strong opinions from me except to be mindful that the interpreter doesn’t always insert semicolons where you’d expect. When writing/reviewing, be wary of declarations that begin with `(` or `[`.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.spec.js
```diff
@@ -1,133 +1,188 @@
...
+          }
+        }
+      }, {
+        authClient: {
+          checkBearerToken: sinon.stub().callsFake(() => Promise.resolve(null)),
+          checkCurrent: sinon.stub().callsFake(() => Promise.resolve(null))
+        },
+        client: {
+          panoptes: {
+            get: sinon.stub().callsFake(() =>
```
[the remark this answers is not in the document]
eatyourgreens: Good point. It should work here. I’ve been using it in other PRs, and I’ve found one bug that I should write up an issue for. There are circumstances where it will return `[ undefined ]` when it should return `[]`.

PR #1555, opened 2020-03-04
Classifier: Consolidate store subject observers

### LINKED ISSUE #1521: Classifier: Last workflow step is still selected when a new classification starts
**Package**
lib-classifier

**Describe the bug**
On starting a new classification for the second, and subsequent, subject in the queue, the first annotation that is added to the new classification is the annotation for the last task in the workflow.

Here we log calls to `classifications.addAnnotation(task, value)` on classification complete. Task `T3`, the last task, is added to the new classification before `T0`, the first task.
Screenshot of the console log, showing annotations being added to a classification out of order.

When the classification is submitted, the order of the annotations array is `[ T3, T0, T1, T2]`, not `[T0, T1, T2, T3]`.

**To Reproduce**
Classify more than one subject on project 1844, which has four steps.
http://localhost:8080/?project=1844

eatyourgreens:
Consolidate the subject observers from the feedback, workflow steps and classifications stores into one function in the root store. When the subject changes, reset the workflow steps, create a new classification, then reset subject feedback. I think this will fix #1521 by making sure the first task is selected before the classification is created.

This PR also fixes a couple of infinite loops that showed up in the Tutorial Store tests (infinitely requesting new subjects) and the UPP Store tests (infinitely requesting a new workflow.) 

I've also moved the Feedback `onAction` listener up to the root in order to mute warnings from that store. Classification and Mark `onAction` listeners are replaced with `beforeDestroy` actions to remove those warnings too. Otherwise, I was finding the console so noisy that failing tests were impossible to debug.

Package:
lib-classifier

Closes #1521.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
@@ -11,18 +11,16 @@ import TaskNavButtons from './components/TaskNavButtons'
 import taskRegistry from '@plugins/tasks'
 
 function storeMapper (stores) {
-  const { active: classification, addAnnotation } = stores.classifierStore.classifications
+  const { active: classification } = stores.classifierStore.classifications
   const { loadingState } = stores.classifierStore.workflows
-  const { active: step, activeStepTasks: tasks, isThereTaskHelp } = stores.classifierStore.workflowSteps
+  const { active: step, isThereTaskHelp } = stores.classifierStore.workflowSteps
   const { loadingState: subjectReadyState } = stores.classifierStore.subjectViewer
   return {
-    addAnnotation,
     classification,
     isThereTaskHelp,
     loadingState,
     step,
     subjectReadyState,
-    tasks
   }
 }
 
@@ -41,9 +39,9 @@ class Tasks extends React.Component {
   }
 
   [asyncStates.success] () {
-    const { addAnnotation, classification, isThereTaskHelp, subjectReadyState, step, tasks, theme } = this.props
+    const { classification, isThereTaskHelp, subjectReadyState, step, theme } = this.props
     const ready = subjectReadyState === asyncStates.success
-    if (classification && tasks.length > 0) {
+    if (classification && step.tasks.length > 0) {
       // setting the wrapping box of the task component to a basis of 246px feels hacky,
       // but gets the area to be the same 453px height (or very close) as the subject area
       // and keeps the task nav buttons at the the bottom area
@@ -51,11 +49,11 @@ class Tasks extends React.Component {
       // but works for now
       return (
         <Box as='form' gap='small' justify='between' fill>
-          {tasks.map((task) => {
+          {step.tasks.map((task) => {
             // classifications.addAnnotation(task, value) retrieves any existing task annotation from the store
             // or creates a new one if one doesn't exist.
             // The name is a bit confusing.
-            const annotation = addAnnotation(task)
+            const annotation = classification.addAnnotation(task)
             task.setAnnotation(annotation)
             const TaskComponent = observer(taskRegistry.get(task.type).TaskComponent)
             if (annotation && TaskComponent) {
@@ -74,7 +72,7 @@ class Tasks extends React.Component {
 
             return (<Paragraph>Task component could not be rendered.</Paragraph>)
           })}
-          {isThereTaskHelp && <TaskHelp tasks={tasks} />}
+          {isThereTaskHelp && <TaskHelp tasks={step.tasks} />}
           <TaskNavButtons disabled={!ready || !step.isComplete} />
         </Box>
       )
@@ -92,15 +90,13 @@ class Tasks extends React.Component {
 Tasks.propTypes = {
   isThereTaskHelp: PropTypes.bool,
   loadingState: PropTypes.oneOf(asyncStates.values),
-  ready: PropTypes.bool,
-  tasks: PropTypes.arrayOf(PropTypes.object)
+  ready: PropTypes.bool
 }
 
 Tasks.defaultProps = {
   isThereTaskHelp: false,
   loadingState: asyncStates.initialized,
-  ready: false,
-  tasks: []
+  ready: false
 }
 
 @inject(storeMapper)
--- a/packages/lib-classifier/src/store/ClassificationStore.spec.js
+++ b/packages/lib-classifier/src/store/ClassificationStore.spec.js
@@ -246,8 +246,14 @@ describe('Model > ClassificationStore', function () {
         helpers.isFeedbackActive.restore()
       })
 
-      // Why is this test here?
-      // The observer is in the feedback store
+      /* 
+        Why is this test here?
+        The observer is in the root store.
+        I'm not sure why these are here either. (JOD)
+        The RootStore tests could be updated to test that feedback
+        is added to classifications on classificationComplete.
+        The tests for invalid feedback would have to be moved too.
+      */
       it('should update feedback', function () {
         expect(feedback.update.withArgs(singleChoiceAnnotationSnapshot)).to.have.been.calledOnce()
       })
--- a/packages/lib-classifier/src/store/ClassificationStore.spec.js
+++ b/packages/lib-classifier/src/store/ClassificationStore.spec.js
@@ -8,42 +8,46 @@ import {
   ProjectFactory,
   SingleChoiceAnnotationFactory,
   SingleChoiceTaskFactory,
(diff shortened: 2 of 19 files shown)


eatyourgreens:
Testing this is throwing up a bunch of infinite loops in the mock Root Stores used in the UPP tests. I'm looking into that to see if I can figure out what's going on.

eatyourgreens:
UPP is in a loop because it constantly requests a workflow but receives nothing.

eatyourgreens:
Tests are fixed now so I think this is ready for review.

I've also moved the feedback onAction listener up to the root in order to mute some of the onAction warnings.

shaunanoordin:
## Quick Review

Functionally, this PR looks legit - the changes to the code correctly ensures that annotations are made in the correct order, _for the 2nd Classification onwards._ (The bug does not affect the 1st Classification submitted. See #1521 for deets.)

e.g. for a Workflow with 3 Tasks, the annotation values submitted to Panoptes _on this branch_ are in the form of `[ { task: 'T0', ... }, { task: 'T1', ... }, { task: 'T2', ... } ]`, which is expected.

This is instead of the form of `[ { task: 'T2', ... }, { task: 'T0', ... }, { task: 'T1', ... } ]`, which is what's going on in `master` (6a913857ace00b1f1abb83678713d66397395dac)

This PR gets a 👍 from a functional point of view, though I'm going to take my time to grok the actual code changes.

eatyourgreens:
I'm going to try changing this to run `onSubjectReady` too. I think that would make it more clear that we're resetting the workflow etc. after a new subject loads.

eatyourgreens:
I've updated this to reset the workflow, create a new classification and reset feedback on subject ready. I've also added tests that check the store state after `onSubjectReady` runs.

Mocking the root store for the tests was a lot harder than I expected. There was a lot of reading through store code to figure what needed to be set and which actions should be called to set up the initial store state. It turns out that the root store can't be created directly from a snapshot. You must use `ResourceStore.setResource` and `ResourceStore.setActive` to set up workflows and subjects. That could be improved, maybe as part of #1554.

eatyourgreens:
Is there a project where we can test this with feedback enabled?

eatyourgreens:
The classification store tests for 'with valid feedback > classification metadata' are throwing MST warnings.
```
Error: [mobx-state-tree] You are trying to read or write to an object that is no
longer part of a state tree. (Object type: 'ClassificationMetadata',
Path upon death: '/classifications/resources/ck7lzu84s0066lktw7beca3gn/metadata',
Subpath: 'feedback', Action: '').
Either detach nodes first, or don't use objects after removing / replacing
them in the tree.
```

I'm not seeing any warnings in the browser so I'm going to put that down to something being not quite right in the mocks for those tests. I can't immediately see what it is.

packages/lib-classifier/src/store/ClassificationStore.spec.js
```diff
@@ -245,7 +255,7 @@ describe('Model > ClassificationStore', function () {
       // Why is this test here?
       // The observer is in the feedback store
```
eatyourgreens: I think I’ve moved this observer in this PR.

packages/lib-classifier/src/store/ClassificationStore.spec.js
```diff
@@ -111,8 +117,10 @@ describe('Model > ClassificationStore', function () {
         workflowSteps: {},
         userProjectPreferences: {}
       })
-      rootStore.subjects.setResource(subjectsStub[0])
-      rootStore.subjects.setActive(subjectsStub[0].id)
+      rootStore.subjectViewer.onSubjectReady()
```
srallen: Same comment about calling `onSubjectReady` here.

packages/lib-classifier/src/store/ClassificationStore.spec.js
srallen: Non-blocking comment: Yep. Will these specs move to the `RootStore` spec file? At the very least the comment about the observer can be removed.

packages/lib-classifier/src/store/RootStore.js
```diff
@@ -33,6 +34,49 @@ const RootStore = types
...
+        })
+      }, { name: 'Root Store Classification Observer autorun' })
+      addDisposer(self, classificationDisposer)
+    }
+
+    function createSubjectObserver () {
+      const subjectDisposer = autorun(() => {
+        onAction(self, (call) => {
+          if (call.name === 'onSubjectReady') {
+            onSubjectReady()
```
srallen: Nice. This is easier to read and clearer on what is triggering the reset.

packages/lib-classifier/src/store/ClassificationStore.spec.js
```diff
@@ -66,8 +70,10 @@ describe('Model > ClassificationStore', function () {
...
```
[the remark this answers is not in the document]
eatyourgreens: This broke the tests when I ran it inside `setupStores`, after the active subject is set there, so I moved it out. I'll take another look.

packages/lib-classifier/src/store/ClassificationStore.spec.js
eatyourgreens: This should move up to the root store tests. So much is being stubbed out here that I find it hard to follow what's really being tested. When these tests are moved, I'd suggest rewriting them to remove the stubs and test expected feedback state after a classification completes instead.

packages/lib-classifier/src/store/ClassificationStore.spec.js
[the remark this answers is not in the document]
eatyourgreens: I remember. This is here to work around a problem with the feedback tests. I've rewritten the setup function, and added a comment about the workaround.

eatyourgreens:
I think the drawing task crash might be happening when the subject renders `onReady` but before the new classification has been created. So it's trying to render annotations from the submitted classification.

eatyourgreens:
>I think this can be fixed by checking for the SubjectViewerStore's loading state in InteractionLayerContainer to be success before rendering.

That worked, thanks. I changed `enableInteractionLayer` to check the loading state too.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
```diff
@@ -41,21 +39,21 @@ class Tasks extends React.Component {
   }
 
   [asyncStates.success] () {
-    const { addAnnotation, classification, isThereTaskHelp, subjectReadyState, step, tasks, theme } = this.props
+    const { classification, isThereTaskHelp, subjectReadyState, step, theme } = this.props
     const ready = subjectReadyState === asyncStates.success
-    if (classification && tasks.length > 0) {
+    if (classification && step.tasks.length > 0) {
```
shaunanoordin: [minor] there's no instance where `step` can be undefined, right?

packages/lib-classifier/src/store/RootStore.js
```diff
@@ -33,6 +34,49 @@ const RootStore = types
   })
 
   .actions(self => {
+    // Private methods
+    function onSubjectReady () {
+      const { classifications, feedback, projects, subjects, workflows, workflowSteps } = self
+      const subject = tryReference(() => subjects?.active)
```
shaunanoordin: [aside] Wait, we have `?.` existence checkers?? This is cool, I didn't realise we had this - where's it coming from? Babel? ES2020?

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
eatyourgreens: I don't think so. I _think_ the model always has a `tasks` property, which defaults to `[]`.

eatyourgreens:
>(If I'm understanding the cause of issue #1521 is that there's some sort of race condition where the indvidualistic sub-stores are basically performing their cleanup operations regardless of what the other sub-stores are doing, so e.g. annotations stores are resetting before the task stores, or some sort of similar madness.)

I don't fully understand it, but I think this is the case. We don't have any control over the order in which the individual subject observers run in the workflow steps, classifications and feedback models. My idea was to consolidate them into a single function and run them in a prescribed order.

packages/lib-classifier/src/store/RootStore.js
srallen: [Optional chaining operator](https://github.com/tc39/proposal-optional-chaining)! It's stage 4 now, so it's part of the default babel support and should be coming to the browser soon!

PR #1556, opened 2020-03-05
Subject Group Viewer

shaunanoordin:
## PR Overview

Package: `lib-classifier`

This experimental PR attempts to add a viewer that lets users view "clusters of Subjects", or a "Subjects consisting of child Subjects". This feature is being created for the SERVOS project.

Design: https://projects.invisionapp.com/share/UAN6AP0JTBC#/screens/407090997_survos_1

- Clusters of Subjects do not exist in any form in the backend, they're currently theoretical constructs only.
- On the Classifier, clusters-of-subjects will be displayed as a grid (with pre-defined width and height), with each cell of the grid representing a single Subject.
  - Each child Subject, for this PR, will be a single image Subject ; though eventually, we might want to support video Subjects, etc.

### Dev Notes

**Workflow Configuration**

- Q: how does the front-end Classifier know when to use the Cluster-of-Subjects Viewer? (e.g. instead of a Single Image Viewer)
- A: we'll specify `configuration: { subject_viewer: 'clusterOfSubjects' }`. _This has been configured for staging workflow 3402._

- Q: how does the front-end Classifier know the size of the grid?
- A: we'll specify `configuration: { clusterGridWidth: '3', clusterGridHeight: '3' }` _Oh crap this hasn't been setup for workflow 3402 yet._
- We _might_ also need to specify `configuration: { clusterCellWidth: '200', clusterCellHeight: '200' }`

- Q: `ClusterOfSubjects` or `ClusterSubject`?
- A: I'm a bit pedantic about descriptive names so I'm defaulting to the former, wordier version, unless someone stops me... IF YOU CAN STOP ME.

**Cluster of Subjects Subjects**

- ⚠️ These types of data structures are entirely theoretical for the moment, so during the early stages of testing the UI rendering/displaying (and not any useful classification functionality), we'll use _Subjects with multiple images,_ instead of Subjects with multiple child images.

**Testing**

For testing purposes, Staging project [darkeshard/survos-testing-2020](http://localhost:8080/?env=staging&project=1900) has been created (Project ID 1900, Workflow ID 3402, Subject Set ID 4776.) Assume this project is temporary and will be obliterated as soon as testing for this PR is complete, or the requirements change.

**Tasks for Clusters of Subjects**

- Q: what kind of Tasks are we building the Subject Viewer to support?
- A: uh... a Task that allows users to select any number of items (or cells, with each cell representing a Subject) in the grid. This might be a modification of the MultiAnswerTask or its own separate custom Task, I haven't thought about it.

- Q: will the Cluster of Subjects viewer support drawing tasks?
- A: uuuhhh...

### Status

Currently VERY experimental and exploratory. Expect a 50% chance that this implementation will be completely overwritten by another iteration of ideas or whatnot.

Tagging @eatyourgreens since we'll be working together on this.

⚠️ WIP DO NOT MERGE DO NOT MERGE


eatyourgreens:
Grouped subjects (I'm leaning towards that name since I discovered that the cluster and the group are different things*) will have their own model in the state tree, so the code could reflect on the model type in order to invoke the viewer. Or we could do what the drawing tools do and have the viewer component as a property of the model: `subject.ViewerComponent`.

Come to think of it, we could do that we other subject types: `SingleImageSubject`, `MultiImageSubject`, `AudioSubject`, `VideoSubject`. All MST requires is a unique snapshot for each type.

*If I've got it right, a grouped subject is a subset of subjects drawn randomly from a pool of subjects. The subject cluster is the entire pool.

eatyourgreens:
Selecting/deselecting individual subjects is going to require a new Task model, similar to the drawing task model. In this case, our annotation will be an array of selected subject IDs, and the viewer needs access to the annotation (or current task value) in order to show individual images as selected or not.

eatyourgreens:
I suspect I'll have to make a `SingleImageSubject` model for this since the grouped subject model can only be composed of single image subjects:
```js
types.model('GroupedSubject', {
  subjects: types.array(SingleImageSubject)
})
```
where `SingleImageSubject` is a refinement of Subject that places constraints on the locations array.

camallen:
> Grouped subjects (I'm leaning towards that name since I discovered that the cluster and the group are different things*)

FWIW in the backend I'm thinking about naming as 'SubjectGroup', to fit the subject model as is. In my head the relation will be a join model `SubjectGroup.has_many :subjects` `Subject.has_many :subject_groups` etc.

eatyourgreens:
I went with `SubjectGroup` in #1557, for consistency. At the moment, it consists of a resource ID and an array of subject resources.

shaunanoordin:
`ClusterOfSubjectsViewer` has been rebranded to `SubjectGroupViewer`.

I haven't hooked the `SubjectGroupViewer` to the new model in #1555 yet, but I'll be able to get a basic view using `YourBasicSubject.location` for now

eatyourgreens:
Shall we use the storybook to wireframe/prototype this?

shaunanoordin:
> Shall we use the storybook to wireframe/prototype this?

Can do, but I'll need to get acquainted with coding for Storybook first. (I'm just so used to building things straight into the classifier with a test project on the backend.)

eatyourgreens:
You can write HTML straight into a story, to try out a rough idea, or mock up a model then pass it into a React component and render that without having to worry about the API.

eatyourgreens:
Here's a small component from the project app, that I developed using the storybook because the API wasn't completely set up when I first started it.

https://github.com/zooniverse/front-end-monorepo/blob/58aeb51c1e7a27020589904a080f0c1595bfd0a9/packages/app-project/src/screens/ClassifyPage/components/YourStats/components/DailyClassificationsChart/DailyClassificationsChart.stories.js#L6-L30

shaunanoordin:
## PR Update

The SubjectGroupViewer now allows panning & zooming as designed!

Screen Shot 2020-03-13 at 19 09 37

- Users can pan all N subject images simultaneously by dragging with a mouse.
- Users can pan all N subject images simultaneously with keyboard arrow keys
  - ⚠️ getting focus on the Subject Viewer so you can use the keyboard is still a small faff.
  - ⚠️ Pan left and pan right seem to be reversed for the keyboard arrow keys for some reason. This might be a separate PR to fix.
- Users can zoom in/out on all N subject images simultaneously by using...
  - keyboard +/- keys
  - mouse wheel
  - the zoom in/out buttons on the Subject Viewer controls (the vertical strip in the middle of the page)
- Users can reset the pan and zoom of all N subject images simultaneously by pressing the Reset button on the Subject Viewer controls.

**Buggy Bugs**

Screen Shot 2020-03-13 at 19 09 48

- The masking/clipping isn't working properly, so the images will "spill over" into their neighbouring cells... but ONLY for the left & top neighbours, not the bottom & right neighbours. Wait, WTF?
- **rotation** rotates the whole grid, not each individual image cell.

**Next On The ToDo**

- Create a new Task for selecting image cells
  - via mouse clicks
  - via keyboard
  - requires its own Task model (reminder to self: check TESS LCV)

shaunanoordin:
HA! I finally solved the problem of the masking/clipping wasn't working properly. Apparently-

```
<DraggableImage clipPath="url('#my-clip-path')" />
```

-didn't work, but-

```
<g clipPath="url('#my-clip-path')">
  <DraggableImage />
</g>
```

-does. Which is very... what? What? The `clip-path` property should applicable to every SVG element, not just the `<g>` elements. Unless I'm seriously misunderstanding the order of clipping and transforming (e.g. if the clipping occurs before the image is translated) or something to that effect...

Anyway, the major bug for this PR is solved, now on to tests.

shaunanoordin:
OK, conflicts solved, now back to those tests.

eatyourgreens:
Looks good to me. I think tests are broken because you can't use refs with shallow mounting. One thought: maybe disable `enableRotation` for this viewer?

shaunanoordin:
## PR Update

It's done! It's finally done! Here are some updates to the SubjectGroupViewer:

- Readme added
- Component broken down further (see: SGVGridCell)
  - Proptypes and default proptypes updated... though I may need second pass to ensure I didn't forget to catch invalid values.
- Unnecessary bits - rotation and InteractionLayer - removed
- Tests added for both SubjectGroupViewer and SGVGridCell

### Status

Ready for review

shaunanoordin:
## PR Update

Correction: not all tests are done! SubjectGroupViewerContainer.js is all sorts of borked - sorry, I was sure I wrote tests for this but apparently it's not part of this PR. Fixing that now -

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/README.md
```diff
@@ -0,0 +1,61 @@
...
+  cell_width: 1200,
+  cell_height: 1000,
+  cell_style: {
+    stroke: '#fff',
+    strokeWidth: '4',
+    fill: '#000'
+  },
+  grid_columns: 4,
+  grid_rows: 3
+}
```
eatyourgreens: Should the subject viewer config options be grouped into a nested object? @srallen might have suggestions because she's been working on how to specify config options for the light curve viewer.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  dragMove (event, difference) {
+    this.onDrag && this.onDrag(event, difference)
+  }
+
+  async preload () {
+    const { subject } = this.props
+    if (subject && subject.locations) {
+      // TODO: Validate for allowed image media mime types
```
eatyourgreens: You can remove this comment. Validation is included in the `Subject` model now.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/README.md
srallen: The options that the scatter plot supports is per subject with the expectation that a chart options object is defined in the subject JSON for the plot type (and there's individual data series option object as well). The reason I went this direction is because working with actual data, there's probably a greater need for granular control like this than other media types. For the plots, I've been using `chartOptions` in the JSON. 

If you know that these settings should always be the same for the viewer type for every subject, then I agree this should probably be an object in the workflow configuration, perhaps, `subject_viewer_settings` or something like that, so that it is clear that what these are for.

eatyourgreens: I _think_ we're okay having these on the workflow. At least, to my mind, it would be a very confusing experience if the look and layout of the subject viewer could change from one subject to the next while I'm classifying.

If it turns out that SuRVoS would like to re-use the same workflow with different clusters of subjects, then we should consider adding config at the subject set level but I don't think we need that right now.

shaunanoordin:
## PR Update

This PR has been updated with more robust tests for GroupSubjectViewerContainer, and a few additional tests for the SGVGridCell.

Much thanks to @eatyourgreens on this one; I was originally basing my test code on an outdated version of SingleImageViewerContainer, which caused havoc with the async calls. [Correct version here.](https://github.com/zooniverse/front-end-monorepo/blob/e928339e28d31ec1f0ac3e3ea54bc8a2cfbd4138/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.spec.js#L56-L86)

One thing to note:
- The SubjectGroupViewerContainer and SingleImageViewerContainer differ greatly in _where_ the image should be rendered. The SingleImageViewerContainer can test when an SVG image is rendered, but for the SGV system, that duty is passed down to the SGVGridCell component. The tests in SubjectGroupViewerContainer mostly ensures the proper values are being passed down to the children components, and the children components are in charge of making sure those values (image URLs) are rendered.

shaunanoordin:
Yaaasss all checks passed!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -0,0 +1,301 @@
...
+      
+      const imageUrls = subject.locations.map(obj => Object.values(obj)[0])
+      const images = await Promise.all(
+        imageUrls.map(url => this.fetchImage(url))
+      )
```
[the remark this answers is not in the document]
eatyourgreens: Oh wait, `onLoad` wraps this in a `try… catch` so that should be fine.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -0,0 +1,301 @@
...
+      
+      this.setState({ images })
+      return images
+    }
+    return {}
```
eatyourgreens: I think this should return an empty array, since `images` is an array.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -0,0 +1,301 @@
...
+      clientWidth,
+      naturalHeight,
+      naturalWidth
+    }
+  }
+
+  async onLoad () {
+    const { onError, onReady } = this.props
+    try {
+      await this.preload()
```
eatyourgreens: `preload()` returns the `images` array but you don't do anything with them here.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -0,0 +1,301 @@
...
+        imageUrls.map(url => this.fetchImage(url))
+      )
+      
+      this.setState({ images })
+      return images
+    }
+    return {}
+  }
+
+  async getImageSize () {
```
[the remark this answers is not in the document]
eatyourgreens: Also, I don't see anything inside this function which would make it asynchronous.

eatyourgreens:
I'd suggest changing the storybook to use some representative SuRVoS subject images.

For better performance, I'd also suggest updating the individual grid cells to use our thumbnail service. I think we might have a component that does that for you.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
```diff
@@ -0,0 +1,128 @@
...
+        width={cellWidth}
+        height={cellHeight}
+        dragMove={dragMove}
+      />
+      <g clipPath={`url(#${clipPathID})`}>
+        <DraggableImage
+          dragMove={dragMove}
+          height={imageHeight}
+          width={imageWidth}
+          xlinkHref={image.src}
```
eatyourgreens: Depending on the size of the original images, should we pass them through the thumbnail service?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewer.js
```diff
@@ -0,0 +1,126 @@
...
+    
+    width,
+    height,
+    
+    panX,
+    panY,
+    zoom,
+  } = props
+
+  const transformLayer = createRef()
```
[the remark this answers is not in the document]
shaunanoordin: It definitely should be! I should change the same line in the `SingleImageViewer.js` code as well, probably in a separate PR.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -0,0 +1,301 @@
...
+  componentDidMount () {
+    this.onLoad()
+
+    // Listen for pan and zoom actions outside of this component.
+    // i.e. zoom in/out actions from the iamge controls. 
+    this.props.setOnPan(this.onPanViaExternalControls.bind(this))
+    this.props.setOnZoom(this.onZoomViaExternalControls.bind(this))
+    
+    this.scrollContainer.current
+    && this.scrollContainer.current.addEventListener('wheel', preventDefault)
```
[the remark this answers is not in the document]
shaunanoordin: 👍

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -0,0 +1,301 @@
...
+  }
+  
+  componentWillUmount () {
+
+    this.setOnPan(() => true)
+    this.setOnZoom(() => true)
+    
+    this.scrollContainer.current
+    && this.scrollContainer.current.removeEventListener('wheel', preventDefault)
```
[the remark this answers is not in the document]
shaunanoordin: 👍

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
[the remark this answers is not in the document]
shaunanoordin: This should be completely renamed and updated, since the _"get the sizes of the visual elements so we can calculate how we can display them"_ function has mutated greatly from the comparable SingleImageViewer code. I originally tried to keep the SubjectGroupViewer and SingleImageViewer as similar as possible, but I think the dev for the SGV has sailed waaay past that point already.

To be precise: this should be named `getGridSize()` and doesn't need to be async.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -0,0 +1,301 @@
...
+      naturalWidth
+    }
+  }
+
+  async onLoad () {
+    const { onError, onReady } = this.props
+    try {
+      await this.preload()
+      const { clientHeight, clientWidth, naturalHeight, naturalWidth } = await this.getImageSize()
+      const target = { clientHeight, clientWidth, naturalHeight, naturalWidth }
```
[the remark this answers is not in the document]
shaunanoordin: In the context of the SubjectGroupViewer, the "naturalHeight" and "naturalWidth" are the WxH dimensions of the explicitly-sized Grid of Image Cells.

You're right that both terms don't make sense in the context of a grid, so I've renamed them to "gridWidth" and "gridHeight", which I hope is more understandable. (These variables are still renamed to naturalHeight/naturalWidth when passed to SubjectViewerStore.onSubjectReady() though.)

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -0,0 +1,301 @@
...
+
+    if (loadingState === asyncStates.error) {
+      return (
+        <div>Something went wrong.</div>
+      )
+    }
+
+    const svg = this.groupViewer.current
+    
+    console.log()
```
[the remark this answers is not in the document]
shaunanoordin: Uh... would you believe me if I said I was trying to print out some silence?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -0,0 +1,301 @@
...
+        || !(cellWidth > 0)
+        || !(gridColumns > 0)
+        || !(gridRows > 0)
+    ) {
+      return null
+    }
+
+    return (
+      <SVGContext.Provider value={{ svg }}>
+        <div ref={this.scrollContainer} onWheel={this.onWheel.bind(this)}>
```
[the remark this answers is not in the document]
shaunanoordin: 👍

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
shaunanoordin: 👍

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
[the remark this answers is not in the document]
shaunanoordin: Yup, I was basing the error-handling on the SingleImageViewerContainer. In case either `preload()` or `fetchImage()` throws an error, we'll get the general purpose `loadingState === asyncStates.error` state and the appropriate "Something Went Wrong" error message.

⚠️ Note: this means if ANY of the Subject's sub-Subject's images don't load properly, the WHOLE grid considers it a failure. I thought this made sense, since if I allowed a "just show everything that loaded properly" approach, it might accidentally introduce an untraceable bias when 1 or more images of the N total images was a 404 while the others displayed perfectly.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
shaunanoordin: I've noted this as a separate issue - https://github.com/zooniverse/front-end-monorepo/issues/1584

The ability to pass images to the thumbnail service is a good idea, but I think should be configurable, and default to false at the start, since some researchers may actually be providing hi-res Subject images so they can ask users to zoom in on the details.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
shaunanoordin: 👌 The images array no longer needed in `onLoad()`; I've now simplified the return for `preload()`

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
eatyourgreens: With our rough limit of ~600k per image, that gives a download of ~5.4MB per subject for a cluster of 9 images. 😨

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -0,0 +1,297 @@
...
+    const { gridRows, gridColumns, cellWidth, cellHeight } = this.props
+    
+    const gridWidth = gridColumns * cellWidth
+    const gridHeight = gridRows * cellHeight
+    
+    return {
+      clientHeight,
+      clientWidth,
+      gridHeight,
+      gridWidth,
```
eatyourgreens: I'm still a bit confused by this code. How does grid width and height differ from the client width and height in the browser window? Looking at the storybook, grid dimensions are set from props, which I think are passed down from the workflow config. I'm missing how it ties in to the original natural height and width of each of the subject images.

In the image viewer code, we want to return the rendered size of the image _and_ its natural size so that the classification records how much the image was scaled by for this volunteer. Here, I was thinking we'd want to record how much each subject image was scaled down in the grid viewer.

shaunanoordin:
## PR Update

This PR has been updated with the following:

- the Storybook component has been fixed - you can now view the SGV with kitten images at http://localhost:6006/?path=/story/subject-viewers-subjectgroupviewer--with-zoom-controls
- Multiple feedback issues & improvements have been addressed.

Side note: I realised the solution for fixing the SGV storybook component was to just delete the whole thing, and re-base the code on the latest SingleImageViewer. It just occurred to me that my baseline had drifted so far apart that a rebuild was the best solution. (It was!)

### Status

Thanks for the review @eatyourgreens , I'm going ahead and merging this with the following items on the ToDo (separate from the classifying bits for the Subject Group Viewer):

- SGV could use thumbnail service - #1584
- SGV need to examine how it records _"the natural width/height of the image file vs the client-viewed width/height of the image as displayed/resized on the browser"_ metadata - #1594
  - Currently there's a lot of confusion on what's going on here. The short explanation is that the SGV doesn't record the _natural sizes_ of all _N_ number of individual image objects, but instead records the _intended width/height of the grid vs the client-viewed width-height of the grid"_
  - Of course, if this is the _short_ explanation, this is certainly an item worth re-examining. Full deets are in the issue writeup.
- SGV might want to consolidate its grid/cell config settings in `myWorkflow.configuration` into a single `viewerConfig` object. - #1595
  - This is fairly straightforward to do; I had just assumed that we wanted our `workflow.configurations` to be flat for some reason, and purposely spread out the config vars.
- SGV storybook component should use sample Survos images instead of kitten placeholder, which while adorable, make less sense.
  - I've got the files, let me do some uploading and patch in a quick fix later.

Issue #1484, opened 2020-02-05
Classifier: SuRVoS subject cluster model


eatyourgreens:
**Package**
lib-classifier

**Describe the solution you'd like**
Add a subject cluster model. Initially for the SuRVoS project but this may have more general applications. A subject cluster is a group of subjects which can all be annotated by a single task at the same time.

>This is similar to the current setup in the Muon hunters 2.0 project. However instead of the images being pre-computed by an offline process, this Zooniverse subject selection system will return randomly sampled subjects from a pre-defined ‘cluster’ group.

eatyourgreens:
Small change to the terminology: #1557 adds a `SubjectGroup` model, where a grouped subject is a subset of subjects chosen from within a larger cluster.

PR #1557, opened 2020-03-06
Classifier: Grouped subject models

eatyourgreens:
Adds a basic model for grouped subjects, which are composed of single image subjects. A `SingleImageSubject` is a refinement of `Subject` which has exactly one image. Not 0. Not 2. Three is right out.

Package:
lib-classifier

Towards #1484.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/SingleImageSubject/SingleImageSubject.js
+++ b/packages/lib-classifier/src/store/SingleImageSubject/SingleImageSubject.js
@@ -1,0 +1,13 @@
+import { types } from 'mobx-state-tree'
+import createLocationCounts from '../../helpers/createLocationCounts'
+import Subject from '../Subject'
+
+const SingleImageSubject = types
+  .refinement('SingleImageSubject', Subject, subject => {
+    const counts = createLocationCounts(subject)
+    return counts.images === 1
+  })
+
+  
+
+export default SingleImageSubject
--- a/packages/lib-classifier/src/store/SingleImageSubject/SingleImageSubject.js
+++ b/packages/lib-classifier/src/store/SingleImageSubject/SingleImageSubject.js
@@ -5,7 +5,7 @@ import Subject from '../Subject'
 const SingleImageSubject = types
   .refinement('SingleImageSubject', Subject, subject => {
     const counts = createLocationCounts(subject)
-    return counts.images === 1
+    return subject.locations.length === 1 && counts.images === 1
   })
 
   
--- a/packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
+++ b/packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
@@ -3,6 +3,10 @@ import Resource from '../Resource'
 import subjectViewers from '../../helpers/subjectViewers'
 import SingleImageSubject from '../SingleImageSubject'
 
+function validateSubjects (subjects) {
+  return subjects.length > 1
+}
+
 const SubjectGroup = types
   .model('SubjectGroup', {
     already_seen: types.optional(types.boolean, false),
@@ -13,7 +17,7 @@ const SubjectGroup = types
     selected_at: types.maybe(types.string),
     selection_state: types.maybe(types.string),
     shouldDiscuss: types.maybe(types.frozen()),
-    subjects: types.array(SingleImageSubject),
+    subjects: types.refinement(types.array(SingleImageSubject), validateSubjects),
     user_has_finished_workflow: types.optional(types.boolean, false),
   })
 
--- a/packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
+++ b/packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
@@ -1,0 +1,12 @@
+import { types } from 'mobx-state-tree'
+import Resource from '../Resource'
+import SingleImageSubject from '../SingleImageSubject'
+
+const SubjectGroup = types
+  .model('SubjectGroup', {
+    subjects: types.array(SingleImageSubject)
+  })
+
+  
+
+export default types.compose('SubjectGroup', Resource, SubjectGroup)
--- a/packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
+++ b/packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
@@ -21,6 +21,30 @@ const SubjectGroup = types
     user_has_finished_workflow: types.optional(types.boolean, false),
   })
 
+  .views(self => ({
+    get locations () {
+      return self.subjects.map(subject => subject.locations[0])
+    },
+
+    get talkURL () {
+      const project = tryReference(() => getRoot(self).projects?.active)
+      if (project) {
+        const { origin } = window.location
+        return `${origin}/projects/${project.slug}/talk/subjects/${self.id}`
+      }
+
+      return ''
+    },
+
+    get viewer () {
+      return subjectViewers.subjectGroup
+    },
+
+    get workflow () {
+      return tryReference(() => getRoot(self).workflows?.active)
+    }
+  }))
+
   .actions(self => {
 
     function addToCollection () {
@@ -47,26 +71,6 @@ const SubjectGroup = types
       toggleFavorite
     }
   })
-
-  .views(self => ({
-    get talkURL () {
-      const project = tryReference(() => getRoot(self).projects?.active)
-      if (project) {
-        const { origin } = window.location
-        return `${origin}/projects/${project.slug}/talk/subjects/${self.id}`
-      }
-
-      return ''
-    },
-
-    get viewer () {
-      return subjectViewers.subjectGroup
-    },
-
-    get workflow () {
-      return tryReference(() => getRoot(self).workflows?.active)
-    }
-  }))
 
   
 
--- a/packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
+++ b/packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
@@ -1,11 +1,68 @@
-import { types } from 'mobx-state-tree'
+import { getRoot, tryReference, types } from 'mobx-state-tree'
 import Resource from '../Resource'
+import subjectViewers from '../../helpers/subjectViewers'
 import SingleImageSubject from '../SingleImageSubject'
 
 const SubjectGroup = types
   .model('SubjectGroup', {
-    subjects: types.array(SingleImageSubject)
+    already_seen: types.optional(types.boolean, false),
+    favorite: types.optional(types.boolean, false),
+    finished_workflow: types.optional(types.boolean, false),
+    metadata: types.frozen({}),
+    retired: types.optional(types.boolean, false),
+    selected_at: types.maybe(types.string),
+    selection_state: types.maybe(types.string),
+    shouldDiscuss: types.maybe(types.frozen()),
+    subjects: types.array(SingleImageSubject),
+    user_has_finished_workflow: types.optional(types.boolean, false),
   })
+
+  .actions(self => {
+
+    function addToCollection () {
+      const rootStore = getRoot(self)
+      rootStore.onAddToCollection(self.id)
+    }
+
+    function openInTalk (newTab = false) {
+      self.shouldDiscuss = {
+        newTab,
+        url: self.talkURL
+      }
+    }
+
+    function toggleFavorite () {
+      const rootStore = getRoot(self)
+      self.favorite = !self.favorite
+      rootStore.onToggleFavourite(self.id, self.favorite)
+    }
+
+    return {
+      addToCollection,
+      openInTalk,
+      toggleFavorite
+    }
+  })
+
+  .views(self => ({
+    get talkURL () {
+      const project = tryReference(() => getRoot(self).projects?.active)
+      if (project) {
+        const { origin } = window.location
+        return `${origin}/projects/${project.slug}/talk/subjects/${self.id}`
+      }
+
+      return ''
+    },
+
+    get viewer () {
+      return subjectViewers.subjectGroup
+    },
+
+    get workflow () {
+      return tryReference(() => getRoot(self).workflows?.active)
+    }
+  }))
 
   
 
--- a/packages/lib-classifier/src/store/SubjectStore.js
+++ b/packages/lib-classifier/src/store/SubjectStore.js
@@ -10,8 +10,19 @@ import SubjectGroup from './SubjectGroup'
 
 const MINIMUM_QUEUE_SIZE = 3
 
-const subjectModels = [ SingleImageSubject, Subject, SubjectGroup ]
-const subjectReferences = subjectModels.map(model => types.safeReference(model))
+/*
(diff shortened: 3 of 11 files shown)


eatyourgreens:
`types.refinement` is very handy. We should make more use of it.

packages/lib-classifier/src/store/SingleImageSubject/SingleImageSubject.js
```diff
@@ -0,0 +1,13 @@
+import { types } from 'mobx-state-tree'
+import createLocationCounts from '../../helpers/createLocationCounts'
+import Subject from '../Subject'
+
+const SingleImageSubject = types
+  .refinement('SingleImageSubject', Subject, subject => {
+    const counts = createLocationCounts(subject)
+    return counts.images === 1
```
eatyourgreens: Will this pass if the subject has one image and one audio file, say?

eatyourgreens:
Tests pass locally but fail in Travis with module loading errors. I have no idea why. 🤔

shaunanoordin:
## Quick Review

1. Changes look good in terms of functionality - testing on basic single-image workflows show no issues (...though this may be because the new SingleImageSubject store and GroupSubject store is never triggered/called by the actual Classifier UI yet, I think?)
2. I need to get a bit more clarity on what I think I'm reading here...
  - the classic Subject store is still there, it's just been moved on folder down. 👌 
  - the new SingleImageSubject store is used _only_ by the GroupSubject store.
  - ❓ the new GroupSubject store is _not being used yet,_ but we'll hook it up to the Classifier UI somehow (something like... if `workflow.configuration.subject_viewer == 'grouped_subjects'`... then `Classifier.store.subject = new GroupSubject()` ? This part is still foggy to me, though I get the general shape of the solution)
3. I need to read up on MST's `.refinement()` feature, which I suspect is... a way of checking and/or enforcing a certain rule, e.g. "a SingleImageSubject store instance must have exactly 1 item in its .location array"
4. Argh! There's a conflict with Subject.js that needs to be resolved

shaunanoordin:
EDIT: I'm just starting to read https://github.com/zooniverse/front-end-monorepo/pull/1556#issuecomment-595740611 onwards, which addresses a few of my question marks.

eatyourgreens:
It turns out the Subject model has `subject.viewer` which tells the code which viewer to use. I've got a test for it here, for the single image viewer.

Something we could do, as part of the Survos work, is document the general API for subject viewers.

eatyourgreens:
#1555 is probably going to conflict with this so I'm wondering whether to let that merge first, then clean up the specs.

eatyourgreens:
Should `SubjectGroup` have metadata, and a `viewer` view?

shaunanoordin:
> #1555 is probably going to conflict with this so I'm wondering whether to let that merge first, then clean up the specs.

I say let #1555 get merged first because it's more fundamental, and this PR is more exploratory.

> Should SubjectGroup have metadata...

Sounds reasonable.

> ...and a viewer view?

On one hand, this sounds like a legit idea. On the other hand, `src/store/Subject` already has the ["Choose The Form Of The Subject Viewer"](https://www.youtube.com/watch?v=3Ctgn7kKYHo) logic in `get viewer ()`, so if we start giving each Subject type its own `.viewer` view we need to be consistent or else the logic will be all over the place.

shaunanoordin:
Currently in #1556, I'm modifying `src/store/Subject.viewer()` so that...

```
} else if (configuration.subject_viewer === 'subjectGroup') {
  viewer = subjectViewers.subjectGroup
```

eatyourgreens:
I'm thinking `subjectGroup.viewer` should return `subjectViewers.subjectGroup`. I'm not sure how much overlap there's supposed to be between subject groups and subjects. They're both things that can be classified but also collected, discussed, favourited etc.?

eatyourgreens:
The MST docs aren't great, I'm afraid, but `types.refinement` is defined (very briefly) here: https://mobx-state-tree.js.org/API/#refinement

`SingleImageSubject` is a `Subject` that has one location, which is an image URL. I could cretainly add it to the subjects store. I had some random thoughts about that here: https://github.com/zooniverse/front-end-monorepo/pull/1556#issuecomment-595740611

eatyourgreens:
>My plan for API implementation (open to change) is to return 1/N subjects, each one will be a special subject with an ordered list of images that represent the underlying subjects used to create the 'Group'. The SubjectGroup will be a distinct read only resource in the API to record the group data and will link to a Subject and will probably serialize some of this group data with the Subject that has a group.

Sounds like it should be ok for me to add `SubjectGroup` as an allowed type in the classifier subject queue, if I'm reading this correctly. Should I add subject metadata to the model?

camallen:
> The MST docs aren't great, I'm afraid, but types.refinement is defined (very briefly) here: https://mobx-state-tree.js.org/API/#refinement
> SingleImageSubject is a Subject that has one location, which is an image URL

The docs are very sparse, does `SingleImageSubject` have the same model attributes as Subject? If so that's great.

> Should I add subject metadata to the model?

I assume if the model metadata will be used for say the classifier buttons on subject metadata then it might be required. That said i'm really not sure what the Subject looks like in the API yet. 

What Subject attributes do you need to wire up the UI correctly for parity with existing features?

eatyourgreens:
From a quick test run, I'm getting `metadata.intervention` is undefined when trying to create a classification from a `SubjectGroup`. I would guess `metadata` at a minimum, to avoid errors but more likely most of the subject state attributes listed here:
https://github.com/zooniverse/front-end-monorepo/blob/58aeb51c1e7a27020589904a080f0c1595bfd0a9/packages/lib-classifier/src/store/Subject.js#L9-L21

eatyourgreens:
>The docs are very sparse, does SingleImageSubject have the same model attributes as Subject? If so that's great.

Yes, it's a `Subject` with an additional constraint on `locations`. I can load I Fancy Cats without any problems if I change `Subject` to `SingleImageSubject` in the store code. PH-TESS throws a type error, as expected:
```
Error: [mobx-state-tree] Error while converting 
`{"id":"81370",
"metadata":{"#ID":"441813045","png":"43d3eecd-d436-4b2b-865a-77588d031442.png","txt":"7e089440-5a8e-423f-b6f5-dc63af4eedd2.txt",
"Radius":"0.910 Solar Radii",
"Sector":"14",
"TIC ID":"441813045",
"#Target":"ulated_data/ALL",
"Magnitude":"9.770 mag",
"Temperature"......ormal"}` to `(SingleImageSubject)`:
```

eatyourgreens:
Weirdly, the subject viewer breaks for I Fancy Cats when I add `SubjectGroup` to the list of allowed subject types. That should not happen, since it has a completely different snapshot to `Subject`, so I'm going to look into that.

eatyourgreens:
I've added `SubjectGroup` as a subject type in the subjects store, but it gives invalid reference errors when I try to add one to the store. 🤔

eatyourgreens:
I've added a bunch of attributes, views and actions to `SubjectGroup` that I've copied across from the `Subject` model. They should act like subjects now, except that they have `subject.subjects` instead of `subject.locations`.

camallen:
> I've added a bunch of attributes, views and actions to `SubjectGroup` that I've copied across from the `Subject` model. They should act like subjects now, except that they have `subject.subjects` instead of `subject.locations`.

Will talk and other UI code need the subject.locations? I was considering keeping this as is to list the media URLs in a known order. I could also support the `subject.subjects` attribute at serializer as well.

eatyourgreens:
>Will talk and other UI code need the subject.locations? I was considering keeping this as is to list the media URLs in a known order. I could also support the subject.subjects attribute at serializer as well.

That's an excellent point. I can map the individual subject locations to the group locations, no problem. Anything that previews subjects as thumbnails (Recents, collections, favourites) will use `subject.locations` too.

eatyourgreens:
Tests are failing because the model snapshots are not unique ie. I can create a `SingleImageSubject` from a `SubjectGroup` snapshot.

eatyourgreens:
I've updated this to add subject groups to the subjects store, with tests, so it could do with another review.

packages/lib-classifier/src/store/SubjectStore.js
```diff
@@ -1,13 +1,30 @@
 import asyncStates from '@zooniverse/async-states'
 import { autorun } from 'mobx'
-import { addDisposer, addMiddleware, flow, getRoot, isValidReference, onPatch, tryReference, types } from 'mobx-state-tree'
+import { addDisposer, addMiddleware, flow, getRoot, getType, isValidReference, onPatch, tryReference, types } from 'mobx-state-tree'
 import { getBearerToken } from './utils'
 import { filterByLabel, filters } from '../components/Classifier/components/MetaTools/components/Metadata/components/MetadataModal'
 import ResourceStore from './ResourceStore'
 import Subject from './Subject'
+import SingleImageSubject from './SingleImageSubject'
+import SubjectGroup from './SubjectGroup'
```
eatyourgreens: The subject models could be further reorganised so that these lines become
```js
import { Subject, SingleImageSubject, SubjectGroup } from './subjects'
```

packages/lib-classifier/src/store/SubjectStore.js
```diff
@@ -5,9 +5,26 @@ import { getBearerToken } from './utils'
...
+  see https://github.com/mobxjs/mobx-state-tree/issues/514
+  for advice about using references with types.union.
+*/
+const SingleSubject = types.union(SingleImageSubject, Subject)
+function subjectDispatcher (snapshot) {
+  if (snapshot.subjects) {
+    return SubjectGroup
+  }
+  return SingleSubject
+}
```
shaunanoordin: [notes to self]

I've been trying to wrap my head around this (unions and dispatchers aren't my forte, alas), so I'm going to write out what I can discern from this code:

- the "dispatcher" is just a function whose sole job is to determine what _type of_ Subject is being seen in a particular instance of a union-ed data model.
- Usually the types.union([list_of_types]) data model would be smart enough (?) to figure out the _type_ of the instance from the shape of its data, but the dispatcher is there to make the selection explicit. By default it selects the first matching type from the 'list_of_types'.
- So basically what we have here is...
```
SubjectStore.active('s data model) = (snapshot) => {
  if snapshot has .subjects, return SubjectGroup
  else,
    if snapshot has only 1 image (as determined by .refinement), return SingleImageSubject
    else, return generic Subject 
}
```

If that's the case, then this dispatcher will become an important logical "selection crossroads" for future development, when potential new Subject types are added.

It might be worth noting/broadcasting this info, because otherwise future devs might start encoding the "selection logic" in individual data model definitions, so you might need to dig through 5 layers of specific sub-types+unions+dispatchers to determine why Subject.active is currently of the type SingleImageWithAnimalsInAfricaWithBlueSkiesButNoAardvarks. (Great for compartmentalisation the code, headachey for instantly grokking what's happening.)

Similar notes can be applied for `Subject`'s `get viewer ()` which has a "selection crossroad" and `SubjectGroup`'s `get viewer ()` which overrides it.

packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
```diff
@@ -0,0 +1,77 @@
...
+        const { origin } = window.location
+        return `${origin}/projects/${project.slug}/talk/subjects/${self.id}`
+      }
+
+      return ''
+    },
+
+    get viewer () {
+      return subjectViewers.subjectGroup
+    },
```
shaunanoordin: [note to self]

I'm making a note of this since it'll probably minorly affect #1555

I'm still curious if the "selection logic" (in this case, determining which SubjectViewer to use) is best served by placing it in individual Subject sub-types' `get viewer()` code, and overriding the generic Subject type's `get viewer()` "selection crossroads".

I mean, pro: it _definitely_ results in more compartmentalised code that fits with the monorepo philosophy, but con: it _might_ make the selection logic harder to grok if the logic is split across several places.

(My feelings for/against this is exactly at the 50% mark, so I'm quite ambivalent.)

srallen: Agreed with @shaunanoordin here. Having logic in different places for which viewer to render is harder to follow. Is there a reason why this isn't following the same logic as other customized viewers (LCV, VSLCV) in using a workflow configuration setting to select which subject viewer to render?

packages/lib-classifier/src/store/SubjectStore.spec.js
```diff
@@ -121,6 +123,43 @@ describe('Model > SubjectStore', function () {
...
+        }
+        subjectGroups.push(subjectGroup)
+      }
+      subjects = mockSubjectStore(subjectGroups)
+    })
+
+    it('should be valid subjects', function () {
+      expect(subjects.active).to.exist()
+      expect(subjects.active.id).to.equal('0')
+      expect(subjects.active.subjects).to.have.lengthOf(25)
```
srallen: `subjects.active.subjects` I think could be confusing. What about `subjects.active.groupedSubjects` instead? We could also rename `subjects` to `subjectsQueue`

srallen:
There's hints of a possible larger refactor which I'm guessing is modeling each subject media type. I'd prefer an open RFC discussion about this rather than have it happen via PR without a clear explanation and communication to the team about implications and changes. @shaunanoordin has noted the some of implications very well.

packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
eatyourgreens: I thought about copying over the `viewer` function from the `Subject` model. We can still do that.

packages/lib-classifier/src/store/SubjectStore.js
eatyourgreens: Yeah, a dispatcher is a function that takes a snapshot and returns a type for that snapshot. It’s used when a type can’t be determined from a snapshot ie. snapshots aren’t unique to models.

packages/lib-classifier/src/store/SubjectStore.spec.js
eatyourgreens: The naming here will ultimately depend on what this attribute is called in Panoptes. I think this model is going to change once we know what the API response looks like.

eatyourgreens:
> There's hints of a possible larger refactor which I'm guessing is modeling each subject media type.

I'm not sure you can model subjects that way, at least without restricting what project owners are allowed to upload as subjects. Having said that, audio and video subjects might work as refinements of `Subject`, where the refinement validates `subject.locations` for audio and video URLs.

packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
eatyourgreens: If we did copy over the `viewer` logic from `Subject`, then `groupSupject.viewer` could return viewer types other than the `GroupSubject` viewer. I'm not sure that's what we want.

srallen:
> Having said that, audio and video subjects might work as refinements of Subject, where the refinement validates subject.locations for audio and video URLs.

This is exactly the refactor I'm talking about. I'm not sure if we want to do this, and now we partially do by way of `SingleImageSubject` from this PR.

PR #1562, opened 2020-03-11
Validate subject mimetypes

srallen:

Package: lib-classifier

Describe your changes:
Adds a mime type validation helper function that I've incorporated into the `createLocationCounts` helper, but can be used on its own too.

I took the approach that if any one of the locations of a subject is invalid to return an empty object with no counts. This isn't entirely incorporated into the `viewer` getter yet; it's not validating the viewers that explicitly look for a workflow configuration, but I wanted to get feedback first before doing any more work.

I also took a slightly different approach from the [original PFE version](https://github.com/zooniverse/Panoptes-Front-End/blob/ec86d183e6d1a1cce27b63ac51e338329ce40131/app/lib/get-subject-locations.js) using regexes and the test method. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/helpers/validateMimeType/validateMimeType.spec.js
```diff
@@ -0,0 +1,46 @@
...
+  'video/mp4',
+  'video/mpeg',
+  'video/x-m4v'
+]
+
+const invalidMimeTypes = [
+  'application/xml',
+  'audio/aac',
+  'audio/wave',
+  'image/tiff'
```
eatyourgreens: Maybe add `text/html` to the list of blocked types?

eatyourgreens:
I think the JSON data for the PH-TESS subjects might be marked as `text/plain` in the subjects served by Panoptes. It shows up in Talk using the text file viewer.

eatyourgreens:
I think this will work very well with the `SingleImageSubject` model in #1557.

packages/lib-classifier/src/helpers/validateMimeType/validateMimeType.spec.js
eatyourgreens: I’m being paranoid, but could you block JS here too?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
```diff
@@ -69,7 +69,6 @@ class SingleImageViewerContainer extends React.Component {
   async preload () {
     const { subject } = this.props
     if (subject && subject.locations) {
-      // TODO: Validate for allowed image media mime types
       const imageUrl = Object.values(subject.locations[0])[0]
```
eatyourgreens: Where is this URL being validated?

packages/lib-classifier/src/helpers/createLocationCounts/createLocationCounts.js
```diff
@@ -1,14 +1,26 @@
...
 
-  const images = types
-    .filter(type => type.startsWith('image'))
-    .length
+  try {
+    const images = mimeTypes
+      .filter(mimeType => mimeType.startsWith('image') && validateMimeType(mimeType))
+      .length
+
+    const json = mimeTypes
```
eatyourgreens: You often see `.js` and `.json` sent as `text/plain` on the web. I think PH-TESS subjects use `text/plain` for JSON data.

packages/lib-classifier/src/helpers/createLocationCounts/createLocationCounts.spec.js
```diff
@@ -15,29 +15,53 @@ const mockLocations3 = [
...
+    it('should return the number of json files', function () {
+      const result4 = createLocationCounts(mockSubject(mockLocations4))
+
+      expect(result4.json).to.equal(1)
+    })
+  })
 
-    expect(result2.images).to.equal(2)
-    expect(result3.images).to.equal(2)
+  describe('when the mime types are invalid', function () {
```
eatyourgreens: Is it worth having a text for a mix of valid and invalid types?

packages/lib-classifier/src/helpers/createLocationCounts/createLocationCounts.js
srallen: I think TESS using that was due to a bug in the python client.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
srallen: `createLocationCounts` now uses the validation and if any of the locations are invalid, an empty object is returned. There's a conditional in the getter in the `Subject` store that will now evaluate as false and the viewer will return null. Then  in the `SubjectViewer` component, the return JSX component will be null, so this code will never run. I can add a test in that component to be sure.

packages/lib-classifier/src/helpers/createLocationCounts/createLocationCounts.js
srallen: I didn't touch the getter that sets the viewer to use the LCV subject viewer to use the validation yet, so this is setup for VSLCV to validate properly and we can speak to the TESS team to make sure future sectors get uploaded with the correct mime type.

packages/lib-classifier/src/helpers/createLocationCounts/createLocationCounts.spec.js
srallen: I've added a test for this.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js
srallen: I've added a couple of tests in `Subject.spec.js` for when the subject is invalid, the viewer should be null.

packages/lib-classifier/src/helpers/validateMimeType/validateMimeType.spec.js
srallen: I've added a few more invalid mime types.

srallen:
@eatyourgreens This is ready for another review. I've added some more robust tests and another helper `validateSubjectLocations` which checks the mime type and also validates the file extension on the url. I'm still not totally certain how best to integrate this with the viewers that are using workflow config directly, so I'd like to have this reviewed as is and continue adding more checks in follow PRs if that's alright.

packages/lib-classifier/src/helpers/createLocationCounts/createLocationCounts.js
eatyourgreens: Sounds good. You may find that the API client uploads `.json` as text, depending on where it's run and what type that machine/OS assigns to JSON files, but we can deal with that when it happens.

PR #1563, opened 2020-03-12
Update Grommet to 2.11

srallen:

Package: all

Describe your changes:
This is an update to Grommet so we can get the latest bug fixes. I'm using the `RadioButtonGroup` in the VSLCV and in particular, there's a bug fix for the keyboard focus that I need in this update.

There are a few notable changes:
- New `themeMode` and `background` props on the `Grommet` component which are necessary for the light/dark theme toggling now.
- The `FormField` component assumes theme is available in context, so you'll notice that I had to do some refactoring for component tests that use the `FormField`
  - To make this easier, I updated enzyme to its latest which includes [a new helpful API method and options](https://enzymejs.github.io/enzyme/docs/api/shallow.html#arguments) for shallow rendering or full mounting. You can set the context wrapper in the options which keeps the wrapper for the component that you actually want to test.
- **Important to note that the form components in Grommet are now controlled.** I consider this a breaking change even though it was in a minor version release. I think it came in by way of some validation improvements, but isn't explicitly noted in their changelogs. What this means is that I had to modify the `TextTask` to be controlled again. This seems to be a minor change of `defaultValue` to use `value` and I haven't been able to replicate the initial issue which led to it being refactored to be uncontrolled. This should definitely be double checked by others, however.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
I've built a local Docker image from this branch and I'm not seeing any problems on a drawing/question project, either on the home page or the classify page. 
A screenshot of a staging project running in the new app.

eatyourgreens:
I don't see any problems on a text workflow either: http://localhost.zooniverse.org:3000/projects/eatyourgreens/-project-testing-ground/classify?env=staging&workflow=2641

eatyourgreens:
If I sign out from the content pages app, I get
```
Unhandled Rejection (TypeError): Cannot read property 'user' of undefined
```
which, I think, is completely unrelated to this PR.

packages/app-project/src/components/AuthModal/components/LoginForm/components/Form/Form.spec.js
```diff
@@ -1,25 +1,31 @@
...
   Form
 } from './Form'
 
-describe('Component > Form', function () {
+describe('LoginForm > Component > Form', function () {
   it('should render without crashing', function () {
-    const wrapper = shallow(<Form />)
+    const wrapper = shallow(
+      <Form />,
+      { wrappingComponent: <Grommet />, wrappingComponentProps: { theme: zooTheme } }
```
eatyourgreens: How weird that they chose to do this instead of
```js
{ wrappingComponent: Grommet, wrappingComponentProps: { theme: zooTheme } }
```
or
```js
{ wrappingComponent: <Grommet theme={zooTheme} /> }
```

packages/app-project/src/components/AuthModal/components/LoginForm/components/Form/Form.spec.js
```diff
@@ -1,25 +1,31 @@
...
   })
 
   describe('fields', function () {
     let rendered
 
     before(function () {
-      const wrapper = shallow(<Form />)
+      const wrapper = mount(
+        <Form />,
+        { wrappingComponent: Grommet, wrappingComponentProps: { theme: zooTheme } }
```
eatyourgreens: If you're re-using this, is it worth declaring it as a constant somewhere?

eatyourgreens: I'm also a little unclear as to whether `wrappingComponent` should be a component or an instance of a component. You seem to be doing both in the code.

packages/app-project/src/components/AuthModal/components/LoginForm/components/Form/Form.spec.js
```diff
@@ -41,7 +47,10 @@ describe('Component > Form', function () {
   describe('when being edited', function () {
     it('should call handleChange on the inputs', function () {
       const handleChangeSpy = sinon.spy()
-      const wrapper = shallow(<Form handleChange={handleChangeSpy} />)
+      const wrapper = shallow(
+        <Form handleChange={handleChangeSpy} />,
+        { wrappingComponent: <Grommet />, wrappingComponentProps: { theme: zooTheme } }
```
eatyourgreens: Maybe declare this as a constant too, since you seem to be re-using it in a bunch of tests.

packages/app-project/src/components/AuthModal/components/RegisterForm/components/Form/Form.spec.js
```diff
@@ -18,17 +18,23 @@ import {
...
   })
 
   describe('fields', function () {
     let rendered
 
     before(function () {
-      const wrapper = shallow(<Form />)
+      const wrapper = mount(
+        <Form />,
+        { wrappingComponent: Grommet, wrappingComponentProps: { theme: zooTheme } }
```
eatyourgreens: Same comment here as to whether we're passing `Grommet` or `<Grommet />` as an option.

packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/Button/Button.spec.js
```diff
@@ -1,18 +1,21 @@
...
 import sinon from 'sinon'
+import { Grommet } from 'grommet'
+import zooTheme from '@zooniverse/grommet-theme'
 import Button from './Button'
 
-describe('Component > Button', function () {
+describe('ImageToolbar > Component > Button', function () {
   it('should render without crashing', function () {
-    const wrapper = shallow(<Button />)
+    const wrapper = shallow(<Button />, { wrappingComponent: <Grommet />, wrappingComponentProps: { theme: zooTheme } })
```
eatyourgreens: Maybe put this on multiple lines.

packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/TextTask.js
```diff
@@ -70,7 +71,7 @@ function TextTask (props) {
           autoFocus={autoFocus}
           disabled={disabled}
           id={`${task.taskKey}-${task.type}`}
-          defaultValue={value}
+          value={value}
```
eatyourgreens: I wonder if this is uncontrolled under the hood, because it still works even though the code never updates `value` on change, as far as I can tell.

packages/app-project/src/components/AuthModal/components/LoginForm/components/Form/Form.spec.js
srallen: I don't have an explanation for this either. When using shallow, I could only get it to work if using the JSX component, `<Grommet />` in the options object, but using mount it only works with `Grommet`. That seems like a bug to me.

packages/app-project/src/components/AuthModal/components/LoginForm/components/Form/Form.spec.js
srallen: Yep. Shallow only worked with `<Grommet />`, and mount only worked with `Grommet`. I couldn't find any explanation in the docs.

packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/TextTask.js
srallen: It's controlled. I checked their source code when debugging this.

srallen: For ref: https://github.com/grommet/grommet/blob/master/src/js/components/TextArea/TextArea.js

All of the form components are using hooks and context now for setting the value.

packages/app-project/src/components/AuthModal/components/LoginForm/components/Form/Form.spec.js
srallen: Almost all the shallows in the tests and the single mount here have slight differences (changes in the props on `Form` itself for instance), so there's not much re-use actually.

srallen: Or you mean the options object itself? That I can do

eatyourgreens: Yeah, I was thinking something like
```js
const formContext = { wrappingComponent: Grommet, wrappingComponentProps: { theme: zooTheme }}
```
might save some repetition.

packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/TextTask.js
eatyourgreens: Interesting. So the value that's passed in here is used as the default value for `FormContext` in the `TextArea` code, then the input updates `FormContext` via `setValue`. I was trying to figure out how this was working since our code doesn't directly control `value`.

Issue #1316, opened 2019-11-20
Transcription task text task with suggestions UI


srallen:
**Package** lib-classifier

Related to #1271 and #1313, the UI for the transcription task sub-task text input can use Grommet's `TextInput` since it supports out of the box suggestions. Here is the story demonstrating this: https://storybook.grommet.io/?path=/story/textinput--suggestions

The advantage of using this is that it'll pick up the zooniverse theme and be styleable.

An HTML input with a datalist is a possibility, but wouldn't be styeable. 

@eatyourgreens comments in #1268 about long lines of text and the text task implements a textarea to deal with this possibility (see: https://github.com/zooniverse/front-end-monorepo/pull/1268#discussion_r348331539).

ASM leverages the pop-up resizing function for long text rather than use a textarea. @beckyrother do you have a preference on dealing with long lines of text?

beckyrother:
Storybook looks great! It would be better if long text wrapped if possible, like on ASM.

eatyourgreens:
This could maybe extend the `TranscriptionLine` mark model to add suggested transcriptions for each line.

beckyrother:
Here's the transcription task on InVision: https://invis.io/UAN6AP0JTBC#/396381419_90_Text_Transcription

beckyrother:
Previous transcriptions: adding a dropdown + 'previous transcriptions' help text:

https://invis.io/UAN6AP0JTBC#/404084878_91-2_Text_Transcription_Line_Text_Box_Collaborative

snblickhan:
A few questions/suggestions:
1. Can the list of Text Modifiers be above the text entry box, so they aren't covered by the dropdown menu when it's open?
2. Will Project Builders have the option of inserting guidelines/instructions on the text entry popup box (in the space below "TRANSCRIBE LINE" but above the text entry field? Thinking of how the transcription box in ASM includes the following: "Please transcribe all of the words in the line of text. Open the dropdown menu to use previous volunteers' transcriptions as a starting point." Could be worth giving PBs the option to give further instruction in that space?
3. Is it worth considering the option to remove the number required from the tool label in this case? With the exception of structured data, it's likely that in most cases there won't be a known number of lines that need to be transcribed, and I worry that this may confuse volunteers.
4. Will the drag-to-adjust option on the text entry field resize the entire popup, or just the text entry field? (This may be a silly question -- apologies if I'm using incorrect terminology regarding any of these components)

beckyrother:
[...]

PR #1576, opened 2020-03-24
[WIP] Refactor tasks to use hooks instead of inject

srallen:

Package: lib-classifier

Describe your changes:
I wanted to get a start on writing a story for #1316 but got a bit sidetracked with some aspects of the task stories being broken. I was encountering this error when trying to use the knobs: https://github.com/mobxjs/mobx-react#the-set-of-provided-stores-has-changed-error

I read this to mean that the Tasks component should be updated to use context and hooks, which I did, but that didn't fix the error. The refactor is still worth keeping because we'll need to do this anyway for our connect components.

I also fixed the background color of the task area in the stories, fixed a key error on the task help, and added the ellipse to the drawing task story.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import { withTheme } from 'styled-components'
 
+import { useStores } from '@helpers/hooks'
 import TaskHelp from './components/TaskHelp'
 import TaskNavButtons from './components/TaskNavButtons'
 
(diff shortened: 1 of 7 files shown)


srallen:
WIP because there are a couple of broken tests.

eatyourgreens:
To get the tests working with `useContext`, you'll need to wrap your components in a context provider, with a mock context, and use full DOM mounting. Here's an example from the `InteractionLayer` tests:
https://github.com/zooniverse/front-end-monorepo/blob/4a52094b61ba73c81a66d4ba5b8806fb5d00ef7d/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js#L82-L88

Full mounting might cause problems here, because the Task Nav components are all connected to stores themselves. I remember that being a headache when I first set up the stories and tests for this component.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
```diff
@@ -24,67 +25,74 @@ function storeMapper (stores) {
...
+    subjectReadyState,
+    step,
+    theme
+  } = storeMapper(stores)
 
-  [asyncStates.error] () {
-    console.error('There was an error loading the workflow steps and tasks.')
-    return (<Paragraph>Something went wrong</Paragraph>)
-  }
+  const stateRenderFunctions = {
```
eatyourgreens: This object will break if the strings set in `asyncStates` change. I think you can use dynamic object properties to map states to handlers:
```js
function onError() {
…error handling
}
function onSuccess() {
…success stuff.
}
const stateRenderFunctions = {
  [asyncStates.error]: onError,
  [asyncStates.success]: onSuccess
}
```

Alternatively, maybe use dynamic method naming, like the original class did?

```js
const stateRenderFunctions = {
  [asyncStates.error] () {
   …error handling stuff
  },
  [asyncStates.success] () {
   …render the tasks
  }
}
```

eatyourgreens:
> I was encountering this error when trying to use the knobs:

The knobs are trying to change loading state, which used to be passed in as a prop but now is passed in via the store-mapper. We could split the `Tasks` component into a container and a view, then run the tests and stories against the view, but I think I ran into problems with that because of the Task Nav buttons being wrapped in `inject/observer`.

eatyourgreens:
I've opened #1577 to fix those storybook errors, which turned out to be more involved than I thought it would be. The problem was the `store` object being created anew each time a story function runs. I moved the store outside the stories, and set the story functions to update the store via actions.

srallen:
I updated enzyme when I updated Grommet and the `wrappingComponent` options should work for when I need context I think: https://enzymejs.github.io/enzyme/docs/api/shallow.html#shallownode-options--shallowwrapper

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
srallen: I tried that yesterday but ran into type errors. I can try again.

eatyourgreens:
Sure, but I think the tests for things like loading state will fail, because they use props to pass the mock state into the `Tasks` component.

srallen:
Closing this as it'll be easier to rework the merge conflict, my storybook refactor, and the hooks refactor separately.

PR #1578, opened 2020-03-25
Multi-frame viewer styling clean-up

mcbouslog:
Package:
- lib-classifier

Describe your changes:
- edit margins between frames
- edit previous/next buttons' focus/hover styling
- pointer cursor for frames
- edit frame list styling

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


mcbouslog:
@beckyrother what's your preferred way of reviewing? I can post screenshots, including animated gifs, though I'm not sure if that's the best way for you to review functionality.

Also, I've struggled with the active border and focus styling, which is on the input/image element, and the parent label element's height/width and margins. I think it's fine as is, but there might be a better way.

eatyourgreens:
Only one element in the DOM can be `document.activeElement` and apply `:focus` styles, so this does sound like a Grommet bug. The Previous button should lose focus styling once `:focus` moves to another element.

srallen:
I think it'd probably work as expected if we were using Grommet components for the input list.

beckyrother:
Hi @mcbouslog sorry for the delay! Sarah's been creating gifs of interactions which work pretty well, so I'm fine with that if you'd like to do that.

mcbouslog:
Attempted to use Grommet's https://v2.grommet.io/list unsuccessfully, as it worked in some ways but by default has certain functionality that created new issues or didn't work. The Grommet [RadioButtonGroup](https://storybook.grommet.io/?path=/story/radiobuttongroup--children) showed promise though with some initial work this afternoon.

mcbouslog:
@srallen do you usually share `lib-classifier` screenshots from that library's dev env or do you rebuild `app-project` with `lib-classifier` changes and share from `app-project`'s dev env? I don't think makes a big difference for this case, but just curious. prob depends.

@beckyrother - here are some mp4s from a test project,
few frames:
frame-viewer-few-frames
clicks:
frame-viewer-clicks
tabs:
frame-viewer-tabs

mcbouslog:
I'll create a new issue for the focus issue.

I also started adding the Media component from `react-components` to utilize its use of the thumbnail service as well as other helpful functions, but that affects how the focus styling is applied as well, so I think adding the Media component and refactoring for proper focus should come in a subsequent PR? I think they're so intertwined it makes sense, but open to suggestion.

eatyourgreens:
It's also worth checking focus in Safari, Chrome and Firefox. The Chrome team have a different opinion, from Mozilla, as to how focus should be managed. You'll probably notice that, in FF, clicking an element does not automatically focus that element.

srallen:
@mcbouslog I think it depends on how far along the feature is in development, whether or not the context of the app is important for reviewing it, etc. I've taken screenshots from storybook, the dev classifier, and from the project app. If you aren't sure, I think just asking what would be useful for the reviewer is probably sufficient. 

Also agree to just note the focus issue separately and not get bogged down in it for this PR.

beckyrother:
These look great, thanks @mcbouslog ! 

Small note that maybe doesn't belong here – when the image is shorter than the toolbar, can it be centered vertically with a white background rather than top aligned:
image

mcbouslog:
@beckyrother 💯 on vertically centering wider images, which touches on a more general dynamic layout based on image dimensions and subject classification purpose that I think we have planned for later this year, but I'm going to leave out of this PR, though will make sure we have noted.

PR #1528, opened 2020-02-19
Remove Docker production-apps build

eatyourgreens:
Move the NextJS app builds back into the bootstrap script.
Remove the local production-apps build image.
Remove all references to named targets.
Build all docker images, dev and production, from the bootstrapped monorepo.

This PR reverts most of the changes from #1109 and moves back to the much simpler build process that we originally used, where bootstrapping the monorepo built all its packages.

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #1589, opened 2020-03-31
Set sentry environment during the Jenkins build

### LINKED ISSUE #1588: Sentry environment should be staging for staging deploys
**Package**
app-content-pages
app-project

**Describe the bug**
Sentry environment is never set for the content pages app, and is always set to `production` for the deployed project app.

The project app sets the sentry environment from the `APP_ENV` environment variable here:
https://github.com/zooniverse/front-end-monorepo/blob/d8e90b403ddb3c986789d931609dce723b639b1f/packages/app-project/src/helpers/logger/initializeSentry.js#L5-L12

As far as I can tell, `APP_ENV` is `production` during the Jenkins build, when the NextJS apps are built. Staging deploys set `APP_ENV` to `staging` but, at that point, the JS code has already been built to use `production`.

https://github.com/zooniverse/front-end-monorepo/blob/07b4e0705972a234b82cb94e7df0104459163710/packages/app-project/Dockerfile#L9-L12

eatyourgreens:
Add APP_ENV to the content pages app.
Split the app builds into staging (master branch) and production (production-release tag.)

Package:
app-content-pages
app-project

Closes #1588.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


Jenkinsfile
```diff
@@ -42,21 +42,25 @@ pipeline {
         }
 
         stage('Build app Docker images') {
+          environment {
+            APP_ENV = "${env.TAG_NAME == "production-release" ? "production" : "staging"}"
```
camallen: I don't think this will work as this stage is only run on master branch events, https://github.com/zooniverse/front-end-monorepo/blob/63a23399074c0b6e2520e78187d256bb6056c979/Jenkinsfile#L28

so we'd have to build on both prod tag events and master to distinguish between staging / prod.

eatyourgreens: Isn't `TAG_NAME` set on master branch events that are tagged with `production-release`? I was going with [this Stack Overflow suggestion](https://stackoverflow.com/questions/44007034/conditional-environment-variables-in-jenkins-declarative-pipeline/53771302#53771302) but I don't know if it's right.

eatyourgreens: Maybe I should go back to #1528 and have separate builds for staging and production.

camallen: tag name won't be available unless the build is triggered by a tag event

eatyourgreens: Cheers! That's very useful to know. How do you find this stuff out? I was googling Stack Overflow posts yesterday to find out how Jenkins environment variables work.

eatyourgreens:
So something like:
```
stage('build production') {
      when { tag 'production-release' }
      environment {
        APP_ENV = "production"
        …
      }
      steps {
        …build the production images
      }
    }
    stage('Deploy staging to Kubernetes') {
      when { branch 'master' }
      environment {
        APP_ENV = "staging"
        …
      }
      steps {
        …build the staging images
      }
    }
```

Jenkinsfile
camallen: with great struggle - it might have changed due to the plugin behaviour changing / upgrading but i ran into this a while back, https://issues.jenkins-ci.org/browse/JENKINS-58751?focusedCommentId=372585&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-372585
>Execute the stage when the build is building a tag. Example: when { buildingTag() }
https://jenkins.io/doc/book/pipeline/syntax/#when)

Jenkinsfile
```diff
@@ -48,6 +48,7 @@ pipeline {
               agent any
 
               environment {
+                APP_ENV = "${env.TAG_NAME == "production-release" ? "production" : "staging"}"
```
camallen, line 51: sadly TAG_NAME will never be available here as it's under the master branch `when { branch 'master' }`

eatyourgreens:
> I'm not sure i follow, master commit hash will diverge from deployed production-release as PRs are merged and deploy to staging, right? They will converge once we deploy production.

That's what's confusing me. We'll build an image from the HEAD of master, with `APP_ENV = staging` and tag that with the commit hash. Later, we'll move the `production-release` tag to that commit. Will the tag build create a new image, with `APP_ENV=production`, or will it use the existing image for that commit hash (which has the staging environment.)

eatyourgreens:
I've split the app builds into a master branch build and a production-release tag build. I'm not 100% sure if the stages are right. The indentation is a bit confusing to follow.

Jenkinsfile
```diff
@@ -24,68 +24,120 @@ pipeline {
...
+          def dockerImageName = "${dockerRepoName}:${BRANCH_NAME}"
+          def newImage = docker.build(dockerImageName, "--target bootstrap ./")
+          newImage.push()
+          newImage.push('latest')
+        }
+      }
+    }
+
+    stage('Build production app images') {
+      when { tag 'production-release' }
```
camallen, line 43: It might be simpler to collapse prod and master to the same stage here using when `anyOf` https://jenkins.io/doc/book/pipeline/syntax/#when-example

then detect the presence of the TAG_NAME to determine if it's prod build vs staging?

eatyourgreens: That's an excellent idea. Have you memorised all the Jenkins documentation? 😉

eatyourgreens: So I could go back to using a ternary based on `TAG_NAME`?

camallen: i think so - best to test. 

I tend to use the replay jobs function of jenkins to test jenkinsfiles, https://pushbuildtestdeploy.com/using-jenkinss-replay-feature-to-work-on-without-repeatedly-committing-the-jenkinsfile/

eatyourgreens: If it doesn't work, I think the worst that happens is `APP_ENV` is staging for all builds. As far as I can tell, it's only used to set the Sentry event environment, and we aren't using those in Sentry anyway. So it wouldn't improve the error reports, but it wouldn't break anything either.

eatyourgreens: I replayed a build for this branch, changing `branch 'master'` to `branch 'sentry-environments'` throughout the Jenkinsfile. It still skips those `when` conditions and I can't see why.
```
Stage "Build app images" skipped due to when conditional
```
https://jenkins.zooniverse.org/job/Zooniverse%20GitHub/job/front-end-monorepo/job/PR-1589/7/console

Maybe I need to check a different value of the branch name for a PR merge?

eatyourgreens: With the correct branch name, Jenkins builds and deploys to staging, and I'm seeing the `staging` environment tagged on events from that build.

Issue #1603, opened 2020-04-06
Logging out crashes the About pages


eatyourgreens:
**Package**
app-content-pages

**Describe the bug**
Signing out on a content page throws the following error. It also leaves your name in the top menu, giving the impression that you haven't been signed out from Panoptes (but you have.)
```
Unhandled Rejection (TypeError): Cannot read property 'user' of undefined
```

PR #1604, opened 2020-04-06
Content pages: Fix login and logout errors

### LINKED ISSUE #1602: The login form crashes the About pages
**Package**
app-content-pages

**Describe the bug**
https://www.zooniverse.org/about/publications?login=true
Opening the login form on a content page crashes the page with the following error.
```
TypeError: Cannot read property 'focus' of null
```

The page then gets stuck in an infinite loop of rendering and crashing. The only way out is to either force close the browser process, or wait for the 'page unresponsive' dialog to pop up and kill the page.

eatyourgreens:
Replace old login and register components with the newer AuthModal component from the project app.
Fix broken sign out function.
Upgrade Panoptes client from 3.0.0-rc.0 to 3.2.1.

Package:
app-content-pages
app-project

Closes #1602.
Closes: #1603.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
The Grommet password input will not autofill my saved password (in Chrome) and I can't see why.

eatyourgreens:
I'm not sure. It's in the page header, so it could give you the impression that you've been logged out from the whole site, if we remove it.

srallen:
We could talk to @beckyrother about a different header design for content parts of the site?

beckyrother:
If this is referring to the black bar at the top, it would be best if it stayed the same across the site

eatyourgreens:
It is the black bar across the top. The bug is slightly nasty in that if you go to publications or the team page while not logged in, and click Sign In, you crash the browser tab. At least, that’s what I found in Firefox. I’ve no idea why the effects are that severe.

eatyourgreens:
Sounds like a good idea. This was a quick fix because the bug is so severe.

eatyourgreens:
Does the POST for the token work for production from the project app?

srallen:
Yes, sign in and sign out works on production for the project app.

eatyourgreens:
How strange. Both apps should be using the same auth client.

eatyourgreens:
No problems here using the production API to get a token on localhost.zooniverse.org. However, I notice that both apps are on an old version of the client, so I'll upgrade.

Screenshot of the production auth process in Chrome on the Team page.

srallen:
I'm still getting a 422 using `http://local.zooniverse.org:3000/` doing a production build and start of the app. I noticed you're adding the environment variable on the url to run in production. Are you trying out the production build or using the dev build with query param?

I tried both dev with the env query param and production builds as well as add `localhost.zooniverse.org` to my hosts file (I've always used just `local.zooniverse.org`) and still get a 422. (BTW adding `localhost.zooniverse.org` and using that fixed fetching contentful content for me, something broken that I assumed was just me, so the configuration assumes we're all just using `localhost.zooniverse.org` even though authentication should work with either).

eatyourgreens:
I confirmed in the network tab that it was using the production API. What error message is Panoptes returning with the 422 error? Or is it an empty response?

eatyourgreens:
Production docker build for me, by the way.

eatyourgreens:
@srallen If I run a Docker build in `app-content-pages` with
```
docker-compose -f docker-compose-prod.yml up
```
I get errors, but that's because the production build doesn't include this branch. That build is using commit `2234b9cae3f0fc967d430e1016b8636564b8a083`, according to the console log in dev tools. I don't get 422 errors, though. I see the bug that this branch fixes, where the sign-in link crashes the browser.

eatyourgreens:
To run a production build locally, by the way, run `yarn build && yarn start` to build the production app and start a local server on port 3000. That should work if your docker build isn't working.

eatyourgreens:
Here's the network request that I see with a local production build.

Screenshot showing a 200 response for an auth token, from the production Panoptes API.

eatyourgreens:
I tried running `docker image list` and my copy of `zooniverse/front-end-monorepo:latest` is 7 weeks old, which explains why it's running from a stale commit. @srallen you could try checking that, if you are testing locally with that image.

A screenshot of the current docker image list on my MacBook Pro.

srallen:
Yes, I know how to run the app using yarn commands as well as docker. I used the nuclear option the other day and pruned all of my local images and re-pulled and built, but I'll double check what commit its running from later today.

eatyourgreens:
Not in the immediate future but definitely open an issue. A lot of the common code seems to be container components and state stores, so maybe this needs its own shared library?

eatyourgreens:
Let's see if merging this updates the linked Sentry issues.

eatyourgreens:
Looks like Sentry issues are only closed if you [manually list the Sentry IDs in the commit message](https://docs.sentry.io/workflow/integrations/global-integrations/#resolving-in-commitpull-request). That's annoying. I'm going to close those issues in Sentry.

PR #1634, opened 2020-04-30
Content pages: remove next/link

eatyourgreens:
Remove client-side navigation from content pages by removing [next/link](https://nextjs.org/docs/api-reference/next/link) from the NavLink component.

Package:
app-content-pages

Closes #1630 .

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

Issue #1423, opened 2020-01-16
[RFC] ADR process


srallen:
There's been discussion on the ADR process and we should evaluate how it's been going, what work works, what doesn't, and any other concerns. This RFC is to centralize a single point for discussion.

camallen:
I think it’s fair to have relevant, in person / slack / emails discussions documented as ADRs as we adopt features into the codebase. 

As long as the ADR processes isn’t blocking folks delivery of features. Remembering that we are a small team working somewhat independently and folks will make choices along the way as they ‘discover’ the system they are building and it might be that going through the ADR processes before implementation might hinder our team. 

I do agree that implementation choices and tradeoffs should be documented as a record as features are accepted into the codebase.

On the flip side  experiments / hacking / playful work should be blocked by an ADR at all, they should only come into play when a team member gets serious about changing the architecture of an app.

srallen:
> As long as the ADR processes isn’t blocking folks delivery of features. Remembering that we are a small team working somewhat independently and folks will make choices along the way as they ‘discover’ the system they are building and it might be that going through the ADR processes before implementation might hinder our team.

I don't think this is actually happening though. Do we have an actual example of hinderance? Or is this just a precautionary concern?

I generally think the process should be RFC > ADR, if we aren't sure on a decision, then it could be, RFC > experiment > proposal ADR > experiment > revised ADR and this has already happened when Rog and I experimented with SSR rendering frameworks finally settling on Next.js. 

Many of the ADRs we *are* fairly sure of because we already know what hasn't been working from PFE and have a reasonable idea of how to go forward. In these cases the initial RFC might be skipped and just go straight into a proposal ADR.

srallen:
Spotify Labs has a really nice article on RFC --> ADR processes especially a helpful guide on using them even for smaller decisions: https://labs.spotify.com/2020/04/14/when-should-i-write-an-architecture-decision-record/

I think what will help us is is to define our own impact assessment and decision flow of when to use RFCs and ADRs?

PR #1635, opened 2020-04-30
Content pages: remove getInitialProps and the content API

### LINKED ISSUE #1566: Content pages: remove getInitialProps
The About pages don't change much, but we generate the HTML on every request. If we remove [getInitialProps](https://github.com/zooniverse/front-end-monorepo/blob/8ff7209acebc71cd586a17e552bb955647553de0/packages/app-content-pages/pages/_app.js#L25-L39) we can [render the pages as static HTML](https://nextjs.org/docs/advanced-features/automatic-static-optimization).

See also [`next export`](https://nextjs.org/docs/advanced-features/static-html-export) for building and serving static pages.

eatyourgreens:
This removes `getInitialProps`, and the page content API, in favour of [`getServerSideProps`](https://nextjs.org/docs/basic-features/data-fetching#getserversideprops-server-side-rendering). On each request, page content is fetched from the local content cache and rendered as static HTML (and JSON) for the Team and Publications pages. That HTML  (or JSON) is then served by the app. The content cache (on the server) refreshes from the Contentful API as before.

Package:
app-content-pages

Closes #1566.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-content-pages/src/screens/Publications/getServerSideProps.spec.js
+++ b/packages/app-content-pages/src/screens/Publications/getServerSideProps.spec.js
@@ -1,12 +1,12 @@
 import sinon from 'sinon'
 
 import cache from '../../api/publications/publicationsCache'
-import getStaticProps from './getStaticProps'
+import getServerSideProps from './getServerSideProps'
 import mockData from './PublicationsContainer.mock'
 
 const DATA = mockData
 
-describe('Component > PublicationsContainer > getStaticProps', function () {
+describe('Component > PublicationsContainer > getServerSideProps', function () {
 
   describe('populates the "publicationsData" props from contentful API', function () {
     let getpublicationsDataStub
@@ -17,7 +17,7 @@ describe('Component > PublicationsContainer > getStaticProps', function () {
 
     it('should handle valid API data', async () => {
       getpublicationsDataStub = sinon.stub(cache, 'get').returns(DATA)
-      const { props } = await getStaticProps({})
+      const { props } = await getServerSideProps({})
       expect(props).to.deep.equal({
         error: null,
         publicationsData: DATA
@@ -26,7 +26,7 @@ describe('Component > PublicationsContainer > getStaticProps', function () {
 
     it('should handle empty API reponse', async () => {
       getpublicationsDataStub = sinon.stub(cache, 'get').returns([])
-      const { props } = await getStaticProps({})
+      const { props } = await getServerSideProps({})
       expect(props).to.deep.equal({
         error: null,
         publicationsData: []
@@ -37,7 +37,7 @@ describe('Component > PublicationsContainer > getStaticProps', function () {
       var errorMsg = 'failed to connect to API'
       var errorPromise = Promise.reject(new Error(errorMsg))
       getpublicationsDataStub = sinon.stub(cache, 'get').returns(errorPromise)
-      const { props } = await getStaticProps({})
+      const { props } = await getServerSideProps({})
       expect(props).to.deep.equal({
         error: errorMsg,
         publicationsData: []
--- a/packages/app-content-pages/src/screens/Team/getServerSideProps.spec.js
+++ b/packages/app-content-pages/src/screens/Team/getServerSideProps.spec.js
@@ -1,12 +1,12 @@
 import sinon from 'sinon'
 
 import cache from '../../api/team/teamCache'
-import getStaticProps from './getStaticProps'
+import getServerSideProps from './getServerSideProps'
 import mockData from './TeamContainer.mock'
 
 const DATA = mockData
 
-describe('Component > TeamContainer > getStaticProps', function () {
+describe('Component > TeamContainer > getServerSideProps', function () {
 
   describe('populates the "teamData" props from contentful API', function () {
     let getTeamDataStub
@@ -17,7 +17,7 @@ describe('Component > TeamContainer > getStaticProps', function () {
 
     it('should handle valid API data', async () => {
       getTeamDataStub = sinon.stub(cache, 'get').returns(DATA)
-      const { props } = await getStaticProps({})
+      const { props } = await getServerSideProps({})
       expect(props).to.deep.equal({
         error: null,
         teamData: DATA
@@ -26,7 +26,7 @@ describe('Component > TeamContainer > getStaticProps', function () {
 
     it('should handle empty API reponse', async () => {
       getTeamDataStub = sinon.stub(cache, 'get').returns([])
-      const { props } = await getStaticProps({})
+      const { props } = await getServerSideProps({})
       expect(props).to.deep.equal({
         error: null,
         teamData: []
@@ -37,7 +37,7 @@ describe('Component > TeamContainer > getStaticProps', function () {
       var errorMsg = 'failed to connect to API'
       var errorPromise = Promise.reject(new Error(errorMsg))
       getTeamDataStub = sinon.stub(cache, 'get').returns(errorPromise)
-      const { props } = await getStaticProps({})
+      const { props } = await getServerSideProps({})
       expect(props).to.deep.equal({
         error: errorMsg,
         teamData: []
--- a/kubernetes/deployment-staging.tmpl
+++ b/kubernetes/deployment-staging.tmpl
@@ -31,6 +31,13 @@ spec:
       containers:
         - name: fe-project-staging
           image: zooniverse/fe-project-staging:__IMAGE_TAG__
+          resources:
+            requests:
+              memory: "200Mi"
+              cpu: "10m"
+            limits:
+              memory: "200Mi"
+              cpu: "500m"
           ports:
             - containerPort: 3000
           env:
@@ -84,6 +91,13 @@ spec:
       containers:
         - name: fe-content-pages-staging
           image: zooniverse/fe-content-pages-staging:__IMAGE_TAG__
+          resources:
+            requests:
+              memory: "200Mi"
+              cpu: "10m"
+            limits:
+              memory: "200Mi"
+              cpu: "500m"
           ports:
             - containerPort: 3000
           env:
--- a/packages/app-content-pages/package.json
+++ b/packages/app-content-pages/package.json
@@ -38,7 +38,7 @@
     "mobx-state-tree": "~3.15.0",
     "morgan": "^1.9.1",
     "newrelic": "^5.10.0",
-    "next": "~9.3.6",
+    "next": "~9.4.1",
     "next-absolute-url": "~1.1.1",
     "node-cache": "~5.0.2",
     "panoptes-client": "~3.2.1",
--- a/packages/app-content-pages/pages/about/publications.js
+++ b/packages/app-content-pages/pages/about/publications.js
@@ -1,1 +1,1 @@
-export { default, getStaticProps } from '../../src/screens/Publications'
+export { default, getServerSideProps } from '../../src/screens/Publications'
--- a/packages/app-content-pages/pages/about/publications.js
+++ b/packages/app-content-pages/pages/about/publications.js
@@ -1,1 +1,1 @@
-export { default } from '../../src/screens/Publications'
+export { default, getServerSideProps } from '../../src/screens/Publications'
--- a/packages/app-content-pages/pages/about/publications.js
+++ b/packages/app-content-pages/pages/about/publications.js
@@ -1,1 +1,1 @@
-export { default, getServerSideProps } from '../../src/screens/Publications'
+export { default, getStaticProps } from '../../src/screens/Publications'
--- a/packages/app-content-pages/pages/about/team.js
+++ b/packages/app-content-pages/pages/about/team.js
@@ -1,3 +1,3 @@
-export { default, getStaticProps } from '../../src/screens/Team'
+export { default, getServerSideProps } from '../../src/screens/Team'
 
 
--- a/packages/app-content-pages/pages/about/team.js
+++ b/packages/app-content-pages/pages/about/team.js
@@ -1,1 +1,3 @@
-export { default } from '../../src/screens/Team'
+export { default, getServerSideProps } from '../../src/screens/Team'
+
+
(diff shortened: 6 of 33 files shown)


eatyourgreens:
Even with totally static rendering, Next `Link` still makes XHR requests on page change. So CORS errors would still happen with this setup and Next handling the page navigation.

eatyourgreens:
Tests are fixed so this is ready for review.

eatyourgreens:
I've only tested it using `yarn build && yarn start` locally. I tried the production docker compose files, but they built a different branch. I'm assuming that was master, pulled down from dockerHub.

I think the dev build will work if you set `ASSET_PREFIX`, but I'm not entirely sure.

Something I would really like to do is test the NextJS apps locally with the browser and the server on different domains. I haven't been able to figure out how to do that. In theory, setting `ASSET_PREFIX` should do it, but I haven't had any luck when I've tried that.

eatyourgreens:
If you see this (with no api paths and the black dots) when the app builds, then you've got a completely static build for both pages.

```
Creating an optimized production build  

Compiled successfully.

Automatically optimizing pages  

Page                                                           Size     First Load JS
┌   /_app                                                      50.2 kB         603 kB
├ ○ /404                                                       3.01 kB         606 kB
├ ● /about/publications                                        1.76 kB         632 kB
├ ● /about/team                                                5.4 kB          636 kB
└ ○ /Index                                                     332 B           603 kB
+ First Load JS shared by all                                  603 kB
  ├ static/pages/_app.js                                       50.2 kB
  ├ chunks/1d1f41e5323d9a4629773c3f197126f23f76d7e1.804244.js  10.2 kB
  ├ chunks/507ea89b563f1354750d19864b4a562a1c1ecc83.0317c2.js  13.5 kB
  ├ chunks/675ed458c968b559aeb04314a5d8b53423f69d69.82f247.js  28.8 kB
  ├ chunks/8313d723.9f49ac.js                                  14.4 kB
  ├ chunks/a289364f.ab84ce.js                                  20.2 kB
  ├ chunks/b7dac8a5.49a6f5.js                                  13.3 kB
  ├ chunks/ff86ee6953622fa94b35780e7362446efc9863e6.141076.js  405 kB
  ├ chunks/framework.f8a63c.js                                 40.4 kB
  ├ runtime/main.4180dd.js                                     5.55 kB
  └ runtime/webpack.c21266.js                                  794 B

λ  (Server)  server-side renders at runtime (uses getInitialProps or getServerSideProps)
○  (Static)  automatically rendered as static HTML (uses no initial props)
●  (SSG)     automatically generated as static HTML + JSON (uses getStaticProps)

✨  Done in 39.61s.

```

eatyourgreens:
If you’d like to experiment with the two for comparison, pull this branch down and change the exported function name to `getServerSideProps`. That’s enough to get Next to run it on each request rather than only at build time.

eatyourgreens:
What you should see is that the page HTML is built from the Contentful response on every page request, even though the content hasn’t changed.

camallen:
I'm actively testing this using the local docker setup and once #1634 is in this will help with testing.

> If we're only going to serve up cached static content that was created at build time, I don't see any reason to continue to use Contentful at all.

I believe the benefit of contentful is for UX for non developers to add content, that and it provides the API for build time data access. 

I think we can look at going even further and try to generate static pages hosted on a webserver / CDN that is update on each build / change in contentful (though it may be harder to deal with CDN caching on `/about/team` url paths). Relying on next here would avoid those issues by the look (testing cache headers on those urls is key), https://nextjs.org/docs/basic-features/pages#summary

eatyourgreens:
#1634 is merged into this branch now, so you can try it with client-side navigation disabled. The team and publications pages should now be requested as HTML on every page change, rather than loaded as JSON and decoded in the browser.

eatyourgreens:
>I think we can look at going even further and try to generate static pages hosted on a webserver / CDN that is update on each build / change in contentful (though it may be harder to deal with CDN caching on /about/team url paths).

Next can be run as a Static Site Generator, although they keep this a little bit quiet. Replace `next build` with `next export` and you'll get HTML pages that could be hosted on S3 in the static content bucket.

You can also [supply the content as `.md` files](https://itnext.io/building-a-static-blog-site-with-markdown-and-next-js-702c515389b3) in the pages directory, which would be really useful for the other About pages. The Markdown content for those is currently in big JSON files that are a pain to edit.

srallen:
> I believe the benefit of contentful is for UX for non developers to add content, that and it provides the API for build time data access.

Yes, I know. I was a part of the decision making process to switch to a headless CMS to make updating this content friendly to non-devs. If we switch to requiring a new build when changes are made, then as I said, we lose one of the main benefits, which is empowering non-devs to make content changes as needed without involving a dev or a deploy.

Both of you will have to present a case to change what we decided [ADR 14](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-14.md) with what the benefits are to either: use a CDN (option @camallen proposing) or to, as I see it, revert to use markdown (option @eatyourgreens is proposing). Perhaps there is an option 3 here?

eatyourgreens:
Sorry, I was proposing to move across all the other Markdown content from PFE. Team and Publications should continue being loaded from Contentful.

eatyourgreens:
Building static content from a headless CMS is a pretty common practice. We can either hook up Contentful to run the Jenkins job when content changes, or stick with our current schedule and publish changes once a week.

srallen:
Why wouldn't we move all static page content to the CMS? Whats the benefit of leaving some as static markdown files in the repository and some in the CMS?

eatyourgreens:
Volunteers have translated those pages for us, so we have them in multiple languages. Easy to localise from the file system, but I don’t think we have localisation in Contentful.

Pages that need to be edited should go into Contentful. Pages that don’t change should be fine in the file system.

srallen:
Next.js 9.4 just released and now has [incremental static regeneration](https://nextjs.org/blog/next-9-4#incremental-static-regeneration-beta). I think this addresses my concerns, so worth trying out. They've also provided an example using Contentful: https://github.com/zeit/next.js/tree/canary/examples/cms-contentful

camallen:
Nice find @srallen - The new features (beta) seem very useful and exactly the use case we want our apps to pursue, i.e. serve static pre built pages, use CDNs for internet propagation and update with fresh content at runtime from API sources without rebuilding, this is a very exciting offering from next 🎉 

@eatyourgreens would you be happy to rework this PR to we try out the new `update existing pages` feature of next? 

FWIW i think this repo is a nice place to try out new features of next, like SSG of pages, and I don't think we should be going through an ADR process to try these features out. Rather we can discuss the merits of the inclusion of these features via the Pull Request and we can write up the Architecture Decision after we've chosen to pursue it (and before merging the PR most likely).

eatyourgreens:
I think I’d upgrade to 9.4 in a separate PR. Their example uses `getStaticProps` so I think this PR will still be needed. 

Their example code makes an API request to Contentful on each page request but they mention using webhooks so that sounds promising.

srallen:
> FWIW i think this repo is a nice place to try out new features of next, like SSG of pages, and I don't think we should be going through an ADR process to try these features out. Rather we can discuss the merits of the inclusion of these features via the Pull Request and we can write up the Architecture Decision after we've chosen to pursue it (and before merging the PR most likely).

Absolutely, the content pages app is meant to be test and learning app for next.js features on a low traffic part of the site before we decide to use it with the project pages app, but my concerns for _this particular change_ was that it impacted a past decision, which would make our documentation inaccurate, and impact the usefulness of the contentful setup by non-dev team members. I'm advocating for us being mindful of this as we try things out. 

Also, hard disagree on waiting to write up a proposal for a impactful change as having a documented cost vs benefit for it is necessary to make the decision. I'm open to the proposal being in the form of a few comparison PRs with the relevant information, though, and is what I asked for in an earlier comment. With this new feature, the update is now theoretically seamless to non-devs on the team using contentful, so this is less impactful and I don't think it's necessary now if it works. As this PR is now, I don't think I'd approve it and would like to see the 9.4 update and feature.

eatyourgreens:
The Contentful example code is the same as this PR (`getStaticProps` makes requests to Contentful, formats the response and returns the formatted response.) That’s reassuring. 

They don’t go into how 9.4 invalidates a page: request headers or the new param that’s added to the return value of `getStaticProps`? I get the impression that `getStaticProps` is run at scheduled intervals, so polling Contentful for changes once per second in their example?

camallen:
>They don’t go into how 9.4 invalidates a page: request headers or the new param that’s added to the return value of getStaticProps? I get the impression that getStaticProps is run at scheduled intervals, so polling Contentful for changes once per second in their example?

Interesting how there isn't much documentation on this, https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation however it appears it is discussed here though https://github.com/zeit/next.js/discussions/11552

I think we could attempt to re-render after a request past say a 60s timeout, i believe that will only request the data from the CMS API after the 60s timeout and if an incoming request is received. Agree the webhooks to trigger the invalidation from the data source look interesting.
e.g. 
```
export async function getStaticProps() {
  return {
    props: await getDataFromCMS(),
    // we will attempt to re-generate the page:
    // - when a request comes in
    // - at most once every second
    unstable_revalidate: 60
  }
}
```
Should we get an next upgrade PR in to rebase this over?

Regarding the following:
>  but my concerns for this particular change was that it impacted a past decision, which would make our documentation inaccurate, and impact the usefulness of the contentful setup by non-dev team members. I'm advocating for us being mindful of this as we try things out.

Agree we should approach this with good intentions and care. I'd argue that as software developers our whole job focus is controlled change. I dare say this is most prevalent in the fast moving world of JavaScript on the web. 

When we see reasons to implement new features or change the way a system works, we should embrace that change while balancing it with the requirement to leave decent documentation trails that are not solely code artifacts. 

In this case we would rewrite add a new ADR outlining the new reasoning and implementation choices. However, I would strongly argue against ensure there is consensus before initial implementation to ensure our team members are 'agile' and have the agency to attempt to change our systems for the better.

eatyourgreens:
Reading through the RFC, the server will rebuild the team and publications pages in the background every 60s. That’s still very inefficient for these pages. I can understand this approach for a news feed but our pages don’t change that fast. 

With content that rarely changes, the best (and cheapest) performance will still be static publishing that triggers when the content is edited. If we can hookup Contentful to publish those pages, in the same way that you can with Netlify (say), that would be ideal.

eatyourgreens:
A good sandbox for NextJS would be a stats page or dashboard that updates frequently. Or maybe a server-side version of the Talk comments feed. https://github.com/eatyourgreens/zooniverse-talk-activity

I started updating the project app to 9.3 earlier this week. It seems to be so far behind the content pages app that I gave up after a couple of hours. It has a completely different Next setup, to the point where I don’t think the content pages app can be used to test new features before applying them to the project app.

srallen:
Ok, the feature seems to be a top priority as indicated by the next.js team in the blog announcement. It was about ~2 months between 9.3 and the 9.4 releases, so I think we can expect another release soon. Here's what I propose:

- We wait and actively monitor the next.js RFC feature discussion since the webhook implementation is what it seems we need
- Once 9.5 is released, update and apply the feature with the webhook integration
- In the interim, open an issue for and update the app-project to 9.3
- Open an RFC issue to discuss another sandbox environment that can better help us assess and try out features if the app-content-pages is no longer serving that role. @eatyourgreens in this issue, could you describe your experience and reasoning that app-content-pages no longer serves this role? It certainly did when we first launched the two, but understandably changes have happened.

As far as process, again, I'm open to discuss about it and have this issue for it: https://github.com/zooniverse/front-end-monorepo/issues/1423

In this particular case, I think the related issue for this would have been sufficient to continue the conversation about best approach before the PR stage. From my perspective, with all of the needs between project deliverables and ongoing maintenance, I prefer to have conversations early and often before code is written so that we can avoid having code written and be wasted effort if there is a strong disagreement amongst the team about it.

eatyourgreens:
We still need this PR to remove the API routes, but switching to `getServerSideProps` is a small matter of renaming the new function.

eatyourgreens:
> Open an RFC issue to discuss another sandbox environment that can better help us assess and try out features if the app-content-pages is no longer serving that role. @eatyourgreens in this issue, could you describe your experience and reasoning that app-content-pages no longer serves this role? It certainly did when we first launched the two, but understandably changes have happened.

From what I can tell from the code, content pages have never used dynamic routing or a custom `App` component. Both of those would be needed in a test or sandbox environment for the project app. 

In addition, content pages don’t have an MST store. The project app’s use of MST is quite odd, though, and I want to issue that up, if only to get help understanding what the project code is doing and why.

eatyourgreens:
I've renamed `getStaticProps` to `getServerSideProps` so this should be ready now.

eatyourgreens:
I've updated the PR description too. The PR is now a refactor that removes the content API but doesn't change how the pages are built.

packages/app-content-pages/src/screens/Publications/getServerSideProps.spec.js
```diff
@@ -0,0 +1,47 @@
...
+      const { props } = await getServerSideProps({})
+      expect(props).to.deep.equal({
+        error: null,
+        publicationsData: []
+      })
+    })
+
+    it('should handle API errors', async () => {
+      var errorMsg = 'failed to connect to API'
+      var errorPromise = Promise.reject(new Error(errorMsg))
```
srallen, line 38: Non blocking: any reason `var` is being used here instead of `const`?

packages/app-content-pages/src/screens/Team/getServerSideProps.spec.js
```diff
@@ -0,0 +1,47 @@
...
+      const { props } = await getServerSideProps({})
+      expect(props).to.deep.equal({
+        error: null,
+        teamData: []
+      })
+    })
+
+    it('should handle API errors', async () => {
+      var errorMsg = 'failed to connect to API'
+      var errorPromise = Promise.reject(new Error(errorMsg))
```
srallen, line 38: Same question about use of `var`

srallen:
@eatyourgreens could you also follow up with opening an RFC with your thoughts about a sandbox next.js app that better helps us test features before applying to the project app? Thanks.

packages/app-content-pages/src/screens/Publications/getServerSideProps.spec.js
eatyourgreens: I wondered about that too.

PR #1636, opened 2020-04-30
Subtask stories drawing tools

srallen:

Package: lib-classifier

Describe your changes:
I refactored the point mark stories to display the mark on top of a subject image and use the svg context, so the mark is movable and we can have a story that shows a subtask with it. The long term goal is to have a story with the transcription task's transcription line tool when the subtask is open.

I used a `setTimeout` to work around the svg ref being null. I still don't know what's missing for the context of the story to get this to work right away or why, even more strangely, the component needs about a second to get the ref to be defined. This gets the story to work, but this is a brittle workaround.

I also, in a brittle way, just guesstimated the position for the subtask based on the current behavior for the snapshot for the stores at the start. The actual positioning is determined from the DOM node... and I didn't want to go down that rabbit hole again. 

Check it out:

point-mark-story

If this is looking good, I'll update in a subsequent PR the rest of the mark stories. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/plugins/drawingTools/components/Point/Point.stories.js
```diff
@@ -1,25 +1,143 @@
...
+    color: zooTheme.global.colors['drawing-red'],
+    type: 'point'
+  }],
+  type: 'drawing'
+})
+
+const subTasksSnapshot = [
+  {
+    instruction: 'Name your favourite fruit.',
+    taskKey: 'T0.0',
```
wgranger, line 37: Is this the new way taskKeys are structured in the monorepo, as a decmial?

packages/lib-classifier/src/plugins/drawingTools/components/Point/Point.stories.js
```diff
@@ -1,25 +1,143 @@
...
+  }
+
+  mockStores.classifications.createClassification(subject, workflow, project)
+  mockStores.workflowSteps.activeStepTasks[0].updateAnnotation()
+  if (activeMark) {
+    mockStores.workflowSteps.activeStepTasks[0].setActiveMark(point.id)
+  }
+
+  if (subtask) {
+    mockStores
```
wgranger, line 100: What is happening here? Is this conditional needed in an upcoming PR?

packages/lib-classifier/src/plugins/drawingTools/components/Point/Point.stories.js
srallen: Yes, check out [ADR 25](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-25.md)

packages/lib-classifier/src/plugins/drawingTools/components/Point/Point.stories.js
srallen: Oh this is cruft I think left over when I was trying to figure out how to implement the story. Good catch

srallen:
@wgranger I think at least two of those are already documented in issues. I'll double check before requesting re-review again. In the meantime, I'm going to switch this back to WIP because I'm refactoring the stories for the other tools now to use the new story format. I know I said I'd do that in a follow up, but decided to go for it. The transcribed line stories need a bit more work than the others, so this is what is still WIP. The other stories are all review ready now.

srallen:
Actually I'm going to re-request review., @wgranger. The stories for the transcription line don't make sense, but they work, so I'll work on that separately.

packages/lib-classifier/src/plugins/drawingTools/components/Ellipse/Ellipse.stories.js
```diff
@@ -0,0 +1,166 @@
...
+  mockStores.classifications.createClassification(subject, workflow, project)
+  mockStores.workflowSteps.activeStepTasks[0].updateAnnotation()
+  if (activeMark) {
+    mockStores.workflowSteps.activeStepTasks[0].setActiveMark(ellipse.id)
+  }
+
+  return mockStores
+}
+
+class DrawingStory extends Component {
```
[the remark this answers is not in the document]
srallen: Yeah could do in a follow up while working on the transcription line story.

PR #1638, opened 2020-04-30
Subject Group Annotations

shaunanoordin:
## PR Overview: New + Updated Components

Package: `lib-classifier`
Follows #1556 
Related to #1627

This PR allows the Subject Group Viewer (a custom Subject Viewer) to accept annotations from the user. Users are now able to mark which cells in the grid are of interest to the Task's question.

Screen Shot 2020-04-30 at 23 25 39

_Subject Group Viewer. The subject viewer (+annotations) can be seen on the left, the Subject Group Task (which just shows the task question) can be seen on the right. The subject viewer shows a 3x3 grid; the middle 3 cells have a teal border indicating that the user has selected (clicked on) them ; the middle-bottom cell has a dashed white border indicating the user's keyboard focus._

- The **SubjectGroupViewer** (custom Subject Viewer) component has been updated to allow for **annotations.**
  - That's right, like with a drawing task, the viewer is both a component to VIEW the Subject and the MAKE ANNOTATIONS.
  - Users can select (annotate) cells by clicking on them with a 🖱️ mouse or 👆 touchscreen pointer OR by tab-navigating to the cell then pressing Enter on the ⌨️ keyboard... 
  - ...but ONLY when the Subject Viewer is in Annotate mode. (i.e. when the ImageToolbar in the middle has the pointy arrowhead icon activated)
  - Users can continue to zoom and pan the cell images (see #1556) but ONLY when the Subject Viewer is in Move mode. (i.e. when the ImageToolbar in the middle has the 4-directional icon activated)
- Our stores have been updated to recognise **SubjectGroupAnnotations.**
  - An annotation for workflows using the Subject Group Viewer should list all the cells that the user selected, similar to a MultipleChoiceAnnotation. 
  - Ideally, this annotation should list the subject IDs of each cell (because each cell is supposed to represent a different single image subject)... ⚠️but we haven't set up actual Subject Group-type subjects in the API yet.
  - So, for the moment, annotations are in the form of `[ selected-cell-index, ... ]`, so for the example 3x3 grid above where the middle row is selected, the annotation is submitted as `[3, 4, 5]`. 
- The Task panel now recognises the (incredibly plain) **SubjectGroupTask.**
  - This Task initialises the annotations in the store (as with most Task component sub-types) and displays the Task question, but doesn't really have any other form of interactability

### Status

Ready for review? Sure, ready for review.


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -238,9 +308,16 @@ class SubjectGroupViewerContainer extends React.Component {
       return null
     }
 
+    // Note: the Task's Annotations are initialised by the SubjectGroupTask component.
+    const annotation = toJS(classification?.annotation(currentTask))
```
shaunanoordin, line 312: Problem: changes to annotations aren't being observed properly.

- Annotations are changed by the SGVGridCell calling it's parent's parent's `SubjectGroupViewerContainer.toggleCellAnnotation()` function.
- The stores are updated correctly (i.e. if I submit a Classification, the annotations are correct). 
- However, the visuals don't reflect changes to the annotation unless another change is triggered (e.g. switching the Subject Viewer from annotate mode to view mode)
- Basically, the `SubjectGroupViewerContainer`'s `render()` method doesn't trigger when the annotation changes.

Solution: `toJS()` somehow works.

Solution's Problem: we generally try to avoid using `toJS()` in our code.

Other solutions tried:

- Wrapping `SubjectGroupViewer` in `observer`. Result: click-to-annotate causes the correct visual change, but drag-to-pan function breaks.
- Wrapping `SGVGridCell` in `observer`. Result: no change

Other solutions considered:

- Wrapping the `annotation` in an `observable`?

beckyrother:
👋 Is this ready for design review as well? Happy to take a closer look if it's ready

eatyourgreens:
>Annotations are changed by the SGVGridCell calling it's parent's parent's SubjectGroupViewerContainer.toggleCellAnnotation() function.

I'm going to look at this today. My initial impression is that one component invoking another component's methods like this might be considered A Bad Thing in React. I'm not sure.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
eatyourgreens: My initial thoughts: The appearance of `SVGGridCell` should reflect its component state (selected/not selected), which is toggled by clicking on it. Think of the grid cell as being a HTML `checkbox` element. Changes in component state are passed to the annotation value via `annotation.update`.

eatyourgreens: I think the grid cell itself can have `role="checkbox" aria-checked=(true|false)`.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
```diff
@@ -54,6 +68,17 @@ function SGVGridCell (props) {
   const imageY = (cellHeight - imageHeight) / 2
 
   const clipPathID = `subjectGroupViewer-clipPath-${index}`
+
```
eatyourgreens, line 71:
```suggestion
  const [checked, setChecked] = useState(cellAnnotated);
```
Initialise the `checked` state of each cell here.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
```diff
@@ -79,14 +104,31 @@ function SGVGridCell (props) {
           transform={`scale(${zoom}) translate(${panX}, ${panY})`}
           transform-origin={`${imageWidth/2}px ${imageHeight/2}px`}
         />
+        <rect
+          fill="none"
+          stroke={(cellAnnotated) ? cellStyle.highlight : cellStyle.stroke}
```
eatyourgreens, line 109:
```suggestion
          stroke={(checked) ? cellStyle.highlight : cellStyle.stroke}
```

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
```diff
@@ -79,14 +104,31 @@ function SGVGridCell (props) {
...
+          strokeWidth={(cellAnnotated)
```
eatyourgreens, line 110:
```suggestion
          strokeWidth={(checked)
```

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
```diff
@@ -1,10 +1,20 @@
 import React from 'react'
```
eatyourgreens, line 1:
```suggestion
import React, { useState } from 'react'
```

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
```diff
@@ -79,14 +104,31 @@ function SGVGridCell (props) {
...
+          strokeWidth={(cellAnnotated)
+            ? cellStyle.highlightWidth * BORDER_MULTIPLIER
+            : cellStyle.strokeWidth * BORDER_MULTIPLIER
+          }
+          width={cellWidth}
+          height={cellHeight}
+        />
+        {annotationMode  && (
+          <ClickableRect
+            tabIndex={0}
```
eatyourgreens, line 119:
```suggestion
            tabIndex={0}
            role='checkbox'
            aria-checked={checked}
            aria-label={`Subject ${index}`}
```
I think the subject ID might be a better label here but array index will work too.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
```diff
@@ -79,14 +104,31 @@ function SGVGridCell (props) {
...
+            fill="transparent"
+            width={cellWidth}
+            height={cellHeight}
+            onClick={() => {
+              toggleCellAnnotation(index)
```
eatyourgreens, line 124:
```suggestion
              setChecked(!checked)
              toggleCellAnnotation(index)
```

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
```diff
@@ -79,14 +104,31 @@ function SGVGridCell (props) {
...
+            }}
+            onKeyPress={(e) => {
+              if (e.key === 'Enter') toggleCellAnnotation(index)
+            }}
```
eatyourgreens, line 128:
```suggestion
            onKeyPress={(e) => {
              if (e.key === 'Enter') {
                setChecked(!checked)
                toggleCellAnnotation(index)
              }
            }}
```

eatyourgreens:
Pan and zoom with the keyboard should work in drawing mode. See [Seabird Watch in the new classifier](https://frontend.preview.zooniverse.org/projects/penguintom79/seabirdwatch/classify) for an example of this. Testing in Chrome, I'm not able to zoom with the keyboard while selecting cells, so I don't think cells are receiving focus after being clicked.

srallen:
Maybe the task could be `subjectGroupComparison`? That is what the task is, yes?

srallen:
A more verbose task naming gives us potential flexibility for another task type with this viewer type I think. If there's a new task type in the future, it could be `subjectGroupX` as the naming convention?

eatyourgreens:
Subject selection task, maybe? Although I like @srallen's suggestion of using `SubjectGroup` as a prefix for tasks that are specific to subject groups.

srallen:
As another aside, it'd be great if we can get some of the (proposed?) tech choices for the `SubjectGroupViewer` and its associated task documented up in an ADR. Definitely can be done separately as this is out of scope of this PR and I know some functionality may change once the back end is in place.

shaunanoordin:
Dev work on this has been resurrected, with current items to address:

- [ ] Redesign SGVGridCell to act more like checkboxes
  - [ ] Can't use HTML <input type="checkbox"> in SVG, so use `useState()` instead.
  - [ ] Use aria-labels where possible
- [ ] Check the issue where everything crashes on a poorly-configured workflow.
  - `http://localhost:8080/?env=staging&project=1900&workflow=3412` works fine - this is the project I've been testing against and is configured correctly
  - `http://localhost:8080/?env=staging&project=1900&workflow=3402` crashes - this uses the Subject Group VIEWER, but its TASK isn't a Subject Group Task. (It's a basic single question task) ⚠️ This should be addressed in case project owners want to ask multiple questions, outside of the initial plan for the Survos questions.
- [ ] Check that keyboard input is working properly
  - Users should be able to navigate/pan image using arrow keys while in annotation mode.
- [ ] Discussion: naming conventions?
  - Sarah proposed `subjectGroupComparison` as the Task type's name.

shaunanoordin:
The error encountered while testing Workflow 3402 made me realise that my scope for what the Subject Group Task should do is perhaps too narrow. At the moment the scope of the Subject Group (Comparison?) Task workflow looks like this:

```
workflow.tasks = [
  Task 1 of exactly 1: "Mark all cells that look different"
]
```

The current design should be expanded to accommodate something like...

```
workflow.tasks = [
  Task 1: "Mark all cells that look different"
  Task 2: "Did any of the cells look yellow? Y/N"  // Basically, any non-SubjectGroupTask kind of task.
]
```

...though would the annotations made in Task 1 _visually persist_ when answering Task 2? Because that leads to a more complicated example:

```
workflow.tasks = [
  Task 1: "Mark all cells that look red"
  Task 2: "Mark all cells that look yellow"
  Task 3: "Mark all cells that look blue"
]
```

...what if there's a project that requires MULTIPLE DIFFERENT SubjectGroupTask tasks? This might be overthinking the scope a bit, but it's worth thinking about before I paint this component into a corner. At the moment, I can implement a solution provided _annotations don't need to be persisted across Tasks._

shaunanoordin:
## PR Update

I've done as much as I can to take care of the remaining bits since the last PR update in June, but I'm still having issues with one particular warning.

Screen Shot 2020-11-18 at 21 06 52
_New visual style: thin white border = cell has focus. Thick teal border = cell is selected_

- Keyboards fixed 👍  Users can now use arrow/+/- keys to move the Subject image(s)
- Visual styles updated so there's less faff. (See image above.)
- A significant warning in the dev console was fixed - I made the mistake of putting a `styled()` call inside `SGVGridCell`'s render().
- The way the SubjectGroupViewer pulls annotations has been refactored to better resemble what's used in `InteractionLayer` (mostly moving the logic up to the store mapper instead of inside the render) but...
- ⚠️ That One Warning still exists - see below
- [minor] Tests fixed

Tested with macOS 10 + Chrome on `https://local.zooniverse.org:8080/?env=staging&project=1900&workflow=3412`, which is a project set up specifically for the Subject Group Viewer. Baseline functionality continues to work:

- Users can view a Subject Group on the viewer
- Users can navigate (pan or zoom) the Subject Group either by 1. click-dragging and mouse-scrolling, or 2. using the keyboard (arrows/+/- keys). Each image in each cell is panned/zoomed independently.
- Users can select cells (i.e. mark a cell as an item of interest) either by 1. clicking on the cell, or 2. tab-navigating across each cell and pressing the Enter key
- Users can submit classifications, and annotations for SubjectGroupAnnotations are in the form of an array of selected indices, e.g. `annotation.value: [ 3, 7, 10 ]`

### That One Warning

There is a React warning whenever the Subject Group Viewer loads up:

```
Warning: Cannot update a component (`Observer`) while rendering a different component (`DecoratedTasks`). To locate the bad setState() call inside `DecoratedTasks`, follow the stack trace as described in https://fb.me/setstate-in-render
```

This can be traced directly to this bit:

```
SubjectGroupViewerContainer.js

  ...  
  const allAnnotations = classification ? Array.from(classification.annotations.values()) : []  // HERE
  const annotation = allAnnotations.find(item => (getType(item).name === 'SubjectGroupAnnotation'))
  ...
}
```

Any attempt to get `classification.annotations.values()` is considered problematic by React? I'm genuinely unsure of why this is the case. If anybody has insights into this, it'd be greatly appreciated.

### Status

I've been unable to make any further progress with That One Warning, but I'd still like to get this PR re-reviewed and merged before it further drifts from `master` and all the extant functionality is lost.

shaunanoordin:
## PR Update (small one this time)

I forgot to address one more thing from previous discussions:

- `SubjectGroupTask/Annotation` has been renamed to `SubjectGroupComparisonTask/Annotation`
- Example workflow (staging) 3412 has appropriately been updated to its Task type is now `"subjectGroupComparison"`
- Note that that VIEWER itself remains a generic SubjectGroupViewer; only the Task and Annotation name have been changed.

As @srallen noted, having a `SubjectGroup` prefix provides more flexibility for future task types.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -39,15 +45,41 @@ function storeMapper (stores) {
...
+  
+  const isCurrentTaskValidForAnnotation = !!currentTask
+
+  // Note: the Task's Annotations are initialised by the SubjectGroupComparisonTask
+  // component. However, do note that it's possible to have a
+  // SubjectGroupViewer without a SubjectGroupComparisonTask.
+  
+  // Currently only supports 1 Subject Group Task in the workflow.
+  const allAnnotations = classification ? Array.from(classification.annotations.values()) : []
+  const annotation = allAnnotations.find(item => (getType(item).name === 'SubjectGroupComparisonAnnotation'))
```
eatyourgreens, line 67: If you're just starting the task, then `classification.annotations` won't contain an annotation for this task yet. So `annotation` might be undefined.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
```diff
@@ -54,6 +79,25 @@ function SGVGridCell (props) {
...
+    const hasCellIndex = annotationValue.includes(index)
+    
+    if (hasCellIndex && !toggledValue) {  // Remove cell index from annotation values
+      const indexInValue = annotationValue.indexOf(index)
+      annotationValue.splice(indexInValue, 1)
+    } else if (!hasCellIndex && toggledValue) {  // Add cell index to annotation values
+      annotationValue.push(index)
+    }
+    
+    if (annotation?.update) annotation.update(annotationValue)
```
eatyourgreens, line 99: If `annotation` is undefined, then we can't store a new value here. That makes logical sense, but creates a case where a task is rendered but never saves annotations. I think it would be better to make `annotation` required, not optional.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
eatyourgreens: Comparing this with the drawing task code, maybe use `.filter` instead of `.find` here in case there's more than one subject group task in the workflow?

eatyourgreens, line 67:
```suggestion
  // Currently only supports 1 Subject Group Task in the workflow.
  const annotation = classification?.addAnnotation(currentTask)
```
This might be a safer way to get the annotation for the current task, without throwing React warnings.

eatyourgreens: `classification?.annotation(currentTask)` might work too. I think there's an `annotation(task)` view that gives you back the annotation for a specific task (or possibly a specific task key.)

eatyourgreens:
After reading comments on https://github.com/facebook/react/issues/18178#issuecomment-595846312, the React warning, when the first subject loads, might be pointing to a `setState` bug in `Observer`. I'm not sure. Accessing the annotation through a classification view seems to fix it for me, locally.

eatyourgreens (approved):

beckyrother:
Hi Shaun, those outlines are looking quite thick. (Sorry if this isn't the place for design feedback)  
I'd shared an InVision link a while back (https://projects.invisionapp.com/d/main#/console/12923997/407098812/preview) showing what a selected and active image might look like – is that doable?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
shaunanoordin: re: use of `.find` -

This line originally used a variant of the `.filter` code from InteractionLayer (`const [annotation] = allAnnotations.filter(item => (getType(item).name === 'SubjectGroupComparisonAnnotation'))`, to be exact) but I thought `.find` would be easier to read since I was purposely constraining the component to 1 Subject Group Comparison Task per workflow. I might need to rethink this if `.filter` makes more sense to others.

(Note: supporting multiple possible Subject Group X Tasks would require a mild rethink of the UI - e.g. colour markings - and I didn't want to jump into that before I had a proper use case.)

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
shaunanoordin: Ah! That scenario has already been accounted for - what if a Workflow wants to use the Subject Group Viewer but not actually use a Subject Group Comparison Task? (i.e. no `annotation`)

Example: https://local.zooniverse.org:8080/?env=staging&project=1900&workflow=3402 (`workflow.config.subject_viewer = 'subjectGroup'; T0.type = 'single'`)

Screen Shot 2020-11-18 at 21 06 52
_Example of a Subject Group Viewer with a Single Choice Task_

In this scenario, the SGVGridCell continues to render individual images and allow users to pan the image (mouse/keyboard), without responding to any click-to-select-cell (or keyboard equivalent).

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
shaunanoordin: Unfortunately, `classification?.annotation(currentTask)` doesn't work either. 😞 See comment: https://github.com/zooniverse/front-end-monorepo/pull/1638#issuecomment-642219935 (sorry, it's kinda buried under so many revisions!)

I'll experiment with `const annotation = classification?.addAnnotation(currentTask)`, thank you for the suggestion. Two things for me to note:
- At the moment, the SubjectGroupViewer assumes that if there's a SGComparisonTask, it'd be initialised by the Task component. (`TaskArea/components/Tasks/Tasks.js`) Gotta check there's no issue with double-initialisation, if that's such a thing.
- Check that the `addAnnotation` doesn't trigger if there's a non-SGComparisonTask on the right-hand panel. (e.g. a Single Choice Task with an SGViewer.) (Shouldn't be the case if I remember to set currentTask = undefined for non-SubjectGroupXTasks)

shaunanoordin:
@beckyrother This is absolutely the right place for design feedback! I'll try to adapt the component to your feedback as best as I can, but if it turns out to be a big change, I'll merge this PR first and spin up a new one dedicated to a design update, I hope you don't mind.

- ⚠️ Important background info: the Subject Group Viewer has a set of _default styles_ that I'm updating to reflect your Invision design
  - However, these styles CAN be potentially overridden by a project owner (not that it's easy to do since there's no PFE Project Builder code for it yet) if, e.g. they need a different border/overlay colour to contrast their Subject images.

- I added a thin white border to every unselected cell to provide visual differentiation. I noticed in your Invision design, the "idle" cells have no borders but are easy to distinguish because each image has wildly different colour palettes. The example data we're working with is more, er, grey. Please see Comparison 1.

Screen Shot 2020-11-19 at 23 00 29

_Comparison 1: if I set no border for unselected cells, it's very very hard to tell them apart._

- I can't use your exact border pixel sizes at the moment, sorry. There's a lot of reasons for this - the SVG scales according to the available webpage space, the arbitrary number of rows/cols/cell size/subject size, SVG rects only showing half the stroke-width due to layout, etc - and that just basically means there's a lot of UI maths I need to compensate for that in a future update.
  - For now, when I see you use "border: 1px ...", I'm taking it to mean _very thin border,_ and "border: 5px ...", that's 5x whatever the thin border was. I hope this is OK for now.

--------

OK, putting all that together, this is how things look like now:

- Unselected cells = very thin (~0.5px visually) ⚪  white border
- ☑️  Selected cells = thin (~1px visually) 🔵  blue border + 🔵 blue tint
- Unselected cells + ⌨️ keyboard focus = thick (~5px visually) 🔵 blue border
- ☑️  Selected cells + ⌨️ keyboard focus = thick 🔵 blue border + 🔵 blue tint

(Blue = #2967FF, blue tint = rgba(41,103,255,0.3), "selected" means user has marked the cell like an annotation, "keyboard focus" means either that user tab-navigated to the cell or the user just clicked the cell)

Screen Shot 2020-11-19 at 23 33 00

_Screen 1: two selected cells, plus one unselected cell that has keyboard focus_

Screen Shot 2020-11-19 at 23 33 07

_Screen 2: three selected cells, the bottom-most which has keyboard focus._

For funsies, I also decided to test how the UI would look like when there's a whole cluster of selected cells.

Screen Shot 2020-11-19 at 23 33 22

_Comparison 2: nine selected cells, in a box. Can you see which cell has keyboard focus?_

Please let me know if I'm misreading your Invision design, or missed something obvious!

beckyrother:
Shaun this looks awesome! Thank you for taking the time. Definitely understand about the border size differences – I think this works really well. The white borders add a lot with this type of data too, thank you.

shaunanoordin:
Sweet! 👍 Much thanks to Becky, Jim, and Sarah for all the feedback on the Subject Group annotations, from code to design - this wouldn't have worked without y'all.

### Status

Merging! 🎉

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
eatyourgreens: Oh wait, you want to continue to display the annotation during subsequent tasks, don't you? My suggestion would only display it while we're doing the subject selection task. It would vanish on moving to the next task in the workflow. I missed that because I only looked at one task in the test workflow, so the current task was always a subject group task.

PR #1643, opened 2020-05-06
Refactor to reset pan, zoom, or rotate on new frame

mcbouslog:
Package: lib-classifier

Towards #1538, however, this refactors for transcription workflows. **Conditionally setting pan, zoom and rotation based on task or workflow is still needed.**

Describe your changes:
- refactors SVGPanZoom to reset on new image
- refactors SubjectViewerStore to reset rotation in `onSubjectReady`

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.js
@@ -177,13 +177,14 @@ class MultiFrameViewerContainer extends React.Component {
         />
         <SVGContext.Provider value={{ svg }}>
           <SVGPanZoom
-            img={img}
+            img={this.subjectImage.current}
             maxZoom={5}
             naturalHeight={naturalHeight}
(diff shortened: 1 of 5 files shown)


mcbouslog:
> This is an interim step toward supporting the two different behaviors for multi-frame subjects and adds the functionality desired for transcription subjects which is to reset the pan/zoom between frames. Going forward, I think we'll have to rely on the new `subject_viewer_config` in workflow configurations for the setting on whether or not to reset. I don't think the code in the classifier should check for task type, but just check for a setting in that config. The reason is that not all transcription projects will want to explicitly use the transcription task, since that is designed for use with the downstream caesar configuration and reductions. We'll plan on the transcription task defaulting to reset and the lab will set this when the transcription task is enabled. Otherwise, transcription projects that don't use the transcription task will have to set this config and the overall default will be the PFE function which is to not reset between frames. Does this sound good?

Yep sounds good to me, consistent with https://github.com/zooniverse/front-end-monorepo/issues/1538#issuecomment-616732071.

I'll merge #1642 and this PR today, leaving #1538 open to implement a default of not resetting for non-transcription task workflows and resetting for transcription task workflows.

PR #1648, opened 2020-05-11
Reorganize task stories. Make TranscriptionLine task type flexible for color

srallen:

Package: lib-classifier

Describe your changes:
Reorganizes the stories to be closer to their code and enables us to add multiple stories per task. Refactored `TranscriptionLine` to be task flexible as we want to allow project builders to use the tool type in either the drawing task or transcription task. The colors are only hardcoded and have meaning for the transcription task, so this adds a `color` prop back to the `TranscriptionLine` and set the default state to be an empty string. The `state` prop will only be set by the TranscriptionTask or TranscribedLines models in a follow up PR.

As an aside, the task stories refactor and the TranscriptionLine refactor don't really have to be in the same PR, so happy to split out if requested. I just got there because I like starting with stories and the organization was just confusing previously. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.stories.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.stories.js
@@ -62,7 +62,7 @@ rootStore.workflows.setActive(workflowSnapshot.id)
 rootStore.subjects.setResource(subjectSnapshot)
 rootStore.subjects.setActive(subjectSnapshot.id)
 
-storiesOf('TranscribedLines', module)
+storiesOf('Drawing Tools | TranscribedLines', module)
   .add('default', () => (
     <Provider classifierStore={rootStore}>
       <Grommet theme={zooTheme}>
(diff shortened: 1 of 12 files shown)


eatyourgreens:
One initial thought about this: plugins should be independent of the classifier. So, ideally, plugins shouldn’t be importing components from the classifier. Importing `Tasks` and classifier stores is going to make it difficult to extract tasks into their own library.

srallen:
Sure, perhaps further reorganization can be done in future PRs. My refactor here isn't changing how these stories are being built, just where they are in the component structure which enables us to have multiple stories about specific task types more clearly and places them next to the relevant code. Several of the components need decoupling from the inject decorator and the upgrade to use React context directly as suggested in mobx-react v6 will help with that.

eatyourgreens:
Sounds good. Maybe tasks should export mocks that can be used in `Tasks` stories? That would keep the `Tasks` code separated from individual tasks. Tasks themselves could also have stories that don’t show the task navigation and task help. I think the second one depends on whether we use tasks in projects outside of the classifier. 

Either way, it’s something to consider but not a requirement for this PR. 

From a quick look through the code on my phone, the container for individual task stories could be put into a shared component, which would keep all the external imports in one place. Looks good, though. 👍🏻

srallen:
@eatyourgreens Yeah, I had similar thoughts on future refactoring of the stories. Generally speaking, I think we should make sure we organizationally keep stories next to the component code. I was struggling to develop against task stories that weren't along side, so I decided to make this change even though it's not immediately relevant to the other half of the PR. 

@mcbouslog no worries on the story breakage. We do our best to keep these working and useful for us in developing! Thanks!

PR #1651, opened 2020-05-13
Hide Previous Transcription Marks

### LINKED ISSUE #1278: Hide previous marks
There's a feature for being able to toggle the visibility of every mark up to the most recent one.

wgranger:

Package: 
lib-classifier

Closes #1278 

Describe your changes:
This PR allows the three-way toggle of transcription marks: all marks, user marks, no marks. This is done by converting the `hidePreviousMarks` value in the drawing and transcription tasks into a three-item enum, which is kept in the helpers directory. 

Tagging @beckyrother for some design input, as I took some liberties with the transcription text and icons. The text and icons remain the same for a drawing task.
toggle

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HidePreviousMarksButton.js
```diff
@@ -1,23 +1,38 @@
...
+    default:
+      text = isDrawingTask ? counterpart('HidePreviousMarksDrawingButton.show')
+        : counterpart('HidePreviousMarksTranscriptionButton.show')
+      icon = <FormViewHide />
+  }
 
   return (
     <MetaToolsButton
-      aria-checked={hidePreviousMarks}
+      aria-checked={shownMarks}
```
eatyourgreens, line 35: I think this is setting `checked` state to a number, but it can only be true/false.

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HidePreviousMarksButton.js
```diff
@@ -1,23 +1,38 @@
...
       disabled={disabled}
       icon={icon}
       role='checkbox'
```
eatyourgreens, line 38: Now that there are multiple states to choose from, this is emulating a select menu or radio group, rather than a checkbox.

wgranger: For accessibility, is the best way to go about this to use `role='menu'` and then remove `aria-checked` in favor of... `aria-label`? I would think the aria item should tell what the current state is: "All Marks", "User Marks", "No Marks".

eatyourgreens: Buttons can’t have the `menu` role, because a menu has to wrap a list of `menuitem` roles.

The button label should tell you what the button does when you press it. I don’t think anyone will understand the label as being the current state. That sounds confusing. 

Button state can be either `pressed` or `not-pressed`. There’s also a `switch` role but that’s either on or off. 

I’m having trouble finding examples of buttons that have more than two states. The closest I can think of would be a group of 3 radio buttons, one of which is selected. However, the expected interaction for radio buttons would be that you change the selected state with the cursor keys.

wgranger: It looks like there is also an [indeterminate state](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Indeterminate_state_checkboxes) for checkboxes that can be set via Javascript. I can try using this with the label being "Show All Marks," "Show User Marks," and "Hide All Marks."

wgranger: Actually, I'm starting to think of reasons this would still conflict with the label. I'll have to think of another solution.

wgranger:
I'm not sure if this is the best fix, but I've removed the `role='checkbox'` designation and used `aria-live: assertive` to notify of changes to the button label. I don't think using a checkbox with `aria-pressed: true|false|mixed` is the best solution considering the inner text changes.

wgranger:
@eatyourgreens Do you believe the implementation I have above would work in lieu of a checkbox?

srallen:
Based on the discussion we had about the accessibility for this component, it might make the most sense to have two separate components now, one as a button for the drawing task version, and one for the menu or select for the transcription task version.

wgranger:
Pinging @beckyrother 

hideTranscriptions

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.spec.js
```diff
@@ -133,5 +136,24 @@ describe('Component > InteractionLayerContainer', function () {
...
+            height={height}
+            width={width}
+            activeInteractionTask={hidingUserMarksInteractionTask}
+            interactionTaskAnnotations={drawingAnnotations}
+            workflow={{ usesTranscriptionTask: true }}
+          />
+        )
+
+        expect(wrapper.find(DrawingToolMarks)).to.have.lengthOf(0)
+      })
```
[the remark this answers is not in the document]
wgranger: Nothing that I see. That test needs to be adjusted a bit to make sure user marks are visible.

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HidePreviousTranscriptionsButton.js
```diff
@@ -0,0 +1,87 @@
...
+  div {
+    margin-top: 10px;
+    padding: 0;
+  }
+
+  span {
+    ${props => css`line-height: ${props.theme.paragraph.small.height};`}
+  }
+
+  &:hover {
```
srallen, line 24:
```suggestion
  &:focus,
  &:enabled:hover {
```

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HidePreviousTranscriptionsButton.js
```diff
@@ -0,0 +1,87 @@
...
+
+  const NONE = {
+    icon: <Hide />,
+    label: <StyledText margin={{ vertical: 'auto' }} {...props}>{counterpart('HidePreviousTranscriptionsButton.hide')}</StyledText>,
+    onClick: () => onClick(SHOWN_MARKS.NONE)
+  }
+
+  let items = new Map([['ALL', ALL], ['USER', USER], ['NONE', NONE]])
+  const current = items.get(shownMarks)
+  items.delete(shownMarks)
```
srallen, line 62: Even if we add the appropriate aria roles and checked states, removing the actively selected item from the menu options means that we won't be communicating which one is selected anymore.

beckyrother:
@wgranger so sorry I missed your message from last week. Do you need different icons for 'show your marks' and 'hide all marks'?

wgranger:
Updated the component to use the Grommet `DropButton` for a couple reasons. The `Menu` component doesn't have great accessibility because the button wraps the children options in divs that are unable to accept the `role=radio` designation. Lighthouse doesn't like this and complains with a low accessibility score.

`DropButton` gives a bit more flexibility with customizing the options, so I am able to assign the children and all options the correct roles that Lighthouse likes and passes an accessibility audit. This meant a bit more code, but I think it's worth it.

Also, @beckyrother the "accent-4" background on an active item looks **very** jarring, so I decided to use the default active button background that I believe is with our theme and is reflected in the screenshot below:
Screen Shot 2020-06-17 at 2 16 36 PM

beckyrother:
@wgranger good call on the color, the grey works better.

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HideTranscriptionButton.js
```diff
@@ -0,0 +1,59 @@
...
+      disabled={disabled}
+      gap='xxsmall'
+      icon={icon}
+      label={
+        <SpacedText color={{ dark: 'accent-2', light: 'neutral-2' }}>
+          {title}
+        </SpacedText>
+      }
+      onClick={onClick}
+      pad='small'
```
srallen, line 38: This doesn't do anything with `plain` being true (see explaination: https://v2.grommet.io/button#plain). The padding is added by the use of styled-components on line 10.

The styling here is now almost identical to the `MetaToolsButton` from the component library. This is non-blocking, but if that shared component is updated to take a pad prop and set padding via styled-components, then I think we can replace this here with the `MetaToolsButton`. Perhaps this can be a minor refactor next?

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HidePreviousTranscriptionsButton.spec.js
```diff
@@ -0,0 +1,52 @@
...
+describe('Component > HidePreviousTranscriptionsButton', function () {
+  before(function () {
+    wrapper = shallow(<HidePreviousTranscriptionsButton />)
+  })
+
+  it('should render without crashing', function () {
+    expect(wrapper).to.be.ok()
+  })
+
+  it('should show the correct text and icon', function () {
```
srallen, line 19: The description here doesn't match what is being tested. Should maybe read `should show the correct aria title`?

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HidePreviousTranscriptionsButton.spec.js
```diff
@@ -0,0 +1,52 @@
...
+      <HidePreviousTranscriptionsButton
+        shownMarks={SHOWN_MARKS.ALL}
+      />)
+    const drop = wrapper.find(StyledDrop).first()
+    const { a11yTitle } = drop.props()
+    expect(a11yTitle).to.equal("Show All Marks")
+  })
+
+  describe('and showing user marks', function () {
+    it('should show the correct text and icon', function () {
```
srallen, line 30: Same comment about the description of this test vs the actual expectation. I think if we test for the displayed text and icon, you'll want to check the icon and label prop?

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HidePreviousTranscriptionsButton.spec.js
```diff
@@ -0,0 +1,52 @@
...
+          shownMarks={SHOWN_MARKS.USER}
+        />)
+      const drop = wrapper.find(StyledDrop).first()
+      const { a11yTitle } = drop.props()
+      expect(a11yTitle).to.equal("Show Your Marks")
+    })
+  })
+
+  describe('and showing no marks', function () {
+    it('should show the correct text and icon', function () {
```
srallen, line 42: Same comment about the description of this test vs the expectation.

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HidePreviousTranscriptionsButton.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  span {
+    ${props => css`line-height: ${props.theme.paragraph.small.height};`}
+  }
+`
+
+function HidePreviousTranscriptionsButton (props) {
+  const { disabled, onClick, shownMarks } = props
+  const [isOpen, setOpen] = React.useState(false)
```
srallen, line 28: I think there should be some expectations to test this state and hook.

packages/lib-classifier/src/plugins/tasks/experimental/TranscriptionTask/models/TranscriptionTask.spec.js
```diff
@@ -80,6 +81,17 @@ describe('Model > TranscriptionTask', function () {
...
+
+    it('should hide previous marks', function () {
+      expect(transcriptionTask.shownMarks).to.equal(SHOWN_MARKS.ALL)
+      transcriptionTask.togglePreviousMarks(SHOWN_MARKS.USER)
+      expect(transcriptionTask.shownMarks).to.equal(SHOWN_MARKS.USER)
+      transcriptionTask.togglePreviousMarks(SHOWN_MARKS.NONE)
+      expect(transcriptionTask.shownMarks).to.equal(SHOWN_MARKS.NONE)
+      expect(transcriptionTask.hidingIndex).to.equal(transcriptionTask.marks.length)
+      transcriptionTask.togglePreviousMarks(SHOWN_MARKS.ALL)
+      expect(transcriptionTask.shownMarks).to.equal(SHOWN_MARKS.ALL)
```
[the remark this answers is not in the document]
srallen: Just bumping this from the last review as part of my spec suggestion in the latest comments.

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HidePreviousTranscriptionsButton.js
```diff
@@ -0,0 +1,114 @@
...
+            a11yTitle={counterpart('HidePreviousTranscriptionsButton.show')}
+            active={shownMarks === SHOWN_MARKS.ALL}
+            aria-checked={shownMarks === SHOWN_MARKS.ALL}
+            icon={<FormView />}
+            margin={{ top: '0' }}
+            onClick={() => {
+              onClick(SHOWN_MARKS.ALL)
+              setOpen(false)
+            }}
+            padding='0.5em'
```
srallen, line 52: The `MetaToolsButton` is also a `plain` Grommet `Button`, so this isn't doing anything either. I may not have been clear about that. It'll have to be modified to take the `pad` prop and do something with it like apply styling via styled-components. This can be in a follow up PR?

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HidePreviousTranscriptionsButton.js
```diff
@@ -0,0 +1,114 @@
...
+            active={shownMarks === SHOWN_MARKS.USER}
+            aria-checked={shownMarks === SHOWN_MARKS.USER}
+            disabled={disabled}
+            icon={<FormViewHide />}
+            margin={{ top: '0' }}
+            onClick={() => {
+              onClick(SHOWN_MARKS.USER)
+              setOpen(false)
+            }}
+            padding='0.5em'
```
srallen, line 67: Same comment about padding here.

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/HidePreviousMarksButton/HidePreviousTranscriptionsButton.js
```diff
@@ -0,0 +1,114 @@
...
+            a11yTitle={counterpart('HidePreviousTranscriptionsButton.hide')}
+            active={shownMarks === SHOWN_MARKS.NONE}
+            aria-checked={shownMarks === SHOWN_MARKS.NONE}
+            icon={<Hide />}
+            margin={{ top: '0' }}
+            onClick={() => {
+              onClick(SHOWN_MARKS.NONE)
+              setOpen(false)
+            }}
+            padding='0.5em'
```
srallen, line 81: Same comment about padding here.

wgranger: My bad. There was a change to `MetaToolsButton` that wasn't included in this commit for whatever reason. I'll push it up now.

packages/lib-react-components/src/MetaToolsButton/MetaToolsButton.js
```diff
@@ -6,8 +6,13 @@ import PlainButton from '../PlainButton'
 export const StyledPlainButton = styled(PlainButton)`
   > div {
     justify-content: flex-start;
+    ${props => css`padding: ${props.padding};`}
```
srallen, line 9: Awesome! This was indeed missing, so the changes make more sense! Non-blocking suggestion, but the `MetaToolsButton` is missing a storybook story, readme, etc. Perhaps in a follow up PR, a story and a readme documenting its function and the props it accepts would be helpful. Thanks!

srallen:
@beckyrother is your last comment an approval? The old requested changes review needs clearing or an update to approval. Thanks!

packages/lib-react-components/src/MetaToolsButton/MetaToolsButton.js
wgranger: If it's missing a story and a bit more, I'll save this for a separate PR.

Issue #1625, opened 2020-04-20
Annotations from previous frames not saved with multi-frame subject


mcbouslog:
**Package**
lib-classifier

**Describe the bug**
Only the most recent frame annotations persist, previous frame annotations overwritten.

**To Reproduce**
1. Go to http://localhost:8080/?project=3496&workflow=13983&env=production
2. Draw some points on one frame
3. Go to a different frame, draw some more points
4. Click done and only recent frame annotations saved, not initial points or go back to initial frame and won't see first points

**Expected behavior**
Annotations on every frame persists and saves to classification.

mcbouslog:
I think current code is creating new InteractionLayer instance with each frame, when a single instance of the InteractionLayer should be part of MultiFrameViewer.

mcbouslog:
I think #1538 might be related in a sense. Similar to scale, for a survey workflow would we expect drawing marks to persist across frames? Clearly for a transcription type workflow we would not want the transcription lines to persist from page to page. Though regardless of workflow type, the markings should save, as this issue states.

eatyourgreens:
There's a workflow setting in the project builder, which toggles whether marks are unique to each frame or cloned across all frames. Solar Stormwatch is an example of a project where you make a different drawing on each frame, because they're interested in measuring changes.

Space Warps is an example of a project where your marks are shown across all frames. From what I remember, it causes problems with PFE's subtask popups. Maybe they only show up on the original mark, or maybe they open on all copies of a mark when you click just one? I don't remember.

Anyway, it might be worth toggling that option on/off in a test project, just to get an idea of how it works.

srallen:
Closed by #1658

PR #1658, opened 2020-05-20
Observe and react to subject advance

srallen:

Package: lib-classifier

Toward #1625

Describe your changes:
This refactors the action that triggers the workflow steps reset, classification creation, and feedback reset to use `advance` rather than `onSubjectReady`. I think this is a better fit as I don't think a failure of the onload event for the subject media should prevent the reset actions to follow. The failure of onload should stop someone from being able to fill out and submit a classification however, and I'm not sure if we're handling an onload event error in this way or testing for it.

What was key to make this work was to set a third parameter for the `onAction` call. [The MST docs](https://mobx-state-tree.js.org/API/#onaction), state the third `afterAttach` parameter is a boolean, which it isn't entirely clear what it does, but after [looking at the source](https://github.com/mobxjs/mobx-state-tree/blob/19012c2c0f50a8b89cecdd045cb5ce98db950287/packages/mobx-state-tree/src/middlewares/on-action.ts#L252), it becomes clear that this boolean controls the order of operations here. When set to false (which it defaults to), the action call that is intercepted by the `onAction` middleware happens after the `onAction` middleware function (the second parameter) is called. This is problem for this particular middleware because we need to have an active subject defined to be able to create the classification. I've set this boolean parameter to true and now this is working as expected: subject advances, then reset the workflow steps, create classification, and reset the feedback.

Except! A wrinkle (also from the MST docs about `onAction` and explained in this issue: https://github.com/mobxjs/mobx-state-tree/issues/1508):

> onAction events are emitted only for the outermost called action in the stack.

What this means for the first subject that is loaded, that the `advance` that the RootStore is listening for, although it is called, is not emitted to be intercepted by the `onAction` middleware. The first subject is initially set by first calling `populateQueue`, then `append`, and then `advance`, so we can't rely on `onAction` to react to the first call of `advance`. So, in the end, we have to write our own custom middleware using `addMiddleware`. I only caught this by manual testing, so we need better integration tests at some point, though, this PR is already big and I'm not sure on an approach on refactoring this. I'd like to save that for another time.

While refactoring the tests, which included moving the test for the on subject advance reset behavior for classifications to the `RootStore` specs since I think this is a better fit, I also refactored `setResource` to be `setResources` and be able to handle an array of resources, which is usually how the API provides them to us. The refactor to `setResources` may be helpful when we do a refactor for #1554 

This is a big change, so I'd like a couple of reviews trying out a variety of staging test projects.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
Apologies if anyone is mid-review of this. I was going to open a separate branch and PR, but I ended up decided to attach the new test to just this one since it was a single spec. I rebased and added the new spec to the `Mark.spec.js` file for testing for the correct frame index on the mark. Manual testing can also be done with this test project: http://local.zooniverse.org:8080/?project=1876&workflow=3405 and here's a screenshot of the submitted classification with the correct frames per mark:

Screen Shot 2020-05-27 at 3 59 53 PM

Issue #1660, opened 2020-06-01
Consensus UI


srallen:
**Package** lib-classifier

For the text transcription feature, the [user stories](https://docs.google.com/document/d/16abI-wkRlEXsWgACfFQVqwO76aEopohIjQiRfNQKWiw/edit) include

> As a user I want to be able to read completed (greyed-out) transcriptions.

This is an enhancement over the ASM version where there was just a tooltip that displayed text saying the line is complete. However, volunteers wanted to see which lines went into the retirement. The aggregation code has been updated to support this, so it's now just a matter of having a UI to display this.

In past in person meetings, we decided the UI should be similar to the sub-task UI, a movable modal and in the modal content area, the consensus text and the transcriptions that went into it would be displayed. We don't have a specific invision design for this, but we may not necessarily need one. We can coordinate with @beckyrother as questions arise.

beckyrother:
InVision attached: https://projects.invisionapp.com/d/main#/console/12923997/420788232/preview

Classifications are listed with clear text explaining what this text is and why it can't be interacted with

srallen:
Closed by #1683

Issue #1546, opened 2020-02-28
Drawing SubTasks - Follow up for PR 1517


shaunanoordin:
## UI Enhancement

Package: `lib-classifier`

In PR #1517, we introduced a basic subtask popup component for drawing tools. This issue acts as a handover document for further development of the Drawing SubTasks popup, focusing on the UI aspects of the component.

### Current Features

PR #1517 introduces the following features:

![Screen Shot 2020-02-28 at 18 48 20](https://user-images.githubusercontent.com/13952701/75577998-f5771500-5a5a-11ea-9f40-d1b1df90d8e3.png)

- If a drawing tool mark has subtasks, then when you finish creating a new one, a popup will open.
- If you click on an existing drawing tool mark with subtasks, a popup will open.
- The subtask popup functions the same as the right-hand Task Area.
  - All changes to the subtask values are **saved automatically.** (This is important to note because there's a 'Save' button in that UI that only closes the popup. _Not_ clicking the save button to close the popup doesn't revert any changes made.)
  - Classifications are correctly submitted with subtask values.
- When the subtask popup is open, interaction with other elements are blocked (e.g. other drawing marks on the Subject Viewer, the right-hand task panel).
  - There's an **invisible modal layer** blocking interaction.
  - Clicking outside the popup, or clicking the save button, or clicking the close (X) button, or pressing the Esc key, will close the popup.
- The selected/active drawing tool mark is now more visually distinguishable from other marks.

Basically, this allows users to classify on workflows that feature drawing tools with subtasks.

### Implementation Notes

- ⭐️ The popup itself uses [React-Rnd](https://github.com/bokuweb/react-rnd)
  - The `<Rnd />` popup component can be reposition and resized by the user.
  - It can do a lot more (see the docs), but for simplicity, I only chose to set the _default starting position_ and _minimum size_ of the popup.
- The **x-y coordinates** specify the top-left corner of the popup, relative to _its closes parent with position: relative|absolute._ (i.e. `<Rnd />` component has `position: absolute`)
  - Currently, the _default x-y coordinates of the popup_ has the top-left corner of the popup appear at the middle of the drawing tool mar.
- Figuring out the appropriate **size of the popup** can be tricky.
  - On one hand, you don't want to hardcode a popup size, so it can auto-calculate width/height to accommodate any number of subtasks.
[...]

PR #1661, opened 2020-06-01
Reusable movable modal

srallen:

Package: lib-react-components

Toward #1660

Describe your changes:
I'm adding modal component that is wrapped by react-rnd to the shared components library based on the sub-task pop up component. The eventual goal is that the consensus transcription line can also be clickable to display information about the consensus as described in #1660. This means reusing the work done for the subtask pop up. Eventually we will swap out the code for the sub-task pop up to use this reusable component and long term also apply it to the field guide since the original design had that be a movable modal too. 

~WIP~:
- [x] specs
- [x] readme
- [x] fixing the issue with the target container for the resizing (can be seen by running the storybook locally):

rnd-issue

The subtask pop up has positioning being calculated, so that may be necessary here. I'm unsure, and would appreciate advice about react-rnd's API. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
Can the box be moved and resized from the keyboard (when it has focus)?

srallen:
Focus and keyboard interaction is part of #1546

beckyrother:
If the modal can be resized, it should have an indicator on the corner. Here, resizing the text box resizes the entire modal, but the same behavior makes sense for the entire modal:
 
Classifier_2018_Preview_Mode_-_InVision

srallen:
@beckyrother the bottom right resize icon was there, but hard to see because the background wasn't filling the whole area. I have that fixed now thanks to @wgranger help. Is this ok?

working-resize

srallen:
I need to write the readme and a few more specs, so this should ideally be ready for full review sometime tomorrow.

packages/lib-react-components/src/MovableModal/MovableModal.js
```diff
@@ -0,0 +1,111 @@
...
+  const {
+    children,
+    closeFn,
+    headingBackground,
+    pad,
+    rndProps: {
+      minHeight,
+      minWidth,
+      position,
+      ...restRndProps
```
wgranger, line 18: Out of curiosity, where does this `...restRndProps` come from?

packages/lib-react-components/src/MovableModal/MovableModal.js
```diff
@@ -0,0 +1,111 @@
...
+        bottomLeft: false,
+        bottomRight: true,
+        left: true,
+        right: true,
+        top: true,
+        topLeft: false,
+        topRight: false
+      }}
+      default={position}
+      minHeight={minHeight}
```
wgranger, line 38: I think this may need to be adjusted a bit to ensure the `Modal` doesn't become too small. However, it could just be how to storybook is rendering this.

packages/lib-react-components/src/MovableModal/MovableModal.js
```diff
@@ -0,0 +1,111 @@
...
+          bottom: 0,
+          left: 0
+        },
+        bottomRight: {
+          bottom: 0,
+          height: '20px',
+          right: 0,
+          width: '14px'
+        },
+        left: {},
```
wgranger, line 53: I believe these empty objects can be removed for the sides that are not being used in the `resizeHandleStyles`.

srallen:
@beckyrother there's no guarantee the content inside resizes. The inner containers have to have the right CSS applied to them for that. In that particular story, they aren't, but I could include in the readme notes to make sure that's done, i.e. use flex grow/shrink CSS to inner flex boxes?

beckyrother:
@srallen yes, that's a good idea

packages/lib-react-components/src/MovableModal/MovableModal.spec.js
```diff
@@ -0,0 +1,106 @@
...
+      )
+      expect(wrapper.find(Rnd)).to.have.lengthOf(1)
+    })
+
+    it('should set the min-height, min-width, and position by props', function () {
+      const minWidth = 800
+      const minHeight = 400
+      const position = { x: 100, y: 100 }
+      const wrapper = shallow(
+        <MovableModal closeFn={() => { }}>
```
wgranger, line 41: Is it necessary to pass in this `closeFn` prop if the default prop is the same?

srallen:
@wgranger I think that's the same issue @beckyrother observed and likely just because of the way I built the stories. There's no guarantee the child elements will resize unless they have the right CSS.

packages/lib-react-components/src/MovableModal/MovableModal.js
srallen: That's any other props that might be coming in with `rndProps` destructured.

packages/lib-react-components/src/MovableModal/MovableModal.js
srallen: The inner elements/child nodes have to have CSS for them to resize. I'll update the readme to make sure devs know this.

packages/lib-react-components/src/MovableModal/MovableModal.spec.js
srallen: Nope. It's a remnant of a copy/paste.

srallen:
@wgranger this is ready again for review.

PR #1673, opened 2020-06-12
Add Kubernetes autoscaling, liveness probe, and dry run

adammcmaster:
This sets up autoscaling with what I think are sensible limits. It also configures a liveness probe to ensure deploys don't cause any downtime (old containers won't be taken down until the new ones are live).

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


kubernetes/deployment-production.tmpl
```diff
@@ -38,6 +37,10 @@ spec:
             limits:
               memory: "2200Mi"
               cpu: "500m"
+          livenessProbe:
+            httpGet:
+              path: /
```
camallen, line 42: i'm not 100% sure the app returns a 200 response on this path...worth asking dev team for this.

kubernetes/deployment-production.tmpl
```diff
@@ -96,6 +111,10 @@ spec:
             limits:
               memory: "275Mi"
               cpu: "500m"
+          livenessProbe:
+            httpGet:
+              path: /
```
camallen, line 116: Again we need to make sure this path responds with 200 else it'll kill the pod.

adammcmaster:
👍 I've added readiness probes now and will wait for word on how the app responds to `GET /`

adammcmaster:
I've just done a quick test and confirmed that `GET /` won't work. Options are:

1. Hardcode some paths that do work – might break if those pages move in the future.
2. Add a dedicated `/_liveness` (or similar) route to each app – probably the best option long term.
3. Change the probes to use `tcpProbe` to just check the app is listening, without actually making a request – doesn't check the app is actually healthy, but better than nothing.

I'm tempted to go with #3 for now just for speed, so that we can get autoscaling deployed this week (I won't have time to implement #2 before the next deploy).

adammcmaster:
Yup that's right, it'll run those every time. The idea is to make the PR status check fail if there's a syntax error in the deployment template. (I added this in a few other repos last week after I opened/merged a PR with a syntax error that failed to deploy.)

adammcmaster:
I've updated the probes to use `tcpSocket` instead of `httpGet`. We can change that later if we want to.

eatyourgreens:
I haven't checked the content pages app, but there's an index page set up for the projects server.
https://github.com/zooniverse/front-end-monorepo/blob/master/packages/app-project/pages/Index.js

In theory, you should get that back from a request for `/` for the projects app.

eatyourgreens:
It turns out macOS is case-insensitive, so `/Index` is served as `/` for local development on a mac.

Both apps have a page at `/Index` which can be used for the health check.

kubernetes/deployment-production.tmpl
```diff
@@ -96,6 +113,12 @@ spec:
             limits:
               memory: "275Mi"
               cpu: "500m"
+          livenessProbe:
+            tcpSocket:
+              port: 3000
```
camallen, line 118: I assume we're switching to use the `/Index` path via HTTP instead of tcp?

adammcmaster: Yup, going to switch that over this afternoon.

PR #1676, opened 2020-06-16
Refactor subtask to use MovableModal

srallen:
_Please request review from `@zooniverse/frontend` team. If PR is related to design, please request review from `@beckyrother` in addition._ 

Package: lib-react-components, lib-classifier

Describe your changes:
Swaps out the Layer and ReactRnd component in the sub-task with the new MovableModal. Here's how it's looking in the transcribed lines story:

Screen Shot 2020-06-16 at 11 38 08 AM

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #1674, opened 2020-06-12
Reusable movable modal

srallen:
_Please request review from `@zooniverse/frontend` team. If PR is related to design, please request review from `@beckyrother` in addition._ 

Package: lib-react-components

Describe your changes:
The heart of this was already approved in #1661, but it was reverted because it was breaking the next.js app builds with a `window` is not defined error when building for SSR. After nearly two days of debugging, I found that the issue was with `react-draggable`, a dependency of `react-rnd`. It seems the likely culprit is what was described in https://github.com/STRML/react-draggable/issues/472. It appears this has been patched, but I can't 100% confirm this because I couldn't figure out how to successfully get `react-rnd` to use an updated version of `react-draggable` using yarn resolutions in the workspace. I can get the builds to work again by locking `react-rnd` to version 10.1.0, which is not a longterm solution, but works for now.

Please do check that the builds are working for you either using `yarn build` or `docker-compose -f docker-compose-prod.yml build --pull`

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #1677, opened 2020-06-16
Refactor mark subtask visibility, add enter and space key events

mcbouslog:
Package: lib-classifier

Towards: #1546 (Item 2, touch loop issue, towards Item 1)

Describe your changes:
- add `role='button'` to marks
- add enter and space key events to show subtask from active mark
- refactors subtask visibility from mark selection to finish (including dragEnd)

Potentially helpful staging workflows:
- http://localhost:8080/?project=1876&workflow=3363
- http://localhost:8080/?project=1764&workflow=3392

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


shaunanoordin:
I'm currently testing this PR on a workflow that's dedicated for sub-tasks (http://localhost:8080/?env=staging&project=1898&workflow=3396) and so far the early tests are looking good. 👍 

Still a lot more to go through, though-

## Testing Plan

Setup:

- Workflow with Drawing task, with 3 tools:
- A. Point-type tool with single answer sub-task
- B. Line-type tool with text sub-task
- C. Rectangle-type tool with multiple sub-tasks

Actions:

- Mouse:
  - Click on an empty spot -> create mark (A only)
  - Click and drag on an empty spot -> create mark (B, C only)
  - Click on an existing mark -> open sub-task
  - Click and drag an existing mark -> move mark
  - Focus on an existing mark, then click Delete button -> delete mark (the mark from the drawing tool, not @mcbouslog !)
- Keyboard
  - Tab navigation -> traverse (focus) through existing marks _without opening_ sub-task popup
  - Space or Enter key while focused on existing mark -> open sub-task popup
  - ⚠️  (unidentified action) -> delete mark

Note: while sub-task popup is open, it should capture all keyboard input, preventing tab-navigating to items "behind" the popup. I'm not testing this however, since that's the purview of the popup/modal component. (see #1674, #1676)

Advanced Scenarios of Paranoia:

- How is the interactivity of the marks affected when we move to the next step of the workflow?
  - (Expectation: previous marks must not be focusable or interactive.)
- How is the interactivity of the marks affected when we move the next step of the workflow, then _go back?_
  - (Expectation: business as usual)
- How about workflows with drawing tasks that have NO sub-tasks?
  - (Expectation: business as usual)
- Did you confirm annotation values are saved on the sub-task popup closing?
  - (Expectation: duh)

shaunanoordin:
### Testing Plan (con't)

UI considerations:

- Is it UI being very clear about which existing mark has focus?
  - Consider testing with different mark types
  - Consider testing with different colours

shaunanoordin:
I've picked up a few issues with my tests, but as far as I can tell, none are related to this branch. (In fact, I'm almost certain these are bugs _I should have caught myself_ when I was working on the sub-tasks. 🤦 )

1. when existing drawing marks are _inert_ (e.g. a user has moved on to the next Step of the workflow, which is no longer drawing marks on the image), you can still tab-navigate to them. They can get focus (question mark?) but can be interacted with (pressing Space/Enter or clicking does nothing)

2. [UI] It's not very clear which drawing mark has focus.

3. clicking on an existing, _inert_ drawing mark causes a hidden `Uncaught TypeError: activeTool.createMark is not a function, at onPointerDown` error in the dev console

4. Say you have a drawing Task that has multiple Tools, and some tools HAVE sub-tasks but other tools DON'T have sub-tasks. If the active tool DOESN'T have a sub-task, and you click on a mark with a sub-task, then the sub-task popup won't appear.

Screen Shot 2020-06-18 at 17 22 56
_While "Point + no subtasks" Tool is active, clicking on the red or yellow marks (which do have subtasks) doesn't cause the sub-task popup to pop up._

5. Similarly, if the active Tool selected has a sub-task, but the clicked mark doesn't, an empty sub-task popup appears.

6. Whatever the heck this UI nightmare is.

Screen Shot 2020-06-18 at 17 28 08
_Click to make a Point mark. Go to the next step. Go back to the first step. Note that the mark is now visually in the focused and inert states._

All these problems require their own issue writeup, which I'll have to take care of later. Good news is though, this PR solves the problems it set out to fix. 👍 Tested on mac OS 10 + Chrome 83

I'mma gonna update this branch and try it on FF and Safari, IIRC at least one of them had funky ideas about handling focus on SVG elements.

shaunanoordin:
⚠️ Holy crap I can barely get any of the drawing tools to work on macOS 10 + Safari 12, but this is applicable to `master` as well as to this branch. 

FF works fine though.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingToolMarks/DrawingToolMarks.js
```diff
@@ -42,12 +42,19 @@ function DrawingToolMarks ({ activeMarkId, marks, onDelete, onDeselectMark, onFi
     }
 
     function deselectMark (event) {
-      onDeselectMark(mark)
```
shaunanoordin, line 45: [cleanup] It seems that the `onDeselectMark` event hook is being removed here, which is fine, because it doesn't seem to be used. (If some future tool needs it, we can always add it back.)

If that's the case, won't it make sense to remove the `DrawingToolMarks.onDeselectMark` property altogether?

It's declared in the constructor, propTypes, defaultProps, and the test specs but not actually invoked anywhere; and heck, I can't seem to find the term `onDeselectMark` anywhere else in the `lib-classifier`.

srallen:
@shaunanoordin I'd double check that the polyfill for pointer events is loading as we would expect. That would be my guess as to why things might be breaking for you in Safari 12.

shaunanoordin:
Thanks @srallen ! I'm going to follow up and figure if something's up with `pepjs` in Safari.

mcbouslog:
Thank you @shaunanoordin! I also noticed points 4 and 5 regarding the subtask showing or not showing unexpectedly. I'm still processing a few other notes, not sure I'll have anything to add, but if I do will add comments in this PR or in related new issues (which let me know if you'd like help with).

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/DrawingToolMarks/DrawingToolMarks.js
mcbouslog: I think you're absolutely right `onDeselectMark` isn't used, but I think for now I'd like to add the hook back in for consistency between DrawingToolMarks and InteractionLayer, though structure `deselectMark` similar to `endMoveMark`. Maybe we can make a point to review/refactor/cleanup the drawing pieces once at a somewhat complete state?

shaunanoordin: 👍

PR #1683, opened 2020-06-18
Consensus UI

srallen:
_Please request review from `@zooniverse/frontend` team. If PR is related to design, please request review from `@beckyrother` in addition._ 

Package: lib-react-components, lib-classifier

Toward #1660

Describe your changes:
This implements the UI of the pop-up alone. It can be previewed in the storybook, but it is not wired up to the `TranscribedLines` component yet because I need to extract out the positioning code into a reusable utility function out of the `SubtaskPopup` component.

I'd like to get design feedback / approval. WIP still because I need to write and update tests. I've included a few minor tweaks to the shared `Modal` and `MovableModal` components.

Here's how this is looking:

Invision: https://projects.invisionapp.com/d/main?origin=v7#/console/12923997/420788232/inspect?scrollOffset=8607

(One note on the font of the transcriptions. I haven't added `Roboto Mono Regular` to our theme's font faces yet. I have the fallback set to `monospace` and that's how that looks here. @beckyrother let me know if you want the fallback to be something else)

Screen Shot 2020-06-18 at 3 08 43 PM

With overflow:

Screen Shot 2020-06-18 at 3 08 05 PM

In dark theme:

Screen Shot 2020-06-18 at 3 08 24 PM

Unfortunately styling scroll bars with CSS isn't well supported yet, so I think we'll just have to deal with that being ugly. https://caniuse.com/#search=scroll%20bar%20styling

# Review Checklist

## General

- [x] Are the tests passing locally and on Travis?
- [x] Is the documentation up to date?

## Components
- [x] Has a storybook story been created or updated?
- [x] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [x] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [x] Is it passing accessibility checks in the storybook?

## Apps

- [x] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [x] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

[...]

PR #1686, opened 2020-06-22
New classifications from previous transcriptions

srallen:

Package: lib-classifier

Toward #1316

Describe your changes:
This is partly a refactor and partly an implementation of the classification from previous transcription feature. I wanted to get this up early and get feedback before continuing forward. This PR does the following:

- Refactors `SubTaskPopup` to be directly connected to the stores with its own container component.
- Refactors `TranscribedLines` to be directly connected to the stores with its own container component.
- Refactors the sub-task state to be managed by the generic `Mark` model
  - The reasoning behind this is because the click/keydown event on a `TranscriptionLine` in `TranscribedLines` now calls `setSubTaskVisibility` with a snapshot for storing the options that went into the reduction for that transcription line.
  - The previous transcriptions are keyed by `mark.id` 
  - Keying by `mark.id` allows us to load up the previous transcriptions for subsequent editing as well
- Adds `subTaskPreviousAnnotations` to the mark model
- Adds a `TextPreviousAnnotations` model

I chose to more generically model this concept even though the text sub-task for the transcription task is the only task types that functionally supports loading reductions, but I think this sets us up to be able to support other kinds (though for question type annotations, only the aggregated result makes sense in loading?). 

These refactors break a bunch of tests, so before I fix those, I would appreciate early feedback on the concepts particularly from @eatyourgreens and @wgranger. Here's a gif of the action (sorry about the low res gif, you can also pull this down and check out the `TranscribedLines` story to see it in action):

previous-annotations

# Review Checklist

## General

## Components
  - [x] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
```diff
@@ -97,12 +123,12 @@ function TranscribedLines ({ lines, scale, task, theme }) {
...
-                role='img'
+                role='button'
                 aria-describedby={id}
                 aria-label={line.consensusText}
                 focusColor={focusColor}
-                onClick={e => createMark(line)}
-                onKeyDown={e => (e.key === 'Enter' && createMark(line))}
+                onClick={event => createMark(event, line)}
+                onKeyDown={event => (event.key === 'Enter' && createMark(event, line))}
                 tabIndex={0}
```
eatyourgreens, line 132: My bad, but @shaunanoordin pointed out that lines are always focusable and interactive, no matter which step of the workflow you are on. They could be marked up as disabled (`aria-disabled="true"`), outside of the transcription task, so that there's no expectation that they can be clicked during question steps, say.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextPreviousAnnotations.js
```diff
@@ -0,0 +1,10 @@
+import { types } from 'mobx-state-tree'
+
+const TextPreviousAnnotations = types.model('TextPreviousAnnotations', {
```
eatyourgreens, line 3: From the name, I expected this model to be a collection of `Annotation` models but I don't have any better suggestions for names either.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextPreviousAnnotations.js
```diff
@@ -0,0 +1,10 @@
...
+  id: types.identifier,
+  taskKey: types.string,
+  taskType: types.string,
```
eatyourgreens, line 6: Can `id`, `taskKey` and `taskType` be composed from the `Annotation` model?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
```diff
@@ -17,24 +18,47 @@ export const ConsensusLine = styled('g')`
...
 
     if (activeTool) {
       const mark = activeTool.createMark(markSnapshot)
       mark.finish()
       setActiveMark(mark)
+
+      let previousAnnotations = []
+      activeTool.tasks.forEach((task) => {
+        const previousAnnotation = {
+          id: mark.id,
```
eatyourgreens, line 45: Are identifiers required to be unique for each of these annotations?

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
```diff
@@ -13,6 +13,9 @@ const BaseMark = types.model('BaseMark', {
     TextTask.AnnotationModel
   )),
   frame: types.optional(types.number, 0),
+  subTaskMarkBounds: types.optional(types.frozen({}), undefined),
+  subTaskPreviousAnnotations: types.map(TextTask.PreviousAnnotationsModel),
+  subTaskVisibility: types.optional(types.boolean, false),
```
eatyourgreens, line 18: Should these be volatile, so that they aren't included in the classification when the mark is serialised?

If they aren't needed globally in the page, or in the classification, then they can be stored in component state for each mark. I'm not 100% sure if that would work.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextPreviousAnnotations.js
srallen: How about `TextPreviousAnnotationValues`?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
```diff
@@ -17,24 +18,47 @@ export const ConsensusLine = styled('g')`
...
+    activeTool,
+    activeToolIndex,
+    setActiveMark
+  } = task
+
+  function onFinishWithRef(event) {
+    onFinish(event, ref.current)
+  }
+
+  function createMark (event, line) {
```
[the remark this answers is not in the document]
srallen: oh, it's cruft from an earlier bit of work that I didn't end up committing. Thanks for the catch.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
```diff
@@ -17,24 +18,47 @@ export const ConsensusLine = styled('g')`
   }
 `
 
-function TranscribedLines ({ lines, scale, task, theme }) {
-  function createMark (line) {
-    const { activeTool, activeToolIndex, setActiveMark } = task
+function TranscribedLines ({ lines, onFinish, scale, task, theme }) {
+  const ref = createRef()
```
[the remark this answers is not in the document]
srallen: Hm... not sure if that would be a problem with refs or with how the state of the input was being managed at the time. `useRef` and `createRef` are pretty different functionally under the hood, and I've had issues with the hooks version not being available immediately for use in the component... but the ref in this case isn't used until an event is triggered, so that might be a moot point.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
srallen: Yes, I think so. I think the variable naming here actually is adding to the confusion, so I'll think about better naming for this and the associated model.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
srallen: Oh, volatile would be a much better pick. Thanks for the suggestion. 

I think the previous annotations and visibility needs to be in the MST store state. Previous annotations need to be stored and available for the `TranscribedLines` and `SubTaskPopup` components. The visibility ensures only one subtask can be open at a time. The bounds, however, may be a candidate for local component state.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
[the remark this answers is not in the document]
eatyourgreens: If this works, then I'm inclined to leave it alone. I've run into issues where `createRef().current` is null and doesn't update without a change of state or props.

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextPreviousAnnotations.js
eatyourgreens: Sounds good. I don't have a strong opinion on the name here.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
eatyourgreens: If the popup uses `React.Portal`, then you could wrap it inside its parent line and pass the line bounds down as a prop.
```jsx
<Line>
  <SubtaskPopup/>
</Line>
```

That might be too big of a change for this PR. Portals are really useful when you want to pass data to another part of the page.

eatyourgreens: If only one subtask can be active at a time then we could rework that, at some point, so that it works similarly to active marks ie. store the active subtask as an optional property of the parent Task. I think that would avoid having to track a bunch of booleans, and enforce that only one is true at any time by only ever storing one at a time.

srallen: All great ideas, but yeah, I agree that it's too much for this PR. Could you open an issue tracking these suggestions?

packages/lib-classifier/src/plugins/tasks/TextTask/models/TextPreviousAnnotations.js
srallen: Looking over the `Annotation` model, there are the snapshot processors, a couple of computed views, and an action that this model doesn't need, so I'm going to forgo the composition.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
srallen: Actually, I switched up the identifier to be the task key because this is already a child node of the mark which is uniquely identified. I do need an identifier, but it can just be the task key which matches the subtask.

srallen:
I think this is pretty much ready to go, but I'll leave it as a draft because there will be a merge conflict once #1683 merges, so I'll wait for that to rebase and then request review.

srallen:
This is now rebased, but I found some bugs that I need to spend a bit of time with to determine if they were introduced by this or my recent related work or if they were pre-existing. If introduced by this branch, then I will work to resolve. If not, I'll just n note it here and in an issue.

srallen:
This branch is breaking the next step's question task after the first step's transcription task:

Screen Shot 2020-07-07 at 11 15 01 AM

Next step is debugging this after confirming that this bug is not present on the master branch.

srallen:
I've added a spec and fix for the bug introduced.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
srallen: I've added `aria-disabled` state to the `ConsensusLine`, but I'm confirming whether or not a volunteer should be able to see the `ConsensusPopup` or not on other tasks. It might be useful, since there is a final question task in the next step asking if the letter is finished.

srallen:
This is ready for final review and can be tested out with the ASM staging workflow for the new classifier at: http://local.zooniverse.org:8080/?project=1764&workflow=3392

I'd like to get a second subject added, so I've asked @mcbouslog to look into this.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
srallen: I've confirmed with @snblickhan and @beckyrother that the consensus should be able to be referenced on other step tasks, but that volunteers should only be able to make a new mark and transcription on the step with the transcription task. I'll need to make a minor change to the completed lines.

srallen: Ok this piece is now ready to go

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
```diff
@@ -32,27 +32,33 @@ class TranscribedLines extends React.Component {
...
+      activeTool.tasks.forEach((task) => {
+        const previousAnnotationValuesForThisMark = {
+          id: mark.id,
+          taskKey: task.taskKey,
+          taskType: task.type,
+          values: line.textOptions
+        }
+        previousAnnotationValuesForEachMark.push(previousAnnotationValuesForThisMark)
+      })
+      mark.setSubTaskVisibility(true, this.ref.current, previousAnnotationValuesForEachMark)
```
mcbouslog, line 60: I think `this.ref.current` is the parent element containing all the lines, including the individual line we want? I think that's causing an issue with the subtask positioning. If I moved `ref={this.ref}` to the each ConsensusLine component the subtask positioning appeared to be what I expect, but have not tested thoroughly.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
```diff
@@ -144,12 +151,17 @@ class TranscribedLines extends React.Component {
                 label={<TooltipLabel fill={fills.transcribed} label={counterpart('TranscribedLines.transcribed')} />}
               >
                 <ConsensusLine
-                  role='img'
+                  role='button'
                   aria-describedby={id}
+                  aria-disabled={disabled}
```
eatyourgreens, line 156: `aria-disabled` should be a string: `aria-disabled="true"`.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
```diff
@@ -64,27 +67,21 @@ describe('Component > InteractionLayer', function () {
...
+      wrapper = shallow(
+        <InteractionLayer
+          activeMark={mockMark}
+          activeTool={activeTool}
+          frame={2}
+          setActiveMark={setActiveMarkStub}
+          height={400}
+          width={600}
+        />, {
+          wrappingComponent: SVGContext.Provider,
```
eatyourgreens, line 82: Do you need to wrap the component in a context provider if you're mocking `useContext`?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
```diff
@@ -156,7 +170,7 @@ describe('Component > InteractionLayer', function () {
           }
         }
         wrapper.find(StyledRect).simulate('pointerdown', fakeEvent)
-        wrapper.find(InteractionLayer).simulate('pointermove', fakeEvent)
+        wrapper.find('g').simulate('pointermove', fakeEvent)
```
eatyourgreens, line 173: Could this select the root node of the wrapper, to avoid finding elements by name?

eatyourgreens:
I'm not sure if it's introduced by this PR, but the `TranscribedLines` component is rendering invalid lists.

TranscribedLines story in the storybook, warning that there are invalid list elements in the DOM.

eatyourgreens:
There's also a bunch of failures for `aria-describedby` attributes.
Screenshot of warnings in the storybook for bad aria-describedby values.

srallen:
> There's also a bunch of failures for aria-describedby attributes.

The failures are because the tester can't trigger the tooltip, but if you actually try it out, the aria labelling works as expected with voice over.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
srallen: I'm opened a separate issue for this and have noted this observation to it: https://github.com/zooniverse/front-end-monorepo/issues/1702#issuecomment-656712459

eatyourgreens:
>The failures are because the tester can't trigger the tooltip, but if you actually try it out, the aria labelling works as expected with voice over.

Sorry, but those errors are triggered because those element IDs are not present in the DOM. I checked. If those elements are being added dynamically to the DOM, then `aria-describedby` should be set dynamically too. It must always point to an element that is present in the current DOM.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.spec.js
```diff
@@ -57,11 +57,11 @@ describe('Component > InteractionLayerContainer', function () {
       expect(wrapper.find(InteractionLayer)).to.have.lengthOf(1)
     })
 
-    it('should render SubTaskPopup', function () {
+    xit('should render SubTaskPopup', function () {
```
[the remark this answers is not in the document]
srallen: We still need some of these tests, but I forgot to update this block. We don't need to test if `TranscribedLines`  or `SubTaskPopup` renders anymore, but the others are testing the frame filters and hide previous marks functionality. Thanks for catching this.

srallen:
@eatyourgreens ok, I can have a look again, but are you requesting changes for this PR or is noting in an issue ok?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
srallen: This is cruft  I think.

eatyourgreens:
I don't think invalid attribute values should hold this up. I think an issue's fine.

PR #1657, opened 2020-05-20
Add marks filter per frame

mcbouslog:
Package: lib-classifier

Towards #1625. 

Describe your changes:
- filters marks by current frame

For a subsequent PR - filtering marks by frame should be dependent on [workflow/subject-viewer configuration](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/Workflow/WorkflowConfiguration/WorkflowConfiguration.js#L9) as while a transcription workflow would likely want this behavior a survey workflow may not, similar to #1538. 

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #1691, opened 2020-06-27
Add ADR 27 Subject Viewer Config

mcbouslog:
Package: ADR for `docs`, related to `lib-classifier`

Relevant issues/PRs: #1538, #1627, #1643, #1657

Describe your changes:
- adds ADR for workflow `subject_viewer_config`

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


docs/arch/adr-27.md
```diff
@@ -0,0 +1,53 @@
...
+
+```javascript
+{
+  cell_width: [number of pixels]
+  cell_height: [number of pixels]
+  cell_style: { [CSS property]: [CSS property value] }
+  grid_columns: [number]
+  grid_rows: [number]
+}
+```
```
mcbouslog, line 42: Could omit noting code and exclusively refer to READMEs?

Regardless of, but related to point above, I could add multi-frame viewer README as part of this PR or leave for separate PR.

docs/arch/adr-27.md
```diff
@@ -0,0 +1,53 @@
...
+## Decision
+
+Subject viewer configuration is an object stored in `workflow.configuration.subject_viewer_config` and is structured as follows for the subject viewers noted:
+
+`multiFrame`:
+
+```javascript
+{
+  filter_marks_per_frame: [boolean]
+  maintain_pan_zoom: [boolean]
```
mcbouslog, line 28: Originally had something like `pan_zoom: ['maintain', 'reset']`, but thought might be better to have as boolean if makes sense?

Not including rotate in the key, but technically I think we would include, so open to suggestion if ok as is or should also note `rotation` in key.

docs/arch/adr-27.md
```diff
@@ -0,0 +1,53 @@
...
+Created: June 26, 2020
+
+## Context
+
+A workflow determines which subject viewer to use with the `workflow.configuration.subject_viewer` property. Subject viewers include `singleImage`, `lightCurve`, `multiFrame`, and `subjectGroup` at creation of this ADR. Some subject viewers can utilize, or require, additional configuration information.
+
+### [multiFrame](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer)
+
+A workflow using the multi-frame subject viewer might have a preference regarding:
+marks per frame: some might prefer marks filtered per frame, like a transcription workflow where each frame represents a unique page to transcribe with marks only relevant to each page, while other workflows might prefer marks persist between frames [, like ?] (`multi_image_clone_markers` in PFE)
```
srallen, line 12: Yes, I think this is currently `multi_image_clone_markers` in PFE, however, this setting is slightly confusing to some project builders. I've had some ask if this setting creates new cloned marks of the original mark onto the other frames, which would add marks into the annotation. I've had to explain that this setting is just for visibility between frames of existing marks on other frames and doesn't auto-create additional marks in the annotation. We should note that confusion about this setting here.

srallen, line 12: For readability, maybe add a bullet point here?

docs/arch/adr-27.md
```diff
@@ -0,0 +1,53 @@
...
+pan and zoom: some might prefer pan and zoom reset per frame, like a transcription workflow where each frame represents a unique page to transcribe, while other workflows might prefer pan and zoom maintained between frames, like [Wildcam Gorongosa](https://www.zooniverse.org/projects/zooniverse/wildcam-gorongosa/classify) or [Backyard Worlds](https://www.zooniverse.org/projects/marckuchner/backyard-worlds-planet-9/classify) (in flipbook mode, not separate frames)
```
srallen, line 13: Maybe bullet point here? Is there a current setting for this on PFE? We should note that here or is a certain function assumed when using flipbook vs display all frames at once?

docs/arch/adr-27.md
srallen: I could see an enumerable being useful if it's possible to have any other states... Just a note on the code example, maybe use < and > instead of brackets since this sort of looks like you mean an array that contains a boolean?

docs/arch/adr-27.md
```diff
@@ -0,0 +1,53 @@
...
+`subjectGroup`:
+
+```javascript
+{
+  cell_width: [number of pixels]
+  cell_height: [number of pixels]
+  cell_style: { [CSS property]: [CSS property value] }
+  grid_columns: [number]
+  grid_rows: [number]
+}
```
srallen, line 41: Same comment here about using <> instead of brackets, expect for the key of the object, because that is the syntax for variable object keys.

docs/arch/adr-27.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+Subject viewers to define the configuration object in the related subject viewer README, if applicable.
+
```
srallen, line 45:
```suggestion
The ScatterPlotViewer and BarChartViewer accept a configuration object directly in the JSON structure of the subject data to support variability in the display settings including the plot labels since it's possible this might vary per subject. The `subject_viewer_config` object should only be used for  configuration options that apply to all subjects linked to the workflow. 
```

docs/arch/adr-27.md
srallen: We should note here in the decision that `filter_marks_per_frame` is replacing `multi_image_clone_markers`.

srallen: > Not including rotate in the key, but technically I think we would include, so open to suggestion if ok as is or should also note rotation in key.

The setting is more than maintain the rotation between frames though? Isn't the scale and pan position maintained as well or no?

docs/arch/adr-27.md
eatyourgreens: It was originally requested by Space Warps, and they might be the only project that have used it.

srallen: We should check on that because it's not behind a feature flag and the checkbox is available to everyone in the lab UI. As I said, I've directly responded to questions in tickets about it.

eatyourgreens: Yeah, if we could get a list of projects that are using it, that would be useful. I remember, from testing on Space Warps, that it didn't work well with subtask popups either.

docs/arch/adr-27.md
srallen: The sample code is useful in the ADR for the context of the decision as a record. I think we should also have a readme for the viewers as well, but the readmes can and will change whereas the ADR might now, so we should have both.

srallen: The readmes for the viewers can be done separately though.

docs/arch/adr-27.md
mcbouslog: That's correct, in Wildcam Gorongosa rotation, scale, and pan position are maintained between frames. Maybe `pan_zoom` could be renamed to something like `positioning` and include pan, zoom, and rotation?

srallen: I wonder if there will ever be a request to maintain only one of these? I was thinking about suggesting this be an enumerable if various values, but then we would have to support each of these configurations and that might be too complex and not actually wanted. I would support making the proposed key `maintain_positioning` though

docs/arch/adr-27.md
```diff
@@ -0,0 +1,54 @@
...
+## Decision
+
+Subject viewer configuration is an object stored in `workflow.configuration.subject_viewer_config` and is structured as follows for the subject viewers noted:
+
+`multiFrame`:
+
+```javascript
+{
+  filter_marks_per_frame: <boolean>  // replaces multi_image_clone_markers in PFE
+  positioning: <enumerable> // includes pan, zoom, and rotation, enumerable i.e. "maintain", "reset"
```
srallen, line 29: `maintain` and `reset` is good for now and represents the current boolean state that we support, but keeps this flexible for supporting more configuration in the future. :+1:

docs/arch/adr-27.md
mcbouslog: Here's a list of launch approved projects with active workflows using `multi_image_clone_markers`:

Agent NEO
Brain Match
Eye for Diabetes
Fishing in the past
ForestEyes
Galaxy Builder
Galaxy Zoo: Clump Scout
Heritage Quest
Hubble Asteroid Hunter
Local Group Cluster Search
Michigan ZoomIN
Monkey Health Explorer
Planetary Response Network & Rescue Global: Caribbean Storms 2017
Poppin' Galaxy
Power to the People
Project Plumage
Radio Galaxy Zoo: LOFAR
Run, Herring, Run!
Snapshot Camdeboo
South Sudan DiversityCam
Space Warps - HSC

I spot-checked 5 - 2 didn't have multi-image subjects for the subject I saw (Fishing in the Past, Hubble Asteroid Hunter), 2 using survey task (Michigan ZoomIN, South Sudan Diversity Cam, which persists the tag choice between frames regardless of `multi_image_clone_markers`), and 1 used as intended (Power to the People), but this is just a rough spot-check, not determinate.

So it's definitely being used, but it's not clear to the project builders what the setting is doing.

srallen: Let's add some of this info into the ADR here. It's good to have confirmation that project builders don't understand this setting.

mcbouslog:
I think this is good enough for initial merge, though happy to make further edits. I'll merge tomorrow morning unless read otherwise.

PR #1709, opened 2020-07-13
VSLCV phased focus

### LINKED ISSUE #1426: [VSLCV] Phase toggles for composite viewer
**Package** lib-classifier

The composite viewer needs a set of radio buttons that sets the phase to display that is being displayed in the scatter plot. The updated data will then be passed down via props to the scatter plot component and bar chart component. There may be some difficulty in implementing some of the UI changes in the scatter plot itself, since the background is custom to this specific VSLCV feature.

Time estimate: 2 weeks

**Additional context**

https://projects.invisionapp.com/d/main#/console/17294981/360635448/inspect

srallen:

Package: lib-classifier

Closes #1426

Describe your changes:
This adds the functionality to pick which phase to use for the calculation of the data series on the phased scatter plot. The visibility and phase focus is now correctly separate controls and control types, checkbox vs radio button, that is now visually distinct.

I had to do several tweaks, that we last discussed in person @beckyrother I think in February? Anyway, here's a preview of how this looks and functions:

separate-controls

This can be reviewed using the storybook. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
@beckyrother I didn't realize how tiny that gif shows up here after I compressed it down. I can make another higher res one to send to you over slack. It was gonna be 10MB originally!!

srallen:
Up close of the controls area:

Screen Shot 2020-07-14 at 11 27 36 AM

I left the gray background on hover/focus from the original pill button design, but I'm not so sure about keeping it. I tried making another gif of the area, and it's not showing up well in that, so I can leave it for now and when this merges, deploy the storybook so you can have a look?

beckyrother:
I see it in the original gif! I think it's ok to leave in – the hover states aren't super obvious so one more thing won't hurt.

srallen:
This is ready for review!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/components/Controls/Controls.spec.js
```diff
@@ -22,75 +24,84 @@ describe('VariableStarViewer > Component > Controls', function () {
...
       const setYAxisInversionSpy = sinon.spy()
       const wrapper = mount(
-        <Grommet theme={zooTheme}>
-          <Controls setYAxisInversion={setYAxisInversionSpy} />
-        </Grommet>
+        <Controls setYAxisInversion={setYAxisInversionSpy} />,
+        {
+          wrappingComponent: Grommet,
+          wrappingComponentProps: { theme: zooTheme }
+        }
```
wgranger, line 31: Out of curiosity, what is the benefit of "wrapping" as an argument to mount rather than explicitly wrapping as before? I know both exist, but I'm not personally aware of the difference.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/components/Controls/components/Label/Label.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { SpacedText } from '@zooniverse/react-components'
+import getDataSeriesColor from '../../../../../../helpers/getDataSeriesColor'
+import getDataSeriesSymbol from '../../../../../../helpers/getDataSeriesSymbol'
```
wgranger, line 6: Is this the quickest way to access this? I thought the monorepo allowed accessing from the root, so maybe it could be `import getDataSeriesSymbol from 'helpers/getDataSeriesSymbol'`. Either way, this is minor.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/components/Controls/components/Label/Label.js
```diff
@@ -0,0 +1,61 @@
...
+
+  /* Could use Grommet gap prop instead */
+  svg {
+    padding: 0px 0.5ch;
+  }
+
+  /* Might remove this */
+  &:hover, &:focus-within {
+    background-color: rgba(216,216,216,0.4);
+  }
```
wgranger, line 23: What would be the reason for removing this? I suppose this comment should have more info if it stays, or either the styling or comment could be removed.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/components/Controls/components/PeriodMultipleControls/PeriodMultipleControls.spec.js
```diff
@@ -0,0 +1,80 @@
...
+
+  describe('FormField', function () {
+    let wrapper, formField
+    before(function () {
+      wrapper = shallow(
+        <PeriodMultipleControls
+          options={options}
+          periodMultiple={1}
+          setPeriodMultiple={() => { }}
+        />
```
wgranger, line 36: wrapper is set three times in this test. The only difference being that on one occasion, `setPeriodMultiple` has a spy attached. Could they all be defined the same way to utilize a `beforeEach` and setting the wrapper once?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewer.js
```diff
@@ -160,6 +165,7 @@ VariableStarViewer.defaultProps = {
     barCharts: []
   },
   setPeriodMultiple: () => { },
+  setPhaseFocusedSeries: () => {},
```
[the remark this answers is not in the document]
srallen: No it is not. I'll fix this.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/components/Controls/Controls.spec.js
srallen: Using these options allows us to wrap a component with a parent component if needed, but guarantees the shallow or fully mounted component for the tests is the intended target. Before the mounted component was `Grommet` and the tests had to always use `wrapper.find(Controls)` to then subsequently test with the intended component. I think there's also some API limitations for the returned components on `find`.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/components/Controls/components/Label/Label.js
srallen: This is a bit confusing unfortunately. The aliased `helpers` path is for the very top level `helpers` folder. The sub-folder for subject viewers also has a `helpers` folder which is what is being used here. The goal is to ideally break off the subject viewers into its own package for eventual reusability with Talk if we ever get around to redoing that (and maybe elsewhere?). I could alias these as `subject-viewer-helpers` for now to make this more clear.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/components/Controls/components/Label/Label.js
srallen: I left a note for myself in case @beckyrother wanted it removed since the styles for these controls had to change. She okayed the background color, so I'll leave it in and remove the comment.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/components/Controls/components/Label/Label.js
srallen: Actually, we already have `@viewers` aliased, so I'll use that

srallen:
> Also, looks like there is a slight misalignment of one of the radio buttons on Safari.

I don't understand why this is happening. The radio buttons right next to it for the period is using the same CSS and is fine. If I adjust it, the inner circle is off center in other browsers. Is it ok if this is noted in issue and I come back to it at another time?

Issue #1645, opened 2020-05-07
[RFC] How do we do pan and zoom?


srallen:
We have four implementations of pan and zoom functionality in the monorepo. In addition, some CFEs have implemented this and may have unique requirements that we should be aware of as we get to parity function for projects to migrate over. Let's discuss the various ways of how we pan and zoom and ideally come to a decision that we can document as an ADR.

The four ways in the monorepo are:

1. PH TESS's `LightCurveViewer`: This is purely done in d3, coupled with the functionality required for PH TESS.
2. `SVGPanZoom`: This is essentially a port of PFE's functionality using a view box and currently used by the image viewers. This has to be wrapped around an svg element to be functional.
3. `VXZoom`: An extension of the vx library's pan and zoom function. Uses a transformation matrix and has some configurability. Currently implemented with the `ScatterPlotViewer` and subsequently the `VariableStarViewer`. Unknown if it is usable with image viewers, but in theory it is.
4. `SubjectGroupViewer`: A variant of `SVGPanZoom` since the transformations have to be applied to multiple image subject simultaneously.

eatyourgreens:
We've also got the [`withKeyZoom`](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/withKeyZoom/withKeyZoom.js) decorator which takes `onPan`, `ZoomIn` and `ZoomOut` functions as props, and wires those up to `keydown` events emitted by the decorated component.

In theory, it should work with any implementation that can pass those three callback functions.

eatyourgreens:
I think my vote would go to a component that can be wrapped around any SVG element (including groups of embedded SVGs for the group viewer) and passes down a transformation that operates on its children. Plus some mechanism for getting the image pixel coordinates of clicks when you are drawing on a photo. At the moment, we invert the current transformation matrix in order to do that. I think ASM might have a better implementation of that than the PFE implementation, but it's been a long time since I looked at that code.

srallen:
@eatyourgreens vx has a utility function for getting point coordinates in events. It's unfortunately undocumented, but you can [read the code here](https://github.com/hshoff/vx/tree/master/packages/vx-event/src). I used it with `VXZoom` for a zoom to point function: 

[...]

PR #1721, opened 2020-07-22
VSLCV zoom controls for individual scatter plot modules

### LINKED ISSUE #1442: [VSLCV] Composite viewer needs separate zoom function for each scatter plot
**Package** lib-classifier

The sidebar zoom tools work great when it's just a single image or you assume all of the displayed images want to zoom simultaneously. For the scatter plot, again, if it is just a single scatter plot, the sidebar tools can be wired up and work. The variable star viewer will have two scatter plots that need to be able to zoom independently.

We discussed leveraging click focus (mouse click or tab and enter or space) for each scatter plot first to be able to enable the zoom side toolbar for that specific scatter plot. The UI should have some kind of discoverability to enable this state and then UI to indicate focus on that scatter plot (maybe outline like other focus UI?). 

We discussed having both plots both zoom at the same time, but this is not viable for the variable star project. We discussed implementing brush zoom, but it will be confusing for volunteers to have two different zoom functions between different scatter plot instances (i.e. TESS). 

**Additional context**

srallen:

Package: lib-classifier

Closes #1442 

Describe your changes:
This adds the functionality to enable pan/zoom for individual scatter plot modules for the VSLCV. This is potentially a model for how we can do this for future composite subject viewers. Notable changes:

- A refactor of the existing `onPan` to be flexible with the VX style matrix transformations for panning. We should circle back #1645 because making sure keyboard zoom and pan worked between the ~two~ three different methods was a bit cumbersome.
- Fixed a bug with the size and positioning of `ZoomEventLayer` used by `VXZoom` so now it's the actual plot area rather than the whole of the parent container which included the axes labels, etc.
- Added `pretendToBeVisual: true` to the JSDOM configuration so tests wouldn't fail because of needing `cancelAnimationFrame`
- This gets keyboard controls working with the basic scatter plot as well, and this can be see in its pan/zoom story
- Adds a few minor updates to the grommet theme, so be sure to build when pulling this down.

The feature can be tested using the pan/zoom story for the VSLCV. Here are some gifs too!:

Pan and zoom using the mouse:

pan-zoom-mouse

Pan and zoom using keyboard:

pan-zoom-keyboard

Pan and zoom using the toolbar:

pan-zoom-toolbar

~WIP~:
- [x] Tests for new components
- [x] Fix broken stories

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/store/SubjectViewerStore.js
```diff
@@ -28,7 +28,7 @@ const SubjectViewer = types
     - 'zoomValue' defines amount zoomed in/out, or current zoom value of 'zoomto'.
      */
     onZoom: function (type, zoomValue) {},
-    onPan: function (value) {}
+    onPan: function (type, value) {}
```
eatyourgreens, line 31: I think this will break the TESS light curve viewer, so panning probably needs to be updated there too.

srallen: Thanks for the heads up. We don't have a test covering this, so I didn't catch it.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
```diff
@@ -7,22 +7,29 @@ import ZoomEventLayer from '../ZoomEventLayer'
...
+  handleToolbarPan (type) {
+    this.onPan(type)
+  }
+
   handleToolbarZoom (type) {
     const doZoom = {
-      'zoomin': this.zoomIn.bind(this),
-      'zoomout': this.zoomOut.bind(this),
-      'zoomto': this.zoomReset.bind(this)
+      zoomin: this.zoomIn.bind(this),
```
eatyourgreens, line 30: You don't need to bind this each time the handler runs. I think you can bind these once in the component constructor, then refer to the bound functions here.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
```diff
@@ -62,6 +69,67 @@ class VXZoom extends PureComponent {
...
+        translateY
+      }
+    } = this.zoom
+    const panDistance = 20
+    const newTransformation = { 
+      down: { translateX, translateY: translateY - panDistance },
+      left: { translateX: translateX + panDistance, translateY },
+      right: { translateX: translateX - panDistance, translateY },
+      up: { translateX, translateY: translateY + panDistance }
+    }
```
eatyourgreens, line 85: Since `dx` and `dy` are passed to `onPan`, the new transformation can be calculated from `translateX + dx * panDistance, translateY + dy * panDistance` for all four directions.

eatyourgreens:
Can the 'Zoom enabled' setting follow the keyboard focus as I tab through the interface? Here, zoom is enabled on the first graph but I've tabbed to the second before pressing + to zoom in. Intuitively, I'd expect the second graph to zoom, but nothing happens because I have to press space to toggle the button first, which is clunky and annoying. On focus, I'd expect zoom enabled to turn on (then off again on blur.)

Screenshot of the interface showing zoom enabled for the first graph, but a keyboard focus ring on the second, so that it receives the keyboard zoom commands.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewerContainer.spec.js
```diff
@@ -557,4 +559,34 @@ describe('Component > VariableStarViewerContainer', function () {
...
+      )
+    })
+
+    it('should default to not have pan and zoom enabled for either scatter plot', function () {
+      expect(wrapper.state().allowPanZoom).to.equal('')
+    })
+
+    it('should set the state for which module is allowed to zoom when setAllowPanZoom is called', function () {
+      wrapper.instance().setAllowPanZoom('rawJSON')
+      expect(wrapper.state().allowPanZoom).to.equal('rawJSON')
```
eatyourgreens, line 585: It's bad practice to test a component's internal state directly like this. Better to test the consequences of a state change instead. I'd merge this test and the following test into something like:
```js
expect(wrapper.find(VariableStarViewer).props().allowPanZoom).to.equal('rawJSON')
```

eatyourgreens: Also, a callback which sets a prop that's then passed back down to the component that fired the callback might mean you've got components whose state is stored outside of those components. I'm not sure.

eatyourgreens:
How about calling `setOnPan` and `setOnZoom` to register the currently selected graph component with the toolbar when the 'Zoom enabled' button is clicked or receives focus? Would that be simpler than registering a handler for the whole viewer, then tracking which component is actually being controlled by the toolbar at any one time?

eatyourgreens:
When the 'Zoom enabled' button has focus, pressing keys to pan and zoom does nothing. I have to tab to the chart itself but it is not obvious that I should do that. I think the keydown events should be picked up by the parent of the button and the chart, so that zooming works when the button has focus.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
```diff
@@ -7,22 +7,29 @@ import ZoomEventLayer from '../ZoomEventLayer'
 class VXZoom extends PureComponent {
   constructor (props) {
     super(props)
-    const { setOnZoom } = props
+    const { setOnPan, setOnZoom } = props
 
+    setOnPan(this.handleToolbarPan.bind(this))
     setOnZoom(this.handleToolbarZoom.bind(this))
```
eatyourgreens, line 13: Instead of calling these on mount, maybe call them when a chart is selected and actively being controlled, so that only one chart is registered with the toolbar at any one time?

eatyourgreens:
Sorry, a couple more comments about the UX. I can't seem to zoom in to look at detail in both graphs. If I zoom in on a detail in the phase graph, then switch to the time graph, enabling zoom on the time graph resets the phase graph back to its initial state.

Zooming both x and y on the phase graph is a bit annoying, in terms of trying to see detail in the light curve. The TESS LCV zooms in x but not y. Is that worth considering as an option for these graphs too?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewerContainer.spec.js
```diff
@@ -557,4 +559,34 @@ describe('Component > VariableStarViewerContainer', function () {
...
+      })
+      wrapper = shallow(
+        <VariableStarViewerContainer
+          subject={subject}
+        />
+      )
+    })
+
+    it('should default to not have pan and zoom enabled for either scatter plot', function () {
+      expect(wrapper.state().allowPanZoom).to.equal('')
```
eatyourgreens, line 580: It would be safer to test the prop that's passed down from state here too.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
```diff
@@ -62,6 +69,67 @@ class VXZoom extends PureComponent {
...
+        case 'ArrowDown': {
+          event.preventDefault()
+          this.onPan('down')
+          return true
+        }
+        default: {
+          return true
+        }
+      }
+    }
```
eatyourgreens, line 128: This function is almost identical to the keydown handler in `withKeyZoom`. Is it possible to merge them and decorate this component with `withKeyZoom(VXZoom)` to add the `onKeyDown` handler?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
```diff
@@ -115,16 +187,21 @@ class VXZoom extends PureComponent {
               {...this.props}
             >
               <ZoomEventLayer
-                onWheel={(event) => this.onWheel(event)}
+                focusable
+                height={height}
+                left={left}
+                onDoubleClick={this.onDoubleClick}
+                onKeyDown={this.onKeyDown}
```
eatyourgreens, line 194: I'm not sure why this isn't catching the pan/zoom keys when the 'Zoom enabled' button has focus.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewer.js
```diff
@@ -68,30 +79,54 @@ const VariableStarViewer = React.forwardRef(function VariableStarViewer(props, r
...
           yAxisLabel={phasedJSON.chartOptions.yAxisLabel}
           yAxisNumTicks={8}
           visibleSeries={visibleSeries}
+          zooming={zoomEnabled.phasedJSON}
         />
       </Box>
       <Box
+        border={zoomEnabled.rawJSON && { color: 'brand', size: 'xsmall' }}
         gridArea='rawJSON'
+        style={{ position: 'relative' }}
```
eatyourgreens, line 110: I think `onKeyDown` handling should happen up here, so that you catch events emitted when the Zoom button has focus.

eatyourgreens:
I think I've found a bug where you can't zoom out again after zooming in to view detail. Here, I've zoomed in to a small group of red dots around 350 days in the time chart. I find that I can't zoom out either by pressing - or by using the toolbar. The only way to get out of this state seems to be to reset the chart with the 'reset' button.

Screenshot showing the Time chart zoomed in on a small clump of red points near 350 days in the light curve.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
srallen: This is a copy of how it's setup elsewhere. It can be changed, but I think out of scope of this PR.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewerContainer.spec.js
srallen: Yes, we're following a state container and view component convention.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
eatyourgreens: Sure, but worth keeping an eye on this for memory leaks if `handleToolbarZoom` is called a lot.

srallen:
@eatyourgreens here's some answers to some of your questions, hopefully this clarifies the feature and UX: 

> Can the 'Zoom enabled' setting follow the keyboard focus as I tab through the interface? Here, zoom is enabled on the first graph but I've tabbed to the second before pressing + to zoom in. Intuitively, I'd expect the second graph to zoom, but nothing happens because I have to press space to toggle the button first, which is clunky and annoying. On focus, I'd expect zoom enabled to turn on (then off again on blur.)

I experimented early on with focus and blur, but chose not to use this and instead have an explicit toggleable state so that the toolbar and mouse is still functional in case a user is using multiple means of input for the interaction.

> When the 'Zoom enabled' button has focus, pressing keys to pan and zoom does nothing. I have to tab to the chart itself but it is not obvious that I should do that.

I chose to have similar UX as the `SingleImageViewer` where the medium has to have focus for the keyboard controls to work. This is for consistency in the UX between viewers, but also to minimize complexity in the code. The `ScatterPlotViewer` won't have this toggle and will just be zoomable by default if it is standalone if you have focus on the plot. I'd like more feedback from others to see if they find this difficult, then we can consider a change. 

> If I zoom in on a detail in the phase graph, then switch to the time graph, enabling zoom on the time graph resets the phase graph back to its initial state.

> Zooming both x and y on the phase graph is a bit annoying, in terms of trying to see detail in the light curve. The TESS LCV zooms in x but not y. Is that worth considering as an option for these graphs too?

This is the requested feature from the project builders. I can see not resetting the zoom going between the two, however, they want zoom in both axes directions. I'll seek feedback from them, but I want the core feature to be reviewed and merged first since this PR is already big enough. By the way, the `ScatterPlotViewer` is configurable to constrain zoom, so it's already a configurable setting and there's knobs in the story for the `ScatterPlotViewer` to check it out.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
srallen: It was difficult to get all three zoom/pan methods to work together. I'd prefer we circle back to RFC #1645 and consolidate on a single approach to panning rather than support panning using viewport and panning using a transformation matrix simultaneously by a single method. I'd like to merge this as is and experiment with swapping the `SingleImageViewer` over to the transformation matrix for pan and zoom instead. Is that ok?

eatyourgreens: Sure, but `withKeyZoom` is a decorator that maps keys to callback functions. I don't think it's tied to a particular implementation of zoom. It's used by both the light curve viewer and single image viewer, which have very different implementations of pan and zoom.

eatyourgreens: The image toolbar uses `withKeyZoom` to connect to the subject viewer store and control the scatter plot viewer from the keyboard when the toolbar has focus.
https://github.com/zooniverse/front-end-monorepo/blob/9f17eea42d4c9926731a81fd0496bbd9dab73122/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/ImageToolbar.js#L14-L15

I've checked in the storybook for this PR. When the toolbar has focus, in Chrome and in Firefox, I can control the scatter plot viewer from the keyboard.

Sorry, but I'm going to ask for changing the keyboard handling here to use `withKeyZoom` since it's already working with the toolbar and I'm not following why the code needs to be implemented in two places.

srallen: I think I got confused about all the implementations and keeping track how this all functioned. I'll have another look, but I still want us to consolidate on a single method.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewerContainer.spec.js
srallen: > It's bad practice to test a component's internal state directly like this. Better to test the consequences of a state change instead. I'd merge this test and the following test into something like:

The props being passed down is tested in the next spec.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewerContainer.spec.js
srallen: The spec on line 590 is testing the props being passed down.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
srallen: The `VXZoom` component is agnostic to what viewer is using it and currently it is a parent component for `ScatterPlot` component when `zooming` is true. It doesn't directly wrap or interact with the `VariableStarViewer`. I think calling these on focus of the parent would couple it to the `VariableStarViewer` with refs, which I don't think I want to do.

eatyourgreens: I was thinking that each scatter plot viewer, when selected, could call `setOnPan` and `setOnZoom` to register itself with the toolbar as the component that is currently being controlled. That's more in line with how those functions were designed to be used but not needed for this PR.

srallen: Have a look sometime at the scatter plot viewer code. `VXZoom` is only parent to the `ScatterPlot` when `zooming` is true, so in effect, these are only being called when the user selects to control it.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
srallen: Ok, this is refactored to use the existing `onKeyDown` handler. I think some documentation would have been helpful here. Would you have time to add some for `onKeyDown`? I can document `VXZoom` in the future, but it'll have to be after I finish up on the VSLCV / upcoming DataImageViewer.

srallen:
After my latest commit, the "variable star light curve data" story for the `ScatterPlotViewer` is now broken, but it is broken because I've connected the stories to the MST stores, and the data isn't quite mocked in the way it expects. Fixing this is out of scope of this PR.

srallen:
> I think I've found a bug where you can't zoom out again after zooming in to view detail. Here, I've zoomed in to a small group of red dots around 350 days in the time chart. I find that I can't zoom out either by pressing - or by using the toolbar. The only way to get out of this state seems to be to reset the chart with the 'reset' button.

@eatyourgreens I haven't replicated this. Any more details for replication?

eatyourgreens:
Sure. Drag the cluster of points around 350 days into the centre of the chart and zoom in to see them in detail. Then zoom out again. Zooming out stops before you return to the full chart. I made this gif after pulling down the latest commits on this branch.
Animation showing the time chart zoomed into a cluster of points at 350 days, then freezing as you zoom out again.

eatyourgreens:
If you try to zoom in on the cluster of red points at the righthand side of the Time chart, the effect is more pronounced. I think it is related to zooming while being near the edge of the chart.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
eatyourgreens: Sure. The `withKeyZoom` component was written while we were rushing to get PH-TESS launched, so documenting it kind of got lost along the way, but it's a lightweight component. I think a short Readme should do. From memory, it connects a component up to the `SubjectViewer` store so that keyboard events on that component will fire the store's `onPan` and `onZoom` actions. (I can't remember the exact names of the actions.)

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewerContainer.spec.js
eatyourgreens: I know. Props should be tested here so that the test doesn't depend on whether the tested component is a class or a function.
```suggestion
      expect(wrapper.find(VariableStarViewer).props().allowPanZoom).to.equal('')
```

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewerContainer.spec.js
eatyourgreens:
```suggestion
      expect(wrapper.find(VariableStarViewer).props().allowPanZoom).to.equal('rawJSON')
```
Same comment here, so that the test is checking behaviour, not implmentation.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js
eatyourgreens: #1726 adds a Readme and some new actions in the subject viewer store.

srallen: Thank you!

srallen:
@eatyourgreens this is ready for review again

srallen:
Oh, I forgot to say, the bug you found appears to have to do with the constrain function with VXZoom, so out of scope of this PR.

eatyourgreens:
That makes sense, because it doesn't happen for the unconstrained plot.

eatyourgreens:
I added a note to #1726 about the `onPan` API being different from the `onZoom` API, which throws me a bit. https://github.com/zooniverse/front-end-monorepo/pull/1726#issuecomment-665125482 I think you're idea of changing it to `onPan(type, { dx, dy })` would be worth exploring.

srallen:
@beckyrother do you have any feedback for the UX for the feature? Thanks!

srallen:
@beckyrother 

> My only other comment would be to turn on the zoom button in the toolbar no matter how 'enable zoom' is selected.

If one of the plots has zoom enabled, then you can tab over to the toolbar and use the keyboard controls from there as well too, so either focus on the plot or focus on the toolbar works. Is this what you are suggesting or something else?

beckyrother:
> If one of the plots has zoom enabled, then you can tab over to the toolbar and use the keyboard controls from there as well too, so either focus on the plot or focus on the toolbar works. Is this what you are suggesting or something else?

Sorry, just meant turning the zoom button in the toolbar teal to indicate that you're using that function

srallen:
> Sorry, just meant turning the zoom button in the toolbar teal to indicate that you're using that function

Ah ok. We don't do this for images either, so I'll track this in a separate enhancement issue.

beckyrother:
In that case, I'll change my review to approved!

Issue #1719, opened 2020-07-21
Transcribe text alpha testing


mcbouslog:
Thank you for testing the new transcription task!! 
While we've identified a few known issues in [the related milestone](https://github.com/zooniverse/front-end-monorepo/milestone/35), please add a comment to this issue noting any issue, unexpected UX, or suggestion.

mcbouslog:
On the second task ("Have all the pages in this letter been completely transcribed?") all previously drawn lines are shown on frame selected, regardless of if drawn lines related to frame selected.

trouille:
![Screen Shot 2020-07-21 at 1 38 56 PM](https://user-images.githubusercontent.com/4975655/88093376-8e681300-cb57-11ea-8601-7f5c4982ec30.png)
If you drag the box to the edge of the page and then use the resize button, the 'save' button gets very big. I think this will be a very rare event and it's easy to 'x' it closed and restart the text box by clicking on the line again, but just wanted to note. (In Chrome on Mac os)

lcjohnso:
Tested using Chrome on Mac OS.

1. The dropdown options sometimes appear awkwardly: above the text box, covering the instructions; dropdown options list extends beyond the boundary of the subtask modal. Neither are serious issues, only suboptimal.  See screenshot below:
<img width="500" alt="Screen Shot 2020-07-22 at 6 33 31 PM" src="https://user-images.githubusercontent.com/2006101/88239468-1fb7b200-cc4a-11ea-8a9c-e702ec8b7486.png">

2. I often use a two-finger scroll (equivalent to mouse zoom wheel) to scroll a page.  Within the subject window, that two-finger scroll is always interpreted as zoom-in/zoom-out -- even for the case where the subtask modal has a scroll bar, as is the case in the screen shot below.  

<img width="500" alt="Screen Shot 2020-07-22 at 6 39 13 PM" src="https://user-images.githubusercontent.com/2006101/88239812-0bc08000-cc4b-11ea-8ec4-ab44c2cb37a4.png">

Is it possible / preferable to disable zoom control unless selected from the tool bar on right side of subject view window?

shaunanoordin:
## PR Review

Previous titles: ~~Quick Review,~~ ~~Semi-Quick Review,~~ ~~not-so-Quick Review~~

Looks mostly good so far in terms of main functionality! I have been able to a break a few things, though, so here are the full deets.

### Functionality Tests

[...]

Issue #1716, opened 2020-07-20
Subject image invert button


srallen:
**Package**  lib-classifier

**Describe the solution you'd like**

If enabled for a workflow, this displays a toggle to CSS color invert the subject image. Currently on PFE, it's a toggle that appears below the subject image (highlighted by me with the red square):

<img width="589" alt="Screen Shot 2020-07-20 at 11 17 22 AM" src="https://user-images.githubusercontent.com/5016731/87961249-10ced500-ca7b-11ea-9b3c-8a900c07be4f.png">

I didn't find an existing design for this. @beckyrother do you want this toggle to take a similar placement and design as the below image viewer options or to be a button in the image toolbar with the pan/zoom functions?

beckyrother:
Thanks for asking. It should be in the toolbar

PR #1726, opened 2020-07-28
Classifier: Upgrade and document the withKeyZoom decorator

eatyourgreens:
- Add a Readme for the withKeyZoom component.
- Add panLeft, panRight, panUp and panDown actions to the SubjectViewer store so that withKeyZoom is independent of the implementation of the image toolbar.
- Update tests.
- Fix event handlers that should return false when cancelling the default handler.
- Replace `inject` with the MobX context provider.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
Theres's no keyboard mapping at the moment for `SubjectViewer.resetView` (or the rotation actions) but it would be straightforward to add one.

eatyourgreens:
This works for me on the drawing step of I Fancy Cats: in Chrome, after drawing an ellipse, I can zoom and pan the subject from the keyboard. However, on a transcription project (https://localhost:8080/?project=1764&workflow=3392), transcription lines swallow the keyboard input when they have focus and you are forced to use the mouse to zoom and pan the page. This happens even when the transcription popup is closed.

srallen:
@eatyourgreens could you open an issue tracking behavior with pan/zoom on the transcription task separately from this PR?

eatyourgreens:
Sure. It may already have been covered by Shaun in his review of transcription tool behaviour.

eatyourgreens:
I've added some more tests to this because the subject viewer `onPan` and `onZoom` actions weren't tested.

eatyourgreens:
I've added a note to #1719 about keyboard control from transcription lines. https://github.com/zooniverse/front-end-monorepo/issues/1719#issuecomment-665072048

eatyourgreens:
Anyone else find it confusing that we pass `type` with calls to `onZoom` but not to `onPan`? That's a holdover from the TESS light curve viewer but it confuses me every time I look at this code.

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/withKeyZoom.spec.js
```diff
@@ -13,12 +13,18 @@ describe('withKeyZoom', function () {
...
   const WithZoom = withKeyZoom(StubComponent)
-  const onPan = sinon.stub()
+  const panLeft = sinon.stub()
+  const panRight = sinon.stub()
+  const panUp = sinon.stub()
+  const panDown = sinon.stub()
   const zoomIn = sinon.stub()
   const zoomOut = sinon.stub()
   const classifierStore = {
     subjectViewer: {
```
eatyourgreens, line 23: Maybe create an instance of `subJectViewerStore` here, rather than mocking it with stubs?

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/README.md
```diff
@@ -0,0 +1,33 @@
+# withKeyZoom
+
+A decorator that connects a component to the pan and zoom actions in the `SubjectViewer` store, which handles the image toolbar functions. The wrapped component is given an `onKeyDown` handler but focus management is the responsibility of the component author, not this <abbr title="Higher-Order Component">HOC</abbr>.
```
srallen, line 3: A non-blocking comment: I'm thinking about renaming it `SubjectToolbar` since now it's wired up with the scatter plot, so once that happens, this should reference the `subject toolbar` rather than `image toolbar`.

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+  withKeyZoom(MyComponent)
+```
+or with another connected component
+```
+  @inject(storeMapper)
+  @withKeyZoom
+  @observer
+  class MyComponent { }
+```
```
srallen, line 33: I think we need to clarify that this is the older mobx-react method of decorating and show and example with v6:

```
with another connected class component using mobx-react v5

  @inject(storeMapper)
  @withKeyZoom
  @observer
  class MyComponent { }


with another connected class or functional component using mobx-react v6 (preferred)

import { observer, MobXProviderContext } from 'mobx-react'
function useStores() {
  return React.useContext(MobXProviderContext)
}

class MyComponent { 
  render () {
    const stores = useStores()
   }
}

export withKeyZoom(observer(MyComponent))

function MyFunctionalComponent ()  {
   const stores = useStores()
   return ()
}

export withKeyZoom(observer(MyFunctionalComponent))

```

Feel free to tweak my suggestion here. I couldn't write this as an actual suggestion and get the formatting right in the github markdown.

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/README.md
srallen: Maybe something about how there's a filter on allowed html tags?

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/README.md
eatyourgreens: Maybe leave out inject and write it as
```js
@withKeyZoom
@observer
class MyComponent { }
```
or
```js
withKeyZoom(observer(MyComponent))
```
depending on how you like to write decorators. Do we have a style preference for decorating classes with the `@decorator` syntax?

We maybe don't need to show how to inject properties into a component, since that would be repeating the MobX documentation.

srallen: Yeah, leaving out the example might be best since there are a few different ways it can be done and just add a note about referencing the correct mobx-react docs depending on the version with a link to it?

eatyourgreens: I can update `withKeyZoom` to replace `inject` with context too.

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/README.md
srallen: Yes, I believe that's why.

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/README.md
srallen: Sounds good. I'd like to start to move us over to the v6 way of doing things.

eatyourgreens: I think all this needs to note is that the HOC should wrap `observer` when used with observed components. Something like that. `observer(withKeyZoom(MyComponent))` won't work.

There might already be a note about decorating observers in the mobx-react docs, in which case I'll link to it.

eatyourgreens:
> Could we add documentation on the actual panning actions?

Sure. Do we have any examples of existing docs for the stores? I had a quick look, but couldn't see any Readme files.

packages/lib-classifier/src/store/SubjectViewerStore/Readme.md
```diff
@@ -0,0 +1,17 @@
+# SubjectViewerStore
+
+## Actions
+
+- **setOnPan:** register an onPan handler with the subject toolbar. Subject viewers should expose a method with the signature `onPan(dx, dy)` to enable panning.
+  - _dx_: -1 (left), 0 (ignored), 1 (right)
+  - _dy_: -1 (up), 0 (ignored), 1 (down)
+- **setOnZoom:** register an onZoom handler with the subject toolbar. Subject viewers should expose a method with the signature `onZoom(type, value)` to enable zooming.
+  - _type:_ `'zoomin'|'zoomout'|'zoomto'`
+  - _value:_ TBD
```
eatyourgreens, line 10: Does anyone know what `value` controls in the `onZoom` handler?

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/README.md
eatyourgreens: >I'm thinking about renaming it SubjectToolbar since now it's wired up with the scatter plot, so once that happens, this should reference the subject toolbar rather than image toolbar.

Good point but I don't know if we're going to use this same toolbar for video and audio. The zoom and pan functions seem like they are specific to images and charts.

srallen: True, but it's not always going to be only pan/zoom controls here. The subject color invert button is going to be added to it otherwise I'd suggest `PanZoomToolbar`. See #1716 

Audio and video may indeed have their own controls that that take the same position in the layout. This is a future discussion though.

packages/lib-classifier/src/store/SubjectViewerStore/Readme.md
srallen: It's a number and it's the optional scale value for the zoom function, defaulting to 1.2 for zooming in and 0.8 for zooming out within the zoom code of both the TESS LCV and VXZoom. I think only the TESS LCV uses `zoomto` and I'm not exactly sure what it does. The VXZoom has a zoom to point function that is self contained in that class specifically for the double click event.

eatyourgreens:
Merge conflicts should be fixed. I had a go at describing the optional number that's passed to `onZoom`. Looking at the LCV code, it only seems to be used for `zoomTo` and is ignored when zooming in and out.

srallen:
Issue documented in https://github.com/zooniverse/front-end-monorepo/issues/1739

Issue #1679, opened 2020-06-17
[RFC] Dropdown Task


srallen:
We've identified the dropdown task as being a task that will be continued to be supported, however, needs potential reworking for continued support. 

## Current implementation

The current implementation on PFE:
 - code: https://github.com/zooniverse/Panoptes-Front-End/tree/master/app/classifier/tasks/dropdown
- demo: https://master.pfe-preview.zooniverse.org/dev/classifier (pick the "Maybe select something" option)

### Functionality

The current functionality of the dropdown includes:
- Has instructions field
- Has help text field
- Setting to be required
- Setting to allow option creation which sets the dropdown task to have dual function of being a text input for free text and a select input to choose pre-defined options.
- Cascading so options selected for one dropdown affects the available options for the next dropdown. Classic example is location based, so Country > State > City, etc.
- Can be set to use a preset of options for:
  - A range of numbers
  - Months
  - Countries
  - US states
  - Canadian provinces
  - Mexican states
- Can be part of the combo task
- Can be part of a drawing sub-task

### JSON task structure

TODO: add example of actual task JSON

### JSON annotation structure

TODO: add example of actual annotation JSON

### Actual usage

These are examples of actual project usage
TODO: Link to actual projects using the dropdown in these ways:
- A short list (TODO: define short) of preset options for known potential values, often used in transcription of tabular data
- A long list (TODO: define long) of preset options for known potential values.
- A range of dates
- A location
TODO: Are there other use cases?

### Performance

- Long selection lists, particularly if there are multiple for cascading select options, can create massive task objects on the workflow resource resulting in slow loading and browser performance and slow exports
  - https://github.com/zooniverse/Panoptes-Front-End/issues/3636
  - https://github.com/zooniverse/panoptes/issues/2297

### Aggregation

- Human readable labels make extraction and aggregation more complicated, particularly because of workflow versioning, translations, as well as user submitted values
  - https://github.com/zooniverse/caesar/issues/841

## UX and design research

tl;dr select inputs / dropdowns are hard to have good UX with and alternatives like text inputs using validation, auto-complete, and/or auto-suggestion or radio buttons are often better choices

UX research resources:
- https://baymard.com/blog/drop-down-usability
- https://www.nngroup.com/articles/drop-down-menus/
- https://uxdworld.com/2018/05/06/7-rules-of-using-radio-buttons-vs-drop-down-menus/

## UX

- Limit dropdown usage by minimum and maximum number of options
[...]

PR #1732, opened 2020-07-30
Add a story about a text task with suggestions

srallen:

Package: lib-classifier

For #1719

Describe your changes:
I added a story for the `TextTask` to demo the behavior when there are suggestions. Right now suggestions can only come in via the `Mark` model if it has `subTaskPreviousAnnotationValues` defined and passed down as a prop from the store connection (when we build out getting suggstions for the new dropdown variant, maybe this prop will be renamed or another added?). I added this to demo the behavior in response to this comment from the text transcription tools alpha testing:

> [ui] the list of "suggested text" can, however, block the Save button, especially when I choose to ignore the list and type something on my own. (Note: this doesn't happen when the popup is positioned near the bottom of the screen, in which case the list appears above the textbox)

Currently if the text task has suggestions and you initially focus on the input, the suggestions menu opens and does not close if you start typing something else. This is because this mostly functions like a native HTML datalist element (can be previewed here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist). What you're typing is compared to the suggestions for a match, which I generally think is desirable behavior and something we want for the long list of options that is asynchronously loaded in lieu of a massive dropdown (what we're discussing in #1679). The difference between the Grommet component and the native datalist element that I can find is that if there isn't an immediate match, the native element no longer displays the suggestions.  Grommet's component expects you to hit enter or esc to close the suggestions menu. A couple of gifs to demo this:

No match, where you must hit esc or enter:
suggestions-no-match

Match, highlights the matched suggestion, then you hit enter (I think this is desirable UX):
suggestions-match

I'm not sure if we should make a UX change, but this at least adds a story where we can demo it.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
The prop on the Grommet `TextInput` is suggestions, so I'm thinking that we may want to allow these suggestions to come in from various sources whether it's previous annotations values via caesar or from the async result against a very long list supplied for the task by the project builder. That's a future feature though maybe we do as a result of the dropdown task discussions.

PR #1738, opened 2020-08-05
Make mark UX consistent and fix associated keyboard events

srallen:

Package: lib-classifier

Describe your changes:
I'm working on doing some of the debugging for the reports from #1719 and came across some code related to the onSelect, onDeselect, and associated DOM events onFocus and onBlur that don't make sense and/or didn't work.

Here's how tabbing through the transcribed lines when there's a new line works currently as seen in the TranscribedLines story:

unfixed-onselect

It's unclear when the new mark is focused and no longer focused. It doesn't have an outline like the transcribed lines do and onblur does nothing.

Here's how this same interaction is on this branch:

fixed-onselect

When the new line mark is focused, it now has an outline consistent with the transcribed lines. The outline color is consistent and now onblur deselects the line mark. I accomplished this by:

- Define an `onDeselect` handler in the `InteractiveLayer` component
- Move the `onBlur` event to the delete button. If left on the mark component, then once you hit tab to try to get to the delete button, the handler for deselecting the mark fires and then since the mark is no longer active, the delete button is no longer rendered, and thus you can never get to it if using the keyboard. 
- Add the `withTheme` HOC to `DeleteButton` and `Mark` to add the focus outline styles we use elsewhere

A few other tweaks to note:
- Removed the `observer` HOC from `Mark` component's export because it's being wrapped by this same HOC in `DrawingToolMarks`, so it's being wrapped twice. This is unnecessary.
- Changed the `tabIndex` attributes to use strings instead of numbers where applicable. I think this DOM attribute is supposed to be set to strings.
- Removed `onSelectMarkWithRef` because the ref is never used by the `onSelectMark` handler.

For the select --> make mark active in store, deselect --> inactivate mark in store integration, we do not have any existing tests for this. I think to test this, I'd like to connect `DrawingToolMarks` to the store directly, but I'm going to consider this out of scope since this would be a fairly big change and no tests exist currently. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
@beckyrother one question I have from this is that the color we use for the focus outline and the color for the active new mark when in the transcription task is really close and maybe too similar. Have a look at the second gif and let me know what you think.

beckyrother:
@srallen good point. Should we try something like #06FE76 for the active new mark instead?

srallen:
@beckyrother here's how green looks in context:

Screen Shot 2020-08-07 at 2 53 33 PM

And in action with tabbing through:

green-line

If this is ok, then I'll push the change up for code review

beckyrother:
I like the green but it's very hard to see any of those lines. Can they be 2px instead of 1px, I think that will help a lot. Sorry for continued changes.

srallen:
@wgranger it seems that the outline isn't working for any SVG group element in Safari and it's an issue on master, so is it ok we document this as a separate issue? I'm not really sure what's happening, but it seems like a browser bug. Some time will have to be spent to figure out a work around. 

@beckyrother here's what a stroke width of 2 looks like:

Screen Shot 2020-08-13 at 3 14 33 PM

While checking this out today, I discovered my branch is breaking the delete button from being dragged with the rest of the mark somehow, so I'm going to investigate that.

eatyourgreens:
I had a similar problem with re-rendering the delete button on drag, which I fixed by wrapping marks in `observer`. I don’t know if that’s the same problem here.

srallen:
Ooh, thank you. That was a super helpful hint. I had removed an observer that appeared redundant. My note from the OP:

> Removed the observer HOC from Mark component's export because it's being wrapped by this same HOC in DrawingToolMarks, so it's being wrapped twice. This is unnecessary.

I wrapped the `DeleteButton` with an observer and this fixed it. Now it's more clear where and why the observers are being used. I wonder if there's an integration test that could be written to test for this.

beckyrother:
Heck yes that 2px looks so much better

srallen:
Documented Safari focus outline issue: https://github.com/zooniverse/front-end-monorepo/issues/1757

PR #1740, opened 2020-08-06
SVGZoom: fix horizontal panning

### LINKED ISSUE #1739: Panning image vs scatter plot
**Package**

**Describe the bug**
How the pan up and pan down events work is reverse between the two viewer types. They use different pan/zoom implementations, so the event handlers in `SubjectViewerStore` do not behave consistently between the two.

eatyourgreens:
Test all four pan directions for the SVGZoom component. Reverse the x direction when panning, so that panning left and right moves the image viewport in the same direction as the key that was pressed.

Package:
lib-classifier

Closes #1739.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
I've copied the directions from Seabird Watch. It should be the same now, between the monorepo and PFE. The arrow keys move the zoomed viewport over the image, so up moves the viewport up etc.
https://www.zooniverse.org/projects/penguintom79/seabirdwatch/classify

eatyourgreens:
Double-checking against https://frontend.preview.zooniverse.org/projects/penguintom79/seabirdwatch/classify and left/right is not the same. You're right about changing the handler globally in the store, but do we want to change the controls for PH-TESS?

srallen:
They're reverse all around from what I can tell. Here's image viewer vs seabird watch using up, down, left, and right in that order:

panzoom-story

seabird-panning

srallen:
No, sorry, looking again, it's just left and right. My brain is not parsing this well. I think partly because when looking at this with the scatter plot, it does *not* make sense to me to have it pan like this.

srallen:
I think this is an example of needing different behaviors depending on the subject viewer?

srallen:
We could return to the refactor I had initially tried and pass in a `type` and optional `value` similar to the zoom function and each viewer can handle it differently as needed?

eatyourgreens:
Adding `type` makes sense to me, and would make `onPan` consistent with `onZoom(type, value)`. I _think_ the subject viewer API should stick with the SVG convention that negative is left/up and positive is right/down, so `pan(0, -1)` is up. How authors handle those values internally in subject viewers probably comes down to what seems natural/intuitive for that viewer.

The left/right scroll in the PH-TESS seems intuitive to me, with the cursor keys. Panning around the Seabird Watch image also seems natural to me, but that could be down to repeated use of PFE. Looks like the directions are the same for both those, so it's the image viewer that's reversed in the monorepo, compared to both PH-TESS and Seabird Watch?

I'm also finding it really hard to tell the difference between left and right after trying this a few times, with different viewers/projects.

I'll update this to fix the image viewer so that it's consistent with PFE and PH-TESS. I'm not sure about flipping the values in the store, because I tend to think of negative direction changes as left/up in SVG. I've not got a strong opinion about that. Do you think flipping them would make it harder to reason about which direction a change is?

eatyourgreens:
The scatter plot viewer here uses the same convention as PFE and PH-TESS, so I think fixing the image viewer to use that convention too would make sense. https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/index.html?path=/story/subject-viewers-scatterplotviewer--pan-and-zoom

srallen:
The currently deployed story for the scatterplot has the panning behavior I'd expect for the scatter plot. This branch reverses the behavior, which I think is confusing.

This is using arrow keys up, down, left, right in that order:
scatterplot-pan-keyboard

And I think I got confused when going between the image subjects and the scatter plot subjects as well trying to compare the behavior, because the way they pan or the way we might want them to pan is seemingly opposite? Now I'm wondering how this interaction is going to be with the `DataImageViewer` having a scatterplot and an image side-by-side.

eatyourgreens:
That’s weird, because this PR doesn’t touch the scatter plot viewer. On this branch, I see the scatter plot viewer and the TESS LCV behave the same way when panning with the keyboard.

eatyourgreens:
OK, I've rewritten this to test all four pan directions, then updated the component to do the maths correctly and pass the tests.

I'm not sure why the scatter plot viewer moves in the opposite direction to the image viewer and the LCV. Looking at the LCV code, it translates the d3 zoom layer by `xMultiplier * PAN_DISTANCE`.
https://github.com/zooniverse/front-end-monorepo/blob/5d89bf72d71a9fbe86419b65cc7396523a2cf9e5/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js#L385-L387

Does the scatter plot viewer translate the zoom layer or the image? If it's the image, I think you want it to translate right when pressing left, up when pressing down and so on. So subtract the passed translation rather than adding it.

srallen:
There's no image with the scatter plot viewer, so that not applicable. The data points themselves get translated within the svg group.

That being said, here's the handler for the key zoom integration:
https://github.com/zooniverse/front-end-monorepo/blob/5d89bf72d71a9fbe86419b65cc7396523a2cf9e5/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VXZoom/VXZoom.js#L72-L85

Subtracting rather than adding does fix it, but expectedly tests break now. I'm happy to make this change myself in a separate PR since I think I wrote this when I was still confused about how this was all working anyway.

eatyourgreens:
Odd that the LCV code adds the same translation as the scatter plot yet moves in the opposite direction. I didn't really delve into how `d3.zoom()` works, so maybe it handles reversing the direction somewhere.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.spec.js
```diff
@@ -50,18 +50,40 @@ describe('Components > SVGPanZoom', function () {
...
+        expect(viewBox).to.equal('0 -10 400 200')
+      })
+    })
+    describe('down', function () {
+      it('should move the viewbox down', function () {
+        onPan(0, 1)
+        wrapper.update()
+        const viewBox = wrapper.find('svg').prop('viewBox')
+        expect(viewBox).to.equal('0 10 400 200')
+      })
```
srallen, line 85: Should each of these have expectations on what the viewBox was previously?

eatyourgreens: Yes! Good catch.

PR #1742, opened 2020-08-06
Classifier Dropdown Task pt1 (v2) - Task and Annotation models

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Associated RFC issue: #1679 
Associated Project: Engaging Crowds

This is part of a series of PRs that introduces the Dropdown Task to the monorepo Classifier

- Part 1 introduces Task and Annotation models for the dropdown task.
- Part 2 introduces the dropdown React component.

For this PR, the goal is to simply be _backwards compatible_ with the dropdown data models being created by PFE/the Zooniverse project builder.

As stated in #1679, there is currently a discussion on how dropdown tasks _should_ be designed moving forward, so expect more design divergences in Part 2.

### Dev Notes

The greatest complication is that classic PFE dropdown tasks have a built-in branching system (separate from the classifier's branching task system). This leads to _shenanigans_ in how the task data is structured (which we won't change, for now), how the annotation data is structured (I'll _try_ not to change it), and how the component is designed (nyerrr).

Dropdown task data structure (simple single dropdown per task/non-cascading selection!), example:

```
"T0":{
   "help":"",
   "next":"T1",
   "type":"dropdown",
   "selects":[
      {
         "id":"070b610fbf5d9",
         "title":"Alignment",
         "options":{
            "*":[
               {
                  "label":"Lawful Good",
                  "value":"4beef18e9baa"
               },
               {
                  "label":"True Neutral",
                  "value":"db939237aa00f"
               },
               {
                  "label":"Chaotic Evil",
                  "value":"fedae0dd4f96d"
               }
            ]
         },
         "required":false,
         "allowCreate":true
      }
   ],
   "instruction":"Dropdown with an 'OTHER' choice - type whatever!"
},
"T1":{
   "help":"",
   "type":"dropdown",
   "selects":[
      {
         "id":"dbd3d991fcd9e",
         "title":"Colour",
         "options":{
            "*":[
               {
                  "label":"Red",
                  "value":"7fda9846fd624"
               },
               {
                  "label":"Yellow",
                  "value":"a2c9fbc881d7a"
               },
               {
                  "label":"Green",
                  "value":"baf1ae42d766f"
               },
               {
                  "label":"Blue",
                  "value":"6d13daea5706a"
               }
            ]
         },
         "required":true,
         "allowCreate":false
      }
   ],
   "instruction":"Dropdown with limited choices"
}
```

Dropdown annotation (classification) data structure, example:

```
"annotations":[
    {
       "task":"T0",
       "value":[
          {
             "value":"THIS IS A FREE ANSWER",
             "option":false
          }
       ]
    },
    {
       "task":"T1",
       "value":[
          {
             "value":"baf1ae42d766f",
             "option":true
          }
       ]
    }
 ]
```

### Status

WIP


shaunanoordin:
## PR Update

This PR has been updated so that the the Task and Annotation data models for the Dropdown Task are "working" in the PFE sense.

- the Task model is now in the correct data format, as when reading from a Panoptes API resource.
- the Annotation model is now in the "correct" data format, as when submitting to Panoptes as a Classification resource. (By "correct" I actually mean "backwards compatible", i.e. the same as what you'd get if you submit via a PFE project.)
-  There is a very basic Dropdown component that lets you try things out in a functioning UI.
  - ⚠️ The proto-component is styled after Single Answer components, doesn't allow chaining, ignores "is required to proceed" flag, and ignores the "use can create custom answers" flag.
- As usual, all my tests and stories are still in a state best described as craaaappppp 🔥 

Recommended testing project:
- `https://local.zooniverse.org:8080/?env=production&project=12754&workflow=15764` (lib-classifier)
- Project has 2 steps/tasks; Task 1 is a dropdown with 3 options (required: no, custom answer: yes), Task 2 is a dropdown with 10 options (required: yes, custom answeR: no)

shaunanoordin:
## PR Update

- Tests added for Dropdown Task and Annotation models.
- Readme added for the Dropdown Task plugin
- Tests & story for Dropdown Task _component_ have been temporarily disabled

Oh hey, this means that this PR is done! This branch is now ready for review, though please ignore anything in the Dropdown Task component folder. That'll be addressed in PR Part 2: Electric Boogaloo: 2 Pull 2 Request: This Time It's PRsonal

shaunanoordin:
### Status (Clarification)

Ready for review, but NOT ready to merge! There's a Part 2 coming up.

shaunanoordin:
@srallen I'm down with creating a `dropdown-simple` task (wait, would the component be called a `SimpleDropdownTask` or `DropdownSimpleTask`?) that's _utterly separate and not backwards-compatible_ with the classic `dropdown` task.

I'll add more thoughts in [ADR29](https://github.com/zooniverse/front-end-monorepo/pull/1767), but basically, _if_ we go with `dropdown-simple`:

- I'll spin off a new branch (`v3`) to accommodate the more simplified Task and Annotation data models of the `dropdown-simple` task.
  - This will only take me a few hours since I'm just snipping out code from `classifier-dropdown-pt1-v2` and `classifier-dropdown-pt2-v2`, so this is cool - the minimal extra effort is worth the cleaner code for everyone.
- ⚠️ The only real dev effort is building the `dropdown-simple` task into **Project Builder.**
  - I'm just going to assume the code for that is written in CoffeeScript, backwards, in Arabic, in a Vignere cipher, so I have no idea how much time this will actually take. 🤷‍♂️ I should look into this.
  - (Project Builder is reason 1 why I built the "simple dropdown" implementation on top of the classic `dropdown` task, btw)
- I will however insist on keeping the current `dropdown` task code around in some form (and separate in responsibility from the `dropdown-simple` task), because the current code was built with **backwards compatibility** in mind.
  - i.e. I believe the `v2` branches still have value, even if I'm cool to spin up a `v3` branch.
  - Yeah, I tried to make sure the code could eventually re-support cascading dropdowns, even though nobody really uses it according to the stats, and assuming I have extra dev time on my hands somehow.
  - My conversations with @camallen informs me that backwards compatibility has significant priority - maybe this is a difference in opinion between the Oxford dev ("we eventually want to port all PFE projects to the monorepo, which means accommodating old data models") and Chicago dev ("monorepo projects should be built on sensible data models, and not spend too much time supporting old wonky bits")? Might be worth a chat.
  - (Backwards Compatibility is reason 2)

camallen:
> My conversations with @camallen informs me that backwards compatibility has significant priority - maybe this is a difference in opinion between the Oxford dev ("we eventually want to port all PFE projects to the monorepo, which means accommodating old data models") and Chicago dev ("monorepo projects should be built on sensible data models, and not spend too much time supporting old wonky bits")? Might be worth a chat.

Not necessarily "significant priority" rather, we need a way to support our existing projects to port to the monorepo code base. That can either be
- switch your workflow task type to the new 'simple dropdown' as long as this supports the research outcomes of the project. 
- provide a backwards compatible 'dropdown' task from ported PFE but with the great UX feature additions in the monorepo

If we can't port these projects then we have to run both PFE & Monorepo code bases in perpetuity which was never a desired outcome of the rebuild decision. 

The work we're doing with engaging crowds provides an opportunity to add new tasks to the monorepo. This work, while focused on UX deliverables for the grant, can also help us:
- provide a pathway to move existing projects to the monorepo codebase with it's improved UX features
- deprecate PFE as quickly and as much possible (looking at you `/lab`)

All while balancing the research outcomes of the grant projects.

In my opinion, this engaging crowds work can help us with project portability via a ported tool for the dropdown which would technically enable some projects to move to monorepo codebase. 

I don't really mind the implementation details TBH but I feel that we should carefully consider providing task types that enable project portability from PFE to the monorepo. 

It seems that Shaun has done the majority of work for a parity dropdown tool type. If it's not much more work to add the simple version, perhaps we could include both dropdown versions? Perhaps only allowing existing projects access to the backwards compatible version, much like we do experimental tools?

eatyourgreens:
Small thing (I think) but remember to add translations strings for the new task model in Panoptes too.

eatyourgreens:
Sorry, one other thought. If the mapping is straightforward, then a snapshot preprocessir can be used to convert the old task JSON to the new task model. That would allow the Engaging Crowds projects to continue editing their tasks in the project builder, without manual intervention.

shaunanoordin:
> If it's not much more work to add the simple version, perhaps we could include both dropdown versions?

@camallen @srallen Looks like we're in agreement - I'll spin up the `v3` branch tomorrow (Wed 26 Aug), to implement the `dropdown-simple` task as currently described, and it'll be updated to match any updates in [ADR29](https://github.com/zooniverse/front-end-monorepo/pull/1767). 👍 

> Perhaps only allowing existing projects access to the backwards compatible version, much like we do experimental tools?

Sold!

I'll keep `v2` branch(es) around as a separate development track for the backwards-compatible `dropdown`. (Note: the preprocessor solution will be unnecessary in this scenario, but thanks nonetheless for the suggestion!)

srallen:
> Perhaps only allowing existing projects access to the backwards compatible version, much like we do experimental tools?

I don't think we need to do this. The PFE version can remain as a task type of `dropdown` which loads up the correct models, and the simple version should be `dropdown-simple`, so they will be completely separate task types set in the workflow. Then in the future, we'll be adding the more complex variants as their own types like `date`, `dropdown-cascading`, `dropdown-location`, `text-async` etc. 

Does that make sense @shaunanoordin? 

That being said, I really don't think we should support `dropdown` and the path forward is to have projects that migrate to update their workflow tasks to fit the new task options.

camallen:
> The PFE version can remain as a task type of dropdown which loads up the correct models, and the simple version should be dropdown-simple, so they will be completely separate task types set in the workflow. Then in the future, we'll be adding the more complex variants as their own types like date, dropdown-cascading, dropdown-location, text-async etc.

As you say, we want folks to switch to the new simple dropdown task type but i'm guessing there will be long lived workflows / projects that just won't switch and we don't want them to be 'stuck' on the PFE codebase. 

If we don't provide a compatible dropdown task in the monorepo then I think at some some point we should disable this task type in the lab to stop folks selecting this task type for their workflows and being locked into PFE.

I see the monoreopo, compatible dropdown task as a stepping stone for project migration to the monorepo code. If we can provide this then great, folks can go on using the same workflow task definitions (albeit with all the drawbacks it has) and still get off of PFE to the monorepo code.

> That being said, I really don't think we should support dropdown and the path forward is to have projects that migrate to update their workflow tasks to fit the new task options.

I 100% agree we should promote projects to move to the new codebase and off PFE. However we need to provide these teams with an easy way to do this. In my opinion, a compatible tool type is this pathway.

srallen:
> i'm guessing there will be long lived workflows / projects that just won't switch

I don't think we should allow this. We've never promised projects the classifier and task types would be permanently available. 
 I think the pathway forward is to deprecate, give them an end date, and the tools to migrate via the lab.

shaunanoordin:
## PR Update

In regards to the Engaging Crowds project, this PR is now superseded by #1772 (Simple Dropdown Task)

In regards to creating a backwards-compatible Dropdown Task, I'm marking this PR as an experiment and switching it back to a draft to prevent accidental merging.

srallen:
@shaunanoordin would it be ok to close this and document in an issue what would be involved with implementing a PFE like dropdown task? I just would like the PR queue cleaned up a bit. Thanks!

shaunanoordin:
👌 Can do.

This PR will now be closed to clear the PR queue; dev notes for resurrecting classic `dropdown` support will be compiled elsewhere.

Issue #687, opened 2019-04-10
Branching in the single choice task isn't supported


rogerhutchings:
**Package**

lib-classifier

rogerhutchings:
Time estimate: 1 dev, 1 week

srallen:
Workflow steps ADR has an example of how branching logic will be supported with workflow steps: https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-05.md#decision

srallen:
This also can be developed against the kitteh zoo staging project as the first step/task should be a branching single choice question.

mcbouslog:
I've created draft PRs for two of likely a few approaches:

#1747 - adds `next` property to `Step`, similar to what intended for recursive workflows as stated in https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-05.md#decision
- Note: as expected with MST/observables, TaskNavButtonsContainer doesn't rerender on a change to the active step next property, as the active step is still the same, the steps haven't changed, so why would it rerender, but I want it to rerender! [console.logging `step.next` forces a rerender](https://github.com/zooniverse/front-end-monorepo/pull/1747/files#diff-491dac2863aa29086de04a585966846fR80), but that's not proper

#1748 - adds `next` to `workflowStepStore`

Going Back/Previous breaks either at the moment.

In this test workflow - https://master.pfe-preview.zooniverse.org/lab/1693/workflows/3428, it should go:
single q (0, 1, 2)
 0 - complete
 1 - drawing, then complete
 2 - text, then complete
However, after the drawing task it advances to the text task, because the workflow step store does not currently account for the next task setting (on any task, including the branching of the single question task). Is this what we refer to as a recursive workflows, and is therefor out of scope of this issue? That's fine, I can just focus on advancing from a single choice question, but they're (single choice branching, any task next's setting, going back/previous) so interrelated I'm not sure if should be considered at same time?

srallen:
After discussing this with @mcbouslog, we've realized that once a branching has occurred, we will always need to explicitly know what step is next and previous. Say a branching workflow looks like this:

|Step A|
|           |
|           |
[B]       [C]
|           |
[D]      [E]

And the step map looks like this:
[[A],[B],[C],[D],[E]]

We can't rely on the map order once the branching happens.

So, we need to tweak how this works and define a `next` and `previous` for all steps:

- by the action that converts tasks to steps
[...]

PR #1758, opened 2020-08-18
Refactor Step and WorkflowStepStore for branching

mcbouslog:
Package: lib-classifier

Towards #687, additional work on previous needed to close.

Describe your changes:
- refactor Step with `next` and `previous` properties and related actions to set
- refactor Step with `isThereBranching` view if there's a single choice task with branching including in step tasks
- refactor WorkflowStepStore to remove `isThereANextStep` and `isThereAPreviousStep` views now included in Step
- add WorkflowStepStore annotation observer if branching to update next step accordingly
- refactor WorkflowStepStore tasks conversion to steps for next, previous, and if combo task comes after tasks included in combo
- refactor TaskNavButtonsContainer to remove `isThereANextStep` and `isThereAPreviousStep` views now included in Step
- update tests

branching with combo - https://localhost:8080/?project=1693&workflow=3428

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/Step.js
+++ b/packages/lib-classifier/src/store/Step.js
@@ -34,7 +34,10 @@ const Step = types
       // so we should disallow that in the editor
       const [singleChoiceTask] = self.tasks.filter(task => task.type === 'single')
       return singleChoiceTask?.answers.some((answer, index) => {
-        return answer?.next && singleChoiceTask.answers[index + 1]?.next && singleChoiceTask.answers[index + 1].next !== answer.next
+        if (singleChoiceTask.answers.length > index + 1) {
+          return answer.next !== singleChoiceTask.answers[index + 1].next
+        }
+        return false
       })
     }
   }))
--- a/packages/lib-classifier/src/store/Step.js
+++ b/packages/lib-classifier/src/store/Step.js
@@ -7,8 +7,10 @@ function taskDispatcher (snapshot) {
 
 const GenericTask = types.union({ dispatcher: taskDispatcher }, ...taskModels)
 
-export const BaseStep = types
-  .model('BaseStep', {
+const Step = types
+  .model('Step', {
+    next: types.maybe(types.string),
+    previous: types.maybe(types.string),
     stepKey: types.identifier,
     taskKeys: types.array(types.string),
     tasks: types.array(GenericTask)
@@ -16,24 +18,35 @@ export const BaseStep = types
   .views(self => ({
     get isComplete () {
       return self.tasks.reduce((isStepComplete, task) => isStepComplete && task.isComplete, true)
+    },
+
+    get isThereANextStep() {
+      return self.next && self.next !== 'summary'
+    },
+
+    get isThereAPreviousStep() {
+      return self.previous && self.stepKey !== 'summary'
+    },
+    get isThereBranching () {
+      // We return the first single choice task
+      // It doesn't make sense to have more than one single choice task in a step
+      // so we should disallow that in the editor
+      const [singleChoiceTask] = self.tasks.filter(task => task.type === 'single')
+      return singleChoiceTask.answers.some((answer, index) => { 
+        return answer.next && answer[index + 1].next && answer[index + 1].next !== answer.next
+      })
     }
   }))
   .actions(self => ({
     reset () {
       self.tasks.forEach(task => task.reset())
-    }
-  }))
-
-export const NextStepReference = types
-  .model('NextStepReference', {
-    next: types.maybe(types.string) // We have this optionally to support recursive workflows, temp edit to types.string - not working as expected otherwise, though should work as types.reference
-  })
-  .actions(self => ({
+    },
     setNext (nextStepKey) {
       self.next = nextStepKey
+    },
+    setPrevious (previousStepKey) {
+      self.previous = previousStepKey
     }
   }))
-
-const Step = types.compose('Step', BaseStep, NextStepReference)
 
 export default Step
--- a/packages/lib-classifier/src/store/Step.js
+++ b/packages/lib-classifier/src/store/Step.js
@@ -21,11 +21,11 @@ const Step = types
     },
 
     get isThereANextStep () {
-      return self.next && self.next !== 'summary'
+      return !!self.next && self.next !== 'summary'
     },
 
     get isThereAPreviousStep () {
-      return self.previous && self.stepKey !== 'summary'
+      return !!self.previous && self.stepKey !== 'summary'
     },
 
     get isThereBranching () {
--- a/packages/lib-classifier/src/store/Step.js
+++ b/packages/lib-classifier/src/store/Step.js
@@ -20,20 +20,21 @@ const Step = types
       return self.tasks.reduce((isStepComplete, task) => isStepComplete && task.isComplete, true)
     },
 
-    get isThereANextStep() {
+    get isThereANextStep () {
       return self.next && self.next !== 'summary'
     },
 
-    get isThereAPreviousStep() {
+    get isThereAPreviousStep () {
       return self.previous && self.stepKey !== 'summary'
     },
+
     get isThereBranching () {
       // We return the first single choice task
       // It doesn't make sense to have more than one single choice task in a step
       // so we should disallow that in the editor
       const [singleChoiceTask] = self.tasks.filter(task => task.type === 'single')
-      return singleChoiceTask.answers.some((answer, index) => { 
-        return answer.next && answer[index + 1].next && answer[index + 1].next !== answer.next
+      return singleChoiceTask?.answers.some((answer, index) => {
+        return answer?.next && singleChoiceTask.answers[index + 1]?.next && singleChoiceTask.answers[index + 1].next !== answer.next
       })
     }
   }))
--- a/packages/lib-classifier/src/store/Step.js
+++ b/packages/lib-classifier/src/store/Step.js
@@ -44,6 +44,10 @@ const Step = types
   .actions(self => ({
     reset () {
       self.tasks.forEach(task => task.reset())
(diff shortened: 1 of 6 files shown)


packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -79,13 +72,41 @@ const WorkflowStepStore = types
...
 
+    function createAnnotationObserver () {
+      const annotationDisposer = autorun(() => {
+        if (self.active?.isThereBranching) {
+          // presumes one single choice task per step
+          const [singleChoiceTask] = self.activeStepTasks.filter(task => task.type === 'single')
+          onAction(getRoot(self), (call) => {
+            if (call.path.endsWith(singleChoiceTask.annotation?.id) && call.name === 'update') {
+              let nextStepKey
+              const singleChoiceTaskAnswers = toJS(singleChoiceTask.answers)
```
srallen, line 83: It's unclear why the conversion here is necessary. I checked and directly using `singleChoiceTask.answers[call.args[0]].next` is working for me. Is there a case when this wouldn't?

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -79,13 +72,41 @@ const WorkflowStepStore = types
...
+              const nextTaskKey = singleChoiceTaskAnswers[call.args[0]].next
+              self.steps.forEach(step => {
```
srallen, line 85: This is only necessary when the single choice task is still the classifier 1.0 setup, that next is a task key rather than a step key. Maybe it can be like this:
```
              let nextStepKey
              const nextKey = singleChoiceTask.answers[call.args[0]].next
              if (nextKey.startsWith('T')) {
                // Backwards compatibility
                self.steps.forEach(step => {
                  if (step.taskKeys.includes(nextKey)) {
                    nextStepKey = step.stepKey
                  }
                })
              } else {
                nextStepKey = nextKey
              }
              self.active.setNext(nextStepKey)
            }
```

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -150,41 +171,81 @@ const WorkflowStepStore = types
...
           stepKey: 'S0',
           taskKeys: [first_task]
         }
 
-        if (workflow.tasks[first_task].type === 'combo') {
-          const combo = workflow.tasks[first_task]
-          firstStep.taskKeys = getStepTasksFromCombo(combo)
+        if (firstTask.type === 'combo') {
+          const combo = firstTask
+          firstStep.taskKeys = combo.tasks
```
srallen, line 196: This could be shorten to:
```suggestion
          firstStep.taskKeys = firstTask.tasks
```

packages/lib-classifier/src/store/Step.spec.js
```diff
@@ -104,4 +104,56 @@ describe('Model > Step', function () {
...
+        // SingleChoiceTaskFactory defaults to a branching single choice task (answers with different next values)
+        SingleChoiceTask.TaskModel.create(SingleChoiceTaskFactory.build({ taskKey: 'T2', required: false }))
+      ]
+    })
+
+    it('should have isThereBranching return true', function () {
+      const step = Step.create({ stepKey: 'S1', taskKeys: ['T1', 'T2'], tasks })
+      expect(step.isThereBranching).to.be.true()
+    })
+  })
```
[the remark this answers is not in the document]
mcbouslog: I included the above examples and a test expecting `isThereBranching` to be false if all answers' `next` is not defined, so several failing tests as desired.

Subsequent commit refactors `isThereBranching` resulting in passing tests.

srallen:
Just to note for myself that this is currently set to merge into the branch for #1768.

packages/lib-classifier/src/store/Step.js
```diff
@@ -7,28 +7,50 @@ function taskDispatcher (snapshot) {
...
+        if (singleChoiceTask.answers.length > index + 1) {
+          return answer.next !== singleChoiceTask.answers[index + 1].next
+        }
+        return false
+      })
     }
   }))
   .actions(self => ({
     reset () {
       self.tasks.forEach(task => task.reset())
```
srallen, line 46: The step reset here also needs to reset the next and previous value for a step with branching logic. 
```suggestion
      self.tasks.forEach(task => task.reset())
      if (self.isThereBranching) {
        self.setNext(undefined)
        self.setPrevious(undefined)
      }
```

I found this issue when testing with Kitteh Zoo. The state of the previous annotation selection for which step is next remained with subsequent subjects, thus causing the 'next' button to display at the start instead of 'done', which is the current behavior. 

Apparently the spec file is missing a test for the reset action, so this needs to be added.

packages/lib-classifier/src/store/Step.spec.js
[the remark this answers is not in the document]
srallen: This looks great. Could we add one more test that tests for backwards compatibility where the `next` property is set to a task like `'T2'` rather than a step? This will test the conversion to steps route.

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -18,35 +26,19 @@ const WorkflowStepStore = types
...
 
     get shouldWeShowDoneAndTalkButton () {
-      const isThereANextStep = self.isThereANextStep()
       const workflow = tryReference(() => getRoot(self).workflows?.active)
       const classification = tryReference(() => getRoot(self).classifications?.active)
 
       if (workflow && classification) {
         const disableTalk = classification.metadata.subject_flagged
-        return !isThereANextStep &&
+        return !self.active.isThereANextStep &&
```
srallen, line 41: Just for safety, let's check for a valid reference. See above comment for the addition.

```suggestion
        return step.isThereANextStep &&
```

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -18,35 +26,19 @@ const WorkflowStepStore = types
...
     get isThereTaskHelp () {
       const tasks = self.activeStepTasks
 
       return tasks.some(task => task.help)
     },
 
     get shouldWeShowDoneAndTalkButton () {
-      const isThereANextStep = self.isThereANextStep()
       const workflow = tryReference(() => getRoot(self).workflows?.active)
       const classification = tryReference(() => getRoot(self).classifications?.active)
```
srallen, line 38: Let's check for a valid step reference. Used below.
```suggestion
const step = tryReference(() => self.active)
```

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -18,35 +26,19 @@ const WorkflowStepStore = types
...
 
       if (workflow && classification) {
```
srallen, line 39:
```suggestion
      if (workflow && step && classification) {
```
Added to the conditional check for valid references.

mcbouslog:
Ah yes that does make sense! I'll add more comments inline, but please let me know what you think.

packages/lib-classifier/src/store/WorkflowStepStore.spec.js
```diff
@@ -173,6 +173,82 @@ describe('Model > WorkflowStepStore', function () {
...
+    it('should generate a step from the combo task', function () {
+      const { workflowSteps } = rootStore
+      const comboStep = workflowSteps.steps.get('S1')
+      expect(comboStep.taskKeys).to.deep.equal(['T2', 'T3'])
+    })
+
+    it('should define the next step as appropriate', function () {
+      const { workflowSteps } = rootStore
+      const firstStep = workflowSteps.steps.get('S0')
+      expect(firstStep.next).to.equal('S1')
```
mcbouslog, line 244: This particular expectation was failing without the additions (subsequent commit) to WorkflowStepStore to check if the task was a single choice task that doesn't have branching, and if that is true, then the next attribute can be set accordingly (initially to the task, converted to the step), if false then next can be undefined and it will be defined once an answer is selected via the annotation observer.

packages/lib-classifier/src/store/WorkflowStepStore.spec.js
```diff
@@ -173,6 +173,82 @@ describe('Model > WorkflowStepStore', function () {
     })
   })
 
+  describe('when the workflow tasks define next steps as tasks (backwards compatibility)', function () {
```
mcbouslog, line 176: This largely copied from the previous describe block, not sure the following tests are really necessary, erred on side of including, though happy to remove if prefer:
- it('should assign each task to a step'
- it('should set the first step to be active'
- it('should generate a step from the combo task'

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -199,6 +209,10 @@ const WorkflowStepStore = types
           firstStep.taskKeys = firstTask.tasks
         }
 
+        if (firstTask.type === 'single' && !isThereBranching(firstTask)) {
```
mcbouslog, line 212: I don't love the `!isThereBranching` with the `!`, but consistent with the related code in Step? Open to suggestion.

srallen: One time I miss coffeescript's style of `not isThereBranching firstTask` :smile:. I'm ok with this. It could be set to a variable for readability like `const isFirstSingleChoiceTaskNotBranching = firstTask.type === 'single' && !isThereBranching(firstTask)`. That's really verbose, but readable?

mcbouslog: I like the improved readability, there's a lot going on in this `convertWorkflowToUseSteps`, making it easier to understand++

PR #1768, opened 2020-08-24
Refactor task required property, Step and WorkflowStepStore for branching

mcbouslog:
Package: lib-classifier

Describe your changes:
- the single choice question, multiple choice question, and text task `required` property becomes `"true"` or `""` (strings) when toggled in the Project Builder
- the [single choice question](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/single/index.jsx#L65-L72) and [multiple choice question](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/multiple/index.jsx#L64-L71) do not define `required` in their `getDefaultTask` method
- the text task [defines `required` as `false` (boolean)](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/text/index.jsx#L144) in it's `getDefaultTask` method
- this PR refactors the single choice question, multiple choice question, and text task to support a `required` property of `"true"` or `""`

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR fixes an issue in the Classifier where a Task that's `.required` can't be rendered.

- Root cause: when the `task.required` value is toggled on/off in the PFE Project Builder, it sets `task.required = "true"` or `task.required = ""` (a string!), whereas the MST store expects a _boolean._
  - Note: if `task.required` is never touched in PFE, the `.required` value simply doesn't exist, which the MST store knows how to handle.
- Current solution: change the Task model (`packages/lib-classifier/src/plugins/tasks/models/Task.js`) so `.required` supports string _and_ booleans. 
  - Also, every Storybook Story and spec.js has been updated to use "true" strings instead of booleans.

Tested OK on localhost + mac OS 10 + Chrome on [a project with two Single Choice Tasks,](https://local.zooniverse.org:8080/?env=production&project=13082&workflow=16096) with each Task being set as 'required' via the PFE Project builder. 👍

Storybook stories seem to work with some quirks, but those quirks seem unrelated to this PR's changes - I'm investigating these as a separate matter. _(Basically, if you switch Required 'on' in Storybook before you make a choice in a Single Choice Task, the 'Done' button stays dead even after you make a choice. I dunno, it's a Storybook think, the Classifier itself works fine. 🤷‍♂️ )_

Code looks good and PR is generally a 👍 , though I have an idea for an improvement.

### Dev Notes

@mcbouslog while the solution works, I think it may be overkill to change 24 files to support "true" strings.

My thoughts:

- I think the original `.required = types.maybe(types.boolean)` is correct. The monorepo is pretty opinionated in how it _demands_ the data _should_ be shaped, and I'm on its side in this case.
- `.required` semantically makes sense to be (boolean) true _or_ false. Allowing `.required` to be true, "true", false, "", "f", "t", "T", "deep fried potatoes", or etc can cause unnecessary confusion for components, stories, etc that might use it.
- The actual problem, I think, lies in the PFE project builder, not the monorepo data model.

As such, I think an improved solution would be to use a snapshot preprocessor to "translate" any non-boolean values from the API/project builder, to a proper boolean for the MST store. You'd only need to change `Task.js` instead of every other file that touches on `.required`, I believe.

```
const Task = types.model('Task', {
  annotation: types.safeReference(Annotation),
  taskKey: types.identifier,
  required: types.maybe(types.boolean),  // You keep on being a boolean, .required.
  type: types.literal('default')
})
  // Add this down here
  // --------
  .preProcessSnapshot(snapshot => {
    const newSnapshot = Object.assign({}, snapshot)
    newSnapshot.required = (snapshot.required === true || snapshot.required === 'true')
    return newSnapshot
  })
  // --------
```

My feelings on using the snapshot preprocessor vs keeping your PR the way it is, on the patented How-Strong-My-Opinion-Is-O-Meter, from a scale of 1 (chill man, we're, like, all children of an ever-loving Universe here) to 10 (I will stand outside your house at 2am with a loudspeaker arguing that Michael Keaton is the best goddamn Batman until you capitulate): **7** or maybe **6** once I've had dinner.  Not sure why I'm craving french fries.

### Status

This PR looks good as is, though I think the Snapshot preprocessor is a better solution.

srallen:
I agree that the snapshot processor and boolean type would be preferable in our code, however, in discussing with @mcbouslog I leaned more toward getting this to work with the resource as is from Panoptes. Initially this issue came up when working on the branching single choice task work and initially was part of that PR, but I advised @mcbouslog to break it apart into two PRs for review. I'd say let's get this merged as is since #1758 depends on this, then open an issue for the refactor enhancement that is recommended? @shaunanoordin would that be a good task to pair with @ErikOstlund on?

srallen:
> The actual problem, I think, lies in the PFE project builder, not the monorepo data model.

Yep, it's how the HTML checkbox input works. Values are strings. It just so happens that with this state what we're trying to represent is actually a boolean state. I agree it's a bit wonky and I think I've been caught out on it before. I don't think we can really update the lab without negative side effects on current projects, so I would be supportive of a longterm solution and refactor to use a pre-processor.

mcbouslog:
Gadzooks, `preProcessSnapshot` indeed looks preferable long-term. I wasn't familiar with, but realize now it's very straightforward - my bad as I think I shied away from when @srallen mentioned when we discussed previously, then updating `required` to work with strings took on a life of its own.

If @shaunanoordin and @ErikOstlund can refactor to use `preProcessSnapshot` there might be a few tests that build single/multiple tasks that could be replaced with using the single/multiple task factories, though I'm not 100% sure. I am 100% sure Michael Keaton is the best Batman.

shaunanoordin:
I'm down with this plan. 👍 Let's get this merged first, then @ErikOstlund and me can run a refactor after the dust from #1758 has settled.

Batman gives a thumbs up. Yes, I know this is from Batman Forever, starring Val Kilmer, but I spent more time than necessary looking for a Michael Keaton Batman GIF to no avail.

PR #1770, opened 2020-08-26
Classifier Dropdown Task pt3 (v2) - Cascading Dropdowns

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Associated Project: Engaging Crowds / general PFE migration
Follows: #1750

This is part of a series of PRs that introduces the Dropdown Task to the monorepo Classifier

- Part 1 is described in #1742 
- Part 2 is described in #1750
- Part 3 is an experimental expansion of work that attempts to add the "cascading dropdowns" feature, which was previously discarded in Pt 1 and Pt 2 in favour of the "simple dropdowns only" goal.

The goal of this new experimental work is to bring the monorepo `dropdown` task closer to feature parity with PFE `dropdown` tasks

### Dev Notes

Let's be clear: _nobody asked for this right now and it's not on any project road map._ But I suddenly had an idea (also, week-long insomnia) and needed to work with it before it escaped my brain. And hey, a lot of it is working!

At the moment:

- Good: we can have multiple cascading `<selects>` per Dropdown task.
- Good: the options in a child Select will change depending of the value of the parent Select.
- Bad: if you change the value of a _parent Select,_ the child Selects that depend on that parent _won't reset to empty._ This seems to be a React thing, nyeh.

It's basically this:

```
<DropdownTask>  // Value from store is: { country: 'uk', city: 'london' }
  <SelectSubComponent title="Country" value="UK" options={[ 'UK', 'USA' ]} />
  <SelectSubComponent title="City" value="London" options={[ 'London', 'Oxford' ]} />
</DropdownTask>
```

If I change the Country to USA, the "City" component should, 1. clear/empty the value, and 2. change the options to ['Chicago', 'Minneapolis']. Unfortunately, that (1) isn't happening. The store is updating properly, the options are changing correctly, but the selected value ain't nixed.

SUDDEN IDEA: wait, maybe that's because I'm letting the `DdSelect` subcomponents manage their own states. Maybe I need to make it so only the Dropdown Task component uses state, and pass an "onChange" function to the DdSelect subcomponents instead.

Also: best look at the updated README.md for all my notes on how cascading dropdowns work, and some quirks we need to factor in. Also also, "free text" answers for cascading dropdowns can go straight to UI/UX hell.

### Status

WIP - purely experimental at this stage and possessing no priority. I'll come back to this if I have the time & energy, or am unable to sleep again.

Note: Pt3 was originally supposed to be design updates based on Becky's feedback, but that's now going to be Pt 4.

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/DropdownTask/README.md
+++ b/packages/lib-classifier/src/plugins/tasks/DropdownTask/README.md
@@ -118,3 +118,90 @@ Dropdown annotation (classification) data structure, example:
     }
  ]
 ```
+
+## Data Models (Cascading)
+
+Addendum - Dropdown task data structure (cascading), example:
+
+```
+"T0": {
+    "help": "",
+    "type": "dropdown",
+    "selects": [
+        {
+            "id": "e6648d9524038",
+            "title": "Country",
+            "options": {
+                "*": [
+                    {
+                        "label": "UK",
+                        "value": "8d2b49f2ab18a"
+                    },
+                    {
+                        "label": "USA",
+                        "value": "a893ad87b8b8c"
+                    },
+                    {
+                        "label": "Japan",
+                        "value": "cc8773d27f0aa"
+                    }
+                ]
+            },
+            "required": true,
+            "allowCreate": false
+        },
+        {
+            "id": "d6697219c5d3a",
+            "title": "City",
+            "options": {
+                "8d2b49f2ab18a": [
+                    {
+                        "label": "London",
+                        "value": "80623599d84d1"
+                    },
+                    {
+                        "label": "Oxford",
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Hi Shaun. Thanks for taking a look at this but I wouldn't spend a lot of time on it now. 

Cascading dropdowns exponentially increase the total number of available options, so the performance problem we'd like to solve here is deferring requests for options, rather than downloading the whole set up front.

srallen:
And again same request as in #1742. Ok to close and document?

shaunanoordin:
👌 Follow up from #1742 and #1750 closing.

This PR will now be closed to clear the PR queue; dev notes for resurrecting classic `dropdown` support will be compiled elsewhere.

Issue #1132, opened 2019-09-18
[RFC] How to handle workflow selection?


rogerhutchings:
In #228, I wrote an ADR on how to handle workflow routing, but a lot of the conversation revolved around how we choose a workflow for a given user; this issue is to continue that conversation.

It should lead to an ADR on how to handle workflow selection.

srallen:
This impacts home page design because with current project builder/PFE configuration it is possible to have a scenario where a project has multiple active workflows, does not have a default, and does not allow a volunteer to choose which workflow to work on. This is effectively the project setting its preference to the last fallback option of random selection for the initial visit.

beckyrother:
Why and how does this happen?

srallen:
@beckyrother, We have a list of project now where we can investigate why they might be using this configuration:

- Radio Meteor Zoo
- Count flowers for bees
- Parasite Safari
- Parochial Archive Project in Rome
- Galaxy Zoo Mobile - Might be a workaround for mobile use
- Anti-Slavery Manuscripts - CFE, so workflow selection does not apply

I think this list _might_ have been greater in the past because when the new project builder launched, only two kinds of workflow selection was supported: Allow volunteer to choose or random selection.

With the addition of several new features, we had to start keeping track of which workflow a user worked on last, which the project wants them to use, setting a default, etc. This greatly has complicated workflow selection, however, has meant over time that very few use random selection at all anymore.

It should be noted, that forcing volunteers to get a random workflow was a feature requested really early on in the development in PFE: https://github.com/zooniverse/Panoptes-Front-End/issues/189

> Question 2: Can we have a single button on the project landing page that goes to 'classify' without specifying workflow (and so get a random workflow)?

> Right now there's a button for each workflow on the project landing page. But I don't want users to have the option of choosing different workflows.

This request didn't really explain _why_ the project wanted this, but there may be a valid research reason why this might be wanted, although, with the current rarity, we want decide against supporting it. We need more information about potential use of random selection.

beckyrother:
Thanks @srallen. I'm receptive to the idea of random selection IF there's a good reason for it. 

In general randomization isn't a great user pattern as it makes it difficult for the user to get back to a specific workflow. Additionally, if someone really enjoyed one task on a project and wanted to see what else they could help with, they wouldn't be able to see other available workflows or dive deeper into the project in the way they might if they were able to see and participate in multiple workflows.

srallen:
[...]

PR #1793, opened 2020-09-08
Classifier: load specific workflows and subject sets

### LINKED ISSUE #1792: Classifier: enable subject selection from a specific workflow
**Package**
lib-classifier

When a volunteer chooses a specific workflow, they should be shown that workflow in the classifier.

When a workflow ID is passed into the classifier, load that specific workflow. Override `getDefaultWorkflowId` here to specify a workflow.
https://github.com/zooniverse/front-end-monorepo/blob/35512d4e620dcd23a88dbf356429b2931c1a0f46/packages/lib-classifier/src/store/WorkflowStore.js#L74-L80

eatyourgreens:
Package:
lib-classifier

Closes #1792 

Explicitly call the `workflows.selectWorkflow(workflowID)` action when the classifier mounts. Pass in any `workflowID` prop that was passed to the classifier, otherwise fallback to default project workflow selection. Remove the `selectWorkflow()` actions that automatically run when the project or UPP loads.

Add an optional `subjectSetID` parameter, which will restrict subject selection to that subject set: `workflows.selectWorkflow(workflowID, subjectSetID)`.

Workflow and subject set IDs are validated against a project's linked active workflows and a workflow's linked subject sets before being loaded.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/dev/components/App/App.js
+++ b/packages/lib-classifier/dev/components/App/App.js
@@ -64,6 +64,19 @@ class App extends React.Component {
     }
   }
 
+  workflowId () {
+    if (window.location && window.location.search) {
+      const { workflow } = queryString.parse(window.location.search) // Search the query string for the 'project='
+      if (workflow) {
+        return workflow
+      }
+
+      return undefined
+    }
+
+    return undefined
+  }
+
   login () {
     oauth.signIn('http://localhost:8080/')
   }
@@ -109,6 +122,7 @@ class App extends React.Component {
               onAddToCollection={(subjectId) => console.log(subjectId)}
               onCompleteClassification={(classification, subject) => console.log('onComplete', classification, subject)}
               project={this.state.project}
+              workflowID={this.workflowId()}
             />
           </Box>
         </Box>
--- a/packages/lib-classifier/dev/components/App/App.js
+++ b/packages/lib-classifier/dev/components/App/App.js
@@ -109,6 +109,7 @@ class App extends React.Component {
               onAddToCollection={(subjectId) => console.log(subjectId)}
               onCompleteClassification={(classification, subject) => console.log('onComplete', classification, subject)}
               project={this.state.project}
+              subjectSetID={this.props.subjectSetID}
               workflowID={this.props.workflowID}
             />
           </Box>
--- a/packages/lib-classifier/dev/components/App/App.js
+++ b/packages/lib-classifier/dev/components/App/App.js
@@ -64,19 +64,6 @@ class App extends React.Component {
     }
   }
 
-  workflowId () {
-    if (window.location && window.location.search) {
-      const { workflow } = queryString.parse(window.location.search) // Search the query string for the 'project='
-      if (workflow) {
-        return workflow
-      }
-
-      return undefined
-    }
-
-    return undefined
-  }
-
   login () {
     oauth.signIn('http://localhost:8080/')
   }
@@ -122,7 +109,7 @@ class App extends React.Component {
               onAddToCollection={(subjectId) => console.log(subjectId)}
               onCompleteClassification={(classification, subject) => console.log('onComplete', classification, subject)}
               project={this.state.project}
-              workflowID={this.workflowId()}
+              workflowID={this.props.workflowID}
             />
           </Box>
         </Box>
--- a/packages/lib-classifier/src/components/Classifier/Classifier.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.js
@@ -60,12 +60,12 @@ export default class Classifier extends React.Component {
   }
 
   componentDidMount () {
-    const { onAddToCollection, onCompleteClassification, onToggleFavourite, project, workflowID } = this.props
+    const { onAddToCollection, onCompleteClassification, onToggleFavourite, project, subjectSetID, workflowID } = this.props
     this.setProject(project)
     this.classifierStore.setOnAddToCollection(onAddToCollection)
     this.classifierStore.classifications.setOnComplete(onCompleteClassification)
     this.classifierStore.setOnToggleFavourite(onToggleFavourite)
-    this.classifierStore.workflows.selectWorkflow(workflowID)
+    this.classifierStore.workflows.selectWorkflow(workflowID, subjectSetID)
   }
 
   componentDidUpdate (prevProps) {
--- a/packages/lib-classifier/src/components/Classifier/Classifier.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.js
@@ -60,11 +60,12 @@ export default class Classifier extends React.Component {
   }
 
   componentDidMount () {
-    const { onAddToCollection, onCompleteClassification, onToggleFavourite, project } = this.props
+    const { onAddToCollection, onCompleteClassification, onToggleFavourite, project, workflowID } = this.props
     this.setProject(project)
     this.classifierStore.setOnAddToCollection(onAddToCollection)
     this.classifierStore.classifications.setOnComplete(onCompleteClassification)
     this.classifierStore.setOnToggleFavourite(onToggleFavourite)
+    this.classifierStore.workflows.selectWorkflow(workflowID)
   }
 
   componentDidUpdate (prevProps) {
--- a/packages/lib-classifier/src/store/WorkflowStore/WorkflowStore.js
+++ b/packages/lib-classifier/src/store/WorkflowStore/WorkflowStore.js
@@ -22,8 +22,7 @@ const WorkflowStore = types
         const validProjectReference = isValidReference(() => getRoot(self).projects?.active)
         if (validProjectReference) {
           self.reset()
-          const queryParamId = getQueryParamId()
-          selectWorkflow(queryParamId)
+          selectWorkflow()
         }
       }, { name: 'Workflow Store Project Observer autorun' })
       addDisposer(self, projectDisposer)
@@ -39,19 +38,6 @@ const WorkflowStore = types
         }
       }, { name: 'Workflow Store UPP Observer autorun' })
       addDisposer(self, uppDisposer)
-    }
-
-    function getQueryParamId () {
-      if (window.location && window.location.search) {
-        const { workflow } = queryString.parse(window.location.search) // Search the query string for the 'project='
-        if (workflow) {
-          return workflow
-        }
-
-        return undefined
-      }
-
-      return undefined
     }
 
     function getDefaultWorkflowId () {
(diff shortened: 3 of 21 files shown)


eatyourgreens:
This works locally, which is promising. Tested in the standalone classifier and in the project app, using both server-side and client-side rendering of the classify page. However, the tests for `WorkflowStepStore` are all broken and I can't see why.

packages/lib-classifier/dev/components/App/App.js
```diff
@@ -109,6 +122,7 @@ class App extends React.Component {
               onAddToCollection={(subjectId) => console.log(subjectId)}
               onCompleteClassification={(classification, subject) => console.log('onComplete', classification, subject)}
               project={this.state.project}
+              workflowID={this.workflowId()}
```
srallen, line 125: This is a pretty minor non-blocking comment since this only affects our dev classifier setup, but I've only seen the pattern of calling a function in the prop definition of a component when you dynamically determining what to render and the return value is a component. I'm not sure if this is a good idea otherwise, since this will be called every time this mini app is re-rendered and short circuits the lifecycle methods and any performance benefits those have. I think it'd be safer to call this in `componentDidMount`? Regardless, this is a non-blocking question.

srallen (commented):
I'm assuming this is ready for review with the change from being a draft PR... is there a project to test this with in the project app using routes? This works with the dev classifier testing between loading different workflows with staging project 1754 (ASM testing) and using query param.

packages/lib-classifier/dev/components/App/App.js
eatyourgreens: Good catch. `workflowId` doesn't even need to be a class method. It could be run once, outside the component, and the workflow ID passed in as a prop to App.

eatyourgreens:
I've tested with the Classic vs. Enhanced workflows for Galaxy Zoo, though it's hard to tell exactly which one is loaded in the classifier. I've also got a staging project with a bunch of different workflows. https://master.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground

packages/lib-classifier/src/store/WorkflowStore/WorkflowStore.js
```diff
@@ -22,8 +22,6 @@ const WorkflowStore = types
         const validProjectReference = isValidReference(() => getRoot(self).projects?.active)
         if (validProjectReference) {
           self.reset()
```
eatyourgreens, line 24: Should this reset move into the `selectWorkflow` action, so that we always reset the store before setting a new workflow?

packages/lib-classifier/src/store/WorkflowStore/WorkflowStore.js
```diff
@@ -35,25 +33,11 @@ const WorkflowStore = types
         const validWorkflowReference = isValidReference(() => self.active)
         if (validUPPReference && !validWorkflowReference) {
           self.reset()
```
eatyourgreens, line 35: Same question about resetting the store here.

packages/lib-classifier/dev/components/App/App.js
srallen: Yeah, I think we had talked about ages ago about the potential of there being a functional workflow selection library that could be used independently by any app and this is one of the strategy types.

eatyourgreens: Yeah, that could happen outside the classifier, or the `selectWorkflow` action could invoke a strategy in order to choose a workflow, with some sort of default fallback. Outside the classifier could mean that it's the project's responsibility to choose a selection strategy, which could be useful for Gravity Spy.

If we made `workflowID` a required prop for the classifier, that might allow some simplification of the stores by removing some of the autorun code, and the default fallback. Do we want consuming apps to be responsible for selecting a workflow and passing it in? That's probably out of the scope of this PR.

I've moved the workflow parameter function out of the classifier component and into the dev app. Should I do the same with the code that gets the project from the URL? At the moment, you can do something like this with the production classifier but it seems like params should only be for development.
https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/classify?project=908&workflow=3370&env=staging

srallen: Agreed that it's out of scope of this PR. Would you add notes to the RFC about workflow selection? I don't want this to get lost in this PR. Thanks! https://github.com/zooniverse/front-end-monorepo/issues/1132

We'll have to address workflow selection soon, so capturing an idea of what we want to do would be good.

eatyourgreens: Will do. I think it's a variation on the idea here: https://github.com/zooniverse/front-end-monorepo/issues/1132#issuecomment-548516888

eatyourgreens:
This seems to work for anonymous users but I'm seeing an error from the state tree if I try to load the classifier while logged in.
```
Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree. (Object type: 'SubjectSetResourceStore', Path upon death: '/workflows/resources/15652/subjectSets', Subpath: 'headers', Action: '/workflows/resources/15652/subjectSets.e()'). Either detach nodes first, or don't use objects after removing / replacing them in the tree.
```
http://localhost.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/15652/subject-set/85773?env=production

I'm also finding that the project app dies with an MST error if I run
```
yarn dev
```
with the dev tools open. It runs fine with the dev tools closed (which isn't very useful) or if I run a production build with dev tools:
```
yarn build
yarn start
```

eatyourgreens:
I'm testing on a staging project withe the project and UPP observers removed:
http://localhost.zooniverse.org:3000/projects/eatyourgreens/-project-testing-ground/classify/workflow/3371

I see workflow selection happen while the classifier is still loading the UPP, but I don't see that make any difference to the subject selection request, so it seems like the changes here load the correct workflow and subject queue for a volunteer, even if the UPP hasn't loaded in the classifier yet.

eatyourgreens:
One potential enhancement of this (but outside the scope of this PR): fetch the workflow and subject set in `getServerSideProps` so that the workflow and subject set titles can be added to the page title (useful for bookmarking and for SEO.) The full workflow and subject set objects could then be passed into the classifier as props, similar to how projects are handled at the moment.

packages/lib-classifier/src/components/Classifier/Classifier.js
```diff
@@ -60,11 +60,12 @@ export default class Classifier extends React.Component {
   }
 
   componentDidMount () {
-    const { onAddToCollection, onCompleteClassification, onToggleFavourite, project } = this.props
+    const { onAddToCollection, onCompleteClassification, onToggleFavourite, project, subjectSetID, workflowID } = this.props
     this.setProject(project)
     this.classifierStore.setOnAddToCollection(onAddToCollection)
     this.classifierStore.classifications.setOnComplete(onCompleteClassification)
     this.classifierStore.setOnToggleFavourite(onToggleFavourite)
+    this.classifierStore.workflows.selectWorkflow(workflowID, subjectSetID)
```
srallen, line 68: I think I would prefer on continuing to rely the reactive paradigm we have setup using observers in the stores for asynchronous request or setup of resources in the store. However, I think that will take some additional refactoring that is out of scope here.

srallen (approved):
> EDIT: removing the WorkflowStores's Project and UPP observers, which reset the store on change, seems to fix this. I'm not sure what those observers are intended to do but nothing seems broken without them.

1) To be able to use the classifier as stand-alone from the project app, the workflow should request a new workflow if the project changes
2) To setup to be able to validate a workflow that is requested against the workflow id stored in the UPP. Therefore, if the UPP changes, request a new workflow. This part isn't finished yet.

Removing these observers limits what the classifier lib can do independently now. However, since neither of these are required by the Engaging Crowds feature, I would suggest we leave in the observers, but with todo comments on future enhancement and don't have them reactively do anything yet:

```
    function createProjectObserver() {
      const projectDisposer = autorun(() => {
        const project = self.project
        const workflow = tryReference(() => self.active)
        if (project && !workflow) {
          // TODO request for the workflow if project has changed and there isn't one already
          // self.reset()
          // selectWorkflow(workflowID, workflow.subjectSetID)
        }
      }, { name: 'Workflow Store Project Observer autorun' })
      addDisposer(self, projectDisposer)
    }

    function createUPPObserver() {
      const uppDisposer = autorun(() => {
        const upp = tryReference(() => getRoot(self).userProjectPreferences?.active)
        const workflow = tryReference(() => self.active)
        if (upp && !workflow) {
          // TODO validate and get workflow id from UPP to select
          // self.reset()
          // selectWorkflow()
        }
      }, { name: 'Workflow Store UPP Observer autorun' })
      addDisposer(self, uppDisposer)
    }
```

>  One potential enhancement of this (but outside the scope of this PR): fetch the workflow and subject set in getServerSideProps so that the workflow and subject set titles can be added to the page title (useful for bookmarking and for SEO.) The full workflow and subject set objects could then be passed into the classifier as props, similar to how projects are handled at the moment.

Agreed that this is out of scope, but again, I think I'd prefer we lean toward taking a snapshot (from prop or somewhere else?), which could be used in the constructor of the Classifier where the store is initialized. The snapshot could be defined to have full resources or to just have ids that the classifier could then use to make requests as needed. I don't want to have to depend a whole lot on logic happening in either a `componentDidMount` or hooks to setup the classifier.

Anyway, I've tested this using the production project app starting with this URL:
http://localhost.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/15652/subject-set/85773?env=production

And confirmed that the requested workflow matched 15652, subject set matched 85773, and the queued endpoint correctly added those as query params.

Then I navigated to the home page and selected a new workflow and subject set:
http://localhost.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/16106/subject-set/85771

And I confirmed the requested workflow id is 16106, subject set is 85771, and the queued endpoint correctly added those query params.

Then back to the homepage and selected a non-grouped workflow, id 15651 aka "Drawing Workflow", and it correctly requested it and used it for the queued endpoint, but the route in the URL did not have the id. This is out of scope here, but I would expect the URL to have this workflow id as well.

eatyourgreens:
I think that might be the default workflow. Those are routed to `/classify` in the workflow selector code.

packages/lib-classifier/src/components/Classifier/Classifier.js
eatyourgreens: I couldn't come up with a good way to pass the classifier props down into the workflow store, except by calling an action directly. If this block could be replaced with `store.create(snapshot)`, that would be great.

eatyourgreens:
I'm going to open a PR to restore the observers. One scenario I tested here in the project app was to log in, check that the subject queue refreshed with my personal subject queue, then log out and check that it refreshed using the random subject selector. I can see a case for observing user changes in order to handle that inside the standalone classifier.

PR #1796, opened 2020-09-11
[VSLCV] Configurable bar chart y-axis domain

### LINKED ISSUE #1440: [VSLCV] Bar chart design enhancements
**Package** lib-classifier

After meeting with the researchers, we have to balance design and accessibility needs with the needs of the project, so there are a few discussion points we need to review with regard to the bar chart and how it is displayed in the composite VSLCV:

- The specific values of the bars in their case shouldn't distract volunteers from making variable star classifications. They're worried that some volunteers may get too focus on specific numerical values and that might lead them to make inaccurate classifications. To this end, they want the axis values to be hidden. What they want volunteers to do is to just compare the bar height size of the red light and blue light spectra to each other for period and amplitude.
- However, generally speaking, we need labels for accessibility. They've recommended just having labels like 'tall' and 'short' be read out. I'm not sure if that's viable because how do you determine what is 'tall' and 'short' as it's inherently an arbitrary distinction.
  - We could choose to allow project builders to submit aria labels in the JSON, but I'd rather not go this route as it may open up the bar chart to more complexity and potentially poor labeling.
  - I showed the tooltips from #1375, and that was an acceptable middle ground for displaying specific values or labels. If the axes are hidden, though, are the tooltips discoverable?

So the question is: How can we make the display of the axes and specific individual bar values configurable and also maintain accessibility? @eatyourgreens do you have any suggestions for this?

Additional TODO:
- [ ] How the labels are read out by the bar chart currently still need adjusting [comment](https://github.com/zooniverse/front-end-monorepo/pull/1375#pullrequestreview-345681695).

srallen:

Package: lib-classifier

Closes #1440

Describe your changes:
This adds a prop and sensible defaults for the bar chart y-axis domain. The researchers asked for a domain of `[0, 10]` to help make it more obvious when the period bar chart changes:

bar-chart-domain

There are some discussion and suggestions in #1440 we're not going to do. I'll open a separate issue about adding animations, if we have time.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


wgranger (approved):
Looks great on all major browsers. Documentation is up to date. Tests are included and have been refactored for these changes.

beckyrother (approved):
Do the period and/or amplitude ever reach 10? Wondering if it's just this example but if they're all closer to 1-2 it's a bit harder to differentiate. Approving because this is a non-blocking question, just a comment

srallen:
> Do the period and/or amplitude ever reach 10? Wondering if it's just this example but if they're all closer to 1-2 it's a bit harder to differentiate. Approving because this is a non-blocking question, just a comment

I don't know. This is what the researchers asked for and I've replied back to them asking if there's any potential subject data where this happens. If not, then we can adjust it; I'm suggesting a max of 5, rather than 10. We  merge this as is now and then if they agree to the change, then a small follow up PR can change it.

adamamiller:
> > Do the period and/or amplitude ever reach 10? Wondering if it's just this example but if they're all closer to 1-2 it's a bit harder to differentiate. Approving because this is a non-blocking question, just a comment
> 
> I don't know. This is what the researchers asked for and I've replied back to them asking if there's any potential subject data where this happens. If not, then we can adjust it; I'm suggesting a max of 5, rather than 10. We merge this as is now and then if they agree to the change, then a small follow up PR can change it.

Amplitude will (essentially) never reach 10. Periods can be much much larger than 10 (some pulsating stars have periods as large 1000 d). For this particular project it would be most useful to show log_10 (period) with a range from -2 to 3. There may be some sources with period < 0.01 d or > 1000 d, but at that stage the difference does not really matter. Similarly, log_10 (amplitude) could also make sense. If a linear range is needed, I would show 0 to 3 (there are some sources with amplitude > 3, but they are very small in number).

srallen:
Thanks @adamamiller. The quickest thing right now would be to adjust the linear scale's domain. As a future enhancement, we could look into supporting other scale types through configuration.

Issue #1798, opened 2020-09-15
React components: Customise the appearance of the Modal component


eatyourgreens:
**Package**
lib-react-components

**Is your feature request related to a problem? Please describe.**
The Modal component follows our [standard modal layout and colours](https://projects.invisionapp.com/dsm/zooniverse/primary-brand/asset/components/5ea995ce07156c97c3cd7985) by default. Modal title bar colours can be changed by passing in `titleColor` and `headingBackground` props but, other than that, modals can't be customised.

Drawing subtasks, transcription tools and the subject set picker (for grouped workflows) all use spacing and layout that is different from the standard Modal spacing and layout. In addition, other projects (such as mapping visualisation) should be able to import the Modal component without having to use the default colours and layout.

**Describe the solution you'd like**
Projects should be able to set their own default Modal styles (maybe by supplying their own theme?)

Individual Modals, within a project, should be able to override the default styles where necessary.

PR #1799, opened 2020-09-15
Add subtask required UI

### LINKED ISSUE #1623: [RFC] Sub-tasks required task UX
**Package** lib-classifier

A common issue on PFE is that some volunteers get confused when they accidentally do not fill out the sub-task when it is required. See https://github.com/zooniverse/Panoptes-Front-End/issues/5571 and https://github.com/zooniverse/Panoptes-Front-End/issues/5463

The current behavior as described in #1546 is the the sub-tasks, whether they're text or question, save the annotation _when the annotations is made_, not when "Save" is clicked. Changing this would be a big refactor on the tasks in general. 

We need to have differing behavior and UX if the sub-task is required. A few ideas:

Option 1: The sub-task has UI, focus, and validation if one of the subtasks is required and isn't changed from the default annotation. If it isn't changed, then do not allow the subtask area to close and focus the task that isn't finished yet.

Option 2: Have an ASM-like UX where the mark is deleted if the subtask isn't filled out and one of the sub-tasks is required. (but this doesn't communicate to the user what happened? how do we do that?)

PFE originally had a UI that notified users of the requirement, but ~it was removed~ it was removed for main tasks and doesn't fix the issue of marks having unfinished required sub-tasks and you don't know which one is not finished. See: https://github.com/zooniverse/Panoptes-Front-End/issues/1666

ASM deletes the mark if the subtask isn't filled out, but this isn't a generalizable solution since some projects may not require the subtask.

mcbouslog:
Package: lib-classifier

Closes #1623.

Describe your changes:
- adds "This step is required." under required subtasks
- if close or click away from subtask with incomplete required subtasks, confirm intention
- if intend to answer unanswered required subtasks, return to subtask
- ^ highlight unanswered required subtasks
- if intend to close subtask without answering required subtasks, close confirm, close subtask and delete mark

Using local.zooniverse.org:8080/?project=1876&workflow=3341 for local testing

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


mcbouslog:
Opening Draft PR for initial feedback, if general functionality is what's intended. If good to continue, this'll need design input, translations, and tests.

mcbouslog:
@beckyrother - this still really rough design, but functionally I think working close to how we'd like (see #1623), so would like to get your input? I'll post some screenshots. The testing workflow I'm including includes a first subtask "Breakfast?" that is required, and a second subtask "Coffee?" that is not required. Let me know if you'd like to see variations on the subtasks.

1. Initially:
Screen Shot 2020-09-15 at 11 11 38 AM

2. Modal if click Save, click close button, or click outside modal _with unfinished required subtasks_:
Screen Shot 2020-09-15 at 11 14 00 AM

3. If click "No" to go back to subtask modal and complete any unfinished required subtasks, highlights unfinished required subtasks:
Screen Shot 2020-09-15 at 11 14 19 AM

mcbouslog:
Also @beckyrother ...

The required subtask text for PFE is "This step is required.", which makes sense for PFE, but within the new frontend Steps are their own thing, maybe we should use "task" or something else?

Also, maybe the text should be moved closer to related subtask question or a visual differentiation should be made between subtasks?

beckyrother:
This is looking great! Couple of things:
1. I think I'd move the 'this is required' text to just below the question so it's clear to which step it's referring 

2. "Modal if click Save, click close button, or click outside modal with unfinished required subtasks"  :  
- To be totally clear, the question can be  "Are you sure you want to close? Marks will be deleted." buttons: "Keep working" / "Close without saving" with  "keep working" as the [secondary button](https://projects.invisionapp.com/dsm/zooniverse/primary-brand/asset/components/5eda8d65f30fad2f9adbd1f1) and "close without saving" as the primary button (styled the way it is now)
- This happens when they click Save as well? In that case the save button should read "Save and Close"

3. Looks good! I like the red outline.

beckyrother:
Missed the second comment!

- Good call, I think we say "task" elsewhere.
- I feel like @shaunanoordin recently did something with an arrow glyph to indicate a subtask? Am I making that up?

mcbouslog:
@beckyrother everything noted sounds great to me! I'm updating accordingly, cleaning other code specific pieces up, then will remove the draft designation and (re-)request reviews.

mcbouslog:
I didn't realize how complicated inserting "This task is required." would be into the subtask component, specifically under the question/instruction, which I realize is why it's the way it is in PFE. It's possible! But I'm going to open a new issue to discuss approaches. 

For now, this PR leaves the subtasks without any required notation (similar to normal tasks, only queue is disabled Next/Done) until the user attempts to Save and Close modal with incomplete required subtasks, then the incomplete required subtasks are emphasized as follows:

Screen Shot 2020-09-23 at 11 20 51 PM
Screen Shot 2020-09-23 at 11 22 02 PM

mcbouslog:
Confirm modal screenshots:

Screen Shot 2020-09-23 at 11 20 26 PM
Screen Shot 2020-09-23 at 11 21 36 PM

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js
```diff
@@ -1,19 +1,64 @@
...
+        margin={{ bottom: 'medium' }}
+      >
+        {counterpart('Task.confirm')}
+      </Paragraph>
+      <Box
+        direction='row'
+        gap='small'
+        justify='center'
+      >
+        <StyledPrimaryButton
```
srallen, line 43: This can be a standard Grommet Button (defaults to the brand teal already w/ no background color):

Screen Shot 2020-09-24 at 2 27 01 PM

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js
```diff
@@ -1,19 +1,64 @@
...
+          label={counterpart('Task.keepWorking')}
+          color='teal'
```
srallen, line 45: This prop can be removed when switched over to the Grommet Button

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js
```diff
@@ -24,75 +69,107 @@ function SubTaskPopup(props) {
   } = activeMark
   if (!subTaskVisibility) return null
 
+  const [confirm, setConfirm] = React.useState('pending')
```
srallen, line 72: I'm not 100% on the naming here... maybe `confirmationState`? This is a non-blocker for me though.

srallen (changes requested):
I'll leave the final say to @beckyrother but I think the background colors are a bit odd in the dark theme. We have modals use #5C5C5C and the task area uses #333333. The single choice and multiple choice task input buttons are designed to also use #333333. I think either the subtask modal background needs to be #333333 or we refactor the buttons to use a transparent background so that whichever the parent container uses will show through.

I've also encountered a strange behavior. Sometimes the confirmation modal triggers a flicker in the browser:

wtf-flicker

What's happening in the console is that the DOM elements are going back and forth between hiding the overflow and not on the document body. My guess is this is a Grommet bug and likely javascript somehow caught in a loop.

Another side note, we still have a style problem with the close button in dark theme in the modal:

Screen Shot 2020-09-24 at 2 23 59 PM

Fixing this is within the scope of https://github.com/zooniverse/front-end-monorepo/issues/1798 I think.

beckyrother:
> I'll leave the final say to @beckyrother but I think the background colors are a bit odd in the dark theme.

Agreed, the primary button on the dark theme should still be #00979d filled in. The light theme looks perfect!

beckyrother (changes requested):
Ack I'm so bad at github, sorry!! Here's my review (the previous comment, in Review form)

The primary button on the dark theme should still be #00979d filled in. The light theme looks perfect!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js
mcbouslog: I like `confirmationState` better 👍

mcbouslog:
I think this is ready for another review.

I changed the dark theme task input background to `transparent`, but happy to change back and open as different issue/PR if out of scope.

I need to isolate and document the flicker issue, that's an open item.

Here are some screenshots:
Screen Shot 2020-10-13 at 4 04 45 PM

Screen Shot 2020-10-13 at 4 11 01 PM

Screen Shot 2020-10-13 at 4 05 43 PM

Screen Shot 2020-10-13 at 4 05 31 PM

beckyrother (approved):
Looks great! Thanks Mark!

srallen (approved):
I just have some non-blocking suggestions:

- Update the transcribed line's story to use a task with a required sub-task, so update lines 41-47 to:
```js
          details: [
            {
              instruction: 'Transcribe the text',
              required: 'true',
              taskKey: 'T0.0',
              type: 'text'
            },
          ],
```

- To move `ConfirmModal` component into its own file in the components folder so it can have its own spec file with smoke test and for organization.

- Failed prop type in the `TextTaskWithSuggestions` component for the `required` prop. It needs to be updated to: `PropTypes.oneOfType([PropTypes.string, PropTypes.bool])`

I'm approving this for now, however this is not fully working with https://localhost:8080/?project=1764 (the ASM test project on stagine). The sub-task is required, but the logic in line 24 of the Task model evaluates as true for the text task even if the sub-task is required: 

https://github.com/zooniverse/front-end-monorepo/blob/3535805f98eedcdeab99d835c869486ba74f1a55/packages/lib-classifier/src/plugins/tasks/models/Task.js#L24

The reason is because the TextAnnotation model doesn't have a defined `isComplete` condition and falls back to the default Annotation model's:

https://github.com/zooniverse/front-end-monorepo/blob/3535805f98eedcdeab99d835c869486ba74f1a55/packages/lib-classifier/src/plugins/tasks/models/Annotation.js#L22-L24

I think adding this is out of scope of this PR, but should be an immediate follow up.

srallen (approved):
LGTM. Thanks Mark!

Issue #1780, opened 2020-09-01
[VSLCV] Alpha testing


srallen:
**Package** lib-classifier

This is an issue to capture the initial alpha testing of the variable star light curve viewer. This can be tested here: https://frontend.preview.zooniverse.org/projects/adamamiller/stellar-sleuths/classify

If any issues are found that are not already documented as a separate issue (https://github.com/zooniverse/front-end-monorepo/milestone/36), please try to use the template below to document it.

### Bug Template

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

trouille:
Mac OS Mojave v10.14.6, with Chrome v84

Minor: the temp-luminosity plot axes are labeled, but then the x-axis label (temperature) is repeated below. 

Minor: when you enable zoom (on either plot), there's a funny movement of the screen when you move the cursor around. Here's a [ video ](https://drive.google.com/file/d/1sog07cwD-31A05QUEoGZpoBVT_lQgPbS/view?usp=sharing) showing the behavior since it's a bit hard to describe. 

Question: When you click the 'visibility' options, should just the color go away, or should the point+color go away?

trouille:
It'll be good to hear from the researchers if the way zoom is working in the plots is as they need/want/expect.

eatyourgreens:
Minor display bug: the tooltip isn't positioned properly for the bar chart in Firefox. I think there's already an issue open for this. I triggered this after scrolling the viewport down a bit, which seems to throw off the position calculation.

![Screenshot in Firefox, showing that the bar chart tooltip is displaced below and to the right of the focussed bar.](https://user-images.githubusercontent.com/59547/91970567-c2665480-ed0f-11ea-85d9-274b5a90009e.png)

srallen:
Yes, we have an open issue for the positioning bug in Firefox. Please check the existing issues first: https://github.com/zooniverse/front-end-monorepo/milestone/36. I'd like this to be new feedback. Thanks!

srallen:
> Minor: the temp-luminosity plot axes are labeled, but then the x-axis label (temperature) is repeated below.

The fig caption is now "HR Diagram"

> When you click the 'visibility' options, should just the color go away, or should the point+color go away?

Resolved by making the label "highlight" instead of "visibility"

[...]

PR #1801, opened 2020-09-15
[VSLCV] Make glyph less visible when toggled off

### LINKED ISSUE #1778: [VSLCV] Improve styling for data series not focused when there's a lot of data
**Package** lib-classifier

Researcher request: 

> We will have some sources that have a LOT of data (>10 x what is in the current example). When turning off a filter, is it possible to make the symbols even lighter (and possibly also transparent)?

Perhaps we should actually consider making unfocused series invisible? This will be an underlying change that affects the `ScatterPlot`. I'll see if they can provide us with an example of what "a LOT of data" is.

srallen:

Package: lib-classifier

Closes #1778 

Describe your changes:
We had a request to make the data points less visible or possibly even transparent both from the researchers and from others during the alpha test (See https://github.com/zooniverse/front-end-monorepo/issues/1780#issuecomment-685167861).

We have two options, one I've removed the color and lightened the stroke:

visibility-still-show

The other is to make them fully transparent:

fully-transparent

I'm not convinced making them fully transparent is a good idea, but open to discussion. @beckyrother thoughts? I haven't been sent an example of a subject that has a lot of data points, but I guess imagine one that has much more than what we see here?

Once we decide on an option, I'll finish this PR and update tests.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


beckyrother (commented):
From a design standpoint, if I were to uncheck a box that said "visibility" I would assume it would turn off the visibility entirely, so I'd be inclined to go with the second option. I'm not a scientist, though, so would defer to the research team as to the usefulness of these options.

srallen:
@adamamiller could you let us know which is the preferred function here? Please see my initial comment for animated gif examples: https://github.com/zooniverse/front-end-monorepo/pull/1801#issue-487587665

adamamiller:
> @adamamiller could you let us know which is the preferred function here? Please see my initial comment for animated gif examples: [#1801 (comment)](https://github.com/zooniverse/front-end-monorepo/pull/1801#issue-487587665)

I should have an example with many more observations shortly (sorry for delay). I agree with @beckyrother that if it says visibility the points should disappear entirely. However, it looks a bit strange to have no points displayed at all when all options are unchecked... 

Perhaps it would make sense to change "visibility" to highlight? And then the observations don't need to disappear entirely? @ageller what do you think?

ageller:
I agree that we don’t want the data to disappear entirely when the box in unchecked.  (I like how the data become grayed out.)  Using the term “highlight” seems reasonable, or something like “colorize” or “emphasize”.

beckyrother:
I like Highlight! 

Wondering if the greyed out should be even lighter just to be sure this function is useful?

srallen:
Sounds good. Thanks for all the guidance everyone! I'll post some updates today for review.

srallen:
@beckyrother here is the updated control label using "highlight" and the gray stroke for the points is lighter:

adjusted-colors

beckyrother (approved):
Looks great @srallen !

wgranger (approved):
LGTM. This works on multiple browsers and is accessible. I can tell series are being toggled, and in doing so the color becomes transparent and the stroke/outline changes from black to a dim gray.

Note: I'm also noticing tests are failing locally in other components, but that is unrelated to this PR.

srallen:
> Note: I'm also noticing tests are failing locally in other components, but that is unrelated to this PR.

@wgranger What failures are you seeing? CI is passing and I don't see any locally either.

PR #1804, opened 2020-09-16
Classifier: specify subject set

### LINKED ISSUE #1791: Classifier: Enable subject selection from a specific subject set
**Package**
lib-classifier

Once a volunteer has selected a workflow and subject set to work from, we should show them only subjects from that subject set.

When a subject set ID is passed into the classifier as a prop, use that subject set ID during subject selection here:
https://github.com/zooniverse/front-end-monorepo/blob/35512d4e620dcd23a88dbf356429b2931c1a0f46/packages/lib-classifier/src/store/Workflow/Workflow.js#L23-L27

eatyourgreens:
Package:
lib-classifier

Closes #1791.

Add an optional `subjectSetID` argument to `workflowStore.selectWorkflow(workflowID, subjectSetID)`. Store any selected subject set ID on the active workflow. Return that value from `workflow.subjectSetId` during subject selection.

Converts `workflowStore.selectWorkflow()` to return a promise, so that we can request a workflow from the API, set an active subject set on that workflow and then set the active workflow.

NB.this makes changes to the workflow store so I've based it off #1793 **not** the main branch.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
You should be able to test this locally with Shaun's test project:
http://localhost.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020?env=production
or Annotate:
http://localhost.zooniverse.org:3000/projects/drrogg/annotate?env=production

eatyourgreens:
Testing with Shaun's project, I'm seeing the subject set ID being correctly set on the classifier.
http://localhost.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/15652/subject-set/85773?env=production

but the subjects request still has the default subject set ID for this workflow.
https://www.zooniverse.org/api/subjects/queued?workflow_id=15652&subject_set_id=85771&http_cache=true

it seems like `workflowStore.setActive('15652')` initiates a subjects request before `workflowStore.active.selectSubjectSet('85773')` is called.

srallen (changes requested):
I think in fitting with the concept of the classifier stores are structured, I would prefer we add a `SubjectSetStore` to model and store the resource when a request is made. This store would be composed with the base `ResourceStore`, be a child node of and observe the workflow store, and handle its own request as needed. This also sets us up to be able to load in an already existing snapshot into the store as well.

eatyourgreens:
We don't ever fetch the subject set, but I could add that in. What would the resource be used for, other than the ID in the subjects query?

eatyourgreens:
Tests are horribly broken on this branch, and I can't immediately see why. 🤔

srallen:
@eatyourgreens clarify if I'm wrong, but display names and potentially descriptions and set metadata are or will potentially be used in the selection tool? We also have an idea to have the project and workflow selection and consequently the subject set selection be available from the classifier page (see: https://github.com/zooniverse/front-end-monorepo/issues/642), so having the subject set resource display names, descriptions, etc available to the library would be helpful. We could potentially load up an already selected subject set via snapshot as well to avoid requesting again. Regardless, having the resource modeled I think will set us up for a lot of flexibility for enhancement in the future.

eatyourgreens:
I don't think those are in the designs at the moment, but worth considering as an enhancement. For testing on staging, right now I think we just need the subject set ID as a query parameter (before I go on leave.)

srallen:
Just to add: even if we don't actually request and store the resource in this iteration, I think having each resource have its own model with the id stored in a map of resources and the currently selected one as the active reference keeps the API consistent and predictable across the state tree. So if we don't request the resource, the resource map would just be the ids and active the selected id.

eatyourgreens:
Changing the `ResourceStore` model broke almost all of the classifier tests, so I've reverted that. I'm trying to figure out why one test is still failing, although I can run projects locally without error.

srallen:
I don't think `ResourceStore` needs any changes to do what I'm suggesting.

eatyourgreens:
No, I just nuked a commit from this morning that was breaking all the tests, where I added a `getResource` method to `ResourceStore`.

eatyourgreens:
Tests for 'Done and Talk', in `WorkflowStepStore` are still failing. I think maybe because I've made workflow selection asynchronous. 🤔

eatyourgreens:
We could pass the full subject set in as a prop, rather than just the ID. The project pages request it from Panoptes anyway.

srallen:
> We could pass the full subject set in as a prop, rather than just the ID. The project pages request it from Panoptes anyway.

Yeah, out of scope from this PR, but I had in mind that we would have some kind of mechanism to pass in snapshots and load those when available. Could be as prop or some kind of utility could look for a snapshot of store trees in local storage or indexedDB.

eatyourgreens:
Tests are passing now! 🎉 The `WorkflowStepSotre` tests had to be updated to use promises for workflow setup, now that `selectWorkflow` is asynchronous.

eatyourgreens:
`workflow.subjectSets.active` seems cleaner to me than `workflow.activeSubjectSet`. I'm not a big fan of the name either, since it's an ID, not a subject set.

packages/lib-classifier/src/store/WorkflowStore/WorkflowStore.js
```diff
@@ -55,8 +55,14 @@ const WorkflowStore = types
       return id
     }
 
-    function selectWorkflow (id = getDefaultWorkflowId()) {
+    function * selectWorkflow (id = getDefaultWorkflowId(), subjectSetID) {
       if (id) {
+        const workflow = yield self.fetchResource(id)
```
eatyourgreens, line 60: This will always fetch the workflow from the API, even if its already loaded.

eatyourgreens: Fixed by adding `getResource` to `ResourceStore`.

eatyourgreens:
I've added `getResource` back in, but more conservatively so that it's only being used by workflows. Yesterday, I added it to `setActive`, which broke the tests. The code seemed fine: I could run projects locally without problem, but running the store tests failed with an infinite loop somewhere.

eatyourgreens:
I've also moved the base `ResourceStore` model into its own folder, so that we can add a Readme.

eatyourgreens:
I've also confirmed that the following refactor of `setActive` breaks all the tests.
```js
setActive: flow(function * setActive (id = null) {
      const active = yield self.getResource(id)
      self.loadingState = asyncStates.success
      self.active = id
    })
```

I can still run the dev classifier without any problems, after that change. There's something very brittle in the mock stores somewhere. Maybe the tests don't explicitly await the outcome of `setActive`?

For the time being, I'm going to leave duplicated code for getting a stored resource in the `getResource` and `setActive` actions.

eatyourgreens:
I've added a subject set resource store, but API requests are now broken. With set 85773 selected in the URL, I'm seeing the API request use the default value of 85771 (for the test workflow.)
https://www.zooniverse.org/api/subjects/queued?workflow_id=15652&subject_set_id=85771&http_cache=true

I might leave the subject set store out of this PR, since it isn't really needed to begin testing for Engaging Crowds.

eatyourgreens:
With #1805 merged, you can now test this in the dev classifier:
https://localhost:8080/?env=production&project=12754&workflow=15652&subjectSet=85773

You can also test a full project at:
http://localhost.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/15652/subject-set/85773?env=production

eatyourgreens:
API behaviour is odd if you supply a subject set ID that isn't attached to the project. Here, 857 is an AnnoTate subject set but subjects still load from Shaun's project. I can't see anything in the `/subjects/queued` response that would allow us to validate the queue against the specified set.
https://localhost:8080/?env=production&project=12754&workflow=15652&subjectSet=857

srallen:
> API behaviour is odd if you supply a subject set ID that isn't attached to the project. Here, 857 is an AnnoTate subject set but subjects still load from Shaun's project. I can't see anything in the /subjects/queued response that would allow us to validate the queue against the specified set.
https://localhost:8080/?env=production&project=12754&workflow=15652&subjectSet=857

The workflow has serialized with it the subject sets that are linked to it, so we could validate against the workflow links before even attempting to make a requested against the queued endpoint.

srallen (approved):
The addition of the subject set model was reviewed and approved in #1805, so this review is for:

- the addition of the `subjectSetId` prop to the `Classifier` component
- the addition of the `getResource` action
- the change to requesting the workflow to be an async action

Automated tests are passing and manual tests with the dev classifier (https://localhost:8080/?env=production&project=12754&workflow=15652&subjectSet=85773) and project app (http://localhost.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/15652/subject-set/85773?env=production) show that the requests to the queued endpoint are correctly adding the query parameters we would expect for the workflow and subject set from the URL.

There was a comment about validation on the subjects and I agree we should add in validation, but I'm recommending we validate the subject set id before a request to the queued endpoint is even made. I don't consider this a blocker and think we can add an issue to add this validation enhancement in. I think this is ready for a merge into the final `classifier-specify-workflow` branch for final review and regression testing. Thanks!

eatyourgreens:
Thanks for the review! I'll add validation against the workflow links in #1793. I'm going to open an issue on Panoptes about the API call.

eatyourgreens:
https://github.com/zooniverse/panoptes/issues/3477 for subject selection when an invalid subject set ID is sent with the API request.

PR #1805, opened 2020-09-17
Add a subject set store to workflows

eatyourgreens:
Add `workflow.subjectSets`, which is a `SubjectSetStore`. Use `workflow.subjectSets.active` for subject selection.

Based off #1804.

~~Set to draft because workflow store tests are failing with invalid reference errors.~~

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/Workflow/SubjectSetStore/SubjectSetStore.js
+++ b/packages/lib-classifier/src/store/Workflow/SubjectSetStore/SubjectSetStore.js
@@ -1,0 +1,12 @@
+import { types } from 'mobx-state-tree'
+import SubjectSet from '../SubjectSet'
+import ResourceStore from '../../ResourceStore'
+
+const SubjectSetStore = types
+  .model('SubjectSetStore', {
+    active: types.safeReference(SubjectSet),
+    resources: types.map(SubjectSet),
+    type: types.optional(types.string, 'subject_sets')
+  })
+
+export default types.compose('SubjectSetResourceStore', ResourceStore, SubjectSetStore)
--- a/packages/lib-classifier/src/store/Workflow/Workflow.js
+++ b/packages/lib-classifier/src/store/Workflow/Workflow.js
@@ -1,4 +1,4 @@
-import { flow, types } from 'mobx-state-tree'
+import { flow, tryReference, types } from 'mobx-state-tree'
 import Resource from '../Resource'
 import SubjectSetStore from './SubjectSetStore'
 import WorkflowConfiguration from './WorkflowConfiguration'
@@ -25,7 +25,8 @@ const Workflow = types
     get subjectSetId () {
       // TODO: enable selection of a subject set from the links array.
       const [ subjectSetId ] = self.links.subject_sets
-      return self.subjectSets.active?.id || subjectSetId
+      const activeSet = tryReference(() => self.subjectSets.active)
+      return activeSet?.id || subjectSetId
     },
 
     get usesTranscriptionTask () {
@@ -38,7 +39,7 @@ const Workflow = types
   .actions(self => {
     function * selectSubjectSet(id) {
       yield self.subjectSets.setActive(id)
-      return self.subjectSets.active
+      return tryReference(() => self.subjectSets.active)
     }
 
     return {
--- a/packages/lib-classifier/src/store/Workflow/Workflow.js
+++ b/packages/lib-classifier/src/store/Workflow/Workflow.js
@@ -1,5 +1,6 @@
-import { types } from 'mobx-state-tree'
+import { flow, types } from 'mobx-state-tree'
 import Resource from '../Resource'
+import SubjectSetStore from './SubjectSetStore'
 import WorkflowConfiguration from './WorkflowConfiguration'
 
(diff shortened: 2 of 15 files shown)


eatyourgreens:
Tests should be passing now.

packages/lib-classifier/src/store/Workflow/SubjectSetStore/SubjectSetStore.js
```diff
@@ -0,0 +1,12 @@
+import { types } from 'mobx-state-tree'
+import SubjectSet from '../SubjectSet'
+import ResourceStore from '../../ResourceStore'
+
+const SubjectSetStore = types
+  .model('SubjectSetStore', {
+    active: types.safeReference(SubjectSet),
+    resources: types.map(SubjectSet),
+    type: types.optional(types.string, 'subject_sets')
```
srallen, line 9: Thinking about this, maybe this should be `types.literal('subject_sets'). I don't know when we would ever want this to be undefined or something else. Out of scope, though, and we would want to make this change across store models.

packages/lib-classifier/src/store/Workflow/Workflow.js
```diff
@@ -35,12 +36,13 @@ const Workflow = types
   }))
 
   .actions(self => {
-    function selectSubjectSet(id) {
-      self.activeSubjectSet = id
+    function * selectSubjectSet(id) {
+      yield self.subjectSets.setActive(id)
+      return self.subjectSets.active
```
srallen, line 41: Should this have a try...catch? What if this fails?

packages/lib-classifier/src/store/WorkflowStore/WorkflowStore.js
```diff
@@ -61,7 +61,8 @@ const WorkflowStore = types
         self.resources.put(workflow)
         if (subjectSetID) {
           const selectedWorkflow = self.resources.get(id)
-          selectedWorkflow.selectSubjectSet(subjectSetID)
+          // wait for the subject set to load before activating the workflow
+          const subjectSet = yield selectedWorkflow.selectSubjectSet(subjectSetID)
```
srallen, line 65: Same question about error catching here, I'm not sure if/where an error would be caught for this part of the loading.

srallen (approved):
LGTM from just a read of the code and passing tests locally. I think we can manually test the branch this merges into. I have a few inline comments, but none are blocking.

packages/lib-classifier/src/store/Workflow/Workflow.js
eatyourgreens: Maybe it should be wrapped in `tryReference`?

srallen: Yeah, that sounds good and maybe a test for this possibility? What do we expect if this is undefined?

packages/lib-classifier/src/store/Workflow/SubjectSetStore/SubjectSetStore.js
eatyourgreens: From memory, literals error if they aren't present in the initial snapshot. So, this line allows us to create a new store from an empty object?

packages/lib-classifier/src/store/Workflow/Workflow.js
eatyourgreens: Return null or undefined I guess. I can't remember which is returned by tryReference for an invalid reference. At the moment, I don't think the return value is used. It's there because MST actions can't handle async/await.

packages/lib-classifier/src/store/WorkflowStore/WorkflowStore.js
eatyourgreens: I'm not sure either. The classifier renders with an empty subject viewer, so the error is being caught somewhere. Here's the classifier when I edit the URL to include an invalid subject set ID.

Screenshot of the classifier with an empty subject set viewer and the dev tools open to show MobX State Tree invalid reference errors for the subject sets store.

packages/lib-classifier/src/store/Workflow/Workflow.js
eatyourgreens: Looking at the tests, they need to be updated to use `async` anyway, since `selectSubjectSet` now returns a promise. And the API client needs to be mocked in the tests. And I've broken the Workflow factory with the changes in this PR.

I'm working on that this morning.

eatyourgreens: I've completely rewritten the tests now so the async functions should be tested correctly and cases where `subjectSets.fetchResource(id)` returns undefined are included.

eatyourgreens:
Tests are passing but the code isn't working for me locally, so I'm setting this back to draft.

eatyourgreens:
This works for me in the dev classifier:https://localhost:8080/?env=production&project=12754&workflow=15652&subjectSet=85773

A production build of the project app also works locally with the test project:
http://localhost.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020?env=production

I've been seeing errors with the development build of the project app, locally, but I'm not sure why that would fail if the production build is ok.

PR #1809, opened 2020-09-18
Render project workflows on the server

### LINKED ISSUE #1803: Project: reduce network requests from the Hero component
**Package**
app-project

The project Hero component makes API requests for a project's active workflows when it mounts. Linked subject sets are also loaded, even if the project home page doesn't display them.
https://github.com/zooniverse/front-end-monorepo/blob/672df219cdd67a7b6d17896fda353c943d68244a/packages/app-project/src/screens/ProjectHomePage/components/Hero/HeroContainer.js#L28-L30

To see this, try the following with the Network tools open. Look for API requests that fetch `workflows` and `set_member_subjects`.
1. Visit https://frontend.preview.zooniverse.org/projects/zookeeper/galaxy-zoo
1. Click through to classify.
1. Click the link back to the project home page.

Active workflows should be requested once. The request isn't authenticated, so could be made on the server.

Ideally, subject sets should only be requested by projects that use grouped workflows (but we don't know if workflows are grouped until we've requested them from the API.)

### LINKED ISSUE #1827: Project home page: workflow loading broken in Safari.
**Package**
app-project

**Describe the bug**
The workflow selector in the project Hero component shows: 'There was an error fetching the workflows :('

**To Reproduce**
Load any project home page that has active workflows eg. https://frontend.preview.zooniverse.org/projects/cseidenstuecker/every-name-counts

Safari 12.1 on macOS 10.14

The culprit is `Promise.allSettled` here, which isn't supported in Safari 12. It should fallback to `Promise.all` instead of failing.
https://github.com/zooniverse/front-end-monorepo/blob/ab839d266676658e253abe2c489679e6fb59f02e/packages/app-project/src/screens/ProjectHomePage/components/Hero/helpers/fetchWorkflowsHelper/fetchWorkflowsHelper.js#L12-L14

eatyourgreens:
Request workflows in `getDefaultPageProps`, then pass the workflow data down to the Hero component as a prop. Wait for user login to complete before displaying the list of workflows in the browser.

This is now based off #1823 so that `getServerSideProps` is consistent across branches.

Package:
app-project

Closes #1803.
Fixes #1827.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
Tests are broken and the code needs re-organising, but this works, which is very promising.

packages/app-project/src/screens/ProjectHomePage/components/Hero/HeroContainer.js
```diff
@@ -7,84 +7,20 @@ import WideLayout from './components/WideLayout'
...
-    }
-  }
 
   render () {
-    const { workflows } = this.state
-    const { isWide } = this.props
+    console.log('hero props', this.props)
+    const { isWide, workflows } = this.props
+    const workflowData = {
+        loading: asyncStates.success,
```
eatyourgreens, line 16: Here I'd hold off on displaying the workflows until login has finished, so this loading state should reflect the user loading state.

packages/app-project/src/screens/ProjectHomePage/components/Hero/HeroContainer.js
```diff
@@ -5,86 +5,21 @@ import asyncStates from '@zooniverse/async-states'
...
-          loading: asyncStates.error
-        }
-      }))
-    }
-  }
 
   render () {
-    const { workflows } = this.state
-    const { isWide } = this.props
+    console.log('hero props', this.props)
```
eatyourgreens, line 12: Stray console log.

packages/app-project/pages/projects/[owner]/[project]/index.js
```diff
@@ -1 +1,20 @@
...
+}
+export async function getServerSideProps({ params, query, req, res }) {
+  const { owner, project } = params
+  const { env } = query
+  const projectQuery = {
+    env,
+    slug: `${owner}/${project}`
+  }
+  const response = await projects.get({ query: projectQuery })
+  const workflows = await getProjectWorkflows(response.body.projects)
```
Comment (eatyourgreens) [inline packages/app-project/pages/projects/[owner]/[project]/index.js:17]: The API environment (production or staging) should be passed along with the workflows query here.

eatyourgreens:
This is neat. NextJS serialises the project and workflows data as JSON, then requests
http://localhost.zooniverse.org:3000/_next/data/pKc0Nl5BF0VaThy5N0Po4/projects/darkeshard/engaging-the-crowds-2020.json
when you visit the home page from the Classify page.

packages/app-project/src/screens/ProjectHomePage/components/Hero/HeroContainer.js
eatyourgreens: Resolved by moving loading state down to the workflow selector itself.

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowSelector/WorkflowSelectorContainer.js
```diff
@@ -4,21 +4,26 @@ import React, { Component } from 'react'
 
 import WorkflowSelector from './WorkflowSelector'
 
-function storeMapper (stores) {
+function storeMapper(stores) {
+  const { project, user } = stores.store
   return {
-    workflowDescription: stores.store.project.workflow_description
+    loadingState: user.loadingState,
+    workflowDescription: project.workflow_escription
```
srallen, line 11:
```suggestion
    workflowDescription: project.workflow_description
```
typo here

srallen (changes requested):
We haven't really had `index.js` files hold code before, just handle import/export of modules. I don't think I would think to look there for that. I would prefer organizationally that the async requests added to the `packages/app-project/pages/projects/[owner]/[project]/index.js` file be put elsewhere ideally with specs. We've used folders called `helpers` elsewhere to contain one off functions with specs.

eatyourgreens:
I'm hoping #1823 will make things cleaner by moving the common page props, like the project, into a shared function. However, with NextJS 9.3 and up, we're going to have to add functions to individual pages. It's probably a good idea to keep those functions short. I wasn't sure how to test the page props here, without just duplicating the `fetchWorkflowsHelper` tests.

eatyourgreens:
I've tidied up getting the page props from the request. #1823 should allow us to replace the duplicated project fetch (once in `getInitialProps` and once here) with something like:
```js
const { props: defaultProps } = await getDefaultPageProps({ params, query, req, res })
const language = defaultProps.initialState.project.primary_language
const { active_workflows, default_workflow } = defaultProps.initialState.project.links
const workflows = getWorkflowsHelper(language, active_workflows, default_workflow)
const props = Object.assign({}, { workflows }, defaultProps)
return { props }
```

eatyourgreens:
For the content pages app, I've experimented with putting the page data fetching functions under `/src/screens`:
https://github.com/zooniverse/front-end-monorepo/blob/master/packages/app-content-pages/src/screens/Publications/getStaticProps.js

I'm not 100% sold on that approach because I wouldn't necessarily look in `src/screens` for functions that handle HTTP requests. The project workflow and subject set pages also don't have components under `src/screens`, since they re-use `ClassifyPage`. Neither of those are big problems, in my mind. We could export `src/screens/ProjectHomePage/getServerSideProps` from `pages/project/[owner]/[project]/index` and `src/screens/ClassifyPage/getServerSideProps` from every page under `pages/project/[owner]/[project]/classify`

eatyourgreens:
I ended up copying the tests for `getServerSideProps` from the tests for `fetchWorkflowsHelper`, so there’s a lot of overlap between those specs.

packages/app-project/src/screens/ProjectHomePage/ProjectHomePage.js
```diff
@@ -23,13 +23,14 @@ const RemainingHeightBox = styled(Box)`
 `
 
 function ProjectHomePage (props) {
+  const { workflows } = props
```
srallen, line 26: Non-blocking, but this should have a prop type definition.

packages/app-project/src/screens/ProjectHomePage/getServerSideProps.js
```diff
@@ -0,0 +1,18 @@
...
+    slug: `${owner}/${project}`
+  }
+  const response = await projects.get({ query: projectQuery })
+  const [projectData] = response.body.projects
+  const language = projectData.primary_language
+  const { active_workflows, default_workflow } = projectData.links
+  const workflows = await fetchWorkflowsHelper(language, active_workflows, default_workflow, env)
+  const props = { workflows }
+  return ({ props })
+}
```
srallen, line 18: What if these requests fail? What kind of handling should these have for different scenarios, say 404 vs 500s?

srallen (changes requested):
Non-blocking, but I don't think it's clear upon reading the `WorkflowSelector` code that it's contingent on the _user_ loading state, not the workflow loading state. I think given the name of the component, one would assume it's the workflow. I think this could be better named to avoid this confusion. 

Also, what happens now when the workflow fails to load? The UI has been changed to only display certain UI messaging based on user load state, not workflow, so what if the workflow request fails?

packages/app-project/src/screens/ProjectHomePage/getServerSideProps.js
eatyourgreens: Good question. What happens at the moment? Doesn't the component display a message about workflows failing to load?

srallen: There's a story for it: https://zooniverse.github.io/front-end-monorepo/@zooniverse/fe-project/index.html?path=/story/project-app-screens-project-home-hero--error

eatyourgreens: We could send a project workflows error state back with the page props. The generated HTML might be cached across page refreshes, which could be a problem if it's a temporary error. 

Uncaught errors on the server will generate a 500 response, but that won't be cached. The error page is customisable, so setting up a project error page might be a good first issue. That ties into having a custom 404 page too.

eatyourgreens:
This can't be deployed until #1826 is fixed in production, so I'm marking it as draft.

packages/app-project/src/screens/ProjectHomePage/getServerSideProps.js
eatyourgreens: NextJS recommend not handling server-side errors, although there's next-to-no official documentation about that, just discussions about best practice. https://github.com/vercel/next.js/issues/11644 has some discussion about error-handling.

If we do handle errors, and send back a 200 response with a project home page, then it's worth noting that this can be indexed by search engines (as well as cached by browsers.)

eatyourgreens: Next 10 adds [`notFound` and `redirect` keys](https://nextjs.org/blog/next-10#redirect-and-notfound-support-for-getstaticprops--getserversideprops) to the returned object from `getServerSideProps`. That could be useful.

Also worth noting that the workflows request can't 404. `notFound` would only be needed when requesting a specific resource.

eatyourgreens:
Rebased to bring in `getServerSideProps` from #1823, so that should be reviewed before this PR.

eatyourgreens:
I've done some reading up on server-side errors in NextJS. I think we have two choices.

1. Any uncaught error in `getServerSideProps` returns a 500 response and the app's error page. The default error page is a blank page saying 'An unexpected error has occurred' (see the screenshot in https://github.com/zooniverse/front-end-monorepo/issues/1826.) NextJS recommends letting Next handle your server-side errors and set up [a custom error page](https://nextjs.org/docs/advanced-features/custom-error-page#500-page). This seems like the appropriate response, to me, if Panoptes is down.
2. We wrap `getServerSideProps` in `try… catch` and return some default set of props in case of an error. In this case we'd send back a 200 response along with a page that includes an error message. That makes me slightly nervous because a 200 response can be cached by the browser and might not refresh immediately on page reload.

packages/app-project/src/screens/ProjectHomePage/getServerSideProps.js
srallen: I'm fine with punting this issue for version 10 if it has a better API for it. Let's track in an issue?

srallen (approved):
LGTM. I'm satisfied having `getServerSideProps` in the index file of the pages for now. However, one recommendation I have is to at least add basic smoke unit tests for these and this would I think be sufficient to signal to another dev that there's code in these files. 

> Any uncaught error in getServerSideProps returns a 500 response and the app's error page. The default error page is a blank page saying 'An unexpected error has occurred' (see the screenshot in #1826.) NextJS recommends letting Next handle your server-side errors and set up a custom error page. This seems like the appropriate response, to me, if Panoptes is down.

This seems like the right choice to me too.

eatyourgreens:
In #1964 I’ve figured how to customise the error message. I haven’t experimented much with it, but that might work for 500 errors too.

eatyourgreens:
One thing I haven’t figured out: how to add tests to pages. Every JS file under `/pages` is a page on the site. Right now, my approach to that is to imported helpers, which have tests, and run those in `getServerSideProps`.

eatyourgreens:
This might work. Put the page tests in `/src/tests`. https://medium.com/frontend-digest/testing-getserversideprops-in-nextjs-b339ebcf3401

PR #1814, opened 2020-09-22
add prod service pod disruption budgets

camallen:
Ensure we have these service pods running in our k8s cluster

Package:

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


zwolf (changes requested):
This is new. If I'm understanding the docs, then currently the minimum number of running pods for `project-production` is 2. So if there are 2 running, and the minAvailable in the PDB is 2, then...will they ever be replaced?

>If you set maxUnavailable to 0% or 0, or you set minAvailable to 100% or the number of replicas, you are requiring zero voluntary evictions. When you set zero voluntary evictions for a workload object such as ReplicaSet, then you cannot successfully drain a Node running one of those Pods. 

In this case minAvailable == minReplicas ==  2 in prod/1 in staging. So a deploy would never voluntarily evict a pod, and the new pod would never deploy. Would it be better to set this to minReplicas - 1 to ensure there's at least 1 running pod?

camallen:
> This is new. If I'm understanding the docs, then currently the minimum number of running pods for `project-production` is 2. So if there are 2 running, and the minAvailable in the PDB is 2, then...will they ever be replaced?

Pod disruption budgets ensure that a minimum number of pods are running at any time in the cluster. This may cause issues when doing `node drain` events etc, TBD i guess.

The HPA & replica set will ensure the number of running pods matches our desired targets, taking into account any issues with the cluster. 

> In this case minAvailable == minReplicas == 2 in prod/1 in staging. So a deploy would never voluntarily evict a pod, and the new pod would never deploy. 

Deploys won't be impacted by this change. I copied the existing setup in panoptes to ensure we have at least 2 api nodes running at all times, https://github.com/zooniverse/panoptes/blob/de7a2c9cae481979f55753366eba1bf0183dde10/kubernetes/deployment-production.tmpl#L243-L251

> Would it be better to set this to minReplicas - 1 to ensure there's at least 1 running pod?

We want at least 2 to handle the traffic profile we want for our apps (and ensure they are available).

zwolf:
The quote from the docs seems to say one thing while our Panoptes deploys indicate another. Maybe I'm still missing something about "voluntary evictions" or "node draining". If 2 pods must always be running, and there are currently two pods running, I don't see how we could evict them to deploy. It clearly works, though, so objection withdrawn.

zwolf (approved):

camallen:
> The quote from the docs seems to say one thing while our Panoptes deploys indicate another. Maybe I'm still missing something about "voluntary evictions" or "node draining". If 2 pods must always be running, and there are currently two pods running, I don't see how we could evict them to deploy. It clearly works, though, so objection withdrawn.

👍 Let's get this out and see how our deploys behave. We can always modify the PDB to better fit our needs

PR #1823, opened 2020-09-24
Project app: use getServerSideProps on all project pages

eatyourgreens:
This removes `getInitialProps` from the project app and replaces it with [`getServerSideProps`](https://nextjs.org/docs/basic-features/data-fetching#getserversideprops-server-side-rendering) for each page. Common props (the project, theme and banner settings) are moved into a `getDefaultPageProps` helper.

I think you'll need to test against a production build in order to use server-side props without building the page on each request so this might best be tested with:
```
yarn build
yarn start
```

Package:
app-project

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
This can't be deployed until #1826 is fixed in production, so I'm marking it as draft.

eatyourgreens:
Removing draft status because we'll need this for the Engaging Crowds subject pages (see #1875.)

shaunanoordin (changes requested):
## Approve

This PR consolidates the `getServerSideProps()` function, which has hitherto been initialised in the parent `_app` and then scattered across several sub-pages (e.g. `/projects/foo/bar/classify`, `/projects/foo/bar/classify/workflow/1234`, etc)

- 👍 Code changes look good to me; this seems to be mostly a refactor which will be relevant as we add more sub-pages.
- Testing using `yarn build ; yarn start` works fine in most scenarios
- ❌❓  Testing using `yarn dev`, however, brings up some question marks.
  - Uh, see the Dev Notes for more deets. Specifically, **Test 1b** and **Test 2**
- ❔ Functionality: I couldn't figure out how I'd break this code, so this is... good?
  - I get the impression that the biggest thing the props-getting code does, at the moment, is getting the cookie values for "announcement banner dismissed?" and "light/dark mode?"
  - However, this functionality doesn't seem to work even on `master`, so I don't have a _functionality baseline_ to compare this branch with, so I'm ignoring functionality tests.

TL;DR: code changes look good, but what is up with `yarn dev`?

### Dev Notes

Tested using macOS+Chrome

**Test 1**

branch: `project-initial-props`
urls:
- `http://localhost:3000/projects/darkeshard/engaging-the-crowds-2020`
- `http://localhost:3000/projects/darkeshard/engaging-the-crowds-2020/classify`
- `http://localhost:3000/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/15764`

**Test 1a**

command: `yarn build ; yarn start`
results:
- 👍 pages loaded & rendered fine. Classifier functions normally.
- ⚠️ switching to Dark/Light Theme isn't saved to cookies?
  - But this is true for `master` also, so shrug? 🤷‍♂️ 

**Test 1b**

command: `yarn dev`
results:
- ❌ the pages keep spitting out errors
  - `engaging-the-crowds-2020` = _"Server Error. Error: Error serializing `.initialState.project.error` returned from `getServerSideProps` in "/projects/[owner]/[project]". Reason: `object` ("[object Error]") cannot be serialized as JSOΩN. Please only return JSON serializable data types."_
  - `engaging-the-crowds-2020/classify` = _"Server Error. Error: Error serializing `.subjectSetID` returned from `getServerSideProps` in "/projects/[owner]/[project]/classify". Reason: `undefined` cannot be serialized as JSON. Please use `null` or omit this value all together."_
  - _`engaging-the-crowds-2020/classify/workflow/15764` = _"Server Error. Error: Error serializing `.subjectSetID` returned from `getServerSideProps` in "/projects/[owner]/[project]/classify/workflow/[workflowID]". Reason: `undefined` cannot be serialized as JSON. Please use `null` or omit this value all together."_
- ❓ BUT, this is ALSO true for `master`. The error on `master`, for the Classifier page, for the Engaging Crowds project is `There was an error in the classifier :( `
  - If you look below at Test 2, I Fancy Cats has _no_ similar issues on master, but _does_ have problems on this branch.

**Test 2**

command: `yarn dev`
url: `http://localhost:3000/projects/brooke/i-fancy-cats/classify` I FANCY CATS

branch: `master`
results: 🆗 

branch: `project-initial-props`
results: ❌ error: _"Server Error. Error: Error serializing `.subjectSetID` returned from `getServerSideProps` in "/projects/[owner]/[project]/classify". Reason: `undefined` cannot be serialized as JSON. Please use `null` or omit this value all together."_

### Status

I wanted to give this a 👍 since the code changes/refactor looks good, but the extended testing with `yarn dev` made me scratch my head. 😖  There are a few problems that only trigger at the specific combination of this branch + yarn dev, while other problems are ongoing issues with master, and it's a bit hard for me to distinguish which are specific to this PR.

**Change Request:**

 @eatyourgreens can you please try checking if your projects run fine with `yarn dev`? If they're OK, I'm going to give this PR an approval and run some follow up tests on `master` instead.

eatyourgreens:
Thanks for the thorough review @shaunanoordin. Undefined values aren't valid JSON, and page props are serialised to JSON in the latest versions of NextJS (since 9.3 I think.) If you switch between the home page and classify page, and back, you should see the browser making requests for the page props as JSON on each client-side page transition.

Interesting that this works in a production build but fails in development. I think I'll have to filter out undefined values when building the page props, or give each page its own `getServerSideProps` function.

eatyourgreens:
>`engaging-the-crowds-2020` = _"Server Error. Error: Error serializing `.initialState.project.error` returned from `getServerSideProps` in "/projects/[owner]/[project]". Reason: `object` ("[object Error]") cannot be serialized as JSOΩN. Please only return JSON serializable data types."_

I think I'm going to have to dive into how API errors are being handled in the project store. This sounds like the JavaScript error object is being stored, but can't be serialised. The `initialState` prop is a snapshot, which is used to hydrate the store in the browser. Here, the project has failed to load and that's broken the snapshot.

eatyourgreens:
There's a property in the Engaging Crowds 2020 project which breaks the project model. I think this might be a separate issue, but I'm not sure. This is what the server logs when I load that project's home page.
```
Initialising Sentry: undefined production 3d64b7b949ae0fb1daca40aec410f1e73153f01a
Error loading project: Error: [mobx-state-tree] Error while converting `null` to `string`:
```

shaunanoordin:
Thanks for confirming, @eatyourgreens ! If it looks like the problem is specific to Engaging Crowds 2020, we'll isolate investigations to that project.

I'm going to run one last test with with I Fancy Cats (post recent update), then I'll pop an approval off.

eatyourgreens:
TIL: giving `workflowID` a default value of null breaks the classifier.

I've added conditional checks before assigning `workflowID` and `subjectSetID` props to the Classify page.

shaunanoordin:
Ahh I was waiting for 4d40629e9b03aa6a2169b33dad741ed607b20d0b ! Yeah, I traced some problems in I Fancy Cats to this specific undefined-value-assignment as well. One sec, re-pulling the branch...

eatyourgreens:
All project use `/classify` for the default workflow, so the classifier must be using that if a workflow ID isn't present in the URL.

shaunanoordin:
I need to check something - I'm looking at commit ad7bf1d , and in this line...

```
// getServerSideProps.js
props.workflowID = project.configuration['default_workflow']
```

where is `project` coming from?

eatyourgreens:
D'oh! One moment while I fix that.

shaunanoordin:
We're almost there Jim! I think there's only one last possible glitch to worry about.

In this bit of code..

```
  if (workflowID) {
    props.workflowID = workflowID
  } else {
    props.workflowID = project?.configuration['default_workflow']
  }

// This works for for `yarn build ; yarn start`, but breaks for `yarn dev`, on I Fancy Cats
```

I'd recommend that it be changed to...

```
  if (workflowID) {
    props.workflowID = workflowID
  } else if (project?.configuration['default_workflow']) {
    props.workflowID = project?.configuration['default_workflow']
  }

// This works for `yarn dev`, on I Fancy Cats. Gonna test for `yarn build ; yarn start` next.
```

I think this is because the `else` clause works against the initial `if (workflow)` check. The `else` clause ensures that no matter what the `props` will have a workflowID, and if there's no project, then that's `props.workflowID = undefined`

eatyourgreens:
If there’s no project, I want the workflow ID to be undefined. At least, I only want you to be able to classify if there’s a project and workflow.

packages/app-project/src/screens/ClassifyPage/getServerSideProps.js
```diff
@@ -0,0 +1,16 @@
...
+  const { props } = await getDefaultPageProps({ params, query, req, res })
+  const { project } = props.initialState
+  if (subjectSetID) {
+    props.subjectSetID = subjectSetID
+  }
+  if (workflowID) {
+    props.workflowID = workflowID
+  } else {
+    props.workflowID = project?.configuration['default_workflow']
+  }
```
shaunanoordin, line 14:
```suggestion
  } else if (project?.configuration['default_workflow']) {
    props.workflowID = project?.configuration['default_workflow']
  }
```

shaunanoordin (changes requested):
## PR Review (Update)

Changes are looking good, with only one suggested change remaining. I think the key takeaway here is that the system super duper hates it if `getServerSideProps()` returns anything with `{ someProperty: undefined }`

### Testing

I've streamlined my testing to use only the prototypical I Fancy Cats project, so I can sidestep the project-specific problems I encountered in the earlier review.

command: `yarn build ; yarn start`
url: `http://localhost:3000/projects/brooke/i-fancy-cats?env=staging`
result: 👍 

command: `yarn build ; yarn start`
url: `http://localhost:3000/projects/brooke/i-fancy-cats/classify?env=staging`
result: 👍 

command: `yarn dev`
url: `http://localhost:3000/projects/brooke/i-fancy-cats`
result: 👍 

command: `yarn dev`
url: `http://localhost:3000/projects/brooke/i-fancy-cats/classify`
result: ❌

The error, as expected, was that `getServerSideProps` was trying to pass a value of `.workflowID = undefined`. This reflects the [earlier comment](https://github.com/zooniverse/front-end-monorepo/pull/1823#issuecomment-755450418) I made.

```
Server Error
Error: Error serializing `.workflowID` returned from `getServerSideProps` in "/projects/[owner]/[project]/classify".
Reason: `undefined` cannot be serialized as JSON. Please use `null` or omit this value all together.
```

### Status

Changes reque- wait,

UPDATE: while I was writing this review, @eatyourgreens replied to my earlier comment with the following:

> If there’s no project, I want the workflow ID to be undefined. At least, I only want you to be able to classify if there’s a project and workflow.

Ah, in that case you need to figure out how to handle the issue when you run `npm run dev` and try to classify on I Fancy Cats. Additional details shortly.

shaunanoordin:
**Follow up details**

If we have the `else` block, then `getServerSideProps()` has the potential to return `props` that looks like...

```
{
  host: 'http://localhost:3000',
  initialState: {
    project: {
      ...bla bla bla...
    },
    ui: { dismissedProjectAnnouncementBanner: null, mode: 'light' }
  },
  isServer: true,
  query: { owner: 'brooke', project: 'i-fancy-cats' },
  workflowID: undefined  // <-- this thing over here
}
``` 

No idea why it _only_ occurs for `yarn dev` though.

eatyourgreens:
Weird. Maybe I Fancy Cats doesn’t have a default workflow set?

eatyourgreens:
For the `/classify` route, it might be safer to check if the project only has one active workflow.

The logic here gets slightly hairy because one `getServerSideProps` function is being used across all the different `/classify/*` routes. Another approach would be to give each page its own `getServerSideProps` function.

eatyourgreens:
Each page under `/classify` has slightly different logic for setting the `workflowID` prop. I've given each route its own `getServerSideProps` function, so that we don't have to deal with a bunch of `if` conditionals in a single function that tries to handle all possible routes.

That's going to have a knock-on effect for #1809 and #1875 but I think it's best to deal with that once this PR is finished.

eatyourgreens:
While working on the tests, I found that the project model tests are broken, but still passing. I've opened #1960 to fix the tests.

shaunanoordin (approved):
## PR Review

The latest iteration of this PR fixes the issues encountered by earlier tests.

- Changes LGTM and tests pass fine 👍 
- I am noting that in the latest iteration, each page type (`.../classify`, `.../classify/workflow/123`, `.../classify/workflow/123/subject-set/456`) has its own version of `getServerSideProps`, which contrasts with the initial PR iteration which attempted to consolidate these versions into a single universal version.

### Tests

Tests look good on...

Different commands: `yarn build ; yarn start` as well as `yarn dev`

Different projects:
- `http://localhost:3000/projects/brooke/i-fancy-cats/classify?env=staging` - a project with one (active) workflow, and NO default workflow (Note: the one and only active workflow is chosen)
- `http://localhost:3000/projects/darkeshard/transformers/classify?env=staging` - a project with multiple (active) workflows, and one default (active) workflow (Note: default workflow is correctly chosen.)

The only blorp I encountered was when you have a default workflow, but that workflow is inactive, but that's an edge case we don't need to worry about in this PR.

### Status

This PR gets a green light for me. 👍

PR #1800, opened 2020-09-15
Add nice to scaleLinear utility functions

srallen:
_Please request review from `@zooniverse/frontend` team. If PR is related to design, please request review from `@beckyrother` in addition._ 

Package: lib-classifier

Toward #1777 

Describe your changes:
This adds a `nice` chained functional call to the d3 `scaleLinear` functions. A description of what this does is in the d3 docs (https://observablehq.com/@d3/d3-scalelinear) under the `Nice Bounds` header (they don't have anchor links for hte headings).

I'm still playing around actually adding padding to the min and max but this isn't as straight forward as it seems. There's layout impacts for other positions, particularly the enable zoom button, so this will be in a follow up.

Before:
Screen Shot 2020-09-15 at 1 15 25 PM

After:
Screen Shot 2020-09-15 at 1 15 11 PM

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #1824, opened 2020-09-24
Upgrade vx to 0.0.199

srallen:

Package: lib-classifier, app-project

Describe your changes:
Updates vx to 0.0.199 in preparation of upgrading to v1. There were breaking changes between the patch version, so this fixes them for the tools we do use. The change to v1 will involve installing newly branded packages, so I decided to scope that into a separate PR. 

The switch to typescript broke all my tests testing the panning behavior because the event object is type checked and thus Enzyme's use of the `SyntheticEvent` object fails. I've temporarily set all of these tests to be pending with this explanation and opened https://github.com/airbnb/visx/issues/811.

The panning UX is a bit jumpy, but I'm unsure how to fix this while the panning is essentially untestable (unless somehow has an idea on how to approach?):

odd-panning

I wanted to get this update done so that I would be fixing the remaining bugs found for VSLCV against a stable API. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
I got the tests working from a suggestion from the issue I opened on visx. Adding `nativeEvent: new Event()` got the internal type checking to pass. Now I'm looking into improving that pan function. From the changelog, it seems like the constrain bug fix might be relevant: https://github.com/airbnb/visx/blob/master/CHANGELOG.md#bug-bug-fix-4

srallen:
The issue with the bounciness is not any vx change, but the addition of using `nice` from #1800. If you watch the gif I made closely, the number of ticks will change from 7 to 8 while panning when it should be consistently 8. Removing it for now is the easiest fix.

mcbouslog (approved):
tested:
- https://localhost:8080/?env=production&project=4878
- http://localhost:3000/projects/adamamiller/zwickys-stellar-sleuths/classify?env=production
- http://localhost:9001/?path=/story/project-app-screens-classify-daily-classifications-chart--plain

everything worked as expected, tests passing in lib-classifier and app-project

PR #1830, opened 2020-09-29
Classifier: Simple dropdown task adapter

eatyourgreens:
This adds an adapter (and tests) for the simple dropdown task, so that we can test it out in staging. Dropdown tasks are converted to simple dropdown tasks if they contain only one select menu. More complex tasks are not transformed, and should throw an error in the classifier.

Snapshot of the dev classifier with a dropdown task that has three answers: One, Two or Three.
https://localhost:8080/?project=908&workflow=3339

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/SimpleDropdownTask/models/SimpleDropdownTask.js
+++ b/packages/lib-classifier/src/plugins/tasks/SimpleDropdownTask/models/SimpleDropdownTask.js
@@ -1,5 +1,6 @@
 import cuid from 'cuid'
 import { types } from 'mobx-state-tree'
+import { panoptesAdapter } from './helpers'
 import Task from '../../models/Task'
 import SimpleDropdownAnnotation from './SimpleDropdownAnnotation'
 
@@ -14,6 +15,9 @@ const SimpleDropdown = types.model('SimpleDropdown', {
   options: types.array(types.string),
   type: types.literal('dropdown-simple'),
 })
+  .preProcessSnapshot(snapshot => {
+    return panoptesAdapter(snapshot)
+  })
   .views(self => ({
     get defaultAnnotation () {
       return SimpleDropdownAnnotation.create({
--- a/packages/lib-classifier/src/plugins/tasks/SimpleDropdownTask/models/helpers/index.js
+++ b/packages/lib-classifier/src/plugins/tasks/SimpleDropdownTask/models/helpers/index.js
@@ -1,0 +1,1 @@
+export { panoptesAdapter } from './panoptesAdapter'
--- a/packages/lib-classifier/src/plugins/tasks/SimpleDropdownTask/models/helpers/panoptesAdapter/index.js
+++ b/packages/lib-classifier/src/plugins/tasks/SimpleDropdownTask/models/helpers/panoptesAdapter/index.js
@@ -1,0 +1,1 @@
+export { panoptesAdapter } from './panoptesAdapter'
--- a/packages/lib-classifier/src/plugins/tasks/SimpleDropdownTask/models/helpers/panoptesAdapter/panoptesAdapter.js
+++ b/packages/lib-classifier/src/plugins/tasks/SimpleDropdownTask/models/helpers/panoptesAdapter/panoptesAdapter.js
@@ -1,0 +1,16 @@
+export function panoptesAdapter(snapshot) {
+  const newSnapshot = {}
+  const { selects } = snapshot
+  if (selects?.length === 1) {
+    const menu = selects[0]
+    newSnapshot.allowCreate = menu.allowCreate
+    newSnapshot.help = snapshot.help
+    newSnapshot.instruction = snapshot.instruction
+    newSnapshot.taskKey = snapshot.taskKey
+    newSnapshot.type = 'dropdown-simple'
+    const options = menu.options['*']
+    newSnapshot.options = options.map(option => option.label)
+    return newSnapshot
+  }
+  return snapshot
+}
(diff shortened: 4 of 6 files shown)


eatyourgreens:
Trying this out in the dev classifier, I'm noticing that the simple dropdown from #1772 throws a store error if you use it to make an annotation.

eatyourgreens:
I guessed at the structure of the dropdown task when writing the tests, but a complex dropdown does (correctly) error when I try to load it in the dev classifier.

I'm not sure why submitting a classification crashes the classifier.

eatyourgreens:
Here's the store error, on submitting a classification and loading a new subject.

```
You are trying to read or write to an object that is no longer part of a state tree. 
(Object type: 'SimpleDropdownAnnotation', 
Path upon death: '/classifications/resources/ckfo2dagk00022a6a2hxqqkn4/annotations/ckfo2csub00012a6aa3d7c6ou', 
Subpath: 'task', 
Action: '/classifications/resources/ckfo2dagk00022a6a2hxqqkn4.addAnnotation()'). 
Either detach nodes first, or don't use objects after removing / replacing them in the tree.
```

eatyourgreens:
I'm seeing that crash on a workflow that uses a single choice question, so I don't think it has anything to do with the dropdown task.
https://localhost:8080/?project=908&workflow=3223

eatyourgreens:
I’m thinking we can use this to get the Engaging Crowds projects up and running, but I’ve tried to keep it out of the main classifier code so that it’s easily removed/reverted when it’s no longer needed.

srallen:
@shaunanoordin are you reviewing this? Just want to be sure and don't want to double up on work already being done. Thanks!

shaunanoordin:
@srallen I am reviewing this, but I'm focusing mostly on functionality and practical tests across different workflow permutations. Please feel free to add your own review; you have a much better eye for how things fit into the larger monorepo code, which I can easily miss. 👍

shaunanoordin (approved):
## PR Review

Requires: #1772 
Package: `lib-classifier`

PR #1772 introduces the "simple-dropdown" Task (data model) and its corresponding "Simple Dropdown" Task component (the UI). This PR adds an adapter that handily transforms classic "dropdown" Tasks (data model) into the "simple-dropdown" Tasks.

This in turn allows classic dropdown tasks (created in the PFE project builder) to "work" using the new Simple Dropdown Task system, without having to 1. make the new Simple Dropdown Task system too backwards-compatible, or 2. change every existing classic dropdown Task in PFE to match the new standard.

- The adapter only works on very basic (non-cascading) classic dropdown configurations.

### Testing

Tested on localhost with macOS + Chrome

**Case 1: simple-dropdown Task**

URL: https://local.zooniverse.org:8080/?env=production&project=12754&workflow=16106&subjectSet=85771

- We're testing against a `simple-dropdown` Task workflow.
- This ensures that changes introduced in 1830 doesn't adversely affect the functionality introduced by 1772.
- Tests look good. 👍 Simple Dropdown tasks continue to function properly. The panoptesAdapter doesn't trigger when the workflow Task is a simple dropdown. This is correct. 👍 

**Case 2: classic dropdown Task**

URL: https://local.zooniverse.org:8080/?env=production&project=12754&workflow=15764
Compare with PFE: https://www.zooniverse.org/projects/darkeshard/engaging-the-crowds-2020/classify?reload=0&workflow=15764

- We're testing the main functionality of this PR. 
- Baseline: without this PR, a classic dropdown Task will not show on the Task panel (i.e. it has no corresponding Task component)
- With this PR: a classic (non-cascading) dropdown Task will appear on the Task panel using the Simple Dropdown component.
- Tests look good. 👍 Classic dropdown tasks are 'transmuted' into Simple Dropdown tasks. The adapter triggers as expected.

⚠️ Important thing to note for any PFE projects being 'imported' to the monorepo: on PFE, `dropdown` tasks will have an annotation that looks like `{ value: "ad49e86d82968", option: true }` while on monorepo, the same task will have annotations like `{ option: true, selection: 1 }`

### Dev Notes

Functionality looks good! 👍 I can't think of any other workflow permutations that need testing.

One thing I'd add is that it's probably a good idea to add a quick bit of documentation to explain `panoptesAdapter.js` and the context of its creation. Reading `if (selects?.length === 1)` makes sense to _me,_ but others might scratch their head for longer than necessary.

Some `// inline comments` would be ace, a small README.md would be overkill but good, or maybe even...

```
const { selects } = snapshot
const isThisAClassicDropdownTaskThatNeedsToBeTransmogrifiedIntoASimpleDropdownTask = selects?.length === 1

if (isThisAClassicDropdownTaskThatNeedsToBeTransmogrifiedIntoASimpleDropdownTask) {
...
```

### Status

This PR LGTM, though I do want to add a note that additional documentation to help people understand the context (i.e. currently very specific use) of the very general-sounding Panoptes Adapter.

srallen (commented):
I would suggest that the snapshot processor check for min and max number of options that is proposed from the ADR (See: #1767) and to not allow those that are less than or greater than the amount we've agreed to. An error could be thrown, an error state set, and a useful UI message communicated. 

Secondly, I agree with @shaunanoordin that a read me for the snapshot preocessor would be helpful and that the naming `panoptesAdapter` isn't really descriptive.

I don't consider these blocking, but would highly recommend it.

eatyourgreens:
Let's merge these, so that Engaging Crowds can get going, then open a new PR with the suggestions.

PR #1831, opened 2020-09-30
[VSLCV] Responsive grid

### LINKED ISSUE #1548: [VSLCV] Fallback for small screen sizes
**Package** lib-classifier

One of the ideas from the last discussion with the variable star researchers is to have a message come up on small screen sizes that states that this data is best viewed on tablets and desktop rather than try to come up with a responsive view of the data. We need to make a few decisions:

- What's the screen size cut off for this?
- What should the message say?
- How should the message be displayed?

This effectively scopes this project to not be able to be participated on with mobile users so we should confirm this is ok before designing for it.

srallen:

Package: lib-classifier

Closes #1548

Describe your changes:
This add's vx's `withParentSize` HOC to the `VariableStarViewer` to get the parent container's width as a prop. If the parent container width is less than or equal to 768, then we fallback to a single column layout. This isn't great, but it's better than nothing and we can consider additional messaging or alternatives based on if we get feedback about it during the beta.

stacked-layout

This can be viewed by running `yarn bootstrap` first from the parent repo folder, then going into the `app-project` folder and running `yarn start` and going to http://localhost:3000/projects/adamamiller/stellar-sleuths/classify. Then you can use dev tools to switch to any phone or tablet layout.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewer.js
@@ -182,13 +182,17 @@ const VariableStarViewer = React.forwardRef((props, ref) => {
       >{Object.keys(barJSON).map((barChartKey) => {
         //Let's keep the rendering of the bar chart flexible in case more plots are added in the future
         return (
-          <BarChartViewer
-            data={barJSON[barChartKey].data}
+          <Box
+            basis='1/2'
             key={barChartKey}
-            xAxisLabel={barJSON[barChartKey].chartOptions.xAxisLabel}
-            yAxisDomain={barJSON[barChartKey].chartOptions.yAxisDomain}
-            yAxisLabel={barJSON[barChartKey].chartOptions.yAxisLabel}
-          />
+          >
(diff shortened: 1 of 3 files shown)


beckyrother (approved):
Nice!

ErikOstlund:
On initial page load, all charts look good. When I resize the window below 1407px, the charts layout change to single column. But, if I resize the window greater than 1407px, the top right chart extends under the task component. This also happens if I resize the window from a large size to a smaller size, but not yet hit the single column breakpoint.
I'm not sure what the best solution would be? Maybe consider only using a single column layout any any screen size?
Screen Shot 2020-10-01 at 12 21 06 PM

ErikOstlund (changes requested):
See previous comment.

srallen:
@ErikOstlund I fixed this by wrapping the bar charts in a Grommet `Box` (this is a component that auto applies flex box styling to a div) and set the wrapper's flex-basis to 1/2. Could you have a look again? Thanks!

ErikOstlund (approved):
The fix is looking good! Charts are now responsive and do not extend under the task component.
All tests are passing locally.
👍🏽

PR #1841, opened 2020-10-05
Task models: remove unused actions and use tryReference to access annotations

eatyourgreens:
Wrap any calls to `self.annotation` inside `tryReference`, since `task.annotation` stores a reference to an annotation and not the annotation itself.
Remove `task.updateAnnotation(value)`. Task components should call `annotation.update(value)` on their annotation prop.
Fix annotation references to explicitly use the annotation ID.

https://mobx-state-tree.js.org/concepts/references

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/models/Task.js
+++ b/packages/lib-classifier/src/plugins/tasks/models/Task.js
@@ -44,7 +44,7 @@ const Task = types.model('Task', {
     },
 
     setAnnotation (annotation) {
-      self.annotation = annotation
+      self.annotation = annotation.id
     },
 
     start () {
--- a/packages/lib-classifier/src/plugins/tasks/models/Task.js
+++ b/packages/lib-classifier/src/plugins/tasks/models/Task.js
@@ -1,5 +1,5 @@
 import cuid from 'cuid'
-import { types } from 'mobx-state-tree'
+import { tryReference, types } from 'mobx-state-tree'
 import Annotation from './Annotation'
 
 const Task = types.model('Task', {
@@ -21,18 +21,20 @@ const Task = types.model('Task', {
     },
 
     get isComplete () {
-      return !self.required || !!self.annotation && self.annotation.isComplete
+      const annotation = tryReference(() => self.annotation)
+      return !self.required || !!annotation?.isComplete
     }
   }))
   .actions(self => ({
     updateAnnotation (value) {
-      self.annotation && self.annotation.update(value)
+      const annotation = tryReference(() => self.annotation)
+      annotation?.update(value)
     },
 
     complete () {
       // set an annotation for this task if there wasn't one already.
-      const { value } = self.annotation
-      self.updateAnnotation(value)
+      const annotation = tryReference(() => self.annotation)
+      self.updateAnnotation(annotation?.value)
     },
 
     createAnnotation () {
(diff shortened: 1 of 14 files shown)


packages/lib-classifier/src/plugins/tasks/models/Task.js
```diff
@@ -21,18 +21,20 @@ const Task = types.model('Task', {
...
+      const annotation = tryReference(() => self.annotation)
+      annotation?.update(value)
     },
 
     complete () {
       // set an annotation for this task if there wasn't one already.
-      const { value } = self.annotation
-      self.updateAnnotation(value)
+      const annotation = tryReference(() => self.annotation)
+      self.updateAnnotation(annotation?.value)
```
srallen, line 37: Do we actually want this action called if the annotation is undefined?

srallen (changes requested):
I discovered while debugging that line 52 is setting the `safeReference` type to the full resource, not the identifier of the resource, so this should be corrected to:
```
self.annotation = annotation.id
```

I would like to see this fix rolled into this PR.

There should be added specs for the behavior when the annotation is defined vs undefined. For instance, like my inline comment, I don't think we should be actually calling the action to update the annotation if it is undefined? There may be unintended side effects?

packages/lib-classifier/src/plugins/tasks/models/Task.js
eatyourgreens: I think this can be deleted completely. It doesn't create a new annotation, like the comment says.

packages/lib-classifier/src/plugins/tasks/models/Task.js
```diff
@@ -21,18 +21,20 @@ const Task = types.model('Task', {
     },
 
     get isComplete () {
-      return !self.required || !!self.annotation && self.annotation.isComplete
+      const annotation = tryReference(() => self.annotation)
+      return !self.required || !!annotation?.isComplete
     }
   }))
   .actions(self => ({
     updateAnnotation (value) {
```
eatyourgreens, line 28: This isn't used by task components. It can be deleted.

packages/lib-classifier/src/plugins/tasks/models/Task.js
eatyourgreens: `updateAnnotation` was only being used to set up tests, so I've deleted it in favour of `annotation.update(value)` throughout the code.

srallen (changes requested):
Generally, this is working, but all of the drawing tool stories are broken now because it still is using the removed `updateAnnotation` in their setup.

eatyourgreens:
D'oh! I thought I'd caught all those.

eatyourgreens:
Drawing stories should be fixed now.

srallen (approved):
LGTM. I'm a bit concerned about the lack of tests for the data vis task. I believe this is partly because of the integration with the d3 viewer, so we should be sure to double check TESS (and its staging equivalent).

eatyourgreens:
It seems to me like it ought to be possible to test the data vis task models similarly to the tests for the drawing tasks: create some mock tool models, then mock the actions that would be run by the classification tools. I haven't looked closely at that task, though.

Issue #1826, opened 2020-09-28
Project: client-side page data requests (/_next/data/ URLs) broken


eatyourgreens:
**Package**
app-project

**Describe the bug**
Client-side workflow routing is broken in mobile Safari.

EDIT: the underlying cause of this bug is that Next's data request URLs are broken, because they aren't being proxied through to a NextJS app. eg. https://www.zooniverse.org/_next/data/cy3jPk61vsMFAGo-Zi51-/projects/msalmon/hms-nhs-the-nautical-health-service/classify/workflow/16926/subject-set/82738.json

**To Reproduce**
1. Go to https://frontend.preview.zooniverse.org/projects/cseidenstuecker/every-name-counts
1. Click one of the workflow links.
1. Get the generic error message: 'An unexpected error has occurred'.

![Screenshot of mobile Safari showing a blank white page saying: An unexpected error has occurred.](https://user-images.githubusercontent.com/59547/94413315-d36d7e80-0172-11eb-9210-7ec49ee5b756.png)

Loading a workflow page directly will work eg.
https://frontend.preview.zooniverse.org/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/16099

**Expected behavior**
The workflow page should load with that workflow shown in the classifier.

 - OS: iOS 13.7
 - Browser: Safari

eatyourgreens:
I think the problem might be related to the page data requests going to a 404 page in PFE.
https://frontend.preview.zooniverse.org/_next/data/_d8G2fIA7qwq8iBc9Fpnt/projects/cseidenstuecker/every-name-counts/classify/workflow/15904.json

eatyourgreens:
The problem goes away if you browse the NextJS app directly on your phone:
https://fe-project.preview.zooniverse.org/projects/cseidenstuecker/every-name-counts

The problem seems to be with proxying requests to `/_next/data/`. Those don't use [the asset prefix](https://nextjs.org/docs/api-reference/next.config.js/cdn-support-with-asset-prefix) so will be requested from the window origin: `https://frontend.preview.zooniverse.org` in this case.

camallen:
> The problem goes away if you browse the NextJS app directly on your phone:
> https://fe-project.preview.zooniverse.org/projects/cseidenstuecker/every-name-counts
> 
[...]

PR #1843, opened 2020-10-06
Add basic demo mode just for admins

srallen:

Package: lib-classifier, lib-grommet-theme

Describe your changes:
This adds the gear menu and option toggle on demo mode if you are an admin user. This can be tested with the project app and with an admin user. There will be a gear only visible to an admin user in the task navigation button area where it can be enabled:

demo-mode

Check it out by bootstrapping and running `yarn dev` with the project app and go to: http://local.zooniverse.org:3000/projects/brooke/i-fancy-cats/classify

and sign in with an admin user.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
In the future, to get to parity with PFE classifier function, this needs to be expanded to project owners and collaborators.

packages/lib-classifier/src/store/ClassificationStore.spec.js
```diff
@@ -235,5 +235,78 @@ describe('Model > ClassificationStore', function () {
...
+          drawing: {},
+          feedback: {},
+          fieldGuide: {},
+          subjectViewer: {},
+          tutorials: {},
+          workflowSteps: {},
+          userProjectPreferences: {}
+        })
+
+        sinon.stub(rootStore.classifications, 'submitClassification')
```
eatyourgreens, line 292: Maybe spy here too, just in case there is a bug in `submitClassification`, which we'll miss by not running the action.

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -236,7 +236,7 @@ const WorkflowStepStore = types
 
           const isSingleChoiceTaskNotBranching = task.type === 'single' && !isThereBranching(task)
           if (isSingleChoiceTaskNotBranching) {
-            stepSnapshot.next = task.answers[0]?.next
+            stepSnapshot.next = task.answers && task.answers[0]?.next
```
eatyourgreens, line 239: Making a note for myself that this fix can be removed from #1859.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/ExpertOptions/ExpertOptionsContainer.spec.js
```diff
@@ -0,0 +1,108 @@
...
+      wrappingComponent: Grommet,
+      wrappingComponentProps: { theme }
+    })
+
+    after(function () {
+      useContextMock.restore()
+    })
+
+    it('should not render ExpertOptions', function () {
+      console.log(wrapper.debug())
```
eatyourgreens, line 86: Stray console log.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/ExpertOptions/ExpertOptionsContainer.spec.js
```diff
@@ -0,0 +1,108 @@
...
+      expect(wrapper.find(ExpertOptions)).to.have.lengthOf(0)
+    })
+  })
+
+  xdescribe('when there is an admin user', function () {
```
eatyourgreens, line 91: Does this test still need to be fixed?

eatyourgreens (approved):
Looks good to me. I have minor questions but nothing blocking.

Logged in, I was able to toggle demo mode with the trackpad and stop classifications from being submitted. I could also operate the toggle from the keyboard, but it took some figuring out and should be improved in a future PR. For keyboard users, I'd recommend cutting down on the number of keystrokes retired to navigate to the toggle and operate it. Ideally, after opening the toggle, I think a single keystroke should operate the checkbox. At the moment, you have to tab to it before you can operate it. Maybe autofocus the checkbox after the popup opens?

After the popup is closed (with Esc), focus is lost and it takes a while to figure out where it's gone. I'd recommend returning focus to the cog button after the toggle is closed from the keyboard. All this was tested from the keyboard in Firefox 82.

Screenshot of the dev classifier showing the deno mode toggle next to the Done button for I Fancy Cats.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/ExpertOptions/ExpertOptionsContainer.js
```diff
@@ -0,0 +1,43 @@
...
+  useEffect(() => {
+    authClient.checkCurrent().then((user) => {
+      if (user) {
+        setIsAdmin(user.admin)
+      } else {
+        setIsAdmin(false)
+      }
+    }).catch((error) => { console.error(error) })
+
+    return setIsAdmin(false)
```
eatyourgreens, line 29: Sorry, missed this on my first pass. I think useEffect is supposed to return a callback function.
```suggestion
    return () => setIsAdmin(false)
```

eatyourgreens: My suggestion could also be a named function, to make debugging easier.
```js
return function onUnmount() {
  setIsAdmin(false)
}
```

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/ExpertOptions/ExpertOptionsContainer.spec.js
eatyourgreens: For reference, this test is skipped because it throws an error inside the `DropButton` that's rendered for the toggle.

```
TypeError: Cannot read property 'length' of undefined
    at phash (/Users/odonnell/zooniverse/front-end-monorepo/node_modules/styled-components/dist/styled-components.cjs.js:683:13)
    at ComponentStyle.generateAndInjectStyles (/Users/odonnell/zooniverse/front-end-monorepo/node_modules/styled-components/dist/styled-components.cjs.js:1271:25)
    at useInjectedStyle (/Users/odonnell/zooniverse/front-end-monorepo/node_modules/styled-components/dist/styled-components.cjs.js:1497:120)
    at useStyledComponentImpl (/Users/odonnell/zooniverse/front-end-monorepo/node_modules/styled-components/dist/styled-components.cjs.js:1525:28)
    at StyledButton (/Users/odonnell/zooniverse/front-end-monorepo/node_modules/styled-components/dist/styled-components.cjs.js:1592:12)
```

My best guess at what might be wrong here is that `styled-components` uses context, but this test is mocking `React.useContext`.

eatyourgreens: One way to get around this would be to add an error boundary around the `DropButton`. It's not ideal, but I think it would allow you to test the container.

packages/lib-classifier/src/store/ClassificationStore.spec.js
```diff
@@ -235,5 +235,78 @@ describe('Model > ClassificationStore', function () {
...
+        taskSnapshot.createAnnotation = () => SingleChoiceAnnotation.create(singleChoiceAnnotationSnapshot)
+        classifications.addAnnotation(taskSnapshot, singleChoiceAnnotationSnapshot.value)
+        classifications.completeClassification({
+          preventDefault: sinon.stub()
+        })
+      })
+
+      it('should not call submitClassification', function () {
+        expect(classifications.submitClassification).to.have.not.been.called()
+      })
```
eatyourgreens, line 273: With demo mode enabled, we still expect the old classification to complete and a new classification to be created. Is it possible to test that `classifications.active` changes?

eatyourgreens:
I'm seeing the current user checked on every step of I Fancy Cats, because the Expert Options component is unmounted and mounted on each change to the task navigation. That might cause us performance problems on popular projects. The user could be stored (or passed down from the project as a prop) to save on redundant API calls.

Screenshot of the dev console for I Fancy Cats, showing three API calls to check the current user during a single classification.

eatyourgreens:
That API performance problem might need to be fixed before this can be used in production, so maybe hold off on merging until we know for certain whether it's a major issue?

srallen:
There's enough comments here that any follow up should necessitate another review, so I'm going to dismiss the approval so this doesn't accidentally merge.

srallen:
I came across a note I left myself about implementing this:

> We could implement this as a URL query param rather than linked to authenticated user roles

Which is still a good idea. There isn't exactly a reason that this can't be available to anyone as long as it's clearly communicated in the UI when it's enabled. So would a refactor to make the gear and toggle available in the UI by query param in the URL be welcome? Then I can eliminate the request to check for admin user (and the later requirement to check for user roles)

eatyourgreens:
The query param sounds like a good idea to me. In Next 9.3 and up, you can get query params from the request in `getServerSideProps()`. #1826 is blocking us from using that in production at the moment, because it will make requests to `https://www.zooniverse.org/_next/data/`.

beckyrother:
Small design comment that the flyout in dark mode needs to be [darker gray](https://projects.invisionapp.com/dsm/zooniverse/primary-brand/asset/colors/5bfd7a8a86b10b0011547f68) to stand out from the rest of the page.

srallen:
Closing this in favor of #1909. I think I addressed all the comments that still applied.

PR #1859, opened 2020-10-21
Classifier: add a generic Task component

eatyourgreens:
Move task rendering out of the Tasks component and into a generic Task component. On mount, Task is responsible for getting an annotation from the store, passing it to the current task and then rendering the appropriate task component, depending on task.type. Adding/creating an annotation should only happen once. At the moment, `classification.addAnnotation(task)` is called on every render, which is leading to problems in the store when tasks try to access a stale annotation.

Key the Tasks area by classification.id so that tasks unmount at the end of a classification.

Wrap exported Task components in `observer`, so that `observer()` isn't running (and generating a new component) on each render of a task.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
```diff
@@ -48,30 +47,22 @@ class Tasks extends React.Component {
...
+          key={classification.id}
+          as='form'
+          gap='small'
+          justify='between'
+          fill
+        >
+          {step.tasks.map((task) => (
+            <Task
+              key={task.taskKey}
+              {...this.props}
```
eatyourgreens, line 60: Should I remove this. I don't think all the task area props need to be passed down to every task, but I'm not 100% sure which props the PH-TESS task needs.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
```diff
@@ -0,0 +1,46 @@
...
+export default function Task (props) {
+  const { classification, disabled, task } = props
+  const [ annotation, setAnnotation ] = useState()
+  const { TaskComponent } = taskRegistry.get(task.type)
+
+  useEffect(function onMount() {
+    const annotation = classification.addAnnotation(task)
+    task.setAnnotation(annotation)
+    setAnnotation(annotation)
+  }, [])
```
srallen, line 16: This makes me nervous to have the annotation stored in the stores and in local component state. It seems risky as there can be potential of getting out of sync with each other. From testing, this works, but I really would like us to rethink how annotations are being stored with an aim of simplification.

srallen (commented):
I'm generally supportive of this refactor, but as noted in line, the storing of the annotations makes me nervous. I'd like to have some more time to think about this before giving an explicit approval.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
eatyourgreens: I've got another commit, which I've pushed up as #1870, which replaces `useEffect` and `useState` with a `useStores` hook, so the `Task` component becomes a generic container which takes a `task` prop, connects to the store and passes `annotation` and `disabled` props down to a specific task.

I haven't tested that branch in depth, but the disadvantage of that might be that the `addAnnotation` action gets called more than once during rendering.

eatyourgreens:
The drawing subtask popup uses the same generic code as `Tasks` to get task annotations from the model. I think that's going to need rewriting too, so that it isn't calling actions during `render()`.
https://github.com/zooniverse/front-end-monorepo/blob/c4f819c2269a9a3937d2f37f447778e593bf635a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js#L78-L87

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
eatyourgreens: I agree. I think the data-vis task only needs a `task` prop, but I'm not sure.

srallen:
I've reviewed #1870 first before looking at this one again since it addresses one of my comments.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
eatyourgreens: #1870 doesn’t fix the issue this PR is addressing (actions being called during render.) It refractors the actions into a function which is called on every render, but `useStores` isn’t a hook. After reading the [docs for custom hooks](https://reactjs.org/docs/hooks-custom.html), I think `useStores` still needs to use `useEffect` and `useState`.

eatyourgreens: `useStores` here should now be a hook which only runs `classification.addAnnotation(task)` on component mount.

eatyourgreens:
I've changed `AnnotationsStore` so that `classification.annotation(task)` and `mark.annotation(task)` are pure views (and can be undefined). Creating an annotation still happens as part of `mark.addAnnotation(task)` and `classification.addAnnotation(task)`, but I've split it out into its own function.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
eatyourgreens: What we'd really like, I think, is to add/create an annotation on mount (if one doesn't already exist), then access `classification.annotation(task)`. `classification.annotation(task)` is a pure view now, so it shouldn't matter how many times we call it (I think.)

eatyourgreens: You can [set references from `useEffect`](https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables), so I've experimented with using `useRef` instead of `useState`. However, that breaks I Fancy Cats. I think that's happening because changing a ref doesn't result in a new render.

srallen: I don't understand why you changed again what was in #1870. This now still has the same issue that I had concerns about before, which was the fact we're storing the annotation in component state and in the MST state. I don't think we need a custom hook or local component state here. It functioned as it was proposed and accepted in #1870 with `useContext` which triggers the rerender as needed: https://reactjs.org/docs/hooks-reference.html#usecontext

eatyourgreens: Without the effect hook, `classification.addAnnotation(task)` is called on every render, which leads to the bug where the store throws errors like this:

```
Uncaught Error: [mobx-state-tree] You are trying to read or write to an object that is 
no longer part of a state tree. 
(Object type: 'TextAnnotation', 
Path upon death: '/classifications/resources/ckgullo6r00002a6anqcvzwmy/annotations/ckgullodx00012a6ae99dqhcj', 
Subpath: 'id', Action: ''). 
Either detach nodes first, or don't use objects after removing / replacing them in the tree.
```

I'm trying to mitigate that here by making sure that `addAnnotation` is only called once for a task.

eatyourgreens: There might be a simpler solution, now that `classification.annotation(task)` doesn't create new annotations. :thinking:

eatyourgreens: I think I'm happier with the code now. It checks the `classification.annotation(task)` on every context change, but only runs `classification.addAnnotation(task)` if there's no annotation already.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
eatyourgreens: The downside of this, I think, is that `useStores`, in the Task component, will run every time one of these extra props changes. That might be why I've been seeing bugs related to the `classification.addAnnotation(task)` action running when it isn't needed.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
srallen: I didn't see those errors in #1870 while testing, but conceptually it makes sense that it could happen. I see that you've updated the branch to remove the local state, so thanks for doing that. I can understand the desire to get `useContext` to run once since that seems like what we want, however I think the mix of `useContext` and `useState`/`useEffect` for the same state seems potentially error prone. `useContext` is intended to rerun if the context changes and the way the `useStores` function is written, it should rerun if either the store or task prop changes. I like the explicit check for the existence of `annotation` now and makes it more clear to me what the intention is: only call `addAnnotation` if `annotation` doesn't exist. I think what we have now is preferred.

FWIW, it's my understanding that the empty dependency array of useEffect is to be avoided... it's often what people have recommended as the `componentDidMount` "replacement" but apparently that can be error prone. I've found these two articles helpful to understand this:

https://reacttraining.com/blog/useEffect-is-not-the-new-componentDidMount/
https://medium.com/better-programming/understanding-the-useeffect-dependency-array-2913da504c44

I find it helpful to remember the array is our way of telling React what the effect in useEffect is dependent on and an empty array should only be used when the effect truly _depends on nothing_. So suppose we had a scenario where useContext wasn't being used, but we wanted something to run once, I _think_ the preferred method is this over an empty dependency array:
```js
const [state, setComponentState] = useState(null)
useEffect(function myEffect () {
  const newState = doSomething()
  setComponentState(newState)
}, [setComponentState])
```
So I think it will only run once upon mount and clearly declares the function as a dependency in the effect. We haven't really been doing this, but the second article helped clarify for me that you can actually put functions in the dependency array. (I definitely could still be wrong about this, as I'm still wrapping my brain around it. It was easy to just grab the simple solution of using empty arrays in previous solutions with `useEffect`, but I'm going to try to avoid it going forward).

eatyourgreens: Thanks! I didn’t know that about `useEffect` either. I will give those articles a read tomorrow. 


I think the acid test for this PR would be: does it fix the focus bug that stops you from typing into text tasks? IIRC that was being caused by the observers.

srallen (commented):
This is looking great! I did a final manual check and the single, multiple, text, drawing, transcription (plus sub-task), and data vis task function as expected with setting their default annotations. I don't know of good test projects for the simple dropdown and subject group, so I'd like confirmation that those are working before giving explicit approval. @shaunanoordin or @eatyourgreens do you know of any? Thanks!

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
srallen: I'm not seeing the focus bug, but would be good to maybe try out each other's test projects. Which project do you use to test the text task?

eatyourgreens:
Try this for a dropdown task. https://localhost:8080/?project=908&workflow=3339

packages/lib-classifier/src/store/AnnotationsStore.js
```diff
@@ -21,10 +21,18 @@ const AnnotationsStore = types
...
+      const taskAnnotation = task.createAnnotation()
+      self.annotations.put(taskAnnotation)
+      return taskAnnotation
+    }
+
     function addAnnotation (task, value) {
-      const annotation = self.annotation(task)
-      // new annotations must be added to this store before we can modify them
-      self.annotations.put(annotation)
+      let annotation = self.annotation(task)
```
eatyourgreens, line 31: This line can be removed at some point, once we’re sure that `addAnnotation`is only called to add a new annotation. The TESS light curve viewer might still be using the old logic.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
eatyourgreens: Mark mentioned it in a PR, but I forget which one. It shows up if you have a text task in the main task area, not in a popup. Every character you type, the task component unmounts and mounts again, which stops you from typing more than one letter.

eatyourgreens: https://github.com/zooniverse/front-end-monorepo/pull/1868#issuecomment-716817616

srallen: Confirmed that there is no focus bug with https://localhost:8080/?project=1693&workflow=3430

srallen:
I don't see a dropdown task rendered with https://localhost:8080/?project=908&workflow=3339. I don't think it has the right number of options for the constraints we've defined.

eatyourgreens:
I've added an extra option to that dropdown, so it should work now.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
eatyourgreens: By the way, referential integrity, from that second post, is the reason to avoid binding handlers with `onClick={handleClick.bind(this)}` in components. Every time the component renders, a new copy of `onClick` (new reference) is created.

eatyourgreens: The closure in the first post reminded me of this example (updated to use JSX.) I remember asking what this code does being a JS interview question, before `let` existed:
```jsx
const items = []
for (var i = 1; i < 4; i++) {
  items.push(<button onClick={() => alert(i)}>{i}</button>)
}
return(<>{items}</>)
```

srallen (approved):
Simple dropdown task looks good. I'm still not sure how to test subject group, but that can be addressed separately if needed. I don't have any reason to think it would have an issue since specs are passing.

PR #1861, opened 2020-10-22
Classifier: validate simple dropdown options

eatyourgreens:
Add a minimum and maximum options count to the simple dropdown task. Add a MenuOptions type which validates the task options against those counts.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
If task creation fails, because the number of options doesn't match the model, then the `WorkflowStepStore` model will error here:
https://github.com/zooniverse/front-end-monorepo/blob/30a8516cb0d0d5b827da00b6b011e058f8f611b6/packages/lib-classifier/src/store/WorkflowStepStore.js#L166-L171

That error handler could be modified to store the error message and display it in the classifier, if we want to make the error visible outside of the dev console. The MST error in the console is a little opaque, but should tell you that the supplied `options` snapshot is not a valid `MenuOptions` type.

eatyourgreens:
Try it out on this workflow, which is a dropdown task with three options: https://localhost:8080/?project=908&workflow=3339

```
Error: [mobx-state-tree] Error while converting `{"help":"","type":"dropdown","selects":[{"id":"c0a8efe878208","title":"Main Dropdown","options":{"*":[{"label":"one","value":"3a5e7c30a1f81"},{"label":"two","value":"ecc0e62c6c11"},{"label":"three","value":"57ca0302a9e4d"}]},"required":true,"allowCreate":false}],"instruc......":"T0"}` to `(DataVisAnnotationTask | DrawingTask | MultipleChoiceTask | SimpleDropdownTask | SingleChoiceTask | TextTask | TranscriptionTask)`:

    at path "/options" snapshot `["one","two","three"]` is not assignable to type: `MenuOptions`
Value does not respect the refinement predicate, expected an instance of `MenuOptions` or a snapshot like `MenuOptions` instead.
```

srallen (approved):
LGTM, but it'd be nice to have a display of an error in the UI. The test project https://localhost:8080/?project=908&workflow=3339 has a dropdown with only 3 options, so no longer meets the requirements and displays a blank task area:

Screen Shot 2020-10-28 at 11 19 01 AM

Perhaps follow up item?

srallen:
Apologies! Just saw that you noted how the error handling works already, but happy to have that be in a follow up PR. Thanks!

eatyourgreens:
Yeah, I think we can modify the step tasks model so that any invalid task shows a message in the task area. Maybe have a `TaskError` model that gets pushed into `step.tasks` or something like that.

PR #1870, opened 2020-10-29
Classifier: generic Task connector

eatyourgreens:
Connect up the Task component so that it gets its annotation and disabled state from the store. Replace useContext and useState with a useStores hook in the Task component. Add an optional store prop for passing stores in during tests.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
@@ -1,19 +1,33 @@
+import asyncStates from '@zooniverse/async-states'
 import { Box, Paragraph } from 'grommet'
-import { bool, func, shape } from 'prop-types'
-import React, { useEffect, useState } from 'react'
+import { MobXProviderContext, observer } from 'mobx-react'
+import { func, shape, string } from 'prop-types'
+import React, { useContext } from 'react'
 
 import taskRegistry from '@plugins/tasks'
 
-export default function Task (props) {
-  const { classification, disabled, task } = props
-  const [ annotation, setAnnotation ] = useState()
-  const { TaskComponent } = taskRegistry.get(task.type)
+function useStores(task, stores) {
+  const { classifierStore } = stores || useContext(MobXProviderContext)
 
-  useEffect(function onMount() {
-    const annotation = classification.addAnnotation(task)
-    task.setAnnotation(annotation)
-    setAnnotation(annotation)
-  }, [])
+  const {
+    classifications,
+    subjectViewer
+  } = classifierStore
+  const { loadingState: subjectReadyState } = subjectViewer
+  const classification = classifications.active
+  const disabled = subjectReadyState !== asyncStates.success
+  const annotation = classification.addAnnotation(task)
+  task.setAnnotation(annotation)
+  return {
+    annotation,
+    disabled
+  }
+}
+
+function Task (props) {
+  const { store, task } = props
+  const { annotation, disabled } = useStores(task, store)
+  const { TaskComponent } = taskRegistry.get(task.type)
 
   if (annotation && TaskComponent) {
     return (
@@ -32,15 +46,11 @@ export default function Task (props) {
 }
 
 Task.propTypes = {
-  classification: shape({
-    addAnnotation: func.isRequired
-  }).isRequired,
-  disabled: bool,
   task: shape({
-    setAnnotation: func.isRequired
+    setAnnotation: func.isRequired,
+    taskKey: string.isRequired
   }).isRequired
 }
 
-Task.defaultProps = {
-  disabled: false
-}
+export default observer(Task)
+
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
@@ -3,7 +3,6 @@ import { Box, Paragraph } from 'grommet'
 import { inject, observer } from 'mobx-react'
 import PropTypes from 'prop-types'
 import React from 'react'
-import { withTheme } from 'styled-components'
 
 import Task from './components/Task'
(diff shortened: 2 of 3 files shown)


eatyourgreens:
https://localhost:8080/?project=1693&workflow=3430
On filling out the text box and pressing Done, I'm seeing the classification submitted then the classifier crashes with
```
Uncaught Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree. (Object type: 'TextAnnotation', Path upon death: '/classifications/resources/ckgullo6r00002a6anqcvzwmy/annotations/ckgullodx00012a6ae99dqhcj', Subpath: 'id', Action: ''). Either detach nodes first, or don't use objects after removing / replacing them in the tree.
```

eatyourgreens:
Restoring `key={classification.id}` to the top-level container fixes that crash. :thinking:

eatyourgreens:
After reading the custom hooks example, maybe the way to go here is to wrap `useEffect` and `useState` inside the `useStores` hook, rather than removing them?
https://reactjs.org/docs/hooks-custom.html#extracting-a-custom-hook

The bug above is happening because `useStores(task)` runs on every render, even when the new subject is still loading. We only want to add a new annotation to the task when the subject is loaded and the new classification is ready, and we only want to do that once.

Ditto for drawn marks: we'd want to add a new annotation to a mark once, when the mark is created. However, a single drawing tool subtask can have multiple annotations (one for each drawn mark.) So, the Task component needs to get the correct annotation from the store, and add it to the rendered task, each time it mounts for a given classification or mark.

eatyourgreens:
In the store, we've got `classifications.addAnnotation`, which is used by the TESS light curve viewer:
https://github.com/zooniverse/front-end-monorepo/blob/c4f819c2269a9a3937d2f37f447778e593bf635a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js#L21-L24

It's a wrapper for `classifications.active.addAnnotaton(task, value)`:
https://github.com/zooniverse/front-end-monorepo/blob/c4f819c2269a9a3937d2f37f447778e593bf635a/packages/lib-classifier/src/store/ClassificationStore.js#L86-L101

`addAnnotation` for individual classifications and marks is used elsewhere to manage task annotations:
https://github.com/zooniverse/front-end-monorepo/blob/c4f819c2269a9a3937d2f37f447778e593bf635a/packages/lib-classifier/src/store/AnnotationsStore.js#L24-L32

`addAnnotation` is a confusing name because that action doesn't add a new annotation to the store. If there's an existing annotation for a task, that is returned instead. One improvement would be to split up creating a new task annotation and retrieving a task annotation into separate functions. Creation only needs to happen once for a given classification or mark, but retrieval can happen multiple times as we work through a workflow's tasks, opening and closing any drawing task popups to edit subtask annotations, or going back to edit annotations from previous steps. I think `addAnnotation` should always add a new annotation for a task, and should only be called if there's no existing annotation for the current task and classification/mark. 

Question: how do we handle recursive workflows, where a classification can store multiple annotations for the same task key? On the second pass through the workflow, we'll want to add a new annotation even though there's an existing annotation for each task key.

eatyourgreens:
After stopping and restarting the dev classifier, on my laptop, I can't trigger that MST bug any more on this branch. 🤔

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
```diff
@@ -1,19 +1,33 @@
...
+  task.setAnnotation(annotation)
+  return {
+    annotation,
+    disabled
+  }
+}
+
+function Task (props) {
+  const { store, task } = props
+  const { annotation, disabled } = useStores(task, store)
```
srallen, line 29: Non-blocking suggestion: I think we can still have a default here for disabled by:
```
const { annotation, disabled = false } = useStores(task, store)
```

srallen (approved):
LGTM. I checked the staging kitty project to make sure the default annotation is working for single, multiple, and drawing task; ASM staging for transcription task (which also is a subtask check, but as noted that hasn't been refactored yet); and TESS staging for data vis task. I'm not sure what is a good test project for the simple dropdown and the subject group and I missed those when checking for the id refactor. @shaunanoordin could you manually check those two tasks types to make sure the default annotation is working as expected with this refactor? Specs are passing, so I'd expect the manual check to be fine.

srallen:
> how do we handle recursive workflows, where a classification can store multiple annotations for the same task key? On the second pass through the workflow, we'll want to add a new annotation even though there's an existing annotation for each task key.

Unless I'm forgetting something, the annotations are keyed and stored by the generated id by cuid, so I think it would work?

srallen:
> addAnnotation is a confusing name because that action doesn't add a new annotation to the store. If there's an existing annotation for a task, that is returned instead. One improvement would be to split up creating a new task annotation and retrieving a task annotation into separate functions. Creation only needs to happen once for a given classification or mark, but retrieval can happen multiple times as we work through a workflow's tasks, opening and closing any drawing task popups to edit subtask annotations, or going back to edit annotations from previous steps. I think addAnnotation should always add a new annotation for a task, and should only be called if there's no existing annotation for the current task and classification/mark.

100% supportive of this suggestion. The naming is what tripped me up while debugging several weeks ago.

eatyourgreens:
> Unless I'm forgetting something, the annotations are keyed and stored by the generated id by cuid, so I think it would work?
 
On the second pass through a workflow, `classification.addAnnotation(taskKey)` will get the existing annotation for `taskKey`, so you’ll end up editing your first answer again.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
eatyourgreens: Looking through `useStores`, `disabled` can’t be undefined.

PR #1887, opened 2020-11-06
Remove the multistage Docker build

eatyourgreens:
Revert the [multi-stage Docker build](https://docs.docker.com/develop/develop-images/multistage-build/) from #1109 and go back to building the NextJS apps in the monorepo bootstrap script. 

With this PR, the NextJS apps are built in `zooniverse/front-end-monorepo-staging` and `zooniverse/front-end-monorepo-production` images and start commands are [defined in the kubernetes templates](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/), as suggested in https://github.com/zooniverse/front-end-monorepo/pull/1528#issuecomment-588971800

I've consolidated the build environment into one set of env vars, and removed the individual Docker images for `fe-project` and `fe-content-pages`. In turn, that meant that the docker-compose production templates for the apps could be removed too.

I've also updated the Jenkinsfile so that a docker build is run on every push. Currently, we only build on merges to master. I think this change will catch any build errors in the NextJS apps before they're deployed to staging.

Working on this, I noticed that the local production setup for the Next apps is using a development build against the staging API. That won't work with a production build of the classifier library, so I've changed the local production environment to use production.
https://github.com/zooniverse/front-end-monorepo/blob/50a9ceeccb037b14ce761ac9110d236e6e0302bc/packages/app-project/docker-compose-prod.yml#L12-L13

Package:
app-content-pages
app-project
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
Ideally, I think we'd like to be able to build NextJS apps in isolation, without deploying the entire monorepo (source code plus tests etc.) for each app.

Something like this maybe, where an app is built in the monorep image but only the built app is copied to the deployed image?

```docker
FROM zooniverse/front-end-monorepo:latest AS bootstrap

FROM node:12

RUN mkdir -p /usr/src
WORKDIR /usr/src/

COPY --from=bootstrap /usr/src/packages/app-project /usr/src

ENV PORT 3000
EXPOSE 3000

CMD [ "yarn", "start" ]
```

At the moment, that would fail for a couple of reasons:

- `git rev-parse HEAD` will fail here because `packages/app-project` isn't a git repo.
https://github.com/zooniverse/front-end-monorepo/blob/c4f819c2269a9a3937d2f37f447778e593bf635a/packages/app-project/next.config.js#L14-L16
- module resolution will fail for modules like `@zooniverse/grommet-theme`.

camallen:
> it rev-parse HEAD will fail here because packages/app-project isn't a git repo.

we can pass in the git revision via the jenkins build script as an env var
https://github.com/zooniverse/designator/blob/e7583b0e55faad24604b022ff7a103fa1428ee1d/Jenkinsfile#L18
https://github.com/zooniverse/designator/blob/e7583b0e55faad24604b022ff7a103fa1428ee1d/Dockerfile#L14

Perhaps for local images, have a fallback to the shell cmd vs the env var lookup?

eatyourgreens:
Travis CI is failing because it doesn't have the Contentful secrets for the content pages build.

camallen:
> > it rev-parse HEAD will fail here because packages/app-project isn't a git repo.
> 
> we can pass in the git revision via the jenkins build script as an env var
> https://github.com/zooniverse/designator/blob/e7583b0e55faad24604b022ff7a103fa1428ee1d/Jenkinsfile#L18
> https://github.com/zooniverse/designator/blob/e7583b0e55faad24604b022ff7a103fa1428ee1d/Dockerfile#L14
> 
> Perhaps for local images, have a fallback to the shell cmd vs the env var lookup?

like you've already done! 👍 

re tavis secrets, travis CI has changed their plans and have been very slow to run CI tooling. Perhaps we could switch to github actions vs travis for this? 

OOI - why does the test runner need real secrets? does it pull real data from contentful for the CI testing?

eatyourgreens:
>we can pass in the git revision via the jenkins build script as an env var

The Jenkins build is passing COMMIT_ID to the Docker build, so that should work. It turns out I was seeing local errors because COMMIT_ID is missing from the `docker-compose` environment.

eatyourgreens:
>OOI - why does the test runner need real secrets? does it pull real data from contentful for the CI testing?

The tests don't need it, but this is in `travis.yml`. I think it's to dry-run a build in case there are errors. With the changes in this PR, `yarn bootstrap` will build the Team and Publication pages but maybe we can work around that?
```yml
install:
  - yarn bootstrap

script:
  - npm run test:ci

after_success:
  - yarn coverage
```

README.md
```diff
@@ -137,9 +137,8 @@ The yarn build scripts default to production for libraries if `PANOPTES_ENV` is
 
 ### Docker images
 
-- `zooniverse/front-end-monorepo`: Built from the Dockerfile in the root directory. It runs `yarn install` and `yarn build` for the `lib-` packages.
-- `zooniverse/fe-content-pages`: Built from the Dockerfile in `packages/app-content-pages`. Runs `yarn build` and `yarn start` in `packages/app-content-pages` from `zooniverse/front-end-monorepo:latest`.
-- `zooniverse/fe-project`: Built from the Dockerfile in `packages/app-project`. Runs `yarn build` and `yarn start` in `packages/app-project` from `zooniverse/front-end-monorepo:latest`.
+- `zooniverse/front-end-monorepo-staging`: Built from the Dockerfile in the root directory. It runs `yarn install` and builds all the libraries and apps from the latest main branch commit.
+- `zooniverse/front-end-monorepo-production`: Built from the Dockerfile in the root directory. It runs `yarn install` and builds all the libraries and apps from the `production_release` tag.
```
eatyourgreens, line 141: Do we need two images, now that the monorepo builds are tagged with the commit hash? Can we use `zooniverse/front-end-monorepo:${GIT_COMMIT}` instead when we deploy to staging and to production?

Dockerfile
```diff
@@ -10,15 +10,38 @@ USER node
...
-RUN mkdir -p /usr/src
-WORKDIR /usr/src/
+ARG COMMIT_ID
+ENV COMMIT_ID=$COMMIT_ID
 
-COPY --from=bootstrap /usr/src /usr/src
+ENV PANOPTES_ENV production
 
-RUN yarn workspace @zooniverse/fe-project build
+ENV NODE_ENV production
```
eatyourgreens, line 22: Does anyone know why `NODE_ENV` is explicitly set for the app builds?
https://github.com/zooniverse/front-end-monorepo/blob/6b73ad555eb98799283f28950e4517d74d029b57/packages/app-project/Dockerfile#L15-L17

eatyourgreens: Previously, this was set in the Jenkins build environment (see #1494) but `yarn install` ignores dev dependencies during a production build, so that setup would break the library builds (there'd be no webpack installed, for a start.)
https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-production-true-false

Setting `NODE_ENV` for the image in the Dockerfile builds production builds of the NextJS apps, but clobbers any tests that use `NODE_ENV=test`. You can try that out on this branch with `docker-compose run --rm bash`, then run the tests in the built image with `yarn test`. That might not be a big deal for us, because we don't run our CI tests in the production Docker images.

README.md
eatyourgreens: I have a vague memory that we switched from deploying `fe-project:${GIT_COMMIT}` to deploying `fe-project-staging:${GIT_COMMIT}` and `fe-project-production:${GIT_COMMIT}` because of a problem where requests for JS files 404'ed in production, but I don't remember the details. Maybe having separate images for staging and production deploys is safer?

docker-compose.yml
```diff
@@ -5,30 +5,41 @@ services:
...
     entrypoint:
       - "yarn"
       - "workspace"
       - "@zooniverse/fe-project"
     command: ["start"]
+    environment:
+      - PORT=3000
+      - ASSET_PREFIX=http://localhost:3000
+      - NODE_ENV=production
+      - PANOPTES_ENV=production
```
eatyourgreens, line 23: I think the Next runtime environment should match the environment that was used to build the classifier and apps, so I've changed these to both use `production`.

README.md
eatyourgreens: I remember now: a staging build and a production build from the same commit ID will build different JS bundles. If we only use commit IDs to identify builds, we can run into a situation where we're accidentally trying to use a staging build with the production HTML, resulting in 404 errors for the JS files.

Dockerfile
eatyourgreens: Maybe `NODE_ENV` should be a build argument?

docker-compose.yml
eatyourgreens: The current values are here. https://github.com/zooniverse/front-end-monorepo/blob/50a9ceeccb037b14ce761ac9110d236e6e0302bc/packages/app-project/docker-compose-prod.yml#L12-L13

camallen: the build env could be modified by injecting these ENV values as build args via compose definition. So we could have a development build for local dev image and the runtime envs would match this 

e.g. Caesar does this via dockerfile args and docker-compose build args https://github.com/zooniverse/caesar/blob/cb14a81b53f43b37025b0b109e5bd7c909402ea9/docker-compose.yml#L14-L17

docker-compose.yml
```diff
@@ -5,30 +5,41 @@ services:
...
       context: ./
-      target: production-apps
     entrypoint:
       - "yarn"
       - "workspace"
       - "@zooniverse/fe-project"
     command: ["start"]
+    environment:
+      - PORT=3000
+      - ASSET_PREFIX=http://localhost:3000
```
camallen, line 21: does this need to be renamed to match the project / content-page apps context?

docker-compose.yml
```diff
@@ -5,30 +5,41 @@ services:
...
       context: ./
-      target: production-apps
     entrypoint:
       - "yarn"
       - "workspace"
       - "@zooniverse/fe-content-pages"
     command: ["start"]
+    environment:
+      - PORT=3000
+      - ASSET_PREFIX=http://localhost:3001
```
camallen, line 39: has this env var been renamed?

packages/app-content-pages/README.md
```diff
@@ -25,11 +25,8 @@ To get the credentials, go to https://app.contentful.com/spaces/jt90kyhvp0qv/api
 
 - `docker-compose up` to run a server on http://localhost:3000.
 - `docker-compose down` to stop the dev server.
-- `docker-compose run --rm dev test` to run the tests.
-
-##### Preview the production Dockerfile image build
-
-- `docker-compose -f docker-compose-prod.yml up` to run a production Dockerfile image server on http://localhost:3000.
+- `docker-compose run --rm content-pages test` to run the tests.
+- `docker-compose --rm content-pages start` to run a production build on http://localhost:3000.
```
camallen, line 29: not blocking at all - I guess the production version will depend on the final setup of the env vars there could be another service in the compose file (or compose file) for explicit prod / dev builds.

camallen (commented):
I haven't tested this thoroughly, rather a first pass review. 

My understanding is this PR will optimize the split builds into one image build with both apps and therefore avoid the multistage image builds via initial bootstrap image and then each app image build. That sounds good.

I've left some comments about the image build env, my main question is if we should be injecting the env at local docker build time via compose build args - thus having a non-production builds for local dev envs.

I'll happily take another look and test this out locally on another review, please re-request one after going over these comments / questions.

Dockerfile
```diff
@@ -10,15 +10,38 @@ USER node
 
 RUN yarn install
 
-RUN yarn bootstrap
+RUN yarn workspace @zooniverse/react-components build
 
-FROM node:12 AS production-apps
+RUN yarn workspace @zooniverse/classifier build
```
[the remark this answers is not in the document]
eatyourgreens: Setting those envs will actually break these builds. That's one of the questions I have about how our build process works. Why do we bootstrap the monorepo in one environment, then deploy the NExt apps in a different environment?
Shouldn't we build and run everything in the same environment. I could see everything failing, for example, if we built for staging/development but ran in a production environment, mostly because the classifier build wouldn't match the runtime environment.

docker-compose.yml
eatyourgreens: Not unless we add a proxy to the local setup. I don't think anyone is using different hostnames for local work (couldbe wrong.)

docker-compose.yml
eatyourgreens: Yes! good catch.

docker-compose.yml
eatyourgreens: Oh wait, you mean this should be `PROJECT_ASSET_PREFIX` now. Yep.

I was thinking about the hostnames for projects vs. content pages, which are different in our production deploys

Dockerfile
eatyourgreens: Do you remember why you hardcoded them in #1494? I think it should be okay to set them via injected args, but we'd have to add  `--production=false` to the install step.

README.md
eatyourgreens: Should befine. Do we still want a build that pulls the production image from Docker Hub and runs it locally? Local `docker-compose` will build and run from the latest local HEAD commit.

camallen: we don't have these for any other apps and can add a custom docker-compose file / manually set it up if needed.

Dockerfile
camallen: hard coded is the default without the injected args but allows for them to be overriden at build time, https://github.com/zooniverse/front-end-monorepo/pull/1494/files#r377658763

camallen (commented):
I think by default (via jenkins CI) this should build production env images

via docker-compose files we should inject the development env values and build non-production images

I think that will allow us to have the best of both worlds, i may be misunderstanding the specific issues around the env specification for the `yarn install` issues etc. 

Let me know if the env injection / specification at build time sounds unfeasible.

eatyourgreens:
>I think that will allow us to have the best of both worlds, i may be misunderstanding the specific issues around the env specification for the yarn install issues etc.

The build issue is a slightly odd one. I figured it out by removing `yarn bootstrap` from the Dockerfile and explicitly listing the build commands, setting the environment at each stage.

You can reproduce the build problem locally by running `NODE_ENV=production yarn bootstrap` and watch the library builds fail because dev dependencies like `webpack` aren't installed. At the moment, I think we're bootstrapping the monorepo in a development environment, then building in a production environment. I find it quite confusing to follow.

I think you're right that we should inject the environment at build time, then use `--production=false` to force a dev install even during production builds.

eatyourgreens:
Oh, I should add that the strange environment setup in the Dockerfile here is to make sure that the build environments here match the build environments we use on master. I thought it might be too complicated to consolidate the builds into one image here and also change the actual build steps that we run during a deploy.

docker-compose.yml
eatyourgreens: The classifier build environment is hard-coded in the webpack config.
https://github.com/zooniverse/front-end-monorepo/blob/71faab0da33610e5263d3b4e0b02b995980166d4/packages/lib-classifier/webpack.dist.js#L4-L8

So, we can easily inject a local build environment via compose, but the classifier build will ignore it and always use production. I think changing that would be a big enough change to our build process that it would merit its own PR and discussion.

If the classifier is always built in a production environment, I think that means the Next apps should be built and run in the same environment. Otherwise, I think we could get strange side-effects where SSR uses the staging API but client JS uses the production API.

I'm happy to add build args to this PR. I'm not 100% sure what the side effects of those would be, without spending some time to build and test different environments.

README.md
eatyourgreens: 👍 

For the benefit of anyone else following this conversation, with this PR `docker-compose up --build` should build and run a local image which matches the Jenkins build (except for running behind a proxy.)

If we were to add a proxy service, as part of the local compose setup, then that could catch bugs like #1826 when we run the apps locally in Docker.

packages/app-content-pages/README.md
eatyourgreens: Move the dev services up into the top-level compose file? That would keep everything in one place. I don't know how people feel about having package-level compose files vs. a single compose file for everything in the monorepo. Currently there are dev services, spread across the different packages, for content-pages, project pages, the dev classifier and all the storybooks (including Zooniverse React Components.)

I've made the assumption that if you're using Docker for an individual package, you're doing development work on that package. If you're using Docker at the top level, it's because you want to start all the services and see how they'll work together in production.

eatyourgreens:
The Panoptes client tests fail inside the built image even if you run `NODE_ENV=test yarn test:ci` in a bash shell in the image.

```
$ NODE_ENV=test mocha --reporter=min



  142 passing (1s)
  2 failing

  1) config.js
       environment from shell
         should use the value of process.env.NODE_ENV:

      AssertionError: expected 'production' to equal 'test'
      + expected - actual

      -production
      +test
      
      at Context.<anonymous> (src/config.spec.js:19:22)
      at processImmediate (internal/timers.js:439:21)

  2) config.js
       environment from shell
         should return the API host values for the test environment from the config export:

      AssertionError: expected 'https://www.zooniverse.org/api' to equal 'https://panoptes-staging.zooniverse.org/api'
      + expected - actual

      -https://www.zooniverse.org/api
      +https://panoptes-staging.zooniverse.org/api
      
      at Context.<anonymous> (src/config.spec.js:23:30)
      at processImmediate (internal/timers.js:439:21)
```

I had a look at the client config code, and that test will fail when `PANOPTES_ENV` is production.

https://github.com/zooniverse/front-end-monorepo/blob/71faab0da33610e5263d3b4e0b02b995980166d4/packages/lib-panoptes-js/src/config.js#L27-L30

eatyourgreens:
@camallen After a bit of thought, I've rearranged the Dockerfile to set the build environment before running any of the build scripts. I've added `yarn install --production=false` so that production builds of the libraries don't fail with missing build tools (like webpack.)

I think this is ready for another look. Note that the classifier webpack config will override `NODE_ENV` and `PANOPTES_ENV` from the Docker build environment, but I think changing that is outside the scope of this PR. https://github.com/zooniverse/front-end-monorepo/blob/71faab0da33610e5263d3b4e0b02b995980166d4/packages/lib-classifier/webpack.dist.js#L4-L8

Dockerfile
```diff
@@ -1,4 +1,34 @@
...
+ENV NODE_ENV=$NODE_ENV
+
+ARG APP_ENV=production
+ENV APP_ENV=$APP_ENV
+
+ENV NEXT_TELEMETRY_DISABLED=1
+
+ARG CONTENTFUL_ACCESS_TOKEN
+
+ARG CONTENTFUL_SPACE_ID
```
eatyourgreens, line 19: Should these arguments also set env vars? They don't in the current setup but I don't quite follow how the Contentful secrets are being passed to the `app-content-pages` build environment.
https://github.com/zooniverse/front-end-monorepo/blob/c0e233b955cf37948566ae92fe19fa6842b71e8a/packages/app-content-pages/Dockerfile#L9-L12

eatyourgreens: I've run a test build in Jenkins, without pushing the built image, and this seems to work fine as it is.
https://jenkins.zooniverse.org/job/Zooniverse%20GitHub/job/front-end-monorepo/job/PR-1887/24/

eatyourgreens: I've also updated the Jenkinsfile so that we run a test build as part of CI.

eatyourgreens: Jenkins is going to be slower now but that does mean we'll try a build of the static pages before changes get merged to staging.

eatyourgreens:
🎉 Jenkins is now building the NextJS apps as part of CI for PRs, rather than after we merge to staging.

Jenkinsfile
```diff
@@ -12,6 +12,9 @@
 //   5. You should get an editor where you can modify the pipeline and run it
 //      again immediately <3
 
+def dockerRepoName = 'zooniverse/front-end-monorepo-${APP_ENV}'
```
camallen, line 15: I'm not sure that APP_ENV is available at this point, it looks to be declared in the `environment` block below.

camallen (approved):
Some minor comments, main one is the APP_ENV in the jenkinsfile

i think we should get this tested on staging env and ensure it works as expected and carefully roll out to production.

Jenkinsfile
eatyourgreens: Good catch.

docker-compose.yml
eatyourgreens: I'm not sure. There's also a test, in the JS Panoptes client config, that fails when `PANOPTES_ENV=production`. There must be a reason the client is set up like that, but it's not clear if the test is wrong, or the client config is wrong. I found that when running the tests in a production image.

Jenkinsfile
eatyourgreens: The Jenkins build logs have the correct image name. I'm going to move this line anyway, because this is confusing.
```
+ docker build -t zooniverse/front-end-monorepo-staging:316fd973152192300e5f859399e0d142b19681bb --build-arg APP_ENV --build-arg COMMIT_ID --build-arg CONTENTFUL_ACCESS_TOKEN --build-arg CONTENTFUL_SPACE_ID --build-arg CONTENT_ASSET_PREFIX --build-arg SENTRY_CONTENT_DSN --build-arg PROJECT_ASSET_PREFIX --build-arg SENTRY_PROJECT_DSN .
Sending build context to Docker daemon  48.72MB
```
https://jenkins.zooniverse.org/job/Zooniverse%20GitHub/job/front-end-monorepo/job/PR-1887/28/console

eatyourgreens:
OK. Let's try this on staging.

PR #1909, opened 2020-11-19
Demo mode by query param

srallen:

Package: lib-classifier

This is a refactor of #1843 to enable the demo mode toggle by url query param rather than restricted by user role. I can't think of a reason this has to be restricted and I don't think this being the way PFE did it is a good enough reason on its own. This updates the `ExpertOptionsContainer` to check for the query param. This enables the classifier to be able to use the demo mode independent of the consuming app. Additionally, it can be enabled by prop, which sets us up to have it enabled in a different way (though some minor refactoring may have to happen once this is truly supported).

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/ExpertOptions/components/DemoModeToggle/DemoModeToggle.js
```diff
@@ -0,0 +1,29 @@
...
+import { SpacedText } from '@zooniverse/react-components'
+import counterpart from 'counterpart';
+import en from './locales/en'
+
+counterpart.registerTranslations('en', en)
+
+export default function DemoModeToggle ({ demoMode, setDemoMode }) {
+  return (
+    <CheckBox
+      checked={demoMode}
```
eatyourgreens, line 13: Adding `autoFocus` here might make the toggle a little easier to use from the keyboard. At the moment, you have to tab to the checkbox after the popup opens and it's not immediately obvious where the focus is.

eatyourgreens (approved):
Looks good to me. Works with a staging project if I add `dev=true` to the URL in the dev classifier. I have some suggestions for changes to the code but nothing blocking.

packages/lib-classifier/src/store/ClassificationStore.js
```diff
@@ -159,6 +160,14 @@ const ClassificationStore = types
         if (process.browser) {
           console.log('Completed classification', classificationToSubmit)
         }
+
+        if (self.demoMode) {
+          // subject advance is observing for this to know when to advance the queue
+          self.loadingState = asyncStates.posting
+          if (process.browser) console.log('Demo mode enabled. No classification submitted.')
+          return Promise.resolve(true)
+        }
```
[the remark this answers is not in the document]
eatyourgreens: Sorry, so the loading state isn't being set in two places. It's only being set once.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/ExpertOptions/components/DemoModeToggle/DemoModeToggle.js
srallen: The expert options context menu will also have the toggle for gold standard classification too once it is implemented, so I'm not sure we want to auto focus by input.

srallen:
@eatyourgreens 

> Works with a staging project if I add dev=true to the URL in the dev classifier

Just want to clarify, did you mean `demo=true` or did it work with `dev=true`?

eatyourgreens:
Sorry. `demo=true`.

Should we enable demo mode if that query param is present, and skip the toggle button?

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/ExpertOptions/components/DemoModeToggle/DemoModeToggle.js
eatyourgreens: Good point. Autofocus could be set by the parent, on the first input. It's not a big deal, but would mean you can toggle demo mode by clicking the cog button, then hitting space/enter.

srallen:
> Should we enable demo mode if that query param is present, and skip the toggle button?

No, I think we should still provide a means in the UI to enable and disable, especially if we change the mechanism to get access to the toggle.

packages/lib-classifier/src/store/ClassificationStore.js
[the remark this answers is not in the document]
srallen: I tried this early on in the previous version of this feature's implementation and I tried again just now to remind myself why I didn't do this. I get MST errors about trying to read or write to an object that is no longer part of a state tree with the classification meaning it's not advancing the subject queue and subsequently resetting the classification correctly.  Perhaps if it's broken out into more separate actions, it would work correctly, but that's more refactoring than I think is in scope of this PR.

PR #1925, opened 2020-12-08
Engaging Crowds: Subject picker tool

### LINKED ISSUE #1884: Project app: subject picker
**Package**
app-project

[Invision design](https://projects.invisionapp.com/share/9WNF9M73AEN#/screens/423863707)
Screenshot of the Invision design for a popup that lists subjects from a subject set

A popup that shows a table of subjects (with metadata) after picking a subject set to work on.

Question: how do we decide whether volunteers have the option to pick a subject? Is this an option that project owners would set when building a workflow?

eatyourgreens:
Work-in-progress towards allowing volunteers to select which subject they work on, from a sequential set of pages in a prioritised subject set. See #1884 for the designs.

You can test this locally on http://local.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020?env=production
After selecting workflow 16106 (the third workflow) and subject set, you should be shown a list of subjects from https://redsea.zooniverse.org. Clicking on any subject ID will take you to the classifier for the selected workflow and subject set. The subject should be selected in the URL.

There's only one subject set indexed at the moment, which is a set of letters from ASM. You should be able to search and sort that set using the Grommet data table headings.

TODO:
- [ ] Tests.
- [ ] Mock the API for the storybook.

A screenshot of the subject picker displayed on the home page of Shaun's test project.

Package:
app-project

Closes #1884.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/src/screens/ClassifyPage/components/WorkflowMenu/WorkflowMenu.js
+++ b/packages/app-project/src/screens/ClassifyPage/components/WorkflowMenu/WorkflowMenu.js
@@ -5,6 +5,7 @@ import { Modal } from '@zooniverse/react-components'
 
 import WorkflowSelector from '@shared/components/WorkflowSelector'
 import SubjectSetPicker from '@shared/components/SubjectSetPicker'
+import SubjectPicker from '@shared/components/SubjectPicker'
 
 /**
   A popup menu which allows you to choose a workflow and optional subject set from the classify page.
@@ -17,7 +18,18 @@ export default function WorkflowMenu({
 }) {
   const router = useRouter()
   const { owner, project } = router?.query || {}
+  const [ activeSubjectSet, setActiveSubjectSet ] = useState()
   const [ activeWorkflow, setActiveWorkflow ] = useState(workflowFromUrl)
+
+  function onSelectSubjectSet(event, subjectSet) {
+    const useSubjectSelection = activeWorkflow.id === '16106'
+    if (useSubjectSelection) {
+      event.preventDefault()
+      setActiveSubjectSet(subjectSet)
+      return false
+    }
+    return true
+  }
 
   function onSelectWorkflow(event, workflow) {
     if (workflow.grouped) {
@@ -32,6 +44,34 @@ export default function WorkflowMenu({
     setActiveWorkflow(null)
   }
 
+  let modalContent = (
+    <WorkflowSelector
+      onSelect={onSelectWorkflow}
+      workflows={workflows}
+    />
+  )
+  let baseUrl = `/projects/${owner}/${project}/classify`
+  if (activeWorkflow) {
+    baseUrl = `${baseUrl}/workflow/${activeWorkflow.id}`
+    modalContent = (
+      <SubjectSetPicker
+        baseUrl={baseUrl}
+        onClose={onClose}
+        onSelect={onSelectSubjectSet}
+        workflow={activeWorkflow}
+      />
+    )
+  }
+  if (activeSubjectSet) {
+    baseUrl = `${baseUrl}/subject-set/${activeSubjectSet.id}`
+    modalContent = (
+      <SubjectPicker
+        baseUrl={baseUrl}
+        subjectSet={activeSubjectSet}
+        workflow={activeWorkflow}
+      />
+    )
+  }
   return (
     <Modal
       active
@@ -39,18 +79,7 @@ export default function WorkflowMenu({
       title={activeWorkflow ? (activeWorkflow.displayName || 'Choose a subject set') : 'Choose a workflow'}
       titleColor={titleColor}
     >
-      {!activeWorkflow ?
-        <WorkflowSelector
-          onSelect={onSelectWorkflow}
-          workflows={workflows}
-        /> :
-        <SubjectSetPicker
-          onClose={onClose}
-          owner={owner}
-          project={project}
-          workflow={activeWorkflow}
-        />
-      }
+      {modalContent}
     </Modal>
   )
 }
--- a/packages/app-project/src/shared/components/SubjectPicker/SubjectPicker.js
+++ b/packages/app-project/src/shared/components/SubjectPicker/SubjectPicker.js
@@ -121,7 +121,7 @@ export default function SubjectPicker({ baseUrl, subjectSet, workflow }) {
             margin='medium'
             weight='bold'
           >
-            {subjectSet.title}
+            {subjectSet.display_name}
           </SpacedText>
         </Paragraph>
       </StyledBox>
@@ -145,7 +145,7 @@ export default function SubjectPicker({ baseUrl, subjectSet, workflow }) {
 SubjectPicker.defaultProps = {
   subjectSet: {
     id: '15582',
-    title: 'Anti-Slavery Letters: 1800-1839',
+    display_name: 'Anti-Slavery Letters: 1800-1839',
     metadata: {
       indexFields: 'date,title,creators'
     }
--- a/packages/app-project/src/shared/components/SubjectPicker/SubjectPicker.js
+++ b/packages/app-project/src/shared/components/SubjectPicker/SubjectPicker.js
@@ -1,0 +1,157 @@
+import { Modal, SpacedText } from '@zooniverse/react-components'
+import counterpart from 'counterpart'
+import { debounce } from 'lodash'
+import { useRouter } from 'next/router'
+import React, { useEffect, useState } from 'react'
+import styled from 'styled-components'
+import { Box, DataTable, Heading, Paragraph } from 'grommet'
+
+import { columns, fetchRows, fetchSubjects, searchParams } from './helpers'
+import en from './locales/en'
+
+counterpart.registerTranslations('en', en)
+
+/*
+  Grommet is opinionated about line-height and links it to font-size.
+  Reset the heading baselines here so that spacing is measured from
+  the tops and bottoms of the letters (without changing the text size.)
+  https://matthiasott.com/notes/the-thing-with-leading-in-css
+*/
+
+const StyledBox = styled(Box)`
+  min-height: 30px;
+`
+const StyledHeading = styled(Heading)`
+  line-height: 100%;
+`
+
+const SubjectDataTable = styled(DataTable)`
+  button {
+    padding: 0;
+  }
+  th button {
+    color: ${props => props.theme.global.colors['dark-1']};
+    font-weight: bold;
+    text-transform: uppercase;
+  }
+  th svg {
+    stroke: ${props => props.theme.global.colors['dark-1']};
+  }
+`
+
+const PAGE_SIZE = 100
+
+export default function SubjectPicker({ baseUrl, subjectSet, workflow }) {
+  const [ rows, setRows ] = useState([])
+  const [ query, setQuery ] = useState('')
+  const [ sortField, setSortField ] = useState('subject_id')
+  const [ sortOrder, setSortOrder ] = useState('asc')
+  const { indexFields } = subjectSet.metadata
+  const customHeaders = indexFields ? indexFields.split(',') : ['date', 'title', 'creators']
+
+  async function fetchSubjectData() {
+    const subjects = await fetchSubjects('15582', query, sortField, sortOrder, PAGE_SIZE)
+    const rows = await fetchRows(subjects, workflow, PAGE_SIZE)
+    setRows(rows)
+  }
+
+  useEffect(function onChange() {
+    fetchSubjectData()
+  }, [query, sortField, sortOrder])
+
+  function search(data) {
+    const query = searchParams(data)
+    setQuery(query)
+  }
+
+  function sort(data) {
+    const { property: sortField, direction: sortOrder } = data
+    if (sortField === 'status') {
+      return true;
+    }
+    setSortField(sortField)
+    setSortOrder(sortOrder)
+  }
+
+  const background = {
+    header: "accent-2",
+    body: ["white", "light-1"]
+  }
+  const pad = {
+    header: "xsmall",
+    body: "xsmall"
+  }
+
+  /*
+    Vertical spacing for the picker instructions.
+    The theme's named margins are set in multiples of 10px, so set 15px explicitly.
+  */
+  const textMargin = {
+    top: '15px',
+    bottom: 'medium'
+  }
+  return (
+    <>
+      <StyledHeading
+        level={3}
+        margin={{ top: 'xsmall', bottom: 'none' }}
+      >
+        {counterpart('SubjectPicker.heading')}
+      </StyledHeading>
+      <Paragraph
+        margin={textMargin}
+      >
+        {counterpart('SubjectPicker.byline')}
+      </Paragraph>
+      <StyledBox
+        background='brand'
+        fill
+        pad={{
+          top: '5px',
+          bottom: 'none',
+          horizontal: 'none'
+        }}
+      >
+        <Paragraph
+          color="white"
+          margin='none'
+          textAlign='center'
+        >
+          <SpacedText
+            margin='medium'
+            weight='bold'
+          >
+            {subjectSet.title}
+          </SpacedText>
+        </Paragraph>
+      </StyledBox>
+      <SubjectDataTable
+        background={background}
+        columns={columns(customHeaders, baseUrl)}
+        data={rows}
+        fill
+        onSearch={debounce(search, 500)}
+        onSort={sort}
+        pad={pad}
+        pin
+        replace
+        sortable
+        step={PAGE_SIZE}
+      />
+    </>
+  )
+}
+
+SubjectPicker.defaultProps = {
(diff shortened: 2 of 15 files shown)


packages/app-project/src/screens/ClassifyPage/components/WorkflowMenu/WorkflowMenu.js
```diff
@@ -17,8 +18,19 @@ export default function WorkflowMenu({
 }) {
   const router = useRouter()
   const { owner, project } = router?.query || {}
+  const [ activeSubjectSet, setActiveSubjectSet ] = useState()
   const [ activeWorkflow, setActiveWorkflow ] = useState(workflowFromUrl)
 
+  function onSelectSubjectSet(event, subjectSet) {
+    const useSubjectSelection = activeWorkflow.id === '16106'
```
shaunanoordin, line 25: So the EC Subject Picker only works for _production_ workflow 16106? This works for me - I'm fine with temporary hardcoding - but I think this might benefit from a huge comment block saying `/* Wakka wakka! The Subject Picker only work for specific Engaging Crowds projects right now, because it has an external dependency at subject-set-search-api.zooniverse.org, which requires specific subject sets pre-loaded into the database. We'll fix this some time after Junetober 2021! Wakka wakka! */`

My basic sanity check is that this line (along with the similar line in the _other WorkflowMenu.js_ ) should explain, to any Zooniverse developer unfamiliar with EC, 1. **why** this code only works for _production_ workflow 16106, and 2. **when** such temporary hardcoding should be re-examined.

eatyourgreens: Datasette is set up now, so I think the next PR will remove this in favour of pinging Datasette to check if the active subject uses indexed subjects.

packages/app-project/src/shared/components/SubjectPicker/SubjectPicker.js
```diff
@@ -0,0 +1,157 @@
...
+    display_name: 'Anti-Slavery Letters: 1800-1839',
+    metadata: {
+      indexFields: 'date,title,creators'
+    }
+  },
+  workflow: {
+    id: '5329',
+    display_name: 'Transcribe Text (Main Workflow)'
+  }
+}
```
shaunanoordin, line 157: I'm not sure how I feel about a component having "example data" as its default values, instead of "blank data". This is my brain right now:

Bad: using "example data" as the default might lead to wonky debugging situations. e.g. subject set fails to load, and suddenly we're told it loaded "Anti-Slavery Letters: 1800-1839"

Good: "example data" provides a good template of how to use the component.

Bad: if you want to use "example data", do it in a component Story instead.

https://www.youtube.com/watch?v=XlcFTbgoeBk

eatyourgreens: I’d forgotten about these. They were set up for a demo of the subject picker in December. 😲

packages/app-project/src/shared/components/SubjectPicker/helpers/checkRetiredStatus.js
```diff
@@ -0,0 +1,20 @@
...
+      page_size,
+      subject_ids,
+      workflow_id
+    })
+  const { subject_workflow_statuses } = response.body
+  subject_workflow_statuses.forEach(status => {
+    const inProgress = status.classifications_count > 0 ? 'In progress' : 'Unclassified'
+    retirementStatuses[status.links.subject] = status.retired_at ? 'Retired' : inProgress
+  })
+  return retirementStatuses
```
shaunanoordin, line 19: ⚠️ Let me talk myself through some sanity checks here.

OK, so on a _successful_ scenario, we get data from /subject_workflow_statuses, and we use the data from that to populate 'retirementStatuses' (an initially empty object `{}`), which is returned when `fetchRows()` calls `const retirementStatuses = await checkRetiredStatus(subject_ids, workflow, page_size)`. Cool!

⚠️ For failure scenarios, here's what can possibly go wrong:
1. panoptes doesn't return a valid result (e.g. endpoint is dead) and returns a bad response. (It probably won't just return undefined.)
2. Will a bad `response` have .body? And will that .body have .subject_workflow_statuses? (Probably not)
3. If subject_workflow_statuses is _undefined,_ that'll throw an error out of `checkRetiredStatus()` and send it up the stack to fetchRows, which can't handle the error either.

Thoughts: `checkRetiredStatus` seems a bit brittle. If it breaks, it'll break everything downstream. Maybe wrap everything in a try-catch so it always returns an empty object `{}` on error?

Counterpoint: Wait, Shaun, do you REALLY want a try-catch to _hide_ the error? If getting the value of subject_workflow_statuses from panoptes _is essential,_ then _it's best_ that an error at checkRetiredStatus() ends up crashing everything downstream, so we know something went wrong at panoptes. Better an explicit error than leading users around with false values.

Counter-counter point: OK, fine, go ahead and crash, but at least spit out an error message that makes sense to users.

packages/app-project/src/shared/components/SubjectPicker/helpers/searchParams.js
```diff
@@ -0,0 +1,10 @@
+export default function searchParams(data) {
+  let query = ''
+  Object.entries(data).forEach(([key, value]) => {
+    if (value !== '') {
+      query += `@${key}:${value}*`
```
shaunanoordin, line 5: ⚠️❓ Is there any chance that somebody will try to search for a value that'll break this query? e.g. `title = 'A Concise Guide to Colons: Volume XI: Deluxe Edition: This Time It's Personal'` and `author = 'shaun@zooniverse'`

shaunanoordin:
OK! I've completed my code read, which is mostly good, and I've left comments in places where I had questions. Jim, I see you already had a chance to see & respond to some of my code comments, but here's the list for easy reference:

- [x] Workflow 16106 is hardcoded, and hardcoded stuff should be signposted better.
  - Response: will be addressed in the next PR, which should hook things up to Datasette.
- [x] Subject Picker has "example data" as its default values, instead of blank values. 
  - Response: this is an error, and a leftover from an earlier demo.
- [ ] Should `checkRetiredStatus()` have error handling?
- [ ] Should `searchParams()` escape special characters for 'value'?

I'll run functionality tests later, but I'll keep an eye on things in case there are newer commits that make my earlier comments redundant (e.g. removing Subject Picker's "example data") 👍

packages/app-project/src/shared/components/SubjectPicker/helpers/searchParams.js
eatyourgreens: Good point. I think this will also be fixed by the move to Datasette, which uses parameterised SQL queries. Queries against the new API will look something like `/subjects/92752?_sort=priority&Date__contains=December`.

packages/app-project/src/shared/components/SubjectPicker/helpers/checkRetiredStatus.js
eatyourgreens: There's an issue with Panoptes where connections will occasionally reset during a query. See https://github.com/zooniverse/panoptes-cli/issues/193

I think you're right that API requests should be wrapped in try/catch. I'll add Sentry logging too.

packages/app-project/src/screens/ClassifyPage/components/WorkflowMenu/WorkflowMenu.js
eatyourgreens: Datasette sends back 404 if you request a table that doesn't exist. I'll use that to check whether indexing is supported for real projects.
https://subject-set-search-api.zooniverse.org/subjects/927520.json

packages/app-project/src/shared/components/SubjectPicker/SubjectPicker.js
eatyourgreens: Removed now. Also, `defaultProps` is deprecated for functional components anyway.

packages/app-project/src/shared/components/SubjectPicker/helpers/checkRetiredStatus.js
shaunanoordin: 👍

packages/app-project/src/shared/components/SubjectPicker/helpers/searchParams.js
eatyourgreens: I've got a branch set up now which gets this working with Datasette. https://github.com/zooniverse/front-end-monorepo/tree/subject-set-search-api

shaunanoordin (approved):
## PR Review

Package: `app-project`

This PR adds a "Subject Selector" functionality (which is an extension of the "Workflow Selector" and "Subject Set Selector" functionality) to a Project's Classifier.



At the moment, this functionality only works for _production workflow 16106._

- Prerequisite:For properly configured _project_ (i.e. can select workflow), with a properly configured _workflow_ (i.e. can select Subject Set, and id=16106), the "Subject Select" feature can be reached by going to Select Workflow -> Select Subject Set -> Select Subject
  - This can be initiated from the Project home page, [(example)](http://local.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020?env=production) or the Project's default classify route, [(example)](http://local.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify?env=production)
- Main feature: the "Select Subject" component _does_ list all members of the specified Subject Set.
  - Clicking on a Subject in the list ⚠️  _should_ open a link to Classify that specific Subject (in that specific subject set, in that specific workflow), e.g. http://local.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/16106/subject-set/85773/subject/13885427?env=production
  - Clicking on the 🔍 search icon and typing some text ⚠️ _does (mostly)_ filter out the search results
- New external requirements:
  - Subject search endpoint: https://redsea.zooniverse.org
  - Specific subject selection endpoint:  https://www.zooniverse.org/api/subjects/selection

Code read looks good, and I think the _main intended functionality of this PR_ looks good, though there are some gotchas I'll discuss.

### Testing

Tested on localhost, [using a production project designed to test Engaging Crowds,](http://local.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/16106/subject-set/85773/subject/13885427?env=production) with macOS10 + Chrome.

Baseline sanity check:
- For a standard non-Engaging Crowds [(example)](http://local.zooniverse.org:3000/projects/darkeshard/transformers/classify)
  - From the standard project home page, I can select a Workflow
  - Using a saved URL, I can open a specific workflow
  - From the root /classify page, I _should be able to theoretically_ select a Workflow...?
    - The "select WF" component is showing correctly, but I keep running into runtime errors.
    - This is probably a problem unrelated to this PR. Filing under the Rabbit Hole.

I'm going to give a 👍 for the baseline sanity check.

Not fully tested:
- Attempting to access the classifier via ID-specific URLs on _projects that allow WF+SSet selection_ (i.e. /classify with workflow ; /classify with workflow + subject set ; /classify with workflow + subject set + subject) (see Rabbit Hole of Madness below)

**Main Testing!**

**Feature 1a:** "Subject Select" feature works from home page

1. Open http://local.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020?env=production
2. A list of workflows should be visible. ("Select Workflow" feature.) Click on "v3 dropdown-simple (grouped)" (aka WF16106)
3. A list of subject sets should be visible. ("Select Subject Set" feature.) Click on "Transformers"
4. A list of Subjects will be shown. ("Select Subject" feature.) Click on any Subject.
5. You'll be taken to the correct URL, e.g. http://local.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/16106/subject-set/85772/subject/13885427?env=production loads correctly
  - ❗ Note: the list of Subjects is actually a cheat, and does _not_ actually list the Subjects inside SSet . i.e. Subject 13885427 is actually a scanned image of a page, not a Transformer cleverly in disguise as one.
  - Opening the "correct URL" will actually result in the Classifier crashing, because 1388527 isn't a part of SSet 85772. The /subjects/selection endpoint will return a 422. **This error is expected.**
  - Curiously though, the endpoint has the error message: _"found unpermitted parameter: subject_set_id"_ , which is... WTF? @eatyourgreens I'm going to highlight this bit for you in case it's something that comes up in the next iteration. Not sure if the endpoint really doesn't like the subject_set_id parameter or if it's the wrong error message (I expected _"yo subject 13885427 ain't a part of the workflow, you dolt"_ )

**Feature 1b:** "Subject Select" feature works from /classify? In theory?

- Repeat previous steps, but replace Step 1 with: Open http://local.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify?env=production
- The "Select Workflow", "Select Subject Set", and "Select Subject" features all load correctly.
- BUT there is runtime error I keep encountering. See Rabbit Hole of Madness (Issue 2) since I think the error isn't directly related to this PR, but is a separate thing to investigate.

**Feature 2:** "Search the Subject List" feature works

1. Open http://local.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020?env=production, select Transformers SSet.
2. List of Subjects should be visible.
3. Click on 🔍 search icon next to "Title", type in "envelope".
4. Results will now be filtered to show only Subjects with title containing "envelope" (11 results) 👍 
5. Click on 🔍 search next to "Creator", type in "may".
6. Results will now be filtered to show only Subjects with title containing "envelope" AND creator containing "may" (2 results) 👍 
7. Change the "Creator" text from "may" to "may:"
8. Page will crash with the runtime error, `TypeError: subjects.map is not a function` (in fetchRows.js) ❌ 

The Search function works well in most cases, EXCEPT when a special character is thrown in. In Step 7 above, we introduce `:` into the mix, which causes shenanigans at the endpoint.

Specifically: `https://redsea.zooniverse.org/search/15582?filter_field=@title:envelope*@creators:may:*&limit=100&sort_field=subject_id&sort_order=asc` returns a 404, with the error message "Syntax error at offset 29 near may"

I'm pretty sure the characters `?&*@` can all lead to wonkiness.

### Rabbit Hole of Madness

The following observations are **not** related to this PR. But, I'm making a note of them since I accidentally bumped into these weirdness while testing, and these may be items that need further investigation.

**Unknown Issue 1:** /queued endpoint seems to ignore subject_set_id

- If I open http://local.zooniverse.org:3000/projects/darkeshard/engaging-the-crowds-2020/classify/workflow/15652/subject-set/85771?env=production , I can consistently get Subjects from OUTSIDE the Subject Set. What??
- OK, see, WF15652 is attached by default to three SSets (85771 Ultraman, 85772 Transformers, 85773 Overwatch) so by default that WF should pull Subjects from across those SSets.
- However, while I can see that the Network call is to `https://www.zooniverse.org/api/subjects/queued?workflow_id=15652&subject_set_id=85771&http_cache=true` (which should show Ultraman), the endpoint is STILL giving me characters from Overwatch and Transformers!

Not sure WTF is going on here but this seems like a big whoops that needs debugging.

EDIT 1: Shaun, can you check if the /queued?subject_set_id requires a whitelist, or another toggle to work with _staging_ WF 15652?

EDIT 2: can you check it the /queued?subject_set_id is only borked on _staging_ and not production?

**Unknown Issue 2:** can't properly select Workflow from /classify URL

- Go to http://local.zooniverse.org:3000/projects/darkeshard/transformers/classify . You should see the "Workflow Selection" component
- Choose a workflow, e.g. "Autobot or Decepticon?"
- You'll be directed to http://local.zooniverse.org:3000/projects/darkeshard/transformers/classify/workflow/2628
- At this point, you'll receive the error: `Unhandled Runtime Error - Error: MobX Provider: The set of provided stores has changed.` (Note: you might also encounter this error earlier, at the first step. This usually ONLY happens the FIRST time you load /classify, which leads me to suspect it's a weird Next.JS thing.)
- Reload the page or go to http://local.zooniverse.org:3000/projects/darkeshard/transformers/classify/workflow/2628 directly. Now, the page loads and the classifier works fine! It's just the _routing from_ /classify that borks everything.

I'm pretty sure we stomped out this issue before?? Gonna test with `master` to see when this problem started re-occuring.

### Status

I may have rambled on for a bit, but the short of it is:

LGTM 👍 Approved!

The main feature ("Select Subject" functionality and component) works perfectly fine. The limitation with hardcoded WF ID and a sorta-hardcoded list of Subjects (that don't reflect the selected Subject Set) is A-OK with me for now since they're temporary fixes which will be addressed in a separate PR.

Similarly, the search feature has some issues with escaping special characters, but that's a small fix for a separate PR.

That only leaves the rabbit hole of issues I need to investigate. I'm going to find out WTF is happening for both Issues 1 and 2,  but don't let that stop this PR from getting merged. Let's gooo!

eatyourgreens:
Thanks @shaunanoordin. #2132 is a quick draft PR to switch the API over to Datasette. It might solve some of those issues you found.

PR #1943, opened 2020-12-14
Restrict the Jenkins builds

eatyourgreens:
Only run the Jenkins builds on changes to master or the production-release tag.

At the moment, the Branch Indexing trigger in Jenkins is launching ~20 builds of the monorepo, which seems to be overwhelming Jenkins when it happens.
https://jenkins.zooniverse.org/job/Zooniverse%20GitHub/job/front-end-monorepo/indexing/console

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


camallen (approved):
I'm not against this change but we lose the visibility of any build tooling changes on PRs.

short term - perhaps it might be best to run this build testing in CI via github actions and offload deployment to jenkins (access to the K8s cluster). That will allow us to test the GHActions tooling and free up the jenkins server to focus on CD.

Finally - I've changed Jenkins (with Jim) to avoid the build storms by ingoring changed target branches. Instead Jenkins will now only run the PR / branch on pushes to the branch.

Jenkinsfile
```diff
@@ -46,12 +52,7 @@ pipeline {
           def buildArgs = "--build-arg APP_ENV --build-arg COMMIT_ID --build-arg CONTENTFUL_ACCESS_TOKEN --build-arg CONTENTFUL_SPACE_ID --build-arg CONTENT_ASSET_PREFIX --build-arg SENTRY_CONTENT_DSN --build-arg PROJECT_ASSET_PREFIX --build-arg SENTRY_PROJECT_DSN ."
           def newImage = docker.build(dockerImageName, buildArgs)
           newImage.push()
-
-          if (BRANCH_NAME == 'master') {
-            stage('Update latest tag') {
-              newImage.push('latest')
-            }
-          }
+          newImage.push('latest')
```
camallen, line 55: I think you still want the `branch == 'master'` conditional here to ensure the production-release tag builds don't clobber the latest master changes.

eatyourgreens: Good point. I was assuming that master branch changes always build the `-staging` image, whereas on the tag change we want to build and push the `-production` image. Maybe I should split those builds up into two separate scripts to clarify what's going on, rather than having on script that builds both images. The code would be more verbose, but I think might be easier to read.

eatyourgreens:
We can wait to see if the Jenkins config change has fixed the build storms, then come back to this if it's still needed.

eatyourgreens:
#1944 runs a test production build as part of CI on GitHub actions. It's much faster than Jenkins (~4 minutes) but does require some small code changes to get the tests passing in a production environment. If we do merge this PR, I think we should merge that PR as well so that we don't get caught out by a broken build on staging.

Jenkinsfile
eatyourgreens: I think I will do that. It would mean writing out the staging and production build environments in separate blocks, which would also make the code easier to read.

eatyourgreens:
@camallen I've rewritten the staging vs. production builds to, hopefully, make them clearer by removing all the ternary expressions.

Jenkinsfile
```diff
@@ -25,33 +25,60 @@ pipeline {
     // Right now, we're *only* building and deploying on the `master` branch;
     // longer-term, we'll want to deploy feature branches as well.
 
-    stage('Build base Docker image') {
+    stage('Build staging Docker image') {
+      when {
+        branch 'master'
```
eatyourgreens, line 30: Remove this check if we want to continue running test builds on every PR, then wrap the push to DockerHub in a conditional.

Jenkinsfile
camallen: ensure only master branch pushes to latest tag
```suggestion
```

eatyourgreens: Don't we want to push `zooniverse/front-end-monorepo-staging:latest`and `zooniverse/front-end-monorepo-production:latest` separately?

camallen: AH 🤦 - the image labels are different my apologies. Ignore my comment.

eatyourgreens: No problem. The builds are slightly different because of the different asset prefixes and sentry environments.

PR #1946, opened 2020-12-16
Fix mobx warning about accessing an undefined index in an array

srallen:

Package: lib-classifier

This fixes something I missed while reviewing #1885. The change on the `Workflow` model type for steps caused this warning to start to be thrown: `[mobx.array] Attempt to read an array index (1) that is out of bounds (1). Please check length first. Out of bound indices will not be tracked by MobX`

This only appears with projects actually configured to use steps, so TESS, Stellar Sleuths and any staging test project we've setup as well and in the specs (if you scope to this model's specs, the warning is thrown, but doesn't fail a test either. Not sure if we want to start testing for warnings...)

The issue is that mobx warns about code that is attempting to read an undefined index in an array since undefined values are not observed. This can and has caused bugs for us in the past, so good to address this kind of warning. As far as I can tell, it didn't actually break anything on TESS.

I've fixed this by checking the index against the length rather than directly checking for a defined value at that index in the array. I'm not exactly sure why the type change caused this. Perhaps the initial union type definition did something else under the hood I'm not understanding at the moment (and I don't have the time right now to invest in that). 

Long term, I'd like to see the data actually stored here to actually be a map. That was always the intention. The storing of an array of pairs is a workaround for JSONB storage requirements on Panoptes. Even _later_ long term, directly storing the steps model would be even better. Both require more refactoring time that isn't really afforded to me right before the holidays. So for another day.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/WorkflowStepStore.js
+++ b/packages/lib-classifier/src/store/WorkflowStepStore.js
@@ -156,7 +156,9 @@ const WorkflowStepStore = types
       const { steps } = workflow
       steps.forEach(([ stepKey, step ], index) => {
         let next
-        if (steps[index + 1]) {
+        // checking for there being a next step without accessing it directly
+        // mobx throws errors for undefined indices because you can't observe those
+        if (steps.length > 0 && index + 2 <= steps.length) {
           const [nextStepKey] = steps[index + 1]
           next = nextStepKey
         }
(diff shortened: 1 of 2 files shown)


mcbouslog (approved):
nice fix! LGTM

Noted while testing this that the TESS related staging project workflow (project 1862, workflow 3317, https://fe-project.preview.zooniverse.org/projects/nora-dot-test/planet-finders-beta/classify?env=staging) is not configured for steps, unlike the production project/workflow. I'll make a note to revisit that, though I don't think urgent.

PR #1961, opened 2021-01-07
Classifier: Project default workflow

eatyourgreens:
In parallel with the changes to the project model in #1823, this refactors the project model, in the classifier, to have a `project.defaultWorkflow` view. `project.defaultWorkflow` replaces the workflow selection logic in `workflows.getDefaultWorkflowId()`.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


shaunanoordin:
Code changes LGTM on a quick reading. These changes will affect the behaviour of default workflow selection (i.e. whenever somebody goes to `projects/foo/bar/classify/`) so I'm going to make a note of my understanding of what was, and what will be, in terms of workflow-loading actions.

Scenario A: Project has 1 (active) workflow, no default 
- Previously: that workflow is selected
  - _Well actually,_ in PFE, a random workflow is selected, it's just that it's a random selection from a pool of one.
- Now: same

Scenario B: Project has 1 (active) workflow, set as default 
- Previously: that workflow is selected 
- Now: same

Scenario C: Project has 5 (active) workflows, no default 
- Previously: the first of the 5 active workflows is selected
  - _Well actually,_ this is true only in monorepo.
  - _Well actually,_ in PFE a _random_ workflow out of the 5 is selected.
  - _Well actually,_ the monorepo was intended to select randomly from the 5 as well, according to the inline comments
- **Now: NO workflow is selected**

Scenario D: Project has 5 (active) workflows, one of which is default 
- Previously: Default workflow is selected? (Shaun: doublecheck pls)
- Now: same. (Default workflow is selected)

_The following are error scenarios that won't hold weight over the PR review, but is listed for thoroughness and future testing._ 

Scenario E: Project has 5 (active) workflows, but the default is an INACTIVE workflow
- Previously: default workflow is selected, even if invalid, causing a crash??? I think??
- Now: default workflow is selected, even if invalid, causing a crash?

Scenario F: Project has 5 (active) workflows, but the default is a DELETED workflow
- Previously: ???
- Now: default workflow is selected, even if invalid, causing a crash?
  - _Well actually,_ this has happened on PFE before, quite a few times, so try to find the old issues that described this problem and the solution was just "oh right, you forgot to set a valid Default Workflow yo"

Please read all the _Well actually..._ context explainers in the same voice as [Comic Book Guy](https://simpsons.fandom.com/wiki/Comic_Book_Guy) explaining some pointless trivia

shaunanoordin (approved):
## PR Review

package: `lib-classifier`

This PR updates the Project model so it has a `.defaultWorkflow` property. This is particularly important when a user accesses a URL such as `zooniverse.org/projects/foo/bar/classify` - i.e. which workflow should the Classifier serve if no explicit workflow ID is specified?

Code changes LGTM and the behaviour looks good. Notably, there's one new behaviour: if a project has _many active workflows, but no default,_ then _NO workflow is served._

See Dev Notes for additional behaviour scenarios.

### Dev Notes

With this PR, the default workflow that's chosen will follow this behaviour pattern:

**Scenario A:** Project has 1 (active) workflow, no default
- that workflow is selected 

**Scenario B:** Project has 1 (active) workflow, set as default
- that workflow is selected

**Scenario C:** Project has 5 (active) workflows, no default
- 🆕 ⭐ NO workflow is selected
- This is a change from previous behaviours.
  - In the monorepo, the first of the 5 workflows is selected.
  - In PFE, a random workflow from the 5 is selected.

**Scenario D:** Project has 5 (active) workflows, one of which is default
- Default workflow is selected.

The following are failure scenarios:

**Scenario E:** Project has 5 (active) workflows, but the default is an INACTIVE workflow
- NO workflow is selected. (See Scenario G)
- Note that in PFE, the classifier gets stuck in "loading workflow..."

**Scenario F:** Project has 5 (active) workflows, but the default is a DELETED workflow
- NO workflow is selected. (See Scenario G)
- Note that in PFE, a random active workflow is selected.

**Scenario G:** An invalid or inactive workflow is selected for a valid project
- NO workflow is selected. Classifier page loads (i.e. doesn't crash), but the Subject Viewer and Task Panel are empty of content.

**Scenario H:** Project has NO active workflow.
- NO workflow is selected. (See Scenario G)

These tests were mostly performed by editing the project `staging:darkeshard/transformers` (URLs: `https://localhost:8080/?env=staging&project=1651`, `https://master.pfe-preview.zooniverse.org/projects/darkeshard/transformers/classify`) and comparing the behaviour differences between this branch, PFE, and the monorepo master branch. macOS10 + Chrome87, _using incognito mode to avoid the "logged in user's saved workflow" feature,_ which would have mucked up the workflow selection logic even further.

### Status

LGTM 👍 Functionality checks out, code looks legit, and I approve of the new behaviour for multiple-workflows-but-no-default.

eatyourgreens:
>Notably, there's one new behaviour: if a project has _many active workflows, but no default,_ then _NO workflow is served._

I think this is ok, because we do want you to actively choose a workflow if there are many workflows and none are set as a default. At the moment, this may crash the classifier but we'll take care of that with new workflow selection features.

srallen:
@eatyourgreens is correct for scenario F, we'll have the selection UI to resolve this. Scenarios G and H should have some kind of messaging rather than just no load with a blank grid area.

packages/lib-classifier/src/store/Project/Project.js
```diff
@@ -0,0 +1,24 @@
...
+    slug: types.string
+  })
+  .views(self => ({
+    get defaultWorkflow() {
+      const activeWorkflows = self.links['active_workflows']
+      let singleActiveWorkflow
+      if (activeWorkflows.length === 1) {
+        [singleActiveWorkflow] = self.links['active_workflows']
+      }
+      const defaultWorkflow = self.configuration['default_workflow']
```
srallen, line 19: I didn't get a chance to leave this comment before this got merged, but we should be checking that the default workflow is active. The lab UI prevents a default being set if it's inactive, but if there's a bug in the lab or if a project owner somehow manages to configure an inactive workflow to be default some other way (maybe python client), then we have an issue. We should validate its activeness.

eatyourgreens: This line's been removed now. When a project has multiple workflows, a volunteer has to specify which one they're going to classify on. The classifier won't fall back to a project default.

PR #1962, opened 2021-01-07
Fix initStore args

eatyourgreens:
The first argument passed to initStore should be the boolean flag `isServer`. This PR calls `initStore(true, snapshot)`, rather than `initStore(snapshot)`.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/src/helpers/getDefaultPageProps/getDefaultPageProps.js
+++ b/packages/app-project/src/helpers/getDefaultPageProps/getDefaultPageProps.js
@@ -7,12 +7,13 @@ export default async function getDefaultPageProps({ params, query, req, res }) {
   // cookie is in the next.js context req object
   const mode = getCookie(req, 'mode') || null
   const dismissedAnnouncementBanner = getCookie(req, 'dismissedAnnouncementBanner') || null
-  const store = initStore({
+  const snapshot = {
     ui: {
       dismissedAnnouncementBanner,
       mode
     }
-  })
+  }
+  const store = initStore(true, snapshot)
 
   if (params.owner && params.project) {
     const { owner, project } = params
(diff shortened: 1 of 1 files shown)


packages/app-project/src/helpers/getDefaultPageProps/getDefaultPageProps.js
```diff
@@ -7,12 +7,13 @@ export default async function getDefaultPageProps({ params, query, req, res }) {
...
   const dismissedAnnouncementBanner = getCookie(req, 'dismissedAnnouncementBanner') || null
-  const store = initStore({
+  const snapshot = {
     ui: {
       dismissedAnnouncementBanner,
       mode
     }
-  })
+  }
+  const store = initStore(true, snapshot)
```
srallen, line 16: I'd prefer `true` be set to a variable to clearly show what this parameter is for. Perhaps:
```suggestion
  const isServer = true
  const store = initStore(isServer, snapshot)
```

srallen (approved):
Loading the setting in the cookie for the theme mode is working now. I have a few non-blocking comments that can be done in follow up because I'd like to get this merged so we can deploy today:
- `getDefaultPageProps` doesn't have specs. I'd like to see at least some basic unit tests for this.
- `initStore` function and specs could be updated to make sure an error like this doesn't occur in the future

eatyourgreens:
I think `isServer` can be deprecated. It seems to be `true` throughout the code. I haven't done a thorough search, though, and the lack of tests for `getInitialProps` made me nervous about outright removing it.

srallen:
@eatyourgreens, I forgot to mention, the `_app.js` file also has an `initStore` in its class constructor. Is this cruft too?

eatyourgreens:
I think that initialises the store in the browser, from the `initialState` prop.

PR #1963, opened 2021-01-08
Remove /classify as the route for a project's default workflow

eatyourgreens:
This PR removes `/classify` as the route for a project's default workflow. If a project has a single active workflow, then `/classify` will set that as the `workflowID` prop for the classifier. If a project has multiple workflows, then you are expected to select a workflow by specifying it in the URL: `/classify/workflow/:workflow_id`. In the second case, `/classify` no longer selects the default workflow for you.

One side effect of this change is that navigation links and workflow links on the project homepage now link to workflows by ID, even for projects that have only one workflow.

Package:
app-project
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/src/components/ProjectHeader/ProjectHeaderContainer.js
+++ b/packages/app-project/src/components/ProjectHeader/ProjectHeaderContainer.js
@@ -1,6 +1,6 @@
 import { inject, observer } from 'mobx-react'
 import { bool, shape, string } from 'prop-types'
-import React, { Component } from 'react'
+import React from 'react'
 import { withRouter } from 'next/router'
 
 import ProjectHeader from './ProjectHeader'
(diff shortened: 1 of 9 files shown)


shaunanoordin (approved):
## PR Review

Package: `app-project` (primary), `lib-classifier` (secondary)

This PR is a major change to the way the `/classify` path functions on `app-project`, e.g. `https://master.pfe-preview.zooniverse.org/projects/darkeshard/transformers/classify`

- if the project has a _single Workflow,_ that workflow ID will be linked in the Project Header's "Classify" link (*️⃣ see Status and Testing notes)
- going to the `/classify` path if the project has _MULTIPLE active workflows and NO default_ will result in an error, **which is expected new behaviour.** ❗ 
  - Note: same is true if the project has no active workflow
- Please see Testing section below for full behaviour record

### Testing

Tested `app-project` on `http://localhost:3000/projects/darkeshard/transformers` with macOS10 + Chrome 87. Staging Project 1651 was configured multiple times for this test.

Scenario 1: project has **one active workflow, which is the default**
- the "Classification" link in the header leads to: `/projects/darkeshard/transformers/classify/workflow/3262`
- navigating directly to the /classify route results in: workflow 3262 loading. /classify path is valid.

Scenario 2: project has **one active workflow, no default**
- Same results as Scenario 1, as expected

Scenario 3: project has **multiple active workflows, one which is the default**
- *️⃣ the "Classification" link in the header leads to: `/projects/darkeshard/transformers/classify` ❓ _(Was this supposed to be /classify/workflow/3262?)_
- navigating directly to the /classify route results in: workflow 3262 (the default) loading. /classify path is valid.

Scenario 4: project has **multiple active workflows, none of which is the default**
- the "Classification" link in the header leads to: `/projects/darkeshard/transformers/classify` _(This is sensible)_
- navigating directly to the /classify route results in: ERROR, No Workflow ID available _(Which IS expected 👍 )_

Scenario 5: project has **no active workflows**
- Same results as Scenario 4

Not tested: projects where the default isn't an active workflow. That's too much of an erroneous case to consider. 

### Status

LGTM

*️⃣ I have a bit of a question of how the ProjectHeaderContainer receives its `defaultWorkflow` variable (since if there are multiple workflows but no default, I think nothing is passed in) but that might be saved as a separate issue to investigate

srallen:
Since work has begun and now merged toward the feature, could we get a proposal ADR written up on these changes in how workflow selection works compared to PFE? We should include this change and the other proposed changes from the associated RFC.

PR #1974, opened 2021-01-12
Add SubjectGroup end point to Classifier

shaunanoordin:
## PR Overview

package: `lib-classifier`
context: this is for the Survos 2020/21 project

This PR changes the Subject API endpoint to the SubjectGroup endpoint, when the Workflow is using the Subject Group Viewer.

- The default Subject API endpoint for the classifier is `/subjects/queued`, of course.
- This is switched to the new `/subjects/grouped` when `workflow.configuration.subject_viewer === 'subjectGroup'`

External dependencies:
- This new SubjectGroup API endpoint was introduced to Panoptes for the Survos project and is still being updated.
  - Initial setup for the endpoint can be seen here: https://github.com/zooniverse/panoptes/pull/3530
- ⚠️ However, as of 12 Jan 2021, the API requires a few issues to be fixed before this PR is ready to be used.
  - [ ] https://github.com/zooniverse/panoptes/issues/3549 - endpoint returns image URLs in an incorrect format. (We can have a temporary workaround on the front end for this, if necessary.)
  - [x] https://github.com/zooniverse/panoptes/issues/3550 - endpoint has a database error for specific Subject Sets
  - [x] https://github.com/zooniverse/panoptes/issues/3551 - endpoint does not recognise `http_cache=true` parameter

### Status

Code is complete and a code-only review can begin, but we're waiting for the API issues to be addressed before any realistic testing can begin. Marking this as a WIP/draft until the API endpoint is sorted.


shaunanoordin:
API update:
- Panoptes update [3554](http://github.com/zooniverse/panoptes/issues/3554) has made the Subject Group API endpoint mostly functional! 👍 
- There is still an issue with the broken URLs, but 0bf98b5 installs a temporary workaround for this.
- Next up: the Subject Group Annotations need to know what _Subject ID in the Subject Group_ was selected, not just the _index of the cell of the Subject Group_ that was selected.

Note to self: tests?

shaunanoordin:
## PR Update

- The SubjectGroupComparison annotation now tracks the index of the selected cell as WELL as the Subject ID of said cell.
  - Previously: `annotations: [0, 5, 10]`
  - Now: `annotations: [ { index: 0, subjectId: '932928' }, { index: 5, subjectId: '938613' }, { index: 10, subjectId: '931177' } ]`
- The SubjectGroup data model has been simplified. It now extends what the Subject data model does, instead of duplicating the functions already in the Subject

Note on the **Subject Group** resource:
- the backend presents the Subject Group resource VERY similarly to the Subject resource. The only way the front end can tell the difference is by looking into the metadata of the resource, or by keeping track of which endpoint the """Subject""" was retrieved from.

What's next?
- [ ] Updating tests

camallen:
>     * the backend presents the Subject Group resource VERY similarly to the Subject resource. The only way the front end can tell the difference is by looking into the metadata of the resource, or by keeping track of which endpoint the """Subject""" was retrieved from.

FWIW - we can add a serialized attribute to the 'Subject' resource indicating it's from a group instead of relying on metadata heuristics / inspection.

shaunanoordin:
## PR Update

And we're done here!

- Temporary workaround for the SubjectGroup API endpoint has been removed
- Tests ready

### Testing Guide

Package: `lib-classifier`
Testing URL: `https://local.zooniverse.org:8080/?env=staging&project=1900&workflow=3412`

External Requirements:
- Subject Group API needs to be alive (`https://panoptes-staging.zooniverse.org/api/subjects/grouped`)
- Workflow and Subject Set needs to be configured like Project 1900 Workflow 3412, linked above. For the record, it's...
  - Workflow setup: 
    ```
    workflow: {
      task: {
        T0: { help: "", type: "subjectGroupComparison", question: "Please select the cells that look weird."}
      }
      configuration: {
        subject_group: { num_rows: 5, num_columns: 5}  // Required for API endpoint
        subject_viewer: "subjectGroup",
        subject_viewer_config: { grid_rows: 5, cell_width: 200, cell_height: 200, grid_columns: 5 }  // Required for front end
      }
    }
    ```
  - Subject Set setup: at least 25 Subjects, each containing a single PNG or JPG image. Preferably of the same size.

Expectations:
(note: the features NEW to this PR will be marked with a 🛠️ )

- SubjectViewer: the classifier's Subject section should display a grid of 5x5 cells
  - The Subject Group Viewer should be interactive (can pan, zoom, be clicked on to annotate)
  - Accessibility: Subject Group Viewer should be interactive to both mouse and keyboard.
- TaskArea: the classifier's Task section should display a Subject Group Task... which is just a set of instructions and the Done/Next button.
- Annotation: when in annotation mode, clicking on an unselected cell should select it. (Indicated by a blue highlight on the cell). And vice versa
- 🛠️ **Network/New Subject Endpoint:** for a correctly configured Workflow, the classifier should fetch subjects from `/subjects/grouped`, instead of `/subjects/queued`
- 🛠️ **New "Subject type", Subject Group:** the Subject Group mimics a standard Subject with notable additional metadata. For example:
  ```
  subject: {
    locations: [ image/png: "https://panoptes-uploads-staging.zooniverse.org/subject_location/8aee8d2a-e309-472d-a4e7-c94d8dc45e04.png", ...24 more images... ],
    metadata: {
      #group_subject_ids: "134717-134726-134830-134744-134823-134765-134696-134699-134872-134884-134855-134664-134650-134740-134681-134834-134634-134713-134687-134720-134651-134604-134827-134825-134633"
      #subject_group_id: 190
    }
  }
  ```
- 🛠️ **Updated Subject Group Annotation:** when submitting the Classification, the annotations for the task should be an _array of marked cells,_ with each _marked call_ consisting of the index of the cell, and the underlying Subject ID of that cell. (Derived from the metadata) For example:
  ```
  annotations: {
    T0: [
      { index: 0, subject: '134717' }, { index: 5, subject: '134765' }
    ]
  }
  ```

### Status

Ready for review!

eatyourgreens:
Subject Group Viewer looks good to me in FF84. I can select frames and submit classifications. Zooming with the keyboard works when the toolbar or any frame has focus. Tabbing through the frames can be tedious: should those be marked up as a grouped control, like the survey task choices are in PFE? With a grouped control, like a tab list or radio button group or select menu, you tab into and out of the control, but use the cursor keys to navigate within the control.

Going to take a look at the code changes now.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewer.js
```diff
@@ -72,6 +73,7 @@ const SubjectGroupViewer = forwardRef(function SubjectGroupViewer(props, ref) {
                 
                 image={image}
                 index={index}
+                subjectId={subjectIds[index]}
```
eatyourgreens, line 76: I'm not seeing updated tests for this. Should it be tested?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -138,7 +138,7 @@ class SubjectGroupViewerContainer extends React.Component {
     if (subject && subject.locations) {
       // TODO: Validate for allowed image media mime types
       
-      const imageUrls = subject.locations.map(obj => Object.values(obj)[0])
+      let imageUrls = subject.locations.map(obj => Object.values(obj)[0])
```
eatyourgreens, line 141: We do this so often in our code (getting URLs from `subject.locations`) that we should probably write a helper for it at some point. Or write a view in the subject model that does it for us.

packages/lib-classifier/src/plugins/tasks/SubjectGroupComparisonTask/models/SubjectGroupComparisonAnnotation.js
```diff
@@ -1,15 +1,20 @@
...
 
-const SubjectGroup = types.model('SubjectGroup', {
-  value: types.optional(types.array(types.number), [])
+const SubjectIndexPair = types.model('SubjectIndexPair', {
+  index: types.number,
+  subject: types.string,
+})
+
+const SubjectGroupComparison = types.model('SubjectGroupComparison', {
+  value: types.optional(types.array(SubjectIndexPair), [])
```
eatyourgreens, line 10: I'm pretty sure `types.array` defaults to an empty array, so `types.optional` isn't needed here.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
```diff
@@ -88,13 +89,16 @@ function SGVGridCell (props) {
...
+    if (isThisCellSelected && !toggledValue) {  // Remove cell index from annotation values
+      const indexInValue = annotationValue.indexOf(isThisCellSelected)
       annotationValue.splice(indexInValue, 1)
-    } else if (!hasCellIndex && toggledValue) {  // Add cell index to annotation values
-      annotationValue.push(index)
+    } else if (!isThisCellSelected && toggledValue) {  // Add cell index to annotation values
+      annotationValue.push({
+        index,
+        subject: subjectId,
+      })
```
eatyourgreens, line 101: Is this tested? I'm not seeing changes to the SVG Grid Cell tests.

packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
```diff
@@ -2,76 +2,24 @@ import { getRoot, tryReference, types } from 'mobx-state-tree'
...
-    favorite: types.optional(types.boolean, false),
-    finished_workflow: types.optional(types.boolean, false),
-    metadata: types.frozen({}),
-    retired: types.optional(types.boolean, false),
-    selected_at: types.maybe(types.string),
-    selection_state: types.maybe(types.string),
-    shouldDiscuss: types.maybe(types.frozen()),
-    subjects: types.refinement(types.array(SingleImageSubject), validateSubjects),
-    user_has_finished_workflow: types.optional(types.boolean, false),
+    subjectIds: types.array(types.string),
```
eatyourgreens, line 9: 👍 So a subject group is a Subject model with an added `subjectIDs` array?

camallen:
>     * **New "Subject type", Subject Group:** the Subject Group (fetched from the new endpoint) should mimic a standard Subject with notable additional metadata. For example:

To be clear the grouped end point returns a 'group' Subject resource not a SubjectGroup resource. This special 'group' Subject resource is for consistency of the data model (classification is on a subject) and for use in talk UI code etc. 

Note the returned Subject is linked to the SubjectGroup resource available at the API via `GET /api/subject_groups/:id`

packages/lib-classifier/src/store/SubjectStore.js
```diff
@@ -18,7 +18,7 @@ const MINIMUM_QUEUE_SIZE = 3
 
 const SingleSubject = types.union(SingleImageSubject, SingleVideoSubject, Subject)
 function subjectDispatcher (snapshot) {
-  if (snapshot.subjects) {
+  if (snapshot.metadata['#subject_group_id']) {
```
eatyourgreens, line 21: It might be possible to write a test for this, by doing something like checking that `getType(subjects.active)` returns a `SubjectGroup` and not a `Subject`.

packages/lib-classifier/src/store/SubjectStore.js
```diff
@@ -152,17 +152,25 @@ const SubjectStore = types
...
+      
       if (workflow) {
         self.loadingState = asyncStates.loading
         const params = { workflow_id: workflow.id }
         if (workflow.grouped) {
           params.subject_set_id = workflow.subjectSetId
         }
+        
+        if (workflow.configuration.subject_viewer === 'subjectGroup') {
+          apiUrl = '/subjects/grouped'
```
eatyourgreens, line 165: Should we have a short ADR so that we don't forget that the API endpoint is now configurable?

eatyourgreens (commented):
Looks good to me. I wasn't sure if some of the changes should also be tested.

Should we have a short ADR to record that subject selection can now use different endpoints, depending on the needs of the workflow?

eatyourgreens (approved):

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
srallen: We should be validating the location here to check that it actually is an image. I've written helpers for this, check out the `DataImageViewerContainer` component, line 49 using `findLocationsByMediaType`.

shaunanoordin: I just got this down to...

```
const locations = findLocationsByMediaType(subject.locations, 'image') || []
const images = await Promise.all(
  locations.map(loc => this.fetchImage(Object.values(loc)[0]))
)
```

⚠️ ❗  ...but this made me realise it could lead to another problem. We're implicitly expecting the 'group' Subject resource from the backend to have exactly (say) 25 _image_ locations and a matching 25 _subject IDs._ (25 is an arbitrary number for illustrative purposes) If there is a mismatch in this - e.g. half the media locations are for videos/JSONs/txts - then we'd have 18 image and 25 subject IDs, causing a _complete mismatch in the image-subjectID pairings._

I'm going to need to think of a better error handling for incorrect/unexpected 'group' Subject Resources. For now, I'll keep the older doesn't-verify-for-media-type implementation since an incorrect media type just straight up **crashes** the Classifier, which is preferable to incorrect classifications, IMO.

packages/lib-classifier/src/plugins/tasks/SubjectGroupComparisonTask/models/SubjectGroupComparisonAnnotation.js
shaunanoordin: 👍

packages/lib-classifier/src/store/SubjectGroup/SubjectGroup.js
shaunanoordin: Yup, that's correct! With the current implementation of the API, there's only a very small difference between a 'standard' Subject and a 'group' Subject.

packages/lib-classifier/src/store/SubjectStore.js
shaunanoordin: This definitely needs a follow up ADR, and an expansion of the README now that the endpoint has more or less been settled.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewer.js
shaunanoordin: I've added tests to 'SGVGridCell' to not only test for SubjectIDs, but for SubjectIDs+cellIndex within the context of annotations.

shaunanoordin: I've also added (lighter) tests to the SubjectGroupViewer to ensure it's rendering those SGVGridCells properly 👍

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/components/SGVGridCell/SGVGridCell.js
shaunanoordin: Tests added! 👍

packages/lib-classifier/src/store/SubjectStore.js
shaunanoordin: I've added existence checks here, just to be safe. Following up with your suggestion for a getType() test.

shaunanoordin: Testes added 👍

shaunanoordin: TESTS ADDED, GODDAMMIT

shaunanoordin:
## PR Update

Much thanks to @eatyourgreens for feedback; this PR has been updated with...

- updated tests for SGVGridCell (to ensure annotations are updated according to events)
- updated tests for SubjectGroupViewer and SubjectStore
- a minor existence check for safety, in SubjectStore

Follow up:
- ADR for the switching between Subject queued vs grouped endpoints
- update README for SubjectGroup system

### Status

I'm going ahead and merging this; I'm looking at deploying this to staging to let users test this tomorrow (Fri 29 Jan) if there are no issues.

PR #1980, opened 2021-01-14
Video controller

### LINKED ISSUE #1880: Build Custom Controls for Video Player
**Package**
`lib-classifier`

The new Video Player needs custom controls for users to interact with the videos.
The video screen will be disabled and only used for annotating.
Features for the custom controls will include:

- Play/Pause
- Adjust playback speed
- Invert Video image
- Display Metadata
- Favorite
- Collect

ErikOstlund:
Package: `lib-classifier`

Closes #1880 

### Describe your changes:
This PR is to implement a custom controller for the video player.

Features for the controller include:

- Play/Pause
- Playback Speed Selection via a select element
- Slider: users can click and drag the slider handle or click anywhere on the slider to adjust the current video time
- Time Display showing the current video time and the total video length
- Accessibility
- Responsiveness

The controller will adjust accordingly to the current theme (Dark or Light).
Light theme:
image

Dark theme:
image

Playback Speed Selection
image

This can be tested by running yarn bootstrap to build the packages. Then cd packages/lib-classifier and run yarn dev.

### Items to consider during review:

- Do we need additional features in the video controller?
- @beckyrother Does the design need adjustments? Are the Dark & Light theme colors good?

This PR is part of ongoing work and part of the [Video Annotator](https://github.com/zooniverse/front-end-monorepo/projects/13) project.


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VideoController/Slider.js
```diff
@@ -0,0 +1,59 @@
...
+    },
+    thumb: {
+      color: '#EFF2F5'
+    }
+  }
+}
+
+const Slider = ({ played, onMouseUp, onMouseDown, onChange }) => (
+  <Grommet theme={customThemeRangeInput}>
+    <Box a11yTitle='Video player slider'>
```
eatyourgreens, line 29: This applies the accessible label to a `div`. Sectioning elements in the DOM won't generally be announced by a screenreader, except for named landmarks like `<header>`, `<nav>`, `<main>` and `<footer>`. I think you want to apply the accessible name to the range input itself, in which case it will be used as the label for that input.

Small wording change: maybe call it 'Video player scrubber' or similar, since the slider is being used as a video scrubber here.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VideoController/VideoController.js
```diff
@@ -0,0 +1,95 @@
...
+          onMouseDown={onSliderMouseDown}
+          onChange={onSliderChange}
+        />
+      </Box>
+
+      <Box direction='row' justify='between'>
+        <Box direction='row'>
+          <Box
+            a11yTitle='Video Play and Pause button'
+            alignSelf='center'
```
eatyourgreens, line 39: Similar comment to before. I think you want to label the button here, not its container. Since Play/Pause is a toggle, then I think the current state (playing or paused) needs to be announced in the accessibility tree too. Here's a great [overview of how play/pause state is announced](https://sarahmhigley.com/writing/playing-with-state/) in various screen reader/browser combinations. It seems you like you need a combination of `aria-label` and `aria-pressed` on the button.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VideoController/VideoController.js
```diff
@@ -0,0 +1,95 @@
...
+            pad={{ horizontal: 'small' }}
+          >
+            <Button onClick={onPlayPause}>
+              {isPlaying ? <PauseFill /> : <CirclePlay />}
+            </Button>
+          </Box>
+
+          <Box a11yTitle='Video playback speed selection' width='140px'>
```
eatyourgreens, line 47: Same comment here. You want to label the `select` control, not its container.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VideoController/VideoController.js
```diff
@@ -0,0 +1,95 @@
...
+              options={[0.25, 0.5, 1]}
+              value={playbackRate}
+              onChange={({ option }) => onSpeedChange(option)}
+              plain={true}
+            />
+          </Box>
+        </Box>
+
+        <Box
+          a11yTitle='Video player time played and total time'
```
eatyourgreens, line 58: I'm not sure you need the label here, but maybe mark this up as a live region since the content is constantly updating. I'm not 100% sure about that, since it might be really annoying for a blind volunteer to here the time announced every second as it changes.

Grommet has an [AnnounceContext](https://v2.grommet.io/announcecontext) component for creating live regions, and setting the politeness level.

eatyourgreens (commented):
Some suggestions about accessible controls for the player, if that's ok?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VideoController/Slider.js
eatyourgreens: Sorry, one other thought. You could move these accessible labels into locale files, in which case they could be made available in other languages at some future point.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VideoController/VideoController.js
eatyourgreens: The Grommet docs say that you can set the politeness level to 'polite', 'assertive' or 'off' but don't tell you what those different settings mean. Here's the [MDN docs for live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) (TLDR: use 'polite' unless it's really important for them to know the current video time every second.)

ErikOstlund:
@beckyrother Updated Video Player Controllers
The video time element was moved close to the Play button and the playback Select element was moved to the far right.
The playback Select element's width was adjusted to be more narrow. This is as narrow as I can get it before the selection option `0.25` gets cut-off.
The Slider color was updated to Gold (#F0B200).
image

image

beckyrother (approved):
lgtm! Thanks for making those changes

eatyourgreens:
@ErikOstlund code changes look good to me. I'll try this branch out in VoiceOver.

For a reference to check against, I made a recording of the native video player (QuickTime on my mac) with VoiceOver and Safari 12. For full compatibility with the native player, we might want the controls to be announced as a group and the slider to announce the time as the video progresses. I'll check how VO handles these custom controls.

https://user-images.githubusercontent.com/59547/105482358-2b8b3c00-5ca0-11eb-8851-191d48141af8.mov

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleVideoViewer/SingleVideoViewerContainer.js
```diff
@@ -74,7 +149,32 @@ class SingleVideoViewerContainer extends React.Component {
...
+      <div>
+        <div>
+          <SingleVideoViewer
+            playerRef={this.handlePlayerRef}
+            url={vid}
+            isPlaying={isPlaying}
+            playbackRate={playbackRate}
+            onProgress={this.handleVideoProgress}
+            onDuration={this.handleVideoDuration}
+          ></SingleVideoViewer>
```
eatyourgreens, line 162: If `SingleVideoViewer` doesn't have children, then this can be a single closed tag.

eatyourgreens:
One small issue. When the video ends, it stops but the play button is still in the playing state, and shows a pause button.
Screenshot of the video player, showing that the video has reached the end and stopped but the Pause button is still showing.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VideoController/Slider.js
```diff
@@ -0,0 +1,64 @@
...
+
+counterpart.registerTranslations('en', en)
+
+const Slider = ({ played, onMouseUp, onMouseDown, onChange }) => (
+  <Grommet theme={customThemeRangeInput}>
+    <Box>
+      <RangeInput
+        a11yTitle={counterpart('VideoController.scrubber')}
+        min={0}
+        max={1}
```
eatyourgreens, line 37: VoiceOver announces the slider value as a number between 0 and 1. Maybe it should be the elapsed time?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VideoController/VideoController.js
```diff
@@ -0,0 +1,111 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Box, Button, Select, ThemeContext } from 'grommet'
+import { CirclePlay, PauseFill } from 'grommet-icons'
+import counterpart from 'counterpart'
+import en from './locales/en.json'
+import FormattedTime from './FormattedTime'
+import Slider from './Slider'
```
eatyourgreens, line 8: Our convention is to put components into a `components` directory, so this would be `./components/FormattedTime` and `./components/Slider`.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VideoController/VideoController.js
```diff
@@ -0,0 +1,111 @@
...
+            <FormattedTime seconds={played * duration} />
+            {' / '}
+            <FormattedTime seconds={duration} />
+          </Box>
+        </Box>
+
+        <Box direction='row' alignSelf='center' pad={{ right: 'small' }}>
+          <ThemeContext.Extend value={customSelectTheme}>
+            <Select
+              a11yTitle={counterpart('VideoController.playbackSpeed')}
```
eatyourgreens, line 75: VoiceOver/Safari is announcing this menu as `Open Drop button` but I think that's a Grommet bug in the Select component. It completely ignores your label. 😞

eatyourgreens (changes requested):
I think the Play button state needs to be changed when the video ends. The controller should have a `components` directory, which can hold the `FormattedTimer` and `Slider` components.

VoiceOver is having trouble announcing the playback speed menu, but I don't think that's an issue with your code.

Ideally, each of these new components should have spec tests too. Are those planned for a future PR?

ErikOstlund:
@eatyourgreens I addressed the Play/Pause button state when a video ends. The `isPlaying` state is updated and the button now reverts back to the `Play` icon.
I also restructured the components into a `components` folder. Hopefully this helps with the VoiceOver issues you mentioned.
I do plan on adding `spec` test files for each component (and `stories` files), but in different PRs. I'm trying to keep things focused and in bite-size chunks for me and anyone who reviews these PRs.

eatyourgreens (approved):
Thanks! The play button is working as expected now, when the video stops.

The word 'button' in the label might be redundant. Here's how VoiceOver announces the play button. It's a button called 'Play video button'.
VoiceOver announcement: press Play video button, button

PR #1969, opened 2021-01-12
ZRC: Add Readme docs to React Components stories

eatyourgreens:
Storybook v6 deprecated `addon-notes` in favour of the Docs page for each component story. This PR replaces the `notes` parameter with the `docs` parameter for all the ZRC stories. This should add each component's Readme to its Docs page in the storybook.
Screenshot of the Favourites Button story, showing the component's Readme on its Docs page.

Package:
lib-react-components

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #1982, opened 2021-01-18
React Components: storybook structure

eatyourgreens:
Add a top-level Components category to the ZRC storybook. ~~Move the admin checkbox back under ZooFooter stories. Move the close button under Modal stories.~~

Follows on from #1969, where we noted that the sidebar structure for the ZRC storybook is a little confusing.

Package:
lib-react-components

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-react-components/src/CloseButton/CloseButton.stories.js
+++ b/packages/lib-react-components/src/CloseButton/CloseButton.stories.js
@@ -17,7 +17,7 @@ const config = {
 
 const darkZooTheme = { ...zooTheme, dark: true }
 
-storiesOf('CloseButton', module)
+storiesOf('Components/CloseButton', module)
   .addDecorator(withActions('click button'))
 
   .add('Light theme (default)', () => (
--- a/packages/lib-react-components/src/AdminCheckbox/AdminCheckbox.stories.js
+++ b/packages/lib-react-components/src/AdminCheckbox/AdminCheckbox.stories.js
@@ -15,7 +15,7 @@ const config = {
   }
 }
 
-storiesOf('AdminCheckbox', module)
+storiesOf('Components/AdminCheckbox', module)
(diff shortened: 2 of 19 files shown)


packages/lib-react-components/src/CloseButton/CloseButton.stories.js
```diff
@@ -17,7 +17,7 @@ const config = {
 
 const darkZooTheme = { ...zooTheme, dark: true }
 
-storiesOf('CloseButton', module)
+storiesOf('Components/Modal/CloseButton', module)
```
srallen, line 20: I don't think this should be nested under the `Modal` story. The `CloseButton` component is available and exported independently of it and therefore should have its own story. This here implies that the `CloseButton` is not and is simply a child component of the `Modal`

eatyourgreens: No problem. `Modal` is the only component that uses the close button, so I assumed it would be nested under `Modal`. Note that `AdminCheckbox` is exported independently but has always been listed as part of `ZooFooter` in our component hierarchy.

srallen: The way I'd correct that is to have the `AdminCheckbox` have its own story. `CloseButton` is used on its own in [our code](https://github.com/zooniverse/front-end-monorepo/blob/e0c725484953612b2c1f6808d8a5095a2fe0909c/packages/app-project/src/components/Announcements/components/GenericAnnouncement/components/WideGenericAnnouncement/WideGenericAnnouncement.js#L2).

eatyourgreens:
Updated so that all the exported components are listed under 'Components'. Only 'Tooltip' has other components nested under it.

mcbouslog (approved):
The storybook organization looks good to me, thanks!

PR #1953, opened 2020-12-18
Add an annotation prop to text tasks

eatyourgreens:
Add the annotation prop to text tasks, and use that to set the value of the text input box.

Package:
lib-classifier

Closes #1951.
Test URLs in the dev classifier:
Text task: https://localhost:8080/?project=908&workflow=3398
Transcription task: https://localhost:8080/?project=11300&env=production

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #1985, opened 2021-01-19
Classifier: use mark.isComplete for subtask validation

### LINKED ISSUE #1986: Drawing and transcription: Creating an incomplete mark makes all your marks incomplete.
Creating an incomplete mark for drawing or transcription tasks makes all your marks incomplete.

1. Make a mark
2. Answer the question and save.
3. Make a new mark.
4. Close without saving.
5. Open your previous mark. It's now incomplete. Closing the dialog will delete it.

The underlying cause is that deleting an incomplete mark sets `task.annotation` to undefined for all marks/lines on a page. Neither the drawing task nor the transcription task will let you save a line that has no annotation(s).

Screenshot of a drawing task with a required question. The question has been answered but you are prompted to close without saving, as if the answer doesn't exist.

_Originally posted by @eatyourgreens in https://github.com/zooniverse/front-end-monorepo/issues/1952#issuecomment-762178339_

eatyourgreens:
Use `mark.isComplete`, not `task.isComplete`, when validating drawing and transcription subtasks. Update `Mark.isComplete` to use `tasks.some`.

Test URLs:
Drawing task: https://localhost:8080/?project=908&workflow=3370
Transcription task: https://localhost:8080/?project=11300&env=production

Package:
lib-classifier

Closes #1986.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js
@@ -47,8 +47,7 @@ function SubTaskPopup(props) {
   }
 
   function close() {
-    const incompleteTask = tasks.some(task => !task.isComplete)
-    if (incompleteTask) {
+    if (!activeMark.isComplete) {
       onOpenConfirm()
     } else {
       setSubTaskVisibility(false)
--- a/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
@@ -61,14 +61,11 @@ const BaseMark = types.model('BaseMark', {
     },
 
     get isComplete () {
-      // Return false after the first incomplete task, true otherwise.
-      let isMarkComplete = true
-      self.tasks.forEach(task => {
-        if (isMarkComplete) {
-          isMarkComplete = !task.required || !!self.annotation(task)?.isComplete
-        }
+      const incomplete = self.tasks.some(task => {
+        const taskAnnotation = self.annotation(task)
+        return task.required && !taskAnnotation?.isComplete
       })
-      return isMarkComplete
+      return !incomplete
     },
 
     get isValid () {


srallen (commented):
I'm trying to follow the steps to recreate the bug outlined in the issue and I cannot because the text task is throwing an error causing the classifier to crash, both on this branch and on master branch:

```
Uncaught Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree. (Object type: 'TextAnnotation', Path upon death: '/workflowSteps/steps/S1/tasks/0/tools/0/marks/ckk4b1kni000p3h6apkiv7gbh/annotations/ckk4b1knp000q3h6asm1ueuvc', Subpath: '', Action: '/workflowSteps/steps/S1/tasks/0/tools/0/marks/ckk4b1kni000p3h6apkiv7gbh/annotations/ckk4b1knp000q3h6asm1ueuvc.update()'). Either detach nodes first, or don't use objects after removing / replacing them in the tree.
```

This MST error is present on staging as well (https://frontend.preview.zooniverse.org/projects/blicksam/transcription-task-testing/classify), though, it doesn't crash the entire classifier in the project app context. I checked out `afd02d1ac49618e651621eaee2ab64e33fb4f6ee` and could not replicate the regression, so I think it is reasonable to think that this was introduced with the removal of the effect from #1953. This also means we're missing specs that would fail with the effect removal. I understand it might not be easy to write specs to test this, so we should add at least an inline explanation of the effect there since it is different from the other tasks. 

Second, I'd expect a spec for this particular bug fix, either updating existing ones or a new one testing what this is fixing. Is it possible to add or update a spec?

eatyourgreens:
I noticed that the pop up tests only use a single mark. Adding a second mark to the tests might catch this bug. I’m happy to have a look at that.

#1953 introduced an MST error, which we fixed by removing `useEffect`. In that case, the effect was running after the annotation had been removed from the store.

srallen:
I believe the bug I found is fixed by #1988

srallen (approved):
> I noticed that the pop up tests only use a single mark. Adding a second mark to the tests might catch this bug. I’m happy to have a look at that.

If you have time to update the specs, then that would be great. Otherwise, let's track it in an issue? 

Otherwise, I can now confirm that the issue is resolved following the testing steps outlined in #1986 for the two test projects linked in the PR. Thanks!

eatyourgreens:
I got as far as sketching out in my head how to test this, by modifying one of the existing tests. I think there's a test that creates an incomplete mark, then checks that the confirm dialog shows (and deletes the mark.) If we add an existing completed mark to that test, then we can also check that the completed mark is not changed to incomplete, and does not open the confirm dialog when edited. I think that would work, but it's just an idea in my head at the moment.

eatyourgreens:
I'll merge this so that it doesn't hold up the transcription work.

Issue #1978, opened 2021-01-13
Transcription Task Beta Feedback


snblickhan:
- [x] 'Back' button is missing -- once you've hit next you can't go back and edit a transcription.
- [ ] Can't zoom out past a full-size page. This is helpful, to be able to see more of the page w/out having to scroll (ASM had this feature)
- [x] Need a 'hide previous marks' feature (use case: previous mark is covering text/punctuation & affecting accuracy of transcription)
- [x] Cursor should automatically move to the text box after the second underline dot is placed, so you can just start typing — really annoying to have to click into the box every time.
- [ ] Page scroll position automatically reverts to the top of the page after an annotation is submitted, meaning when you're working on text at the bottom of the page, you keep having to scroll back down for every new annotation. If possible, this automatic move should be disabled.
- [x] When Previous Transcriptions box is open, mouseover scroll on the Previous Transcription box also causes the image to Zoom in (while the transcriptions box is open, and is also scrolling)
- [ ] The "You've Seen Everything" banner doesn't appear automatically -- e.g. I classified all 5 subjects in the set, and was given one again with newly-greyed-out lines but without the "Finished" banner. Banner appears if I refresh the page.
- [x] POSSIBLE BUG: noticed that when I add a transcription for a given line, all the previous (navy blue) transcriptions are changing to the most recent text I’ve input. e.g. I transcribe a line "Pop goes the weasel" and move on to the next one, and transcribe "Mary had a little lamb." If I click back onto the weasel line, it now reads "Mary had a little lamb."

eatyourgreens:
The second bug is #1951. #1953 should fix it.

srallen:
Two more issues, demonstrated in this screenshot:

<img width="1342" alt="Screen Shot 2021-01-20 at 10 56 27 AM" src="https://user-images.githubusercontent.com/5016731/105209172-28147b00-5b0f-11eb-90a7-7219c89b158e.png">

1) The previous transcription marks should be rendered on the second step to be able to answer the question
2) Back button is not rendering when it should

@snblickhan are either of these blockers for continuing ahead with beta? I'll start looking into both of these regardless.

snblickhan:
> The previous transcription marks should be rendered on the second step to be able to answer the question

Can you explain what this means? I'm not entirely sure what the issue is.

srallen:
[...]

PR #2001, opened 2021-01-22
Fix previous marks from caesar render for transcription task

srallen:

Package: lib-classifier

Toward #1978, fixes the previous marks render described in this comment:  https://github.com/zooniverse/front-end-monorepo/issues/1978#issuecomment-763791441

You can test by running the dev classifier with https://localhost:8080/?project=11300&env=production

The previous transcription marks loaded from the caesar reductions are now viewable throughout the workflow tasks and this enables us to answer the question from the second step question task:

Screen Shot 2021-01-22 at 5 05 56 PM

The hide previous marks button is always available too:

Screen Shot 2021-01-22 at 5 06 14 PM

Compare this to another project with a drawing task (https://localhost:8080/?project=908&workflow=3370), the hide previous marks button is only available on the step with drawing tasks.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
There's a minor styling bug introduced with this PR. On a drawing task with marks from a previous step, the previous marks show the hand cursor on hover. They can't be selected, so it's only the styling that's broken. 

Screenshot of a drawing task showing the hand cursor over a mark that can't be selected.

eatyourgreens:
Hide/Show previous marks is broken on a workflow with two drawing steps.
1. Hide previous marks on the second step.
2. Press Back.
3. Previous marks are rendered again for the first step.
4. Press Next.
5. Previous marks are hidden again, and can be revealed with the Show Previous Marks button.

I can reproduce that behaviour on master, so it isn't a blocker here.

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -47,6 +47,27 @@ const WorkflowStepStore = types
       }
 
       return false
+    },
+
+    findTasksByType (type) {
+      const tasksByType = Array.from(self.steps).map(([stepKey, step]) => {
+        if (step?.tasks) {
+          return Object.values(step.tasks).filter(task => {
```
eatyourgreens, line 55: `step.tasks` is an array so I think you can filter that directly here.

packages/lib-classifier/src/store/WorkflowStepStore.js
```diff
@@ -47,6 +47,27 @@ const WorkflowStepStore = types
...
+          })
+        }
+        
+        return []
+      })
+
+      return tasksByType.flat()
+    },
+
+    get interactionTask () {
```
eatyourgreens, line 66: Can we name this `markingTask`? I find the generic name confusing because all tasks involve some sort of interaction (choosing an answer, making a mark etc.)

eatyourgreens (approved):
Code changes look good to me. There's a couple of places where I thought the tests were hard to understand. There's a styling bug introduced here for previous marks from drawing tasks. I'm not sure if you want to fix that here, or open a separate issue?

The styling bug shouldn't be a blocker for the transcription project beta test.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js
```diff
@@ -87,6 +77,7 @@ class InteractionLayerContainer extends Component {
             width={width}
           />
         }
+        <PreviousMarks scale={scale} />
```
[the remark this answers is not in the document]
srallen: Yes, this is intentional because when someone makes new marks from previous transcriptions from Caesar and goes to the next question step, if it isn't rendered on top, they no longer can see that they made a new mark and see the mark from caesar on top instead.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/PreviousMarks/PreviousMarksConnector.spec.js
```diff
@@ -0,0 +1,179 @@
...
+
+  describe('when there is an interactive task', function () {
+    it('should pass along the type of shownMarks from the drawing task', function () {
+      mockUseContext = sinon.stub(React, 'useContext').callsFake(() => {
+        return {
+          classifierStore: mockStoresWithDrawingAnnotations
+        }
+      })
+      const wrapper = shallow(<PreviousMarksConnector />)
+      expect(wrapper.find(PreviousMarks).props().shownMarks).to.equal('USER')
```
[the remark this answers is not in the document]
srallen: Not for drawing tasks, but it is for transcription tasks. The `PreviousMarks` component, it is valid and needed for transcription tasks. It's impossible for the value to be set as `USER` during a drawing task since the correct binary toggle will be rendered for the control.

packages/lib-classifier/src/store/WorkflowStepStore.js
srallen: It's `interactionTask` because it's being used by the `InteractionLayer`.

srallen:
> Should I expect any of the purple and blue lines to turn grey?

Not yet, the grey state is the retired state from Caesar. We need more classifications to get them to turn grey.

packages/lib-classifier/src/store/WorkflowStepStore.js
srallen: Ah, yes it's an object on the workflow.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/PreviousMarks/PreviousMarks.js
```diff
@@ -0,0 +1,37 @@
...
+  } = props
+  const marksToShow = shownMarks === SHOWN_MARKS.ALL || shownMarks === SHOWN_MARKS.USER
+
+  if (interactionTaskAnnotations?.length > 0 && marksToShow) {
+    // Wrapping the array in an react fragment because enzyme errors otherwise 
+    // React v16 allows this, though. 
+    return (
+      <>
+        {interactionTaskAnnotations.map((annotation) => {
+          const annotationValuesPerFrame = annotation.value.filter(value => value.frame === frame)
```
goplayoutside3, line 21: @eatyourgreens this is a long-ago PR, but I have a question particularly about this line introduced here, which is still used in production. `annotationValuesPerFrame` is intended to filter marks based on the `frame` passed to PreviousMarks component. However, Travis and I have found that's not always the case and are unsure how long the following bug has been around ~

On projects with multi-frame subjects and more than one workflow step, previous annotations do not filter correctly when switching between frames. For instance, [How Did We Get Here](https://www.zooniverse.org/projects/communitiesandcrowds/how-did-we-get-here?demo=true) is a transcription project with drawing as the first step and a yes/no as the second step. If I draw marks on all the frames in the first step, while on the 2nd step navigating between frames renders marks on only _one_ frame. I think the expected behavior is that previous marks should always render on their respective frames. Any thoughts on how to fix this filter?

(More dicussion [here](https://github.com/zooniverse/front-end-monorepo/pull/5913#issuecomment-1945219415)).

eatyourgreens: I can confirm the bug. Here, I've added two lines to frame 0 and one line to frame 1. `classification.annotations` only contains one line, not three.

Are annotations being lost on live projects that transcribe more than one page per subject?

Screenshot of the dev classifier for How Did We Get Here. The console is open to show that classification.annotations only contains one transcribed line.

eatyourgreens: When I go back to the drawing step, my drawn lines are there. They haven't been added to the classification. So my thought would be that the bug is in the code that adds mark IDs to the current classification.

Screenshot of the dev classifier for How Did We Get Here. There are two drawn lines on the page, but neither are stored in classification.annotations yet.

eatyourgreens: If I then add a third mark to frame 0, `classification.annotations` updates to show the three lines from that frame, but loses the line from frame 1.

Screenshot of the dev classifier for How Did We Get Here. I have added a third line to the first frame of the active subject, and `classification.annotations` now shows the three mark IDs for those lines, but has lost the ID of the fourth line (on the second frame.)

eatyourgreens: Based on that, I would say that `previousAnnotations` is working as expected, but the classification isn't being updated correctly when you draw on more than one frame of a subject. `classification.annotations` only contains mark IDs for the active frame. I'd expect it to contain the mark IDs for all the transcription lines that I've made.

Since this bug is in production, and there are projects with subjects that have more than one page, I would definitely double check that classifications are being recorded properly when a volunteer adds lines to more than one page.

eatyourgreens: I've suggested a fix for the `markIDs` array in https://github.com/zooniverse/front-end-monorepo/pull/5919.

eatyourgreens: Here is the bug, by the way. `marks` here is filtered by frame before being added to the classification in `InteractionLayer`.
https://github.com/zooniverse/front-end-monorepo/blob/6b421edae1e92322bf3676cc558612e886983321/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js#L92

eatyourgreens: #5923 is a more robust fix, replacing the manual update with an autorun reaction whenever the task’s marks change.

PR #2011, opened 2021-02-01
Classifier: Fix duplicate marks from the back button with drawing tasks

### LINKED ISSUE #1869: Classifier: Drawing task duplicates marks on previous steps
**Package**
lib-classifier

**Describe the bug**
1. Mark a couple of red points and cyan points for the first step
2. Press Next
3. Press Back to return to the first step

You should find that you can select and edit your marks, but each mark is duplicated when you select it. In the following screenshot, you can see a large, selected cyan point on top of an inactive mark. Both are the same mark. The classification annotations aren't duplicated, so this seems to be a case of the task's marks being rendered twice: once as inactive marks from a previous step, and once as active marks for the current step. 

A screenshot of the drawing task, where one of the drawn point marks has been rendered twice.

**Expected behavior**
I think the drawing layer should be rendering marks from previous steps (just the views with no event handlers) here:
https://github.com/zooniverse/front-end-monorepo/blob/ff9276c03fdc8cfd04d784ff95b87be052deac2a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js#L64-L72

then marks for the current task should be rendered here (with event handlers for editing)
https://github.com/zooniverse/front-end-monorepo/blob/ff9276c03fdc8cfd04d784ff95b87be052deac2a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.js#L74-L88

I think that perhaps the filtering of marks, so that the current task annotations aren't passed to the `DrawingToolMarks` component, might not be working. I'm not sure.

I've only tested this with the point task but that workflow has 3 steps:
- point drawing task
- line drawing task
- single answer question
so that can be used to test the line tool too.

Tested in Firefox 82 on macOS 10.14 but I don't think this is browser specific.

eatyourgreens:
Package:
lib-classifier

Closes #1869.

Move `PreviousMarks` out of the drawing interaction layer. Pass annotations from previous drawing steps to the `PreviousMarks` component. Marks from the active drawing task are already rendered by the `InteractionLayer` component, and shouldn't be rendered twice.

Testing URLs:
Drawing: https://local.zooniverse.org:8080/?project=908&workflow=3370
Transcription: https://localhost:8080/?project=11300&workflow=13898&env=production

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR fixes a presentational issue on the Classifier for Drawing Tasks.

- The issue:
  - if task 1 is a Drawing Task in a multi-task workflow, then we expect to see the "previous annotations" displayed while we look at task 2 onwards. ("Previous annotations" are inert marks, different from "current annotations" which are interact-able marks)
  - however, if we _go back to task 1 from task 2,_ then "current annotations" appear ALONG WITH the "previous annotations"; essentially duplicating the number of marks (the inert marks underlaid directly beneath the interact-able marks) on screen.
- The fix:
  - `PreviousMarksConnector` now makes a distinction between "interactionTaskAnnotations" (which lists all drawing-style annotation tasks) and the new "previousTaskAnnotations" (which is a subset of the previous annotations, containing all annotations EXCEPT those from the current active task)
  - [minor] Side note: theoretically this could mean previousTaskAnnotations could also contain annotations of _future_ tasks, but since we haven't built a "preserve annotations of Task N+1 if we go back to Task N" function _nor a flying DeLorean,_ the shouldn't be an issue)
  - `<PreviousMarks />` now moved out of InteractionLayer and into the viewer components

Testing looks good, code looks legit. 👍 

### Testing

Tested with macOS10+Chrome88 on `https://local.zooniverse.org:8080/?env=staging&project=1651&workflow=3457`

Setup:

- Workflow 3457 is a temporary/quick-test workflow with the following configuration:
  - Task 1: Drawing task (Point tool, Point tool)
  - Task 2: Text task

Testing:

I'm able to...
- ...make 5 marks (all of which are interact-able) in Task 1, then move to the next task.
- ...see all 5 previous marks (now inert, non-interactive) while answering Task 2.
- ...go back to Task 1, and see ONLY 5 marks (all of which are once again active and interact-able). No duplicate inert marks are visible. 👍 
- ...then proceed to move on to Task 2 again, and then submit.

Also...
- when in Task 1, I'm able to use the Drawing Task normally. I'm able to add new marks, and edit/delete existing marks. 👍 
- when going back from Task 2 to Task 1, if there was an active mark (i.e. currently selected, e.g. for editing) in Task 1, that mark immediately becomes active again. This has no deleterious effect, though I suspect that if the active mark has a subtask, this means the SubTask component will pop up as soon as you traverse back to Task 1, which might cause a very minor UI/UX whoopsie. I _could_ test this but this seems too small to worry about. 

### Status

LGTM 👍 This fixes the bug, and the primary functionality of Drawing Tasks seem to work with no issues.

[bonus] Also tagging @ErikOstlund as an FYI since I know he's working on the video-drawing annotations.

_Erik, TL;DR: the <PreviousMarks /> has been moved out of the pluggable InteractionLayer (the drawing layer) to the containing component. In this PR's case, that's SingleImageViewer. In your case, you might need to explicitly add <PreviousMarks /> to your SingleVideoViewer if you're implementing the drawing layer._

srallen:
Could merging of this be held off until we can confirm this does not cause regressions with the transcription task? Thanks.

eatyourgreens:
Sure. I think the bug itself is cosmetic. Lines are duplicated on screen if you go back, but annotations aren’t changed, so I don’t think there’s a hurry to merge this. 

I couldn’t test the bug on the transcription project, because the back button is broken from the second task. If you could go back, there might be some weirdness with lines being rendered twice.

eatyourgreens:
> This has no deleterious effect, though I suspect that if the active mark has a subtask, this means the SubTask component will pop up as soon as you traverse back to Task 1, which might cause a very minor UI/UX whoopsie. 

I’ve had this happen on my test project. Clearing the active mark, when exiting or entering a task, might fix it. It is very annoying.

srallen (commented):
Organizationally I strongly dislike the change of moving `PreviousMarks` out of the `InteractionLayer`. What this means is that there's drawing/transcription task related code that renders along with the `SingleImageViewer` regardless of whether that kind of task is enabled or not. 

Secondly, I think this could be resolved by refactoring some of the computed views I wrote for #2001. I've opened #2013 as a proposed alternative fix and would appreciate a consideration.

eatyourgreens:
>there's drawing/transcription task related code that renders along with the SingleImageViewer regardless of whether that kind of interaction is enabled or not.

This is exactly what we want. Static drawing marks from previous steps should be shown for any task type. For example, a question task after a drawing or transcription step. PFE rendered static marks and interactive marks together in the same component. We agreed not to repeat that in the new classifier, because it lead to bugs. See [ADR-15](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-15.md).

eatyourgreens:
I've changed the name of this because the bug exists on workflows with a single drawing or transcription task. Go back to a previous task from any other task and your drawing/transcription marks will be rendered twice: once by `InteractionLayer` and once by `PreviousMarks`.

I think the code must clearly distinguish between drawing annotations that are being edited and updated (`InteractionLayer`) and annotations that are displayed but can't be altered (`PreviousMarks`.) I'm nervous about putting both of those into the `InteractionLayerContainer` because it makes that logic harder to follow, for me at least.

srallen:
@eatyourgreens Hard disagree that this is what we want and the interpretation of that ADR. This now makes the SingleImageViewer render drawing/transcription task related code regardless of it is enabled. An example of what I mean is that in the composite VariableStarViewer, we render a single image that will _never_ use the drawing task. This change will make it so that SingleImageViewer within that composite viewer runs code that will never be needed. 

I agree that the InteractionLayer could use refactoring, and there's a lot I would do with it, but I would strongly prefer encapsulation of the drawing/transcription task related mark rendering code so that it is easier to understand how it works together and scope appropriately.

eatyourgreens:
If there are no drawing tasks in the workflow, then there are no drawing annotations in `classification.annotations`. When there are no drawing annotations, `<PreviousMarks />` should render nothing. I'm not really following why that causes problems for the variable star viewer. If it does, it would indicate to me that there's a bug in the code for rendering marks.

My concern here is that we are painting ourselves into the same corner as in PFE, where code for editing and updating drawings is also responsible for rendering static marks. Surely separating them makes them easier to test and less likely to introduce bugs? That was the original intention of our discussion about drawing tools, in ADR-15.

srallen:
> My concern here is that we are painting ourselves into the same corner as in PFE, where code for editing and updating drawings is also responsible for rendering static marks.

I agree, but I don't think the solution is moving PreviousMarks out of the InteractionLayer. There's refactors that can be done within the InteractionLayer code still to scope responsibilities to individual components.

eatyourgreens:
True, and `InteractionLayer` itself is no longer responsible for handling pointer events. It's gone from being the drawing canvas, to being more of a container component responsible for rendering drawing annotations, including transcription lines contributed by other volunteers.

PR #2013, opened 2021-02-02
Refactor PreviousAnnotations and associated utility/computed view functions

### LINKED ISSUE #1869: Classifier: Drawing task duplicates marks on previous steps

srallen:

Package: lib-classifier

Closes #1869. Alternative proposal for a fix for #2011.

Testing URLs:
Drawing: https://local.zooniverse.org:8080/?project=908&workflow=3370
Transcription: https://localhost:8080/?project=11300&workflow=13898&env=production

I disagree with the change to move `PreviousMarks` out of the context of the `InteractionLayer`. This means code related to the drawing/transcription task will always run whether the interaction layer is enabled or not. In addition, the computed view I added in #2001 to retrieve the related annotations should be refactored to how it's actually needed. It's not used anywhere else, so let's update it to filter by the basis of the task's step's activeness. It takes a parameter now expecting to filter on the active task key. 

The `WorkflowStepsStore` now has two utility computed views for `interactionTask` which will retrieve an active drawing task and a transcription task regardless of step activeness. The second is `activeInteractionTask` which returns the drawing or transcription task only if the step is active. I think this can be cleaned up a bit, but keeps the solution for #1869 with the available utility functions. 

I removed some cruft and added some missing prop type definitions also.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/store/Classification/Classification.spec.js
```diff
@@ -132,31 +133,50 @@ describe('Model > Classification', function () {
...
-      expect(model.interactionTaskAnnotations).to.be.empty()
+      const previousAnnotations = model.previousInteractionTaskAnnotations('T1')
+      expect(previousAnnotations).to.be.an('array')
+      expect(previousAnnotations).to.be.empty()
       model.removeAnnotation('T0')
       model.removeAnnotation('T1')
     })
 
-    it('should return an array of annotations if the stored annotations are for drawing tasks', function () {
+    it('should return an empty array if the active task key is the same as the annotations task key', function () {
```
eatyourgreens, line 153: This isn't quite true. It should return an empty array if all of our annotations are for the current task. I'm not quite sure how to word that.

eatyourgreens (approved):
Changes look good to me but I found the tests a little hard to follow.

The bug in #1869 can be triggered by building a workflow with a single drawing or interaction task, followed by a question task. To reproduce the bug, draw some marks then go back from the question task to the drawing/transcription task. Are we testing that case here? I found the tests a bit confusing to read.

eatyourgreens:
I think I'd find the tests easier to understand if they described the behaviour of drawing task annotations in different contexts. Something like this, for one mock workflow:
```js
describe('in a workflow with a drawing task followed by any other task`, function () {
  describe('the drawing task', function () {
    it('should not render its own annotations as previous marks.', function () {
      …
    })
  })
  describe('the second task', function () {
    it('should render the drawing task's annotations as previous marks.', function () {
      …
    })
  })
})
```

packages/lib-classifier/src/store/Classification/Classification.spec.js
srallen: That's what I think I have worded here, but I can try to reword it.

srallen:
> Are we testing that case here?

Yes we are, with the classification store tests. I can reword for better clarity. It can also be manually demonstrated as fixed with your own test project.

eatyourgreens:
Sorry, my test project has two drawing tasks. I completely failed to test it by going back in a workflow with a single drawing task. I think the tests use a case with two drawing tasks too. I think it should be fine, but I haven't tested the transcription project as the back button is broken there.

shaunanoordin:
## PR Quick Review

Package: lib-classifier

I'm running tests similar to [the ones I ran for 2011](https://github.com/zooniverse/front-end-monorepo/pull/2011#pullrequestreview-580776297). Here's what I've got so far:

Tested with macOS10+Chrome88 on `https://local.zooniverse.org:8080/?env=staging&project=1651&workflow=3457`

Setup:
- Workflow 3457 is a temporary/quick-test workflow with the following configuration:
  - Task T1: Drawing task (Point tool, Point tool)
  - Task T2: Text task

Testing Actions:
- **(step 1)** on Task T1, draw 3 marks.
  - Expectation: I should be able to add, edit, and remove 3 marks.
- **(step 2)** proceed to Task T2.
  - Expectation: While working on Task 2, I should be able to see the 3 _previous marks_ created in (step 1)
  - Expectation: those 3 marks should be _inert_
- **(step 3)** go back to Task T1
  - Expectation: I should ONLY see those 3 marks I created in (step 1)
- **(optional step 4)** edit marks, removing one and adding two more to make 4 marks. 
  - Expectation: marks should be editable, etc etc
- **(step 5)** proceed to Task T2, then submit.
  - Expectation: classification should be submitted with the correct annotations.

Results, for this PR:
- (step 1) works fine
- ⚠️ in (step 2) however, the previous marks are _invisible_
- (step 3) works fine, and confirms that the previous marks are _still recorded_ in the annotations.
  - I'm noting a console warning: `Did you mean to call previousInteractionTaskAnnotations without an active task key? / at PreviousMarksConnector ...`
- (step 4) and (step 5) confirm that the annotations are _correctly_ submitted.

I think this means that the marks are only visible for the active drawing task.

❓ ** Question:** is this planned new behaviour, or is that console warning more important that I thought?

Comparison with `master` :
- (step 1) and (step 2) work fine
- (step 3) shows the error described in 1869
- (step 4) and (step 5) work fine

shaunanoordin:
Wait, whoop, I took so long to type my review there's a new commit. Updating and retesting...

shaunanoordin:
OK, yup, as of b15bb07 I'm still seeing the oddity where the previous marks for Task T1 don't render when I'm working on Task T2. I think that React warning may be on to something.

@srallen if you want to test this yourself, I'm going to leave ?env=staging&project=1651&workflow=3457 locked to the aforementioned config until tomorrow; after which 3457 will once again be my ever-mutating guinea pig of a workflow. 🔬

eatyourgreens:
Maybe I was too fast in closing and deleting my other PR.

srallen:
@shaunanoordin thanks Shaun. Sounds like that warning is working as intended and I missed something. I'll have a look later in the afternoon and I'll dismiss @eatyourgreens incomplete review.

packages/lib-classifier/src/store/Classification/Classification.js
```diff
@@ -27,9 +27,15 @@ const Classification = types
...
-      return annotations.filter(annotation => (getType(annotation).name === 'DrawingAnnotation' || getType(annotation).name === 'TranscriptionAnnotation'))
+    previousInteractionTaskAnnotations (activeTaskKey) {
+      if (activeTaskKey) {
+        const annotations = Array.from(self.annotations.values()) || []
+        const interactionTaskAnnotations = annotations.filter(annotation => (getType(annotation).name === 'DrawingAnnotation' || getType(annotation).name === 'TranscriptionAnnotation'))
+        return interactionTaskAnnotations.filter(annotation => annotation.task !== activeTaskKey) 
+      }
+
+      if (console) console.warn('Did you mean to call previousInteractionTaskAnnotations without an active task key?')
+      return []
```
eatyourgreens, line 38: This will remove previously drawn marks on any step that doesn’t include a drawing task or transcription task.

eatyourgreens (changes requested):

srallen:
@shaunanoordin and @eatyourgreens this is ready for review again. I confused myself about whether the parameter should be required or not when fixing a typo and left a warning for myself to catch mistakes, so worked as intended. Thanks.

eatyourgreens:
@shaunanoordin feel free to dismiss my review if you get to this before I do.

packages/lib-classifier/src/store/Classification/Classification.spec.js
```diff
@@ -132,35 +149,86 @@ describe('Model > Classification', function () {
...
+    })
+
+    describe('when the annotations are from the current transcription task', function () {
+      it('should return an empty array', function () {
+        const transcriptionMark = TranscriptionLine.create({ id: 'transcriptionline1', frame: 0, toolIndex: 0, x1: 100, y1: 200, x2: 150, y2: 200, toolType: 'transcriptionLine' })
+        model.addAnnotation(transcription, [transcriptionMark])
+        model.addAnnotation(singleChoice, 0)
+        const previousAnnotations = model.previousInteractionTaskAnnotations(transcription.taskKey)
+        expect(previousAnnotations).to.be.an('array')
+        expect(previousAnnotations).to.have.empty()
```
eatyourgreens, line 218: typo?
```suggestion
        expect(previousAnnotations).to.be.empty()
```

packages/lib-classifier/src/store/Classification/Classification.spec.js
```diff
@@ -132,35 +149,86 @@ describe('Model > Classification', function () {
...
-      expect(annotations).to.have.lengthOf(1)
-      expect(annotations[0].task).to.deep.equal(transcription.taskKey)
-      model.removeAnnotation('T3')
+    describe('when the annotations are from the current drawing task', function () {
+      it('should return an empty array', function () {
+        const pointMark = Point.create({ id: 'point1', frame: 0, toolIndex: 0, x: 100, y: 200, toolType: 'point' })
+        model.addAnnotation(drawingOne, [pointMark])
+        const previousAnnotations = model.previousInteractionTaskAnnotations(drawingOne.taskKey)
+        expect(previousAnnotations).to.be.an('array')
+        expect(previousAnnotations).to.have.empty()
```
eatyourgreens, line 193:
```suggestion
        expect(previousAnnotations).to.be.empty()
```

eatyourgreens (approved):
A couple of typos in the tests. Otherwise, looks good!

shaunanoordin (approved):
## PR Review

package: `lib-classifier`
supersedes #2011

This PR fixes a presentational issue on the Classifier for Drawing Tasks.

- The issue:
  - if task 1 is a Drawing Task in a multi-task workflow, then we expect to see the "previous annotations" displayed while we look at task 2 onwards. ("Previous annotations" are inert marks, different from "current annotations" which are interact-able marks)
  - however, if we _go back to task 1 from task 2,_ then "current annotations" appear ALONG WITH the "previous annotations"; essentially duplicating the number of marks (the inert marks underlaid directly beneath the interact-able marks) on screen.
- The fix:
  - `Classification` now makes a distinction between listing _all interaction (drawing) tasks,_ and listing _all interaction (drawing) tasks EXCEPT for the current one, if the current task is a drawing task._ (i.e. Classification. previousInteractionTaskAnnotations() replace Classification.interactionTaskAnnotations)
  - [minor] side note: in theory this could mean 'previousInteractionTaskAnnotations()' could record annotations from _future tasks,_ but since the classifier doesn't keep annotations for task N+1 when we go Back to task N, this ain't an issue.

Testing looks good, code looks legit. 👍 

### Testing

See https://github.com/zooniverse/front-end-monorepo/pull/2013#issuecomment-771863898 for testing setup.

This PR meets all expectations based on the manual test 👌 

### Status

Giving this PR an additional Functionality Thumbs Up on top of Jim's pre-existing approval. 👍

PR #2015, opened 2021-02-03
Classifier: Trap scroll wheel events on drawing tools popups

### LINKED ISSUE #1835: Image zoom by mouse wheel/touch event still functional while sub-task is open
**Package** lib-classifier

**Describe the bug**

From transcription task alpha https://github.com/zooniverse/front-end-monorepo/issues/1719#issuecomment-662751492

> I often use a two-finger scroll (equivalent to mouse zoom wheel) to scroll a page. Within the subject window, that two-finger scroll is always interpreted as zoom-in/zoom-out -- even for the case where the subtask modal has a scroll bar, as is the case in the screen shot below.

**To Reproduce**
Open up drawing sub-task with an image subject. Try to scroll the subtask, it zooms instead.

**Expected behavior**

Either we make sure zoom is disabled or don't make the pop up be a true modal and allow interaction below it and make sure scroll events are correctly z-index.

**Screenshots**

image

eatyourgreens:
Cancel event bubbling for scroll wheel events on the subtask popup and the consensus popup. This stops wheel events from bubbling up to the containing SVG image, in the React tree, and triggering wheel zoom in/out.

I've also taken the opportunity to add some documentation to the consensus popup component, and converted its stories to CSF.

Package:
lib-classifier

Closes #1835.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js
@@ -54,6 +54,10 @@ function SubTaskPopup(props) {
     }
   }
 
+  function onWheel(event) {
+    event.stopPropagation()
+  }
+
   const defaultPosition = getDefaultPosition(subTaskMarkBounds, MIN_POPUP_HEIGHT, MIN_POPUP_WIDTH)
 
   return (
@@ -62,6 +66,7 @@ function SubTaskPopup(props) {
         active
         closeFn={close}
         headingBackground='transparent'
+        onWheel={onWheel}
         pad={{ bottom: 'medium', left: 'medium', right: 'medium' }}
         plain
         position='top-left'
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.spec.js
@@ -28,7 +28,8 @@ describe('SubTaskPopup', function () {
     const fakeEvent = {
       stopPropagation: sinon.stub()
     }
-    const wrapper = shallow(<SubTaskPopup />)
+    const activeMark = { subTaskVisibility: true }
+    const wrapper = shallow(<SubTaskPopup activeMark={activeMark} />)
     wrapper.find(MovableModal).simulate('wheel', fakeEvent)
     expect(fakeEvent.stopPropagation).to.have.been.calledOnce()
   })
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.spec.js
@@ -24,6 +24,15 @@ describe('SubTaskPopup', function () {
     expect(wrapper.find(ConfirmModal)).to.have.lengthOf(0)
   })
 
+  it('should trap scroll wheel events', function () {
+    const fakeEvent = {
+      stopPropagation: sinon.stub()
+    }
+    const wrapper = shallow(<SubTaskPopup />)
+    wrapper.find(MovableModal).simulate('wheel', fakeEvent)
+    expect(fakeEvent.stopPropagation).to.have.been.calledOnce()
+  })
+
   describe('with confirm modal', function () {
     let wrapper
 
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/components/ConsensusPopup/ConsensusPopup.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/components/ConsensusPopup/ConsensusPopup.js
@@ -33,6 +33,10 @@ export default function ConsensusPopup (props) {
 
   line.textOptions.forEach((option, index) => itemProps[index] = { border: false, pad: { horizontal: 'none', vertical: 'xsmall'} })
 
+  function onWheel(event) {
+    event.stopPropagation()
+  }
+
   return (
     <MovableModal
       active={active}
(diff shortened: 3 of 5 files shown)


shaunanoordin (commented):
## PR Quick Review

Package: `lib-classifier`

This PR _attempts_ to **stop** the **mouse wheel/scroll action** from scrolling the page and/or zooming on the Subject Viewer, _if/when_ there's a modal/popup.

- This makes sense, since if there's a modal/popup, the modal/popup should capture all mouse/keyboard/touch actions instead of passing those actions to the rest of the page to handle.
- ⚠️ Unfortunately, I can't confirm that this PR works at 100%, when testing the mouse scroll on Chrome and Firefox.
  - the mouse wheel action _still_ causes a page scroll and/or subject viewer zoom when there's a SubTask modal/popup ; sometimes the page scroll and s.v. zoom happens _at the same time._
  - Basically it's functioning the same as with `master`, save for one small difference.

More deets in the following section.

### Testing

I'm testing this with I Fancy Cat on localhost; macOS10 (with touchpad and external USB mouse) + Chrome 88 + Firefox 85 (privacy mode, because it gets angry at local.zooniverse.org)

Setup:
- Any workflow with a drawing task, with a drawing tool that has tool with a subtask works.
- For my example, I'm using I Fancy Cats, task 3, "Catface" tool with subtask (drawing task: "Mark each cat's face and tail") https://local.zooniverse.org:8080/?env=staging&project=335&workflow=693

Testing steps: 
- I'm interested in seeing how the mouse scroll (or the similar touchpad scroll) works when there's a subtask popup, and when there isn't.
- First, I shrink the browser window size so that the content of `lib-classifier` exceeds the height of the viewport, meaning I have content to "scroll down to". This is to test page scrolling.
  - Note to self: check if this app-project doesn't somehow magically solve the page scroll issue.
- Second, I will test the classifier in four different scenarios:
  - ❤️ No subtask open. Using the mouse wheel on the **subject viewer area**, I expect the subject viewer to **zoom in/out**, and the page to NOT scroll. 
  - 💙 No subtask open. Using the mouse wheel on the **non-subject viewer area**, I expect the **page to scroll**, and the subject viewer to NOT zoom.
  - 💛 Subtask open. Using the mouse wheel on the **subtask popup component**, I expect the subject viewer to NOT zoom.
    - (I also think the page shouldn't scroll, but I'm not 100% sure on this expectation)
  - 💚 Subtask open. Using the mouse wheel on the **non-subtask area**, I expect the same as the previous (yellow) scenario.  

pr-2015-1

_No subtask open. ❤️ Red area indicates subject viewer, 💙 blue area indicates non-subject viewer area._


pr-2015-2

_Subtask available, popup visible. 💛 Yellow area indicates popup component, 💚 green area indicates non-popup area (which can be clicked on to close the popup)._


Results:

- ❤️ works fine in this PR
- 💙 ditto.
- 💛 mouse wheel on the Subtask popup correctly prevents the Subject Viewer from zooming. However, the page STILL scrolls.
  - this is the only functional improvement over `master`. On master, both the SV would zoom and the page would scroll.
- 💚 mouse wheel here causes BOTH the Subject Viewer to zoom AND the page to scroll.
  - this is bonkers because my mouse could be very far away from the Subject Viewer, yet the SV will still register a zoom interaction.
  - this misbehaviour is consistent on `master`

### Status

🤷‍♂️ I dunno, man?

I suppose I can break down my puzzlement to two distinct items.

First, should we allow page scrolling when the subtask popup/modal is open? If no, no problem, I'll strike that off my list of expected behaviour. If yes, then... hmm, would `event.preventDefault()` do the trick?

Second, the 💚 green scenario's a bit bonkers. I know there's a hidden Layer component that contains the movable modal, but I think it's not accepting the onWheel callback function you're passing to it. Not sure why that's causing the weird zoom+scroll interaction though/

Taking a look now.

shaunanoordin:
I'm looking in `lib-react-components` at `MovableModal.js`, and I'm noting that the render code looks like...

```
<Rnd
    cancel={cancel}
    ...bla bla bla...
    {...restRndProps}    // ...like this. There's nothing in this container layer that handles mouse wheel actions by default.
  >
    <Modal
      closeFn={closeFn}
      ...bla bla bla...
      {...rest}
    >
      {children}
    </Modal>
  </Rnd>
```

Maybe in SubTaskPopup.js, we can add something like...

```
<MovableModal
  ...
  onWheel={onWheel}
  ...
  rndProps={{
    ...
    onWheel={onWheel}  // ... something like this?
  }}
  ...
>
```

shaunanoordin:
UPDATE:

> Maybe in SubTaskPopup.js, we can add something like...

That didn't work. 😞 I don't think the `<Rnd>` [component](https://github.com/bokuweb/react-rnd) knows what to do with an onWheel callback function.

eatyourgreens:
Just to be clear, here I was aiming to trap wheel events on the pop up itself.

Wheel events are captured by whatever’s under the pointer, so I’m not sure how to handle scrolling elsewhere on the page either. Capture the pointer when a modal is open?

I don’t think we should stop people from scrolling the page, particularly on pages for transcription projects.

eatyourgreens:
I was only aiming to fix zoom too. Still not sure about scrolling but `preventDefault` might have weird side effects eg. If you’re trying to scroll content inside the popup itself.

eatyourgreens:
I found the original issue (#1835), so I've linked it to this PR. To reproduce:
1. Go to https://frontend.preview.zooniverse.org/projects/blicksam/transcription-task-testing/classify/workflow/13898
2. Click on grey lines until you find a consensus popup that has a scrollbar.
3. Try scrolling the popup content.

As far as I can tell, from local testing, that specific bug is fixed here.

shaunanoordin:
> Just to be clear, here I was aiming to trap wheel events on the pop up itself.

> I was only aiming to fix zoom too.

Ah, OK! That makes much more sense. I had mistakenly thought this PR was trying to address a wider scope than it intended. If we're only worried about preventing _zoom_ while _the pointer is on the popup,_ then the 💛 yellow scenario described in the test ticks all the boxes. 👍

shaunanoordin (approved):
## PR Review (Update)

This PR fixes an issue where the _subject viewer would zoom,_ when the user uses a _mouse wheel/scroll_ action, when the cursor is on a _subtask popup/modal._

- This PR review amends a [previous PR review](https://github.com/zooniverse/front-end-monorepo/pull/2015#pullrequestreview-582725508), which had mistakenly assumed a larger scope of action/interaction (i.e. when subtask is open _all_ mouse scroll action should be capture to prevent a page scroll or subject zoom interaction.)
- Referring to that previous PR review, the 💛 scenario tests pass everything this PR actually intends to do. 👍 

### Status

LGTM! 👍

eatyourgreens:
Cheers! We might still need to open issues if zoom is a problem with popup text tasks. Page scroll involves messing about with [passive event listeners](https://developers.google.com/web/updates/2016/06/passive-event-listeners) so I'm holding off on looking at that.

PR #1964, opened 2021-01-08
Project: Add 404 responses for projects and workflows

eatyourgreens:
Check the project model for a project ID after the project API call resolves. If there's no project ID, set a 404 status code and show the default Next error page.

When there's a workflow ID in the URL, return 404 if it isn't listed in the project's active workflows.

Split data-fetching, and page props that only depend on the page URL, into a `getStaticPageProps` helper.

Screenshot of a blank white error page, saying: 404, project btooke/i-fancy-cats was not found.

Package:
app-project

Closes #1694.

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2018, opened 2021-02-04
Projects: Localise the project model

eatyourgreens:
Add an optional `language` query param to projects. Localise the project model by fetching translation strings when a project loads.

I'm leaving this as draft for the time being, because it will conflict with the changes to page props in #1964 and it still needs tests.

PH-TESS in English: http://local.zooniverse.org:3000/projects/nora-dot-eisner/planet-hunters-tess?env=production
PH-TESS in French: http://local.zooniverse.org:3000/projects/nora-dot-eisner/planet-hunters-tess?env=production&language=fr

Package:
app-project

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/src/helpers/getDefaultPageProps/getDefaultPageProps.js
+++ b/packages/app-project/src/helpers/getDefaultPageProps/getDefaultPageProps.js
@@ -16,20 +16,20 @@ export default async function getDefaultPageProps({ params, query, req, res }) {
   }
   const isServer = true
   const store = initStore(isServer, snapshot)
+  const { env, language } = query
 
   if (params.owner && params.project) {
     const { owner, project } = params
     const projectSlug = `${owner}/${project}`
-    const { env } = query
-    await store.project.fetch(projectSlug, { env })
+    await store.project.fetch(projectSlug, { env, language })
   }
 
   const { project, ui } = getSnapshot(store)
   const { headers, connection } = req
-  const { env } = query
-  const language = project.primary_language
   const { active_workflows, default_workflow } = project.links
-  const workflows = await fetchWorkflowsHelper(language, active_workflows, default_workflow, env)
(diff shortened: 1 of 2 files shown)


srallen (commented):
I'm not sure if we want to load translations quiet like this based on our experience with the mismatch bug on PFE. I had thoughts that we'd store the original resource as is and the translation resource would be stored as a child node with the links between the two validated. The react component would render the translation resource if available, but fallback on the original parent if not present by way of computed view functions.

eatyourgreens:
Sounds good to me. We can't run into that bug here, because we fetch resources by `self.id`. If we do change this, we should change the workflow code too.

srallen:
How we request the original resource, by `self.id`, should remain. I don't understand how or why that would change based on my suggestion.

srallen:
I could sketch out what I was thinking in a discussion? (Good first time to try that out?)

eatyourgreens:
Sorry, I meant that the API query validates the translation resource. It will always be the correct translation for the current project ID. Ditto the code in `fetchWorkflowsHelper`, where we match translations to workflows based on the workflow ID. That was the validation that we didn't have, originally, in PFE.

eatyourgreens:
I think `page.title` and page.content` could definitely be computed views for project pages. That would definitely work. I'm not sure about the nested structures used in tutorials, field guides and workflow tasks eg. from Wildcam Gorongosa's survey task:
```js
workflow.translation.strings['tasks.T0.questions.ARETHEREANYYOUNGPRESENT.answers.NO.label']
```

eatyourgreens:
This isn't really needed any more.

eatyourgreens:
Re-opening this as a reference for the next round of translations work.

PR #2031, opened 2021-02-12
Classifier tasks: Add SubjectViewerStore to task stories

### LINKED ISSUE #2030: Classifier storybook: subject loading state is undefined for tasks
## Package
lib-classifier

## Describe the bug
Subject loading state is undefined for all the task stories, which crashes them in the storybook.
https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/index.html?path=/story/tasks-text--default
Screenshot of the Text Task story, showing a javascript error because 'loadingState is undefined'.

eatyourgreens:
Move the `createStore` function into a shared helper. Add the SubjectViewerStore to the mock store, so that we can mock the subject loading state.

Package:
lib-classifier

Closes #2030.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
There's a couple of other things that I could add to these stories:
1. convert them all to CSF functions.
2. fix the subject loading state toggle so that it updates the store, and properly disables the task in each story.

I wasn't sure whether to do that here or leave it to a second PR.

srallen (changes requested):
Let's add the ProjectStore to the mock and could you include refactoring the story for `Tasks` found in `/components/TaskArea/Tasks` as well? I think that would resolve all fo the errors related to the project store not existing.

eatyourgreens:
@srallen Thanks for the suggestions. I've added projects to the mock store and fixed the stories for `Tasks`. I've also moved `MockTask` into a shared helper, and updated it to disable/enable the task depending on the subject viewer loading state.
Screenshot of the text task story, disabled while the subject is loading.

eatyourgreens:
This should be ready for another review.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/DrawingTask.stories.js
```diff
@@ -1,81 +1,8 @@
 import { withKnobs, boolean, select } from '@storybook/addon-knobs'
 import asyncStates from '@zooniverse/async-states'
 import { storiesOf } from '@storybook/react'
-import zooTheme from '@zooniverse/grommet-theme'
```
srallen, line 4: Import of the theme should remain as its used throughout lines 89-125

srallen (changes requested):
Drawing task story is still broken. I've left an inline comment about what to fix. 

Both the Multiple Choice and Single Choice task stories have an annotation selected already which makes it impossible to demo what the behavior is for making the task required. I think maybe some code for Drawing task is being reused, where it does make sense and is reflecting its real behavior to have the first tool select.

One non-blocking suggestion is to put the new `createStores` and `MockTask` into a `storybook` folder to make it clear they are for that and not like the rest of the utilities in that helpers folder.

eatyourgreens:
I bit the bullet and converted all the stories to CSF, since I was rewriting them anyway. I've also changed `createStore` so that it works more like `initStore` in the project app. Given a snapshot, it now returns a new classifier store from that snapshot. That might be more generally useful than a function that's specific to the task stories. I've also updated `MockTask` to initialise the store from an initial snapshot, which is currently the default empty object.

I'm going to rearrange the helpers and then this should be ready for review again.

packages/lib-classifier/src/plugins/tasks/DrawingTask/components/DrawingTask.stories.js
eatyourgreens: Done!

eatyourgreens:
>Both the Multiple Choice and Single Choice task stories have an annotation selected already which makes it impossible to demo what the behavior is for making the task required. 

I'm not seeing this when I start up the storybook. Both tasks start with an empty annotation.

eatyourgreens:
Ok, this should be ready now. `@storybook` is already in use, for resolving storybook packages, so I've added `@stories` as an alias for any custom storybook components and utilities (just the `MockTask` component right now.) I've also added `@components` to make component imports a little easier to write.

packages/lib-classifier/src/helpers/index.js
```diff
@@ -1,8 +1,7 @@
 export { default as createLocationCounts } from './createLocationCounts'
-export { default as featureDetection } from './featureDetection'
```
eatyourgreens, line 2: It turns out `featureDetection` doesn't have a default export, so I deleted it.

srallen: Could you not? It's part of the service worker code I wrote, and although the service worker is not in use right now, if we're going to remove it, it should be removed altogether with that code where it's used.

eatyourgreens: The `featureDetection` exports are still there. It’s just that `featureDetection` has no default export so this line is erroring in the console during module resolution.

eatyourgreens: I'll change this to export all the named feature detection functions. That should fix the following warning on master.
```
WARNING in ./src/helpers/index.js 2:0-65
"export 'default' (reexported as 'featureDetection') was not found in './featureDetection'

```

eatyourgreens:
@srallen this should be fixed now so that annotation state doesn't persist across stories.

packages/lib-classifier/src/helpers/index.js
srallen: Ok, I misunderstood your first note as meaning you removed the helper function entirely. Thanks!

srallen (approved):
LGTM. Thank you! One idea for a future small PR is that we should put the aliases into a definition list in the classifier lib readme. On the surface, it's not obvious what's our aliases for convenience and what's an actual npm package.

eatyourgreens:
That's a good point. I've also tidied up the aliases because they were out of sync between the two webpack configs.

Issue #379, opened 2019-01-21
Classifier should have 'persist annotations' setting


shaunanoordin:
## Feature Required
package: `lib-classifier`

- The Classifier, by default, removes the annotations made by the user on the current step/task, whenever the user goes back to a previous step/task.
- In PFE, the user is given a (small) warning before they go back a previous step and this happens.
- However, there is a 'persist annotations' setting that can be enabled. (See the [dev Classifier](https://master.pfe-preview.zooniverse.org/dev/classifier), with the 'persist annotations' option at the bottom of the page.)

Follows discussion on https://github.com/zooniverse/front-end-monorepo/pull/378#issuecomment-456138485 . Also see full Slack discussion on `#dev-frontend` between @shaunanoordin @eatyourgreens @srallen on 21 Jan 2019.

srallen:
The todo for this issue should be the second point: 

In PFE, the user is given a (small) warning before they go back a previous step and this happens.

We need to add the tooltip warning.

srallen:
Note the actual TESS project will only have a single step, so, the warning isn't strictly necessary for its launch.

srallen:
We should make sure that this works with recursive workflows.

rogerhutchings:
It may be that instead of this being available as a setting that toggles visibility in the classifier, we have some other UI for showing what you've added already (since this is primarily used in transcription projects)

srallen:
It was noted in https://github.com/zooniverse/front-end-monorepo/issues/1272#issuecomment-553828531 that projects may be using this as a means to have a in-progress summary when using text and dropdown tools, usually used by humanities projects. Maybe we should totally rethink this instead as an UI/UX in-progress summary that is needed. Are there any use cases for annotation persistence outside of the transcription and humanities use case?

srallen:
If we re-concept this feature, then we need 1 week of design time and unknown amount of time for dev. 
If we implement as is, then 1 week dev time.

eatyourgreens:
Do we know how often this gets used to go back and edit annotations for complex drawing and transcription workflows? That would influence whether or not it's a high priority item.

PR #1989, opened 2021-01-19
Add survey task characteristics filter components

mcbouslog:
Package: lib-classifier

Closes #1902, closes #1901.

Staging workflow: https://local.zooniverse.org:8080/?project=1634&workflow=2434

Describe your changes:
- [x] adds Filter drop button
- [x] adds Characteristics (container), CharacteristicSection, and FilterButton components
- [x] adds functionality (temporarily using state, until #1990) to select filter, show selected filters (in container and next to Filter button), and clear filter
- [x] adds Clear Filters button

ToDo:
- [x] tests
- [x] stories
- [x] cleanup styling (somewhat accomplished, to be revisited at complete task level)

# Review Checklist

## General

- [x] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [x] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2050, opened 2021-02-26
Classifier: add support for recursive, branching workflows

eatyourgreens:
Includes #2048.

This adds support for branching, recursive workflows by adding a history manager to the `Tasks` component. Task annotations are rendered from the history stack if this is a step that we’ve previously completed. 

The Back button undoes the current step. The Next button adds a new step to the stack if we are starting a step for the first time. When the subject queue advances, the stack is cleared and a new stack started from the first workflow step.  The first step in the stack cannot be undone. 

The history API is simplified down to four actions:
- `annotatedSteps.start()`: Start recording annotations, step by step, for a new subject.
- `annotatedSteps.back( persistAnnotations = true )`: Move back one step. Undo the current step and restore annotations from the previous step. Keep a redo history if `persistAnnotations` is set.
- `annotatedSteps.next()`: Redo annotations for the next step. Supports branching and recursion by clearing future annotations and steps if the selected step has diverged from history.
- `annotatedSteps.clearRedo()`: delete the redo history, and remove any pending annotations from `classification.annotations`.

UPDATE: now with support for `undoManager.redo`. The Next button redoes history, if possible, with `undoManager.redo()`. The latest step key, from recorded history, is compared with `step.next`. If we've started down a new path, from a branching question, then the redo stack is discarded and a new history is started from the branching point. This allows annotations to persist without being lost, even if we go back and choose a new path.

On classification complete, any pending annotations from the redo history are deleted, so that the submitted classification only contains annotations that we've actually done.

New features supported here:
- recursive workflows: return to a previous step and complete it with new annotations, rather than editing existing annotations for that step.
- persistent annotations: annotation IDs are stored in history and can be undone or reapplied with undo/redo. You should be able to move backwards and forwards through a workflow without losing any text that you've entered.

Some testing URLs:
- Shaun's 9-room dungeon maze. This is probably about as complicated as a branching workflow can get. Each room is a workflow step.
https://github.com/zooniverse/front-end-monorepo/pull/2050#issuecomment-790812025
- A test workflow with two branching questions, one of which takes you back to the beginning of the workflow.
https://localhost:8080/?project=908&workflow=3468&demo=true
- The National Archive's recursive transcription workflow. This is the workflow that we are aiming to launch for the Engaging Crowds project.
https://localhost:8080/?project=12268&workflow=17076&env=production&demo=true

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.js
```diff
@@ -63,16 +59,11 @@ class TaskNavButtonsContainer extends React.Component {
...
     } = storeMapper(this.context)
     this.completeStepTasks()
-    const annotations = tasks.map(task => classification.annotation(task))
-    annotatedSteps.next({
-      step,
-      annotations
-    })
-    selectStep()
+    const { next } = annotatedSteps.latest.step
+    selectStep(next)
```
eatyourgreens, line 66: I think it would be good to have the Next button update history here, so that both Back and Next buttons are responsible for updating history. Something like
```js
annotatedSteps.next({
  step: nextStep
  classification
})
```
where `nextStep` is the next step from the workflow's steps.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js
```diff
@@ -13,10 +13,31 @@ import counterpart from 'counterpart'
...
+  if (readyToClassify) {
+    const latestSavedStep = latest?.step
+    const stepIsNew = latestSavedStep?.stepKey !== step.stepKey
+    if (stepIsNew) {
+      console.log(`Adding ${step.stepKey} to history`)
+      annotatedSteps.next({
+        step,
+        classification
+      })
+    }
```
eatyourgreens, line 38: This is needed to start a new history stack, on the first step for each subject. Maybe there's a better way to do this in the stores themselves?

It may also be needed if we've changed the answer to a branching question ie. the stored history no longer matches our new path through the steps tree.

eatyourgreens: Now handled in the store, using `annotatedSteps.start()` when a new subject loads.

eatyourgreens:
This is working for me now on the TNA Index workflow (choose Meeting from the first list of options, then 'add a new heading' later on to repeat the heading and details steps.) Tests are passing too so this might be close to being ready.
https://localhost:8080/?project=12268&workflow=17076&env=production&demo=true

eatyourgreens:
Redo is enabled now, which persists annotations by default. I'm not sure if that closes #379. If we add the option for project owners to turn off redo then we'll also need the warning on the back button, that going back will erase your work.

shaunanoordin:
I've created a **Choose Your Own Adventure** dungeon maze to test the robustness of our branching system. ⚔️ 🛡️ 👹 

PR testing: https://local.zooniverse.org:8080/?env=staging&project=1873&workflow=3467

PFE baseline: https://master.pfe-preview.zooniverse.org/projects/darkeshard/test-project-2021/classify?reload=0&workflow=3467

**Spoilers:** dungeon map here.

Screen Shot 2021-03-04 at 17 36 57

Asides from the lol factor of making a Dungeon of the Doomiverse, here's what I'm actually testing:

- [ ] Classifier should start at Room 8. (Checking for: 'workflow.first_task')
- [ ] Every room **requires** an answer (i.e. a direction to walks towards) (Checking for: 'task.required')
  - This means the "Next" button should be blocked until the user chooses an answer. 
  - [ ] The exception is Room 8, see below.
- There are only two ways a user can "complete the dungeon" (i.e. submit a Classification)
  - [ ] At Room 8 (Entrance), the user clicks "Done" _without_ choosing an answer. (In-universe, this is the equivalent of giving up.)
  - [ ] At Room 6 (Exit), the user chooses "Exit the Dungeon" and clicks "Done."
- Note that at Room 6, the user MUST choose "Exit the dungeon". There's no going back West to Room 5.
- [ ] Submitted classification should keep track of the user's path through the dungeon.
- [ ] Optional: the "Back" button should be tested, but I dunno, my instincts tell me this will have its own check sheet.

shaunanoordin:
## Quick Tests

Some quirks I'm noting...

- Initially, things look good.
  - ✅ I start at Room 8. 
  - ✅  When starting at Room 8, I'm able to choose no answer click "Done". (But ONLY if I don't move to another room and revisit Room 8. See below.)
  - ✅ I can travel between rooms (rooms are linking correctly).
  - ✅ I can submit the Classification either at Room 6. 
- ⚠️ The `task.required` rule seems to be extremely inconsistent.
  - Let's say I start at Room 8. 
  - ❌ If I go East to Room 9, the "Next" button can be clicked _without_ choosing an answer.
  - ❌ If I travel West again to Room 8, there's no "Done" button, only a "Next" button.
  - ❌ If I click on "Next" without choosing an answer, I'll just go back and forth between Room 9 and 8! (This will be recorded in the Classification.)
- ⚠️ The "Done" button can also appear when it shouldn't!
  - Let's say I start at Room 8.
  - ❌ If I move West to Room 7, the Done/Done & Talk buttons appear!

❗❓ To repeat: there seems to be a bunch of weirdness, which I can only articulate as...

1. The **task.isRequired** logic seems to be ignored. I need to check what happens when the first task of a workflow is Required.
2. **Premature Conclusion** -  The "Done" button appears prematurely IF moving to a room for the first time. EXCEPTION: this doesn't seem to apply to the "final" room (last Task in the array of Tasks in a workflow)
3. **No-Answer Loop** - the "Next" button appears IF visiting a room that's been visited before. This might have something to do with that room already having an "answer" from a previous visit.

This will require some further digging to figure out WTF is happening in this dungeon.

eatyourgreens:
Required tasks will be broken because the logic for those assumes that a task only has one annotation in the classification. That should be straightforward to fix.

eatyourgreens:
This PR doesn’t touch the logic for the Next and Done buttons, so that’s weird behaviour. I’d expect Back to appear on any step that isn’t the very first one, since it’s wired up to `history.canUndo` and the first step in a workflow is the only one that can’t be undone.

eatyourgreens:
Oh, Next might appear if there’s a previous annotation for the current task. There’s an assumption of one annotation per task that’s built into the existing code in a few places.

eatyourgreens:
Anything that uses `classification.annotation(task)` to get an annotation will probably break when `task` can have more than one annotation.

eatyourgreens:
I've set up a test workflow, with required tasks. It has two branching questions, one of which takes you back to the beginning of the workflow.
https://localhost:8080/?project=908&workflow=3468&demo=true

eatyourgreens:
D'oh! I accidentally broke the `disabled` prop for the buttons in #2048. Required tasks should be fixed now.

eatyourgreens:
@shaunanoordin I just did a quick run through your workflow:

8 > next > 9 > back > 8 > next > 7  > next > 4 > next > 5 > next > 6 > back > 5 > back > 4 > next > 5 > next > 6.

It seemed to work. Required tasks were disabled until answered and Next took me back to the previous step after undoing with the Back button.

packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.js
```diff
@@ -26,46 +40,87 @@ const AnnotatedSteps = types.model('AnnotatedSteps', {
...
+    history.undo()
+    history.clearRedo()
+    workflowSteps.selectStep(self.latest.step.next)
+    self.createStep()
+  }
+  /** Clear stored steps and history. Should be run before classifying a new subject. */
   function reset() {
     const { history, steps } = self
     history.withoutUndo(() => {
       steps.clear()
```
eatyourgreens, line 104: `reset` and `start` are called from disposers, which can lead to the history changing unexpectedly when observed values change in other stores. I've stopped that from happening by wrapping those actions in `withoutUndo`. I'll add a comment about that to the store. It might also be possible to make these actions private, if they aren't expected to be called from React components or other models.

eatyourgreens:
`step.isComplete` does have the same bug that we've seen before for drawing subtasks. When a task has more than one annotation, the `Task` component state reflects the most recent annotation.

Here, 'Answer an extra question?' has two annotations in the classification. The first one is answered 'Yes', the second one is unanswered and incomplete. I've clicked Back, from that second task, until I got back to the first instance of the same task, at which point the task is answered but I'm unable to go back forward again, because the Next button is connected to the most recent annotation.

Screenshot of 'Answer an extra question?' with 'Yes' selected but the Next button disabled and not clickable.

packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.js
eatyourgreens: Done.

eatyourgreens:
If you go forward a few steps on the recursive workflow, then go back and  change your mind, `classification.annotations` now contains annotations that won’t be submitted when you press Done. It would be good if orphaned annotations could be deleted when `clearRedo()` runs. At the moment, I’m finding it confusing that the submitted classification doesn’t match the classification store (although the submitted classification is correct, because it is serialised from the history of steps that you’ve completed.)

packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.js
```diff
@@ -27,45 +41,82 @@ const AnnotatedSteps = types.model('AnnotatedSteps', {
...
+      console.log(`Adding ${historyStep.step.stepKey} to history`)
+      steps.put(historyStep)
+    }
+  }
+  /** Clear the redo history and restart history from this point with a new step. */
+  function _replace() {
+    const { history } = self
+    const { workflowSteps } = getRoot(self)
+    history.undo()
+    history.clearRedo()
```
eatyourgreens, line 69: This leaves orphaned annotations in `classification.annotations`. It would be good if those could be cleaned up when we replace history.

eatyourgreens: I guess the MST answer to this would be that `step.annotations` should store the annotations for a step, not references. Safe references will be automatically cleaned up when the original is destroyed.

eatyourgreens: The downside of this might be that every letter in a text annotation gets recorded as a new undo step in the history stack.

shaunanoordin:
Jim, I've gone through the latest updates (d089a14ac486919a9fd3fcc88a2c2964c63e7770) and here are my thoughts:

- Code changes look good, as far as I can tell with so many interconnected parts.
- Functionality tests look good for MOST cases, particularly for the [National Archives workflow](https://localhost:8080/?project=12268&workflow=17076&env=production&demo=true) and [Jim's test workflow](https://localhost:8080/?project=908&workflow=3468&demo=true)
- ⚠️ I can still find bugs and quirks with the [Dungeon workflow](https://local.zooniverse.org:8080/?env=staging&project=1873&workflow=3467)...
  - ❗ but I'm thinking it's best that we don't count these bug against this PR for this review.
  - I'm afraid trying to fix what's turning out to be very fundamental flaws (e.g. several assumptions where only one instance of a task will be seen per classification) will snowball into a huge rewrite of several fundamental components, and delay getting this PR done for the project it's targeting.

Basically, I think this PR should be safe to merge into #2048 . I'll make note of my discoveries, and any bugs _specific to only the Dungeon workflow_ should be noted for a future update, not for this PR.

**The good:**
- Looks like the disabled prop is working again _prop_-erly! 👍 
  > D'oh! I accidentally broke the disabled prop for the buttons in #2048. Required tasks should be fixed now.
- Branching paths work in most normal cases.
- Recursive/repeating Tasks seem (going from A to B to A again) to work in most normal cases.
- There is no "only one Task of its kind" problem. When there are multple instances of a Task, each instance has its own answer. (e.g. if I see "Task A: what's your favourite food?" three times, I can answer "apple", "banana", "cherry" for each instance and this will be recorded in the Classification correctly.) 
- the 'Back' button works in most normal cases.

Note: "normal cases" refer to tests based on [National Archives workflow](https://localhost:8080/?project=12268&workflow=17076&env=production&demo=true) and [Jim's test workflow](https://localhost:8080/?project=908&workflow=3468&demo=true)

----

**Quirk 1:** Persistent "Future Answers"
- If you choose answer A in Step 1, then choose answer X in Step 2, go "Back", then choose answer A in Step 1 again... you'll end up win Step 2 with answer X already chossen.
- To replicate on https://localhost:8080/?project=908&workflow=3468&demo=true
  - Q1 : answer "whatever". -> Next.
  - Q2 : Answer an extra question? Choose "Yes" -> Next
  - Q3 : Guinea pigs or chinchillas? Choose "Guinea Pigs" -> BACK
  - Q2 : "Yes" should already been chosen -> Next
  - Q3 : _Guinea pigs is already chosen_

Not sure if this is related to Bug 1. This is a quirk and not a full-blown bug, because if you "No" at Q2 and proceed, the stores/Classification correctly delete any previous answers to Q3 from memory.

**Bug 1:** That thing you reported - "step.isComplete" uses answers from earlier instances
- I can confirm this is happening.
  > step.isComplete does have the same bug that we've seen before for drawing subtasks. When a task has more than one annotation, the TaskNavButtons component state reflects the most recent annotation.
- Replication steps is exactly as you said
  > Here, 'Answer an extra question?' has two annotations in the classification. The first one is answered 'Yes', the second one is unanswered and incomplete. I've clicked Back, from that second task, until I got back to the first instance of the same task, at which point the task is answered but I'm unable to go back forward again, because the Next button is connected to the most recent annotation.

Yeah, Bug 1 this is the only thing I'd consider a potential blocker, since it definitely affects your test workflow, and might also affect the National Archives workflow. (I'm still testing the latter.)

**Bug 1.1:** Probably another iteration of  Bug 1
- Replication steps:
  - Go to the [Dungeon](https://local.zooniverse.org:8080/?env=staging&project=1873&workflow=3467)
  - In Room 8. Choose "West" -> Next
  - In Room 7. Choose "East" -> Next
  - In Room 8. No answer is chosen, but "Next" is clickable.
    - We expect that there is a "Done" option here since Room 8 should allow users to submit without an answer.
    - Let's say we don't choose an answer and click on "Next"...
  - In Room 7.
    - ...Ah! It seems that our previous "West" answer for Room 8 was saved, just as you mentioned in Bug 1.
    - Note that despite the fact that I'm moved to Room 7, the answer for my second instance for Room 8 is recorded as `null` because I didn't choose an answer.
    - But wait! The _"uses answer from previous instance"_ problem isn't triggering for Room 7, because without choosing a new answer for this new instance of Room 7, the "Next" button is disabled!
    
So here's what I'm wondering... is there a difference in the "step is complete?" logic when an answer _isRequired_ (Room 7) and when it isn't (Room 8)? Hmm.

**Bug 2** Sometimes Undo doesn't Undo all the way
- Replication steps:
  - Enter the [Dungeon](https://local.zooniverse.org:8080/?env=staging&project=1873&workflow=3467)
  - In Room 8. Choose "West" -> Next
  - In Room 7. Choose "East" -> Next.
  - In Room 8.
  - Back.
  - Back.
  - We're now in the first instance of Room 8, with "West" still selected.
  - Choose "East" (heading towards Room 9) -> Next.
  - We end up in Room 7 instead!

This one's a Dungeon-only bug, but hell, it threw me for a loop.

----

That's about all that I can comfortably document for one night. That leaves only Bug 1 (the one you discovered) as the only potential blocker - what are your thoughts? If the fix is going to be a fundamental rewrite of the workflow, then I say it should be a separate rewrite - I'll approve this PR for merging first, if that the case. As I said, everything else is fine for most _normal cases._

eatyourgreens:
I’m not sure this is ready yet. The tests are passing but I think there might still be missing cases in the specs. 

Future history should persist. That’s a feature that is needed for Engaging Crowds. Imagine filling out all those text boxes for the TNA workflow, then losing your work when you go back! The code’s using redo on Next, except when the chosen answer changes for a branching question. There’s a test block for that case too.

eatyourgreens:
Maybe there’s a bug in `step.next` when there are two annotations for that step? Your example sets `step.next` to room 7, from room 8, goes forward two steps, then back two steps. You found that `step.next` is room 7, even if room 9 is selected. 

I haven’t changed any of the code for `step.next` but maybe that logic becomes confused when there’s more than one annotation for the same step. 

Maybe the code here would be easier to follow if the `next()` looked directly at which answer you chose for a branching question, rather than getting it indirectly via `step.next`?

eatyourgreens:
This line here won’t work if there is more than one annotation for `singleChoiceTask`. You’ve uncovered that bug by annotating room 8 twice, I think. https://github.com/zooniverse/front-end-monorepo/blob/22fbe7a3566b3013f3187c2824aa9a6558b797fe/packages/lib-classifier/src/store/WorkflowStepStore.js#L110

In retrospect, `task.annotation` was a bad idea on my part. It’s caused us bugs for transcription, where the same task is annotated many times. It’s now causing problems on recursive workflows, when the same task is annotated more than once.

eatyourgreens:
That bug could be fixed if the `next()` action checked `latest.annotations` for a new answer, before checking `step.next`.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.spec.js
```diff
@@ -85,142 +115,117 @@ describe('TaskNavButtonsContainer', function () {
...
-      wrapper.instance().goToPreviousStep()
-      expect(removeAnnotationSpy).to.have.been.called()
-      expect(removeAnnotationSpy.withArgs('T1')).to.have.been.calledOnce()
+      it('should select the previous step', function () {
+        const activeStep = classifierStore.workflowSteps.active
+        expect(activeStep.stepKey).to.equal('S0')
+      })
+
+      it('should remove annotations for the current task', function () {
+        const { annotations, steps } = classifierStore.annotatedSteps
```
eatyourgreens, line 190: TODO: figure out how to remove orphaned annotations from `classification.annotations` when `clearRedo()` runs.

packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.js
```diff
@@ -0,0 +1,131 @@
...
+    const { workflowSteps } = getRoot(self)
+    if (undoManager.canUndo) {
+      undoManager.undo()
+      const { step } = self.latest
+      workflowSteps.selectStep(step.stepKey)
+    }
+  }
+  /** Redo the next step, or add a new step to history if there is no redo. */
+  function next() {
+    let nextStepKey = self.latest.nextStepKey()
```
eatyourgreens, line 97: Maybe break this method up into `_redo()` and `_replace()`?

packages/lib-classifier/src/store/AnnotationsStore.js
```diff
@@ -24,6 +24,7 @@ const AnnotationsStore = types
     function createAnnotation(task) {
       const taskAnnotation = task.createAnnotation()
       self.annotations.put(taskAnnotation)
+      task.setAnnotation(taskAnnotation)
```
eatyourgreens, line 27: I don’t think this is needed any more.

eatyourgreens:
I think the bugs with `step.isComplete` and `step.next`, which triggered when `classification.annotations` contained many annotations for the same task, should be fixed now.

I've also added a `clearRedo` action, which clears out orphaned annotations from `classification.annotations` when we reset the redo history.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.js
```diff
@@ -40,74 +49,80 @@ class TaskNavButtonsContainer extends React.Component {
...
     }
   }
 
   goToNextStep () {
-    const { selectStep } = this.props
+    const { classifierStore } = this.props.store ?? this.context
+    const {
+      annotatedSteps
+    } = storeMapper({ classifierStore })
     this.completeStepTasks()
```
eatyourgreens, line 66: I’m not sure if this line is needed any more.

shaunanoordin:
## Quick Review

Jim, I'm happy to report that the [previously reported](https://github.com/zooniverse/front-end-monorepo/pull/2050#issuecomment-791849058) Quirk 1, Bug 1, Bug 1.1, and Bug 2 have _all been eradicated_ by your latest changes! 🌟 

It looks like my main functionality tests have all passed, so I should be able to wrap up my review for this PR. I'm going to take my time to do a thorough code read so I can grok what precisely what happened.

brb!

eatyourgreens:
There's a lot of code to read! I might move some of the test set-up into a helper, as I notice that I'm using the same setup code to create a mock store in different test files.

I've just run through the dungeon workflow on PFE and in the new classifier, going backward and forward a bit to change my classification annotation history. The submitted classification in the new classifier seems to be identical to the PFE version.

One bug not yet noted here, I think, is pending annotations from history creeping into the submitted classification. If you go forward, then back, your future annotations aren't removed immediately, in case you redo them. I think I've fixed that by cleaning the classification on submit.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.spec.js
```diff
@@ -1,77 +1,107 @@
...
-const project = ProjectFactory.build()
-const subject = SubjectFactory.build()
-const workflow = WorkflowFactory.build()
+import RootStore from '@store'
+import Step from '@store/WorkflowStepStore/Step'
+import { SubjectFactory, WorkflowFactory, ProjectFactory } from '@test/factories'
+import stubPanoptesJs from '@test/stubPanoptesJs'
 
 describe('TaskNavButtonsContainer', function () {
   function setupMocks () {
```
eatyourgreens, line 15: It might be possible to import this function from a helper. It gets used in a couple of different test files.

eatyourgreens:
I should note that I've left logging in the code, so you should see messages like this in the dev console during branching workflows.
```
Adding S7 to history AnnotatedSteps.js:78:15
Adding S4 to history AnnotatedSteps.js:78:15
Adding S5 to history AnnotatedSteps.js:78:15
Adding S2 to history AnnotatedSteps.js:78:15
Adding S3 to history AnnotatedSteps.js:78:15
redo S5 AnnotatedSteps.js:104:13
redo S2 AnnotatedSteps.js:104:13
redo S3 AnnotatedSteps.js:104:13
Adding S2 to history AnnotatedSteps.js:78:15
Adding S5 to history AnnotatedSteps.js:78:15
redo S2 AnnotatedSteps.js:104:13
replace S2 with S6 AnnotatedSteps.js:107:15
Adding S6 to history AnnotatedSteps.js:78:15
```

eatyourgreens:
I've just pushed one more change, which is to convert `TaskNavButtonsContainer` to a functional component.

packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedStep/AnnotatedStep.js
```diff
@@ -0,0 +1,32 @@
...
+    /** Get the next step key from any single choice answers, or the current step otherwise. */
+    nextStepKey() {
+      // assume only one branching question per step.
+      const [ singleChoiceAnnotation ] = self.annotations.filter(annotation => annotation.taskType === 'single')
+      if (singleChoiceAnnotation) {
+        const [singleChoiceTask] = self.step.tasks.filter(task => task.taskKey === singleChoiceAnnotation.task)
+        const selectedAnswer = singleChoiceTask?.answers[singleChoiceAnnotation.value]
+        return selectedAnswer?.next
+      }
+      return self.step.next
```
eatyourgreens, line 27: Maybe this can also be moved down into the Step model. Something like
```js
return self.step.nextStepKey(annotations)
```

That would make this more consistent with `step.isComplete`, and keeps all the step logic in one place.

eatyourgreens: Done in da32e387e29ca3cf7874d19d3034f019ec1ed931

eatyourgreens:
Sorry, one more commit. I'd removed tests for the Step branching logic, so those are back now, updated for the new code.

packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.js
```diff
@@ -0,0 +1,148 @@
...
+function setUndoManager(targetStore) {
+    undoManager = UndoManager.create({}, { targetStore })
+}
+/**
+  A history manager which records completed steps, from store.workflowSteps, and annotations, from store.classifications.
+*/
+const AnnotatedSteps = types.model('AnnotatedSteps', {
+  /** Completed steps, with their annotations */
+  steps: types.map(AnnotatedStep)
+})
```
shaunanoordin, line 17: Shaun's dev notes 2:
- OK, I can mostly grok this now.
- Tasks/Task components now required Annotated Steps.
- Annotated Steps is a logical container that manages where we are in a series of Annotated Step(s)
- An Annotated Step is just a classic Step with the annotations. (To grok: so what happened to classification.annotation? It's still there, so does _AnnotatedStep.annotations_ just point to the same reference, or a sub-tree of classification.annotation ? Check/re-read.)
- So _Task_ has _AnnotatedSteps_ has many _AnnotatedStep(s)_ is _Step_ + annotations.

shaunanoordin (commented):
## PR Review

Package: `lib-classifier`
Note: includes #2048 , so if functionality tests for 2050 pass, I have no reason to doubt 2048 is safe to merge too.

This PR is **major overhaul** to the way we organise the flow of, er, Workflows, ultimately allowing us to have **branching** and/or **recursive** 'steps' of Tasks.

- **AnnotatedSteps** store now introduced to manage which Step a user is currently at, and what their Next step should be. Also manages undo/redo shenanigans.
  - Task/Tasks now use AnnotatedSteps. (Instead of checking for classification.annotations directly.) AnnotatedSteps has multiple AnnotatedStep(s). AnnotatedStep is a Step (of Tasks) wrapped together with the annotations for that step.
- **Important changes to base assumptions:**
  - ❗ the code no longer assumers there can only be one instance of Task X in the classification's annotations. 👍  
  - ❗ the code now assumes there can be only ONE branching question per Step.
- New Classifier functionality:
  - Workflows allow branching steps. (If, in Task 1, you answer X, you'll go to Task 2. If you answer Y instead, you'll go to Task 3)
  - Workflows allow recursive steps. (Task Z can be encountered multiple times in a single classification. Each instance will have its own instance of annotation.)

### Testing

What we're checking:
- Can a standard (non-branching, non-recursive) workflow go through fine? (i.e. I Fancy Cats)
- Branching workflows:
  - For a Question Task where Answer A takes me to Task 2, and Answer B takes me to Task 3, does this actually work? 
  - For a Question Task with branching answers... but where NO answer is required, what will happen when a null answer is declared? (Expected: "Done" button should show instead of "Next", and clicking it will immediately submit the classification)
- Recursive workflows:
  - When visiting the second (or third, etc) instance of Task X, I should not see the answers I submitted for earlier instances of Task X. (i.e. each instance needs its own annotation)
  - By extension this means answers from earlier instances of Task X shouldn't affect the consequence of future answers from future instances of Task X.
- Misc:
  - will the "Next" and "Done" buttons correctly reflect the Task's "isRequired" status?
  - the "Back" button should be able to take the user aalll the way back to the first Task/step.
  - when clicking "Back" to move from Task B to Task A, all of Task B's annotations should be wiped. ("No Future But What We Make" rule.)
  - submitted Classification should have annotations corresponding to the path of branching and recursive answers, MINUS any answers that were undone by an undo/"Back" button.

EDIT: testing interrupted, **I gotta pull the latest update and try again.** No worries, everything was looking good so far!

Need to test? (Question mark?)
- steps with multiple Tasks.
- multiple browser types? (Optional)
- the maximum amount of steps/tasks that can go into the Undo/Redo pile before a standard Mac's allocated memory for Chrome/Firefox/Safari explodes.

Tested on localhost with macOS10 + Chrome88.

### Status

LGTM, but I [just noticed](https://github.com/zooniverse/front-end-monorepo/pull/2050/#issuecomment-794240792) that there's a new update (da32e387e29ca3cf7874d19d3034f019ec1ed931) so I'm going to restart my tests. brb!

shaunanoordin (changes requested):
## PR Review (Update)

Follow up to [initial review](https://github.com/zooniverse/front-end-monorepo/pull/2050#pullrequestreview-607694710)

Code read looks good, testing looks MOSTLY good with a few WTFs.

### Testing

localhost, macOS10 + Chrome 88

URLs:
- Jim's testing workflow: https://localhost:8080/?project=908&workflow=3468&demo=true
- National Archives workflow: https://localhost:8080/?project=12268&workflow=17076&env=production&demo=true
- Dungeon workflow: https://local.zooniverse.org:8080/?env=staging&project=1873&workflow=3467
- I Fancy Cats: https://local.zooniverse.org:8080/?env=staging&project=335&workflow=693

See [initial review, Testing section,](https://github.com/zooniverse/front-end-monorepo/pull/2050#pullrequestreview-607694710) for what we're testing. There's quite a lot going on so I'll avoid repeating myself and just go through the highlights.

Scenarios:
- ❌ Test basic (non-branching, non-recursive) workflow
  - _Wait, wtf? Are Drawing Tasks broken?_ See New Bugs.
- ✅ Test to see if branching tasks work. (e.g. if answer A go to Task 2, if answer B to to Task 3)
- ✅ Test to see if recursive tasks work. (i.e. every instance of Task X should have its own answer)
- ✅ Test to ensure Task.isRequired continues to work.
- Test to ensure 'Back' action works.
  - ❌ When going back from Task B to Task A, then to Task B again, then Task B should not 'remember' any previous answers. (**Quirk 1:** Persistent "Future Answers")
  - ❌ When I go Back to a previous task that branches, I should be able to choose a separate path. (See New Bugs)
  - ✅ Submitted Classification should not record any answers that 
    - That's right, the submitted classification looks fine.

### New Bugs (Aaargh)

Returning: [**Quirk 1**](https://github.com/zooniverse/front-end-monorepo/pull/2050#issuecomment-791849058)
- Not sure why this is happening. I was pretty sure this problem was gone when I tested earlier this evening! But whatever's causing it might be causing the next problem.

New: **Bug 3** you can't go back and answer a branching decision more than twice
- Replication steps:
  - Go to [Jim's testing workflow](https://localhost:8080/?project=908&workflow=3468&demo=true)
  - Q1: "How many times have you answered this question?" -> answer "once" -> Next
  - Q2: "Answer an extra question?" -> answer "YES" -> Next
  - Q3: "Guinea pigs or chinchillas?" -> _Back_
  - Q2: "Answer an extra question?" -> answer "NO" -> Next
  - Q4: "Do it again?" -> _Back_
  - Q2: "Answer an extra question?"
    - At this point, we notice the error: `Uncaught Error: [mobx-state-tree] A map of objects containing an identifier should always store the object under their own identifier. Trying to store key 'ckm2o6nnr000h3h62y7f568c2', but expected: 'ckm2o6l7700053h624zzdbner'` (this error appears twice)
    - ⚠️ now here's the crazy part... once we trigger the error, we CAN'T go BACK any more. (i.e. in this example, we can never revise the answer in Q1.) BUT, we can proceed to move FORWARD with no problems, and we can SUBMIT the Classification with no issues. (Annotations are correct with no weird orphaned answers.)

New: **Bug 4** drawing tasks are broken (cannot be submitted)
- Replication:
  - Go to I Fancy Cats: https://local.zooniverse.org:8080/?env=staging&project=335&workflow=693
  - Answer the first two questions.
  - At the third/final question, the Drawing Task, click "Done".
    - Expected: classification is submitted.
    - Actual: classification is not submitted. Classifier will not proceed past the final question. Error message: `Uncaught TypeError: Cannot read property 'update' of undefined`
  - Note: you can still draw annotations, you just can't submit and(?)/or(?) move to the next Task.

For bug 4, I've traced it to `src/plugins/tasks/DrawingTask/models/DrawingTask.js`...
```
function complete () {
  // self.annotation is undefined!
  self.annotation.update(self.marks)
}
```

`master` works fine btw, so my guess is that this PR changed one of those implicit assumptions (i.e. not goshdarn spelled out in a comment) that the Task will auto-create annotations somehow.

(I'm looking at you, removed line `annotation = classification.addAnnotation(task)` in Task.js. Or maybe the removed line `let annotation = classification.annotation(task)`? Didn't one of these lines automagically try to create an annotation via default annotation if one didn't exist already?)

### Status

OK, I think we ONLY need to fix **Bug 4:** _drawing tasks are broken_ before this PR is good to go. I _strongly recommend_ worrying about Quirk 1 and and Bug 3 in another separate PR, since this branch has gotten megahuge already with a chunky amount of legit good logic overhauling.

The only reason I don't just approve this PR right now and mark Bug 4 as a "fix separately" item is because I don't think we should break what's already working. I'm going to see if I can dig up any further info on the issue.

shaunanoordin:
OK, this is as far as I can get tonight.

- Let's look at the generic `Task` model in `src/plugins/tasks/models/Task.js`
- Prior to this PR, `task.annotation` will have a value. This value is generated here, in `components/Task/Task.js`
  ```
  let annotation = classification.annotation(task)
  if (!annotation) {
    annotation = classification.addAnnotation(task)  // <-- here! This makes sure Task.annotation is initialised or something. 
  }
  ```
- `task.complete()` is triggered upon step completion (clicking Next or Done). Thing is, `task.complete()` is a BLANK function for every task EXEPT drawing tasks. `drawingTask.complete()` requires `drawingTask.annotation` to have a value, which is why _only_ Drawing Tasks are crashing with this PR.

Good news is we've figured out what needs to be changed. I'm going to go back through the commits to see why `annotation = classification.addAnnotation(task)` and if it can't just be reinserted somewhere somehow.

shaunanoordin:
Note: as a side effect of whatever's going on with the Drawing Tasks, Drawing Tasks is _always_ "complete", so it always passes the "isRequired" check.

eatyourgreens:
@shaunanoordin thanks for the thorough review. I think the answer for tasks is to always pass the current annotation as an argument to actions, so:
```js
task.complete(annotation)
```
Otherwise, we run back into the bug where `task.annotation` points to the wrong annotation.

The identifier mismatch is almost certainly `undoManager` trying to apply a patch after annotation IDs have changed in the store. Possibly trying to reapply `classification.removeAnnotation`. `undoManager` records patches for all actions in the target tree, which means it'll be recording patches for `clearUndo`. I think that that's where the bug is occurring, and that the answer will be that `clearRedo` shouldn't be recorded in the undo/redo history.

With this PR, annotations should persist on undo/redo. We'll need that for Engaging Crowds, but it can be turned off by removing `redo()`.
>New features supported here:
>
>- recursive workflows: return to a previous step and complete it with new annotations, rather than editing existing annotations for that step.
>- persistent annotations: annotation IDs are stored in history and can be undone or reapplied with undo/redo. You should be able to move backwards and forwards through a workflow without losing any text that you've entered.

eatyourgreens:
Note to self: double-check that the classifier storybook still works.

eatyourgreens:
>At this point, we notice the error: Uncaught Error: [mobx-state-tree] A map of objects containing an identifier should always store the object under their own identifier. Trying to store key 'ckm2o6nnr000h3h62y7f568c2', but expected: 'ckm2o6l7700053h624zzdbner' (this error appears twice)

Wrapping `clearRedo` in `withoutUndo` fixed this. It looks like changes to the `classification` model can't be recorded and replayed. The underlying reason for this is that annotation snapshots don't record `annotation.id`, so a new, random ID is generated which won't match the ID in the patch path.

packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.js
eatyourgreens: If you take snapshots, `AnnotatedSteps.steps` looks something like this:
```
[
  { id: 'ckm2o6nnr000…', step: 'S2', annotations: ['ckm2o6nnr000h3h62y7f568c2', 'ckm2o6l7700053h624zzdbner'] }
]
```

It records identifiers, which point to models elsewhere in the store. MST automatically resolves those to the corresponding models in `rootStore.workflowSteps` and `rootStore.classifications`. `AnnotatedStep` here is being used to tie a step key to a list of annotation IDs. A given workflow step key might be annotated zero, once or many times so step keys aren't used as identifiers.

eatyourgreens:
@shaunanoordin bugs should be fixed now, and the task stories should also be working in the storybook.

eatyourgreens:
Moving the Task components from 'inject' to 'MobXProviderContext' broke the classifier storybook. I've included fixes for that in this PR. In retrospect, those changes should have been in their own PR.

eatyourgreens:
Trying to think of a good term for classification annotations that are in your actual past, because you sat down and entered them, but are in your classification’s future, because you’ve gone back a couple of steps but might still redo them. I’ve called them pending annotations here. I’m not sure if that’s quite the right term.

eatyourgreens:
> If a component asks for a store and receives a store via a property with the same name, the property takes precedence. Use this to your advantage when testing!

https://github.com/mobxjs/mobx/tree/main/packages/mobx-react#provider-and-inject

That might be a better approach to fixing those stories.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.spec.js
```diff
@@ -85,153 +30,135 @@ describe('TaskNavButtonsContainer', function () {
...
-      wrapper.instance().onSubmit({ preventDefault: preventDefaultSpy })
+      const buttons = wrapper.find(TaskNavButtons)
+      buttons.props().onSubmit({ preventDefault: preventDefaultSpy })
       expect(preventDefaultSpy).to.have.been.calledOnce()
     })
 
-    it('should call completeClassification', function () {
-      wrapper.instance().onSubmit({ preventDefault: preventDefaultSpy })
-      expect(completeClassificationSpy).to.have.been.calledOnce()
+    it('should call complete the classification', function () {
```
eatyourgreens, line 158:
```suggestion
    it('should complete the classification', function () {
```

eatyourgreens:
I've gone through the task code, changing `task.isComplete` to `task.isComplete(annotation)` so that we can support tasks that have a mixture of complete and incomplete annotations in the active classification. I've also removed `task.setAnnotation(annotation)` from the examples and tests, since that becomes dangerous once a task has multiple annotations.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/TaskNavButtonsContainer.js
```diff
@@ -1,113 +1,112 @@
...
+      goToPreviousStep={goToPreviousStep}
+      showBackButton={canUndo}
+      showNextButton={latest.nextStepKey()}
+      onSubmit={onSubmit}
+    />
+  )
 }
 
-TaskNavButtonsContainer.wrappedComponent.defaultProps = {
+TaskNavButtonsContainer.contextType = MobXProviderContext
```
eatyourgreens, line 86: This is leftover from the class component.

eatyourgreens:
The Back button is now wired up to use `workflow.configuration.persist_annotations` so that we enable redo for workflows that have that option checked.

eatyourgreens:
I've cleaned up the tests so that you can now mock a full store, with client, from a workflow snapshot:
```js
import mockStore from '@test/mockStore'

const store = mockStore(workflowSnapshot)
```

shaunanoordin (approved):
## PR Review (Update 2)

Follows [initial review](https://github.com/zooniverse/front-end-monorepo/pull/2050#pullrequestreview-607694710) and [follow up](https://github.com/zooniverse/front-end-monorepo/pull/2050#pullrequestreview-608226753)

Yaaarrr we're there! 👍 👍 👍 All tests are looking good, and I can't think of any other ways to break this PR.

Please refer to my earlier reviews for testing methods, etc, since at this point I'm just copy-pasting what I said earlier except with "IT WORKS NOW" where all the bug reports used to be.

Code read looks good too, and for anyone following up on this review on the future, pay close attention to `AnnotatedSteps.js` + everything it touches, as that's where a lot of the new logic comes into play.

### Status

LGTM! Let's get this chonky PR merged!

shaunanoordin:
DERP, tests are failing, but I know exactly where the problem is.

PR #1989 was recently merged (about 5 hours ago) to master, and master was recently merged to this PR. The 1989 update added parts of a Survey Task, and the new `FilterStatus.spec.js` uses `task.setAnnotation(annotation)`.

Affected file: [lib-classifier/src/plugins/tasks/SurveyTask/components/components/Chooser/components/CharacteristicsFilter/FilterStatus/FilterStatus.spec.js](https://github.com/zooniverse/front-end-monorepo/blob/4aff5a7e24c81a322388ebd46aa39314f6b08f18/packages/lib-classifier/src/plugins/tasks/SurveyTask/components/components/Chooser/components/CharacteristicsFilter/FilterStatus/FilterStatus.spec.js)

Failing line:
```
before(function () {
  types.model('MockStore', {
    annotation: Task.AnnotationModel,
    task: Task.TaskModel
  })
    .create({
      annotation,
      task
    })
  task.setAnnotation(annotation)  // Beep boop! Just delete this line and the test will run fine. 👍 

  wrapper = mount(
    <FilterStatus
      task={task}
    />
  )
})
```

eatyourgreens:
Thanks Shaun! Looks like an easy fix. Mock tasks should be passing `annotation` as a prop. https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/plugins/tasks#react-components

eatyourgreens:
Ok, CI tests should all pass now.

PR #2054, opened 2021-03-04
Project: test null project properties

eatyourgreens:
Add `researcher_quote:null` to the mock project in tests, so that we can test the project model with an invalid snapshot (see #2053.)
Check for a 404 response after creating the project model, so that `getStaticPageProps` returns early if the project model fails.

Package:
app-project

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/src/helpers/getStaticPageProps/getStaticPageProps.js
+++ b/packages/app-project/src/helpers/getStaticPageProps/getStaticPageProps.js
@@ -17,10 +17,10 @@ export default async function getStaticPageProps({ params, query }) {
   if (params.owner && params.project) {
     const projectSlug = `${params.owner}/${params.project}`
     const project = await fetchProjectData(projectSlug, { env })
-    if (!project.id) {
+    applySnapshot(store.project, project)
+    if (!store.project.id) {
       return notFoundError(`Project ${params.owner}/${params.project} was not found`)
     }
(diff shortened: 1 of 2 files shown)


packages/app-project/src/helpers/getStaticPageProps/getStaticPageProps.js
```diff
@@ -17,10 +17,10 @@ export default async function getStaticPageProps({ params, query }) {
   if (params.owner && params.project) {
     const projectSlug = `${params.owner}/${params.project}`
     const project = await fetchProjectData(projectSlug, { env })
-    if (!project.id) {
+    applySnapshot(store.project, project)
+    if (!store.project.id) {
       return notFoundError(`Project ${params.owner}/${params.project} was not found`)
     }
-    applySnapshot(store.project, project)
```
eatyourgreens, line 23: @goplayoutside3 this was the source of the `cannot read property length of undefined` error that you saw yesterday. `applySnapshot` failed here, because of the null researcher quote, so the project has no active workflows later, when this function tries to fetch linked project workflows.

I've restored the check that we had previously, which checked if the project model has an ID, rather than the returned data from the API helper.

goplayoutside3 (approved):

PR #2062, opened 2021-03-15
Refactor task navigation

### LINKED ISSUE #782: Refactor Task Nav Buttons
There is some refactoring required in the task area nav buttons (Back, Done, Done & Talk, Next):

- There is a bunch of checks for gold standard mode, demo mode etc, but we don't actually have the code for this, and currently no plans to build it, so it should really be removed.
- The nav buttons container is doing a bunch of checks and passing down props to the buttons, which it really doesn't need to do; responsibility for showing buttons and their behaviour should be devolved to the button components themselves (i.e. each button has a container connected to the store), which will make things easier to test and reason about.
- We may be able to share some common theme properties at the task nav level, rather than duplicating them within the buttons.

Follows on from #759, #781

eatyourgreens:
Refactor each of the task navigation buttons (Back, Next, Done & Talk, Done) into a view and a HOC that connects that button to the store. Remove the `TaskNavButtonsContainer` component and split the container's tests of button behaviour up into individual tests for each connected button.

Package: lib-classifier

Closes #782 .

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/store/WorkflowStepStore/Step/Step.js
```diff
@@ -56,6 +56,13 @@ const Step = types
     }
   }))
   .actions(self => ({
+    completeTasks(annotations) {
```
srallen, line 59: This should be tested in in the specs for the Step store.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButtonConnector.js
```diff
@@ -0,0 +1,42 @@
...
+          active: step
+        }
+      }
+    } = props.store || useContext(MobXProviderContext)
+
+    if (!hasNextStep) {
+      return null
+    }
+
+    function onClick() {
```
srallen, line 27: I'd prefer this to be an action in the stores. Maybe `Step` has an action responsible for completing tasks and calling next. I'd like to keep connectors simple as possible and akin to the previous injectors functions so components can potentially be more agnostic of internal store function.

srallen (changes requested):
The refactor generally looks good, but I'd prefer the new action logic written being used by the next button to be wholly in the store and to keep the connector as simple as possible.

eatyourgreens:
The Done & Talk button has broken on master, which has broken the tests for `DoneAndTalkButtonConnector` here. They're failing now because the button isn't rendered when it should be.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/NextButton/NextButtonConnector.js
eatyourgreens: Good idea. I've also wrapped up `step.completeTasks(annotations)` into a `finish()` action which is fired by Done, or Done & Talk.

eatyourgreens:
Tests are updated, so I think the only blocking issue now is to figure out why, after rebasing to the latest master, `workflowSteps.shouldWeShowDoneAndTalkButton` is false when it should be true.

srallen (approved):
Non-blocking comment: I'm not sure what the point of using a HOC approach is with `withStores` is unless you're planning to abstract it at some point? If so, would you document that in an issue?

I don't know if there's any particular project I should double check with these changes, so I'll leave that to others to have a look.

PR #2073, opened 2021-03-21
Video annotation

ErikOstlund:
Package: `lib-classifier`
Project: `Video Annotator`

This PR is a benchmark in the progress made. This feature is still a work-in-progress, but it's at a good point to commit current changes.
This update has a drawing layer that covers the video screen where users can add annotations. Currently the annotation will be visible throughout the duration of the video. This will be a future update.

### **Testing**
In the front-end-monorepo folder, run `yarn bootstrap` to build the packages. Then cd packages/lib-classifier and run `yarn dev`.

### **Items to consider during review:**

- Create a few annotations and resize the window. Annotations should remain on original point of interest in relation to the video
- Annotations can be selected, deleted, moved, resized(rectangle tool)
- Known layout issue: there is a 6px gap between the video screen and video controller. This will be addressed in a future PR.
- Unit tests will be created in a future PR.


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleVideoViewer/SingleVideoViewerContainer.js
```diff
@@ -153,20 +199,55 @@ class SingleVideoViewerContainer extends React.Component {
...
+                  tabIndex={0}
+                  viewBox={`0 0 ${naturalWidth} ${naturalHeight}`}
+                >
+                  {/* {title?.id && title?.text && (
+                  <title id={title.id}>{title.text}</title>
+                )} */}
+                  <g ref={transformLayer} transform={transform}>
+                    {enableInteractionLayer && (
+                      <InteractionLayer
+                        scale={1}
```
eatyourgreens, line 240: I think you can set the scale here to something like `clientWidth / naturalWidth`, which will scale the drawn marks down to size. At the moment, they're drawn with oversized lines and points.

eatyourgreens (approved):
Looks good to me. I can draw on a video, in Firefox and Chrome. The Show/Hide marks button works. The size of the drawn marks is a little off, so I think the initial scale of 1 might be wrong.

mcbouslog (approved):
So cool @ErikOstlund! This is looking good to me too.

Point and rectangle work well for me. Noting I did have trouble resizing the rectangle while the video was playing, but not while the video was stopped or paused. All other drawing actions with point and rectangle (initial placement, move, delete) worked well. Might be intentional, regardless, not blocking.

eatyourgreens:
Looks good to me with the scaled-down marks. 👍 

A video subject with some red points and a purple rectangle drawn on top of it.

eatyourgreens (approved):

Issue #1934, opened 2020-12-09
RFC: Make `SubTaskPopup` a child of each drawn mark


eatyourgreens:
**Package**
lib-classifier

**Is your feature request related to a problem? Please describe.**
`SubTaskPopup` is currently a child of the `InteractionLayer` component. When we want to open a subtask popup, we have to pass in the active mark and a reference to that mark's DOM node, or client bounding rectangle. This means storing DOM data in the state store, so that it can be passed between drawing tasks, marks and the popup. I think there's also some passing of refs which can be hard to follow.

**Describe the solution you'd like**
If each mark has its own subtask form as a child (see PFE for an example of this in action) then mark properties such as the bounding rectangle can be passed down from the Mark to the SubTaskPopup as props. `SubTaskPopup` uses `React.Portal`, so events fired by the popup will bubble up through the React tree, not the DOM, and can be handled in the Mark. I think this will greatly simplify the code for subtasks, particularly for transcription lines (eg. focus management when the popup opens and closes.)

```jsx
<TranscriptionLine>
  <SubTaskPopup />
</TranscriptionLine>
```

The popup could be added to the generic Mark component. I think that wraps all drawn marks.

eatyourgreens:
After working on the drawing task stories a bit more, this pattern would also allow us to write stories for the drawing tools without having to mock the `InteractionLayer` for subtasks.

srallen:
I think I experimented with this some time ago and I recall that the positioning of the subtask broke when making it a child of the mark. So if anyone else has a look at this, that's something to watch out for.

PR #2078, opened 2021-03-23
Fix focus on new transcription line subtask

srallen:

Package: lib-classifier

Toward #1978

> Cursor should automatically move to the text box after the second underline dot is placed, so you can just start typing — really annoying to have to click into the box every time.

I've added an `event.preventDefault()` when the real event is defined for the `onFinish` call. I suspect that the pointer up event is happening after the subtask opens and shifting the focus back to the SVG. I don't have high confidence in this because the onFinish handler is doing more than handling an event, given it can be undefined sometimes, and the specs were missing for pointer up events. I added a couple of specs to make sure and it doesn't seem like I'm breaking anything, but a good manual check should happen.

Before, no focus on the text area:
Screen Shot 2021-03-23 at 4 38 36 PM

After, focus on the text area:
Screen Shot 2021-03-23 at 4 39 08 PM

Testable with: https://localhost:8080/?project=1764&workflow=3392

Be sure to create a new mark to test this out. The previous mark subtask with the suggestions text input properly received focus previously.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
@@ -13,6 +13,7 @@ describe('Component > InteractionLayer', function () {
   let wrapper
   let activeTool
   const mockMark = {
+    finish: sinon.stub(),
     initialDrag: sinon.stub(),
     initialPosition: sinon.stub(),
     setCoordinates: sinon.stub(),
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Changes look good to me after a quick read of the code. I'll pull this down this morning and take a look.

Re. the specs, from what I remember I didn't add tests for `pointerup` because up events don't happen in isolation. I think I had some problems setting up a test that mocked `pointerdown` followed by `pointerup`, but I don't remember the exact details.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
```diff
@@ -180,53 +266,11 @@ describe('Component > InteractionLayer', function () {
...
-              setPointerCapture: sinon.stub(),
-              releasePointerCapture: sinon.stub()
-            }
-          }
-          wrapper.find(DrawingCanvas).simulate('pointerdown', fakeEvent)
-          wrapper.find(DrawingCanvas).simulate('pointerdown', fakeEvent)
-          expect(activeTool.handlePointerDown).to.have.been.calledOnce()
-        })
+        wrapper.find(DrawingCanvas).simulate('pointerup', fakeEvent)
+        expect(fakeEvent.target.releasePointerCapture.withArgs('fakePointer')).to.have.been.calledOnce()
```
eatyourgreens, line 273: Since writing the original code, I've discovered that the browser releases pointer capture for you, when you end a pointer interaction. I think we can get rid of `releasePointerCapture` from our code. That might also explain invalid pointer errors in Sentry, where our code is trying to release a pointer that's already been released by the browser.

>The following example shows a pointer capture being released (when a pointercancel event occurs. The browser does this automatically when a pointerup or pointercancel event occurs.

https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events#pointer_capture

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
```diff
@@ -107,57 +108,142 @@ describe('Component > InteractionLayer', function () {
...
+      before(function () {
+        mockedContext = sinon.stub(React, 'useContext').callsFake(() => { return { svg, getScreenCTM } })
       })
 
-      it('should drag the new mark on pointer down + move', function () {
+      after(function () {
+        mockedContext.restore()
+      })
+
+      it('should set the mark to finished', function () {
```
eatyourgreens, line 246: I think this is true for most drawing tools but only true for transcription tasks on the second `pointerup`. That might have been what tripped up when trying to write tests before.

eatyourgreens (approved):
Changes look good to me (tested in Firefox.) I can transcribe from the keyboard and only need the trackpad in order to create new transcription lines. I can also navigate lines with tab/shift-tab, open them with Enter and focus automatically shifts to the text box. Tab and Enter saves my work again from a popup.

The only bug I've noticed, which is a bug in the text input and not this PR, is that focus always goes to the beginning of the text box, not to the end of its text content.

If you'd like, I think you can also remove the explicit call to `target.releasePointerCapture` from `onFinish`.

eatyourgreens:
I looked a bit more into what `onFinish` is doing and I'm confused. There's a `useEffect` hook in all Marks, which calls onFinish with the mark's DOM node but no event. If I remove it, then subtask popups don't open after first creating a mark. That might be something that could be fixed by #1934. The hook runs after `onFinish` has handled the `pointerup` event and set `mark.finished`, and seems to exist in order to pass a ref to the popup somewhere.
https://github.com/zooniverse/front-end-monorepo/blob/8d37b1a8be2884bccd9c51936ba8bd1c29b04f5a/packages/lib-classifier/src/plugins/drawingTools/components/Mark/Mark.js#L42-L46

I also thought of a case where I could fire `pointerup` without a corresponding `pointerdown` to create a mark: tap/click outside the subject but near it, move the pointer onto the subject, lift your finger. That would fire `pointerup` but the code should do nothing when `creating` is false. Should we have a test for isolated `pointerup` events?

eatyourgreens:
@srallen I've got a change that fixes the calls to `onFinish` where `event` is undefined. I'll open a separate PR for that, so that this isn't held up.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
srallen: I'd prefer to keep this scoped to the focus bug fix. I can track this suggestion in an issue.

PR #2080, opened 2021-03-24
Classifier drawing tools: fix focus and window scroll position for new marks

### LINKED ISSUE #1763: Tab trap occurs when creating new mark from previous annotation
**Package** lib-classifier

**Describe the bug**
When using the transcription tools, I've noticed it's easy to get into a tab trap where I'm unable to tab through each transcription line and am instead stuck tabbing back and forth between the address bar and main window. 

Oddly enough, this _only_ happens when clicking on a previous magenta transcription to place a new transcription. If I try drawing a new line to create a transcription, I do not have this tab issue.

**To Reproduce**
1. Go to the lib-classifier package and visit https://localhost:8080/?project=1764&workflow=3392
2. Once on that project and workflow, click on a magenta line to add a new transcription
3. After closing the sub task visibility (either with the X or close button), you'll notice that pressing tab cycles between the address bar and window

Project 1764
Workflow 3392

**Expected behavior**
I should be able to continue cycling through transcription lines after placing a transcription over a previous annotation.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] macOS Mojave
 - Browser [e.g. chrome, safari] Chrome
 - Version [e.g. 22] 84

**Additional context**
This could be a wonky `onBlur` being used somewhere to release focus from the screen, but I'm not quite sure. The culprit is an action that occurs right after clicking on a magenta previous transcription line.

### LINKED ISSUE #1837: New mark created from previous mark doesn't have focus
**Package** lib-classifier

**Describe the bug**

In an effort to prevent bad classification data from being submitted (see: https://github.com/zooniverse/front-end-monorepo/issues/1719#issuecomment-663275184), https://github.com/zooniverse/front-end-monorepo/pull/1759 removed focus from the previous mark when it was selected. The new mark should be focused, but this hasn't been done yet. 

**To Reproduce**
Use the story for the transcribed lines in the storybook with keyboard to create a new mark from the previous mark.

**Expected behavior**

The new mark should have focus. Implementing this has been difficult do to the complixity of how events are managed on the marks and refs.

### LINKED ISSUE #1838: Zoom by keyboard control no longer works when previous mark has focus
**Package** lib-classifier

**Describe the bug**

Transcription task alpha comment https://github.com/zooniverse/front-end-monorepo/issues/1719#issuecomment-665072048

> Found while testing out the keyboard controls for the image toolbar in #1726. When a transcription line has focus, the keyboard controls for zooming and panning the page don't work. See #1726 (comment)

> The expected behaviour, from Seabird Watch and (I think) Notes from Nature, is that you should be able to place a mark and then pan and zoom the subject from the keyboard while the mark has focus.

**To Reproduce**
Can be observed using the Transcribed Lines story in the storybook using keyboard controls.

**Expected behavior**

Zoom controls still work while mark has focus.

eatyourgreens:
This refactors drawing tool marks to remove `onFinishWithRef` from the `DrawingToolMarks` component. I've removed `setSubTaskVisibility` from the handlers for `onFinish` and `dragEnd`. Instead, opening a subtask popup is handled by the `Mark` component and called on mark creation (when `mark.finished` changes to true), on Enter or Space for marks that have focus, or on `pointerup` for existing marks.

This should mean that refs are all handled in one place (inside `Mark`) rather than passed around between `Mark`, `DrawingToolMarks` and `InteractionLayer`. It should also fix cases where `onFinish(event)` is called with `event` undefined (see #2078.)

I've also updated focus behaviour so that marks with subtasks automatically receive focus after they're created, including marks created from transcription lines with previous annotations. You should be able to open a line with Enter/Space, select previous annotations with the arrow keys, save and close and continue to edit transcriptions without having to use a mouse.

Window scroll position is also preserved on focus (see https://github.com/zooniverse/Panoptes-Front-End/pull/4125), which fixes an annoying bug in Firefox. The subject would scroll back up to the top after transcribing a line at the bottom of a page.

Test URLs:
Transcription task: https://localhost:8080/?project=1764&workflow=3392&demo=true
Drawing task with subtasks: https://localhost:8080/?project=908&workflow=3370&demo=true

Package:
lib-classifier

- Towards #1978.
- Towards #1934.
- Closes #1838.
- Closes #1837.
- Closes #1763.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/plugins/drawingTools/components/Mark/Mark.js
```diff
@@ -39,11 +40,16 @@ const Mark = forwardRef(function Mark ({
...
 
-  React.useEffect(() => {
-    if (mark.finished && !mark.subTaskVisibility) {
-      onFinish()
+  function openSubTaskPopup() {
+    if (mark.finished &&
+      !mark.subTaskVisibility &&
+      mark.tasks.length > 0
+    ) {
+      mark.setSubTaskVisibility(true, markRoot.current)
```
eatyourgreens, line 48: I _think_ we could pass the mark's bounding client rectangle here, as the second arg, instead of a ref. I'm not sure.

eatyourgreens: Seems to work, so I've committed that change.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.js
```diff
@@ -56,8 +55,8 @@ class TranscribedLines extends React.Component {
         }
         previousAnnotationValuesForEachMark.push(previousAnnotationValuesForThisMark)
       })
-      mark.setSubTaskVisibility(true, ref?.current, previousAnnotationValuesForEachMark)
-      ref?.current?.blur()
+      mark.setPreviousAnnotations(previousAnnotationValuesForEachMark)
+      mark.finish()
```
eatyourgreens, line 59: This is the fix for #1837.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js
```diff
@@ -51,6 +51,7 @@ function SubTaskPopup(props) {
       onOpenConfirm()
     } else {
       setSubTaskVisibility(false)
+      activeMark.setPreviousAnnotations()
```
eatyourgreens, line 54: `setSubTaskVisibility(false)` was also clearing out previous annotations, so I've left that behaviour unchanged. I can remove this line if that wasn't intended.

eatyourgreens:
I tried adding some window scrolling code to this, to fix the scrolling bug described in this comment: https://github.com/zooniverse/front-end-monorepo/issues/1978#issuecomment-806695663

I found that scrolling behaviour is fine, in Firefox, for my test drawing project or the ellipse drawing task in I Fancy Cats, which show a question task as a subtask popup. Transcription tasks, on the other hand, still scroll to the top of the page even when the line with focus is out of the viewport, and even if you explicitly call `window.scrollTo` to scroll the window.

I'm going to leave this PR as is

eatyourgreens:
I've used https://github.com/zooniverse/Panoptes-Front-End/pull/4125 to fix the window scroll position in Firefox when a subtask opens. The subject still scrolls back to the top, in Firefox, when you close the popup.

eatyourgreens:
Testing this with the keyboard, I also noticed the Firefox doesn't automatically scroll to active lines when you tab through the page. I've added an `element.scrollIntoView()` to marks, on focus, to fix that.

eatyourgreens:
Previous (magenta) lines are still reachable with Tab, after adding your own transcription. Their tabindex needs to be changed from 0 to -1 after completing them (and back to 0 again if their green line is deleted.)

eatyourgreens:
Also found while testing with the keyboard: keyboard events from drawing marks are being ignored by the `withKeyZoom` component. Adding `g` to the listed of allowed tags seems to have fixed that.

That might also fix #1838.

packages/lib-classifier/src/plugins/drawingTools/components/Mark/Mark.js
```diff
@@ -39,11 +41,35 @@ const Mark = forwardRef(function Mark ({
...
+      window.scrollTo(x, y)
+    }
+  }
+
+  function openSubTaskPopup() {
+    if (mark.finished &&
+      !mark.subTaskVisibility &&
+      mark.tasks.length > 0
+    ) {
+      focusMark()
```
eatyourgreens, line 59: I think I've left this in by accident.

srallen:
I appreciate the contribution, but I'd just ask in the future that the PR stays focused on one issue or bug fix in the future. It's gets harder to review and check when the PR grows.

eatyourgreens:
Good point. I wanted to avoid having three PRs open for review, particularly if they might cause merge conflicts.

I think this one is good to go now.

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/withKeyZoom.js
```diff
@@ -2,7 +2,7 @@ import { MobXProviderContext } from 'mobx-react'
 import React, { Component, forwardRef, useContext } from 'react'
 
 function withKeyZoom (WrappedComponent) {
-  const ALLOWED_TAGS = ['svg', 'button', 'rect']
+  const ALLOWED_TAGS = ['svg', 'button', 'g', 'rect']
```
srallen, line 5: This should update the specs. I would've expected a failure since there are explicit tests for  when the target is a certain element or not

packages/lib-classifier/src/plugins/drawingTools/components/Mark/Mark.spec.js
```diff
@@ -197,45 +215,90 @@ describe('Drawing tools > Mark', function () {
...
+        window.scrollTo = sinon.stub()
+        newMark = pointTool.createMark()
+        newMark.finish()
+        wrapper = mount(markWrapper(newMark))
+      })
 
-      afterEach(function () {
+      after(function () {
         onFinish.resetHistory()
+        window.scrollTo = undefined
```
srallen, line 246: Was this previously undefined by jsdom? It'd be safer to store the original in a variable and reset this back in case what jsdom does changes in the future. If using sinon's API:
```suggestion
        scrollToStub.restore()
```

packages/lib-classifier/src/plugins/drawingTools/components/Mark/Mark.spec.js
```diff
@@ -197,45 +215,90 @@ describe('Drawing tools > Mark', function () {
...
-      }
+        </svg>
+      )
+    }
+
+    describe('with finished mark and no subTaskVisibility', function () {
+      let newMark
+
+      before(function () {
+        window.scrollTo = sinon.stub()
```
srallen, line 238: We could leverage sinon's API here if it's defined originally:
```suggestion
        scrollToStub = sinon.stub(window,  'scrollTo')
```

Or otherwise store it in a variable so we can restore the original that jsdom has set.

packages/lib-classifier/src/plugins/drawingTools/components/Mark/Mark.spec.js
```diff
@@ -197,45 +215,90 @@ describe('Drawing tools > Mark', function () {
...
 
-      describe('when the mark is not finished', function () {
-        it('should not call onFinish', function () {
-          const newMark = Object.assign({}, point, { finished: false })
-          wrapper = mount(markWrapper(newMark))
-          expect(onFinish).not.to.have.been.called()
-        })
+      after(function () {
+        onFinish.resetHistory()
+        window.scrollTo = undefined
```
srallen, line 269: Same question and comment about what jsdom expects and restoring that as well as future proofing.

packages/lib-classifier/src/plugins/drawingTools/components/Mark/Mark.spec.js
```diff
@@ -197,45 +215,90 @@ describe('Drawing tools > Mark', function () {
...
+      before(function () {
+        window.scrollTo = sinon.stub()
+        newMark = pointTool.createMark()
+        newMark.setSubTaskVisibility(true)
+        wrapper = mount(markWrapper(newMark))
+      })
+
+      after(function () {
+        onFinish.resetHistory()
+        window.scrollTo = undefined
```
srallen, line 293: Same question  / comment here too.

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
```diff
@@ -89,24 +89,26 @@ const BaseMark = types.model('BaseMark', {
...
+      if (previousAnnotationValues?.length > 0) {
+        previousAnnotationValues.forEach((previousAnnotationValue) => {
+          self.subTaskPreviousAnnotationValues.put(previousAnnotationValue)
+        })
+      } else {
+        self.subTaskPreviousAnnotationValues.clear()
+      }
+    }
+
+    function setSubTaskVisibility (visible, markBounds) {
```
srallen, line 102: Small naming quibble here, I'd use `visibility` rather than `visible` to  more clearly indicate it should be a boolean. Non-blocking suggestion as that was what it was previously.

srallen (changes requested):
Blocking questions / comments:
- Spec update for the addition of `g` to the allowed nodes for keyboard zoom
- Question about jsdom's behavior

Non-blocking:
- The activeness state of a mark doesn't always consistently change to inactive when tabbing through via keyboard. This happens when tabbing between the transcribed marks and the new marks. I can replicate on the current frontend.preview deploy, so this is preexisting. Notice in the screenshot that the focus is on another mark, but the drawing active state remains on the non-focused mark:
Screen Shot 2021-03-26 at 10 11 56 AM
- New to this branch is sometimes we get flash of focus on a new drawn mark. This only happens if another mark previously had focus:

focus-flash

It's a bit jarring to me, but maybe it wouldn't be to others so I think this is something we can let our beta testers try out and get feedback on.
- Curious about the changes to the text task spec. Seems unrelated, so would like to know what these are for. Just a refactor?

packages/lib-classifier/src/plugins/drawingTools/models/marks/Mark/Mark.js
eatyourgreens: Good catch.

packages/lib-classifier/src/plugins/drawingTools/components/Mark/Mark.spec.js
eatyourgreens: Yeah, `window.scrollTo` isn't supported by JSDOM. I'm not sure if there's a better workaround for that.

packages/lib-classifier/src/components/Classifier/components/withKeyZoom/withKeyZoom.js
eatyourgreens: Sorry, I missed that.

eatyourgreens:
>Curious about the changes to the text task spec. Seems unrelated, so would like to know what these are for. Just a refactor?

I think I should revert those. I started playing with a version of the Text task that only shows suggestions if the initial annotation value is empty, but shows an ordinary text input if you initialise it with a string of text. I was trying to figure out how suggestions work at the moment, in that they don't show when you click to edit the green line.

eatyourgreens:
I think we (@shaunanoordin and me) decided not to activate marks on focus because that would open subtasks, which would then block tabbing through the drawn lines, because focus shifts to the popup. Instead, you tab through drawn marks and activate one with Enter/Space (I think.)

packages/lib-classifier/src/plugins/drawingTools/components/Mark/Mark.spec.js
eatyourgreens: I'm seeing this warning in the terminal, but nothing actually crashes or fails. I should have changes up shortly.
```
Error: Not implemented: window.scrollTo
```

srallen:
> I think we (@shaunanoordin and me) decided not to activate marks on focus because that would open subtasks, which would then block tabbing through the drawn lines, because focus shifts to the popup. Instead, you tab through drawn marks and activate one with Enter/Space (I think.)

I think you're misunderstanding my comment. I'm not asking for the mark being tabbed to to be activated, but for the previous mark that was tabbed away from to be inactivated.

srallen:
> I think I should revert those. I started playing with a version of the Text task that only shows suggestions if the initial annotation value is empty, but shows an ordinary text input if you initialise it with a string of text. I was trying to figure out how suggestions work at the moment, in that they don't show when you click to edit the green line.

I don't understand what you're trying to change. Perhaps issue it up with examples?

eatyourgreens:
@srallen Thanks! Those changes should be addressed now. It turns out sinon will stub a method that isn't supported on the target object (I didn't know that), so that should future-proof us if `window.scrollTo` is added to jsdom.

>I don't understand what you're trying to change. Perhaps issue it up with examples?

Probably not worth an issue. I was playing with the code to figure out how the Text task knows when to show you a suggestions dropdown, and when to just use the plain text input. I've reverted those changes now.

eatyourgreens:
By the way, did you find adding new lines from the keyboard to be awkward, in that new lines are added to the end of the DOM order, which is different from the visible order on the page? I found that, after entering a new transcription from the keyboard, I expected Tab to take me to the next magenta line. Instead, I had to shift-Tab backwards to get back to where I was.

eatyourgreens:
>I think you're misunderstanding my comment. I'm not asking for the mark being tabbed to to be activated, but for the previous mark that was tabbed away from to be inactivated. ETA: The inactivation is the behavior that is inconsistent. Sometimes it is inactivated and sometimes it's not and I believe it's not consistently when going between the transcribed marks and new marks.

We could deselect the active mark on blur. Looking at the code, I think it might be safe to activate a mark on focus, because changes to the active mark don't trigger the subtask popup. That's fired by Enter/Space on key down. Could be worth experimenting with.

eatyourgreens:
If the flash of focus gets a negative reaction during testing, I think we'll have to look at overriding the default Grommet `Layer` behaviour on close. At the moment, it seems to be remembering `document.activeElement`, and restoring that on close. I'm taking advantage of that by switching focus to the new mark just before the layer opens.

eatyourgreens:
>The activeness state of a mark doesn't always consistently change to inactive when tabbing through via keyboard. This happens when tabbing between the transcribed marks and the new marks. I can replicate on the current frontend.preview deploy, so this is preexisting. Notice in the screenshot that the focus is on another mark, but the drawing active state remains on the non-focused mark:

Agh, I get this now. Apologies. This is related to my earlier comment that magenta lines should change tabindex from 0 to -1 after you've added a green line. It looks like you're tabbing to/from a green line, but you're actually tabbing to the magenta line underneath it. When you've tabbed through all the magenta lines and reach the green lines, then active status does change as expected when you tab between the new lines.

I think that what should happen is that the magenta line should stop being focusable, and the green line should replace it in the DOM/Tab order. When/if the green line is deleted, the magenta line should be restored to tabindex 0 again. I had conversations with @shaunanoordin about a year ago, about how ASM handled this. Essentially, there has to be some sort of association between the magenta line and the green line that's created from it, so that we can replace one with the other as required.

eatyourgreens:
Re. my previous comment, I think there might be a subtle bug on staging (probably production too): create a bunch of green transcription lines from previous magenta lies. Now tab through the page and I think you should be able to tab though each magenta line in turn and add a new transcription (by pressing Enter), even if there's already a green line for that magenta line.

That's the scenario that Shaun was telling me about last year, discussing how ASM should replace previous marks rather than just drawing on top of them. That would make the `TranscribedLines` component more complex.

eatyourgreens:
I've opened issues for creating multiple transcription lines from a single previously transcribed line (#2082) and the confusing tab order (#2081.) #2083 suggests some changes which would restrict each magenta line so that it can only create one transcription line.

I've also removed some unused refs from the `TranscribedLines` component.

srallen:
@eatyourgreens is this ready for review again? Are there new changes that I should consider in a new review? It's unclear to me since this PR doesn't clearly address a single issue. Thanks.

eatyourgreens:
It's ready for review again. Just the changes from your review and I took out some cruft.

eatyourgreens:
Focus should be fixed and the annoying window scroll in Firefox.

eatyourgreens:
Sorry, I accidentally deleted an extra line when removing cruft. It's restored now and this should definitely be ready. I've checked it in Chrome and Firefox.

srallen (approved):
LGTM. Thanks!

PR #2097, opened 2021-04-12
Fix Done & Talk for single answer tasks

eatyourgreens:
The workflowSteps store creates or overwrites `step.next` for each workflow step, which then breaks the Done & Talk button for single choice questions. This PR removes that code so that the value of `step.next` is taken from `workflow.steps`, for each step.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/WorkflowStepStore/WorkflowStepStore.js
+++ b/packages/lib-classifier/src/store/WorkflowStepStore/WorkflowStepStore.js
@@ -130,16 +130,9 @@ const WorkflowStepStore = types
     }
 
     function setSteps (steps) {
-      steps.forEach(([ stepKey, step ], index) => {
-        let next
-        // checking for there being a next step without accessing it directly
-        // mobx throws errors for undefined indices because you can't observe those
-        if (steps.length > 0 && index + 2 <= steps.length) {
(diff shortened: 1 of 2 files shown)


mcbouslog (approved):
LGTM, thanks! Confirmed fix on applicable workflow, TESS Beta no change.

eatyourgreens:
Noting that this also fixed a broken workflow for the Engaging Crowds project Scarlets & Blues, which we discovered after the PR merged. The code removed here overwrote `step.next` in their branching workflows, so that the workflow was linked up as a linear set of steps. This broke steps that ended the workflow: the Done button isn't shown if `step.next` exists for a step (and isn't set to `summary`.)

There's a test case for that scenario in the specs for `convertWorkflowToUseSteps`. A single choice branching workflow which can be ended either by a drawing step or a multiple choice step. I think we should consider testing the `WorkflowStepsStore` model with these cases as inputs too.
https://github.com/zooniverse/front-end-monorepo/blob/813f0334eb9543a9163ac5c312c7654722f3458f/packages/lib-classifier/src/store/helpers/convertWorkflowToUseSteps/convertWorkflowToUseSteps.spec.js#L170-L217

PR #2103, opened 2021-04-16
Remove git commit history from the Docker build

eatyourgreens:
Ignore `.git` during Docker builds. I've also updated the top-level Dockerfile so that only yarn and  the source packages are copied across to an image, before building the libraries and apps.

I don't think we've got anything sensitive in the git history, but [removing it from the build context](https://codefresh.io/docker-tutorial/not-ignore-dockerignore-2/) seems like a good idea.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen (commented):
I did see in the logs `fatal: Not a git repository (or any of the parent directories): .git` while building... is this from the `commitID` function failing? I can build successfully otherwise.

eatyourgreens:
Yeah, if `COMMIT_ID` isn’t specified for a build then the `git` command will fail. We specify it in the build environment on Jenkins and in GitHub actions.

srallen (approved):
:+1: sounds good. Thanks.

PR #2104, opened 2021-04-16
Classifier: Set step.next for transcription tasks

### LINKED ISSUE #2102: Transcription Task workflows do not work with additional tasks
## To Reproduce
1. Go to the [Davy Notebooks Project lab page](https://www.zooniverse.org/lab/9006/workflows/18244) and activate Workflow #18244  
2. Then visit https://frontend.preview.zooniverse.org/projects/humphrydavy/davy-notebooks-project
3. Click on "Davy Notebooks Beta Test Workflow (2021)"
3. Respond to the Transcription Task prompts and click next
4. See error (does not move past Transcription Task screen to T2)

## Expected behavior
After the question task "Have all the volunteer-made underline marks turned grey?" the volunteer should click "Next" and be taken to the next Task in the workflow.

 - OS: Mac OS Catalina
 - Browser: Chrome
 - Version: 89

## Panoptes info
Project: 9006
Workflow: 18244

## Additional context
The team was hoping to go to beta on April 20, but I've told them we'd need to delay until April 27. Note that this bug may also be affecting any NEH institute teams using the Transcription Task (though I haven't had any additional bug reports other than the DNP team).

eatyourgreens:
Transcription workflows have steps defined, so they bypass `convertWorkflowToUseSteps`. However, the Next Task selector in the project builder still sets `task.next` to a task key for each task and single choice answer. This PR runs `convertWorkflowToUseSteps` on all incoming workflow snapshots. It checks `step.next` for each step, and tries to set it from `task.next` if it doesn't exist.

https://localhost:8080/?env=production&project=9006&workflow=18244&demo=true

Package:
lib-classifier

Closes #2102.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/helpers/convertWorkflowToUseSteps/convertWorkflowToUseSteps.js
+++ b/packages/lib-classifier/src/store/helpers/convertWorkflowToUseSteps/convertWorkflowToUseSteps.js
@@ -23,14 +23,14 @@ function getStepKeyFromTaskKey(steps, taskKey) {
   return taskStepKey
 }
 
-function getTaskKeysIncludedInComboTasks (tasks) {
+function getTaskKeysIncludedInComboTasks(tasks) {
   let taskKeys
   const comboTasks = Object.values(tasks).filter(task => task?.type === 'combo')
   taskKeys = comboTasks.map(combo => combo.tasks)
   return taskKeys.flat()
 }
 
-function isThereBranching (task) {
+function isThereBranching(task) {
   return task?.answers?.some((answer, index) => {
     if (task.answers.length > index + 1) {
       return answer.next !== task.answers[index + 1].next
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Looking at this again, maybe it would be better to put these fixes into `convertWorkflowToUseSteps` and run that across all workflows, even ones that do have `steps` defined.

srallen (changes requested):
Blocking this because I think the fix should go into the lab. When the new lab page is built, would this be removed? I'd prefer not editing the classifier for a scenario of configuration that shouldn't exist to begin with.

eatyourgreens:
I think the plan is to remove `convertWorkflowToUseSteps` eventually. It was originally part of the `WorkflowSteps` model and my understanding is it's only needed for workflows that are expected to work in PFE?

srallen:
Ok, for posterity, after I got more fully caught up, the issue for #2102 isn't how the lab works, but with the slightly differing behavior between `AnnotatedSteps` and `WorkflowStepStore`. I've opened #2107 as an alternative fix to this because it addresses the differing behaviors and refactors `step.getNextStep` to derive the next step using the way `WorkflowStepStore` built it to do so.

eatyourgreens:
I'm afraid #2107 introduces a regression that was fixed in #2097 for branching workflows. That bug isn't present on this branch. The missing piece in #2107 is a check on each step to see if it's the last step the workflow. Branching workflows can have more than one last step.

I started down the same route here, initially. The first version of this PR modified the `WorkflowSteps` model to check each step for a next step. However, I realised that I was duplicating code from the `Workflow` model, and the fix was simpler if I added ~5 new lines to `convertWorkflowToUseSteps`, so I rewrote this PR and went down that route instead.

packages/lib-classifier/src/store/helpers/convertWorkflowToUseSteps/convertWorkflowToUseSteps.js
```diff
@@ -97,8 +97,21 @@ export default function convertWorkflowToUseSteps ({ first_task, tasks }) {
     }
   })
 
+  return steps
+}
+
+export default function convertWorkflowToUseSteps ({ first_task, steps = [], tasks }) {
+  steps = steps.length > 0 ? steps :createStepsFromTasks(first_task, tasks)
```
srallen, line 104:
```suggestion
  steps = steps.length > 0 ? steps : createStepsFromTasks(first_task, tasks)
```

srallen (approved):
There's a typo, but it doesn't seem to be breaking anything.

PR #2107, opened 2021-04-19
Refactor step.getNextStep to use workflowSteps store

### LINKED ISSUE #2102: Transcription Task workflows do not work with additional tasks

srallen:

Package: lib-classifier

Closes #2102 

Two issues are the cause of #2102:

We have two stores that sort of do the same thing, `WorkflowStepsStore` and `AnnotatedSteps`. The issue with this is that they both sort of control the step behavior, but behave slightly differently and make different assumptions.

`WorkflowStepsStore` assumes the step order is generally the order of the map. `step.next` is only used if the step is configured to recurse. Prior to the refactor to use `AnnotatedSteps`, `step.next` was also set and used when there was single choice branching. This behavior is consistent with the concept laid out in [ADR 5](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-05.md).

`AnnotatedSteps` was built to handle step annotation history, but took over control of the back/next/done behavior of the task step UI. It assumes `step.next` is always defined unless there is a single choice task branching, in which, `answer.next` would take precedence. 

Two issues arise from this: 

- The `Step` model is a child node of both `WorkflowStepsStore` and `AnnotatedSteps`. When a child of `WorkflowStepsStore`, the `answer.next` value of a pre-existing PFE style workflow gets correctly converted from the next task to the next step. When `Step` is a child node of `AnnotatedSteps`, no such conversion occurred and `answer.next` tried to select a PFE style next task rather than a next step. (Note: PR #2104 proses a fix to have all workflow types, whether for PFE or classifier 2.0, to go through the converter and always set `step.next`.)
- The differing assumptions of how the next step is derived has longer term impact on other development, including the workflow lab page, where right now, the plan is to follow the logic laid out in ADR 5. The interim lab editor for the transcription task followed this, and does not set the next step by `step.next`, but instead relies on nested arrays which gets converted to a map in the store to set the order.

This PR proposes an alternative fix to #2104:
- Start to use `workflowSteps.getNextStepKey` again rather than have it be unused cruft and derive the next step by step map order or use `step.next` if defined (only for recursion).
- Use `workflowSteps.active.tasks` for when there is a single choice task branch because the next task key is correctly converted already here, so we don't need to do it again. 

Long term I'd prefer we consolidate this logic more so we don't have two stores doing slightly similar things but behave enough differently to introduce bugs. 

This can be tested with Davy Notebooks:
https://localhost:8080/?env=production&project=9006&workflow=18244&demo=true

The workflow now flows through all of the defined steps as expected. 

And by the testing kitty project, to confirm we don't have a regression with single choice branching and PFE workflow conversion:
https://localhost:8080/

When a choice isn't selected, disable the buttons:
Screen Shot 2021-04-19 at 3 06 11 PM

When none is selected, end the classification:
Screen Shot 2021-04-19 at 3 06 17 PM

When 1 is selected, go to the next step and task:
Screen Shot 2021-04-19 at 3 06 23 PM

As for the differing assumptions, I'm happy to discuss changes from what we laid out in ADR 5, but this should be a discussion first and if we arrive at a idea to try or a decision, let's document it so it can be communicated across the team so we don't run into issues like this in the future.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.spec.js
+++ b/packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.spec.js
@@ -1,12 +1,4 @@
-import { Factory } from 'rosie'
-import sinon from 'sinon'
-
-import AnnotatedSteps from './'
-
-import RootStore from '@store'
-import { SubjectFactory, WorkflowFactory, ProjectFactory } from '@test/factories'
 import mockStore from '@test/mockStore'
-import stubPanoptesJs from '@test/stubPanoptesJs'
 
 describe('Model > AnnotatedSteps', function () {
   let store
--- a/packages/lib-classifier/src/store/WorkflowStepStore/Step/Step.js
+++ b/packages/lib-classifier/src/store/WorkflowStepStore/Step/Step.js
@@ -1,4 +1,4 @@
-import { types } from 'mobx-state-tree'
+import { getRoot, types } from 'mobx-state-tree'
 import taskRegistry, { taskModels } from '@plugins/tasks'
 
 function taskDispatcher (snapshot) {
@@ -45,14 +45,23 @@ const Step = types
     },
 
     nextStepKey(annotations = []) {
+      const { workflowSteps } = getRoot(self)
       // assume only one branching question per step.
-      const [ singleChoiceAnnotation ] = annotations.filter(annotation => annotation.taskType === 'single')
-      if (singleChoiceAnnotation) {
-        const [singleChoiceTask] = self.tasks.filter(task => task.taskKey === singleChoiceAnnotation.task)
-        const selectedAnswer = singleChoiceTask?.answers[singleChoiceAnnotation.value]
-        return selectedAnswer?.next
+      if (self.isThereBranching) {
+        const [singleChoiceAnnotation] = annotations.filter(annotation => annotation.taskType === 'single')
+        if (singleChoiceAnnotation) {
+          // We use the workflowSteps' active step tasks 
+          // because they've been correctly converted for backwards compatibility
+          const [singleChoiceTask] = workflowSteps.active.tasks.filter(task => task.taskKey === singleChoiceAnnotation.task)
+          const selectedAnswer = singleChoiceTask?.answers[singleChoiceAnnotation.value]
+          return selectedAnswer?.next
+        }
+
+        return undefined
       }
-      return self.next
+
+      // Get next step in map order or step.next if recursing
+      return workflowSteps.getNextStepKey()
     }
   }))
   .actions(self => ({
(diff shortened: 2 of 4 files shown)


eatyourgreens (changes requested):
The National Archives told Sam that their branching workflows for their Engaging Crowds project,  Scarlets & Blues, were broken last week but then fixed after #2097 merged. The reason for that is that the code removed in #2097 wires up branching workflows as if they are a single, linear chain of steps. The last step on each branch should have `step.next` undefined, but instead is linked up to the next step in the steps map.

This PR re-introduces the behaviour that was removed in #2097, re-opening that bug as a regression.

I have a reproducible test case here:
https://localhost:8080/?demo=true&project=908&workflow=3488
Each of the tasks `T0` and `T1` is set to submit the classification and load the next subject. However, the code in this PR overrides those settings and links up `T0` to `T1`. Here it shows a Next button, not a Done button as expected.
Screenshot of a multiple choice task, asking you to pick a colour and press Done. Instead of Done, the task navigation shows Next.

There's a test case in the specs for `convertWorkflowToUseSteps` where `step.next` should be undefined for all three steps in a branching workflow. If it's added to the tests here, I think we'd expect `step.next` to be undefined for `S1` and `S2`.
https://github.com/zooniverse/front-end-monorepo/blob/7775f105d9c9b6d461da856c3cc937c605556368/packages/lib-classifier/src/store/helpers/convertWorkflowToUseSteps/convertWorkflowToUseSteps.spec.js#L187-L216

srallen:
I had a revision that involved specifying when the workflow ends, using `'summary'` string as we've prepped for, and it fixed the regression, but the backwards compatibility (which I think is a mistake to build so much for) is overcomplicating this, specifically tasks out of order. Rather than spend more time on this, I'll review #2104 for a temporary fix, but the longer term fixes are:

- Consolidating the logic / removing cruft between `WorkflowStepStore` and `AnnotatedSteps`
- Adding support to the lab to properly set the next step order with branches

PR #2101, opened 2021-04-15
Classifier: add the subjectID prop

eatyourgreens:
Refactor the `Classifier` component to use hooks and add the `subjectID` prop, which allows us to classify a specific subject. When a subject ID (or list of subject IDs) is specified, the subjects store requests subjects from `subjects/selection` rather than `subjects/queued`.

I've also updated `stubPanoptesJS` so that different responses can be specified for `/subjects/queued`, `/subjects/grouped` and `/subjects/selection`. That's useful when you want to test that selection is using a specific endpoint.

https://localhost:8080/?demo=true&workflow=693&subject=4237
https://localhost:8080/?demo=true&workflow=693&subject=4218

Package:
lib-classifier

Towards: #1876.
Closes: #2074.

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2141, opened 2021-04-30
Classifier: fix store re-renders

### LINKED ISSUE #2140: Engaging Crowds: workflows don't load on the Classify page
## Package
lib-classifier

## Describe the bug
https://frontend.preview.zooniverse.org/projects/msalmon/hms-nhs-the-nautical-health-service/classify
Choose a workflow, then a subject set. The classifier stays empty and this error is thrown.
```
ReferenceError: unable to load workflow 18110 for project undefined
```

## To Reproduce
1. Go directly to a project's Classify page
2. Choose a workflow and then a subject set to work on

## Expected behavior
The selected workflow should load in the classifier

## Additional context
The workflow selector works if you choose a workflow and subject set from the project home page. The issue here seems to be caused by the classifier's props changing while the classifier is mounted.

eatyourgreens:
Define a global store for the `useStore` hook, so that useStore doesn't generate a new, empty store on every re-render of the Classifier. Check for a valid project before updating the store on project change. Add the passed project prop to the initial store snapshot.

When I added the `useStore` hook to the classifier in #2101, I missed that [`store` should be global](https://github.com/vercel/next.js/blob/5201cdbaeaa72b54badc8f929ddc73c09f414dc4/examples/with-mobx-state-tree/store.js#L4). This PR fixes that oversight, and updates the other classifier hooks so that we never initialise the classifier store with an empty project (if there's a `project` prop.)

To check this against the original issue, you'll need to build the classifier, then start up the project app (with the newly built classifier.) Load up a page which mounts the classifier but waits for classifier props to change before loading a subject:
http://local.zooniverse.org:3000/projects/msalmon/hms-nhs-the-nautical-health-service/classify?env=production&demo=true

Package:
lib-classifier

Closes #2140.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/Classifier.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.js
@@ -45,12 +45,16 @@ const client = {
 // So we'll unregister the worker for now.
 unregisterWorkers('./queue.js')
 
+let store
+
 function initStore({ authClient, client, initialState }) {
-  const store = RootStore.create(initialState, {
-    authClient,
-    client
-  })
-  makeInspectable(store)
+  if (!store) {
+    store = RootStore.create(initialState, {
+      authClient,
+      client
+    })
+    makeInspectable(store)
+  }
   return store
 }
 /**
@@ -58,8 +62,8 @@ function initStore({ authClient, client, initialState }) {
   https://github.com/vercel/next.js/blob/5201cdbaeaa72b54badc8f929ddc73c09f414dc4/examples/with-mobx-state-tree/store.js#L49-L52
 */
 function useStore({ authClient, client, initialState }) {
-  const store = useMemo(() => initStore({ authClient, client, initialState }), [authClient, initialState])
-  return store
+  const _store = useMemo(() => initStore({ authClient, client, initialState }), [authClient, initialState])
+  return _store
 }
 
 export default function Classifier({
@@ -77,7 +81,14 @@ export default function Classifier({
   const classifierStore = useStore({
     authClient,
     client,
-    initialState: {}
+    initialState: {
+      projects: {
+        active: project.id,
+        resources: {
+          [project.id]: project
+        }
+      }
+    }
   })
 
   const {
@@ -94,8 +105,10 @@ export default function Classifier({
   }, [])
 
   useEffect(function onProjectChange() {
-    projects.setResources([project])
-    projects.setActive(project.id)
+    if (project.id) {
+      projects.setResources([project])
+      projects.setActive(project.id)
+    }
   }, [project.id])
 
   useEffect(function onURLChange() {


shaunanoordin (commented):
## PR Review

Package: `app-project`, `app-classifier`
Fixes #2140 

This PR fixes an issue where workflows don't load properly when you access a multi-workflow project's general `/classify` page.

- The code read makes sense - the solution is to create a global `store` value that the root Classifier store _will not_ re-initialise _if_ it already exists.
- I'm a bit fuzzy on how Next.js handles the _scope_ for these globals when there server-side rendering, but that's something to untangle. In practice (see Testing), I can see it's addressing the functionality issues documented in 2140

Code read looks good, and _most functionality_ looks fine.

### Testing

Tested on localhost using macOS 10 + Chrome, using a specific production project. Ran `yarn bootstrap` since we're testing changes to _lib-classifier_ using _app-project._ I definitely remembered to do this and did not waste 20 minutes running tests on outdated code, no sir.

URLs for testing:
- (A) Project home: http://local.zooniverse.org:3000/projects/msalmon/hms-nhs-the-nautical-health-service?env=production&demo=true
- (B) General classify: http://local.zooniverse.org:3000/projects/msalmon/hms-nhs-the-nautical-health-service/classify?env=production&demo=true
- (C) Specific workflow+subject set classify: http://local.zooniverse.org:3000/projects/msalmon/hms-nhs-the-nautical-health-service/classify/workflow/18109/subject-set/82738?env=production&demo=true
- (C2) Specific workflow+subject set classify of **a different WF+SSet:** http://local.zooniverse.org:3000/projects/msalmon/hms-nhs-the-nautical-health-service/classify/workflow/18110/subject-set/83397?env=production&demo=true

Scenario 1:
- start from (A) project home page
- use Workflow selector to pick WF 18109, then Subject Set selector to pick SSet 82738
- I'm taken to the (C) specific classify page.
- Classifier loads correctly on both master and PR 👍 

Scenario 2:
- start from (B) general classify page
- use Workflow selector to pick WF 18109, then Subject Set selector to pick SSet 82738
- I'm taken to the (C) specific classify page.
  - On `master`, this results in an error: _Unhandled Runtime Error. Error: MobX Provider: The set of provided stores has changed._
  - With this PR, Classifier (with WF + SSet) loads correctly 👍 

Scenario 3: 
- go directly to (C) specific classify page
- Classifier loads correctly on both master and PR 👍 

Scenario 4:
- start from (A) project home page
- click on the 'Classify' link to go to the (B) general classify page
- use Workflow selector to pick WF 18109, then Subject Set selector to pick SSet 82738
- I'm taken to the (C) specific classify page.
  - On `master`, this results in an error: _(same thing)_
  - With this PR, Classifier (with WF + SSet) loads correctly 👍

Scenario 4a:
- Perform the same actions as Scenario 4. And then...
- click on the 'Classify' link to go to the (B) general classify page
- choose a _different WF + SSet_ 
- I'm taken to the (C2) specific classify page...
  - but with this PR, I get the error: `There was an error in the classifier :( TypeError: Cannot read property 'isComplete' of undefined` ❌ 

Scenario 5:
- go directly to (C) specific classify page
- click on the 'Classify' link to go to the (B) general classify page
  - On `master`, this results in an error: _(same thing)_
  - Side note: with this PR, we can also see that the WF + SSet loaded in the classifier in the background, meaning those values are preserved as the WF selector and SSet selector components are shown. Nice.
- choose a **different WF + SSet** 
- I'm taken to the (C2) specific classify page...
  - but with this PR, I get the error: `There was an error in the classifier :( TypeError: Cannot read property 'isComplete' of undefined` ❌ 

Scenario 5a:
- Basically scenario 5, but at step 3, choose the _same WF + SSet_
- So we go from (C) to (B) to (C) again.
- with this PR, Classifier (with WF + SSet) loads correctly 👍

UPDATE: Additional tests with Scenario 4a and 5 show that the `Cannot read property 'isComplete' of undefined` always occurs when changing _workflow._ Changing subject sets seems to have no issue, which makes sense. I think it's _Task.isComplete_ that's tripping the system.

### Status

Ooh, this is a tricky one. Strictly speaking, **this PR does what it intends to do,** which is fix #2140. 👍  However, as we can see in Scenarios 4a and 5, there is an issue when you _choose one WF, then switch to a different WF._ 👎 

The `Cannot read property 'isComplete' of undefined` _might_ be a straightforward fix that can be addressed in this PR (I think something's causing the new WF's Tasks to not render properly, or maybe the old WF's Tasks are still lingering when the old WF is removed, and the error occurs at the edge of the old WF being unloaded and the new PR being loaded) but I genuinely won't know until I dig deeper.

Alternatively, it's safe to give this PR an Approve, (it's not _breaking anything new_ ) and then follow up on that specific bug as a separate fix. How do you want to play this, @eatyourgreens ?

eatyourgreens:
I'd like to open an issue for that other bug. It could be `task.isComplete` that's broken, or maybe `step.isComplete`.

eatyourgreens:
I can reproduce that other bug. I'm seeing `e.step is undefined`, so I think `step.isComplete(annotations` is broken. Maybe changing the active workflow, while the classifier is mounted, gets the store into a weird state in between the two workflows? We've never had a case on any of the live projects where the classifier's props might change, leading to a store update. That's a scenario that can't be debugged in the dev classifier.

Here's the Sentry error from my development build: https://sentry.io/share/issue/3b2302e0866a47b9ba7857c632c9d031/

Screenshot of the classifier showing the error message: e.step is undefined

eatyourgreens:
I've opened #2154 for that other bug and I'm looking into it. I'd like to get this fix in for HMS NHS, if it fixes the original issue. It looks like the other bug might take a while to figure out.

eatyourgreens:
#2155 should fix that bug, but might need this to be merged before it can be tested with a real project.

shaunanoordin (approved):
## PR Review (Update)

Following up from the initial review, I'm giving this a 👍 to get merged. The "classifier crashes when Workflow changes" issue is logged in #2154 with a potential fix in #2155 . Let's get this PR in so we can start testing that new PR!

PR #2160, opened 2021-05-07
Classifier: catch git errors during builds

eatyourgreens:
`gitCommit()` fails on production Docker builds because `.git` is excluded from the Docker image now. This PR wraps gitCommit() in try/catch. The Jenkins build should pick up the commit ID from the build environment, with `gitCommit()` setting a default value.

Here's an example of a broken build:
https://jenkins.zooniverse.org/job/Zooniverse%20GitHub/job/front-end-monorepo/job/master/1114/console

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/webpack.dev.js
+++ b/packages/lib-classifier/webpack.dev.js
@@ -4,8 +4,13 @@ const path = require('path')
 const webpack = require('webpack')
 
 function gitCommit() {
-  const commitHash = execSync('git describe --always').toString('utf8').trim()
-  return commitHash
+  try {
+    const commitHash = execSync('git describe --always').toString('utf8').trim()
+    return commitHash
+  } catch (error) {
+    console.log(error)
+    return 'Not a git repository.'
+  }
 }
 
 const EnvironmentWebpackPlugin = new webpack.EnvironmentPlugin({
--- a/packages/lib-classifier/webpack.dist.js
+++ b/packages/lib-classifier/webpack.dist.js
@@ -3,8 +3,13 @@ const path = require('path')
 const webpack = require('webpack')
 
 function gitCommit() {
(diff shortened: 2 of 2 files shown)


mcbouslog (approved):
LGTM, thanks!

camallen (approved):

packages/lib-classifier/webpack.dev.js
```diff
@@ -4,8 +4,13 @@ const path = require('path')
 const webpack = require('webpack')
 
 function gitCommit() {
-  const commitHash = execSync('git describe --always').toString('utf8').trim()
-  return commitHash
+  try {
+    const commitHash = execSync('git describe --always').toString('utf8').trim()
```
camallen, line 8: OOI - could this be 
```
const commitHash = process.env.COMMIT_ID || 'Not a git repository'
```

eatyourgreens: webpack's EnvironmentPlugin should take care of that.
https://webpack.js.org/plugins/environment-plugin/#use-case-git-version

As of webpack 5, you're advised not to reference `process.env` directly but instead use `EnvironmentPlugin` or `DefinePlugin` to set the environment.

camallen: got it - so the `gitCommit()` function should only run when the `process.env.COMMIT_ID` is missing (i.e. the default fallback), https://webpack.js.org/plugins/environment-plugin/#usage-with-default-values

eatyourgreens: You’re right though: if we don’t need the commit hash in local development, then this could be replaced by a plain string.

Issue #2100, opened 2021-04-13
Transcription task somtimes sends classifications with no text task


CKrawczyk:
## Package
lib-classifier

## Describe the bug
Sometimes the transcription task creates a classification that has a line, but no associated text task (e.g. with `details: []`).  Looking at the classification data of the lines with no text they seem to be **duplicates** of lines that do have text (see example below).

## To Reproduce
I am unsure how to reproduce this bug, but some example classifications are:

```
id: 310540545
"user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.182 Safari/537.36"
"subject_id": 55414222
"user_id": None
"workflow_id": 13898
---
"id": 310580237
"user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.150 Safari/537.36"
"subject_id": 55414317
"user_id": None
"workflow_id": 13898
---
"id": 310607565
"user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_5) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.1.1 Safari/605.1.15"
"subject_id": 55414433
"user_id": 2158146
"workflow_id": 13898
---
"id": 310624327
"user_agent": "Mozilla/5.0 (Linux; Android 11; SM-G998B) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.181 Mobile Safari/537.36"
"subject_id": 55414216
"user_id": 1654808
"workflow_id": 13898
---
"id": 310641679
"user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/88.0.4324.152 Safari/537.36"
"subject_id": 55414220
"user_id": 1988515
"workflow_id": 13898
---
"id": 311198050
"user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:85.0) Gecko/20100101 Firefox/85.0"
"subject_id": 55414309
"user_id": 2237331
"workflow_id": 13898
---
"id": 311199130
"user_agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:85.0) Gecko/20100101 Firefox/85.0"
[...]

PR #2167, opened 2021-05-11
Add TranscriptionLine mark validation

srallen:

Package: lib-classifier

Toward #2100. Note this doesn't quite fix the reported bug in #2100 because volunteers can create a partially completed mark without a sub-task transcription and submit the classification still. Fixing this will be coming in a follow up PR.

The `TranscriptionLine` mark model was still set to be valid all the time. This adds validation to the mark. Also prevents the `onPointerUp` event handler on the `Mark` component from firing which opens the sub-task modal if the mark is deleted because of being invalid.

As an aside, all of the specific drawing mark models are missing specs. I have no idea why, but maybe this would've been caught sooner if we did.

Anyway, the current behavior allows volunteers to create transcription line marks with the exact same point for the two points:

transcription-line-single-point

Try this out at by clicking once on the subject to create the mark and click in the exact same spot, add a transcription, and close the subtask. The mark will have the same points for both ends: https://frontend.preview.zooniverse.org/projects/blicksam/transcription-task-testing/classify/workflow/13898?demo=true

This branch does not allow this:

transcription-line-invalid

And deletes the mark if you click in the exact same spot. Try it out at: https://localhost:8080/?project=11300&env=production&demo=true

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


goplayoutside3 (approved):
Tests and storybook look good, and confirmed expected behavior of deleting a `Mark` if clicking the same spot twice!

PR #2176, opened 2021-05-14
Project: remove inject from the classifier wrapper

eatyourgreens:
Refactor the Classifier wrapper, which loads the classifier into the Classify page, to use `MobXProviderContext` instead of `inject`. I've used a HOC pattern to add store props, so that the tests can continue to run against the undecorated component.

Package:
app-project

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen (commented):
Are you intending to abstract out the HOC that I see you reusing at some point? If not, I find it harder to read as well as think it'll be harder to swap out MST if we need to by keeping the store connections in the same file. My intention of having separate files was to make it as easy as possible to swap out MST if we find that we need to in the future.

eatyourgreens:
The HOC was a quick refactor, to replace the `inject` function without having to rewrite `ClassifierWrapper` or write new tests. Abstracting it out doesn't make much sense, to me, since the store mappings are unique to each connected component.

Looking back at this,I think the classifier wrapper could do with some documentation too.

eatyourgreens:
OK, I've split the wrapper/container into separate components and had an initial go at documenting them, including a usage example that's copied straight from the project classify page.

eatyourgreens:
Having said that, about premature abstraction, a reusable `useStore` hook is starting to look like it might be useful.

srallen:
A HOC pattern as I see it is meant to be a reusable abstraction. If it can't actually be abstracted, then let's just have clearly separate files. I'd like to see at a glance if a component is connected and reliant on a store, and a separate connector file will show that.

srallen (approved):
LGTM

PR #2184, opened 2021-05-18
Classifier: Updated focus management for modal popups

### LINKED ISSUE #2183: Opening a modal or dropdown from a modal breaks keyboard tabindex for the whole page
## Package
app-project
lib-classifier

## Describe the bug
On a page with previously transcribed lines, click on a purple line and complete it by filling out the subtask box. After closing the subtask, all focusable elements in the classifier have `tabindex=-1`, as if disabled. This includes the image toolbar buttons and radio inputs in the task area.

## To Reproduce
1. Go to https://frontend.preview.zooniverse.org/projects/blicksam/transcription-task-testing/classify/workflow/13898
2. Click on a purple line (you might need to refresh to get a subject which isn't 100% complete.)
3. Choose an option from the dropdown and close the subtask box.
4. Try navigating to any other part of the classifier with Tab.

## Additional context
I've been able to reproduce this bug, in the dev classifier, as far back as commit 858ab002.

eatyourgreens:
The subtask confirmation modal opens from a modal popup, so the page is already disabled. This removes modal behaviour so that the page isn't disabled twice, triggering #2183.

To test this, create and save an empty transcription:
https://localhost:8080/?project=11300&workflow=13898&env=production
After the confirm dialog closes, you should still be able to tab through the classifier controls.

With the confirm modal open, the keyboard now tabs through the browser chrome, the active mark and both modals. I'm not sure if that's the behaviour that we want. ~~Should the text task modal be disabled while the confirm modal is open? If so, I can close this and maybe work on a fix in Grommet itself.~~ This seems impossible to fix in Grommet, without introducing more focus bugs, so I've disabled the task popup while confirmation is in progress.

~~#2183 can also be triggered by the text inputs with suggestions dropdowns, so isn't closed by this.~~ I've fixed dropdowns by adding the `trapFocus: false` prop. You can test that by adding a subject to a collection, then tabbing through the page, or classifying on a transcription task where you choose from other volunteers' transcriptions of a line.

To reproduce the bug, add a subject to a collection on PH-TESS, then try to tab through the Classify page, You should find that keyboard navigation is impossible after using the dropdown menu.
https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/classify/workflow/11235

Package:
app-project
lib-classifier

Closes #2183.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen:
> Should the text task modal be disabled while the confirm modal is open? If so, I can close this and maybe work on a fix in Grommet itself.

Yes, I think it needs to be a modal. Without it being one, you can get into a confusing state like this:

Screen Shot 2021-05-18 at 9 41 48 AM

Where the original subtask is still interactive.

mcbouslog:
Are we thinking this should be merged until a Grommet fix can be implemented or should this be closed?

mcbouslog:
I lean towards creating a new issue once this merges to revisit the confusing state noted where the confirm is open but the SubTask can still be typed in, but merging this for now so tabindex isn't totally borked? Maybe moving the confirm modal so it initially appears over the Subtask would be helpful, not sure though.

srallen:
> If so, I can close this and maybe work on a fix in Grommet itself.

I agree that we shouldn't have broken accessibility, but I am confused because @eatyourgreens offered to close this and work on the Grommet fix if it should remain a modal.

eatyourgreens:
I don't have a fix for this in Grommet, at least not without regressing https://github.com/grommet/grommet/issues/4721. I've been reading up on how focus traps are implemented in Grommet, particularly the description of https://github.com/grommet/grommet/pull/3648, and I'm not sure there is a generic solution for this. Rather, I'd see Grommet as providing utilities to handle focus management but it's up to us, as designers and developers, to implement the specific focus management requirements of our components. It certainly seems easier to handle focus ourselves, in the subtask popup, than figure out generic versions of Grommet's [`makeNodeFocusable`](https://github.com/grommet/grommet/blob/6fbf1668da4c412f43819fc6fcb89492d280f1d4/src/js/utils/DOM.js#L104-L123) and [`makeNodeUnfocusable`](https://github.com/grommet/grommet/blob/6fbf1668da4c412f43819fc6fcb89492d280f1d4/src/js/utils/DOM.js#L127-L156) that handle all possible combinations of multiple focus traps open on the screen.

I've been working on disabling tasks, while waiting on a user decision, in #2198 so I've started applying those same changes here. Essentially disabling the subtask and requiring the user to confirm their decision to cancel the annotation.

I also saw that Grommet added `trapFocus`, to control focus traps in dropdowns, in response to a request from us.
https://github.com/grommet/grommet/pull/4378
I'm going to play around with that. It seems like it might fix the focus problems with text task suggestions and also with the collections search dropdown.

eatyourgreens:
I'm going to look into why the new tests won't run, but focus works for me now on a local copy of this. To test it fully, you'll need to build the classifier and  run it inside a local copy of the project app.

mcbouslog (approved):
This LGTM! Thanks for updating and great to see focus persist as expected. Tested on projects and components noted without issue, good UI.

PR #2181, opened 2021-05-14
Observe TranscriptionLineMark

srallen:
_Please request review from `@zooniverse/frontend` team or an individual member of that team._ 

Package: lib-classifier

Closes #2180 

Describe your changes:
I extracted out the transcription line mark into a separate component in #2159, so I could conditionally render a tooltip, but forgot to wrap it in a `mobx-react` observer. This adds the observer to it. 

If anyone has ideas on how to test to make sure a component is decorated by a particular decorator, I'm open to suggestion.

Manually testable with: https://localhost:8080/?project=11300&env=production&demo=true

And the transcription line story, to check that it works outside of the transcription task context. 

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2187, opened 2021-05-19
Classifier: Observe drawing marks

eatyourgreens:
Package:
lib-classifier

Continues #2181.

`DrawingToolMarks` is dynamically wrapping mark components in `observer()` on every render. This removes that code and explicitly wraps each drawn mark in an `observer`.

The `observer` is needed so that marks re-render when we drag them or change their size via drag handles, but it only needs to be added once. At the moment, `DrawingToolMarks` is creating a new observed mark component on every render.

Try it out on the drawing step for I Fancy Cats, or this test drawing workflow: https://localhost:8080/?project=908&workflow=3370

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


shaunanoordin:
Code read looks good, running tests now

shaunanoordin:
### Ongoing Tests

package: `lib-classifier`

Issue 1: **can't go back to a previous step and delete mark.**

Replication:
- Open a workflow with Task 1: drawing with a Point tool, Task 2: text task. (e.g. https://localhost:8080/?env=staging&project=1873&workflow=3493)
- In Task 1, draw a single Point, Annotation X.
- Proceed to Task 2.
- Go Back to Task 1.
- Delete Annotation X
- Proceed to Task 2 again.
  - Expected: Task 2 should load
  - Actual: page crashes to white.
  - Error message:
    - _Uncaught Error: [mobx-state-tree] Failed to resolve reference 'ckoywfhfg00043h6792lv6jiz' to type '(Ellipse | Line | Mark | Point | Rectangle | TemporalPoint | TranscriptionLine)' (from node: /classifications/resources/ckoywfc8j00003h675ki5guw6/annotations/ckoywfc8n00013h67fxusda3f/value/0)_ ; _The above error occurred in the <PreviousMarks> component:_

❗ That said, I don't think this issue is caused by this branch, but is extant on `master` as of a09284a37e06666ac1285fa12179783fc2442c88 . I'm making a note of this in case I need to spin up a new issue. (TODO: check if this has been documented already.)

shaunanoordin:
### Ongoing Tests

Issue 2: **if I go back to a drawing step, new annotations wont appear in future PreviousMarks**
- Related to issue 1.
- ❗ Applicable on `master` too, so this issue isn't exclusive to this branch.

Replication:
- Same WF as Issue 1
- In Task 1, draw Annotation X
- Proceed to Task 2, then return to Task 1
- draw Annotation Y
- Proceed to Task 2.
  - Expected: the PreviousMarks layer should show both Annotations X and Y
  - Actual: Only Annotation X is shown.

shaunanoordin (approved):
## PR Overview

This PR is, if I understand correctly, primarily a _performance improvement._

- The "observe for changes" functionality is now the responsibility of _individual mark types_ (Rectangle, Line, Ellipse, Point)
  - These individual marks will no longer be re-rendered upon every tiny change to their values? I haven't been able to find a way to properly test this, however.
- The Delete button is no longer observed. (Er, did it need to be observed originally?)

Code read looks good and functionality looks good. 👍  (Despite notes from Ongoing Tests above!)

### Testing

Tested on localhost on a staging project ([Task 1: drawing with multiple tools + subtasks + min/max reqs, Task 2: question, Task 3: drawing](https://localhost:8080/?env=staging&project=1873&workflow=3493)) using macOS 10 + Chrome 90

- On the drawing tasks...
  - I can create annotations
  - I can delete annotations
  - I can move annotations
  - I can work with drawing annotations that have a simple subtask
- I can move from Task 1 to Task 2 to Task 3, no problemo.
  - Drawing annotations from previous tasks will be visible - but not interactive - on the PreviousMarks layer
- I can submit my classification
  - Annotations on the classifications are confirmed to correctly reflect the drawing marks I made on the subject.

I can't _go back_ to modify annotations from a previous Drawing Task, but that's been noted in Issue 1 and 2 in the Ongoing Tasks. These issues are likely unrelated to this branch, so ignore them for this PR - I'll spin up new issues for those if necessary.

### Status

LGTM! 👍 

Bonus notes for @ErikOstlund :
- If components.TemporalPoint extends components.Point, does _TemporalPoint also need to be observed?_ Will Temporal Point inherit the observable-ness of Point? Or does Temporal Point need to be observed separately? If so, does Temporal Point need to inherit from an _pure, unobserved_ Point?
  - _(Sub-note: if you need an un-observed components.Point to extend components.TemporalPoint, just add an `export Point` next to the `export default observer(Point)` and change your import to `import { Point } from ...`)_
- Ditto for RotateRectangle

eatyourgreens:
Thanks for the thorough review @shaunanoordin. Davy Notebooks are testing a transcription task workflow with three steps. I think the bugs you found might affect that workflow.

eatyourgreens:
Confirmed on Davy Notebooks that the store crashes if you go back and delete a transcription line. Deleting a mark leaves a reference to the deleted mark in the task annotation, even though the annotation is using `safeReference`. 
https://www.zooniverse.org/projects/humphrydavy/davy-notebooks-project/classify/workflow/18244

The Davy Notebooks project on my phone, showing a store reference error in place of the classifier.

eatyourgreens:
I think that bug might be that `shownMarks` isn’t being observed here but I’m not sure. https://github.com/zooniverse/front-end-monorepo/blob/a09284a37e06666ac1285fa12179783fc2442c88/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/PreviousMarks/PreviousMarksConnector.js#L19-L35

eatyourgreens:
I've opened https://github.com/zooniverse/front-end-monorepo/pull/2199 to investigate that bug. At the moment, I'm adding tests to check that the annotation updates as expected when you create or delete a mark.

eatyourgreens:
I think I see the bug: The Next button uses an undo/redo pattern. `step.complete(annotations)` runs the first time you press the Next button, which saves your initial marks. After that, going back and forward again uses undo/redo, which redoes your initial drawing and ignores changes. The fix wouldbe to always call `step.complete(annotations)` on Next.

eatyourgreens:
I've confirmed that when I refactored the Next button (#2062) I introduced this bug by running `step.completeTasks(annotations)` inside the undo manager. It should be run outside the `next()` action.

PR #2191, opened 2021-05-19
Project app storybook and webpack 5

srallen:

Package: app-project

The storybook for the project app fails to build. The failures show several core node modules failing to load which is indicative of a breaking change with webpack 5 not polyfilling these anymore. From my research this afternoon, I found that if you define any customizations in a webpack config for storybook that conflicts can arise since storybook is still on webpack 4. 

This branch installs a beta support for webpack 5 as well as updates to the config to either skip certain module polyfills or to load them up as needed. The majority of those that were failing seem like they can be skipped, but process is needed but I can't seem to get it to work from the recommended way from the webpack 5 docs. I don't know yet if this is because the webpack 5 support for storybook is in beta or if I'm doing something wrong. Regardless, this now is building and loading most of the stories.

I also removed a customization for the storysource add-on because it uses a webpack 4 style loader. Webpack 5 significantly changed how this works: https://webpack.js.org/guides/asset-modules/

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/app-project/webpack.config.js
```diff
@@ -1,13 +1,31 @@
...
+    fallback: {
+      console: false,
+      domain: false,
+      fs: false,
+      http: false,
+      https: false,
+      net: false,
+      os: false,
+      process: 'process/browser',
+      tls: false
```
srallen, line 28: I haven't tested this with the actual app. The next.js webpack config loads up the aliasing portion of this, so I don't think this will conflict, but we need to confirm.

eatyourgreens:
We're using webpack 5 to build the React components and classifier libraries. Those storybooks are still being built with webpack 4. Should we upgrade all the storybooks together?

packages/app-project/webpack.config.js
eatyourgreens: I think the only package that uses these might be `@sentry/node`, and that shouldn't be included in webpack builds anyway.

The NextJS build doesn't use these, as far as I can tell, and runs fine in webpack 5 (unless Next is adding fallbacks itself in the core webpack config.)

eatyourgreens:
I think I'd also hold off on upgrading the storybook builds until webpack 5 support is out of beta. Beta should be ok, as far as I can tell, but we don't know if there will be further changes before it's fully supported.

packages/app-project/webpack.config.js
eatyourgreens: Having said that, I think it's going to be confusing to have a webpack config file that's only partly loaded into `next.config.js`. I'd assume that the entire config here is being applied to the NextJS build. If fallbacks are only needed for the storybook build, then I'd add them to the custom storybook config.

If they don't make any difference at all to either build, then I'd leave them out.

srallen:
Ok, sure. If there are any more issues because of webpack 4 and 5 running together, then you can make the update later.

PR #2193, opened 2021-05-20
Engaging Crowds: Subject Picker: add Already Seen status

eatyourgreens:
Convert `checkRetirementStatus` to use the `subjects/selection` endpoint (with the current user's auth token), which gives us the already seen and retired status of each subject. Subjects have three possible statuses now: 'Unclassified', 'Already seen' and 'Retired'.

I've also localised the subject statuses so that they can be translated.

You can try it out on the Index or Minutes workflows for Scarlets & Blues, although those subjects are all unclassified at the moment.
http://localhost.zooniverse.org:3000/projects/bogden/scarlets-and-blues?env=production

Package:
app-project

Closes #1955.

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2198, opened 2021-05-21
Engaging Crowds: block the task area for finished subjects

### LINKED ISSUE #2027: Engaging Crowds: block the task area for already seen or retired subjects
## Package
lib-classifier

Engaging Crowds should allow volunteers to see previously classified, or retired, subjects but not classify them.

Returning to an already-classified subject would display a modal, in the task area, explaining why they can still look at this subject; same with a retired subject. 
- [Already seen](https://projects.invisionapp.com/d/main#/console/12924056/438011816/preview)
- [Retired](https://projects.invisionapp.com/d/main#/console/12924056/438011817/preview)

The modal has two options: to return to the project home page or to choose another subject from the subject set. Choosing another subject opens the modal to choose another subject within the classify page: https://projects.invisionapp.com/d/main#/console/12924056/440898093/preview  

Should be able to reuse the modal from the project home page on the classify pages, but should not include the 'back to workflow selection' button and has different text ('choose a subject' rather than 'choose a subject to get started').

Screenshot of a modal blocking the task area for a subject which has already been retired.

eatyourgreens:
When a subject is retired, or has already been seen by the logged-in volunteer, disable the current task and show a popup giving options to move on to a new subject. Enable the task area when the popup is closed/cancelled.

The subject remains completely interactive while the popup is open, so that we can zoom in and read it before deciding whether or not to skip it.

Popup options are:
- Choose another subject: reset the classifier subject store and pick a new subject from the current subject set.
- Next available subject: reset the subject store and seed a new queue with an unclassified subject from `/api/subjects/queued`.
- Dismiss: close the popup and continue with classifying the current subject.

Test URL: http://local.zooniverse.org:3000/projects/bogden/scarlets-and-blues/classify/workflow/17077/subject-set/92752/subject/58047422?env=production&demo=true&finished=true

(Use the `finished` query parameter to force subject finished status for testing.)

Screenshot of a Scarlets & Blues page with the task area disabled and blocked by a popup giving options to select another subject or dismiss the popup.

Package:
app-project
lib-classifier
lib-react-components

Closes #2027.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-react-components/src/helpers/withLayer/withLayer.js
```diff
@@ -41,7 +41,10 @@ function withLayer (WrappedComponent) {
     closeFn: PropTypes.func,
     modal: PropTypes.bool,
     plain: PropTypes.bool,
-    position: PropTypes.string
+    position: PropTypes.string,
+    target: PropTypes.shape({
+      current: PropTypes.node
```
eatyourgreens, line 46: The Grommet Layer docs are wrong. This is a DOM node, and is expected to have a `getBoundingClientRect` method (in the Layer code.)

eatyourgreens:
This is in a working state now, so I'm going to remove the draft status.

packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskArea.js
```diff
@@ -1,100 +1,103 @@
...
+  const [ disabled, setDisabled ] = useState(false)
+  const taskArea = useRef(null)
 
-class TaskArea extends React.Component {
-  constructor () {
-    super()
+  useEffect(function onSubjectChange() {
+    // TODO: remove this once testing is complete.
+    const URLParams = queryString.parse(window?.location?.search)
+    const finished = (subject && URLParams?.finished) || subject?.retired || subject?.already_seen
```
eatyourgreens, line 32: I think this will need an additional filter condition, otherwise the task area is going to be disabled for finished subjects in all projects.

shaunanoordin:
Zoinks, this is going to need a manual rebase/conflict resolve (I can't even click the "Resolve conflicts" button this time) but I should be able to at least do a code read and basic functionality test in the meantime.

ErikOstlund:
Can someone clarify for me:
Why do we allow users to classify a subject if the subject has already been seen or the subject is retired?
Instead of disabling the Task area, can we hide it and only show the new modal? If so, we can remove this option: Dismiss: close the popup and continue with classifying the current subject.

shaunanoordin:
@ErikOstlund 

Q: Why do we allow users to classify a subject if the subject has already been seen or the subject is retired?

A: Turns out, a number of our users _specifically ask_ for the ability to Classify on subjects, even if the subject's fully classified! While individual motivations for such a request would differ, if I'm to guess, I'd say that most of our volunteers are just interested in the science itself. (i.e. they want to continue exploring Subjects even if "all the work is done".)

Personal opinion: Additional classifications for a retired subject doesn't do any harm anyhoo, and if providing the users a simple choice like this keeps them motivated in the science and invested in the platform as a whole, I'm all for it 👍

srallen:
Finished workflows and projects get used for educational classroom purposes, demonstration purposes (think conference talks), and citation purposes (from papers).

shaunanoordin:
## Quick update

- ✅ Code read looks good
- ✅ Basic project looks fine: [without](http://localhost:3000/projects/brooke/i-fancy-cats/classify/workflow/693?env=staging&demo=true) and [with](http://localhost:3000/projects/brooke/i-fancy-cats/classify/workflow/693?env=staging&demo=true&finished=true) the "Finish" popup, I can still classify cats with no issues.
- ⚠️  Issue discovered: If Task Panel is too small/short, the "Finished" popup won't display properly.
  - The "Finished" popup will try to fit into the teeny tiny space available, getting cropped off in the process.
  - Additionally, the popup's overflow scroll doesn't work (touchpad scrolling, mousewheel scrolling, and simply trying to drag the component's scrollbar), preventing access to the action buttons.
    Screen Shot 2021-07-07 at 01 06 01
  - local URL: http://localhost:3000/projects/darkeshard/survos-testing-2020-subject-group-viewer/classify/workflow/3412?env=staging&finished=true
- ❔ The "Finished" popup mostly works, I think?
  Screen Shot 2021-07-07 at 01 16 15
  - When the "Finished" pop appears, I can't interact with the tasks behind it, ~~and I can't add annotations on the Subject~~
  - UPDATE: wait, I can actually still selecting drawing tools and draw on the Subject Image, although (thankfully) I can't submit/go to the next step.
  - ⚠️ Issue discovered?: if the first task is a Drawing Task, that task is automatically active, and I can draw annotation marks on the Subject Viewer/Interaction Layer.
    Screen Shot 2021-07-07 at 01 28 30
    - Also, I can (for some reason??) STILL click on Drawing Tools behind the "Finished" Popup.
    - I consider this a NON-BLOCKING issue though. (More details in full review.)
  - ✅ "Dismiss" button works as advertised.
  - ✅❔ "Next Available Subject" works.
    - [Personal note][TODO: check] specifically, it RELOADS the /queued endpoint with a queue size of 1, and then RELOADS AGAIN with the full size (default 10). Need to re-read the code for this.
  - ❓ "Choose A New Subject" button... not sure what this is supposed to do, I need to re-read the code. [TODO: check]

I'm still throwing a bunch of different scenarios at the code to test functionality, so brb.

shaunanoordin:
(Mumbling to self) OK need to think this through a bit. Anyone, feel free to comment, but this may just be me trying to talk through the logic.

Q: In what setting would a user receive a finished (either retired, or already_seen) Subject?

Assumption: if the /queued endpoint serves the user a retired or already_seen Subject, there's a good chance the /queued endpoint has no further fresh Subjects to serve the user, right?

Follow up 1: If so, would the "Next Available Subject" (see SubjectStore.nextAvailable()) action, which re-fetches data from /queued, end up with more retired Subjects? Need to test this on a more realistic project. (Note to self: add scenario for an actually mostly-completed WF)

Follow up 2: If a user gets a stack of retired/already-seen Subjects, won't the "Finished" popup be constantly popping up, which has the potential to be annoying? (Compare the general good UI guideline of "if I dismiss a reminder once, it should remain dismissed for the rest of the session")

Follow up 2a: the popup does make sense if the user gets a queue of, e.g. [retired, fresh, retired, fresh, fresh, already-seen] Subjects. Check the assumption.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/DisabledTaskPopup/DisabledTaskPopup.js
```diff
@@ -0,0 +1,84 @@
...
+      title={counterpart('DisabledTaskPopup.title')}
+    >
+      <Paragraph>
+        {counterpart('DisabledTaskPopup.body')}
+      </Paragraph>
+      <PlainButton
+        alignSelf="center"
+        onClick={reset}
+        text={counterpart('DisabledTaskPopup.options.select')}
+      />
```
eatyourgreens, line 59: 'Select a new subject' only makes sense in the context of a workflow that allows subject selection, so this maybe needs to be wrapped in a check for `workflow.hasIndexedSubjects` or similar.

packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskArea.js
eatyourgreens: Maybe add a check for `workflow.hasIndexedSubjects`, so that this is limited to indexed subject sets for the time being.

eatyourgreens:
@shaunanoordin thanks for the thorough review. This feature is limited to Engaging Crowds at the moment, so I've left a couple of notes to myself to add checks that we're on a workflow, and subject set, that uses indexed subjects.

The content of the popup is definitely influenced by the context that the classifier appears in (project, workflow and subject.) #2165 discusses that, to a certain extent.

eatyourgreens:
>Follow up 2: If a user gets a stack of retired/already-seen Subjects, won't the "Finished" popup be constantly popping up, which has the potential to be annoying? (Compare the general good UI guideline of "if I dismiss a reminder once, it should remain dismissed for the rest of the session")

That's a good point. Once #2193 is merged, every subject should list its status (Available/Already Seen/Finished) when you open the popup to pick a subject.

If we add this popup to workflows with queued selection, then we'll need to think about how it popping up constantly could drive volunteers away, similar to how the Finished banners can discourage volunteers at the moment.

shaunanoordin (changes requested):
My actual review is finally done! @eatyourgreens the review below is kinda lengthy, so the TL;DR is:

- CR1: "Choose A New Subject" should only appear on projects/workflows with individual Subject selection (e.g. WF 17077 of Scarlets & Blues)
- CR2: Task Area needs a minimum height.
- CR3: The "Dismiss" action should ensure the "Finished" popup is dismissed for the rest of the session.

Alternatively, if time is an issue, hardcode the "Finished" popup to only appear for certain WFs (e.g. WF 17077) while the above two CRs are addressed in separate PRs. Whatever works! 👍 

And now, the feature presentation...

--------

## PR Review

This PR adds the "Finished" popup.

- The popup should appear only when a retired or already seen Subject is served
- Purpose of the popup is to inform (or rather, reinforced to the) users that the subject is retired/already seen, _and_ to give them a choice of whether to continue classifying a retired subject, or do something else.

The "Finished" popup has 4 points of function.

- When the popup is visible, it should block interaction with the Tasks.
- The "Dismiss" button should remove the popup, and allow users to continue classifying as normal.
- The "Next Available Subject" button basically reloads the Subject queue.
- The "Choose A New Subject" button opens the Subject selection menu, _if available._
  - ⚠️ This should only be visible on WFs that allow Subject selection. (Projects with Workflow/Subject Set selection do NOT count. i.e. this button shouldn't work on HMS NHS, but should work on Engaging Crowds.)

While most of the main functionality + code read looks good, there are some issues that need to be addressed before this is good to merge, since it affects multiple projects.

### Testing

Tested on localhost, using macOS 10.12.16 + Chrome 91 unless otherwise noted. Tested with `app-project`, running `yarn build && yarn start`

**Scenario 1:** basic project (I Fancy Cats), with no "Finished" popup

- URL: http://localhost:3000/projects/brooke/i-fancy-cats/classify/workflow/693?env=staging
- Basic project looks fine. 👍
- I can load the home page, navigate to Classify
- I can view the image (Single Image Viewer), and pan (mouse/keyboard), zoom (mouse wheel/UI buttons), and rotate (UI).
- I can answer single/multiple answer tasks.
- I can create Point and Ellipse annotations (Drawing Task).
  - I can edit existing marks (drag to move Point + Ellipse, drag to resize Ellipse), and delete existing marks.
- I can submit the Classification, and the annotations are correct.

**Scenario 2:** main functionality test - the "Finished" popup

- URL: http://localhost:3000/projects/brooke/i-fancy-cats/classify/workflow/693?env=staging&demo=true&finished=true
- The "Finished" popup's meets 3 of 4 points of function:
  - When the popup is visible, it block interaction with the Tasks, but this is partially wonky for Drawing Tasks. (see Scenario 2b) 👌 
  - The "Dismiss" button works as advertised. 👍 
  - The "Next Available Subject" button seems to work 👍 (I need to have a custom /queued endpoint to fully test that the next Subject is always not retired, but what I'm getting is good enough)
  - ❗ The "Choose A New Subject" button does NOT function, and CAN'T function because this Workflow does not allow individual Subject selection.

**Scenario 2a:** Subject Group Viewer with "Finished" popup
- URL: http://localhost:3000/projects/darkeshard/survos-testing-2020-subject-group-viewer/classify/workflow/3412?env=staging&finished=true
- WF has a very, very short Task in the Tasks Panel.
- ❗ Issue discovered: if Tasks Panel is too small/short, the "Finished" popup won't display properly.
  - The "Finished" popup will try to fit into the teeny tiny space available, getting cropped off in the process.
  - Additionally, the popup's overflow scroll doesn't work (touchpad scrolling, mousewheel scrolling, and simply trying to drag the component's scrollbar), preventing access to the action buttons.
  - While I initially dismissed this as a low priority bug since it doesn't affect any active FEM projects, I realised that this is getting launched soon: [Galaxy Zoo: Weird and Wonderful](http://localhost:3000/projects/zookeeper/galaxy-zoo-weird-and-wonderful/classify/workflow/18510?env=production&demo=true&finished=true)

**Scenario 2b:** Drawing Tasks
- URL: http://localhost:3000/projects/darkeshard/prototype-2015/classify/workflow/18372?env=production&finished=true
- WF is a single step Drawing Task with 3 tools.
- ⚠️ Issue discovered: if the first task is a Drawing Task, that task is automatically active, and I can draw annotation marks on the Subject Viewer/Interaction Layer.
  - Also, I can (for some reason??) STILL click on Drawing Tools behind the "Finished" Popup.
  - I consider this a NON-BLOCKING issue, however, since it's just a "decorative" UI issue. (Users can draw unexpectedly, but they can't submit, and it's SUPER obvious what they need to do if they want to submit annotations.) (I'm sure a more stringent UI designer will have a fit over this though!)

**Scenario 2c:** TESS + Light Curve Viewer
- URL: http://localhost:3000/projects/nora-dot-eisner/planet-hunters-tess/classify/workflow/11235?env=production&demo=true&finished=true
- I can view the Subject on the Light Curve Viewer, I can make annotations, I can submit. All good.
- "Finished" popup mostly works. (note 1)
- I'm noting that the Tutorial popup still appears on top of everything else, cool.

**Scenario 2d:** Zwicky's Stellar Sleuths + Variable Star Viewer
- URL: http://localhost:3000/projects/adamamiller/zwickys-stellar-sleuths/classify/workflow/16135?env=production&demo=true&finished=true
- I can view the Subject on the variable star viewer, I can answer the standard single answer question task, I can submit. All good.
- "Finished" popup mostly works. (note 1)

**Scenario 2e:** HMS NHS + Workflow/Subject Set Selection
- URL: http://localhost:3000/projects/msalmon/hms-nhs-the-nautical-health-service/classify?env=production&demo=true&finished=true
- This project allows users to select the Workflow and Subject Set.
- Since HMS NHS has a LOT of finished (retired) Subjects, I can confirm that the "Finished" popup appears correctly for retired Subjects without having to force it with &finished=true. Yay!
- Workflow selection and Subject Set selection works as expected.
- "Finished" popup mostly works. (note 1)

**Scenario 2f:** Davy Notebooks + Transcription Tasks
- URL: http://localhost:3000/projects/humphrydavy/davy-notebooks-project/classify?env=production&demo=true&finished=true
- Transcription tasks work as expected.
- As noted in scenario 2b, I can still draw while/make text transcriptions while the "Finished" popup is visible. Otherwise...
- "Finished" popup mostly works. (note 1)

**Scenario 2g:** Scarlets & Blues + SUBJECT SELECTION 👍 
- URL: http://localhost:3000/projects/bogden/scarlets-and-blues/classify/workflow/17077/subject-set/92753/subject/58047430?env=production&demo=true&finished=true
- Finally, here's an example of a project with a workflow (17077) that allows Subject selection.
- In this specific scenario, the "Finished" popup FULLY works. 👍 
  - The "Choose A New Subject" action correctly opens the Subject selection menu.
  - All other actions work as advertised.

(note 1) "mostly works" means 3 of the 4 functionality points are met. The "Choose New Subject" button, however, continues to be visible despite not doing anything. This should be hidden for this project/workflow.

### Dev Notes

- ❗ There's a possible _UX annoyance_ with _constant popups._
  - If a user gets 1 finished (retired/already-seen) Subject, they're likely going to get a lot more retired/already-seen Subjects from the /queued endpoint.
  - As a result, the "Finished" popup will likely appear for every Subject henceforth, which may be very annoying for users who must Dismiss multiple popups.
  - Suggestion: consider the better UI behaviour of "clicking Dismiss once indicates I want to ignore the popup for the rest of the session"

### Status

As mentioned, this PR is mostly good, and works perfectly for its target scenario (2g, projects with Subject Selection.) I do have some Change Requests, to make these changes OK for every _other_ project on FEM. (scenarios 2a to 2f)

Changes requested:

- [x] "Choose A New Subject" should only appear on projects/workflows with individual Subject selection (e.g. WF17077 of Scarlets & Blues)
- [ ] The Task Area needs a minimum height to function correctly with the "Finished" popup.
  - I think adding an arbitrary `min-height: 10em` or something to the Task Area should suffice.
  - Side note: it's worth figuring out why the scroll doesn't work when the popup is in overflow mode.
- [ ] The "Dismiss" action should ensure the "Finished" popup is dismissed for the rest of the session.
  - I trust your judgement to define what a "session" is.
  - Idea: add "finished_popup_dismissed = true/false" to the SubjectViewerStore or WorkflowStore? It doesn't have to be complicated, i.e. once set to true, it stays that way until the user _reloads the Classify page._ No need to cleanup/reset on new Workflow or new Subject Set or etc.

ALTERNATIVELY: if time is an issue, _temporarily hardcode_ the "Finished" popup to only appear for specific projects/workflows. With this patch, the CRs above can be addressed at a later date.

```
// In TaskArea.js
const finished = project_id = 1234 && ((subject && URLParams?.finished) || subject?.retired || subject?.already_seen)
```

eatyourgreens:
Thanks for the thorough review, @shaunanoordin. I'm paging @snblickhan for comments on how the popup works: particularly whether dismissing it once should dismiss it for all subsequent finished pages.

I've noticed a couple of accessibility bugs while testing.

First: the new popup steals focus from the tutorial. Have a look at where the focus ring is in this screenshot. The tutorial is open, but the active element is the 'Select another subject' button. If possible, we should avoid having two popups open at the same time.
Screenshot showing the modal tutorial open for Scarlets & Blues, but focus is on the 'Select another subject' button, which is beaneath the modal overlay.

Second (which also applies to the Field Guide): modals are added to the DOM at the end of the page, after the footer. Reading through the page in VoiceOver, when I get to the task area I don't get the options to choose another subject. Instead, I just hear that the task area inputs are disabled, without explanation as to how to proceed.

That second bug isn't specific to the new classifier. We've also had comments that the Field Guide is hidden below the bottom of the page for blind volunteers using PFE:
>My biggest criticisms so far are many sounds are extremely short, (privacy concerns limit all sounds outside the field guides to 500MS or less, but many are much shorter), and the field guides are hidden below what are normally bottom of page overall zooniverse.org generalized stuff like the Privacy Policy.

https://www.zooniverse.org/talk/18/1002723?comment=2012101&page=2

eatyourgreens:
Oh, looks like I haven't added tests for the task area either.

eatyourgreens:
@shaunanoordin I've updated the task area so that it checks `workflow.hasIndexedSubjects` before disabling tasks. That should limit the feature to workflows that support indexing.

snblickhan:
Hey @eatyourgreens and @shaunanoordin 

Good point re: the Finished pop ups becoming annoying, but this is the reason we built in the "Take me to the next available subject" option in that pop up. Because of that option already being available, I'm not particularly inclined to make these pop ups stop appearing. I think I'm more comfortable having people be annoyed with a pop up and potentially be "over-informed" that a set is complete, than not realize it's done and keep working on it without realizing it's done (which is a problem we know we currently have, even with banners, etc.).

@mrniaboc any thoughts to add here?

srallen:
We'll want the modal to be dismissable for a session if the project is used for educational reasons, for example like how Wildcam Gorongosa ran for quite some time, where we may want to allow subjects that are finished to be classified. 

A dismissal function can be local component state; I don't see the need to have it be global state in the MST stores.

eatyourgreens:
At the moment, I've got this wrapped in `if workflow.hasIndexedSubjects` so it only appears for Engaging Crowds. I can see it being useful in other contexts, but I could also see the logic becoming quite complicated if the popup's available in other contexts eg. showing different content in the modal if it's shown in a classroom context.

I also need to figure out how to test `TaskArea` with an indexed workflow, which means figuring out how to write tests for `useEffect`.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/DisabledTaskPopup/DisabledTaskPopup.js
eatyourgreens: The task area now checks for `workflow.hasIndexedSubjects` before showing this modal. The content is only relevant to Engaging Crowds, at the moment.

shaunanoordin (approved):
## PR Review (Update)

This PR is looking good! The major change, as Jim outlined, is that the "Finished" popup _can only_ appear on Workflows with indexed Subjects. At time of writing, only [Scarlets & Blues, WF 17077](http://localhost:3000/projects/bogden/scarlets-and-blues/classify/workflow/17077?env=production&demo=true) has such feature.

- Since the majority of other FEM project no longer need to worry about the "Finished" popup, all my CRs from the earlier review are either resolved (by definition, the "Choose a New Subject" button will only appear when the popup appears for WFs with Subject selection) and bumped to low priority. 👍 
- Relevant code is in `TaskArea.js`: `setDisabled(finished && workflow.hasIndexedSubjects)`
  - (bonus points for the safeties that ensure workflow will never be undefined.)

I'm happy with how things are now, so I'm giving this a LGTM! 👍 

### Testing

I'm repeating earlier tests.

**Scenario 1** works fine, base project looks good.

**Scenarios 2, Scenarios 2a to 2f** look good, albeit modified.
- The new expectation is that the "Finished" popup _should not appear_ even with _&finished=true_ in the parameters. And it doesn't appear! 👍 
- All other functionality, unique to each scenario, continues to work fine. 👌 

**Scenario 2g** looks good, with a minor note of mild exasperation.
- The new test URL is: http://localhost:3000/projects/bogden/scarlets-and-blues/classify/workflow/17077/subject-set/96236/subject/63413753?env=production&demo=true
- Alternatively, go to http://localhost:3000/projects/bogden/scarlets-and-blues/classify/workflow/17077?env=production&demo=true&finished=true and then _choose the correct Subject Set,_ and then Subject. The "Finished" popup appears as expected, and all 4 points of functionality are met.

Now, the _minor note of mild exasperation_ comes from the fact that it seems someone changed the active Subject Sets while we were testing. Whaaat?

_Of the four indexed Subject Sets, only one Subject Set is now active for WF 17077_

Screen Shot 2021-07-09 at 00 20 11

_IIRC, the Subject Set Selection component relies on our subject indexing API, so it wouldn't know of changes to WF17077, which made things very confusing for a brief moment when I was testing._

Screen Shot 2021-07-09 at 00 20 46

### Status

LGTM 👍 

The previous remaining CRs (min height for Task Area, and "dismiss button should dismiss for whole session") are still on the "suggested TODO" list, but I'm happy to see those tackled in a separate PR.

As far as I'm concerned, this PR works as advertised, doesn't interfere with other existing FEM projects, and is chonky enough as it is.

shaunanoordin:
EDIT: I just saw the 49-comment thread on Engaging Crowds regarding why certain Subject Sets were disabled. 😱 Erm, never mind then!

eatyourgreens:
Let’s get this merged and I’ll figure out how to test `useEffect` in a separate PR. 

I’ll also write up issues for the accessibility of this modal and the field guide.

eatyourgreens:
Also noting that Grommet’s sizing for targeted layers seems to fill the vertical height of the target area, even when that height is shorter than the content of the layer. We may have to return to that, since it makes it impossible to get past the popup.

eatyourgreens:
No, wait: the height of the task is supposed to match the subject. The sizing bug is that the task area is too short on Galaxy Zoo: Weird and Wonderful.

PR #2223, opened 2021-06-02
Fix Subject Group Viewer (aka Grid Viewer) zoom on Safari

shaunanoordin:
## PR Overview

This PR fixes an issue where the zoom functionality for the Subject Group Viewer, on Safari, causes the image(s) in the cells to zoom from the _top left of the image,_ instead of from the _centre of the image._

- Issue is specific to Safari
- Turns out that Safari doesn't support `transform-origin`, which allows us to alter the "anchor" for the transformation process. Without this parameter, the origin point will always be the default (0,0) coordinate at the top left of the image.
- Solution is relatively crude but functional: _translate twice_ to offset the and then un-offset the size of the image before/after the scaling transform.
- This could be compressed into a single Matrix transform, actually, but I think the _translate-scale-translate-translateAgain_ transform is more readable.

### Testing

Test URL:
- https://local.zooniverse.org:8080/?env=staging&project=1900&workflow=3412
- This is a sample project that uses the Subject Group Viewer with a 5x5 grid

Testing actions:
- Select your browser. The one we're testing for is **Safari 14,** though I've confirmed this works back to Safari 12.
- Open the test URL
- Click on the Zoom In button on the image toolbar.
  - Expected: The image in each grid cell should zoom in, _centred around the centre of the image._

Notes:
- Compare your results with zooming in using Chrome.
- The various methods of zooming - ImageToolbar button, mouse scroll, and +/- keyboard keys - are equivalent.
- Please note that the smallest zoom is 1x, so you can't "zoom out" to see a tiny image in the grid cell.

### Status

Ready for review!


shaunanoordin:
Goshdangit, I need to fix the tests - what happened there?

shaunanoordin:
## PR Update

Tests fixed! Do your thing, CI Tests.

eatyourgreens (approved):

camallen (approved):
> // This hack is necessary since Safari doesn't support transform-origin to
   // manually define the origin point, i.e. we can't use
   // transform-origin={`${imageWidth/2}px ${imageHeight/2}px`}

Reading the MDN docs it looks like you might be able to use `transform-origin: center;` as that seems supported in safari https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin

However i haven't looked into this so trusting your approach is fine. It seems to be the long hand version of the use of transform-origin as outlined in the docs

PR #2229, opened 2021-06-03
Add ADR-32: Subject Viewer Layouts (aka subject viewer sizing/fitting behaviour)

shaunanoordin:
## PR Overview

This PR adds a new ADR for Subject Viewer Layouts.

- The topic of _"how do we fit/size the Subject Viewer in a way that makes sense to the available screen real estate?"_ was brought up during today's frontend dev standup.
- Pertinent information is in the ADR itself.
- See also: the relevant [Slack conversation](https://zooniverse.slack.com/archives/C06LHUC6R/p1622734773089200) following the standup.

### Status

Ready for review!


docs/arch/adr-32.md
```diff
@@ -0,0 +1,33 @@
...
+
+3 June 2021
+
+## Context
+
+Currently, the FEM Classifier's default "Subject Viewer sizing/fitting" behaviour is _"fit to width, no max height"._
+
+This is analogous to PFE's special "no-max-height" behaviour, which had to be _explicitly set._
+
+While the default "fit to width, no ma height" behaviour is good in many cases, there can be issues for certain workflows, e.g. when the Subject is a tall image (such as a portrait photo) and the user's window is very wide, causing the bottom half of the image to be "cut off".
```
srallen, line 11:
```suggestion
While the default "fit to width, no max height" behaviour is good in many cases, there can be issues for certain workflows, e.g. when the Subject is a tall image (such as a portrait photo) and the user's window is very wide, causing the bottom half of the image to be "cut off".
```

docs/arch/adr-32.md
```diff
@@ -0,0 +1,33 @@
...
+- Whether or not a specific subject viewer size configuration is supported will depend on the Layout.
+
+See also:
+- The current [Layout code](../../packages/lib-classifier/src/components/Classifier/components/Layout), as of June 2021, currently only has DefaultLayout.
+
+## Consequences
+
+- As of June 2021, the designs & plans for creating Layouts exist, but we don't have dev resources to focus on such an enhancement.
+- While we wait to get dev time on that, each "special" Subject Viewer (e.g. SubjectGroup, DataViz) should handle Subject Viewer sizing/fitting in whatever manner that suits them, in their own component code.
+
```
srallen, line 30: I'd add a note here that depending on the subject viewer, we have a config object now where specific width and height could be set for more refinement beyond a layout. The ScatterPlotViewer has support for setting even margin and padding by way of subject JSON object (and later by config object) for example. Add link to the ADR about subject viewer config?

srallen (changes requested):
Looking good so far. I have a couple of minor suggestions.

docs/arch/adr-32.md
shaunanoordin: Notes added! 👍

shaunanoordin:
## PR Update

I've added an additional note informing readers that certain Subject Viewers can be further customised via workflow configs (Subject Viewer Config) or in some cases, the Subject data itself. 👌 

This PR went stale, sorry about that! Glad I finally remembered to come back to this.

srallen (approved):
Looks great! Thanks!

PR #2231, opened 2021-06-04
Subject Group Viewer: set optional max grid width/height

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Related PR: [Add ADR-32: Subject Viewer Layouts (aka subject viewer sizing/fitting behaviour)](https://github.com/zooniverse/front-end-monorepo/pull/2229)

This PR allows project owners to set an optional maximum grid width/height in the config.

- Context (the problem): due to the Subject Viewer's default behaviour of "fit to width", there are scenarios where tall or square grids will have their bottom half "cut off" in the visible browser space on _wide_ browser screens.
- Implementation (the solution): using the workflow config, allow project owners to limit the visible dimensions of their grids
  - since each project owner will have a different grid setup, it makes sense that they should be able to also decide how the grid should "best" look in terms of "fitting" into the available visible space.
  - This obviously won't work for EVERY project for EVERY browser (that would require dynamic calculations) but it's a good enough solution for most users.
- To enable the optional max grid width/height, the workflow configuration needs some additional fields:
  ```
  workflow.configuration = {
    subject_viewer_config: {
      grid_max_width: '1000px',
      grid_max_height: '50vh',
      ...
    },
    ...
  }
  ```
  - This example shows a limit for both width (1000 pixels) and height (50% viewport height). I mean, obviously, in practice you just limit one of the dimensions, and you pick either pixels or viewport height, you don't mix and match like a madman. 
  - If no `grid_max_width` or `grid_max_height` is set, the default "fit grid to available width" behaviour triggers.

### Testing

URL: https://local.zooniverse.org:8080/?env=staging&project=1900&workflow=3412 (Survos Testing, staging)

Testing steps:
- Apply (via workflow config) the appropriate max-width/max-height you want to test for.
- Open the testing project URL.
- Resize your browser window.
  - Expect that the grid should fit the available width BUT will be limited to the maximum max-width/max-height that you've set, whichever is lower.
- With various browser window sizes, attempt to zoom in, zoom out, pan, and make annotations.
  - Expect that all these actions should work as expected 
- Remove the max-width/max-height and test again.
  - Confirm that by default, the "fit grid to available width" behaviour triggers.

⚠️ As of Friday evening, I haven't edited the testing project to have the correct workflow configuration yet. (I'mma haveta do some PFE editings, yerp.) If you want to test this PR on its own, I'm afraid you need to open up `SubjectGroupViewerContainer` and change `const DEFAULT_GRID_MAX_WIDTH = ''` to `const DEFAULT_GRID_MAX_WIDTH = '200px'` or something.

⚠️ Please note that the max-width/max-height logic is implemented via CSS, so... keep an eye out for CSS shenanigans across browsers. (e.g. some browsers might now support `vw` and `vh`. I know Safari 12 has different logic for fitting SVGs via height, but I think I've for solved that at least.)

### Status

Ready for review? You can definitely review the code, but to make functional testing easier I _really_ should add the workflow config to the test project.


shaunanoordin:
Whoops, fixed the tests.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewer.js
```diff
@@ -21,6 +27,29 @@ const Container = styled.div`
...
 `
 
+/*
+Note on Subject Viewer sizing/fitting:
+- The grid should fit the height OR width of the available visible space.
+- This is implemented with the container div AND the <svg> having
+  width/height=100% (which 'fits') an optional max-width/height (which
+  'restricts').
+- Curious note: the max-width/height has to be repeated in <svg> due to
+  Safari 12. If we ignore Safari, we only need max-width/height on the
```
srallen, line 37: We're not supporting Safari 12 anymore, but no harm in the application of the workaround.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewer.js
```diff
@@ -21,6 +27,29 @@ const Container = styled.div`
...
+  container div, not the <svg>
+ */
+const SVG = styled.svg`
```
srallen, line 40:
```suggestion
export const SVG = styled.svg`
```

See comment on line 51...

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewer.js
```diff
@@ -21,6 +27,29 @@ const Container = styled.div`
...
+  height: 100%;
+
+  ${props => props.gridMaxWidth
+    ? css`max-width: ${props.gridMaxWidth};`
+    : ''}
+  ${props => props.gridMaxHeight
+    ? css`max-height: ${props.gridMaxHeight};`
+    : ''}
+`
+SVG.displayName = 'styledSvg'  // Required for testing
```
srallen, line 51: Styled components generates unique names for the components wrapped by styled. It can be disabled by way of the babel plugin: https://styled-components.com/docs/tooling#control-the-components-displayname. I'm not sure if or what impact it might have to overwrite the display name manually like this. If this is just for testing purposes, however, you can export the component itself and import it so you can find the component instance itself rather than find by display name.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewer.js
```diff
@@ -54,15 +85,22 @@ const SubjectGroupViewer = forwardRef(function SubjectGroupViewer(props, ref) {
     
   const annotationMode = interactionMode === 'annotate' && isCurrentTaskValidForAnnotation
   
+  console.log('+++ C ', gridMaxWidth, gridMaxHeight)
```
srallen, line 88: Leftover console log?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewer.spec.js
```diff
@@ -75,7 +75,7 @@ describe('Component > SubjectGroupViewer', function () {
   })
   
   it('should have a viewBox proportional to width and height', function () {
-    const viewBox = wrapper.find('svg').prop('viewBox')
+    const viewBox = wrapper.find('styledSvg').prop('viewBox')  // The styled <svg> needs a custom .displayName
```
srallen, line 78: Import the exported styled SVG component and you can do something like `wrapper.find(SVG)` instead.

srallen (approved):
LGTM. I went through your testing steps with all the latest versions of Safari, Chrome, Firefox, and Edge (on browserstack) and it's behaving as it should. I have a non-blocking suggestion with the tests when having to do with styled components. Thanks!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewer.js
shaunanoordin: Yikes, it is! Thank you for spotting this.

shaunanoordin:
Thank you, @srallen ! 😄  I've adjusted the styled SVG testing based on your recommendation, and it looks much better now. 👍

Alright, just need to update with the `master` branch and re-run the tests, and we're good to go!

Issue #2205, opened 2021-05-26
Workflow assignment prompting


srallen:
## Is your feature request related to a problem? Please describe.

"Back end" for #2143. We need code to manage when to prompt.

## Describe the solution you'd like

User project preferences should be requested for and stored with the actions for:

- Fetching the UPP
- Updating the UPP (will be passed down as prop to classifier)
- Last seen tutorial ids (will be passed down as prop to classifier)
- A computed view for comparing the project assigned workflow to the currently chosen workflow

## Describe alternatives you've considered
We considered putting this logic in the classifier, but discovered that the UPP is being requested for twice and the fact the routing is managed by the app too, the project app is a better choice for the logic

## Additional context
The stats model also relies on the UPP, so consideration needs to be made for this.

PR #2238, opened 2021-06-10
Integrate UPP store in project app

srallen:

Package: app-project

Toward #2205 

This has three main parts:

- A refactor of `YourStats` to only handle the stats server request and data
- The addition of `UserPersonalization` to handle the shared bits between `YourStats` and UserProjectPreferences`
- The async fetch and storing of the UPP resource

I initially thought about using MST's [`types.compose`](https://mobx-state-tree.js.org/API/#compose) in `UserPersonalization` of two stores, but eventually to remove the redundancy between the project app and the classifier, the UPP store will have other HTTP requests it is handling for updating the UPP and creating it initially, so it async state needs to be tracked separately from the user stats. 

`UserPersonalization` now handles the merging of the two different stats tracked by the stats service and on the UPP.

A future refactor should include reworking the `YourStats` and `Collections` stores to not automatically make requests when their nodes attached to their parents. It made the tests for separate stores more complicated having to stub or mock for them when the tests in the UPP store have nothing to do with them. (We need to do similar refactors in the classifier stores too).

I chose nock rather than mocking the fetch requests as I find it a bit easier to work with in this case.

The components store connections are also updated and when running the app, the component runs fine while signed out. Testing the signed in user is impossible locally (due to CORS and browser SSL security now), so the full manual QA will happen with the staging deploy. 

@goplayoutside3 asking for review from you since I think you have some flexibility right now still, but let me know if you need to go work on the tile feature and we can ask someone else. Thanks!

# Review Checklist

## General

## Components

## Apps

- [x] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapperConnector.js
+++ b/packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapperConnector.js
@@ -14,8 +14,7 @@ function useStore() {
     user,
     ui: {
       mode
-    },
-    yourStats
+    }
   } = store
 
   return ({
@@ -24,7 +23,7 @@ function useStore() {
     project,
     recents,
     user,
-    yourStats
+    yourStats: user.personalization
   })
 }
 
--- a/packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapperConnector.spec.js
+++ b/packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapperConnector.spec.js
@@ -5,7 +5,6 @@ import React from 'react'
 import asyncStates from '@zooniverse/async-states'
 
 import initStore from '@stores/initStore'
-import ClassifierWrapper from './ClassifierWrapper'
 import ClassifierWrapperConnector from './ClassifierWrapperConnector'
 
 describe('Component > ClassifierWrapperConnector', function () {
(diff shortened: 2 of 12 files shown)


goplayoutside3 (approved):
LGTM! I looked through all the changes, but understand this can’t be tested locally due to user sign-in and CORS.

Also, noting that I dismissed a Github warning about using ‘Bearer’ authorization because it was only used in the spec files 👍

srallen:
I tested out https://frontend.preview.zooniverse.org/projects/blicksam/transcription-task-testing/classify/workflow/13898 after deployment and started with 2 classifications from a previous session loaded from my UPP. Then made one classification which increased my session count to 1 from 0 and my total count from 2 to 3. The bar chart shows 1 classification for Monday.

I refreshed the page and I got 1 classification for today's session and 2 total. This is expected because Panoptes's counters take a moment to update. The bar chart showed 1 classification for Sunday and 1 for Monday, which is incorrect. I checked the data for `thisWeek` prop in the react dev tools:

```
[{count: 1, period: "2021-06-14"},
{count: 0, period: "2021-06-15"},
{count: 0, period: "2021-06-16"},
{count: 0, period: "2021-06-17"},
{count: 0, period: "2021-06-18"},
{count: 0, period: "2021-06-19"},
{count: 0, period: "2021-06-20"}]
``` 

And double checked the network response from the stats server, and they both look correct, showing 1 count for today. But the child component, which I think is `DailyClassificationsChart` and has the prop, `stats`, has a count for both Sunday and Monday of one:

Screen Shot 2021-06-14 at 1 38 07 PM

There's a function in that component converting the `thisWeek` data into a format for the plot, `stats`, so I think I found a bug. I checked production and this is incorrect there too, so unrelated to the changes of this PR: 

Screen Shot 2021-06-14 at 1 41 57 PM

I'll double check tomorrow what the stats are when loading up.

PR #2243, opened 2021-06-14
Refactor SurveyTask Choice render, from Drop, to Stack

### LINKED ISSUE #2242: Render Choice inline (refactor from Drop to Box)
Per Grommet Drop scroll/focus issue noted in #2235 refactor the Choice component inline.

mcbouslog:
Package:
- lib-classifier

Closes #2242.

Staging:

Describe your changes:
- remove Grommet Drop component
- adds Grommet Stack component
- updates SurveyTask tests accordingly

Notes:
- subsequent PR incoming to disable Done or equivalent (only visible if choice content is small, like Fire)
- subsequent PR to fix focus to first input
- subsequent PR to fix dark theme

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/SurveyTask/components/SurveyTask.js
+++ b/packages/lib-classifier/src/plugins/tasks/SurveyTask/components/SurveyTask.js
@@ -1,4 +1,4 @@
-import { Box, Drop } from 'grommet'
+import { Stack } from 'grommet'
 import PropTypes from 'prop-types'
 import React from 'react'
 
@@ -20,17 +20,13 @@ function SurveyTask (props) {
     task
   } = props
 
-  const choiceTargetRef = React.useRef()
-
   function handleCancel () {
     handleAnswers({})
     handleChoice('')
   }
 
   return (
-    <Box
-      ref={choiceTargetRef}
-    >
+    <Stack fill>
       <Chooser
         autoFocus={autoFocus}
         disabled={disabled}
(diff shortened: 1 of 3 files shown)


packages/lib-classifier/src/plugins/tasks/SurveyTask/components/SurveyTask.js
```diff
@@ -1,4 +1,4 @@
-import { Box, Drop } from 'grommet'
+import { Stack } from 'grommet'
```
mcbouslog, line 1: Stack covers the Done or equivalent buttons if the content is long enough, feels close (the same?) to the original design, though clicking outside of and other aspects Drop provided aren't included in Stack, thought would include for initial thoughts, though can easily remove and just use Box, let me know what you think.

srallen (approved):
LGTM. A few non-blocking comments though:

The filter uses a drop button that has the same scroll to top problem. Ok for this to discuss and have this be a separate refactor.

We may have to disable the choices' buttons as well when a choice is absolutely positioned on top. Imagine if the nothing here choice didn't have the image:

Screen Shot 2021-06-14 at 2 38 13 PM

The other choices are partially visible. Clicking on them doesn't seem to do anything, but maybe visually indicating and disabling those buttons would be a good idea? Maybe this can be part of the PR for disabling done as well?

I think the thumbnail images have an issue. I'm seeing a bunch of 415s maybe because the width being passed as a parameter to the service is `26.25`? We may need to make it a whole number?

mcbouslog:
Thanks and def agree about filters refactor and choices' buttons. Next PR will be to disable choices' buttons and Done while making a choice, and depending on how big it is (not looking big atm) also initial focus for first input.

Issue #1991, opened 2021-01-19
Review survey task accessibility, focus, and key controls


mcbouslog:
After #1990, survey task selections made (choice, filter, question, etc.) will persist and function as expected, at which point it will be important to re-review accessibility, focus, and key controls throughout the UX of the survey task.

This issue can be used to note specifics identified in other related issues or PRs, or however helpful.

ToDos:
- [x] radio input can be unselected/cleared (#2194)
- [x] clicking Not This in Choice deletes choice from annotation (#2247)
- [x] Choice Question inputs focus (#2250)
- [x] refactor Choices as [grid columns](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Auto-placement_in_CSS_Grid_Layout#auto-placement_by_column) or other way to improve accessibility, [related comment](https://github.com/zooniverse/front-end-monorepo/pull/1979#pullrequestreview-572530164) (#2256) 
- [x] Choices onKeyDown similar to PFE (#2256) 
- [x] refactor StyledHorizontalRule to border on FilterStatus (#2265)
- [x] ChoiceButton custom theme, [related comment](https://github.com/zooniverse/front-end-monorepo/pull/1979#discussion_r561327574) (#2265)
- [x] Choice dark theme styling (#2265)
- [x] radio input clear on keydown (PFE = space or backspace) (#2305)
- [x] with disable prop, disable Choices and FilterStatus (#2308)

mcbouslog:
Great suggestion from @eatyourgreens - make list of choices an HTML list for improved accessibility and navigatablity

eatyourgreens:
If we keep the list of choices as a grouped control, which it is in PFE, then it should probably have a menu role, with each choice wrapped in a menuItem. I forgot that in the original code.

eatyourgreens:
The list of choices should also be a single list, styled into grid columns, not three lists.

PR #2250, opened 2021-06-17
Add conditional focus to survey task Choice components

mcbouslog:
Package:
- lib-classifier

Towards #1991.

Staging:

Describe your changes:
- conditionally adds focus to Choice components, as follows:
- Choice with more than one image focus (after 1 tab) on Carousel
- Choice with no images, yes confusions, focus on first confusion button
- Choice with no images or confusions, yes questions, focus on first question
  - if first question is checkbox then first input gets focus
  - if first question is radio and no existing answer then first input gets focus
  - if first question is radio and existing answer then existing answer gets focus
- Choice with no images, confusions, or questions focus is on "Identify" button

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/SurveyTask/components/components/Choice/Choice.js
+++ b/packages/lib-classifier/src/plugins/tasks/SurveyTask/components/components/Choice/Choice.js
@@ -33,6 +33,15 @@ export default function Choice (props) {
   const questionIds = getQuestionIds(choiceId, task)
   const allowIdentify = allowIdentification(answers, choiceId, task)
 
+  let hasFocus = 'identify'
+  if (choice.images?.length > 1) {
+    hasFocus = 'images'
+  } else if (choice.confusionsOrder?.length > 0) {
+    hasFocus = 'confusions'
+  } else if (questionIds.length > 0) {
+    hasFocus = 'questions'
+  }
+
   return (
     <Box
       background='light-1'
@@ -42,6 +51,7 @@ export default function Choice (props) {
     >
       {choice.images?.length > 0 && (
         <Carousel
+          autoFocus={hasFocus === 'images'}
           controls='arrows'
         >
           {choice.images.map((filename, index) => (
@@ -57,6 +67,7 @@ export default function Choice (props) {
       <Paragraph>{choice.description}</Paragraph>
       {choice.confusionsOrder?.length > 0 && (
         <ConfusedWith
+          hasFocus={hasFocus === 'confusions'}
           choices={choices}
           confusions={choice.confusions}
           confusionsOrder={choice.confusionsOrder}
@@ -67,6 +78,7 @@ export default function Choice (props) {
       {questionIds.length > 0 && (
         <Questions
           answers={answers}
+          hasFocus={hasFocus === 'questions'}
           choiceId={choiceId}
           questionIds={questionIds}
           questions={questions}
@@ -92,6 +104,7 @@ export default function Choice (props) {
           onClick={() => handleDelete(choiceId)}
         />
         <PrimaryButton
+          autoFocus={hasFocus === 'identify'}
           disabled={!allowIdentify}
           fill='horizontal'
           label={counterpart('Choice.identify')}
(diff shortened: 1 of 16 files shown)


packages/lib-classifier/src/plugins/tasks/SurveyTask/components/components/Choice/Choice.js
```diff
@@ -42,6 +51,7 @@ export default function Choice (props) {
     >
       {choice.images?.length > 0 && (
         <Carousel
+          autoFocus={hasFocus === 'images'}
```
mcbouslog, line 54: `autoFocus` isn't a property of the `Carousel`, this should be deleted. I wasn't able to determine a simple way to focus the Carousel if rendered, please let me know if something comes to mind.

srallen: I had a look at the `Carousel` and under the hood, it passes all other props to a `Stack` and that under the hood is a styled div, so I don't think it'll work.

Check it out: https://github.com/grommet/grommet/blob/d4934e7d9749e67f953dcfb4a571b3a514d7d2e6/src/js/components/Carousel/Carousel.js#L178

If we want to focus this, we'll have to use javascript. I don't think we really need this to be auto-focusable. Getting to the inputs is more important.

srallen (changes requested):
While trying out https://local.zooniverse.org:8080/?project=1634&workflow=2434 I found that I couldn't click or keyboard space/enter the "Not this" option of a choice. I'm not getting an errors related to it. The button doesn't appear to be disabled in the DOM and has an onClick event handler defined, so I'm not sure what's going on here. I checked the behavior on master, and it's broken there too, so this is out of scope for this branch, but something to be aware of.

As far as requested changes, I'd say we can remove the autofocus from the carousel as noted.

mcbouslog:
I made a mistaken during a previous PR rebase and broke the "Not this" functionality, but include a fix in #2256 [here](https://github.com/zooniverse/front-end-monorepo/pull/2256/files#diff-440b4febc47c080a86bf7666488637387d0cd9cdae19300a515b5f58ca646036R34).

packages/lib-classifier/src/plugins/tasks/SurveyTask/components/components/Choice/components/Questions/components/CheckBoxInputs/components/CheckBoxInput.js
```diff
@@ -15,11 +15,16 @@ export const StyledBox = styled(Box)`
     opacity: 0.01;
     position: absolute;
   }
+
+  &:focus-within {
+    box-shadow: 0 0 2px 2px ${props => props.theme.global.colors['accent-1']};
```
srallen, line 20: Non-blocking: the theme has a designation for a focus color at `theme.global.colors.focus` which right now is mapping to `accent-1`, but if we change it, then this wouldn't need to be updated in that case.

packages/lib-classifier/src/plugins/tasks/SurveyTask/components/components/Choice/components/Questions/components/RadioInputs/components/RadioInput.js
```diff
@@ -15,11 +15,16 @@ export const StyledBox = styled(Box)`
     opacity: 0.01;
     position: absolute;
   }
+
+  &:focus-within {
+    box-shadow: 0 0 2px 2px ${props => props.theme.global.colors['accent-1']};
```
srallen, line 20: Same comment about using the focus attribute in the theme object.

srallen (approved):
There's some awkwardness with the focus scrolling to the focused input while the choice image is still loading, but it's not a blocker for me. I can explain more on call if needed.

packages/lib-classifier/src/plugins/tasks/SurveyTask/components/components/Choice/components/Questions/components/CheckBoxInputs/components/CheckBoxInput.js
mcbouslog: forgot about that, always good to set programmatically 👍

mcbouslog:
> There's some awkwardness with the focus scrolling to the focused input while the choice image is still loading, but it's not a blocker for me. I can explain more on call if needed.

Added a note about this to #1991 to revisit shortly.

PR #2253, opened 2021-06-18
Use Markdownz to render text task instructions.

### LINKED ISSUE #2252: Markdown inconsistency
## Describe the bug
This bug has been reported by our National Archives collaborator on the Engaging Crowds project:
Project - https://frontend.preview.zooniverse.org/projects/bogden/scarlets-and-blues
Lab - https://www.zooniverse.org/lab/12268

## To Reproduce
From our collaborator:

"I promised to send an example of markdown-in-tasks behaving inconsistently. So here are some screenshots.
 
good_markdown_input is task T9 of the MinutesSimple workflow (workflow #18504, and I think a Checkbox task)
good_markdown_output is visible in the MinutesSimple workflow. Press “Attendance page” on the first screen to see it at the top of the next screen.
bad_markdown_input is task T1 of the IndexSimple workflow (workflow #18505, and a Text task)
bad_markdown_output is visible in the IndexSimple workflow. Press “Yes” on the first screen to see it at the top of the next screen (but I’m going to change this back to a form that doesn’t rely on markdown in a moment).
 
I’m calling this a markdown inconsistency, but the behaviour I’m seeing is just rendering or not rendering line breaks. I guess that happens via markdown, but that could be a bad assumption on my part."

## Expected behavior
A line break should either appear or not appear each time (consistent behaviour)

## Screenshots
bad_markdown_input
bad_markdown_output
good_markdown_input
good_markdown_output

srallen:

Package:

Closes #2252 

The `TextTask` instructions should be rendered using `Markdownz`. This fixes that.

Before: https://frontend.preview.zooniverse.org/projects/srallen/untitled-project-3-25-2021-12-33-27-pm/classify/workflow/18818

Screen Shot 2021-06-18 at 1 09 54 PM

After: https://localhost:8080/?project=15565&workflow=18818&env=production

Screen Shot 2021-06-18 at 1 12 38 PM

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/components/DefaultTextTask/DefaultTextTask.js
+++ b/packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/components/DefaultTextTask/DefaultTextTask.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { Box, Text, TextArea } from 'grommet'
+import { Box, TextArea } from 'grommet'
+import { Markdownz } from '@zooniverse/react-components'
 import TextTagButtons from '../TextTagButtons'
 
 export default function DefaultTextTask (props) {
@@ -26,7 +27,7 @@ export default function DefaultTextTask (props) {
       <label
         htmlFor={`${task.taskKey}-${task.type}`}
       >
-        <Text>{task.instruction}</Text>
+        <Markdownz>{task.instruction}</Markdownz>
         <TextArea
           ref={textArea}
           autoFocus={autoFocus}
--- a/packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/components/DefaultTextTask/DefaultTextTask.js
+++ b/packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/components/DefaultTextTask/DefaultTextTask.js
@@ -1,8 +1,14 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { Box, TextArea } from 'grommet'
+import { Box, Text, TextArea } from 'grommet'
 import { Markdownz } from '@zooniverse/react-components'
+import styled from 'styled-components'
 import TextTagButtons from '../TextTagButtons'
+
+const StyledText = styled(Text)`
+  display: block;
+  margin: 10px 0;
+`
 
 export default function DefaultTextTask (props) {
   const {
@@ -20,6 +26,18 @@ export default function DefaultTextTask (props) {
     updateAnnotation(textArea)
   }
 
+  const components = {
+    a: StyledText,
+    h1: StyledText,
+    h2: StyledText,
+    h3: StyledText,
+    h4: StyledText,
+    h5: StyledText,
+    h6: StyledText,
+    p: StyledText,
+    span: Text
+  }
+
   return (
     <Box
       direction='column'
@@ -27,7 +45,7 @@ export default function DefaultTextTask (props) {
       <label
         htmlFor={`${task.taskKey}-${task.type}`}
       >
-        <Markdownz>{task.instruction}</Markdownz>
+        <Markdownz components={components}>{task.instruction}</Markdownz>
         <TextArea
           ref={textArea}
           autoFocus={autoFocus}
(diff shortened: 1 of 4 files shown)


eatyourgreens:
Block level elements, like `<p>` aren’t valid content for `<label>`. If we allow text formatting for input labels, we should do it in a way that produces valid, accessible HTML.

srallen:
@eatyourgreens the `Markdownz` component takes a prop to customize which React components to render for certain Markdown. Do you have a suggestion on what this should be?

eatyourgreens:
The [MDN docs for label](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#technical_summary) say [phrasing content ](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#phrasing_content) is allowed, so any inline element, I'd guess. No headings, paragraphs, blockquotes, lists, code blocks or links. (Links are excluded because `label` is a clickable element already.)

srallen:
@eatyourgreens I defined a component mapping that basically uses styled like paragraph spans. I think I've mostly seen project owners use the markdown in task instructions for bold, italics, images, and paragraph styling, so I think this should cover most of it. 

I made another test workflow using instructions copied from the UK Tides just to demo it more (paragraph like spans and the bold renders as expected): https://localhost:8080/?project=15565&workflow=18862&env=production

If you want to propose implementing a character limit or removing markdown from task instructions/questions, then I'd recommend opening an RFC so we can discuss the impact and come up with a migration plan.

eatyourgreens:
Sounds good. Looking at the project builder, the task instruction is a plain text input, not a Markdown editor. I don't think we'd expect project owners to put much Markdown format into a textarea. I am still hesitant about changing Becky's design for the text task but that might be more a case of guiding project owners to write accessible task instructions.

#2252 also raised an issue with form legends in the task area. Bernard asked why the text task instruction is formatted differently to the multiple choice instructions for [workflow 18504](https://www.zooniverse.org/lab/12268/workflows/18504). It turns out that question and drawing tasks are using fieldset legends (which is absolutely correct for an checkbox instruction) but legends, like labels, can't contain paragraphs of text. A legend is a short instruction which JAWS and NVDA will read out as part of the label for each input in that fieldset eg. if the legend were 'Pick some colours' then JAWS would read the checkboxes as 'Pick some colours: red', 'Pick some colours: blue' etc.

That totally fails if the legend is long. Maybe not a big problem for Scarlets & Blues but Davy Notebooks has a transcription tool radio button input that JAWS would announce as 
>Radio button: Welcome to the transcription task. Let's transcribe at least one line of Davy's handwritten text. There are two ways to transcribe, depending on whether anyone else has worked on this notebook page previously: If the page has no previous volunteer-made marks (coloured lines drawn under Davy's text): underline a single line of Davy's text by clicking at the start and end of the line (draw your marks in the order you’re reading the text), and follow the instructions on the pop-up for transcribing the text you’ve underlined. If the document has previous annotations: click on an already-existing underline mark to view, select, edit, and/or agree with previous transcriptions. Single line of text 0 drawn

Legends also need to be fixed, so that they validate. I think we've also got an accessibility problem where project owners are typing long passages of text into the Main Text text area, when they should be using the task help.

FWIW long form task help (and error messages) can be linked up to an input with [`aria-describedby`](https://www.tpgi.com/describing-aria-describedby/) although, as noted in that article, linked content will be concatenated into a single string of text.

eatyourgreens (commented):
Code changes look good. I'm holding off on approval until I've had a chance to double-check against the design, but 10px margin seems fine to me.

I'm not sure about closing #2252 while legends are still broken for the question tasks.

eatyourgreens (approved):
The margins looks good to me. I'm seeing some tasks without labels, but I'm guessing that's broken in the workflow itself.
https://localhost:8080/?project=12268&workflow=17077&subjectSet=92753&env=production&demo=true
Screenshot of a Scarlets & Blues transcription task, showing that some of the text boxes are not labelled.

srallen:
> Looking at the project builder, the task instruction is a plain text input, not a Markdown editor. I don't think we'd expect project owners to put much Markdown format into a textarea.

I don't know why it isn't a markdown editor, but the instructions in the lab state that markdown is allowed and it's been this way for years. Whenever we get around to actually build a new workflow editor, then this can change, either way. 

As for the other issues, could you write up issues for them separately? We may not want to do the same resolutions for other similar problems. Thanks.

PR #2254, opened 2021-06-21
Classifier: add image placeholder for Multi Frame Viewer

### LINKED ISSUE #2221: Transcription Task "Fan Lines" bug
## Fan Lines appearing
When leaving/returning to the classify page, if in-progress transcription isn't submitted the volunteer-drawn underline markings will all gather together in the upper left corner of the image and 'fan out'. 

## To Reproduce
1. Go to https://www.zooniverse.org/projects/humphrydavy/davy-notebooks-project/classify/workflow/18575
2. Start a classification
3. Click on the project home page link
4. Click on the workflow and return to the classify page

NB: this also works if you log out of your account mid-classification, start a classification while not logged-in and log into your account mid-classification, or press the 'back' button on your browser mid-classification & return to the classify page.

project number 9006
workflow number 18575

## Expected behavior
Honestly, I'm not entirely sure what I would expect here from navigating away & returning. Seems to me like a good use case for the "Are you sure you want to leave?" prompt, though I'm less sure of the expected sign-in behavior -- I don't think I would expect classification data to persist, but am also unsure if we support that for other projects (e.g. I can see a user forgetting to log in, getting halfway through a transcription and realizing it, and wanting to sign in so that the page will be in their Recents).

## Screenshots
['fan' bug + console images](https://imgur.com/a/OmNWD8t)
NB these screenshots were taken after generating the bug through mid-classification sign-out.

 - OS: macOS Catalina
 - Browser Chrome 90

## Additional context
[Talk thread](https://www.zooniverse.org/projects/humphrydavy/davy-notebooks-project/talk/2586/2013546) with original report

shaunanoordin:
## PR Overview

packages: code change in `lib-classifier`, but can only be tested on `app-project`
Fixes #2221
Follows #2241 

This PR is a follow up to 2241, which adds an image placeholder for the Single Image Viewer, to solve the issue of drawing components not fully working when the Classifier page is _re-visited after navigating away from it._

- The problem context and solution is the same, **just on different viewers.**
- This PR addresses the issue where the Davy Notebooks projects is STILL encountering the issue "solved" by 2241, since D-Notes (as the cool kids call it) is using an MF Viewer not an SI Viewer.

### Testing

BE SURE TO TEST THIS ON DAVY NOTEBOOKS

1. Bootstrap everything
2. Open the Classify Page of Davy Notebooks: http://localhost:3000/projects/humphrydavy/davy-notebooks-project/classify?env=production (this WF should be using Multi-Frame Viewer)
3. Draw some Transcription Lines, annotate some text, etc.
4. Navigate to the Project home page
5. Return to the Classify page by clicking on the 'Classify' link on the project nav
  - Expected: the Transcription Lines should _look_ correct. (Without this PR, the scale of the drag handles shift wildly since it defaults to a scale of 1.0 instead of calculating the size relative to the browser window)
  - Expected: you should be able to modify existing Transcription Lines by dragging their start/end handles
  - Expected: you should be able to delete existing T-Lines and draw new T-Lines.

### Status

WIP. TODO:

- update tests.
- update Stories.
- check that everything still works with multi-image subjects (while testing today, I've only seen D-Notes subjects with only 1 image.)

Tagging @snblickhan so she has eyes on this for D-Notes.

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewer.stories.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewer.stories.js
@@ -10,6 +10,7 @@ import ImageToolbar from '../../../ImageToolbar'
 import SubjectViewerStore from '@store/SubjectViewerStore'
 import readme from './README.md'
 import backgrounds from '../../../../../../../.storybook/lib/backgrounds'
+import asyncStates from '@zooniverse/async-states'
 
 const config = {
   notes: {
@@ -88,6 +89,7 @@ storiesOf('Subject Viewers / MultiFrameViewer', module)
       <ViewerContext subject={subjectWithMore} theme={zooTheme} themeMode='light'>
         <Box>
           <DecoratedMultiFrameViewerContainer
+            loadingState={asyncStates.success}
             enableInteractionLayer={false}
             subject={subjectWithMore}
           />
@@ -101,6 +103,7 @@ storiesOf('Subject Viewers / MultiFrameViewer', module)
       <ViewerContext subject={subjectWithMore} theme={darkZooTheme} themeMode='dark'>
         <Box>
           <DecoratedMultiFrameViewerContainer
+            loadingState={asyncStates.success}
             enableInteractionLayer={false}
             subject={subjectWithMore}
           />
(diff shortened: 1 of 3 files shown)


eatyourgreens (commented):
Changes look good to me. I’ve opened #2255 because I think SVGContext can be provided in one place.

shaunanoordin:
## PR Update

Alright, this is ready! I've updated the tests and Storybook, so now just to wait for the automated tests to finish running. 👍

eatyourgreens (approved):
Looks good to me. At some point, the stories should be migrated away from `storiesOf`.

shaunanoordin:
Thanks Jim! 👍

Issue #2273, opened 2021-06-26
Classifier: drag-to-pan is wonky when Subject Image is rotated in Subject Viewer


shaunanoordin:
## UI Bug

Package: `lib-classifier`
Affects: the Single Image Viewer (possibly Multi-Frame Viewer too), _if_ it has Rotation enabled.

On the Classifier's Single Image Viewer, users can perform a **"drag-to-pan"** interaction. (Enable Move mode on the image toolbar. Click & drag the mouse left on the subject image. The image should now move left) This ONLY works when the subject image is **not rotated.** (i.e. rotation = 0º)

- If the subject image is rotated (via the image toolbar), then the behaviour is harder to predict. e.g. dragging left might cause the image move right.
- Of note, the "press-keyboard-keys-to-pan" interaction is WORKING FINE. Pressing the Right keyboard key will always move the image to the right, no matter how it's rotated.

### Testing

- open lib-classifier with any project/workflow that uses a Single Image Viewer, e.g. https://localhost:8080/?env=production&project=1292&workflow=18372 . or I Fancy Cats works too.
- Drag ➡️  right to pan the view. Observe that the image moves to the ➡️ right, as expected.
- Rotate the image once, so it's 90º.
- Drag ➡️  right to pan the view. Observe that the image moves ⬇️ ➡️ DOWN-RIGHT.
- Rotate the image again, so it's 180º.
- Drag ➡️  right to pan the view. Observe that the image moves ⬅️ LEFT.

Additional:
- Repeat the steps above, but instead of dragging with a mouse, press the ➡️ Right Arrow key on your keyboard. Observe that the image always moves to the ➡️ right, no matter the rotation.

### Dev Notes, Short version

Analysis:

- SVGPanZoom's `onDrag(event, difference)` function (which handles "drag-to-pan", and behaves incorrectly) should learn from its sibling `onPan(dx, dy)` function (which handles "press-keyboard-keys-to-pan")
  - `onPan()` does NOT factor for the rotational transform in `<g#transform-layer>`, and works fine.
  - `onDrag()` DOES, and has bugs.
- The `onDrag()` hander is tied to the `draggable.js`'s `dragMove()` function, which relies on its sibling `getEventOffset()` function, which _converts DOM coordinates (relative to the screen) to SVG coordinates (including rotational transformations)._

Proof of concept:
[...]

PR #2275, opened 2021-06-28
Classifier: replace SVGPoint with DOMPoint

### LINKED ISSUE #2268: Classifier: SVGPoint deprecated, replacement needed
## Code Update (Refactor?) Required

Package: `lib-classifier`
Follows #2255

Context: while refactoring how the classifier passes svg references around with the SVGContext, @eatyourgreens noticed that a necessary component of our svg-related calculations is now out of date and in danger of no longer being supported.

Overview:
- **SVGPoint** has been deprecated by modern browsers and a replacement is needed.
- SVGPoint is used by the Classifier's 'InteractionLayer' component and 'draggable' class to convert _event coordinates_ (i.e. DOM coordinates) to _SVG coordinates_ (i.e. where the user clicked on the Subject image)

Further Reading:
- [MDN documentation for SVGPoint,](https://developer.mozilla.org/en-US/docs/Web/API/SVGPoint) indicating its deprecation

### Dev Notes

SVGPoints are used in two locations in `lib-classifier`:
- [InteractionLayer.js](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js) - used by drawing tasks, and is the layer "on top" of the Subject image where drawing marks appear.
- [draggable.js](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/plugins/drawingTools/components/draggable/draggable.js) - base class used to implement the "drag-to-move", "drag-to-resize", "drag-to-rotate" interactions used by various drawing-related components.

In `InteractionLayer.js` for example, we have this code:

```
function getEventOffset(event) {
  const svgPoint = svg.createSVGPoint()
  svgPoint.x = event.clientX
  svgPoint.y = event.clientY
  const svgEventOffset = svgPoint.matrixTransform(getScreenCTM().inverse())
  return svgEventOffset
}

// For context, getScreenCTM() belongs to the main transform layer, aka the <g> element that contains the transform ops performed on the subject image + its annotations (the InteractionLayer).
```

SVGPoint's .matrixTransform() is basically the lynchpin that automagically transmogrifies event coordinates/DOM coordinates to SVG-relative coordinates.

_If all else fails,_ I think it's possible to reverse-engineer the matrix transformation calculations. mainTransformLayer.getScreenCTM() [still works](https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement) and returns a [DOMMatrix](https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrix) with all the necessary values that someone smarter than me can mathemagically figure out.

Surely this is solvable. I mean... _it's just a matrix, right?_

The Matrix

Notes to self:
- A quick scan of the [most recent article](https://www.sitepoint.com/how-to-translate-from-dom-to-svg-coordinates-and-back-again/) (Dec 2020) I could find on the topic of DOM-SVG coordinate transformations still uses SVGPoint, so time to dig deeper into Stackoverflow.

### Status

Solution requires research.

At the moment, our Classifier continues to work, and there are no indications that modern browsers are going to pull out SVGPoint support any time _soon,_ but that "deprecation alarm" is unsettling enough as is.

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Affects: drawing tools
Closes #2268 

This PR replaces [SVGPoint](https://developer.mozilla.org/en-US/docs/Web/API/SVGPoint) with the funky fresh face of [DOMPoint](https://developer.mozilla.org/en-US/docs/Web/API/DOMPoint), which is so funky fresh that it's barely documented.

- But trust me, DOMPoint does indeed have `matrixTransform()`, and it _shooould_ work on all major modern browsers.
- Please see #2268 for full details.

### Testing

Test Project: basically any single-image workflow with a Drawing Task.
Test URL: https://localhost:8080/?env=production&project=1292&workflow=18372 , or I Fancy Cats work too

Testing steps:
- Switch to Move mode. Drag to pan the view on the Subject Image.
- Switch to Annotation mode. Draw a Point, Ellipse, or Rectangle. 
- Drag these marks to move them.
- For Ellipse or Rectangle, drag their "resize" handles to resize. 
- Rotate the image to test that drags continue to work with rotation.
- Drag-to-move and drag-to-resize the Points/Ellipses/Rectangles again.

Expectations for these actions should be pretty clear cut.

⚠️ NOTE: drag-to-pan (the view of the Subject Image) does NOT work when the image is rotated. This is documented in #2273

⚠️ NOTE: I noticed that with Rotating Rectangle, the resize handles don't work too well when the rectangle is rotated. This is unique to the Rotating Rectangle and I'm going to take a look at that separately.

### Status

Ready for review!


shaunanoordin:
Ah, _crap,_ I'm getting a lot of problems on the tests. "ReferenceError: DOMPoint is not defined" ??

shaunanoordin:
Setting this PR as a draft while I figure a fix for the tests. Sorry for the noise.

eatyourgreens:
@shaunanoordin check out #2277, where I've tried to use `DOMPointReadOnly` in a way that doesn't break with `jsdom`.

shaunanoordin:
Oh dang, forgot to update this branch! This branch is superseded by #2277 which has a more thorough approach to upgrading the SVG systems as a whole. I'm going to go ahead and close this.

In case #2277 is delayed and we just want to get 2275 in with a quick fix for SVGPoints (without worrying about the SVG system as whole), the key lessons from the former PR is this:

- When creating a DOMPoint, have a fallback to create a default point type to satisfy jsdom. (This is what was holding back this PR)
    ```
    function createPoint(event) {
      const { clientX, clientY } = event
      // SVG 2 uses DOMPoint
      if (window.DOMPointReadOnly) {
        return new DOMPointReadOnly(clientX, clientY)
      }

      // jsdom doesn't support SVG
      return {
        x: clientX,
        y: clientY,
        matrixTransform: () => { bla bla bla }
      }
    }
    ```
- We do NOT need to support SVGPoint, as long as we support DOMPoint + default point.

PR #2276, opened 2021-06-28
Classifier: fix "Reset Image/View" action (fix 2274)

### LINKED ISSUE #2274: Classifier: "Reset Image/View" action doesn't reset image correctly
## UI Bug

Package: `lib-classifier`
Affects: "Reset Image/View" button/action on the Image Toolbar ; Single Image Viewer, possibly Multi-Frame Viewer

The "Reset Image/View" action is slightly wonky. Clicking the button on the Image Toolbar will cause the the _zoom_ and _rotation_ to work correctly, but the _pan_ (or translate, or offset) is slightly off.

_When the Classifier loads, the 1000x1000 Subject image is positioned and rescaled perfectly:_

Screen Shot 2021-06-28 at 17 30 13

_When I press the "Reset Image/View" button, the 1000x1000 Subject Image is offset by 100px horizontally, 200px vertically:_

Screen Shot 2021-06-28 at 17 30 20

_The expected offset should be 0px, 0px._

Inspecting the code, we see that the SVG's has an attribute of `viewBox="-100 -200 1000 1000"` instead of the expected `viewBox="0 0 1000 1000"`

This was probably introduced by #2241 (and by extension #2254 ) which introduces a placeholder image (800px by 600px) for when the Image file is loading in a SingleImageViewer (and MultiFrameViewer).

### Analysis

OK, so...

- The problem is in SVGPanZoom.
- The SVGPanZoom's `onZoom()` function (which is used by the "Reset Image/View" action) seems to _remember the naturalWidth and naturalHeight of the **placeholder image.**_

**Proof of concept.** In SVGPanZoom.js, add the following console.logs:

```
function SVGPanZoom({ ... }) {
  ...
  function onZoom(type) {
    switch (type) {
      ...
      case 'zoomto': {
        setZoom(1)
        setViewBox({
          x: 0,
          y: 0,
          width: naturalWidth,
          height: naturalHeight
        })
        console.log('+++ zoomto: ', naturalWidth, naturalHeight)
      }
    }
  }

  const { x, y, width, height } = scaledViewBox(zoom)
  const scale = imageScale(img)
  
  console.log('+++ render: ', naturalWidth, naturalHeight)

  return ( ... )
}
```

Load the classifier and click "Reset Image/View". You should see the following in the console...

> resetting view
> +++ zoomto:  800 600
> +++ render:  1000 1000

Yup, those naturalWidth and naturalHeight values are different. Looks like the onZoom function exists with its own variable scope instead of pulling from the context/scope of the SVGPanZoom.

### Solutions?

If the onZoom() function has its own scope that is created _only when the SVGPanZoom component is initialised,_ and the problem occurs _when the image changes (from placeholder to actual image),_ then logically a solution would be to re-create the onZoom() function (and hence its scope (wait, is the term _closure?_ )) when the image changes.

i.e. change

```
useEffect(() => {
    if (zooming) {
      enableZoom()
      return disableZoom
    }
  }, [zooming])
``` 

to

```
useEffect(() => {
    if (zooming) {
      enableZoom()
      return disableZoom
    }
  }, [zooming, src])  // Just add src!
```

Ta da! A quick test (i.e. just applying the code and clicking "Reset Image/View") seems to work. However, I'm wary that this might lead to other gotchas, _(it's simple. too simple.)_ so I need to run additional tests. e.g.
- Does everything reset correctly when moving to a new Subject in the queue?
- What about MultiFrameViewers? Are there any issues when jumping between between frames?

If anybody can help sanity check me on this, that'd be ace.

### Status

Medium priority. Needs fixing since it's a broken behaviour, but it won't fully obstruct any user's experience.

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Fixes #2274

This PR fixes a bug (introduced in #2241) where the "Reset Image/View" doesn't correctly reset the image, and actually adds a horizontal+vertical pan offset.

- The problem is that the `onZoom()` function of SVGPanZoom stores/remembers the naturalWidth/naturalHeight of the _placeholder image_
- This fix refreshes the `onZoom()` function when the subject image loads and replaces the placeholder image.
- Please see #2274 for more details.

### Testing

Project: any project with a single image viewer
Testing URL: [I Fancy Cats!](https://localhost:8080/)

Testing Steps:
- Open the classifier and wait for subject image to load.
- Subject image should load with a zoom of 1.0, and translation/pan/offset of (0,0) , and a rotation of 0.
- Move the subject image around (drag to pan, or use keyboard arrow keys to pan), zoom in/out, and/or rotate the image.
- Press the "Reset Image/View" button on the Image Toolbar.
- Expectation: subject image should return to a zoom of 1.0, and translation/pan/offset of (0,0) , and a rotation of 0.

### Status

Ready for review! I actually remembered to check the tests before saying that this time, thank goodness.

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
@@ -42,7 +42,7 @@ function SVGPanZoom({
       enableZoom()
       return disableZoom
     }
-  }, [zooming])
+  }, [zooming, src])
 
   useEffect(
     function onZoomChange() {


eatyourgreens (approved):
Looks good to me.

PR #2277, opened 2021-06-29
Classifier: use DOMPointReadOnly for drawing

eatyourgreens:
Remove `SVGContext` from `InteractionLayer`. Use `DOMPointReadOnly`, if supported, instead of `SVGPoint`, which allows us to remove the passed SVG element from context. I've also renamed the passed element from `svg` to `canvas`, and this is passed down in order to perform matrix transformations on dragged tools etc.

Follows up on the changes in #2255.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
@@ -63,7 +63,7 @@ function InteractionLayer({
   function createPoint(event) {
     const { clientX, clientY } = event
     // SVG 2 uses DOMPoint
-    if (document.DOMPointReadOnly) {
+    if (window.DOMPointReadOnly) {
       return new DOMPointReadOnly(clientX, clientY)
     }
     // SVG 1.1 uses SVGPoint
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
@@ -34,7 +34,7 @@ function InteractionLayer({
   duration
 }) {
   const [creating, setCreating] = useState(false)
-  const svg = useRef()
+  const canvas = useRef()
 
   useEffect(
     function onDeleteMark() {
@@ -66,13 +66,6 @@ function InteractionLayer({
     if (window.DOMPointReadOnly) {
       return new DOMPointReadOnly(clientX, clientY)
     }
-    // SVG 1.1 uses SVGPoint
-    if (svg.current?.createSVGPoint) {
-      const svgPoint = svg.current.createSVGPoint()
-      svgPoint.x = clientX
-      svgPoint.y = clientY
-      return svgPoint
-    }
     // jsdom doesn't support SVG
     return {
       x: clientX,
@@ -83,7 +76,7 @@ function InteractionLayer({
   function getEventOffset(event) {
     const svgPoint = createPoint(event)
     const svgEventOffset = svgPoint.matrixTransform ?
-      svgPoint.matrixTransform(svg.current?.getScreenCTM().inverse()) :
+      svgPoint.matrixTransform(canvas.current?.getScreenCTM().inverse()) :
       svgPoint
     return svgEventOffset
   }
@@ -157,11 +150,9 @@ function InteractionLayer({
   }
 
   return (
-    <svg
-      ref={svg}
-      xmlns="http://www.w3.org/2000/svg"
-    >
+    <>
       <DrawingCanvas
+        ref={canvas}
         disabled={disabled || move}
         pointerEvents={move ? 'none' : 'all'}
         width={width}
@@ -185,8 +176,8 @@ function InteractionLayer({
           scale={scale}
           played={played}
         />
-      )}
-    </svg>
+    )}
+    </>
   )
 }
 
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
@@ -1,8 +1,7 @@
 import cuid from 'cuid'
 import PropTypes from 'prop-types'
-import React, { useEffect } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
 import styled, { css } from 'styled-components'
-import SVGContext from '@plugins/drawingTools/shared/SVGContext'
 import DrawingToolMarks from './components/DrawingToolMarks'
 import TranscribedLines from './components/TranscribedLines'
 import SubTaskPopup from './components/SubTaskPopup'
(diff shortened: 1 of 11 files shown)


eatyourgreens:
Using nested SVGs has done something weird to the viewbox. 🤔 

A zoomed-in cat photo from I Fancy Cats. The photo is cropped, rather than filling the subject viewer.

shaunanoordin:
> Using nested SVGs has done something weird to the viewbox. 🤔

That's weird, the annotations (layer) is being clipped to the... inner SVG's (transform layer's) viewbox? Let me see if I can pinpoint what's up

eatyourgreens:
It might be that it needs to be explicitly styled to fill its parent.

shaunanoordin:
I was looking if SVG element had an `overflow` property, but [it's not widely supported yet,](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/overflow) blagh

shaunanoordin:
Not sure if this [SVG 2.0 tidbit about clipping is relevant,](https://www.w3.org/TR/2013/WD-SVG2-20130409/masking.html#InitialClippingPath) since it's hard to grok.

> When an ‘svg’ element is either the root element in the document or is embedded within a document whose layout is determined according to the layout rules of CSS or XSL, then the user agent must establish an initial clipping path for the SVG document fragment. 

But what's really baffling me is that _in SVGs if you have any clipping/masking, you usually have to be explicit about it._

shaunanoordin:
Findings so far:

- Clipping is the result of nesting `<svg>`s. Each `<svg>` will limit the visibility of its children to its respective viewbox.
  - (Edit: huh, this might be intended. It seems that Inkscape is rendering _nested SVGs without clipping,_ and [somebody's complaining about it](https://gitlab.com/inkscape/inkscape/-/issues/1671))
- Upon further review, I'm not sure if we even need to support SVG 1.0's SVGPoint any more.
  - Anything that supports only SVG 1.0 and not SVG 2.0 will probably encounter other breaking issues with drawing tools, TBH.
  - Just supporting 1. DOMPoint and 2. a fallback for jsdom seems sufficient. 👌

eatyourgreens:
Aha! Then I will change them back to groups tomorrow, and drop `SVGPoint`. 

I think nesting SVGs reset the coordinate frame of their children too.

eatyourgreens:
@shaunanoordin I've updated this to remove the nested SVGs and drop `SVGPoint`. I think you're right: browsers that support pointer events will also support [`DOMPoint`](https://developer.mozilla.org/en-US/docs/Web/API/DOMPoint). All the matrix calculations are now handled by a styled group (`<g>`), which I've renamed from `svg` to `canvas` in the context. Dropping `SVGPoint` means we no longer have to pass around a ref to the parent SVG element, which simplifies a bit of code.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -1,8 +1,7 @@
 import cuid from 'cuid'
 import PropTypes from 'prop-types'
-import React, { useEffect } from 'react'
+import React, { useEffect, useRef, useState } from 'react'
```
shaunanoordin, line 3: [minor] I'm not sure where this branch is in regards to its base, but we've got an `import React` here that'll be reintroduced(??) after that standard was removed recently in #2288

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`
Affects: Drawing task, drawing tools, image navigation (pan/zoom)
Fixes #2268 

This PR is a refactor of the SVG system used to display Subject images and support Drawing Tasks.

- [major] The Drawing tasks/tools system no longer relies on the old SVGPoint, and now fully supports DOMPoint for creating (x,y) coordinates on our <svg>s
  - DOMPoint is the way to go! 👍 
  - A fallback has been added to ensure tests continue to work. (jsdom has no idea wtf a DOMPoint is.)
- Interaction Layer is now disconnected from the svg context. (But still works! See dev notes)
- Only the transform layer (of each Subject Viewer type) is passed around the svg context, as `canvas`

Code read looks good, tests look good! 👍 

### Dev notes

- The Interaction Layer no longer needs a connection to the svg context.
  - This is interesting! I thought you needed a handle on the transform layer - i.e. the previous svgContext.transformLayer.getScreentCTM - but turns out the Interaction Layer (DrawingCanvas) itself can detect the transforms applied by using DrawingCanvas.getScreenCTM()
- [minor][code style] There might be a very minor naming confusion between the `canvas` ref in InteractionLayer and svgContext.`canvas` used everywhere else.
  - The `canvas` in InteractionLayer refers to _the interaction layer itself_ which captures mouse/touch input, so that's 👍 
  - The `canvas` passed around via svgContext (on a per subjectViewerType responsibility basis) actually refers to a _transform layer, or <g#transform />_
  - No suggestions and no strong feelings, just an observation for now.

### Testing

Tested on localhost, using macOS 10.12.16 + Chrome 91 unless otherwise noted. Testing on `app-project`, not lib-classifier, to catch any project-page-specific gotchas. (There were none.)

**Scenario 1:** basic project (I Fancy Cats)
- URL: http://localhost:3000/projects/brooke/i-fancy-cats/classify/workflow/693?env=staging
- Basic project looks fine. I can load the home page, navigate to Classify, answer single/multiple answer tasks, submit Classification.
- I can pan using mouse (drag), zoom/rotate with UI buttons. 👍 
- I can create Point and Ellipse annotations (Drawing Task) 👍 
  - I can edit existing marks (drag to move Point + Ellipse, drag to resize Ellipse), and delete existing marks.

**Scenario 2:** precision coordinates + image rotation + rotating drawing marks:
- URL: http://localhost:3000/projects/darkeshard/prototype-2015/classify/workflow/18372?env=production
- I can create Rectangle, RotateRectangle, and Ellipse annotations (Drawing Task) 👍 
  - I can edit existing marks (drag to move all, drag to resize _Rectangle and Ellipse_), and delete existing marks.
- I can rotate the Subject image and perform the previous actions. 👍 
- All annotations are submitted with the correct coordinates (i.e. relative to the Subject image, not relative to the screen), regardless of Subject image rotation. 👍 

**Scenario 3:** grid tool/Subject Group Viewer
- URL: http://localhost:3000/projects/darkeshard/survos-testing-2020-subject-group-viewer/classify/workflow/3412?env=staging
- I can pan using mouse (drag), zoom/rotate with UI buttons. 👍 
- I can click to mark cells for annotation. 👍 
Additional tests:
- Scenario 2 looks good on Safari 14.

Known issues (not related to this PR):
- When the Subject image is rotated, dragging-to-pan the subject image doesn't work properly.
- RotateRectangle tool has weird drag-to-resize behaviour once rotated
- Safari 12 just borks with drawing tasks.

### Status

LGTM! 👍

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
eatyourgreens: I ran the React import codemod across the NextJS apps but I've left the libraries alone. This shouldn't make a difference at the moment.

PR #2281, opened 2021-06-30
[Security] Upgrade to Next 11.1 and React 17

eatyourgreens:
Upgrade both NextJS apps to [Next 11](https://nextjs.org/blog/next-11). Upgrade both apps to React 17. Upgrade peer dependencies for lib-classifier and lib-react-components.

Package:
app-content-pages
app-project
lib-classifier
lib-react-components

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2291, opened 2021-07-06
NextJS apps: fix linter errors

eatyourgreens:
#2281 identified some code changes which are build errors in Next 11. This PR ports those over without upgrading Next or React. Anonymous objects and functions are given names. Hooks are [run at the top of the main app function](https://reactjs.org/docs/hooks-rules.html), before any conditional blocks. An undefined `React` variable is fixed in `withRequireUser`. 

Package:
app-content-pages
app-project

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/app-project/src/shared/components/withRequireUser/withRequireUser.js
```diff
@@ -15,7 +15,7 @@ function storeMapper (stores) {
 
 function withRequireUser (WrappedComponent) {
   @inject(storeMapper)
-  class RequireUser extends React.Component {
+  class RequireUser extends Component {
```
eatyourgreens, line 18: `React` is undefined here, in `withRequireUser`, but there's a bug in NextJS, with React 16, which defines `window.React` as a global variable.

packages/app-project/pages/projects/[owner]/[project]/logger.js
```diff
@@ -38,7 +38,7 @@ class Logger extends Component {
           </Heading>
           <div>
             <p>
-              You're in development mode (<code>NODE_ENV !== 'production'</code>), so you're seeing this message. In production, this route will 404, so make sure you check the CloudFront config so it's picked up by PFE.
+              You are in development mode (<code>NODE_ENV !== &#39;production&#39;</code>), so you are seeing this message. In production, this route will 404, so make sure you check the CloudFront config so it is picked up by PFE.
```
Comment (srallen) [inline packages/app-project/pages/projects/[owner]/[project]/logger.js:41]: Does the linter not like contractions? 😆

packages/app-project/pages/_app.js
```diff
@@ -32,17 +32,17 @@ function useStore(initialState) {
 }
 
 export default function MyApp({ Component, pageProps }) {
-  try {
-    const { initialState } = pageProps
-    const store = useStore(initialState)
-    makeInspectable(store)
+  const { initialState } = pageProps
+  const store = useStore(initialState)
```
srallen, line 36: Should the `useStore` hook have a try/catch in case instantiating the store fails? I'm not sure if this whole block of code was wrapped with the try in case that failed. The `user.checkCurrent()` has its own error handling.

srallen (approved):
I'm not sure if we'll catch store instantiation errors anymore with this change, but otherwise looks good.

packages/app-project/pages/projects/[owner]/[project]/logger.js
Comment (eatyourgreens) [reply packages/app-project/pages/projects/[owner]/[project]/logger.js:41]: I thought about encoding the apostrophe but I have come across volunteers, who don’t speak English, who find contractions hard to understand.

packages/app-project/pages/_app.js
eatyourgreens: I’m not sure. We probably want the error logged in Sentry, but we probably want NextJS to generate an error page too. I might play around with this in a separate PR. 

Here, I think the hook inside a block was breaking the Rules of Hooks, but I’m not sure if it’s dangerous in the same way that conditional hooks are dangerous.

packages/app-project/pages/projects/[owner]/[project]/logger.js
Comment (eatyourgreens) [reply packages/app-project/pages/projects/[owner]/[project]/logger.js:41]: Also not sure why unencoded single quotes break the build in Next 11. That seemed a bit severe.

PR #2295, opened 2021-07-08
Disable previous transcription marks when there is an transcription mark unfinished and invalid.

### LINKED ISSUE #2292: Transcription task: Single-click annotations blocking movement through workflow
## Package
FEM - Transcription task

## Describe the bug
On projects that use the Transcription Task, it is possible to create an annotation with a single click (i.e. 1 instead of 2 clicks, which is what's required to trigger the transcription modal). These single-dot annotations are behaving like empty 2-dot annotations, and are blocking volunteers from being able to click "Next" and move on in the workflow. 

## To Reproduce
1. Go to the [Davy Notebooks Project classify page](https://www.zooniverse.org/projects/humphrydavy/davy-notebooks-project/classify/workflow/18759).
2. Click on the image to create a first annotation dot. 
3. Click anywhere outside the subject viewer. The dot will remain on the image and be green.
4. Interact with another line on the subject (e.g. click on an existing annotation and select a transcription from the dropdown menu). Once you have done this, the single dot will turn blue.

Davy Notebooks Project ID: `9006`
DNP workflow: `18759`

## Expected behavior
Volunteers should not be able to place a single dot and have it create an annotation. 

 - OS: macOS Catalina version 10.15.7
 - Browser: Chrome
 - Version: 91

## Additional context
Related to a previous issue where volunteers were able to create 'empty' annotations and that blocked them from moving forward. This is primarily happening when volunteers try to click on an existing annotation but don't quite hit it, and don't realize they've started a new annotation.

[Talk thread reporting this issue](https://www.zooniverse.org/projects/humphrydavy/davy-notebooks-project/talk/2586/2050706?comment=3357094&page=1).

srallen:

Package: lib-classifier

Closes #2292 and #2294 

The bug described has to do with transcription task when previous transcriptions are loaded from Caesar. These marks are still interactive when a volunteer has a partially made new mark with a single click. If a previous transcription is clicked, then the partial mark is "finished" in an invalid state. Here is a demonstration:

invalid-mark-bug

This branch prevents this from happening by disabling the previous transcription marks if there is a mark in an invalid state:

Screen Shot 2021-07-08 at 10 45 54 AM

Once the in progress mark is finished, the previous marks become interactive again.

This can be manually tested using: https://localhost:8080/?project=11300&env=production&demo=true

@snblickhan is this an ok solution?

Issue #2294 is fixed because the TypeError is thrown when a volunteer goes click on and attempt to actually finish the invalid mark. When you click on it, the argument passed into `onFinish` isn't actually an event object.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


goplayoutside3 (approved):
Looks good! Tests are passing locally, storybook looks correct, and I was able to see the new disabled transcribed line behavior 👍

snblickhan:
@srallen Thanks -- this seems like a good solution to me, but I worry that it will be confusing to volunteers without some sort of dialogue box. The current blocker is that, when this happens, the "Next" button is disabled. I'm worried that now we'll just get the same questions from volunteers as to why they can't engage with other markings.

Is there a way we could add a pop-up if someone tries to click into another annotation while they have one in progress? Happy to chat more if you think this is a bad idea.

srallen:
@snblickhan I agree that there needs to be some kind of UI that refocuses the unfinished mark. I don't think I want to roll development of something like that into the scope of this particular PR as this is focused specifically on preventing the bad data with creation of an invalid mark where the x, y points of both ends are equal to each other and has no transcription (and that's been the source of aggregation errors). 

I'm imagining maybe some kind of notice or tooltip associated with the disabled state of the task area navigation one can select or click that would then refocus the unfinished work.

Issue #2249, opened 2021-06-17
Limit workflow list when assignment is being used


srallen:
## Describe the solution you'd like

Depending on the current assignment in the user's UPP, the workflow listing on the homepage and on the selector modal needs to disable access to certain workflows based on the level label the workflow is given. 

## Additional context

This is to reach parity with Gravity Spy's feature set.

PR #2304, opened 2021-07-13
Refactor selector stories

srallen:

Package: app-project storybook

Toward #2249

I've refactored the stories to clarify where they live in the code, several renamed to be under the "Shared" section. I removed one redundant `WorkflowMenu` story that was exactly the same as the `WorkflowSelector` story in the shared section (just adds maintenance imo) and renamed the `WorkflowMenu` that exists for the classify page to `WorkflowMenuModal` to clarify that it indeed has slightly different function that the `WorkflowMenu` on the home page.

This will make it a bit easier to use the stories for new development and for maintenance. @goplayoutside3, this refactors the stories I mistakenly broke in a previous PR review you did. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowMenu/WorkflowMenu.stories.js
```diff
@@ -1,85 +0,0 @@
...
-    </Grommet>
-  )
-}
-
-export default {
-  title: 'Project App / Screens / Project Home / Workflow Menu',
-  component: WorkflowMenu,
-  args: {
-    dark: false,
-    workflows: [ ...WORKFLOWS, mockGroupedWorkflow ]
```
eatyourgreens, line 75: `mockedGroupWorkflow` here adds set selection to the story. That should be moved to the Hero component, if this story is being deleted. Otherwise, we don't have a story for subject set selection on the project home page.
https://zooniverse.github.io/front-end-monorepo/@zooniverse/fe-project/index.html?path=/story/project-app-screens-project-home-workflow-menu--default

goplayoutside3 (approved):
Looks good to me!

I don't think I know enough about when grouped workflows show up in the `Hero`'s `WorkflowMenu` to agree/disagree with @eatyourgreens's comment, but the rest of the stories function as expected 👍

eatyourgreens:
The Workflow Menu story has four workflows, and the last one opens a subject set menu.
https://zooniverse.github.io/front-end-monorepo/@zooniverse/fe-project/index.html?path=/story/project-app-screens-project-home-workflow-menu--default

The Hero story only has three workflows, so I think that fourth workflow needs to be added to the Hero story. I can't remember why I didn't just do that in the first place.
https://zooniverse.github.io/front-end-monorepo/@zooniverse/fe-project/index.html?path=/story/project-app-screens-project-home-hero--default

packages/app-project/src/screens/ProjectHomePage/components/Hero/components/WorkflowMenu/WorkflowMenu.stories.js
srallen: I restored this story, but renamed it to make it clear what it was for. I still didn't catch this when I did my side-by-side comparison previously.

eatyourgreens: It's really confusing that we have two stories for the same thing. You're right about consolidating them in one story.

PR #2307, opened 2021-07-15
Use as prop to render a span when href is defined and disabled is true

srallen:

Package: lib-react-components

Toward #2249.

Grommet's `Button` allows us to specify an `href` which will then render an anchor tag that is styled as a button. This is convenient and is definitely used in our code base, however, all other `Button` props are still accepted including `disabled`. This is a problem because links cannot be disabled via attribute. What happens is with this combination you have a link styled like a disabled button, but is still functional. There is a method to pseudo-disable a link, but it's better to just render text rather than do that. 

At first, I thought I'd create a new component for our shared components library that was a span styled like a button, but remembering Grommet's `as` prop that all their components support, we can more succinctly do that in the button components themselves. 

This adds a conditional where if `href` is defined and `disabled` is true to render a span instead of an anchor for the `PrimaryButton` and `PlainButton` components in the shared component library. I've destructured `href` out of the props to pass along for `CloseButton` so this shouldn't be possible for that one. The stories and tests are updated as well.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-react-components/CHANGELOG.md
+++ b/packages/lib-react-components/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Changed `LinkList` component to use `fit-content` CSS styling to render correct height in Safari.
 - Modified matchRegex in `ping` of `Markdownz` to avoid creating username-links in typed email addresses.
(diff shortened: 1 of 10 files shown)


goplayoutside3 (approved):
This looks good! I have just one suggestion that's non-blocking:

I noticed in storybook that a disabled CloseButton shows `cursor: not-allowed` styling, but PlainButton does not affect the cursor when disabled. Is adding `not-allowed` cursor styling easily done for the other disabled buttons?

PR #2230, opened 2021-06-04
Remove dev dependencies and src from Docker builds

eatyourgreens:
Add a second stage to the Docker build. The second stage takes the packages built by the first stage and builds a new image from them. The new image has a production install of `node_modules` (no dev dependencies) and deletes source code from the built packages. Removing dev dependencies shaves ~500MB off the final image size.

The build is loosely based on [the NextJS Docker example](https://nextjs.org/docs/deployment#docker-image), adapted for our monorepo.

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2309, opened 2021-07-16
Docker build and compose improvements

camallen:

This PR attempts to improves the docker build process for development / persistent CI machines by adding in a docker build cache volume for the yarn install steps. 

by using a docker build cache - a local volume for caching the yarn install artifacts - the cached data is not persisted in the resulting image but is available on the build system (VM / dev machine etc) for re-use across builds. Thus the build system doesn't have to rebuild the yarn system each build - in theory this speeds up these steps across the docker builds. Note this doesn't help for ephemeral VM systems like GH actions but is useful for system like out jenkins CI and our development machines

https://github.com/moby/buildkit/blob/2a4577efabed1b4404e2884ef56873b8c0a42e95/frontend/dockerfile/docs/syntax.md#run---mounttypecache
https://vsupalov.com/docker-buildkit-features/#using-a-cache-mount

This PR also 
- ensure the shell service boots properly and has the correct dev dependencies installed. It does this by 
    - targeting the `builder` image
    - ensuring the `lerna.json` config is available in the image
    - using the `CMD` syntax to boot the shell, e.g. `yarn test` now works in the shell
- adds a node_modules volume cache via compose for the shell service. This is useful as on the first use it copies the node_modules from the underlying image to the docker volume for re-use. When starting the shell service will have the node_modules persisted which means any work on the shell console will reuse these, e.g. `yarn install / upgrade` etc will persist state between shell service boots and thus save install times.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


Dockerfile
```diff
@@ -108,7 +103,7 @@ ADD .yarnrc /usr/src/
 
 COPY --from=builder /usr/src/packages ./packages
 
-RUN yarn install --production --frozen-lockfile --ignore-scripts --prefer-offline
+RUN --mount=type=cache,uid=1000,gid=1000,target=/home/node/.yarn YARN_CACHE_FOLDER=/home/node/.yarn yarn install --production --frozen-lockfile --ignore-scripts --prefer-offline
```
eatyourgreens, line 106: Won't this re-use the cache from the previous step, which we don't want to do because we don't want the dev dependencies? See https://github.com/yarnpkg/yarn/issues/6373#issuecomment-771261476

I mentioned in #2230 that we have to do a clean install on the second step because of this bug. If we're building from the cache, then this entire second stage of the build can be removed completely ie. revert #2230.

eatyourgreens (commented):
We should double check, but I don't think you can re-use the yarn cache for a production build, because of a bug in yarn v1.

That bug is fixed in yarn 2, so we could maybe update the build scripts to use yarn v2.

eatyourgreens:
`docker compose` changes look good to me, but Icaching dev dependencies to include in the final production image will put us back where we were before #2230, which I think almost doubled the image size?

Dockerfile
camallen: it'll have the cache artifacts from the `builder` stage available for this yarn cmd `yarn install --production --frozen-lockfile --ignore-scripts --prefer-offline` via the `--mount=type=cache` so it can re-use any packages in the cache and avoid redownloading. 

However it won't persist the cache data in the underlying image, rather it mounts the cache volume at build time for data re-use.

However based on the yarn bug you outline above then this will be a problem. so it's probably best to remove this 2nd re-use of the cache volume and install afresh.

camallen:
```suggestion
RUN yarn install --production --frozen-lockfile --ignore-scripts --prefer-offline
```

camallen:
> `docker compose` changes look good to me, but caching dev dependencies to include in the final production image will put us back where we were before #2230, which I think almost doubled the image size?

dev dependencies aren't in the final images. 

> `docker compose` changes look good to me, but caching dev dependencies to include in the final production image will put us back where we were before #2230, which I think almost doubled the image size?

I'll rebuild the changes locally and report the underlying volume sizes from my local machine.

camallen:
> We should double check, but I don't think you can re-use the yarn cache for a production build, because of a bug in yarn v1.
> 
> That bug is fixed in yarn 2, so we could maybe update the build scripts to use yarn v2.

that's interesting - and possibly a good solution long term.

Dockerfile
eatyourgreens: Yeah, that change should do it. My understanding of https://github.com/yarnpkg/yarn/issues/6373 is that `yarn install --production` installs everything from the yarn cache, even dev dependencies.

eatyourgreens:
I've run a local build from this branch. The `dev` image is smaller than the `shell` image, so maybe that `yarn` bug is a red herring and dev dependencies aren't being added to `node_modules`.

```
odonnell$ docker images
REPOSITORY                         TAG              IMAGE ID       CREATED          SIZE
front-end-monorepo_dev             latest           839f19300b00   2 minutes ago    927MB
front-end-monorepo_shell           latest           22b4acbbf6ab   12 minutes ago   1.27GB
```

eatyourgreens:
However, when I run `du-sh` inside each image, I see `node_modules` has the same size, so I'm still a bit confused as to what this PR is doing.

`shell` image:
```
/usr/src $ du -sh *
4.0K	lerna.json
786.4M	node_modules
4.0K	package.json
599.3M	packages
840.0K	yarn.lock
```

`dev` image:
```
/usr/src $ du -sh *
786.4M	node_modules
4.0K	package.json
546.2M	packages
840.0K	yarn.lock
```

I'd expect the final `dev` images to have a much smaller `node_modules` install, without the dev dependencies.

eatyourgreens:
>that's interesting - and possibly a good solution long term.

If it wasn't for that `yarn` bug, I think we could add this to the build to remove dev dependencies from `node_modules` for the final production deploy.
```sh
# install all dependencies and build everything before this step
rm -rf node_modules
yarn install --production --frozen-lockfile --ignore-scripts --prefer-offline
```

camallen:
> so I'm still a bit confused as to what this PR is doing

It's an experimental use of new docker buildkit features that tries to optimize the build times for the yarn package installs for dev (and as a consequence CI) machines when the docker image layer becomes invalid. I've found the build times for this app incredibly slow and thus the development cycle can slow down.

eatyourgreens:
Oh, I hear that. The build I ran this morning took ~1 hour. Even on Jenkins, I think the build times are very long (around 20 minutes maybe?)

eatyourgreens:
Would it be possible to use separate caches for the `builder` and `runner` images, so that we can cache both yarn installs across builds?  I'm not sure if the caches can be named, but thought I'd ask.

camallen:
> Would it be possible to use separate caches for the `builder` and `runner` images, so that we can cache both yarn installs across builds? I'm not sure if the caches can be named, but thought I'd ask.

yep - i think that's viable through the `id=` of the mount cache directive, https://github.com/moby/buildkit/blob/2a4577efabed1b4404e2884ef56873b8c0a42e95/frontend/dockerfile/docs/syntax.md#run---mounttypecache

eatyourgreens:
Cool. Then I think the changes here will work if the two caches have different IDs. Otherwise we'll see the problem where `yarn install --production` ignores the production flag and installs everything that's saved in the yarn cache.

eatyourgreens:
FWIW I see these sizes locally for `node_modules`:
- development build: 738M
- production build: 277M

so the production build should save us a lot in image size.

camallen:
> FWIW I see these sizes locally for `node_modules`:
> 
>     * development build: 738M
> 
>     * production build: 277M
> 
> 
> so the production build should save us a lot in image size.

I've just rebuilt locally with split caches via abf014f508b5eeaffe86808d3b3a95f4ab03bd2b and i see the following `node_modules` sizes now which matches the expected behaviour of dev-dependencies removal
```
$ docker-compose run --rm shell
...
/usr/src $ du -sh *
4.0K	lerna.json
786.3M	node_modules
4.0K	package.json
598.4M	packages
840.0K	yarn.lock
/usr/src $ exit

$ docker-compose run --rm fe-project
...
/usr/src # du -sh *
290.5M	node_modules
4.0K	package.json
545.4M	packages
840.0K	yarn.lock
/usr/src # 
```

eatyourgreens (approved):
LGTM. Let’s see how this performs.

PR #2311, opened 2021-07-20
Add react-testing-library

srallen:

Package: several

Related to #2301 

This just installs react-testing-library as a dev dependency so we can start experimenting with it. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-content-pages/package.json
+++ b/packages/app-content-pages/package.json
@@ -69,6 +69,7 @@
     "jsdom": "~16.5.2",
     "mocha": "~8.3.2",
     "proxyquire": "~2.1.0",
+    "react-testing-library": "~8.0.1",
     "sinon": "~10.0.0",
     "sinon-chai": "~3.6.0"
   },
--- a/packages/app-project/package.json
+++ b/packages/app-project/package.json
@@ -82,6 +82,7 @@
     "jsdom": "~16.5.2",
     "mocha": "~8.3.2",
     "nock": "~13.0.11",
+    "react-testing-library": "~8.0.1",
     "sinon": "~10.0.0",
     "sinon-chai": "~3.6.0"
   },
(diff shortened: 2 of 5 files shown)


shaunanoordin (approved):
## PR Review?

This PR adds the `react-testing-library`, with an intent to add code that _actually uses_ the new dependency _at a later date._ I have no specific issues with experimenting with a new library, nor having this PR be the precursor step to the experiments.

I've read through the [discussion](https://github.com/zooniverse/front-end-monorepo/discussions/2301) and the gist is:
- We're going to upgrade to React 17 one way or another.
- Enzyme won't work with React 17 any time soon, so we need to look into alternatives.
- react-testing-library is on the top of the list.

Sounds legit to me.

Misc notes:
- I'm assuming we're separating the _(1) "add testing library to project"_ and _(2) "add experimental testing code that uses this library"_ steps to make step 2 easier to develop? So when we jump in/out of step 2's branch, we wouldn't need to _yarn bootstrap_ too much? (If my assumption is wrong, then I'm mildly curious to know why we didn't just combine the two steps.)  
- The only practical downside I can imagine is that by adding this dependency, our build processes will install the 12KB library without actually utilising it, but this extremely minor bit of waste is fine.

Anyway, thumbs up emoji! 👍

srallen:
> (2) "add experimental testing code that uses this library" steps to make step 2 easier to develop? So when we jump in/out of step 2's branch, we wouldn't need to yarn bootstrap too much? (If my assumption is wrong, then I'm mildly curious to know why we didn't just combine the two steps.)

Good question. The reason that this is just setup is that I'm trying to organize a few days where several of us can be working on testing at the same time so we can all be learning and tackling this together all at once. I hope to do this very soon (within the next 2-3 weeks). Several people should be already trying it out on their own at the moment and it's possible we'll have a PR or two prior to an organized group effort.

PR #2315, opened 2021-07-22
Add hover styling to survey task FilterButton

mcbouslog:
Package:
- lib-classifier

Describe your changes:
- adds hover styling to FilterButton

Staging:

Note:
- unable to add FilterButton focus styling, I think bug with Grommet RadioGroupButton with children (https://storybook.grommet.io/?path=/story/input-radiobuttongroup-children--children) compared to focus styling working as expected with simple RadioButtonGroup (https://storybook.grommet.io/?path=/story/input-radiobuttongroup-simple--simple), I asked about in Grommet Slack
- Choices and ChoiceButton steal focus on filtering [here](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/plugins/tasks/SurveyTask/components/components/Chooser/components/Choices/Choices.js#L32-L36), would need larger refactor to address this issue, like track FilterStatus DropButton open state and only update Choices focus when DropButton not open, though likely other ways too
- per above issues, if do larger refactor, might also want to consider refactoring from FilterStatus with DropButton to moveable modal? mentioned in prev conversations, I think we're going to get alpha/beta feedback and go from there, just noting so I don't forget =) 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


srallen (approved):
I think the lack of the focus state being passed down to the children for styling purposes is a bug. I checked out the source and found this line to be the problem: https://github.com/grommet/grommet/blob/b001ef3a3ef668edf2c71dd32258408f9dd1f793/src/js/components/RadioButton/RadioButton.js#L85

I patched the local installed version in node_modules to pass along the focus state and then used to to style successfully. I'll open an issue / PR with Grommet and ideally they'll accept and publish with next release. If they don't, then we'll have to refactor to not use the `RadioButtonGroup`. Could you open an issue here to track it on our end?

srallen:
Opened https://github.com/grommet/grommet/pull/5473

PR #2318, opened 2021-07-22
Added videoViewer ref back to svg element

ErikOstlund:
Package: `lib-classifier`

## Describe your changes:
During the refactor from #2277, the `videoViewer ref` in `SingleVideoViewerContainer.js` was removed.
With the `videoViewer ref` removed, videos were not loading properly and users would see the 'Something went wrong' message.

The `videoViewer ref` is still needed in this file to calculate the video dimensions.
With this `ref` now in place, video dimensions are properly being calculated and loaded!

All unit tests are passing and UI is behaving as expected.

## Testing
In the front-end-monorepo folder, run yarn bootstrap to build the packages. Then cd packages/lib-classifier and run yarn dev.
Project URL: [https://localhost:8080/?env=production&project=13789&workflow=16843](https://localhost:8080/?env=production&project=13789&workflow=16843)

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleVideoViewer/SingleVideoViewerContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleVideoViewer/SingleVideoViewerContainer.js
@@ -27,6 +27,7 @@ class SingleVideoViewerContainer extends React.Component {
   constructor() {
     super()
 
+    this.videoViewer = React.createRef()
     this.transformLayer = React.createRef()
 
     this.state = {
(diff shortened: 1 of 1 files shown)


shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This fixes an issue where Video subjects wouldn't display properly (i.e. the Video Viewer would crash with the error, "can't find .current of undefined") by adding a missing ref.

Code & tests are looking good 👍 

### Testing

Tested on localhost with macOS 10.12.6 + Chrome 91

- Project: [Erik's video project](https://localhost:8080/?env=production&project=13789&workflow=16843&demo=true)
  - Any project with video Subjects should work too.
- Open the Classify page.
- Expected: the video should render and should be playable.
- That's it.

### Dev Notes (Optional Details)

In case anybody's curious, here are some additional details of why the SingleVideoViewer was accidentally bricked. (Sorry about that!)

- The breaking change was introduced in #2277 , which simplified our SVG refs.
- Prior to that PR, there were two references to SVG elements that were passed around via the SVG Context. `<SVGContext.Provider value={{ svg, getScreenCTM }}>` essentially passed around refs to the SVG "container" (to calculate width/height of the viewable element) and the SVG "transform layer" (to compensate for image rotation etc on draggables, etc) to lower-ordered components.
- After that PR, the ref-passing was simplified to `<SVGContext.Provider value={{ canvas }}>`, which meant only the "transform layer" was passed around via context.
- As a result, we removed the refs to the SVG "container" since it didn't need to be passed around to lower-ordered components any more...
- ...but we overlooked the fact that the SingleVideoViewerContainer actually still kinda needed that ref for "getVideoSize()". Derp! 🤦 

Anyway, the fix is pretty straightforward and everything works now!

### Status

LGTM 👍

PR #2319, opened 2021-07-23
Add ADR-33 for survey task feedback

mcbouslog:
Package:
- lib-classifier

Describe your changes:
- adds ADR 33 Survey Task Feedback

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


docs/arch/adr-33.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Decision
+
+To implement a survey task feedback strategy within [the existing feedback framework](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/store/feedback/strategies). The initial survey task annotation feedback will presume the annotation has one annotation value per survey task choice. The subject metadata will include:
+
+- #feedback_N_id (required) - ID of the corresponding workflow task rule.
+- #feedback_N_choice_labels (required) - comma separated target choice label(s) (not values, i.e. "Blip" not "BLP").
+- #feedback_N_choice_strict (optional) - defaults to true, target label(s) must match annotation label(s) exactly for success. If false then any target label(s) match with annotation label(s) will be a success.
```
mcbouslog, line 17: I'm not sure about this, curious what others think. Can leave for subsequent iteration and just keep strict match initially.

lcjohnso:
Hi @mcbouslog -- Two questions:
1. Is there a specific reason for having the `#feedback_N_choice_labels` be the user-provided label (e.g., "Blip", "None of the Above") instead of the processed choice (e.g., "BLIP", "NONEOFTHEABOVE")?  I could see this being confusing, as Caesar uses the processed choice names for defining rules.  Also, the possibility of special characters in the label and the assumption that the `en`-version of the label is the one that should be provided are two extra complications, too.
2. I started out wondering whether `single-answer-question` and this survey strategy could be combined / generalized.  After some though, I think the answer is **no** due to index vs. text checking (and this survey strategy's possibility of allowing/requiring multiple choices -- which I like).  But: are their any other task types that might use the same strategy, or are we keeping the focus very strict?

docs/arch/adr-33.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Decision
+
+To implement a survey task feedback strategy within [the existing feedback framework](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/store/feedback/strategies). The initial survey task annotation feedback will presume the annotation has one annotation value per survey task choice. The subject metadata will include:
+
+- #feedback_N_id (required) - ID of the corresponding workflow task rule.
+- #feedback_N_choice_labels (required) - comma separated target choice label(s) (not values, i.e. "Blip" not "BLP").
```
mcbouslog, line 16: @lcjohnso great question:
> 1. Is there a specific reason for having the `#feedback_N_choice_labels` be the user-provided label (e.g., "Blip", "None of the Above") instead of the processed choice (e.g., "BLIP", "NONEOFTHEABOVE")?  I could see this being confusing, as Caesar uses the processed choice names for defining rules.  Also, the possibility of special characters in the label and the assumption that the `en`-version of the label is the one that should be provided are two extra complications, too.

I didn't realize Caesar was using the processed choice ID. I don't have a strong opinion either way. 

I've initially noted user provided labels because I thought it would be straightforward from the labels noted in the CSVs, whereas the processed ID takes checking the Raw task data or other steps. 

Any inconsistency between configurations is certain to cause issues though.

If users are already comfortable using the choice processed ID then I lean towards that for consistency, but I'm curious what others think?

mcbouslog: Noting for reference here's the PFE survey task editor code that converts user provided labels to the processed choice ID - https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/survey/editor.cjsx#L382

mcbouslog:
@lcjohnso -

> 2. I started out wondering whether `single-answer-question` and this survey strategy could be combined / generalized.  After some though, I think the answer is **no** due to index vs. text checking (and this survey strategy's possibility of allowing/requiring multiple choices -- which I like).  But: are their any other task types that might use the same strategy, or are we keeping the focus very strict?

You're correct the index vs text(choice ID/label) checking makes the success/failure determination different for the single choice and survey task, but many of the [components](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/components/Classifier/components/Feedback) and [strategy helpers](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/store/feedback/helpers) will indeed be reused. 

I'm not sure about other task types, though maybe dropdown would be similar to this survey task approach? Certainly worth looking into! But I think initially good to keep the scope strict, get working for a specific use case (Gravity Spy survey task), then take a step back and assess if it can be generalized to other tasks, imho.

docs/arch/adr-33.md
srallen: I don't know how we would do a fuzzy match without an algorithm. Did you have something in mind? This implies we have the capability to do this (or will have the capability), and if we don't, then I don't think we should give people the option.

srallen:
I don't think we can use labels for the comparison because of translations. Same reason we don't use labels on single choice task. Using the index from the list is unreliable because the survey task allows users to filter the options, so unless we kept track of the original indexed position from the full array of options as a unique id, then I don't think we can rely on this either. (Note: single choice task is vulnerable to mismatch since it also uses indexes if someone edits the workflow midway after launch and rearranges the answers, but typically we strongly recommend against this to project builders and assume they don't.)

So, I think we'll have to use the expected annotation, which is the processed choice, as an array of strings. We can use `toLowerCase` to exclude the case differences when doing the comparison. I think case is used more for human readers to be able to quickly identify what string is intended for the machine vs the UI label. 

Thoughts?

srallen:
Another thing to consider for generalized feedback for survey is if we'll be comparing and processing any of the questions for the choice selection. For Gravity Spy we don't because they don't ask additional questions, but most survey task projects do. So the comparison may have to be more complicated as an array of objects with the choice selection as the key to an object with the object containing the questions and answers?

srallen:
One way we can get around the potential complexity with the choice questions is to have two feedback strategies: a simple that just compares an array of strings and a complex that compares an array of objects?

docs/arch/adr-33.md
mcbouslog: This would be if the subject metadata feedback says the answer is 'BLIP' and the user has answered 'BLIP' and 'KOIFISH', or vice versa (feedback says two choices, user has answered with one) is the classification a success or failure. I think if strict is true [this line stays as is](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/gs-gold-standard-summary.jsx#L18), if strict is false it would use `Array.some`, I think, though likely oversimplifying in my mind. I think that's possible without an algorithm or additional dependency, but could be misunderstanding?

For Gravity Spy it'd be strict = true, but maybe camera trap projects would want strict = false? This can 💯  be a subsequent iteration if it's even something camera trap researchers want.

mcbouslog:
Apologies, I think using "labels" is misleading. For the subject metadata field that's used to compare expected vs actual, I'd propose `feedback_choice_names` instead. The alternative we've referred to as the processed choice ID could be `feedback_choice_ids` (open to suggestion on these names, just hopefully helpful defining two options for discussion).

`feedback_choice_names` would be the name per the CSV the researcher has created. For feedback we'd process the name the same as we do when the CSV is imported, then compare it to the `classification.annotation.value[0].choice` ([link to survey annotation store](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/plugins/tasks/SurveyTask/models/SurveyAnnotation.js#L6)).

The CSV choice "name" is used to create a choice's ID and human readable label (which can then be translated). Using the name for feedback wouldn't be an issue with translations, but it could still be an issue if there are special characters and would be an issue of consistency with Caesar. It would be easier for a researcher to take the name they've entered in the CSV and use it for feedback, but edits can be made to the project builder survey task editor to make the processed choice ID clear.

A lot of writing to say I still lean towards using the processed choice ID, but hopefully I've better explained why using the choice "name" is an option to consider.

docs/arch/adr-33.md
srallen: I misunderstood what this meant. I thought it was referring to the string comparisons itself which would require fuzzy string matching (basically an algorithm). Array.every vs Array.some is ok though. We just need to make sure this is clearly documented about the array or expected list. Thanks!

lcjohnso:
Thanks for this info and discussion.  I'm not feeling strongly about the adopted plan moving forward -- I see pros and cons for either `names` or `ids`.  I'm also fine having a larger number of more specific feedback strategies -- such as sticking to a survey task specific strategy here -- as long as it is documented how they should be used, limitations, etc.

docs/arch/adr-33.md
srallen: Could we add to the docs here that true means array.every and false means array.some?

docs/arch/adr-33.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Context
+
+We plan to implement feedback for the survey task.
+
+Currently in PFE, the only feedback related to the survey task is for the Gravity Spy project, and is referred to as "Gravity Spy Gold Standard". Gravity Spy Gold Standard is a PFE experimental feature (per workflow configuration object) that checks a subject's metadata for `'#Type' === 'Gold'` and provides feedback after classification completion with the [`gs-gold-standard-summary`](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/gs-gold-standard-summary.jsx). Survey task annotation values are compared to subject metadata `'#Label'` values. A message is then shown accordingly (i.e. success, failure, or special message per subject metadata `'#post_classification_feedback'`).
+
+## Decision
+
+To implement a survey task feedback strategy within [the existing feedback framework](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/store/feedback/strategies). The initial survey task annotation feedback will presume the annotation has one annotation value per survey task choice. The subject metadata will include:
```
srallen, line 13: Let's have this be a simple feedback strategy for just comparing the string choices and make a note in the consequences section that there may be a request in the future for a more complex one that compares the choice questions too? This strategy will not implement this.

docs/arch/adr-33.md
mcbouslog: Apologies for the noise around this, but in a similar mindset that the initial focus is a simple survey task feedback strategy, and subsequently we can create more complex survey task feedback strategies that include question answers, I think this "strict" attribute should be dropped, and only exact matches should be a success?

Then based on input from researchers if it's something desired we can revisit, it wouldn't even have to be a new strategy type, but could be an iteration on the simple survey task feedback strategy if the default is going to be a strict comparison. Or it could be part of a new complex type. Important point I think is that it's not necessary initially?

I'll update accordingly, but happy to still include if others feel strongly it should be included.

srallen: Sounds good to me. This could be in a future iteration if asked for.

lcjohnso (approved):
LGTM!

srallen (approved):

PR #2321, opened 2021-07-26
Disable workflows

### LINKED ISSUE #2249: Limit workflow list when assignment is being used

srallen:

Package: lib-react-components, app-project

Closes #2249

This refactors the workflow selector to display a disabled looking button-like span for workflows the user does not have access to yet when a project is using workflow assignment. 

How this looks now on PFE:

Screen Shot 2021-07-26 at 10 45 32 AM

How it looks implemented on this branch as seen with https://local.zooniverse.org:3000/projects/srallen086/workflow-assignment-testing:

Screen Shot 2021-07-26 at 10 45 42 AM

We should see the same render in the modal for selection when navigating to the classify page from the navbar without a workflow chosen:

Screen Shot 2021-07-26 at 10 49 24 AM

And projects that do not use workflow assignment should see the preexisting styling:

Screen Shot 2021-07-26 at 10 47 48 AM

To prevent a flash of a rendered button links list without the unlocked and locked filtering, we now have to check if the user project preferences have finished loading.

As a technical side note, all of the existing tests continue to use enzyme for now, but the new tests for `WorkflowAssignmentButtons` are using react testing library. 

This is best tested with https://local.zooniverse.org:3000/projects/srallen086/workflow-assignment-testing

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-react-components/src/PlainButton/PlainButton.js
```diff
@@ -28,7 +28,7 @@ function PlainButton (props) {
     color,
     ...rest
   } = props
-  const renderAs = as || href && disabled && 'span'
+  const renderAs = (href && disabled) ? 'span' : as
```
srallen, line 31: I had to change how this conditional works because previously the falsey condition returned false which is an invalid prop for `as`. Now the default prop of an empty string is used if as isn't defined or if href and disabled isn't defined.

packages/app-project/src/shared/components/WorkflowSelector/components/WorkflowSelectButtons/WorkflowSelectButtons.spec.js
```diff
@@ -0,0 +1,57 @@
...
+      }, {
+        completeness: 0.9,
+        configuration: {
+          level: 3
+        },
+        displayName: 'workflow 3',
+        id: '3'
+      }
+    ]
+    it('should only render links for unlocked workflows', function () {
```
mcbouslog, line 49: Maybe this could be two different tests (same code), like `render links for unlocked workflows` are first two assertions and `render, but not as links, other workflows` are the second two? I'm not sure that's really an improvement 🤷‍♂️

packages/app-project/src/shared/components/WorkflowSelector/components/WorkflowSelectButtons/WorkflowSelectButtons.js
```diff
@@ -0,0 +1,106 @@
...
+      as="ul"
+      fill='horizontal'
+      gap='xsmall'
+      margin={{ top: 'small' }}
+      pad='none'
+      style={{ listStyle: 'none' }}
+      width={{ max: 'medium' }}
+    >{workflows.map(workflow => (
+        <li key={workflow.id}>
+          <WorkflowSelectButton onSelect={onSelect} workflow={workflow} />
```
mcbouslog, line 95: if the `project.experimental_tools.includes('workflow assignment')` I think all but the first button should include `disabled`, for a first time or very few classifier who doesn't have a workflow assignment yet?

mcbouslog (dismissed):
This looks good, no blocking comments, maybe a little story cleanup, optional if the following is for this PR or subsequent. Tested on other projects, with and without workflow selection, various browsers.

~In PFE if logged out, a project with the workflow assignment experimental feature won't show workflow buttons, but just Get Started [per these lines](https://github.com/zooniverse/Panoptes-Front-End/blob/45399f3267f45fa1d67d5bcd162b4050ed11e5ac/app/pages/project/home/index.jsx#L105-L114). I don't think that's been added to the FEM repo yet, which can be subsequent PR, if even want to replicate PFE exactly (maybe we don't, maybe showing all workflows and only enabling first level workflow button is preferable for logged out user).~ using new design (show workflow buttons) in favor of PFE 

If logged in, but never classified on, therefore no workflow assignment per user project preference, PFE does show workflow buttons, but disables all except the first level workflow button. I think with this PR if a user hasn't or doesn't have a workflow assignment, it shows all the workflow buttons (no unlocked/locked sections) enabled, when I think we'll want to disable all but the first level workflow?

packages/app-project/src/shared/components/WorkflowSelector/components/WorkflowSelectButtons/WorkflowSelectButtons.js
mcbouslog: Wait this is not the right place for possible editing, it'd be in the first conditional block. Closing/resolving this thread, will post where noted...

packages/app-project/src/shared/components/WorkflowSelector/components/WorkflowSelectButtons/WorkflowSelectButtons.js
```diff
@@ -0,0 +1,106 @@
...
+import { SpacedText } from '@zooniverse/react-components'
+import WorkflowSelectButton from '../WorkflowSelectButton'
+import en from './locales/en'
+import counterpart from 'counterpart'
+import PropTypes from 'prop-types'
+
+counterpart.registerTranslations('en', en)
+
+export default function WorkflowSelectButtons ({ assignedWorkflowID = '', onSelect, workflows = [] }) {
+  let assignedWorkflow
```
mcbouslog, line 11: I think a new user, or user without workflow ID assigned, on a project with the experimental feature "workflow assignment" we still want in this first block where workflow buttons are shown in unlocked and locked sections (with only first level workflow unlocked), but currently that scenario is going to the second block and showing all workflow buttons enabled.

srallen:
> In PFE if logged out, a project with the workflow assignment experimental feature won't show workflow buttons, but just Get Started per these lines. I don't think that's been added to the FEM repo yet, which can be subsequent PR, if even want to replicate PFE exactly (maybe we don't, maybe showing all workflows and only enabling first level workflow button is preferable for logged out user).

I don't think this UI will ever come back for projects. Becky has all her designs of the project homepage always showing workflow button-styled links.

srallen:
@mcbouslog you are correct that for signed out users, the UI should still show a filtered button link list. So I've updated it so that if you're signed out or if you're new to the project, only the first level workflow is available:

Screen Shot 2021-07-28 at 1 40 21 PM

Stories are fixed as well.

Note that since I refactored the button links to consistently use the `PrimaryButton` (what was existing was prior to the styles being abstracted out to the reusable component) that with dark theme, the buttons appear as the border outlined version. Becky never provided dark theme styling for this area of the site nor consistent guidance generally on dark theme for buttons. I think I'm going to separately in a new PR update the theme to leverage Grommet's newer [button theme API](https://v2.grommet.io/button) and guidance. Buttons that are primary will retain the color fill. Buttons that are secondary will switch to a border outline and fill on hover/focus. Default buttons will just be the standard grommet style border outline and the width of the outline changes on hover/focus. In the classifier area, when switching to dark theme, the buttons will switch from primary to secondary. Other places like for the call to action in modals, those kinds of buttons will stay primary.

packages/app-project/src/shared/components/WorkflowSelector/WorkflowSelector.stories.js
```diff
@@ -84,6 +93,21 @@ export function Default({ dark }) {
...
+  )
+}
+
+export function WithLevels({ dark }) {
+  return (
+    <StoryContext theme={{ ...zooTheme, dark }}>
+      <WorkflowSelector
+        assignedWorkflowID='7890'
+        onSelect={onSelect}
+        uppLoaded
```
mcbouslog, line 110:
```suggestion
        uppLoaded
        workflowAssignmentEnabled
```

mcbouslog (approved):
Updates look great, all users (logged out, new, promoted) have workflow buttons as expected. Thank you for updating tests and stories as well, very helpful.

srallen:
I'm going to revert the refactor on the button swap from the custom themed Grommet Button to PrimaryButton for now, just so we can keep consistency in the current styling between light & dark themes. The theme update I've been working on locally has ended up being more complicated that anticipated, so I won't have an immediate follow up ready.

PR #2324, opened 2021-07-29
Talk in Classify (aka QuickTalk)

shaunanoordin:
## PR Overview

Package: `lib-classifier`

This is an experimental PR that adds the Quick Talk component.

- The Quick Talk component is intended to promote Subject discovery & discussion, by embedding basic Talk functionality into the Classify page.
- Design: https://projects.invisionapp.com/d/main?origin=v7#/console/12923997/380590378/preview

### Status

WIP


shaunanoordin:
## PR Update

The big questions of this PR are: _how do I know if a user is logged in, and which user is it?_

TL;DR: as a ⚠️ _ short term quick fix,_ the QuickTalk component will _independently_ use the authClient to find out which user is logged in.

### Dev Notes

Full discussion: https://github.com/zooniverse/front-end-monorepo/discussions/2362 

Some context to the issue:

- `lib-classifier` doesn't (or rather, _shouldn't_ ) know about the user. At least, as currently designed, the classifier's _responsibility_ is to only to handle Subjects + Classifications.
- `app-project` is the bit that's supposed to be in charge of managing the User resource (including figuring out when a login/logout occurs, and who's logged in).
- However, the QuickTalk project DOES require the User resource. So how do we get it from `app-project`?
- Passing the `User` resource down as a param is the sensible way of doing it, but requires a large re-jiggering of the codebase, which is inadvisable for this **experimental feature.**
- The solution is a quick fix: QuickTalk will _independently_ use the authClient to find out which user is logged in.
  - This is a _duplication of work/responsibility_ done by `app-project`, but it's a clean way of separating this experimental feature from the rest of Classifier's mechanisms. (i.e. we can pull it out without negatively affecting the rest of the Classifier)
  - Anyway, as it turns out, something else in the Classifier is _already_ doing this duplication of work/responsibility: the UserProjectPreferences store. It's not supposed to do so, as per the original intent of "separation of responsibilities between packages", but there you have it.
- Long term, we will need to rip out UserProjectPreferences store and QuickTalk's short-term solution, and properly pass down the User resource. But let's do that after we have this experimental feature running and we have a better idea of how the User resource will fully be used.

shaunanoordin:
## PR Update

This PR is now complete!

### QuickTalk Component Documentation

**Overview**

- This component is intended to promote Subject discovery & discussion, by embedding basic Talk functionality into the Classify page.
- Caveats: 
  - As of August 2021, this is an **experimental feature**
  - To enable it, add `quicktalk` to a Project's `experimental_tools`
  - As part of its experimental nature, it _independently checks the User resource._

**Features**

Screenshot 2021-09-02 at 20 57 19

- **QuickTalk Button:** QuickTalk starts in a 'collapsed' state, i.e. as a button
  - The QuickTalk button maintains a fixed position (bottom right of the screen)
  - This button will have a badge indicating the number of comments, if there's 1 comment or more. (note 1)
  - Clicking on the button will expand QuickTalk.

Screenshot 2021-09-02 at 20 57 28

- **QuickTalk Panel:** in its expanded state, QuickTalk will list all Talk Comments for this Subject's discussion
  - The QuickTalk panel maintains a fixed position (bottom right of the screen), and a maximum size (HTML overflow allows users to scroll through all comments) 
  - Clicking on 'Subject Discussion' in the component header will open the Subject's Talk link _in a new window_
  - Clicking on the close button in the component header will collapse the QuickTalk component.
  - Comments will be listed oldest-to-newest (note 2)
    - Similar to classic Talk, Comments will contain: the user's avatar (with a default avatar as a fallback), display name, login, role(s), and comment text (using standard markdown)
    - TODO: maybe host that default avatar somewhere less arbitrary?
    - TODO: maybe add a different style for deleted comments? (This is in PFE)
  - At maximum, this should list 20 comments (note 2) (note 3)
  - Excluded functionality:
    - ✖️ You can't click on a user's name to view their profile on Talk. (note 4)
    - ✖️ You can't edit your comments within QuickTalk. (note 5)

Screenshot 2021-09-02 at 20 57 36

- **Posting Comments:** a logged-in user can post comments.
  - The "Post Comment" form ONLY appears IF the user is signed in AND QuickTalk is in its expanded state. It's a simple text area and "submit" button.
  - Upon successfully posting a comment, all comments are re-fetched.
  - Users cannot post empty comments.
  - The "Post Comment" form can show error messages.
  - The "Post Comment" form will be disabled when a comment is being sent (to prevent multiple sends), and will have a "posting comment..." message.
  - ⚠️ Dev note: the "Post Comment" functionality checks that a user is logged in before each post. This is due to user authentication cautions on lib-classifier.
  - ⚠️ Dev note: be aware that there is a logic difference between *posting the first comment for a Subject* vs *posting the second/third/etc comment for a Subject.* The former requires the code to create a 'discussion' resource first, the latter simply adds a new 'comment' resource to an existing 'discussion' resource.
  - Excluded functionality:
    - ✖️ "Post Comment" form uses a basic textarea, not a Markdown Editor. (note 5) 

Footnotes: 🦶📓 
- (1) grommet.Button.badge crashes tests and Storybook for some reason, so have been removed from .spec.js and .stories.js
- (2) these design choices are subject to change once we have testers try out QuickTalk
- (3) 20 is the Talk default, _I think._ In any case, the reason is to avoid adding pagination unless necessary.
- (4) Reason: didn't want to give users reason to keep opening new windows/tabs to Talk.
- (5) Reason: this would have introduced complexity for a supplementary feature. Users can open the Talk link for full functionality.

### Testing

Package: `lib-classifier`

- Test URL: https://local.zooniverse.org:8080/?env=staging&project=1651&workflow=3262&demo=true (a test project with `quicktalk` enabled in its experimental tools)
- Verify all the "Features" listed above.
- Note: don't forget to ensure you're logged in to test "Posting Comments". Logging in on localhost can be a slight faff sometimes.
- Note: it may be difficult to test "adding _the first comment_ to the Subject's discussion" since I've kinda commented a lot on all my Subjects. You can either add the `quicktalk` experimental tool to your own test Project, or admin you way into Staging Project 1651 WF 3262 and add some new Subjects of your own to test.

Package: `app-project`

- Test URL: https://local.zooniverse.org:3000/projects/darkeshard/transformers/classify/workflow/3262?env=staging
- Same notes as above.

### UPDATE

Dangit I just noticed a few new things as I type this PR Update.

❔ Minor UI Issue: on `app-project`, the QuickTalk button (collapsed) kinda clips with the "switch to Dark Theme". I need to bump the distance a bit

❔ Minor UI Issue: the QuickTalk panel "jumps" to the left once you start typing a comment. I think the fixed position: right doesn't recognise that the scrollbar has eaten up a part of the screen until it's nudged. I have no idea why this is, and am open to suggestions.

❗ **Data Issue:** we must explicitly define `?env=staging` to get the correct Users
- Compare https://local.zooniverse.org:3000/projects/darkeshard/transformers/classify/workflow/3262 with https://local.zooniverse.org:3000/projects/darkeshard/transformers/classify/workflow/3262?env=staging . Keep refreshing until you get the same subjects
- You should mostly see comments from (staging) `@darkeshard` and `@zootester1`. These are my test accounts.
- HOWEVER, if you don't have `?env=staging`, then you'll actually get users  from PRODUCTION. e.g. `@darkeshard` becomes `@Anna_Kennis`
- The comments are correct, only the user avatars are wrong!
- I have... no idea why this is happening.
- Actually, I do. I just think it's stupid. I think `talkClient` (from 'panoptes-client/lib/talk-client') respects the implicit env set by console, but `apiClient` (from 'panoptes-client/lib/api-client') doesn't?!?

### Status

This PR is ready for review! Functionality is complete, and this is ready to be rolled out on test projects. Do note that I may come back to tweak some styles here and there.

eatyourgreens:
Preliminary testing looks good: I can login, read comments and leave a reply.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
```diff
@@ -0,0 +1,295 @@
...
+  subject: PropTypes.object,
+}
+
+QuickTalkContainer.defaultProps = {
+  authClient: undefined,
+  enabled: false,
+  subject: undefined,
+}
+
+@inject(storeMapper)
```
eatyourgreens, line 290: Check out `helpers/withStores`. It’s a helper I wrote to replace `inject` in new components.

eatyourgreens (approved):
LGTM. Let’s get this out onto staging.

PR #2330, opened 2021-08-02
Classifier: Add a useStores hook and withStores decorator

eatyourgreens:
This adds a `useStores` [custom hook](https://reactjs.org/docs/hooks-custom.html), which makes connectors easier to write. It takes an optional `storeMapper` function, and returns the store filtered through that function. Default behaviour is to return the entire store.

I've used [`@testing-library/react-hooks`](https://react-hooks-testing-library.com/usage/advanced-hooks) to test the new hook.

Usage example:
```js
  import { observer } from 'mobx-react'
  import { useStores } from '@helpers'

  function storeMapper(store) {
    const { workflows } = store
    return { workflows }
  }

  function ConnectedComponent(props) {
    const { workflows } = useStores(storeMapper)
    // do stuff with the store props here.
   return <ViewComponent workflows={workflows} {...props} />
  }

  export default observer(ConnectedComponent)
  ```

I've also added a `withStores` decorator, which can be used to build connectors that have no internal state of their own. Here's an example, from the `DisabledTaskPopupConnector`, which generates `nextAvailable` and `reset` props for `DisabledTaskPopup`.
```js
import React from 'react'
import { withStores } from '@helpers'
import DisabledTaskPopup from './DisabledTaskPopup'

function storeMapper(store) {
  const {
    subjects: {
      nextAvailable,
      clearQueue
    }
  } = store

  return {
    nextAvailable,
    reset: clearQueue
  }
}

export default withStores(DisabledTaskPopup, storeMapper)
```

I’ve used this to generate connectors for the classifier task area (`TaskArea`, `DisabledTaskPopup`, and `Tasks`.)

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
This broke drawing tool tests that stubbed `console.error`, so I've rewritten those in this PR too.

srallen (commented):
Could you elaborate on this?:

> This broke drawing tool tests that stubbed console.error

It's not really clear to me why the changes here broke this.

In addition, could the comments in `useStores` and `withStores` describing what they do be moved into a readme.md file in their respective folders? This is just so we have consistency in where documentation is placed within the library.

eatyourgreens:
>It's not really clear to me why the changes here broke this.

I'm not sure either but `git checkout HEAD~2` (to remove those changes) and the `Tool` tests fail with:
```
  2405 passing (1m)
  9 pending
  1 failing

  1) Model > DrawingTools > Tool
       tool.createTask
         should error for invalid subtasks:
     TypeError: console.error.withArgs is not a function
      at Context.<anonymous> (src/plugins/drawingTools/models/tools/Tool/Tool.spec.js:131:28)
      at processImmediate (internal/timers.js:456:21)
```

Any advice you might have for fixing that would be appreciated.

eatyourgreens:
>In addition, could the comments in useStores and withStores describing what they do be moved into a readme.md file in their respective folders? This is just so we have consistency in where documentation is placed within the library.

I'd prefer to continue using JSDoc comments, going forward. We could generate Readme files from those, if necessary.

I've been using JSDoc comments to document components since Storybook 6 came out, and I'm now in the habit of writing JSDoc comments for anything I work on. It seems straightforward, and gives us self-documenting components via the Docs page in stories. I'd prefer not to use a mix of JSDoc for components and Readme files for helper functions. Now that support for `addon-notes` has officially ended (in the latest Storybook release?) I'm not sure how useful it is to write Readme files by hand for new code.

eatyourgreens:
`DisabledTaskPopup` is an example of documentation that I’ve written with JSDoc, rather than writing a Readme. 
https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/index.html?path=/docs/tasks-disabled-task-popup--default

eatyourgreens:
Alternatively, we could carry on writing Readme files and import them into stories as a doc description, like the Modal component here:
https://github.com/zooniverse/front-end-monorepo/blob/053bf8389ae9a1ae765aad5be646b388948126eb/packages/lib-react-components/src/Modal/Modal.stories.js#L34-L40

Personally, I prefer documenting code with JSDoc (it keeps descriptions next to the code that they describe) but I don't think I've got a strong opinion about this.

srallen:
@eatyourgreens I'd rather not have a mix of styles. This was a problem in PFE and makes that codebase hard to navigate. We need predictable places to find information and what is established is separate readme files. I tend to not like documentation generation tools because often those docs are really quite poorly written and hard to read (here's looking at you API docs for MST).

We could consider migrating documentation to JSDoc, but I'd rather have that be a holistic change than sprinkled in here and there.

eatyourgreens:
Reading through the Storybook docs again, I think JSDoc is required for component props, but optional for components descriptions.
https://storybook.js.org/docs/react/writing-docs/doc-blocks

srallen:
@eatyourgreens the testing-library plugin installed with this branch is stubbing console.error automatically in before and after hooks which I think is interfering with the Tool specs: https://react-hooks-testing-library.com/reference/api#consoleerror

About documentation:

> Reading through the Storybook docs again, I think JSDoc is required for component props, but optional for component descriptions.

The inline comments are required if you want Storybook to generate documentation using react docgen. Props can be described using argTypes object as well. 

I agree we shouldn't have a mix of styles and I think we could consider react docgen as well as other document generation as an option more holistically, but for now, I'd like consistency with the docs for helper functions to follow what the repo has already done with markdown readme files. If you'd like to have a discussion about switching over to generation for all documentation, then could you open a discussion thread about that? I don't think anyone else has been using these generation tools, so if we do this, then an opportunity for discussion, knowledge sharing, and buy in with the whole team should happen first. Generally speaking, I tend to really dislike documentation generation because those docs tend to be very poor, both in writing and in organization. However, perhaps we can come to an understanding about how best to use them and when, particularly when Storybook is involved.

eatyourgreens:
Thanks for the testing library link. That's useful. We patch `console.error` in the feedback tests too, but scope the stub to that function. Maybe that's a way to avoid clobbering it globally?
https://github.com/zooniverse/front-end-monorepo/blob/9ed2da0233a509f5b123d6b10a427b3010ccd3f5/packages/lib-classifier/src/store/feedback/helpers/metadata-to-rules.spec.js#L52-L56

Doing a quick search for `sinon.stub(console, 'error')`, we're stubbing it globally in quite a few places in the tests.
https://github.com/zooniverse/front-end-monorepo/search?q=sinon.stub%28console%2C+%27error%27%29

I've moved the Tool changes into #2386, so that they can be reviewed separately.

srallen (approved):

PR #2346, opened 2021-08-09
Engaging Crowds: test DisabledTaskPopup

### LINKED ISSUE #2297: Engaging Crowds: add tests for the task area popup
## Package
lib-classifier

#2198 added a popup, which disables the task area when a retired subject loads for a workflow with indexed subjects.
https://github.com/zooniverse/front-end-monorepo/blob/6d60de4374146540f55119c57eaf3850b8a1476d/packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskArea.js#L32-L37

`useEffect` here needs tests to check that we don't accidentally introduce this for other workflows.

eatyourgreens:
Mock a grouped workflow, with an indexed subject set, for the TaskArea tests. Render the task area with and without an indexed workflow, and with and without finished subjects, and test that the popup renders correctly for indexed, finished subjects.

I've set the tests up to run when `store.subjects` is ready (copying a technique from the UPP store tests) and test whether the popup text and buttons are shown. I've also added tests to check whether the task inputs are disabled. I haven't added any tests to check what happens when the popup buttons are pressed.

Package:
lib-classifier

Closes #2297.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskArea.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskArea.js
@@ -65,7 +65,6 @@ export default function TaskArea({
         flex
       >
         <Tab
-          disabled={disabled}
           title={counterpart('TaskArea.task')}
         >
           <Box
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskAreaConnector.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskAreaConnector.spec.js
@@ -1,58 +1,204 @@
-import * as React from 'react'
+import zooTheme from '@zooniverse/grommet-theme'
 import { render, screen } from '@testing-library/react'
 import { expect } from 'chai'
-import { Provider } from 'mobx-react'
 import { Grommet } from 'grommet'
-import zooTheme from '@zooniverse/grommet-theme'
+import { when } from 'mobx'
+import { Provider } from 'mobx-react'
+import nock from 'nock'
+import React from 'react'
+import { Factory } from 'rosie'
+
+import { SubjectFactory, SubjectSetFactory, WorkflowFactory } from '@test/factories'
+import mockStore from '@test/mockStore'
 import TaskAreaConnector from './TaskAreaConnector'
 
+import popupText from './components/DisabledTaskPopup/locales/en'
+
 describe('TaskAreaConnector', function () {
-  const mockStore = {
-    annotatedSteps: {
-      latest: {
-        isComplete: () => {},
-        tasks: []
-      }
-    },
-    classifications: {
-      active: { id: '10' },
-      demoMode: false
-    },
-    subjects: {
-      active: { id: '2' }
-    },
-    subjectViewer: {
-      loadingState: 'success'
-    },
-    tutorials: {
-      disableTutorialTab: false,
-      setActiveTutorial: () => {},
-      tutorial: { id: '1' }
-    },
-    workflows: {
-      active: {
-        configuration: {},
-        id: '5'
-      },
-      loadingState: 'success'
-    },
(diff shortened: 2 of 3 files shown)


packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskArea.js
```diff
@@ -65,7 +65,6 @@ export default function TaskArea({
         flex
       >
         <Tab
-          disabled={disabled}
```
eatyourgreens, line 68: Grommet warns that an active tab can't be disabled, so I've removed this to fix that warning.

packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskAreaConnector.spec.js
```diff
@@ -1,58 +1,277 @@
...
+      selectButton = screen.queryByText(popupText.options.select)
+      nextButton = screen.queryByText(popupText.options.next)
+      dismissButton = screen.queryByText(popupText.options.dismiss)
+      inputs = screen.queryAllByRole('radio')
+    })
+
+    it('should render the active task', function () {
+      expect(inputs).to.not.be.empty()
+      inputs.forEach(input => {
+        expect(input.disabled).to.be.false()
```
eatyourgreens, line 65: Outside the scope of this PR, but I think this should be `true` until `store.subjectViewer.onSubjectReady()` runs. Otherwise, we can classify before the subject image has loaded in the browser.

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR updates tests for the TaskArea to test for the _"This subject has enough classifications, maybe do something else?"_ popup.

- Context: the "DisabledTaskPopup" only appears for special workflows + subject sets that allow individual Subject selection, and only when the user chooses a Subject that is fully classified/retired (and therefore is prompted to choose another Subject)
- No functional changes to components. lib-classifer compiles fine. Cool 👌 
- All tests for lib-classifier pass.
- New tests are now split between "WF + SSet have no indexed Subjects" and "WF + SSet have indexed Subjects"
  - [extremely minor] If I were to be super nitpicky, I'd add a line or comment that says "btw, 'no indexed subjects' is the default for like 98% of projects", but honestly nobody who works with these specs.js files wouldn't know that already.

### Status

LGTM 👍

packages/lib-classifier/src/components/Classifier/components/TaskArea/TaskAreaConnector.spec.js
```diff
@@ -1,58 +1,277 @@
...
+            required: false,
+            taskKey: 'T0',
+            type: 'single'
+          }
+        },
+        links: {
+          subject_sets: [subjectSetSnapshot.id] 
+        }
+      })
+      const subjectSetAPI = mockSubjectSetAPI(subjectSetSnapshot)
```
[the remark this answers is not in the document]
eatyourgreens: `nock` sets up request responses that can be used once, unless you call `nock().persist(true)`. In that case, every request is intercepted until you call `nock.clearAll()`. 

I think they’re scoped to the current block but I’m not sure about that.

I can’t remember why I named this. Maybe to make it clear which API is being mocked?

eatyourgreens:
I rebased this to fix the merge conflict, and bring in the changes to the mock store from #2402. Now, these new tests don't run at all. This is going to need more time than I've got at the moment. I'll come back to it.

eatyourgreens:
I think I figured out the problem with the mock store: if a mock subject is supplied, it should be included as the first subject returned in the mock Panoptes API response.

eatyourgreens:
Tests seem to be failing due to a timeout in survey task choices, so I’m going to re-run them.

PR #2350, opened 2021-08-10
Classifier: remove Ellipse component's unused `.children`

shaunanoordin:
## PR Overview

Package: `lib-classifier`

This PR remove the Ellipse component's `.children` property, which _isn't used_ but _is required._ I'm guessing this is a holdover from some older version of the code.

- Removing the unused `.children` property also fixes the _"Warning: Failed prop type: The prop `children` is marked as required in `Ellipse`, but its value is `undefined`."_ React error that appears whenever we create a new Ellipse mark.

### Testing

Tested on macOS 11.5.1 + Chrome 92 on localhost.

- Pick any project with a Drawing task with an Ellipse tool, e.g. I Fancy Cats
- Draw an ellipse.
- Expected: no React error should appear in the console. The Ellipse mark should be drawn as expected.

### Status

Ready for review


ErikOstlund (approved):
Ellipse tool works as expected. Can draw, move, change x_radius, change y_radius and delete ellipse. All unit tests are passing.
Looks good! 👍

shaunanoordin:
Oh pants! Thank you for the reminder

PR #2351, opened 2021-08-11
reinstate PR 2309 - Docker build and compose improvements

camallen:
Reverts zooniverse/front-end-monorepo#2345 which in turn reverted #2309. 

PR #2309 was reverted as it was blocking deploys due to Jenkins failing to build the docker building images by using BUILDKIT features but BUILDKIT was not enabled on jenkins docker.  

This PR
1. adds DOCKER_BUILDKIT env varaible to Jenkinsfile to access BUILDKIT features for this repo only** (not globally), https://docs.docker.com/develop/develop-images/build_enhancements/#to-enable-buildkit-builds
2. and reinstates the features in #2309 (BUILDKIT build features of caching yarn dependencies for faster CI builds and improved compose setup for testing and running apps)

**Longer term buildkit will be the default build engine of docker so this ENV opt in won't be needed once that happens

Linked to https://github.com/zooniverse/talk-api/pull/271 which enabled buildkit docker build features using Jenkins ENV var


eatyourgreens:
Looks good to me. When I removed dev dependencies from the production Docker image, I also broke the local development environment for the classifier and the Next apps. If you do this:
```
cd packages/lib-classifier
docker compose up
```
the dev classifier crashes because `webpack` isn't available.

I've started on a branch that keeps `_dev` as a dev image with all dependencies, and `_prod` as a production image with only the production deps installed. Maybe that should be rolled into this? My changes are in 09cd68ed79a012f2d01a0c72b0b82794ae213514.

eatyourgreens:
I'm seeing a warning about a missing yarn cache during the build steps:
`warning Cannot find a suitable global folder. Tried these: "/usr/local, /home/node/.yarn"`

eatyourgreens (approved):
Changes look good to me. `docker compose build` builds two images locally:
- `front-end-monorepo_shell`: a build with all dependencies installed.
-  `front-end-monorepo_dev`: a build with no dev dependencies.

I'd maybe change the names of those images to make it clear which one is for development work and which one matches the production deploy. It's a bit confusing that the `_dev` image is actually production. If we kept `_dev` for the development image, that would also work with the docker compose configs for the classifier library and NextJS apps in local development.

`docker compose up` starts the two NextJS apps on ports 3000 and 3001. Tested with:
- http://local.zooniverse.org:3000/projects/msalmon/hms-nhs-the-nautical-health-service
- http://local.zooniverse.org:3001/about/publications

I can also run a shell with `docker compose run --rm shell`.

I'm seeing a yarn warning during the individual build steps:
`warning Cannot find a suitable global folder. Tried these: "/usr/local, /home/node/.yarn"`
When I run `yarn cache dir` inside a container, I get `/home/node/.cache/yarn/v6` so maybe the path needs changing in the Dockerfile?

EDIT: does `YARN_CACHE_FOLDER=/home/node/.yarn` need to be set in the build environment? At the moment, it's only set for the `yarn install` command and I'm seeing that warning during the `yarn build` steps for each individual package.

camallen:
> I'm seeing a yarn warning during the individual build steps:
> `warning Cannot find a suitable global folder. Tried these: "/usr/local, /home/node/.yarn"`
> When I run `yarn cache dir` inside a container, I get `/home/node/.cache/yarn/v6` so maybe the path needs changing in the Dockerfile?

I've ensured the cache folder is mounted for the build steps now as well. There are no warnings now, just a verbose docker file.

> EDIT: does `YARN_CACHE_FOLDER=/home/node/.yarn` need to be set in the build environment? At the moment, it's only set for the `yarn install` command and I'm seeing that warning during the `yarn build` steps for each individual package.

I've opted to re-use the docker cache mount across the relevant Dockerfile run cmds rather than setup a cache dir ENV and ensuring that path exists and is writable by node user. This way the cache is-reused for the builds as well we could look at doing it another way long term.

eatyourgreens:
Just ran one more check locally. Everything builds and runs. I can run a production image with both the project app and the content pages app. 👍

PR #2355, opened 2021-08-13
Add Circle Drawing Tool

### LINKED ISSUE #1288: Circle drawing tool
> a point and a radius.

## Related code in PFE

- [circle.cjsx](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/drawing-tools/circle.cjsx)

## Related Issues

- #500

ErikOstlund:
Package: `lib-classifier`

Closes #1288 

## Describe your changes:
This PR adds the `circle` drawing tool to the available tools in the `front-end-monorepo`.
The `circle` is created from its center point (x_center, y_center) and radius.
A README file was created to note some gotcha's!
The unit test is a work in progress. This will be completed in a different PR as I gain knowledge of the `React Testing Library`.

If a circle is drawn like so:
Pasted Graphic

The following data will be sent to our api:
frame 0

## Testing
In the front-end-monorepo folder, run `yarn bootstrap` to build the packages.
Then cd packages/lib-classifier and run `yarn dev`.

Project url: [https://localhost:8080/?env=production&project=16201&workflow=18675](https://localhost:8080/?env=production&project=16201&workflow=18675)


shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR adds the `circle` drawing tool to the Classifier's drawing tasks.

- The Subject Viewer's Interaction layer can now render 'circle' marks
- The Task Area can now support Drawing Task -> Circle Tool
- 'circle'-type annotations can now be submitted in the classification.
  - Annotation attributes: `radius` (number), `x_center` (number), `y_center` (number)
  - See dev notes regarding annotation shape.

### Testing

Tested on localhost with macOS 11.5.1 + Chrome 92.

- Open a project with a drawing task with the circle tool: https://localhost:8080/?env=production&project=16201&workflow=18675
  - Note that for this project, we have a Subject image with a 1000x1000 grid for precise coordinate testing 
- I can click & drag to create a new Circle mark
- I can click & drag on a circle mark's drag handle to resize the circle.
- I can click & drag the circle to reposition it
- I can delete the circle mark
- I can perform all the above even when the Subject Viewer is zoomed and/or panned and/or rotated.
- When I submit the classification, the circle mark's annotation correctly reflects the `radius`, `x_center`, and `y_center` as determined by the 1000x1000 grid.

Functionality looks good to me! 👍  There is one quirk, though.

**Quirk:** depending on screen size and a number of other indeterminate factors, the position of the Delete button can be wildly off-position

_In the example below, I opened my dev console which caused the browser window to shrink to ~900px width. The Delete button for a circle looked fine UNTIL I refresh the page, after which point all circles will have off-position Delete buttons._

Screenshot 2021-08-19 at 22 19 49

_By "off-position", I mean that I expect the Delete button to be around (500,400), not (600,400). It's something to do with the dynamic scaling of the subject viewer, I believe._

### Dev Notes

I wanted to highlight a few things about the way the circle marks are annotated,
1. since they're different from PFE,
2. because they follow FEM's new rules for creating standardised drawing annotations: https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-24.md

The PFE circle mark (see the [PFE dev classifier](https://master.pfe-preview.zooniverse.org/dev/classifier)) creates an annotation like so:

```
{
  "x": 200,  // centre of the circle
  "y": 350,  // centre of the circle
  "r": 100,  // radius, shortened to 'r'
  "angle": 30,  // in degrees, anti-clockwise positive (!!!)
}
```

Compare that to the FEM circle mark:

```
{
  "x_center": 200,  // centre of the circle
  "y_center": 350,  // centre of the circle
  "radius": 100,  // spelled out as 'radius'
}
```

Asides from the changed names, the FEM mark _removes the angle_ attribute, which I think shoould be intentional since the angle doesn't make any sense to a circle. (In PFE, you can actually rotate a circle, it just has very very VERY little visual effect.) To be safe, I'm going to ask the #project-building channel about this, but don't let my bonus investigation hold up this PR.

#### Bonus Dev Notes

SIDE note: while I was writing this up, I wanted to also make a comparison to the FEM ellipse mark, which creates an annotation like so:

```
{
  "x_center": 200,
  "y_center": 350,
  "rx": 100,  // radius in the x-direction, also shortened to 'rx'
  "ry": 150,  // radius in the y-direction, also shortened to 'ry'
  "angle": 30,  // in degrees, clockwise positive
}
```

- I'm pretty sure that according to ADR-24, the angle is _supposed_ to be clockwise positive now. That's what _RHC_ is, right? ~~Gosh, I should double-check.~~ UPDATE: yes, RHC is defined in the ADR, I shouldn't overthink and confuse it with _general physics_ RHC rules.
- I'm not sure if the ellipse tool should have radius_x and radius_y instead of rx and ry, that seems a bit inconsistent with ADR-24. Again, I should check this on the dev channel.

### Status

This PR overall looks good; code's readable and functionality checks out. My only notes are:

- I suggested some minor wording changes in regards to the "cx, cy" comments. But overall, documentation is good! 👍 
- That quirk needs investigating, but that should be its own standalone UI investigation issue/PR. Don't let it block this main PR.

LGTM! 👍

srallen (changes requested):
The annotation model should use `r` as its key rather than `radius`. Me spelling it out in the ADR is either a mistake or an undocumented requested change. We can't confirm right now, so let's go with what PFE used, which is `r`. We can always change this later.

PR #2365, opened 2021-08-18
Always hide workflow summaries

eatyourgreens:
Remove `hide_classification_summaries` from `workflowSteps.shouldWeShowDoneAndTalk`, so that we always show Done & Talk on the last step of a workflow (except for flagged classifications.)

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/WorkflowStepStore/WorkflowStepStore.spec.js
+++ b/packages/lib-classifier/src/store/WorkflowStepStore/WorkflowStepStore.spec.js
@@ -344,41 +344,37 @@ describe('Model > WorkflowStepStore', function () {
       expect(rootStore.workflowSteps.shouldWeShowDoneAndTalkButton).to.be.false()
     })
 
-    it('should be true by default', async function () {
-      const project = ProjectFactory.build({}, { activeWorkflowId: workflow.id })
-      const panoptesClientStub = stubPanoptesJs({ workflows: workflow, subjects })
-      const rootStore = await setupStores(panoptesClientStub, project, workflow)
-      rootStore.classifications.createClassification(subject, workflow, project)
-      rootStore.workflowSteps.selectStep('S2')
-      expect(rootStore.workflowSteps.shouldWeShowDoneAndTalkButton).to.be.true()
-    })
+    describe('for any value of hide_classification_summaries', function () {
+      let workflows
 
-    it('should be false if the classification subject has been flagged', async function () {
-      const project = ProjectFactory.build({}, { activeWorkflowId: workflow.id })
-      const panoptesClientStub = stubPanoptesJs({ workflows: workflow, subjects })
-      const rootStore = await setupStores(panoptesClientStub, project, workflow)
-      rootStore.classifications.createClassification(subject, workflow, project)
-      rootStore.classifications.active.metadata.update({ subject_flagged: true })
-      rootStore.workflowSteps.selectStep('S2')
-      expect(rootStore.workflowSteps.shouldWeShowDoneAndTalkButton).to.be.false()
-    })
+      before(function () {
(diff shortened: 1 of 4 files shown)


packages/lib-classifier/src/store/WorkflowStepStore/WorkflowStepStore.spec.js
```diff
@@ -363,13 +363,13 @@ describe('Model > WorkflowStepStore', function () {
       expect(rootStore.workflowSteps.shouldWeShowDoneAndTalkButton).to.be.false()
     })
 
-    it('should return false if configured to show the summary', async function () {
+    it('should return true if configured to show the summary', async function () {
```
eatyourgreens, line 366: Maybe the wording here should be more like 'should ignore workflow configuration' to make it easier to understand what the code's intended to do? I'm not sure.

srallen (changes requested):
I think we should remove the `hide_classification_summaries` from the workflow configuration model too. The setting there is irrelevant now to the classifier and no longer used. 

- FEM classifier has no UI built to display summaries
- If and when we do build one, possibly what determines to show it could be part of workflow steps property rather than workflow configuration, as suggested in the ADR: https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-05.md. This of course, is really contingent on what is being summarized and if it should be pre or post classification (cases have been made for either).

There's no plan to do either of these things in the upcoming milestones for parity with PFE, so it's slated for deprecation, unless someone can come up with a good use case.

To that end, the tests need a bit more updating:

- The test description on line 347 no longer makes sense
- The tests on line 366 and 375 are duplicative of the test for 347, so should be removed.

eatyourgreens:
I've removed `hide_classification_summaries` from the config.

I think we'll still need regression tests for the three types of workflows from Panoptes, which can have `hide_classification_summaries` set to 'undefined', 'true' or 'false' in the database but should show 'Done & Talk' regardless.

packages/lib-classifier/src/store/WorkflowStepStore/WorkflowStepStore.spec.js
```diff
@@ -344,41 +344,37 @@ describe('Model > WorkflowStepStore', function () {
...
-    it('should return false if configured to show the summary', async function () {
-      const project = ProjectFactory.build({}, { activeWorkflowId: showSummaryWorkflow.id })
-      const panoptesClientStub = stubPanoptesJs({ workflows: showSummaryWorkflow, subjects })
-      const rootStore = await setupStores(panoptesClientStub, project, showSummaryWorkflow)
-      rootStore.classifications.createClassification(subject, showSummaryWorkflow, project)
-      rootStore.workflowSteps.selectStep('S2')
-      expect(rootStore.workflowSteps.shouldWeShowDoneAndTalkButton).to.be.false()
-    })
+      it('should be true by default', async function () {
+        const awaitWorkflows = workflows.map(async function (testWorkflow) {
```
eatyourgreens, line 355: Fun fact: you can't use `forEach` with promises, so this uses `map` and `Promise.all`. Otherwise, the tests run before the store has been set up and pass by default.
https://zellwk.com/blog/async-await-in-loops/

srallen:
> I think we'll still need regression tests for the three types of workflows from Panoptes, which can have hide_classification_summaries set to undefined, true or false in the database but should show 'Done & Talk' regardless.

If it's removed from the model, the those values aren't available to the classifier MST stores at all. The extra safety is fine, I suppose, but it's testing for something that isn't possible to affect the function any longer.

srallen (approved):

Issue #1851, opened 2020-10-15
[VSLCV] Highlight should move data to the "background"


adamamiller:
**Package**

VSLCV

**Is your feature request related to a problem? Please describe.**

The highlight feature is an extremely useful way to focus on a single filter in the light curve viewer. Right now the observations are plotted such that the first filter is plotted, then the second (and in the future subsequent) filters are plotted "on top" of the first filter. If the highlight button is used to de-emphasize the 2nd filter, those points are still shown "on top" of the first light curve. See the screenshot below.

![image](https://user-images.githubusercontent.com/2434744/96157831-6bd86280-0ed8-11eb-8437-1f2f1ebb2c15.png)

**Describe the solution you'd like**
It would be nice if observations that are not being highlighted "recede" to the background. In the screenshot above it is much easier to see the "g" band observations than it is when both filters are shown:
![image](https://user-images.githubusercontent.com/2434744/96158010-a5a96900-0ed8-11eb-9498-0f40424cfe31.png)

However, focusing on just r is much easier than just g because the g band observations are "behind" the r-band observations (when r is the only filter highlighted): 

![image](https://user-images.githubusercontent.com/2434744/96158118-c5d92800-0ed8-11eb-9ad8-d2a9cf235e07.png)

**Describe alternatives you've considered**
If this request leads to performance issues (because the entire plot would need to be re-rendered every time then this feature may not be worthwhile. 

**Additional context**
This feature is *not required* prior to the planned upcoming beta test. But it would be nice to have before the project goes fully live before the public.

srallen:
The z-index of SVG elements is determined by its rendering order, so right now, the data points are programmatically rendered in the array order that they're supplied in the subject JSON. There may be a way to manipulate this via CSS however.

srallen:
@adamamiller this is now fixed and deployed.

PR #2366, opened 2021-08-18
Simplify highlight state

srallen:

Package:

Toward #1851

I've refactored the highlight state to be more simple to use in upcoming refactors to fix the issue described in #1851. The basic idea is that originally the highlight state relied on array order of the data points to figure out which series is "highlighted" (meaning should have color fill), but once #1851 is fixed, the logic can't rely on the order any longer. SVG doesn't have the ability to be z-index by CSS, so how something renders on "top" of another SVG object is determined by the literal order of the SVG elements in the DOM, so the data series points themselves will have to be reordered (coming in a future PR).

This can be manually checked to make sure there's no regression by running the storybook and trying out the highlight toggles in the VariableStarViewer stories.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VariableStarViewer/VariableStarViewerContainer.js
```diff
@@ -222,8 +222,9 @@ class VariableStarViewerContainer extends Component {
     return scatterPlotJSON.data.map((series, index) => {
       if (series?.seriesData.length > 0) {
         const fallbackLabel = counterpart('VariableStarViewer.label', { id: index + 1 })
-        const label = series.seriesOptions?.label || fallbackLabel
-        return { [label]: true }
+        series.seriesOptions.label ??= fallbackLabel
```
mcbouslog, line 225: Cool use of `??=`!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/helpers/isDataSeriesHighlighted/isDataSeriesHighlighted.js
```diff
@@ -1,4 +1,4 @@
-export default function isDataSeriesHighlighted (highlightedSeries = [], seriesIndex = 0) {
-  const [highlighted] = (highlightedSeries?.length > 0) ? Object.values(highlightedSeries[seriesIndex]) : [true]
-  return highlighted
+export default function isDataSeriesHighlighted ({ highlightedSeries: highlightedSeries = [], seriesOptions: seriesOptions = {} } = {}) {
+  if (highlightedSeries.length > 0 && seriesOptions.label) return highlightedSeries.includes(seriesOptions.label)
+  return true
```
mcbouslog, line 3: I think `return true` is causing issues when attempting to uncheck both series, should be `return false`? Though I'm not sure in practice when one would ever want to uncheck both series.

mcbouslog (changes requested):
One question relating to `isDataSeriesHighlighted` otherwise looks great.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/helpers/isDataSeriesHighlighted/isDataSeriesHighlighted.js
srallen: This is now fixed. I do want this to default to true, but the problem was having the default parameter for `highlightedSeries`. The reason is that this helper is used both by the `VariableStarViewer` and underlying `ScatterPlotViewer` on its own, so if you have multiple data series in the base scatter plot, we want them to all have fill colors since the highlight state is not toggleable there.

srallen:
@mcbouslog issue with highlighting is now fixed. The highlight toggles for the `VariableStarViewer` should work as expected as well as the multiple data series story for the `ScatterPlotViewer` should still render a plot with all the data series having fill colors.

mcbouslog (approved):
Changes LGTM!

PR #2375, opened 2021-08-19
Remove `plop`

shaunanoordin:
## PR Overview

Packages: _root_

This PR removes `plop` from the Front End Monorepo.

- Plop is used to create new apps (e.g. lib-classifier) or components
- Unfortunately, this system hasn't been updated in yonks so I'm not sure anyone is even using this any more. (Please let me know if you're still using plop, though!)
- As a result, this old system is generating quite a few security alerts
  - These aren't serious security issues because they're only dev-side, but they're darn annoying.

### Status

Ready for review. Low priority. 💤 

I'm assuming nobody's using this system anymore, so I'm favouring removing it.

However, if someone on the team _is_ indeed still using plop, then I'm happy to change my mind on this, but this is what you'll need to do to keep the plop system up to date:

- Revise `plop/templates/app/package.json` to use the latest dependencies.
  - See `app-project` for the benchmark
  - At _minimum,_ this involves updating to next 11.1.0 and react 17.0.2
- Replace Enzyme tests with React-Testing-Library
- _I'm pretty sure the component templates themselves are still OK to reuse, but best check just in case._


eatyourgreens (approved):
Looks good to me. I've used `plop` to set up the boilerplate for new components, but I haven't done that in ages now, and the component templates look very outdated now.

PR #2379, opened 2021-08-23
Engaging Crowds: Show the subject set name for grouped workflows

eatyourgreens:
Add the subject set display name to the the `SubjectSetProgressBanner` for grouped workflows, so that we can see which subject set we are working on.

Try this out on HMS NHS:
https://localhost:8080/?project=11429&workflow=18619&subjectSet=82738&env=production&demo=true
Screenshot of an HMS NHS subject. The subject banner says: 1826-27 (DSH/1): Subject 51/101 (Finished).

Package:
lib-classifier

Towards #2278.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/Banners/components/SubjectSetProgressBanner/SubjectSetProgressBanner.js
+++ b/packages/lib-classifier/src/components/Classifier/components/Banners/components/SubjectSetProgressBanner/SubjectSetProgressBanner.js
@@ -8,6 +8,7 @@ import Banner from '../Banner'
 counterpart.registerTranslations('en', en)
 
 function SubjectSetProgressBanner({ subject, workflow }) {
+  const setName = workflow?.subjectSet.display_name
   const subjectTotal = workflow?.subjectSet.set_member_subjects_count
   const background = (subject?.alreadySeen || subject?.retired) ? 'status-critical' : 'status-ok'
   const color = (subject?.alreadySeen || subject?.retired) ? 'neutral-6' : 'neutral-7'
@@ -15,6 +16,7 @@ function SubjectSetProgressBanner({ subject, workflow }) {
   statusText = subject?.retired ? `${counterpart('SubjectSetProgressBanner.finished')}` : statusText
   const progressText = counterpart('SubjectSetProgressBanner.bannerText', {
     number: subject?.priority,
+    setName,
     total: subjectTotal
   })
   const tooltipText = counterpart('SubjectSetProgressBanner.tooltipText')
--- a/packages/lib-classifier/src/components/Classifier/components/Banners/components/SubjectSetProgressBanner/SubjectSetProgressBanner.js
+++ b/packages/lib-classifier/src/components/Classifier/components/Banners/components/SubjectSetProgressBanner/SubjectSetProgressBanner.js
@@ -8,7 +8,7 @@ import Banner from '../Banner'
 counterpart.registerTranslations('en', en)
 
 function SubjectSetProgressBanner({ subject, workflow }) {
-  const setName = workflow?.subjectSet.display_name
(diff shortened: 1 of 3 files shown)


packages/lib-classifier/src/components/Classifier/components/Banners/components/SubjectSetProgressBanner/SubjectSetProgressBanner.spec.js
```diff
@@ -19,26 +19,22 @@ describe('Component > SubjectSetProgressBanner', function () {
...
       const component = render(<Default />)
-      const bannerText = 'Subject 37/56'
+      const bannerText = 'Hello there!: Subject 37/56'
       expect(component.getByText(bannerText)).to.exist()
     })
 
     describe('when the subject is already seen', function () {
       it('should show the Already Seen banner', function () {
-        // Unsure why RTL isn't picking up the updated args from the story
-        // so explicitly setting already_seen to true here
```
eatyourgreens, line 29: The default story args were imported as `WithAlreadySeenSubject.args`, but weren't passed to the rendered component as props, so I've added that to the tests here.

srallen: I don't understand this. There aren't any default story args. I removed them from the default export of the stories and defined them explicitly on each story. It seems like you're doing the exact same thing I did, but in a different way. I would've expected the static `args` defined in the stories would've been already defined and imported.

packages/lib-classifier/src/components/Classifier/components/Banners/components/SubjectSetProgressBanner/SubjectSetProgressBanner.js
```diff
@@ -8,13 +8,15 @@ import Banner from '../Banner'
 counterpart.registerTranslations('en', en)
 
 function SubjectSetProgressBanner({ subject, workflow }) {
+  const setName = workflow?.subjectSet.display_name
```
srallen, line 11: There's a chance `subjectSet` could be undefined on the model, so I think for safety, there should be default:
```suggestion
  const setName = workflow?.subjectSet?.display_name || ''
```

srallen (approved):
I have some non-blocking change suggestions. I still don't understand what's going on with the `args` export and import.

packages/lib-classifier/src/components/Classifier/components/Banners/components/SubjectSetProgressBanner/SubjectSetProgressBanner.js
eatyourgreens: Good point, but subject selection would fail if there was no subject set selected, so I don't think this banner renders until after the subject set is set on the workflow.

packages/lib-classifier/src/components/Classifier/components/Banners/components/SubjectSetProgressBanner/SubjectSetProgressBanner.spec.js
eatyourgreens: I'm taking the static `WithAlreadySeenSubject.args`, imported with the story, and passing them to `WithAlreadySeenSubject` when it runs. They aren't explicitly defined as default props for the `WithAlreadySeenSubject` functional component, so they have to be explicitly passed in as props. Storybook does that for you, but React will use the defaults that are declared in the function definition.

If the story function was declared with default props, React would use those and you could render the component without having to set props:
```js
export function WithAlreadySeenSubject({ already_seen = true, dark = false, retired = false })
```
I'm not really sure why storybook doesn't use ES6 destructured defaults here. Maybe so it can support objects and complex types as default values?

PR #2384, opened 2021-08-23
Classifier: add 'Literal Insertion' text tags

### LINKED ISSUE #2331: & text modifier should be a single character, not brackets
For FEM transcription projects, the Text Modifiers menu includes an ampersand &.

This button currently creates bracketed modifiers: `[&][/&]`, which is not the expected behavior. It should generate a single character, `&`, without brackets. It's not a 'modifier' so to speak, but instead an option for volunteers to use in order to generate consistency for transcribing the `&` character (instead of `+` for example, which teams frequently see).

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Closes #2331 

We have a special request for the [Davy Notebooks project](https://www.zooniverse.org/projects/humphrydavy/davy-notebooks-project/) on FEM. Users want to have the `&` text modification tag, but they want it to _behave differently._ Instead of the standard tag insertion that goes `[&]...[/&]`, the insertion is literally just `&`. _I know, I know._

- ⚠️ This PR is a quick fix to address the feature request and should be treated as a short term measure.
  - A proper long-term solution is to split the "text modification tags" into a few categories: standard `[tag]...[/tag]`, single `[tag]`, and literal/raw `tag`
  - A long term solution also needs to visually distinguish the various categories of tag insertions, as _different UI elements should indicate different interaction behaviour._
- This PR adds an array of "special exceptions" to text tags, which will be treated as **literal insertion text tags.** (Actually, tell me if you can think of a better term.)
  - If a user attempts to use a "literal insertion" text tag, the standard `TextTask.setTagSelection()` will be overridden by `TextTask.setTagSelection_literalInsertion()`
  - As of now, only the `&` tag is a _hardcoded global exception._

### Testing

- Use a Text Transcription project that's configured to have the `&` text tag modifier, e.g. https://local.zooniverse.org:8080/?env=staging&project=1929&workflow=3459
- Transcribe a line. Type some text. Click on the `&` button to insert the `&` symbol. (Yes, the same thing that happens when you press Shift-7 on a US keyboard.)
- Baseline check: every other text tag should behave normally, with start `[tag]`s and end `[/tag]`s!

_Step 1a: either set the cursor to where you want to insert the `&` symbol, OR..._

Screenshot 2021-08-23 at 21 26 57

_Step 1b: highlight the text you want to REPLACE with the `&` symbol..._

Screenshot 2021-08-23 at 21 27 14

_Step 2: and click the `&` button. Done!_

Screenshot 2021-08-23 at 21 27 26

### Status

Ready for review. Tagging @snblickhan for comments [&][/&] feedback.


packages/lib-classifier/src/plugins/tasks/TextTask/components/TextTask/TextTask.js
```diff
@@ -44,6 +55,33 @@ function TextTask (props) {
...
+  - i.e. 'tag' instead of '[tag]...[/tag]'
+  - Requested for Davy Notebooks on Aug 2021: https://github.com/zooniverse/front-end-monorepo/issues/2331
+  - This is a temporary measure. Long term solution is to create a new category of text tags.
+ */
+  function setTagSelection_singleInsertion (e, ref) {
+    const currentRef = ref.current
+    const text = currentRef.value
+    const textTag = e.currentTarget.value
+    const startTag = `[${textTag}]`
+    const endTag = `[/${textTag}]`
```
eatyourgreens, line 71: I don't see these used anywhere, so I think you can get rid of them?

eatyourgreens (changes requested):
Changes look good to me, but I'd also ask for `&` to be added to the task story, and for the this new function to be added to the text tagging tests in the spec file.

shaunanoordin:
Thanks for the review, @eatyourgreens ! Feedback has been addressed:

## PR Update

- Tests now include _TextTask (component) > text tagging (special case for literal insertions)_
  - I did _not_ update TextTask.spec.js from enzyme to react-testing-library, however. I tried to, but only got halfway before my event tests started breaking, so this has to be a separate thing.
- TextTask's Storybook Story now has the `&` text tag.
- Code cleaned up, notably removing those unused startTag/endTag in the literalInsertion code.

eatyourgreens (approved):

eatyourgreens:
:shipit:

snblickhan (approved):
Looks good to me!

PR #2383, opened 2021-08-23
VariableStarViewer client sizes for classification metadata

srallen:
_Please request review from `@zooniverse/frontend` team or an individual member of that team._ 

Package: lib-classifier

Closes #1832 

This adds the storing of the client sizes of the surrounding parent div stored into the `SubjectViewerStore` and then gets added to the classification metadata. 

A demo mode classification can be made with the project and the classification sent inspected to manually confirm this is working now: https://local.zooniverse.org:8080/?project=4878&workflow=16454&env=production&demo=true

# Review Checklist

## General

- [x] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [x] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2387, opened 2021-08-24
Add storing of client sizes for ScatterPlotViewer

srallen:

Package: lib-classifier

Similar to #2383, this adds the storing of the client sizes to the `SubjectViewerStore` and classification metadata. This is for projects that just use the scatter plot on its own. Can be manually tested with this project: https://local.zooniverse.org:8080/?project=14770&workflow=17386&env=production. If you use the demo mode, then make a classification, you can check the POSTed classification metadata for the expected subject dimensions metadata. The client sizes should be defined. Natural sizes are zero because it's a vectored SVG plot. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


shaunanoordin:
Code read looks really good and straightforward 👍 I'm running tests shortly.

- I'm assuming we're not worrying about if/when the browser window is resized in the middle of the classification. If I'm reading this right, the size will correct itself when the next Subject loads anyway. I'll test for this to confirm.
- _Natural sizes are zero because it's a vectored SVG plot._ - Noted 👌

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR adds the client width/height to the Classification metadata, when submitting classifications for the Scatter Plot Viewer.

- Specifically, `clientWidth` and `clientHeight` (of the viewer's outer `<svg>`) are added to `classification.metadata`
- The recorded height and width of the viewer are set **when each Subject loads**
  - This means that we're not recording the height/width changes if the browser window is resized mid-classification, but the measurements will correct themselves when the next Subject comes in.
  - (This is good, btw - it doesn't make more or less sense than recording the size when the _classification is submitted,_ and tracking every single size change would be mad. Just worth noting here, because I can't remember at which point PFE takes a snapshot of the clientWidth/Height)

Code read looks good, functionality checks out 👍 

### Testing

Tested on localhost with [the provided testing URL,](https://local.zooniverse.org:8080/?project=14770&workflow=17386&env=production&demo=true) in demo mode, using macOS 11.5.2 + Chrome 92

- Browser window was set to a certain size, and reloaded page to ensure a fresh "snapshot" of the clientWidth/Height are taken as the Subject loads
- Dimensions of viewer's `<svg>` is noted (via Inspect Element).
- Answered "Yes" or "No" then Done to submit.
- Checked submitted classification to confirm `classification.metadata.subject_dimensions[0].clientWidth/Height` match the noted dimensions above.
- Additional test 1: continued submitting 2 more times to test repeated results.
- Additional test 2: Resized in the middle of classification to confirm that only the size on subject load is recorded.

### Status

LGTM! 👍

PR #2402, opened 2021-09-02
Track already seen subjects during session

### LINKED ISSUE #2111: Already seen banners not showing during session
## Describe the bug

Originally reported in #1978

> The "You've Seen Everything" banner doesn't appear automatically -- e.g. I classified all 5 subjects in the set, and was given one again with newly-greyed-out lines but without the "Finished" banner. Banner appears if I refresh the page.

## Expected behavior

The banner should show up correctly during the session without page refresh.

srallen:

Package: lib-classifier, app-project

Closes #2111 

This is an unfinished feature in the FEM classifier. PFE tracks seen subjects during the session regardless of the state that is returned from Panoptes (here are the original utility functions: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/lib/seen-this-session.coffee). The feature is that on successful classification submission, an in memory array of string id pairs (workflow id and subject id) are stored and then checked against on subsequent subject loads. Note that this temporary store is *not* recorded in the classification metadata; only the serialized state from Panoptes on the subject is: https://github.com/zooniverse/Panoptes-Front-End/blob/45399f3267f45fa1d67d5bcd162b4050ed11e5ac/app/classifier/classifier.jsx#L259-L282

When implementing this on the FEM classifier, I enhanced this to use browser session storage since volunteers can browse between this project and talk, so we need to reliably load up the already seen state for the session between apps. 

Not displaying the already seen state for subjects has been an issues for subjects that are very quickly classified or very quickly returned to from Talk. The two scenarios are thus:

- A project has a very small subject that someone can get through quickly and Panoptes hasn't had enough time to mark the subjects as being seen before they're loaded up again. This is only really apparent on development or test projects or certain education only use projects. 
- Someone using Done & Talk and clicking to go back on a project that uses ordered subjects and the subject the user just classified is first in queue. For whatever reason the downstream services haven't marked it yet as seen yet (and this is not a fix for this!). In addition, since the subject set is ordered, you're guaranteed to get the subject you just saw if it hasn't been filtered out yet by its already seen status.

We will focus on the latter since this is the reported issue experienced by actual volunteers. To reliably simulate this, load up the HMS project and use demo mode while signed in: https://localhost.zooniverse.org:3000/projects/msalmon/hms-nhs-the-nautical-health-service/classify/workflow/18612/subject-set/83581?env=production&demo=true

You'll initially see a subject with the ordered subject progress banner:
Screen Shot 2021-09-02 at 4 29 45 PM

Click through to simulate submitting a classification and use Done & Talk. This will navigate you away to a 404 page as expected since we don't have Talk working for staging of the new app. Use the browser back button and you'll get the already seen message with the same subject loaded:

Screen Shot 2021-09-02 at 4 30 13 PM

A follow up to this would be to filter out the already seen subject in the queue if it's been stored in the session storage, but this PR is big enough already.

A side effect of this addition is that I did refactor the tests and stories for `SubjectSetProgressBanner` and a mock store utility function. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


shaunanoordin:
Code read looks good.

Main change is that instead of using `subject.already_seen` straight off the Subject resource, we have `.alreadySeen` which checks if already_seen is on the Subject resource OR if the _workflow + subject ID_ is stored in the _session_ storage. 👍 

- The way that the _workflow + subject ID_ token is stored into/retrieved from session storage looks legit to me.
- Clearing the session storage when user logs out also looks good.
- I remember an [old PFE issue](https://github.com/zooniverse/Panoptes-Front-End/issues/2058) where (old) Safari in Private mode just _completely crashes_ due to session storage. (You _can_ see `window.sessionStorage.setItem()`, but actually _trying to run the function_ throws an error.) Fortunately, a quick check in (new) Safari 14.1.2's Private mode + dev console shows that simple commands like `window.sessionStorage.setItem('name', 'Shaun')` no longer causes that crash. Rest assured, I'll still run full tests on Safari Private mode to be safe

Speaking of which, testing go!

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR adds "local memory" for "already seen" Subjects.

- Prior to this PR, whether or not a Subject is "already seen" by a user (and therefore visibly marked by a "you've seen this already" banner) is determined entirely by the Subject resource. (i.e. `subject.already_seen`)
- With this PR, we have an 'queue' in window.sessionStorage that tracks SubjectIDs that a user has seen (specifically, _"workflow ID + Subject ID"_ , since different workflows may share Subjects)
- This means the "already seen" logic no longer requires a Subject queue refresh to be accurate for the user experience. This is most notable for projects with small Subject Set sizes (e.g. < 10 Subjects per Set)
Code read looks good, and functionality checks out 👍 

### Testing

Tested on `app-package`, on localhost, using macOS 11.5.2 + Chrome 92 (Incognito) and Safari 14 (Private)

**Scenario 1: Testing for local memory**

- Test URL: https://local.zooniverse.org:3000/projects/darkeshard/transformers/classify/workflow/3262 (This is a simple Workflow with only 4 Subjects)
- As an anonymous user, classify all Subjects in the WF.
  - (Note: as an anonymous user, it seems that every Subject will come back as `already_seen = false`)
- Result: after the 4th Subject, I start seeing the "you've already seen this Subject" banner, as expected. 👍 This confirms local "already seen" memory

**Scenario 2: Testing for parity**

- Close the window and start afresh.
- Log in as user `darke_shard` (or any user who has already seen all 4 Subjects) and open the test URL again
- Result: the first Subject already has the "you've already seen this Subject" banner. 👍 This confirms that the `subject.already_seen` flag is still respected.

**Scenario 3: Testing for resets**

- Continue for Scenario 2
- Proceed to log out.
- Result: `subjectsSeenThisSession` is cleared from sessionStorage, as expected. 👍 

**Advanced: hacking sessionStorage**

- Repeat Scenario 1, except...
- In the dev console, add or remove Subject IDs to the `subjectsSeenThisSession` value. The "already seen" banner will appear/disappear as expected for the specific Subject IDs. 👍 
- ⚠️ Note: a corrupted `subjectsSeenThisSession` value (i.e. anything that can't be JSON-parsed into an array of strings) _will crash the classifier,_ but this is unlikely to happen in normal user interactions.

**Special note: Safari in Private mode**

- No issues found in Safari 14 in Private mode! 🎉 
- As mentioned in an earlier comment, earlier versions of Safari would crash altogether if you try to run `sessionStorage.setItem()`, but this seems to no longer be an issue.
- And since FEM no longer supports earlier versions of Safari anyway, there should be nothing to worry about for this PR breaking things.

### Dev Notes

- We fully expect some quirks when the user logs in and out in the middle of Classification, simply because logging in/out _currently does not immediately reset the subject queue._
  - e.g. if a user logs out and becomes anonymous, the remaining Subjects already in the queue will retain the already_seen values for the logged-in user identity.
- It's possible for a logged in user to amass a very large queue of "already seen" Subject IDs in the session storage, but I'm genuinely not sure how to test for this in a sane way, nor the limit before this remotely becomes a problem.
  - Only 2 ways this could theoretically be a problem is if the array is so large it takes a long time to search for Subject IDs, or if it's so large that it exceeds the generous ~5MB limit of session storage.
  - I'm filing this under 🤷 (shrug) category for now.

### Status

LGTM 👍 This PR functions as advertised.

eatyourgreens:
>A follow up to this would be to filter out the already seen subject in the queue if it's been stored in the session storage, but this PR is big enough already.

There's a minor complication here, in that we can't filter the queue for all workflows automatically. In projects like Scarlets & Blues, where you can see every subject, already seen or retired subjects should remain in the queue. Similar to what I'm finding in #2392, the subject queue behaviour will depend on `workflow.hasIndexedSubjects`.

Sarah's out today, so shall we merge this and get the fix out onto staging?

I'd like to keep #2111 open, because HMS NHS and Davy Notebooks will still be broken if returning from Talk shows you an Already Seen subject that you can't skip past without classifying it.

packages/lib-classifier/src/components/Classifier/components/Banners/components/SubjectSetProgressBanner/SubjectSetProgressBanner.stories.js
```diff
@@ -1,79 +1,171 @@
...
-    workflows: {
-      resources: {
-        [workflowSnapshot.id]: workflowSnapshot
-      }
-    }
-  })
-  const subject = store.subjects.active
-  const workflow = store.workflows.resources.get(workflowSnapshot.id)
-  return { subject, workflow }
+  const store = mockStore({ subject: subjectSnapshot, subjectSet: subjectSetSnapshot, workflow: workflowSnapshot })
```
eatyourgreens, line 21: Nice!

packages/lib-classifier/src/components/Classifier/components/Banners/components/SubjectSetProgressBanner/SubjectSetProgressBanner.stories.js
```diff
@@ -1,79 +1,171 @@
...
+    >
+      <Provider classifierStore={store}>
+        <Box background={{ dark: 'dark-3', light: 'light-3' }} width='large'>
+          <Banners />
+          <img src="https://placekitten.com/800/400" alt='placeholder' />
+        </Box>
+      </Provider>
+    </Grommet>
+  )
+}
```
eatyourgreens, line 142: Does this duplicate the standalone story for the retired subject banner?

I've been wondering, after working on the other PRs for banners, if we should move the tests and stories up one level, and just have a set of tests and stories which render the `Banners` component for different combinations of subject and workflow, like @srallen has done here.

shaunanoordin:
> Sarah's out today, so shall we merge this and get the fix out onto staging?

Sounds good to me. 👍  I'll go ahead and merge since PR's safe & functional; any improvements and feedback responses can be done in a separate branch.

Issue #1305, opened 2019-11-14
Add survey task feedback strategy


rogerhutchings:
Will allow us to deprecate the Gravity Spy gold standard experimental feature in PFE.

Related issue: #1274

srallen:
@mcbouslog I think this can close, yes?

PR #2403, opened 2021-09-02
Add survey task feedback

mcbouslog:
Package:
- lib-classifier

Describe your changes:
- as per ADR 33 adds feedback for survey task

Staging:
- https://local.zooniverse.org:8080/?project=1810&workflow=3516&demo=true (click subject metadata for "correct" answer, i.e. duplicated feedback columns without hidden tag)

Questions
- what should the strategy name be? currently I used "Survey", but if we know there'll be more complex versions (comparing choice answers) then maybe it should be something like "Survey Simple"/`surveySimple`?
- open to any variable renaming, particularly in the reducer

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/store/feedback/strategies/survey/index.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  it('should have a `id` property', function () {
+    expect(index.id).to.equal('survey')
+  })
+
+  it('should have a `title` property', function () {
+    expect(index.title).to.equal('Survey')
+  })
+
+  it('should have a `labComponent` property', function () {
+    expect(index.labComponent).to.equal(null)
```
srallen, line 21: Non-blocking. This can use `null()` 
```suggestion
    expect(index.labComponent).to.be.null()
```

packages/lib-classifier/src/store/feedback/strategies/survey/index.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  })
+
+  it('should have a `validations` property', function () {
+    expect(index.validations).to.equal(null)
```
srallen, line 25:
```suggestion
    expect(index.validations).to.be.null()
```

packages/lib-classifier/src/store/feedback/strategies/survey/reducer.js
```diff
@@ -0,0 +1,22 @@
+function surveyReducer (rule, choices = []) {
+  const ruleIds = rule.choiceIds
+    .split(',')
+    .map(idString => idString.replace(/\W/g, '').toUpperCase())
+
+  const chosenIds = choices.map(choice => choice.choice)
```
srallen, line 6: Maybe just a bit of clarity here:
```suggestion
  const chosenIds = choices.map(choiceObject => choiceObject.choice)
```

srallen (changes requested):
I have a few minor syntax suggestions, but otherwise this is looking great. I'm submitted this as requesting changes because I do think we should name the strategy both in code and in the docs as "Survey: Simple" or `surveySimple` just in case we do get asked to setup a strategy in the future that evaluates the choice questions as well. 

I ran through the test project's subjects and submitted a variety of success and failure annotations and each evaluated as expected. Thanks for setting this up!

srallen (approved):
LGTM. Thanks!

mcbouslog:
Closes #1305

PR #2411, opened 2021-09-13
Engaging Crowds: Add annotation._inProgress

eatyourgreens:
Add the `annotation._inProgress` flag, which is false by default, but true once an annotation has been updated by the classifier.
Add some generic tests which create and update an annotation for each registered task type.

For Engaging Crowds, we'd like to be able to detect when a volunteer has begun to work on a subject, so that we can warn them before clearing their classification when changing to a new subject. `annotation._inProgress` here should hopefully provide a flag that we can use to detect when classification work has begun.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/index.spec.js
+++ b/packages/lib-classifier/src/plugins/tasks/index.spec.js
@@ -1,0 +1,57 @@
+import taskRegistry from './'
+
+describe('Task models', function () {
+  const taskTypes = Object.keys(taskRegistry.register)
+  taskTypes.forEach(function (taskType) {
+    let task
+    describe(taskType, function () {
+      before(function () {
+        const taskSnapshot = {
+          answers: [],
+          instruction: `${taskType} instructions`,
+          options: [ '1', '2', '3', '4' ],
+          question: `${taskType} question`,
+          taskKey: 'init',
+          type: taskType
+        }
+        const { TaskModel } = taskRegistry.get(taskType)
+        task = TaskModel.create(taskSnapshot)
+      })
+
+      it('should exist', function () {
+        expect(task).to.be.ok()
+      })
+
+      describe('annotations', function () {
+        let annotation
+
+        before(function () {
+          annotation = task.createAnnotation()
+        })
+
+        it('should exist', function () {
+          expect(annotation).to.be.ok()
+        })
+
+        it('should store the task key', function () {
+          expect(annotation.task).to.equal(task.taskKey)
+        })
+
+        it('should store the task type', function () {
+          expect(annotation.taskType).to.equal(taskType)
+        })
+
+        it('should not be in progress', function () {
+          expect(annotation._inProgress).to.be.false()
+        })
+
+        describe('on update', function () {
+          it('should be marked in progress', function () {
+            annotation.update(annotation.value)
+            expect(annotation._inProgress).to.be.true()
+          })
+        })
+      })
+    })
+  })
+})
--- a/packages/lib-classifier/src/plugins/tasks/models/Annotation.js
+++ b/packages/lib-classifier/src/plugins/tasks/models/Annotation.js
@@ -18,6 +18,9 @@ const Annotation = types.model('Annotation', {
     delete newSnapshot.id
     return newSnapshot
   })
+  .volatile(self => ({
+    _inProgress: false
+  }))
   .views(self => ({
     get isComplete () {
       return true
@@ -30,6 +33,7 @@ const Annotation = types.model('Annotation', {
   .actions(self => ({
     update (value) {
       self.value = value
+      self._inProgress = true
     }
   }))
 


shaunanoordin:
PR Review in progress. Code read looks good, quick testing notes:

- Survey Tasks trigger `annotation.update()` when a species (plus behaviour, plus number, plus etc) is selected 👍 
- Question Tasks trigger `annotation.update()` when an answer is selected 👍
- Text Tasks trigger `annotation.update()` when anything is typed 👍
- [Subject Group Comparison Tasks](https://local.zooniverse.org:8080/?env=staging&project=1900&workflow=3412&demo=true) trigger `annotation.update()` trigger `annotation.update()` when any cell is clicked
- [Multi-branch workflows with Question Tasks](https://local.zooniverse.org:8080/?env=staging&project=1873&workflow=3467&demo=true) work fine. 👍 
  - When going 'Back' to previous steps, the ._inProgress value is correctly preserved. 
  - When going 'Back' and then _switching to different branches,_ "formerly future step" annotations are correctly reset.
- Drawing Tasks only trigger `annotation.update()` (and hence update ._inProgress) when proceeding to the next Step/Task, or when submitting. 🤔❓
  - I'm going to take a peek into this.

Review ongoing, brb.

shaunanoordin:
Further quick testing notes:

- Drawing Tasks are making me scratch my head. _(But this is quickly spiralling out into its own investigation, so take the following notes in moderation!)_
  - Ah, OK, the Drawing Task only calls annotation.update() on the complete() step, gotcha.
  - I'm fairly sure we can link `annotation._inProgress = true` to `createMark()` (or some other activity drawing-specific activity), but I can't find the connection between the mark and the annotation because...
  - ❓ ...is `models/DrawingTask.js`'s `Drawing.annotation` actually doing anything? I actually deleted line 17, `annotation: types.safeReference(DrawingAnnotation),` and I was still able to successfully draw marks, see the number of marks listed on the Drawing Tasks component, and submit the classification with the correct annotations
- I can't seem to find Dropdown Tasks to test. I won't let that hold up this PR, but that's something I need to follow up on.
- I'm not going to test for other special Tasks, e.g. Light Curve Viewer, to keep the scope of this PR manageable.

eatyourgreens:
🤔hmmm, the primary target of this work will be transcription projects, and the transcription task is a drawing task.

shaunanoordin (commented):
## PR Review

This PR adds a new `._inProgress` value to Annotations.

- Context: we're starting to want to know when a user has _started making annotations for a task._ This is so we can warn the user when they might _lose any classifications in progress_ , e.g. if they're trying to navigate to another workflow/subject set/subject.
- Implementation is simple but effective - any time `annotation.update()` is called, we mark a small boolean called _inProgress. 👌 

Code read looks good, but testing revealed some missing coverage.

### Testing

Tested on localhost with macOS + Chrome, as usual. Testing steps are very simple:
- add a `console.log()` to `annotation.update()` to monitor when `self._inProgress` triggers
- find WFs for each Task type. For each Task, make an annotation (e.g. select an answer for a Question Task, type some text in a Text Task) and observe that the `console.log()` triggers as expected.

Here's our coverage for each task (i.e. whether or not we can tell a user has started making annotations for each task), as of 1539bce.

- ✔️ **Question** Tasks updates when an answer is selected
- ✔️ **Text** Tasks update when anything is typed
- ❌ **Drawing** Tasks only update when complete() is triggered (i.e. when moving to next Task or submitting)
- ❓ **Transcription** Tasks is a bit of a mixed bag.
  - Drawing a line doesn't update annotations.
  - Typing in some text DOES update annotation. (Note that this is the sub-task's annotation)
- ✔️ **Survey** Tasks update when a species (plus behaviour, plus number, plus etc) is selected
- ✔️ **Subject Group Comparison** Tasks update when any cell is clicked
- Unknown: **Dropdown** Tasks, **DataVisAnntotation** Tasks
- Side note: going back to previous Tasks and going to new branches in branching workflows will update annotations as expected

### Status

UPDATE: I was going to go "good enough, we can worry about Drawing Tasks later" for this PR, but as this comment pointed out...

> 🤔hmmm, the primary target of this work will be transcription projects, and the transcription task is a drawing task.

...gah, you're right, this needs to be sorted out!

Things I'm going to follow up on tomorrow:

- How are Drawing Task annotations organised?
  - Is each mark an annotation in an of itself?
  - Shouldn't there be a master "Drawing Task", which contains tools, which contain marks? How are they linked? Study the store schema.
  - There should be a way to link `createMark()` with the master "Drawing Task" annotation
- ~~How do sub-tasks' inProgress values affect the parent tasks~~ EDIT: nvm, I just realised we don't really care if a sub-task's annotation is _inProgress_ or not, we just want to know if any of the top-level tasks are in progress.

shaunanoordin:
4:22am ideas: I might be overthinking it previoisly. A DrawingAnnotation can easily tell if it's in progress by checking to see if DrawingAnnotation.value.length > 0

The catch here is that if a user creates an annotation and then deletes it, the code thinks they never started. Hmm will need to think about this tomorrow after I zzzzzz

eatyourgreens:
I think this is about where I got to, lst week, while trying to think through a generic method for figuring out when you'd started classifying for any type of task.

Tasks like questions and dropdown menus call `annotation.update(value)` on change, in the component code.
https://github.com/zooniverse/front-end-monorepo/blob/031f7eb3b8beabaff65a934d2dc24785c8fcf2c6/packages/lib-classifier/src/plugins/tasks/SimpleDropdownTask/components/SimpleDropdownTask.js#L66-L71

We could refactor drawing tasks so that creating a mark also updates the drawing annotation:
https://github.com/zooniverse/front-end-monorepo/blob/031f7eb3b8beabaff65a934d2dc24785c8fcf2c6/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js#L84-L99
Perhaps in a separate PR, so that we can test whether that breaks anything?

eatyourgreens:
@shaunanoordin I'm going to work on rewriting the drawing tasks to update annotations incremementally, like other tasks. If that works, then I think we can use the approach here to detect drawing in progress, by flagging calls to `createMark`.

eatyourgreens:
If #2415 works, then that would set `annotation._inProgress` to true the first time that you draw something, which would allow us to detect when transcribers have started work on a transcription task.

eatyourgreens:
If you look at the changes, you'll see that the annotation is only updated when a mark is created. Drawing annotations are an array of mark ID strings, so values are only updated in the Tool models. 

There won't be continuous calls to annotation.update while drawing is taking place.

shaunanoordin:
## PR Review (Update)

The previous review for this PR was looking good _except for_ the Drawing Tasks, which was missing coverage.

Good news, #2415 addresses the Drawing Tasks, and that PR's now approved. Once 2415's merged and this branch is updated accordingly, I'll re-test this PR. 👌

eatyourgreens:
This is rebased to include #2415 now. `annotation._inProgress` should be true for any task, after you've started work.

shaunanoordin (approved):
## PR Review

Testing looks good, this PR is now good to go!

### Testing

Follow up from [earlier tests:](https://github.com/zooniverse/front-end-monorepo/pull/2411#pullrequestreview-753240752)

- ✔️ Question Tasks updates when an answer is selected
- ✔️ Text Tasks update when anything is typed
- 🆕 ✔️ Drawing Tasks update when a new mark is created
- 🆕 ✔️ Transcription Tasks update when...
  - drawing the start point of a line
  - typing in some text for the sub-task
  - choosing any of the single answer to the "is it done?" question
- ✔️ Survey Tasks update when a species (plus behaviour, plus number, plus etc) is selected

### Status

LGTM 👍

PR #2415, opened 2021-09-15
Classifier: Refactor drawing task annotations

eatyourgreens:
Pass the active drawing task annotation down to the `InteractionLayer` components as a prop. Update the annotation incrementally, when a new mark is created, rather than waiting until drawing is complete for the current step.

Here's a test project with a couple of drawing steps, including some subtasks.
https://localhost:8080/?project=908&workflow=3370

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
@@ -21,14 +21,15 @@ const DrawingCanvas = styled('rect')`
 function InteractionLayer({
   activeMark,
   activeTool,
-  activeToolIndex,
-  disabled,
-  frame,
+  activeToolIndex = 0,
+  annotation,
+  disabled = false,
+  frame = 0,
   height,
-  marks,
+  marks = [],
   move,
-  setActiveMark,
-  scale,
+  setActiveMark = () => {},
+  scale = 1,
   width,
   played,
   duration
@@ -96,6 +97,8 @@ function InteractionLayer({
     mark.setSubTaskVisibility(false)
     // Add a time value for tools that care about time. For most tools, this value is ignored.
     mark.setVideoTime(timeStamp, duration)
+    const markIDs = marks.map(mark => mark.id)
+    annotation.update([ ...markIDs, mark.id ])
   }
 
   function onPointerDown(event) {
@@ -185,6 +188,11 @@ InteractionLayer.propTypes = {
   activeMark: PropTypes.object,
   activeTool: PropTypes.object.isRequired,
   activeToolIndex: PropTypes.number,
+  annotation: PropTypes.shape({
+    task: PropTypes.string,
+    taskType: PropTypes.string,
+    value: PropTypes.array
+  }).isRequired,
   frame: PropTypes.number,
   marks: PropTypes.array,
   setActiveMark: PropTypes.func,
@@ -192,16 +200,6 @@ InteractionLayer.propTypes = {
   disabled: PropTypes.bool,
   scale: PropTypes.number,
   width: PropTypes.number.isRequired
-}
-
-InteractionLayer.defaultProps = {
-  activeMark: undefined,
-  activeToolIndex: 0,
-  frame: 0,
-  marks: [],
-  setActiveMark: () => {},
-  disabled: false,
-  scale: 1
 }
 
 export default InteractionLayer
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.spec.js
@@ -46,6 +46,7 @@ describe('Component > InteractionLayer', function () {
         ],
         type: 'drawing'
       })
+      const annotation = mockDrawingTask.createAnnotation()
       const setActiveMarkStub = sinon.stub().callsFake(() => mockMark)
       activeTool = mockDrawingTask.activeTool
       sinon.stub(activeTool, 'createMark').callsFake(() => mockMark)
@@ -53,6 +54,7 @@ describe('Component > InteractionLayer', function () {
         <InteractionLayer
           activeMark={mockMark}
           activeTool={activeTool}
+          annotation={annotation}
           frame={2}
           setActiveMark={setActiveMarkStub}
           height={400}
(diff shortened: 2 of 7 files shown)


eatyourgreens:
If this works, then we can use `annotation.update` to detect when a transcriber has begun work on a transcription subject, which will help with #2411.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -194,15 +202,5 @@ InteractionLayer.propTypes = {
   width: PropTypes.number.isRequired
 }
 
-InteractionLayer.defaultProps = {
```
eatyourgreens, line 197: This isn't strictly required for this PR, but [`defaultProps` is being deprecated for functions](https://github.com/facebook/react/pull/16210), so it seemed like a good opportunity to remove it.

eatyourgreens:
This could be a good opportunity to rewrite `InteractionLayerContainer` to use [`withStores`](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/helpers/withStores), and remove `inject` from drawing tasks too.

srallen:
I'm recommending an alternative to be considered because of potential performance impact: https://github.com/zooniverse/front-end-monorepo/pull/2411#issuecomment-920075116

shaunanoordin (commented):
## PR Review

This PR is a refactor of Drawing Annotations.

- Functionally, **we expect no changes to user interactions with Drawing Tasks,** nor with the Drawing Annotations submitted in the classification.
- The biggest change is that for Drawing Tasks, the `annotation.update()` call will be moved from `DrawingTask.complete()` to `InteractionLayer.createMark()`
- Context: starting with #2411, we want to know when a user starts interacting with the Classifier, and we're doing that by checking when annotation.update() is called. Unfortunately, at the moment, you can't tell when a DrawingTask has been .update()ed since that's only called when it's complete.

Code read looks good, but...

### Testing

...testing has not started yet. I have a testing plan, though:

- What happens when you submit a drawing Task with 0 Marks?
  - ...with 1 Mark?
  - ...with 3 Marks?
- What happens when you make 3 Marks, then delete 1?
- Will Marks be updated after they're edited? (e.g. place a mark, then drag it sideways)
  - (Note to self: if this is a problem - i.e. drawing mark data isn't updating after an editing action, check to see if simply re-introducing annotation.update() back to DrawingTask.complete() solves the issue.)

Sarah raised an issue about potential performance problems. Based on the code, this shouldn't be an issue, since the change only causes `annotation.update()` to trigger the first time a user clicks on the InteractionLayer. To be safe, I'm going to trace the function calls to determine how often the annotations are actually updated.

### Status

Testing in progress.

CCing @ErikOstlund as an FYI. Erik, I know that you're working on the Freehand Drawing Tool, and I want to help make sure this PR works well with the code you build. Let me know if you've got a dev branch I can _locally rebase_ on top of this PR, for testing.

shaunanoordin (approved):
## PR Review (Continued)

Package: `lib-classifier`

Testing is now complete, and everything looks good 👍 

### Testing

Tested on localhost with macOS 11.5.2 + Chrome 93

Setup:
- Test URL: https://local.zooniverse.org:8080/?env=staging&project=1873&workflow=3520&demo=true
  - This is a Workflow with 1 Drawing Task, with 3 tools: [ Point, Rectangle (with 1 Text Sub-Task), Line] 
  - WF has 1 Subject, an image of a 1000x1000 pixel grid for precision coordinate testing.
- Annotation.update() modified to provide console logs.

Testing steps:
- On the Classifier, make 0 annotations, then submit. Check submitted classification. 🆗 
- Make 1 Point mark, then submit. Check submitted classification. 🆗 
  - Classification submitted with the correct coordinates.
  - (For brevity, assume checking submitted classification is implicit for all steps.)
- Make 3 Point marks, then submit. 🆗
- Make 1 Point mark, delete it, then submit. 🆗
- Make 3 Point marks, delete 2, then create 1 more. Submit. 🆗 
  - Deleted marks are also deleted from the annotations! 👍
  - I'm glad to see this works, since I was initially wary that `annotation.update()` only triggered when the mark is created.
- Make 1 Point mark, then edit the mark (i.e. by moving it to a new coordinate). Submit. 🆗 
  - Edited marks are also edited in the annotations! 👍 
- Repeat steps above with Rectangle marks (with Text sub-task)
- Repeat steps above with Line marks.

Everything looks good!

Not tested:
- Multi-step workflows. It's safe to assume there shouldn't be much of an effect here.

### Dev Notes

From the testing, I can confirm that the annotations submitted with the classifications are correct. 👍 As far as the refactor is concerned, user-side functionality is preserved. This is looking good.

One thing to note is that `annotation.update()` **isn't** called when a Drawing Mark is either **deleted** or **edited**. This is fine for this PR (as noted, tests show that everything works fine) but I'm wondering if there is any "developer assumption" that .update() is (or """should""") be called on editing/deleting, similar to what happens with other tasks. e.g. TextTask calls `annotation.update()` on every key press. (i.e. onChange)

Devil's advocate: updating on every key press seems _very performance un-optimised,_ so maybe TextTask isn't the best example to follow.

Devil's advocate pt 2: attempting to call annotation.update() on every drawing mark _edit action_ will almost certainly incur the performance costs that Sarah was warning about.

Recommendation: do nothing for now, the code is fine as-is.

### Status

Approved, though I'd love to follow up testing with Erik's Freehand Tool.

eatyourgreens:
>One thing to note is that annotation.update() isn't called when a Drawing Mark is either deleted or edited. This is fine for this PR (as noted, tests show that everything works fine) but I'm wondering if there is any "developer assumption" that .update() is (or """should""") be called on editing/deleting, similar to what happens with other tasks. e.g. TextTask calls annotation.update() on every key press. (i.e. onChange)

Drawing task annotations are arrays of mark IDs, references to the actual marks which are stored with their respective tools. Deleting a mark from the store automatically removes all references too, so the annotation automatically updates when you delete a mark from a tool. Likewise, editing a mark on a tool updates the annotation, because the annotation holds a pointer to the stored mark.

`TextTask`, I think, is a good example of what we found happening with freehand drawing. Imagine a text task if the value had ~1,000 characters. Every keystroke would manipulate a large string, in memory, to change one character.

The performance heavy code for drawing tasks is in `annotation.toSnapshot()`, which turns the annotation into a flattened array of marks. This code only runs once, when you press Done and complete a classification. @srallen is right that this would be prohibitive to run on every edit, so it was designed to only run once, when you've finished work on a subject.
https://github.com/zooniverse/front-end-monorepo/blob/690b3aef60a8a24e3597ebc58c37508cc98d7880/packages/lib-classifier/src/plugins/tasks/DrawingTask/models/DrawingAnnotation.js#L14-L39

PR #2417, opened 2021-09-15
Revert "Temporarily move about pages to preview route"

srallen:
Reverts zooniverse/front-end-monorepo#2407

Chris gave a verbal approval of the new about pages, so let's revert the temporary route.


goplayoutside3 (approved):
Looks good!

PR #2470, opened 2021-10-01
Engaging Crowds: add confirmation dialog for Next/Prev navigation

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Project: Engaging Crowds

Context: Engaging Crowds introduces the ability for users to _select individual Subjects_ in an _indexed, organised Subject Set._ (e.g. Subject Set = a book, Subject = pages, in order) This feature also allows a user to go to the Previous or Next Subject in the indexed set (similar to reading a book). However, there's a possibility that a user might inadvertently move to the Prev/Next Subject while they're still working on the current Subject, not realising they might lose any work in progress.

The intent of this PR is to add a confirmation dialog, prompting users "Are you sure you want to navigate to the Next/Previous Subject?" IF they've already started annotating the current Subject.

Components:
- [ ] Add a mechanism in the annotations (AnnotatedSteps, actually) to check if any there is ANY annotation with their `._inProgress` value set to true.
- [ ] Connect that "checkForProgress" mechanism up to the `SubjectProgressBanner` (which contains the Next/Prev nav buttons, and only displays for grouped WFs with indexed Subject Sets)
- [ ] Modify the Next/Prev nav button logic to check the "checkForProgress" mechanism
- [ ] Add the "Are you SURE??" modal

Intended Behaviour:
- Applicable only for grouped WFs with indexed Subject Sets which allow single Subject selection
- On the Classifier, with a Subject selected:
  - If user has NOT started making any annotations (current step, previous step, or next step, whatever), clicking the Next or Previous nav button should take the user to the next/previous Subject immediately.
  - If user HAS started making any notations, clicking the Next or Previous nav button should open the "Are you sure?" modal
    - If user says yes, the modal closes and then navigation happens as intended.
    - If user says no, the modal closes with no further action.

### Testing

Currently testing on an EC project that has Indexed Subject Sets and Individual Subject Selection enabled, i.e. Scarlet and Blues. Please be aware that this is a rather unique setup with several external dependencies, so describing the full process of replicating that project + external dependencies for "standalone testing" is beyond the scope of this PR.

Basically, just click this link to test, and hope it's still alive when you're reading this: https://frontend.preview.zooniverse.org/projects/bogden/scarlets-and-blues/classify/workflow/18505/subject-set/98217/subject/68961909?env=production

### Status

WIP

### DIFF
--- a/packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.js
+++ b/packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.js
@@ -133,11 +133,19 @@ const AnnotatedSteps = types.model('AnnotatedSteps', {
     })
   }
 
+  /** Checks if the user has started making any annotations. Returns boolean */
+  function checkForProgress() {
+    // TODO
+    return Math.random() <= 0.5
+  }
+
+
   return {
     back,
     finish,
     next,
-    start
+    start,
+    checkForProgress
   }
 })
 
--- a/packages/lib-classifier/src/components/Classifier/components/Banners/Banners.js
+++ b/packages/lib-classifier/src/components/Classifier/components/Banners/Banners.js
@@ -10,7 +10,7 @@ import UserHasFinishedWorkflowBanner from './components/UserHasFinishedWorkflowB
 function useStores(stores) {
   const { classifierStore } = stores ?? React.useContext(MobXProviderContext)
   return {
-    annotations: classifierStore.annotations,
+    annotatedSteps: classifierStore.annotatedSteps,
     subject: classifierStore.subjects.active,
     subjects: classifierStore.subjects,
     workflow: classifierStore.workflows.active
@@ -20,14 +20,14 @@ function useStores(stores) {
 const environment = process.env.APP_ENV
 
 function Banners({ stores }) {
-  const { annotations, subject, subjects, workflow } = useStores(stores)
+  const { annotatedSteps, subject, subjects, workflow } = useStores(stores)
   const subjectNumber = subject?.priority ?? -1
   const hasIndexedSubjects = workflow?.hasIndexedSubjects
   const hasGroupedOrderedSubjects = workflow?.grouped && workflow?.prioritized
   if (environment !== 'production' && hasIndexedSubjects && subjectNumber > -1) {
     return (
       <SubjectSetProgressBanner
-        annotations={annotations}
+        checkForProgress={annotatedSteps.checkForProgress}
         onNext={subjects.nextIndexed}
         onPrevious={subjects.previousIndexed}
         subject={subject}
--- a/packages/lib-classifier/src/components/Classifier/components/Banners/Banners.js
+++ b/packages/lib-classifier/src/components/Classifier/components/Banners/Banners.js
@@ -10,6 +10,7 @@ import UserHasFinishedWorkflowBanner from './components/UserHasFinishedWorkflowB
 function useStores(stores) {
   const { classifierStore } = stores ?? React.useContext(MobXProviderContext)
   return {
+    annotations: classifierStore.annotations,
     subject: classifierStore.subjects.active,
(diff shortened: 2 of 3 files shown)


shaunanoordin:
## PR Update

This PR is now functional! Please check Components and Intended Behaviour noted on the previous post, along with the test URL. I'm marking this ready for review, but I'll need to swing back tomorrow to check for tests, storybook, etc.

packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.js
```diff
@@ -133,11 +133,21 @@ const AnnotatedSteps = types.model('AnnotatedSteps', {
     })
   }
 
+  /** Checks if the user has started making any annotations. Returns boolean */
+  function checkForProgress() {
```
eatyourgreens, line 137: This doesn’t change the model state, so it can be a view instead of an action.

packages/lib-classifier/src/store/AnnotatedSteps/AnnotatedSteps.js
```diff
@@ -133,11 +133,21 @@ const AnnotatedSteps = types.model('AnnotatedSteps', {
...
+    let progressFlag = false
+    self.annotations.forEach(annotation => {
+      progressFlag ||= annotation.inProgress
```
eatyourgreens, line 140: I think you can check `annotation._inProgress` directly here, which saves another view.

eatyourgreens (commented):
A couple of quick initial comments. Nothing major.

shaunanoordin:
## PR Update

Thanks for the feedback, Jim!

- `intent` now uses a function callback, instead of a "next"/"previous" string with if-else statements
  - Curiously, I learnt that to set a function as a state value, you need to use `setMyState(() => myFunction)`, not `setMyState(myFunction)`. Doing the latter will instead immediately trigger myFunction()
- `<ConfirmModal>` is now an actual _modal_
  - This prevents the user from interacting with other bits of the classifier, e.g. clicking the Prev/Next buttons while the modal is open.
  - ❔ Ahhah, but there's still a small accessibility issue. The ConfirmModal does NOT automatically gain focus, even though it has `ConfirmModal.GrommetModal.active = true`. When the modal loads, the focus goes to an invisible `<a>` anchor that's part of the GrommetModal instead. Maybe I need to explicitly spell out that the Primary Button should get the focus.
- `Annotation.inProgress` is no longer needed since I just realised I no longer need to a getter function that I can extend for different Annotation data types. `Annotation._inProgress` is used in its place.

eatyourgreens:
>Curiously, I learnt that to set a function as a state value, you need to use setMyState(() => myFunction), not setMyState(myFunction). Doing the latter will instead immediately trigger myFunction()

I forgot about the callback pattern for setting state: `const state = setState(oldState => newState)`. When you pass a function, it runs and the return value is used as the new state. `useRef` might be more relevant here, for storing a ref to the callback function but 🤷‍♂️ .

A pattern similar to this might work but I'm not sure if it's any better than what we've got. The state change triggers a re-render, but with a new value for the stored reference.
```jsx
const [ showModal, setShowModal ] = useState(false);
const confirmCallback = useRef(null);

function deferredNext() {
  confirmCallback.current = onNext;
  setShowModal(true);
}

function deferredPrevious() {
  confirmCallback.current = onPrevious;
  setShowModal(true);
}

function onConfirm() {
 confirmCallback.current();
 confirmCallback.current = null;
 setShowModal(false);
}
return <ConfirmModal active={showModal} onConfirm={onConfirm} />
```

eatyourgreens:
The stories are broken with
```
MobX Provider: The set of provided stores has changed. See: https://github.com/mobxjs/mobx-react#the-set-of-provided-stores-has-changed-error.
```

That's from #2402. The controls should be updating component props, not generating a MobX store when they change.

PR #2481, opened 2021-10-12
Fix: Next and Previous button should only appear when onNext and onPrevious is defined

shaunanoordin:
## PR Overview

Package: lib-classifier

#2470 introduced a bug which causes the "Next" and "Previous" button to ALWAYS appear on the Subject Set Progress Banner, regardless of whether or not it has indexed Subjects.

This was caused by `Banner.onNext` and `Banner.onPrevious` being ALWAYS defined, in turn due to `SubjectSetProgressBanner.onNext/onPrev` now being wrapped in tryToGoNext/Previous. This PR fixes that by adding existence checks for the appropriate functions.

### Status

Testing in progress.


shaunanoordin:
## PR Update

This PR is now ready for review! @eatyourgreens Thanks for your earlier review, but I'm re-requesting the review since I had to put in one additional fix: _I had to remove the default () => {} I added to onNext and onPrevious,_ d'ooohhh

shaunanoordin:
Oh thank goodness they're back in order.

Scarlets & Blues: (has Next/Previous)
image

HMS NHS: (doesn't have Next/Previous)
image

PR #2349, opened 2021-08-10
Refresh user project preferences

srallen:
_Please request review from `@zooniverse/frontend` team or an individual member of that team._ 

Package: app-project

Closes #2205

This is the final piece for UPP prompting. I realized that I was getting ahead of myself if I implemented the web socket idea since the Caesar portion of this feature has yet to be completed and I'm unsure if/how it actually will end up being used. So, I decided that implementing polling is much simpler rather than implementing a web socket connection that would be hard to test and take more time to remove if we end up not doing this. We also already keep track of the classification session count, so this was relatively straight forward. 

What this does is refresh the user project preferences stored every 5 classifications. The prompt is connected to the store and fires an effect when the loading state of the UPP changes, which then calls the function which checks if the assigned workflow is different from the current workflow.

This can be tested manually at: https://local.zooniverse.org:3000/projects/srallen086/workflow-assignment-testing

Sign in as `zootester`, password in passbolt. This user is assigned to workflow id `'2334'`, so has access to the beginner and journey labeled workflows. Pick to start with workflow id `'2333'`. You'll be prompted, select "No thanks". Make five classifications, and you should be prompted again. This time select "take me to the next workflow." You should then be routed to `/workflows/2334`. 

Now go back to the homepage and select the beginner workflow. This time select "Don't show me this again." and "No thanks". Do five classifications, and you won't be prompted. 

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

[...]

PR #2515, opened 2021-11-03
Refactor UPP data-fetching

### LINKED ISSUE #2501: Every 5 classifications, the classifier unmounts and remounts
## Package
app-project

## Describe the bug
https://frontend.preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground/classify/workflow/3223?env=staging

Try submitting some classifications on that test Yes/No workflow. Watch today's session count in the 'Your Stats' box on the Classify page. At 5, 10, 15 classifications etc. the classifier is replaced by a loading icon and completely reloads.

## Expected behavior
`UserProjectReferences` should refresh in the background, every 5 classifications, but the classifier shouldn't completely reload, which forces a reload of the workflow and other resources from Panoptes.

eatyourgreens:
Fetch the full UPP resource when the stored user changes. When refreshing the UPP resource, fetch data in the background and only refresh stored settings, so as to avoid overwriting the stored classification count.

#2349 introduced a couple of undesired side effects for all projects:
- every 5 classifications, the user's `activity_count` is refreshed from Panoptes. `activity_count` is updated every 5 minutes, so this can lead to weird behaviour where your classification count, on the Classify page, can go up, then down, then up again.
- every 5 classifications, the classifier is replaced by a loading icon and reloads, including clearing the stored subject queue and refreshing it from Panoptes.

This PR should fix both those effects, by silently requesting UPP (every 5 classifications) and only refreshing `UPP.settings`.

Package:
app-project

Closes #2501.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
+++ b/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
@@ -45,27 +45,19 @@ const UserProjectPreferences = types
 
   }))
   .actions(self => {
-    const _fetch = flow(function * _fetch() {
+    async function _fetch() {
       const { client, project, user } = getRoot(self)
-      self.setLoadingState(asyncStates.loading)
-      try {
-        const token = yield auth.checkBearerToken()
-        const authorization = `Bearer ${token}`
-        const query = {
-          project_id: project.id,
-          user_id: user.id
-        }
-
-        const response = yield client.panoptes.get('/project_preferences', query, { authorization })
-        const [preferences] = response.body.project_preferences
-        if (preferences) {
-          self.setResource(preferences)
-        }
-        self.setLoadingState(asyncStates.success)
-      } catch (error) {
-        self.handleError(error)
+      const token = await auth.checkBearerToken()
+      const authorization = `Bearer ${token}`
+      const query = {
+        project_id: project.id,
+        user_id: user.id
       }
-    })
+
+      const response = await client.panoptes.get('/project_preferences', query, { authorization })
+      const [preferences] = response.body.project_preferences
+      return preferences
+    }
 
     return {
       reset() {
@@ -97,13 +89,29 @@ const UserProjectPreferences = types
       },
 
       fetchResource: flow(function* fetchResource() {
-        if (!self.id) {
-          yield _fetch()
+        try {
+          if (!self.id) {
+            self.setLoadingState(asyncStates.loading)
+            const preferences = yield _fetch()
+            if (preferences) {
+              self.setResource(preferences)
+            }
+            self.setLoadingState(asyncStates.success)
+          }
+        } catch (error) {
+          self.handleError(error)
         }
       }),
 
       refreshResource: flow(function * refreshResource() {
-        yield _fetch()
+        try {
+          const preferences = yield _fetch()
+          if (preferences) {
+            self.preferences = preferences.preferences
+          }
+        } catch (error) {
+          self.handleError(error)
+        }
       })
     }
   })
--- a/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
+++ b/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
@@ -2,6 +2,7 @@ import { applySnapshot, flow, getRoot, types } from 'mobx-state-tree'
 import auth from 'panoptes-client/lib/auth'
 import asyncStates from '@zooniverse/async-states'
 
+import { logToSentry } from '@helpers/logger'
 import numberString from '@stores/types/numberString'
 
 const Preferences = types
@@ -83,12 +84,6 @@ const UserProjectPreferences = types
         applySnapshot(self, resource)
       },
 
-      handleError(error) {
-        console.error(error)
-        self.error = error
-        self.setLoadingState(asyncStates.error)
-      },
-
       setLoadingState(state) {
         self.loadingState = state
       },
@@ -104,7 +99,10 @@ const UserProjectPreferences = types
             self.setLoadingState(asyncStates.success)
           }
         } catch (error) {
-          self.handleError(error)
+          console.error(error)
+          logToSentry(error)
+          self.error = error
+          self.setLoadingState(asyncStates.error)
         }
       }),
 
@@ -115,7 +113,7 @@ const UserProjectPreferences = types
             self.settings = preferences.settings
           }
         } catch (error) {
-          self.handleError(error)
+          console.error(error)
         }
       })
     }
--- a/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
+++ b/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
@@ -34,12 +34,15 @@ const UserProjectPreferences = types
     settings: types.maybe(Settings)
   })
   .views(self => ({
+    get assignedWorkflowID() {
+      return self.settings?.workflow_id
+    },
+
     promptAssignment(currentWorkflowID) {
       const project = getRoot(self).project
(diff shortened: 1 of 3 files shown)


packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
```diff
@@ -97,13 +89,29 @@ const UserProjectPreferences = types
...
       refreshResource: flow(function * refreshResource() {
-        yield _fetch()
+        try {
+          const preferences = yield _fetch()
+          if (preferences) {
+            self.preferences = preferences.preferences
+          }
+        } catch (error) {
+          self.handleError(error)
+        }
```
eatyourgreens, line 114: This will crash the classifier if the refresh request fails (eg. I'm on mobile and temporarily lose network.) It might be better to ignore the error and let the volunteer continue to classify with their old UPP.

mcbouslog: I think it might be better to ignore the error and let the volunteer continue to classify.

mcbouslog (commented):
Apologies, this took me some time to familiarize myself with. I confirmed the issue on staging project Planet Finders Beta, confirmed this branch fixes the issue.

A few questions for my understanding:
- why was `_fetch` refactored to remove `flow/yield` to `async/await`? was it because it's just requesting the project preference resource now, and not updating loading states at all (loading state updates moved to `fetchResource`, omitted from `refreshResource` to fix bug)
- this is outside of this PR, but I think important to confirm promotion still working as expected - I don't understand [the last test in UserProjectPreferences.spec.js](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.spec.js#L346-L368). 
    - should there be an assertion that equals `true`, like `expect(rootStore.user.personalization.projectPreferences.promptAssignment('123')).to.be.true()`, to test that if a user is on a different workflow than assigned they should be prompted to change to workflow assigned?
    - if the workflow the user's assigned(?) per `upp.settings.workflow_id` is '999' [code link](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.spec.js#L49), should that workflow be included in the test's project active workflows [code link](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.spec.js#L351), so an assertion could be made to confirm `promptAssignment` returns `true`?
    
I'll dig into the following a little more asap, but the reason not approving is that I just want to double-check if I'm seeing things correctly that on master I'm prompted again after 5 to try assigned workflow, but on this branch I don't think I am ever prompted again, on this testing project - https://local.zooniverse.org:3000/projects/srallen086/workflow-assignment-testing/classify/workflow/2333. Is that expected, and if so, should we create an issue to track how this breaks subsequent promotion prompting (but still merge this to fix the bug on all active projects, no active projects are using promotion (I think), so the no subsequent prompt issue can wait)?

eatyourgreens:
That’s a good question about the tests. I’m not sure what they’re testing either. 

This was a refactor, so the behaviour of the store shouldn’t change. I’ll have another look and see if I can see what’s happening with workflow assignment. I’d expect that the new workflow is set when the preferences change, and then the front-end reacts to that change.

eatyourgreens:
Reading back through the code, workflow assignment uses `settings`, not `preferences`. This has broken workflow assignment and the tests should be failing.

packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
```diff
@@ -97,13 +89,29 @@ const UserProjectPreferences = types
...
+          self.handleError(error)
         }
       }),
 
       refreshResource: flow(function * refreshResource() {
-        yield _fetch()
+        try {
+          const preferences = yield _fetch()
+          if (preferences) {
+            self.preferences = preferences.preferences
```
eatyourgreens, line 110: This should update `self.settings`.

eatyourgreens:
```suggestion
            self.settings = preferences.settings
```

eatyourgreens:
@mcbouslog I’ve now vanished down a rabbit hole of trying to figure out how the tests work. I think you’re right, that they pass when they should fail, so there’s an evergreen test in there somewhere. 

I think there’s also a bunch of old code in the setup that isn’t needed, now that the `User` model has been refactored to remove some of the `autorun` observers. I can clean up old code here, but the evergreen tests might be worth their own issue.

eatyourgreens:
Sorry, forgot to say that I would prioritise fixing the classifier for live projects over breaking workflow assignment, if workflow assignment isn’t live. You’re right that we can circle back to fix that when it’s needed.

eatyourgreens:
There were a couple of problems with the tests that I've started to fix here:
- `rootStore` was defined in the top level scope, so that store state leaked across tests. I've moved it down to the scope of each individual test block. That means more setup code but loading state etc. is properly reset for each set of tests.
- tests weren't waiting for the mock UPP data fetching to complete. I've added MobX's [`when`](https://www.mobxjs.com/refguide/when.html) to run the tests when loading is complete. That technique is more-or-less copied straight from [the classifier's `UPPStore` tests](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/UserProjectPreferencesStore.spec.js).

I'm still not sure what the workflow assignment tests should be testing, so I've left those out of this PR.

eatyourgreens:
@mcbouslog I added logging, and here's why those tests are broken:
```
Stores > UserProjectPreferences
    Views > promptAssignment
      when a workflow is not currently selected
{ assignedWorkflowID: '555', currentWorkflowID: undefined }
        ✓ should not prompt the user
      when the assigned workflow is not active
{ assignedWorkflowID: undefined, currentWorkflowID: '123' }
        ✓ should not prompt the user
      when the assigned workflow is the same as the current workflow
{ assignedWorkflowID: undefined, currentWorkflowID: '555' }
        ✓ should not prompt the user
      when the assigned workflow is not the same as the current workflow
{ assignedWorkflowID: undefined, currentWorkflowID: '123' }
        ✓ should prompt the user
```

eatyourgreens:
I think all that's left to do here is to write some tests that refresh the assigned workflow without unexpected changes to the user classification count or classifier loading state.

eatyourgreens:
I'm not sure how to test workflow assignment on the test project, but I've added some tests that refresh your settings and they update the assigned workflow ID without changing your classification count or the classifier loading state.

packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
eatyourgreens: I've changed this to log the error without changing the loading state.

mcbouslog (approved):
This looks great! Confirmed this fixes the classifier reloading issue and stats resetting issue. 

tl;dr - this breaks prompting every 5 classifications if promoted to a different workflow than the workflow classifying on, but that's ok - that needs further refactoring and still prompts initially. 

longer story - This still requests and updates user project preferences (`settings.workflow_id`) every five classifications to see if a user is promoted for the workflow promotion feature, however, it breaks the prompting every 5 classifications because the WorkflowAssignmentModalContainer is using `useEffect` based on `loadingState` which is no longer updated on settings updates. I think fixing the subsequent prompt should be revisited in subsequent work. We could change the WorkflowAssignmentModalContainer `useEffect` to listen to `assignedWorkflowID` and then I think it would prompt only if promoted. That might be a nice quick fix for a subsequent PR. Or as discussed in standup dig into using Sugar, which would be a better longterm solution (then can remove requests every 5 classifications).

Issue #2496, opened 2021-10-15
Freehand Line Undo Feature


ErikOstlund:
## Package
`lib-classifier`

## Describe the solution you'd like
Add an `UNDO` button in the `Freehand Line` toolbar.
The `UNDO` button should only appear when a line is active.
Clicking this button will remove a length of the path where the user last drew.
Currently the length is unknown. Experiment with how much should be removed to see what feels right.

![image](https://user-images.githubusercontent.com/9533007/137520882-1958665c-6452-445d-a44f-00a52ef23448.png)

Use case:
A user is drawing a line with the `Freehand Line` tool. They get pretty far into their drawing and make a mistake! The user can click `UNDO` and the last bit of line will be deleted (lets say 20 px). The user can click `UNDO` as many times as needed to return to a 'happy' place, then continue drawing! 😊 

If a user continues to click `UNDO` where the line is shorter than the minimum required length (need to experiment with this as well), the line should be completely deleted.

## Additional context
This is part of the Freehand Line tool feature. For the complete list of features, see: [GitHub Project](https://github.com/zooniverse/front-end-monorepo/projects/14).

PR #2517, opened 2021-11-03
Freehand Line: Continue and Undo

### LINKED ISSUE #2494: Continue Drawing and Close Path
## Package
`lib-classifier`

When a user selects the `Freehand Line` tool, does a `mousedown` and begins to drag, a path will begin to render on screen. At the initial start point where the user did a `mousedown`, a handle will be created. A user can continue drawing a shape and return to the start point, release the mouse button and the path will close.

If a user does a `mouseup` without returning to the start point, the path will be open and a handle will be created. In this scenario, there should now be two handles (start point and end point). When a user hovers the cursor over a handle, a `tooltip` will appear showing the user what can be done.

When a user clicks and drags on the end point handle, drawing will continue. This will not create a new path, but instead add coordinates to the current path.

image

A user should easily be able to start drawing, stop if needed, then continue drawing the same line. When drawing a path back to the start point, the line will auto-close when user releases the mouse button.

## Additional context
This is part of the Freehand Line tool feature. For the complete list of features, see: [GitHub Project](https://github.com/zooniverse/front-end-monorepo/projects/14).

ErikOstlund:
Package: `lib-classifier`
Project: `Freehand Line Tool`
[GitHub Project URL](https://github.com/zooniverse/front-end-monorepo/projects/14)

Closes #2494 and #2496 

Describe your changes:
Added `Continue Drawing`, `Close Path`, and `Undo` features to the `Freehand Line` tool.

### Methods

- `appendPath()`

At the starting point, one will see a solid handle having a stroke and fill matching the `currentColor`. If the user releases the mouse button before closing the path, a handle will appear at the last drawing point. The handle will have a stroke matching the `currentColor` and no fill. This transparent handle is an indicator to the user that they can continue drawing!
A user can continue drawing by clicking and dragging on the transparent handle. All coordinates will be appended to the active path via the `appendPath()` method.

### Methods

- `isCloseToStart()`

When a user's drawing path returns to the starting point and is within 10px, the path will close and both handles will no longer be visible. This is done by simply appending `Z` to the end of the `<path>`.

### Methods

- `shortenPath()`

Near the starting point is an `Undo` button. When a user clicks this button, 20 coordinates will be removed from the end of the `<path>`. This will be the 20 most recently drawn coordinates.

A user can use the `Undo` feature after a path has been closed!

## Testing

Checkout this branch. In the front-end-monorepo folder, run yarn bootstrap to build the packages. Then cd packages/lib-classifier and run yarn dev.
Good testing project with Freehand Line tool enabled: [Drawing Grid](https://localhost:8080/?env=production&project=16201&workflow=18675)


goplayoutside3 (approved):
Looks good! In this PR I can continuously draw a freehand line from the 'handle' end. When the active line ends within 10px of the initial point, the svg line closes. Pressing the undo button on an active line removes the last 20 coordinates.

PR #2531, opened 2021-11-10
Use Incremental Static Regeneration to build project pages

### LINKED ISSUE #3341: Project app: slow page response times
## Package
app-project

## Describe the bug
New Relic reports response times ~2s for project pages.
New Relic screenshot showing response times between 1.5s and 2.5s for dynamic NextJS json files.

## To Reproduce
Load any FEM project in your browser and examine the response times for either the initial page HTML, or subsequent requests for new page props (`.json` URLs served from NextJS.)

## Expected behavior
Response times should be ~100ms, maybe a few hundred milliseconds if we have to wait for a page to be built.

## Additional context
This screenshot of a request for a Poets & Lovers Classify page shows that the bottleneck seems to be a 2000ms wait on getting data from the Panoptes API.
Screenshot of the timings for /classify/workflow/21363/subject-set/103996.json, showing a 2000ms API request for set_member_subjects.

### LINKED ISSUE #3342: Project app: availability problems (404 and 500 errors for live projects)
## Package
app-project

## Describe the bug
Pages on live projects occasionally return either '500: Internal server error' or '404: not found'. I've seen this happen on [PH-TESS](https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess) and [Maria Edgeworth Letters](https://www.zooniverse.org/projects/mariaedgeworthletters/maria-edgeworth-letters) but it probably affects all projects served by the NextJS app.

## To Reproduce
The problem seems to be intermittent but it's characterised by error reports from `fe-project.zooniverse.org`: 'FE: Project is down'. If you see that error happening, you should be able to reproduce the problem by visiting any live project.

## Expected behavior
The NextJS app shouldn't return 404 for live projects.

## Additional context
The underlying cause seems to be the app running into rate limits for I/O, causing API requests to Panoptes to time out. If a project isn't returned from Panoptes here, the app will return 404:
https://github.com/zooniverse/front-end-monorepo/blob/d577fe4ee435e2630331be8d66365a9ce3a7206a/packages/app-project/src/helpers/getStaticPageProps/getStaticPageProps.js#L21-L26

eatyourgreens:
This replaces [`getServerSideProps`](https://nextjs.org/docs/basic-features/data-fetching#getserversideprops-server-side-rendering) with [`getStaticProps`](https://nextjs.org/docs/basic-features/data-fetching#getstaticprops-static-generation) for all project pages. I've used `fallback: blocking`, so that requests block during the initial build, and an arbitrarily chosen refresh time of 60s, which can be tweaked for each page. For an overview of how this works, see [Incremental Static Regeneration](https://vercel.com/docs/concepts/next.js/incremental-static-regeneration) in NextJS. Static pages are served using [`Cache-Control: s-maxage=60, stale-while-revalidate`](https://web.dev/stale-while-revalidate/) to make sure that the end user always sees something.

Static page generation can't support query parameters, so I've added [middleware](https://nextjs.org/docs/middleware) (new in Next 12!) to inject `env` into the page path. Production pages are published at `/projects/production/:owner:/:project:`, staging at `/projects/staging/:owner:/:project:`.

~~Static builds don't have access to request cookies, so theme cookies are broken on this branch. See also #2555 and #2579.~~ Theme cookies are handled in the browser now. 

Try it out with Galaxy Zoo
https://local.zooniverse.org:3000/projects/zookeeper/galaxy-zoo?env=production
or I Fancy Cats
https://local.zooniverse.org:3000/projects/brooke/i-fancy-cats?env=staging

### Updated: 24 August 2022

Here are some deployed projects on https://fe-static.zooniverse.org:
- [HMS NHS](https://fe-static.zooniverse.org/projects/msalmon/hms-nhs-the-nautical-health-service).
- [PH-TESS](https://fe-static.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess).
- [PH-TESS in French](https://fe-static.zooniverse.org/projects/fr/nora-dot-eisner/planet-hunters-tess) (demonstrates localised text and URLs.)
- [Maria Edgeworth Letters](https://fe-static.zooniverse.org/projects/mariaedgeworthletters/maria-edgeworth-letters).
- [I Fancy Cats](https://fe-static.zooniverse.org/projects/brooke/i-fancy-cats?env=staging) (demonstrates the `env` query parameter with static builds.)
- [I Fancy Cats in French](https://fe-static.zooniverse.org/projects/brooke/i-fancy-cats?language=fr&env=staging) (demonstrates the `language` and `env` query parameters with static builds.)
- [Deciphering Secrets](https://fe-static.zooniverse.org/projects/profdrrogerlouismartinez-davila/deciphering-secrets-unlocking-the-manuscripts-of-medieval-spain/classify/workflow/19276/subject-set/106935/subject/79594592) (demonstrates subject selection and routing with the subject ID in the URL.)
- [The error logger](https://fe-static.zooniverse.org/projects/brooke/i-fancy-cats/logger?env=staging) (use this to test Sentry logging.)
- [The health check page](https://fe-static.zooniverse.org/projects/Index) (kubernetes uses this to test whether the app is running.)

Sentry errors for this branch are logged under [the `static` environment](https://sentry.io/organizations/zooniverse-27/issues/?environment=static&project=1492691).

## Package

app-project

- Includes #3345.
- Fixes #3341.
- Fixes #3342.

### General

### Components

### Apps

- [ ] Does it work on mobile?

### Publishing

### Post-merging


eatyourgreens:
Rebased to the latest master.

eatyourgreens:
Bumping this because it should fix slow first byte times for HMS NHS, which I think are due to `getServerSideProps` being slow. 
https://www.webpagetest.org/result/220121_BiDcSY_322f741abacc9e45350f9b7add831fff/

goplayoutside3:
I want to discourage us from going this route before having a discussion team-wide about the UX consequences. Yes, static builds do increase frontend app performance by generating cached HTML, but changing `/pages` data fetching to `getStaticProps` will directly effect project builders’ experience. @lcjohnso is out for a couple weeks, but I remember him pointing out the need for a discussion about UX when this PR was opened a couple months ago.

For example, app-content-pages uses `getStaticProps` and there is a delay between updating data in our Contentful CMS and when users see the changes live. The delay time is set as 1 hour. If app-project fetches data with `getStaticProps`, teams building projects will have to wait to see their changes live (the wait is set at 10 minutes in this PR). This app behavior is in contrast to the expected behavior when refreshing a browser window.

### Dev Environment Notes
- While in a local dev environment, `getStaticProps` is called on every request, which is different than the behavior of a live production site
- I couldn’t run a project locally with this branch until I modified `server/dev-server.js` to include `const app = next({ dev, customServer: true, hostname: 'local.zooniverse.org', port })`

### Notes on an empty array in getStaticPaths
- `paths: []` is telling Next.js to render the root-most route. This means no paths are pre-rendered in app-project. The empty array is a way to use dynamic urls without a long build time, which is good for routes where there's lots of possible dynamic values (like in Zooniverse’s case!)
- /projects/:owner/:project/:anything is not statically generated at deploy time, so Next.js will SSR on the first request
- For example, with the current `getStaticProps` setup the i-fancy-cats home page html is not generated until someone in the world first visits its url. Then, the home page’s html is generated. Data on the home page is cached for 10 minutes before any new changes can appear. SEO will apply when the home page is generated at the first request.

### Pros ✔️
- Users experience faster page load when that page's data is cached
- Lighthouse metrics like First Byte Time will improve

### Concerns ❌
- Data changes will only render on app-project pages after the `revalidate` time is up. Unexpected UX consequences could include delay between editing projects in the project builder and seeing those changes reflected live. 
- I didn't investigate deeply yet, but what happens when `getDefaultPageProps` is in `getStaticProps`? That function will only be called after the `revalidate` time is up, but it includes UI preferences like `dismissedAnnouncementBanner`.

### Major Question
- Is using `getsStaticProps` worth it if `revalidate` is set to less than 1 second in order to "immediately" render changes in data?

eatyourgreens:
@goplayoutside3 Thanks for taking a look. Those are all good points.

This is just a quick note to say that I've changed the revalidate time to 60s, to match the maxage that we send in `Cache-Control` headers. It seems reasonable, to me, to run builds on roughly the same timescale that we catch the generated HTML and JSON in the CDN and in browser caches.

For projects that are still in development, I think we're going to need to use NextJS preview mode to get immediate feedback after changes. Otherwise, the 60s cache in production might catch us out anyway.
https://nextjs.org/docs/advanced-features/preview-mode

Cookies for the theme mode and dismissed announcement banner are currently broken in production. You can try that out on my test project. Dismiss the banner, then reload the page and it reappears again (ignoring the cookie.)
https://frontend.preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground?env=staging
#2579should fix that by moving cookie handling from the server to the browser. You're absolutely right that cookie handling needs to happen in the browser for static pages too.

Right now, we're seeing slow page builds in production on HMS NHS and Davy Notebooks. HMS NHS is relaunching with new data in a couple of weeks (@snblickhan have I got this right?) Ideally, I'd like to have the slow page builds fixed before that launch. I think that static builds, here, should help to fix that.

goplayoutside3:
> This is just a quick note to say that I've changed the revalidate time to 60s, to match the maxage that we send in Cache-Control headers. It seems reasonable, to me, to run builds on roughly the same timescale that we cache the generated HTML and JSON in the CDN and in browser caches.

Good point! 

> For projects that are still in development, I think we're going to need to use NextJS preview mode to get immediate feedback after changes. Otherwise, the 60s cache in production might catch us out anyway.
https://nextjs.org/docs/advanced-features/preview-mode

Agreed! I think preview mode is a good option for project development.

Other than project development while using the project builder, are there any other aspects of FEM that update immediately that might be affected by switching to `getStaticProps`?

eatyourgreens:
I think the only things that might change rapidly are the classification stats on the home page and the classify page. Having said that, I think the browser fetches those numbers. We wouldn't expect to be fetching them in page props.

The classifier runs entirely in the browser, so we don't need to worry about accidentally caching subjects. Recent classifications is also generated client-side, so that wouldn't be a problem.

Recent Talk comments, on the home page, might be generated on the server, but Talk comments should be ok with a 60s refresh rate. Similarly, workflow and subject set completeness counters change over time, but not fast enough to require rebuilding every few seconds.

Off the top of my head, I think those are the only parts of the page where we might expect to see changes when we reload.

eatyourgreens:
@goplayoutside3 I've rebased this to add internationalisation and the `locale` parameter for project About pages.

HMS NHS relaunched this week and I'm still seeing slow build times on dynamic routes eg. over 2s here to get 12k of page props JSON for a classifier workflow page. I'll open an issue for that but it is down to our using `getServerSideProps` to query Panoptes on every page transition.

https://frontend.preview.zooniverse.org/projects/msalmon/hms-nhs-the-nautical-health-service/classify/workflow/18611
Screenshot of the network console, showing the request timings for page props, which take 2.32s to download.

eatyourgreens:
Noting, as per @goplayoutside3's review, that middleware is broken for `yarn dev` because the dev server is misconfigured.
```
Error: To use middleware you must provide a `hostname` and `port` to the Next.js Server
```

eatyourgreens:
The merge conflicts have turned out to be straight forward: adding `locale` and the new props to each page. No trouble at all.

eatyourgreens:
Sentry link for the middleware error:
https://sentry.io/organizations/zooniverse-27/issues/2985962292/?environment=development&project=1492691&query=is%3Aunresolved

eatyourgreens:
Updated for NextJS 12.1. This should build and run locally now.

eatyourgreens:
Rebased to bring in project translations, which work nicely with static pages.

eatyourgreens:
The NextJS app is now deployed to `fe-static.zooniverse.org` for testing:
- https://fe-static.zooniverse.org/projects/eatyourgreens/-project-testing-ground?env=staging
- https://fe-static.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess
- https://fe-static.zooniverse.org/projects/msalmon/hms-nhs-the-nautical-health-service

eatyourgreens:
New Relic still shows response times ~2s for requests that use server-side props.
New Relic screenshot showing response times between 1.5s and 2.5s for dynamic NextJS json files.

eatyourgreens:
@goplayoutside3 If I edit my test project in production, I don't see the changes immediately on the home page:
https://frontend.preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground?env=staging

Looking at the Network panel, we're setting a cache time of 60s currently, so I don't think this PR will change the behaviour of live projects. There'll still be a ~60s delay between making a change in the project builder, and seeing that change on the live site.

eatyourgreens:
I had revalidation set to the default, [which is one year](https://nextjs.org/docs/going-to-production), for About pages. Should be fixed now.

eatyourgreens:
@lcjohnso @goplayoutside3 here are the  NextJS response times for the PH-TESS classify page props with dynamic vs. static builds:

- dynamic (https://frontend.preview.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess) 1.05s 
Screenshot showing a 1.05s response time for /classify/workflow/11235.json
- static (https://fe-static.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess) 101ms 
Screenshot showing a 101ms response time for /classify/workflow/11235.json

eatyourgreens:
Switching the API requests over to using `panoptes-production-app`, instead of `www.zooniverse.org`, drops the response time down to 23ms here.
New Relic screenshot, showing the API response times during a page build in the project app.

eatyourgreens:
@goplayoutside3 @lcjohnso If I make a change to the description of my test project, I see the new edit on the static page here:
https://fe-static.zooniverse.org/projects/eatyourgreens/-project-testing-ground?env=staging

but the old text with the current code here:
https://frontend.preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground?env=staging

`frontend.preview` does update eventually, but I have to wait for my browser cache to expire.

I suspect the difference is that `fe-static` uses `stale-while-revalidate` but, in answer to question about UX above, ISR might actually improve the experience for project owners who are editing their projects, then refreshing the page to check for changes. I've only run very quick tests, though, editing my project in the project builder then refreshing both those pages to see which changes first.

eatyourgreens:
Here's a screen recording of edits to a project page. The changes take a couple of page refreshes to load on both `frontend.preview` and `fe-static`. `frontend.preview` is slightly faster, but not by much.


https://user-images.githubusercontent.com/59547/175910349-bf08f450-4cf1-4f19-b336-2b328f889cc5.mov

eatyourgreens:
I haven't looked at this recently, but links direct to the NextJS app might bypass the caching headers that we set.
eg. https://fe-project.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess

eatyourgreens:
12.2 is blocked by #3429 at the moment.

eatyourgreens:
This is a useful link for the middleware changes. 
https://nextjs.org/docs/messages/middleware-upgrade-guide#executing-middleware-on-internal-nextjs-requests

eatyourgreens:
@goplayoutside3 I've added a check for requests to `/_next`, so this is now compatible with NextJS 12.1 and NextJS 12.2.

eatyourgreens:
Updated to add an example with the `language` query param, now that #3152 has merged.

goplayoutside3:
I encountered a couple of bugs while reviewing.
1. Visiting the linked [i-fancy-cats](https://fe-static.zooniverse.org/projects/brooke/i-fancy-cats?env=staging) project in the PR description causes some components on the home page to flicker, or in Safari the page tries to load over and over and over.
2. Running this branch locally after `yarn bootstrap` gives a ReferenceError: PropTypes is not defined caused by WorkflowSelectButton.

eatyourgreens:
Login is broken too. The 'Sign In' button does nothing on the [PH TESS home page](https://fe-static.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess).

eatyourgreens:
#3569 fixes the PropTypes error in `WorkflowSelector`.

eatyourgreens:
A local build with NextJS 12.1.0 doesn't have that bug, so I'm going to try it again with 12.1.6.

eatyourgreens:
Logging in throws this error, so the router is still using `as` somewhere.
https://nextjs.org/docs/messages/incompatible-href-as

eatyourgreens:
The test `logger` page loads without redirecting, so the redirect bug is restricted to project pages, but it does happen on all project pages when there's a query string in the URL.
https://fe-static.zooniverse.org/projects/brooke/i-fancy-cats/logger?env=staging

eatyourgreens:
Removing the URL rewrite from middleware fixes the redirect bug, but then you have to specify the API environment in the URL: eg. https://local.zooniverse.org:3000/projects/fr/production/nora-dot-eisner/planet-hunters-tess/about/research?login=true will work without reloading.

eatyourgreens:
A temporary workaround is to specify the API environment in the request URL. That's not useful for production, but allows us to debug the problem without the page reloading.

This link should not get trapped in a redirect/rewrite loop: https://fe-static.zooniverse.org/projects/production/nora-dot-eisner/planet-hunters-tess?login=true

eatyourgreens:
Paging through a subject set here is slightly broken, in that each new subject mounts the classifier and opens the tutorial again.
https://fe-static.zooniverse.org/projects/profdrrogerlouismartinez-davila/deciphering-secrets-unlocking-the-manuscripts-of-medieval-spain/classify/workflow/19276/subject-set/106935/subject/79594592

eatyourgreens:
Upgrading NextJS to 12.2.5 seems to have fixed that reload bug, when the page URL has a query string.

eatyourgreens:
All the test URLs should be working again, once the latest commit is deployed.

eatyourgreens:
The `/classify` link is broken, in the project navigation, for projects with more than one workflow. All the other page routes work, but the page params for `/classify` look like this, with the owner name being passed as the API environment.
```js
{
  locale: 'en',
  params: {
    env: 'msalmon',
    owner: 'msalmon',
    project: 'hms-nhs-the-nautical-health-service'
  }
}
```

eatyourgreens:
Updated the PR description to include the Sentry log for this branch.

goplayoutside3 (approved):
Is there any language in the project builder that needs to be updated when this change is merged? For instance, do FEM project teams need to be notified to wait 60 seconds for new changes to about pages or homepage details to display?

eatyourgreens:
Good point. I don't think so, since we have a 60s cache in place already but I haven't tested this branch recently.

eatyourgreens:
From what I remember, the current situation with live projects is that if I edit a project, and the page I'm editing is in my browser cache, then changes don't show up until the cache expires or is cleared. If the page isn't in the browser cache, changes show up immediately.

PR #2535, opened 2021-11-11
Drawing tools models cleanup

ErikOstlund:
Package: `lib-classifier`

Describe your changes:
mobx actions previously `explicitly` returned functions.
image

This PR refactored all the drawing tools model actions to `implicitly` return functions.
This approach allows for less code, is easier to read and is a modern clean style.
image

This PR does not change any functionality.


eatyourgreens (approved):
Looks good to me!

PR #2542, opened 2021-11-15
Vertical layout for Project Announcements

### LINKED ISSUE #2471: Banner paragraph splits text into columns
## Describe the bug
FEM banner text splits up paragraphs into separate columns, see screenshot below.

## To Reproduce
1. Enter text in a banner on a FEM project
2. Include a paragraph break
3. View project

## Expected behavior
Paragraphs should show up below one another, as normal text would. Also, it seems like there are huge margins here, which should be reduced.

## Screenshots
Screen Shot 2021-10-05 at 10 47 20 AM

shaunanoordin:
## PR Overview

Package: `app-project`
Fixes #2471

Currently, the Project Announcements banner will sometimes split the announcement into multiple "columns". This is actually caused by the Markdown splitting each paragraph into its own `<p>` element, and those elements sitting in a `row`-directioned (flex)Box element

_Example: a Project Announcement with 3 paragraphs_

Screenshot 2021-11-15 at 16 22 20

This PR fixes the issue by giving project announcements a vertical layout. (Plus, there's a small styling tweak to space the paragraphs apart a bit.)

_Example: that same Project Announcement, but with the new vertical layout, on a wide screen_

Screenshot 2021-11-15 at 17 23 49

_Example: that same Project Announcement, but with the new vertical layout, on a narrow screen_

Screenshot 2021-11-15 at 17 23 58

### Testing

- This fix can be tested on any project with multiple paragraphs in the "project announcements" settings
  - Note: "project announcements" are rendered in Markdown, so a new paragraphs are created after two newlines, e.g.
  ```
  This is paragraph 1.

  And after two newlines, here is paragraph 2.
  ``` 
- For convenience, here's Test Project 2021 with those 3 paragraphs: https://local.zooniverse.org:3000/projects/darkeshard/test-project-2021/

### Status

Ready for review


shaunanoordin:
Some context: this issue was first noticed in the Engaging Crowds projects - Scarlets & Blues, I think?

It's a low-priority bug since not that many projects have long project announcements, but it's still classified as a visual UI bug since nobody expected long announcements to be broken into columns.

eatyourgreens:
Should we allow block level elements inside those banners? I’d be in favour of limiting them to inline content.

shaunanoordin:
> Should we allow block level elements inside those banners?

I'm neutral about this idea, but if you feel strongly about it, I'll need to ~~find out if [Markdownz](https://github.com/zooniverse/markdownz) has a _"pls don't turn text blocks into paragraphs bro"_ toggle~~ (EDIT) just change `{announcement}` to `{announcement?.replace(/\n/g, ' ')}` to remove all line breaks.

eatyourgreens:
On the other hand, if this changed to support full Markdown, then I think the editor in the project builder should also be changed from a plain text input to a rich text editor. 

I’m cautious about project owners using this to insert large blocks of content into the page. I think the banner is supposed to show short announcements.

packages/app-project/src/components/Announcements/components/GenericAnnouncement/components/NarrowGenericAnnouncement/NarrowGenericAnnouncement.js
```diff
@@ -1,9 +1,18 @@
...
 
 import markdownzComponents from '../../helpers/markdownzComponents'
 
+// Spaces the paragraphs created by Markdown a little more nicely.
+// Box.gap='large' doesn't seem to work.
+const SpacedBox = styled(Box)`
+  p {
+    margin: 0.25em 0;
+  }
+`
```
eatyourgreens, line 14: I think this is possible with Box props in Grommet, but I can't remember how off the top of my head.

eatyourgreens (approved):
Changes look good to me. I'm not 100% sure we should allow long runs of formatted content in the announcement banner, but this will fix Scarlets and Blues.

PR #2520, opened 2021-11-06
Refactor annotatedSteps as subject.stepHistory

eatyourgreens:
- Move the subject models, and their dependent code, into the `SubjectStore` directory, so that models are next to the code that uses them.
- Refactor `store.annotatedSteps` as `subject.stepHistory`, so that each subject now has its own `undoManager`. `subject.stepHistory` remembers the step keys and annotation IDs for a single subject. `stephistory.back()` undoes the current step and its annotations. `stepHistory.next()` either redoes an existing step and annotations, from the subject's redo history, or creates a new step if redo isn't available.
- Refactor the `StepHistory` tests to test history for individual subjects, rather than a global history manager.
- The changes here break the tests for the classification store, so I've stubbed `classifications.reset` in those tests, in order to get them working again. Resetting the classification store destroys the active classification, making it impossible to test classification state with those tests as they are currently written (`classifications.active` is always a new classification, because of the reset.)

#2524 is the associated ADR, for context.

Package:
lib-classifier

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2548, opened 2021-11-17
Classifier: Run addMiddleware, onAction and onPatch outside of autorun

eatyourgreens:
#2520 revealed a subtle bug in the way that hooks like `onPatch` are being run in the classifier. `onPatch` was called inside an `autorun` listener, which ran every time a new classification was created. A side effect of this was that a new patch listener was created for each subject classified. So the first subject advanced the subject queue by one subject, the second by two subjects, the third by three subjects and so on.

This PR removes `onPatch` from the subjects store and moves it to the root store, where it can observe changes in the `classifications` sub-tree and call actions in the `subjects` sub-tree.

I've also unwrapped `onAction` and `addMiddleware`, which were being run inside `autorun` listeners, and removed some stubs that were needed to pass tests in #2520. In retrospect, I think those stubs were covering up for the bug mentioned above.

With this PR, `onAction`, `onPatch` and `addMiddleWare` are run only once, after the root store is created. The listeners that they create will then run when their respective triggers are triggered.

I've also removed anonymous functions nested inside other functions. Personally, I find anonymous, nested functions hard to read and debug so I've replaced them with named functions in each case.

[Listeners in MobX State Tree](https://mobx-state-tree.js.org/concepts/listeners)

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/FeedbackStore.js
+++ b/packages/lib-classifier/src/store/FeedbackStore.js
@@ -1,5 +1,5 @@
 import { autorun } from 'mobx'
-import { addDisposer, addMiddleware, getRoot, isValidReference, types } from 'mobx-state-tree'
+import { addDisposer, addMiddleware, getRoot, isValidReference, tryReference, types } from 'mobx-state-tree'
 import { flatten } from 'lodash'
 
 import helpers from './feedback/helpers'
@@ -37,29 +37,22 @@ const FeedbackStore = types
     get messages () {
       return self.applicableRules
         .map(rule => rule.success ? rule.successMessage : rule.failureMessage)
+    },
+    get shouldShowFeedback() {
+      return self.isActive && self.messages.length && !self.showModal
     }
   }))
   .actions(self => {
-    function setOnHide (onHide) {
-      self.onHide = onHide
-    }
-
-    function afterAttach () {
-      createSubjectMiddleware()
-    }
-
-    function onNewSubject () {
-      const validSubjectReference = isValidReference(() => getRoot(self).subjects.active)
-      if (validSubjectReference) {
-        const subject = getRoot(self).subjects.active
-        self.reset()
-        self.createRules(subject)
+    function _subjectsMiddleware(call, next, abort) {
+      if (call.name === 'advance') {
+        _onSubjectAdvance(call, next, abort)
+      } else {
+        next(call)
       }
     }
 
-    function onSubjectAdvance (call, next, abort) {
-      const shouldShowFeedback = self.isActive && self.messages.length && !self.showModal
-      if (shouldShowFeedback) {
+    function _onSubjectAdvance(call, next, abort) {
+      if (self.shouldShowFeedback) {
         if (process.browser) {
           console.log('Aborting subject advance and showing feedback')
         }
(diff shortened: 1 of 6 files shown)


packages/lib-classifier/src/store/FeedbackStore.js
```diff
@@ -70,17 +60,16 @@ const FeedbackStore = types
...
-          if (call.name === 'advance') {
-            onSubjectAdvance(call, next, abort)
-          } else {
-            next(call)
-          }
-        })
-      }, { name: 'FeedbackStore Subject Middleware autorun' })
-      addDisposer(self, subjectMiddleware)
+    function afterAttach () {
+      addMiddleware(getRoot(self).subjects, _addMiddleWare)
```
eatyourgreens, line 64: Why is feedback adding middleware to the subjects store? This is a bit confusing.

packages/lib-classifier/src/store/RootStore.js
```diff
@@ -49,6 +50,30 @@ const RootStore = types
...
+        annotations.forEach(annotation => self.feedback.update(annotation))
+      }
+    }
+
+    function _onPatch(patch) {
+      // TODO: why are we doing this rather than observe classifications.loadingState for changes?
+      const { path, value } = patch
+      if (path === '/classifications/loadingState' && value === 'posting') {
+        self.subjects.advance()
+      }
```
eatyourgreens, line 74: Could this be replaced by an `autorun` function that reacts to changes in `classifications.loadingState`?

packages/lib-classifier/src/store/FeedbackStore.js
eatyourgreens: The tests mock feedback in isolation, so `getRoot(self).subjects` is undefined when they run.

mcbouslog (commented):
I'm in the process of trying to troubleshoot, but noting on https://local.zooniverse.org:3000/projects/nora-dot-test/planet-finders-beta/classify/workflow/3317 the feedback modal doesn't show when expected (can check subject metadata which will note mock transits for feedback). The classification metadata correctly has the feedback (and related success or failure), but no modal showing feedback.

eatyourgreens:
>I'm in the process of trying to troubleshoot, but noting on https://local.zooniverse.org:3000/projects/nora-dot-test/planet-finders-beta/classify/workflow/3317 the feedback modal doesn't show when expected (can check subject metadata which will note mock transits for feedback). The classification metadata correctly has the feedback (and related success or failure), but no modal showing feedback.

`store.subjects` hasn't been set up yet when the feedback store initialises, so `getRoot(self).subjects` is undefined.

eatyourgreens:
@mcbouslog I've fixed feedback by adding `autorun` back in, so that it now observes `getRoot(self)?.subjects` before adding the middleware. I think the difference, with that model, is that it's adding middleware to another model, so we need to wait for `subjects` to be ready.

A better model might move `feedback` from `store.feedback` to `store.subjects.feedback` or `store.subjects.active.feedback`, so that feedback can react to subjects.

I've been testing this on
https://localhost:8080/?project=nora-dot-test/planet-finders-beta&demo=true
using the support for project slugs from #2551.

mcbouslog (approved):
This LGTM, nice refactor!

PR #2549, opened 2021-11-17
Engaging Crowds: track classifier subject changes in the page URL

### LINKED ISSUE #2537: Subject URLs should update for projects that use subject selection
## Package
app-project
lib-classifier

## Describe the bug
- Start Scarlets & Blues on subject 3/21: https://www.zooniverse.org/projects/bogden/scarlets-and-blues/classify/workflow/18505/subject-set/98879/subject/69734438?demo=true
- Submit a couple of demo classifications so that you're now on subject 5/21.
- Hit Done & Talk.
- Use the Back button to return to the Classifier.
- You're now back on page 3/21 because that's the subject ID in the page URL.

## Expected behavior
You should go back to the subject you were working on.

## Additional context
You can use the Next Available Subject button to jump back to an unclassified subject, but it's not an ideal experience.

Tested in Firefox 94 on MacOS 11.6.1.

eatyourgreens:
Add an `onSubjectChange` prop to the classifier, which passes a snapshot of the active subject to the containing page. The `ClassifierWrapper` component listens for subject changes and performs a shallow update of the page location, if we're on a page that uses the `subjectID` parameter.

As a result, if we go to Talk and back, or if we refresh the page, the page should always reload the active subject ID.

Package:
app-project
lib-classifier

Closes #2537.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapper.js
+++ b/packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapper.js
@@ -53,7 +53,7 @@ export default function ClassifierWrapper({
       const newSubjectRoute = `${baseURL}/workflow/${workflowID}/subject-set/${subjectSetID}/subject/${subject.id}`
       const href = addQueryParams(newSubjectRoute, router)
       const as = href
-      router.push(href, as, { shallow: true })
+      router.replace(href, as, { shallow: true })
     }
   }
 
--- a/packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapper.js
+++ b/packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapper.js
@@ -1,8 +1,10 @@
 import Classifier from '@zooniverse/classifier'
+import { useRouter } from 'next/router'
 import auth from 'panoptes-client/lib/auth'
 import { func, string, shape } from 'prop-types'
 import asyncStates from '@zooniverse/async-states'
 
+import addQueryParams from '@helpers/addQueryParams'
 import { logToSentry } from '@helpers/logger'
 import ErrorMessage from './components/ErrorMessage'
 import Loader from '@shared/components/Loader'
@@ -29,6 +31,7 @@ export default function ClassifierWrapper({
   workflowID,
   yourStats
 }) {
+  const router = useRouter()
   function onCompleteClassification(classification, subject) {
     yourStats.increment()
     recents.add({
@@ -41,6 +44,17 @@ export default function ClassifierWrapper({
   function onError(error, errorInfo={}) {
     logToSentry(error, errorInfo)
     console.error('Classifier error', error)
+  }
+
+  function onSubjectChange(subject) {
+    const { query } = router
+    const baseURL = `/${query.owner}/${query.project}/classify`
+    if (query.subjectID && subject.id !== query.subjectID) {
+      const newSubjectRoute = `${baseURL}/workflow/${workflowID}/subject-set/${subjectSetID}/subject/${subject.id}`
+      const href = addQueryParams(newSubjectRoute, router)
+      const as = href
+      router.push(href, as, { shallow: true })
+    }
   }
 
   function onToggleFavourite(subjectId, isFavourite) {
@@ -74,6 +88,7 @@ export default function ClassifierWrapper({
           onAddToCollection={onAddToCollection}
           onCompleteClassification={onCompleteClassification}
           onError={onError}
+          onSubjectChange={onSubjectChange}
           onSubjectReset={onSubjectReset}
           onToggleFavourite={onToggleFavourite}
           project={project}
--- a/packages/lib-classifier/src/components/Classifier/Classifier.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.js
@@ -71,6 +71,7 @@ export default function Classifier({
   onAddToCollection = () => true,
   onCompleteClassification = () => true,
   onError = () => true,
+  onSubjectChange = () => true,
   onSubjectReset = () => true,
   onToggleFavourite = () => true,
   project,
@@ -103,6 +104,7 @@ export default function Classifier({
   useEffect(function onMount() {
     classifierStore.setOnAddToCollection(onAddToCollection)
     classifications.setOnComplete(onCompleteClassification)
+    classifierStore.setOnSubjectChange(onSubjectChange)
     subjects.setOnReset(onSubjectReset)
     classifierStore.setOnToggleFavourite(onToggleFavourite)
   }, [])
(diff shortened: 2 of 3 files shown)


packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapper.js
```diff
@@ -43,6 +46,17 @@ export default function ClassifierWrapper({
...
 
+  function onSubjectChange(subject) {
+    const { query } = router
+    const baseURL = `/${query.owner}/${query.project}/classify`
+    if (query.subjectID && subject.id !== query.subjectID) {
+      const newSubjectRoute = `${baseURL}/workflow/${workflowID}/subject-set/${subjectSetID}/subject/${subject.id}`
+      const href = addQueryParams(newSubjectRoute, router)
+      const as = href
+      router.push(href, as, { shallow: true })
+    }
```
eatyourgreens, line 57: This will break URLs like https://frontend.preview.zooniverse.org/projects/brooke/i-fancy-cats/classify/workflow/693/subject/4224?env=staging&demo=true so maybe the check should be explicitly for a workflow that uses indexed subjects.

packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapper.js
```diff
@@ -43,6 +46,17 @@ export default function ClassifierWrapper({
     console.error('Classifier error', error)
   }
 
+  function onSubjectChange(subject) {
+    const { query } = router
+    const baseURL = `/${query.owner}/${query.project}/classify`
+    if (query.subjectID && subject.id !== query.subjectID) {
+      const newSubjectRoute = `${baseURL}/workflow/${workflowID}/subject-set/${subjectSetID}/subject/${subject.id}`
```
shaunanoordin, line 53: Writing notes here to make sure that I understand this bit properly.

The `if` triggers when (1) the URL specifies a Subject ID, and (2) the _new Subject_ (i.e. the next subject we're getting from the queue) isn't the same as the _current Subject._ OK, makes sense.

So if the URL is currently `/classify/workflow/TEST/subject-set/FRUITS/subject/APPLE`, and the queue is `[APPLE, BANANA, CHERRY]`, then as soon as I submit the Classification for APPLE, the BANANA subject will load, and the URL will change to `/classify/workflow/TEST/subject-set/FRUITS/subject/BANANA`.

Cool, I understand this new behaviour. 👍 

Question: is it possible for the new route to have the wrong Subject Set? e.g. the queue combines two subject sets, e.g. `[APPLE, CAT, BANANA, DOG]`, so you're moving from `/classify/wf/TEST/sset/FRUITS/sbj/APPLE` to `/classify/wf/TEST/sset/FRUITS/sbj/CAT` ? Instead of `/classify/wf/TEST/sset/ANIMALS/sbj/CAT`

Answer: no, this shouldn't be possible. If we have `/subject-set/` in the path, then by definition we're querying a specialised endpoint that ONLY returns Subjects from the specified Set. If we get `[APPLE, CAT, BANANA, DOG]` when we specify the FRUITS subject set, then the API or the call to the API must have borked.

Question: what about URLs that don't specify the Subject Set?

Answer: [see Jim's comments](https://github.com/zooniverse/front-end-monorepo/pull/2549/files#r753154466). This is a known possible problem. That `if` check might need to be updated to be more strict about what qualifies for the "change Path to match Subject" action.

shaunanoordin (commented):
## PR Review

Package: `app-project`
Affects: Projects that allow single Subject selection

With this PR, when the user is classifying on a Project that allows for _single Subject selection,_ the path/URL now _updates when a new Subject is loaded_ (from the queue (note-1))

Existing behaviour:
- Say we have a Project that lets you select a single Subject...
- When we choose Subject 333 (in WF 100, in SSet 200), the URL will be: `zooniverse.org/projects/foo/bar/classify/workflow/100/subject-set/200/subject/333`
- Then, when we submit 333 and the next Subject 444 loads from the queue, the URL _will remain the same_

New behaviour:
- Same as above, except that...
- When we submit 333 and the next Subject 444 loads from the queue, the URL will change to `zooniverse.org/projects/foo/bar/classify/workflow/100/subject-set/200/subject/444`
- No other behaviour should change.

(note-1) not sure if this also affects when a new Subject loads via the special Next/Prev navigation (on the Indexed Subject Set banner), but I assume the answer is no. I should double check this.

### Testing

Testing plan:
- confirm that new behaviour _doesn't_ trigger on normal Projects without single Subject selection.
  - check stores on the 'Submit Classification' action
- confirm that new behaviour _does_ trigger on a Project with single Subject selection. (e.g. [Scarlets & Blues](https://frontend.preview.zooniverse.org/projects/bogden/scarlets-and-blues/classify/workflow/18505/subject-set/98880/subject/69734487))
  - check stores on the 'Submit Classification' action
  - figure out if 'Go to Next Available Subject' action (on the "Finished" popup) also triggers a URL rewrite
- what happens when you press the Back button?

### Status

Code read looks good. Testing in progress.

shaunanoordin (approved):
## PR Review (Update)

Testing complete. Functionality checks out mostly, but with one caveat.

### Testing

- ✅ The new behaviour doesn't trigger on normal Projects without single Subject selection.
  - Test URL: [a standard test project](https://local.zooniverse.org:3000/projects/darkeshard/single-workflow-project/classify/workflow/3535?demo=true)  
  - When the 'Submit Classification' action triggers: a new Subject is loaded, `onSubjectChange` triggers, but the `if` clause doesn't trigger. The URL doesn't change. Nice.
- ✅ The new behaviour DOES trigger on a Project with single Subject selection.
  - Test URL: [Scarlet & Blues](https://local.zooniverse.org:3000/projects/bogden/scarlets-and-blues/classify/workflow/18505/subject-set/98880/subject/69734490?env=production&demo=true), a random un-retired Subject from a random Subject Set.
  - When the 'Submit Classification' action triggers: a new Subject is loaded, `onSubjectChange` triggers, AND the `if` clause does trigger. The URL changes to reflect the ID of the newly loaded Subject. 👍 
- ✅  On S&B, if I choose a Subject I've already seen, clicking the 'Next Available Subject' button (on the "Finished" popup), a new Subject is loaded, the `onSubjectChange` triggers and the URL changes.
  - ❔  I had some weird intermittent issues where clicking on "Next Available Subject" causes the _same Subject to load,_ but I haven't been able to replicate it consistently. 
- ❌ On S&B, if I advance from Subject 111 to Subject 222, and then I press the 'Back' button on my (Chrome) browser, the URL changes back to Subject 111, but the Classifier still shows Subject 222 (no change).

Tested on localhost with macOS + Chrome (I haven't tested the Back button behaviour on Firefox)

### Status

The "Back" button continues to haunt us. 👻  I think that needs a fix (if we can somehow figure out the Back button logic), but that fix can be separate from this PR.

Note: I can't comment much on what we need to think about, when the URL specifies WF + Subject, but not Subject Set (as Jim pointed out in his [comment](https://github.com/zooniverse/front-end-monorepo/pull/2549#discussion_r753154466)). I know it exists, but I'm not 100% sure when/why we'd use the "WF + Subject only" functionality. 🤔

eatyourgreens:
The Back button is a good point. I don't know if the router has an option to replace, rather than push.

Issue #2553, opened 2021-11-18
Light Curve Viewer / Planet Hunters TESS - annotations disappear when going 'Back' to the Classify page


shaunanoordin:
## Functionality Bug

@eatyourgreens pointed an issue where current annotations on a Light Curve Viewer (the "drag-and-draw columns to mark areas of interest") disappear if you navigate away from the Classify page, then press the 'Back' button on the browser (Chrome? Firefox?).

Steps:
- Go to [Planet Hunters TESS](https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess), which is an FEM project with the Light Curve Viewer
- Go to /classify (the Classify page).
- **Make one annotation/mark** on the light curve (which is a D3.js graph)
- Click on the general Zooniverse 'About' link, or any link that breaks you out of the FEM environment (which only covers the Project home page and Classify page)
- Press the 'Back' button to return to the Classify page
- Observe that the same Subject reloads (the Subject queue remains in memory *️⃣ ), but **the annotation/mark is no longer visible on the light curve.**

This is part of the _"Classifier acts odd (some things are saved, some things aren't, depending on the browser) any time you press the Back button"_ issue that Jim's investigating.

⚠️ However, of special interest to the Light Curve Viewer is this: **is the one annotation STILL saved in the Classifier memory, DESPITE visually disappearing from the graph?**
- If the annotation data is STILL stored in memory BUT not visible on the graph, then _this needs a fix quickly_
- I need to check this because the D3.js graph isn't as tightly bound to the React stores. I can't 100% recall if I set some code for _"on component load, sync annotations from the Classifications store."_ (Maybe? Shaun tends to be very paranoid about edge conditions.) 

*️⃣ Actually, I need to double check that _all_ the Classifier data remains the same. I mean, it'd also be a problem if the Classification data is saved, but now a different Subject loads. (I'm not sure if it's just a coincidence that the Classifier keeps serving the same Subject, in Jim's example video.)

eatyourgreens:
Home, About and Classify are all part of the project app, which is a single page React app, so the store global should be preserved across those pages at least. I think you can also reproduce this by going Home => Classify => Home => Classify.

Firefox and Safari, I think, go one step further and preserve window JS state across pages on the same origin. At least, I can get my annotations back by going back from any page on https://www.zooniverse.org, as far as I can tell.

eatyourgreens:
[...]

PR #2559, opened 2021-11-22
LightCurveViewer: fix "invisible annotations" when returning to Classify page

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Fixes: #2553 

This PR fixes the issue where the LightCurveViewer's existing annotations "disappear" (i.e. go invisible, but still exist in the stores so they'd be submitted in the Classification!) when navigating away from the FEM Classify page to another FEM page, then back to the Classify page.

Basically, there's a desync between the UI visuals (D3.js graph) and the data (Mobx store).

- The problem is that...
  - right now, D3.js is the "source of truth" and syncs one-way to the Mobx store (you draw something on D3.js, it tells the stores "this is what's happening")
  - This is fine in most cases, UNLESS you navigate away (to another FEM page) and _return_ to the Classify page, because _returning_ to the Classify page resets the D3.js graph but NOT the MobX stores.
- The planned solution is to draw all the "brushes" (i.e. visible annotations), based on existing MobX store data, _on component load._ (i.e. once when visiting/revisiting the Classify page)
  - This should "close the loop" for syncing the D3.js and the MobX stores, with the D3.js being the source of truth for all cases _except_ on component load, in which case the MobX store is the source of truth for one fleeting moment.

⚠️ ALSO I've done my best to clean up the code to modern FEM standards, but there's still a lot I don't want to touch.

### Testing

Testing URL: [Planet Finders Beta (staging)](https://local.zooniverse.org:3000/projects/nora-dot-test/planet-finders-beta/classify/workflow/3317)
- This is the precursor to [Planet Hunters TESS (production)](https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess)
- It's a Workflow with the LCV Subject Viewer enabled, and exactly 1 Task (dataVisAnnotation, with the tool graph2dRangeX), and specialised JSON-type Subjects.

Testing steps:
- Run `app-project` (not lib-classifier!) and go to the URL
- Open the Classify page
- Make 1 annotation
- Click on the project title to go to the Project home page
- EITHER press the 'Back' button on the browser OR click on the 'Classify' link, to return to the Classify page
- Observe that the 1 annotations you made is still visible
  - (On `master`, you will see 0 annotations instead)
- From here, you need to test two paths:
- PATH A: Submit the classification. The 1 annotation you made should be reflected in the submitted data.
- PATH B: Add another annotation
  - Observe that there are now 2 annotations
  - Submit the classification, and ensure all 2 annotations are reflected in the submitted data.

NOT in the testing scope:
- Workflows with multiple tasks, WFs with NO dataVisAnnotation tasks, etc.

### Status

WIP

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
@@ -180,11 +180,15 @@ class LightCurveViewer extends Component {
   initBrushes () {
     if (!this.annotationBrushes.length) {
       this.createAnnotationBrush() // Create the initial brush
-      this.updateAnnotationBrushes()
-    }
 
-    // WIP  // TODO
-    // - Create 'loadBrushesFromAnnotations()'
+      // Load any existing brushes from the stores. This is ONLY relevant when
+      // the user is re-visiting the Classifier after leaving
+      // mid-classification. In most normal cases (i.e. the classifier loads, or
+      // a new Subject is loaded), there are 0 existing brushes in the stores
+      this.loadBrushesFromAnnotations()
+
+      this.updateAnnotationBrushes()  // Draw!
+    }
 
     // TODO: Check for the following scenarios:
     // - Change of Subject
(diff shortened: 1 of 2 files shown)


shaunanoordin:
## PR Update

Solution is here!

- I've created a `loadBrushesFromAnnotations()`, the mirror universe counterpart to `saveBrushesToAnnotations()`
  - As mentioned under "planned solution", this grabs "annotation data" from the stores and plonks it into the D3.js graph as "annotation brushes" (i.e. those UI markings)
- On that note, I've also updated the LCV code to use more current methods of connecting to the annotation stores.

### Status

Ready for review!

There is this incredibly odd quirk where the brushes created on `loadBrushesFromAnnotations()` is created in _reverse layer order,_ (which results in restored brushes being _initially un-editable_ when returning Back to the Classify page, at least until you create a new brush) but I'm trying not to invest too much more time sanding off the edge(case)s here since it risks re-writing the whole LCV logic.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
```diff
@@ -228,13 +248,16 @@ class LightCurveViewer extends Component {
...
       id: nextAvailableId,
       brush: brush,
       minX: undefined, // x, relative to the data range (not the SVG dimensions)
       maxX: undefined,
       zoomLevelOnCreation: undefined
-    })
+    }
+
+    this.annotationBrushes.push(newAnnotationBrush)
+    return newAnnotationBrush
```
eatyourgreens, line 260: Out of interest, why does `saveBrushesToAnnotation` return a brush?

PR #2563, opened 2021-11-23
Classifier: Check bearer token on annotation update

eatyourgreens:
When we update an annotation, call [`authClient.checkBearerToken()`](https://github.com/zooniverse/panoptes-javascript-client/blob/9438015064bfd5fb0576b6554adf643b9003f3ca/lib/auth.js#L184-L192). `checkBearerToken()` will refresh the current token if it's close to expiring, allowing volunteers to stay authenticated for another two hours.

This should help volunteers stay logged in on projects with lengthy workflow tasks, like Scarlets & Blues. `annotation.update()` is used by all workflow tasks **except** creating marks from purple transcribed lines for the transcription task.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
I think this will be superseded by https://github.com/zooniverse/panoptes-javascript-client/pull/149, which fixes token refresh so that it works even if we leave several hours between checking the token status.

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`
Context: if a user spends a long time per Subject/Classification, their session might expire without any visual indicators.

This PR updates the Classifier so that logged-in users will be _less likely_ to be silently logged out between long classification sessions, and therefore _less likely_ to submit anonymous Classifications.

Code read looks good, and functionality appears to work fine, although I'm unable to do more thorough feature tests at the moment.

### Testing

I can confirm that checkBearerToken() is firing as expected, but trying to confirm that _checkBearerToken() will ensure a user's login state will continue to be refreshed_ will require far more testing time to simulate properly.

### Dev Notes

- This update works by adding a bearer token check (i.e. "are you still logged in?" check) every time an annotation updates (e.g. when a Single Question Task has an answer selected, or on _every key press_ of a Text Task)
- [checkBearerToken()](https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/oauth.js#L28-L37) works as follows:
  - When a user logs in, a token (read: blob of data) gets saved on local storage. This token has an expiry date on it.
  - When checkBearerToken() is called, it first checks if there's a valid un-expired token in local storage. If yes, that token is returned. (No further network calls are required.)
  - If no, a network call is made to fetch a fresh token from the (Panotes?) login/auth service.

Possible caveats:
- The implementation of checkBearerToken() doesn't _guarantee_ that a valid refreshed token will be returned by the time the user submits the classification, but it sure as heck tilts the odds in favour of a logged-in state.

Advanced notes:
- checkBearerToken() returns a Promise, and can be chained into something like... `authClient?.checkBearerToken().then(token => { if (user_is_logged_in_BUT_no_token_is_returned()) { then_alert_user_or_something() } })`
- That's the more advanced implementation if want to implement a more guaranteed "you can't submit unless you're logged in properly" check, but that's way, way, way beyond the scope of anything we'd want to tackle right now.

### Status

LGTM 👍 I say we get this merged & deployed so we can see that this improves volunteers' experience on those long-classification transcription projects.

PR #2579, opened 2021-12-10
Fix mode cookie and persist project store state with mst-persist

### LINKED ISSUE #2555: Dark theme cookies aren't applied instantly
## Package
app-project

## Describe the bug
In this video, I switch from dark mode to light, then back to dark mode again. After switching back to dark mode, light mode persists in the page props when I navigate to About and back to Classify.

https://user-images.githubusercontent.com/59547/142620576-d5e3d647-67c4-460b-9b35-c2a42baca74d.mov

## To Reproduce

1. Go to https://planethunters.org
2. Switch between light mode and dark mode.
3. Navigate between pages to check that downloaded page props match the theme mode cookie.

## Expected behavior
Theme changes should be applied instantly, and remembered across page transitions.

## Additional context
Requests for page props, like the following URL, are cached for 60s so there's probably a 60s lag between changing a cookie locally and seeing those changes reflected on our servers.
https://www.zooniverse.org/projects/_next/data/sveH-W7XOnvZ_IHl6uSt1/nora-dot-eisner/planet-hunters-tess/about/research.json?owner=nora-dot-eisner&project=planet-hunters-tess

I think we can fix this by reading `ui.mode` and `ui.dismissedAnnouncementBanner` from local cookies when the page loads, rather than getting them from page props.

Alternatively, use [`mst-persist`](https://github.com/agilgur5/mst-persist) to persist the store in the browser, including the UI settings.

eatyourgreens:
This adds [`mst-persist`](https://github.com/agilgur5/mst-persist) to `app-project`, which will persist the MST store in local storage, then reload it when you re-enter a page.

I've removed server-side cookie parsing, along with the `initialState.ui` prop, so that the local store state becomes the source of truth for UI settings like the theme mode. This should fix #2555, where cached server-side props are overriding local settings immediately after you change them. It does mean that SSR only renders light-mode pages, so dark mode users may get an initial flash of light background before the store hydrates.

Since this uses local storage, it does mean that store state is shared across open tabs. I've added the project ID to the storage key, to allow different projects to be open in different tabs, all sharing the same `localStorage`.

Package:
app-project

Closes #2555.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/package.json
+++ b/packages/app-project/package.json
@@ -51,6 +51,7 @@
     "mobx-react": "~7.2.1",
     "mobx-state-tree": "~5.1.0",
     "morgan": "^1.10.0",
+    "mst-persist": "~0.1.3",
     "newrelic": "~8.7.0",
     "next": "~12.0.4",
     "next-i18next": "~10.0.1",
--- a/packages/app-project/pages/_app.js
+++ b/packages/app-project/pages/_app.js
@@ -1,6 +1,7 @@
 import { Box } from 'grommet'
 import makeInspectable from 'mobx-devtools-mst'
 import { Provider } from 'mobx-react'
+import { persist } from 'mst-persist'
 import Error from 'next/error'
 import { useEffect, useMemo } from 'react'
 import { createGlobalStyle } from 'styled-components'
@@ -34,11 +35,20 @@ function useStore(initialState) {
 function MyApp({ Component, pageProps }) {
   const isServer = typeof window === 'undefined'
   const { initialState } = pageProps
-  const store = useStore(initialState)
+  let store = useStore(initialState)
   makeInspectable(store)
 
-  function onMount() {
+  async function onMount() {
     console.info(`Deployed commit is ${process.env.COMMIT_ID}`)
+    if (window?.localStorage && initialState?.project) {
+      const key = `project-${initialState.project.id}`
+      await persist(key, store)
+      console.info('Store hydrated from local storage.')
+      /*
+       Apply page props, in case stored state is stale.
+      */
+      store = initStore(false, initialState)
+    }
     /*
       Another project may have set the mode cookie, so check it now.
     */
@@ -49,7 +59,9 @@ function MyApp({ Component, pageProps }) {
     store.user.checkCurrent()
   }
 
-  useEffect(onMount, [])
+  useEffect(() => {
+    onMount()
+  }, [])
 
   try {
     if (pageProps.statusCode) {
--- a/packages/app-project/pages/_app.js
+++ b/packages/app-project/pages/_app.js
@@ -1,7 +1,6 @@
 import { Box } from 'grommet'
 import makeInspectable from 'mobx-devtools-mst'
 import { Provider } from 'mobx-react'
-import { persist } from 'mst-persist'
 import Error from 'next/error'
 import { useEffect, useMemo } from 'react'
 import { createGlobalStyle } from 'styled-components'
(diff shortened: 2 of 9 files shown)


shaunanoordin:
## Quick ~~Review~~ Notes

My Next.js skills are fairly limited, so I won't be able to pick up on any server-rendering niche scenarios that may happen, but the general code read looks good. I do have some dev notes:

**Note 1:** Each _Zooniverse Project_ will have its own store, OK. So if I'm reading this correctly: if I enable 'dark mode' in WildWatch Cotswolds, then that'll persist for all WildWatch Cotswolds pages (even if opened in a new tab). If I then navigate to a new project, e.g. GalaxyGrue, that project will be in Light Mode (default). Cool.

If I open a new browser tab with WildWatch Cotswolds, that tab should also be in Dark Mode (also, cool). If I change to Light Mode in WildWatch Cotswolds Tab 2, then look at Tab 1, then... hmm? I guess it'll switch to Light Mode, but only after an activity/action causes the store to update? **Note 1a:** note to self, test out activity on the _same project across multiple browser tabs._

Also, if I'm reading this correctly: the store _is consistent across users,_ so if I'm logged in as Alice at WWCotswolds and turn on Dark Mode, then log out and log in as Bob, then WildWatch Cotswolds should still be in Dark Mode. Cool.

**Note 2:** We're adding a new dependency, `mst-persist`, which was [last updated](https://www.npmjs.com/package/mst-persist) in _Dec 2019._ I'm not sure how to feel about this. On one hand, adding a dep that's no longer updated is going to bite us in the long run, e.g. when a Dependabot update tells us "mst-persist needs to update sub-dependency foobar to >= v8.0.0", or when mst-persist is no longer compatible with mobx-state-tree v7.7.7-omicron etc etc.

On the other hand, _I can't find an alternative to mst-persist._ So, it's better to have a solution now which we might need to revisit later, than to have no solution at all, right?

This won't be a blocker in my proper review, but I will view the dependency's "last publish" field date with some trepidation. ☢️

eatyourgreens:
`persist` is a single function, with no dependencies, so should be safe as long as the MST API stays stable.

shaunanoordin:
Update on **Note 2:**
> persist is a single function, with no dependencies, so should be safe as long as the MST API stays stable.

My response: Sweet! 👍 

**Note 3:** The local storage key looks like `project-123`, which (in very narrow circumstances) may cause a collision when you're viewing `/projects/shaun/project123?env=staging` and `/projects/shaun/project123?env=production`.

- Possible solution: append the env, so the key looks like `project-123-staging`.
- Worry level: low. (Normal users will almost never encounter this)

**Note 4:** will there be an issue with "Project Stores" filling up the user's localStorage? _(assumption: window.localStorage has a finite size limit per domain)_

Some quick maths: 
- The store for Scarlets & Blues (localStorage key: `project-12268`) is 27872 characters, which translates to something like 28KB of data.
- Galaxy Zoo is 47775 characters, which is about 27KB of data. So let's say 40KB is the average
- Most web browsers have a default local storage capacity of 5MB [(maybe?? I think???)](https://www.google.com/search?q=local+storage+max+size) so - if we ignore other stored data such as "failed Classifications queue" (which can be huge!) - then a user would on average be able to view 128 before weird data shenanigans happen.
  - EDIT: [maybe 10MB](https://en.wikipedia.org/wiki/Web_storage), so that's 256 projects.

So an average user should be able to view a lot of projects with no issues, but I still feel this might be important to note, since we're not doing anything to clean up/remove past projects from local storage.

Thought 1: we could move the store from window.localStorage to window.sessionStorage, so that'll clean up old projects that are no longer relevant to the user. Counterpoint: wait, this defeats the purpose of using storage to remember the "you've dismissed the banner" and "Light/Dark Mode" settings.

Thought 2: what if we _only persisted_ `store.ui` into window.localStorage, and leave `store.project` and `store.user` as is? This might make some sense, since having _transitional_ data like `store.project.loadingState = "success"` saved into _permanent_ local storage seems _weird._ Counterpoint: I have no idea how to split the store in this context or if it's even possible.

Worry level: moderate discomfort at unverified hypothetical scenarios. Could be valid, could be paranoia.

**Note 5:** What happens if the localStorage data is corrupted for whatever reason? 

- If mst-persist knows how to handle invalid JSON, (e.g. if JSON invalid, just treat as if store doesn't exist). This should be easy to check by just borking my localStorage data.
- Worry level: low

eatyourgreens:
Check out #2627 for an implementation that uses session storage. Good point about how long this data should persist for.

eatyourgreens:
Rebased to the latest master, which includes localisation.

eatyourgreens:
This branch has gotten out of synch with the main branch. I'm going to split the fix for theme cookies out into its own, smaller PR.

eatyourgreens:
Closing in favour of #2947.

PR #2597, opened 2022-01-04
Classifier: Export tasks by task type and remove TaskRegistry

eatyourgreens:
Refactor `@plugins/tasks` to organise tasks by task type. Use the task type as the named export for each task, which then allows us to remove the task registry.

When I added `@plugins/tasks` to the classifier, I added the `TaskRegistry` object to map task types (from Panoptes workflows) to task components and models. However, the register is kind of clunky and I think it's easier to import tasks directly as named imports without having to register them in an external directory eg.
```js
import * as tasks from '@plugins/tasks'
const singleChoiceTask = tasks.single
```
or maybe even use dynamic imports eg.
```js
const type = 'single'
const singleChoiceTask = await import(`@plugins/tasks/${type}`)
```

This PR removes the registry and reorganises tasks by task type. The only problem I've run into is that `dropdown-simple` isn't a valid identifier in JS. I've mapped that to `tasks.dropdownSimple` but there might be better ways to handle dropdown tasks. I'd be interested in @shaunanoordin's thoughts there, as the author of the original simple dropdown task.

One advantage of this, I think, is that unique task types are automatically enforced by the file system. You can't accidentally create a new task model with an existing type. The disadvantage, I think, is that the type has to be entered in multiple places: the named export, the task folder name and the task type in the task model.

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


mcbouslog (approved):
I think refactoring to remove the TaskRegistry and importing tasks by type is a good idea. I've reviewed the code and it looks good to me, largely just renaming and reorganizing.

I've confirmed the following tasks/projects work as expected:
- [x] single question
- [x] multiple question
- [x] drawing (point, line, circle, rectangle)
- [x] text
- [x] light curve viewer (data vis)
- [x] dropdown
- [x] transcription
- [x] survey
- [x] I fancy cats
- [x] TESS
- [x] HMS NHS
- [x] Corresponding with Quakers

eatyourgreens:
Let’s try this out on the main branch.

PR #2627, opened 2022-01-13
Classifier: persist fetched data with mst-persist

### LINKED ISSUE #2610: Browser back button leads to already seen subject

eatyourgreens:
This uses [`mst-persist`](https://github.com/agilgur5/mst-persist) to store partial snapshots of the classifier store in session storage. When you leave the classifier and return during a session (eg. by visiting Talk or reloading the page), the project, field guide, tutorials and workflows should be fetched from session storage. You can check this by watching network requests after reloading the page. On the return visit, only the subject queue and user preferences should be requested from Panoptes.

Longer term, we could consider also remembering the subject queue across visits, which would fix #2610. That needs some thought, though, since we don't want to save a stale subject queue forever, and we do want reloading the page to choose a new random subject for projects like PH-TESS. I've left the `subjects` store out of this PR because preserving it raises these questions, and is a bit tricky to implement with the current state tree.

Test URL: HMS NHS (classifier state should be saved across page reloads, even after switching workflows or switching subject sets within a workflow.)
https://local.zooniverse.org:3000/projects/msalmon/hms-nhs-the-nautical-health-service/classify?demo=true&env=production

Package:
lib-classifier

Towards #2610 (but doesn't fix it here.)

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


eatyourgreens:
I've tested locally with I Fancy Cats and HMS NHS< and haven't seen any problems. There's one failing test, however, because the mocked store loads field guides twice.

packages/lib-classifier/src/plugins/drawingTools/models/tools/Tool/Tool.js
```diff
@@ -26,6 +26,12 @@ const Tool = types
     /*
     Create tasks from details if we have details but no tasks.
     */
+    if (!snapshot.max) {
```
shaunanoordin, line 29: [minor][rambling] Sanity check: this implies that a Drawing Tool can never have "maximum number of marks = 0". Response: that's legit.

shaunanoordin: [minor][rambling] I just realised it's theoretically possible to have odd logic like max and min being negative numbers, or max < min, but those logic checks aren't actually necessary since if a drawing tool is _that_ borked, the Project Owner would know pretty darn quickly.

packages/lib-classifier/src/components/Classifier/Classifier.js
```diff
@@ -83,50 +86,62 @@ export default function Classifier({
...
-    workflows
-  } = classifierStore
+  const [loaded, setLoaded] = useState(false)
 
-  useEffect(function onMount() {
+  async function onMount() {
+    try {
+      const storageKey = `fem-classifier-${project.id}`
+      await persist(storageKey, classifierStore, {
+        storage: asyncSessionStorage,
```
shaunanoordin, line 98: ✅  Dev Note: this where we inform `mst-persist` to use window.sessionStorage. By default, mst-persist uses _localStorage,_ which has ramifications on long-term storage

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR adds persistent storage to the Classifier, on a **per-project** basis, for the duration of the **session**.
- What this means is that when the Classifier loads for a specific project, the MST store data (specifically: _store.fieldGuide, store.projects, store.subjectSets, store.tutorials, store.workflows_ ) are all saved to window.sessionStorage.
- When the Classify page (for that project) is revisited after first navigating away, the saved data is loaded back into the Classifier.

Code read looks good. I have some notes from #2579 that I want to repeat & revise here:
- 🕐 Duration of storage: by default, mst-persist stores data to localStorage, which raises the issue of "how long are we keeping this transient data for?". However, this implementation uses sessionStorage, which sidesteps a lot of issues. 👍
  - ~~The storage key can have collision issues when visiting a staging project and a production project that share the same ID.~~ Not likely given the temporary-ness of sessionStorage.
- ☢️ `mst-persist` was last updated Dec 2019, which makes this a no-longer-supported dependency which can cause potential problems down the line. (e.g. if MST has a major upgrade with breaking changes)
  - This was raised in 2579 and recognised as an accepted risk.

### Testing

Testing is limited as the effects of persistent data would most noticeably be felt on a live server, running the full app-project/Next.js setup, where I'm able to navigate to non-Classify pages.

I'm able to compile and run the Classifier with no issues on localhost ([random project,](https://local.zooniverse.org:8080/?env=staging&project=1651&workflow=3262&demo=true) macOS+Chrome) so that's a bit reassuring.

### Status

LGTM 👍

PR #2657, opened 2022-01-19
Classifier: Persist the subjects store in session storage

### LINKED ISSUE #2610: Browser back button leads to already seen subject

### LINKED ISSUE #2690: Cached Stores: when switching Workflows from the Project home page, the wrong Subject can appear on the wrong Workflow
## ~~Major~~ Situational Bug

Package: `app-project`
Initial discovery: #2677
UPDATE: only affects projects that use cached stores (i.e. mst-persist)

If I switch between two workflows _on the Project home page_, I can cause the Subject Set of Workflow A to appear while classifying Workflow B
- The trigger is very very weird. You need to start at the _Project home page,_ open WF A, go back to the home page, open WF B, back to the home page, then open WF A. 
- The issue doesn't trigger if you switch workflows from the `/classify` page 🤷 
- ⚠️ Again, I'm not sure if this is related to `swr` or `mst-persist`! Additional tests tomorrow.

Steps to replicate:
- In a new incognito window, open the test project home page: https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022
- click the Transformers WF
- click the project title to return to the home page
- click the Ultraman WF
- click the project title to return to the home page
- click the Transformers WF
  - Expected: a Transformers subject should be displayed alongside the Transformers task
  - Actual: an Ultraman subject is displayed alongside the Transformers task

image

_Above: that's an Ultraman, not a Transformer! After switching back and forth between the Transformers WF and the Ultraman WF, I'm able to confuse the Classifier to serve me Ultraman subjects with Transformers tasks._

### Status

~~I'm marking this as a major bug. Investigation ongoing.~~

UPDATE: This is now a situational bug, not a major bug.

eatyourgreens:
Follows on from #2627.

This PR fixes #2610, where volunteers are shown seen subjects on returning from Talk, because the subject workflow observer in the classifier requests a new subject queue from Panoptes and the first subject returned is the last subject that they classified.

Here, `store.subjects` and `store.workflowSteps` are persisted in session storage, so that the new active subject should be remembered for prioritised workflows, and rendered in the classifier after returning from Talk.

I've added support for `workflow.prioritized` to the project app, and enabled `cachePanoptesData` for all prioritised workflows.

TODO:
- [x] Reset the subject queue in `workflows.selectWorkflow(workflowID, subjectSetID)`. Remove `workflow.selectSubjects()`. Remove the subject workflow observer, which is the cause of #2610.
- [x] Persist `store.subjects` and manually start a new classification when cached subjects load.
- [x] Enable `cachePanoptesData` for prioritised workflows.

Package:
app-project
lib-classifier

Closes #2610.
Closes #2690. 

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
```diff
@@ -120,29 +120,11 @@ const SubjectStore = types
       next(call)
     }
 
-    function _onWorkflowChange() {
-      const workflow = tryReference(() => getRoot(self).workflows.active)
-      const subjectSet = tryReference(() => workflow?.subjectSet)
-      if (workflow || subjectSet) {
-        self.reset()
-        self.populateQueue(workflow.selectedSubjects)
-      }
```
eatyourgreens, line 129: It would be nice if we could keep this, but modify it so that it doesn't run if we load in an existing subject store from a snapshot for a prioritised workflow. That would remove a bunch of extra conditional code from the classifier, which now manually resets the queue when the workflow or subject set or initial subject changes.

packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
```diff
@@ -332,6 +322,7 @@ const SubjectStore = types
     }
 
     function reset () {
+      // self.active = undefined
```
eatyourgreens, line 325: Oops. This can go.

eatyourgreens:
I think this works for a single workflow, but breaks, with a store crash, when you change workflow (notebook) on Davy Notebooks.

eatyourgreens:
Hard to tell, on Davy Notebooks, whether the correct page is loading for each notebook, but I'm not seeing any problems.I think this is ready for review.

packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
eatyourgreens: There's two cases where we'd want to reset and repopulate the queue:
1. There's a workflow (and subject set for grouped workflows) but `self.resources` is empty, so we need to load subjects.
2. `self.resources` is not empty, but the subjects are not valid for the current workflow (or set.) This is the case that's impossible to validate with the current model.

mcbouslog (approved):
This LGTM! 🚀 
- went through the steps to replicate the wrong subject in a workflow per #2690 and did _not_ (👍 ) see the wrong subject in a workflow on this branch
- ClassifierContainer nicely separates code
- other than separating code between ClassifierContainer and Classifier, changes to code largely only apply to if `cachePanoptesData`, code changes LGTM
- workflow select in lib-classifier dev will certainly be helpful

confirmed:
- issue on production with HMS NHS (subject-talk-back=>already seen subject)
- issue fixed on local with HMS NHS (subject-talk-back=>new subject)
- no issue on local with Davy Notebooks (subject-talk-back=>new subject)
- no issue on production and local with TESS (subject-talk-back=>new subject)
- on local TESS feedback works

could not confirm already seen issue on Davy Notebooks, strange, but based on all of above that's ok

Issue #2670, opened 2022-01-21
Classifier: Classifier 'persists' the tutorial for the first Workflow in a session


shaunanoordin:
## Classifier Bug

Package: `lib-classifier`
Related: #2668

This bug affects FEM Projects with multiple WFs that each have a distinct tutorial. When you load the Classifier for a WF, that tutorial is "remembered". When you switch to another WF in the same session, you will still see the first tutorial.

- This is related to the `mst-persist` update added in #2627, since the Tutorial data is stored in localStorage after the initial fetch

![image](https://user-images.githubusercontent.com/13952701/150586150-2144cbf8-f7af-45c3-9eb1-14a422a12cee.png)

_In the example above, after visiting the Transformers WF first (and getting the Transformers tutorial with Optimus Prime's face, I went changed WF to the Ultraman WF. I'm still seeing the Transformers tutorial._
- Note that reloading the page does NOT fix the issue. You'll need to close the tab and open it again, or clear your session storage. See 2668 for the discussion on this.

### Testing

Setup:
- Create a Project with at least 2 distinct Workflows, and 2 distinct tutorials.
- Assign one tutorial to a WF

Steps:
- Open the Project home page
- Select WF 1. Observe that you'll see Tutorial 1.
- Click on Classify then select WF 2
- Note that while the Classifier correctly loads WF 2's Tasks and Subjects, the tutorial is still Tutorial 1.

Tested on staging (929e8f1) with macOS + Chrome (incognito window)

Example project: https://frontend.preview.zooniverse.org/projects/darkeshard/test-project-2022?env=staging
- This project has the "Transformers WF" which has a tutorial featuring an image of Optimus Prime, and the "Ultraman WF" which has a tutorial featuring the titular Ultraman.

### Status

Medium priority bug(? Any arguments for high priority?) It's definitely a problem if we're showing the wrong tutorials to volunteers, though the conditions for this happening isn't so rampant that it might affect too many live projects right now. _(I think.)_

eatyourgreens:
The only project I can think of is Davy Notebooks, which has one tutorial that’s linked to every workflow, so isn’t affected. 

I think this should be fixed ASAP if it is affecting live projects.

shaunanoordin:
## Investigation Update

[...]

PR #2677, opened 2022-01-24
Classifier: Refresh workflows with SWR (stale-while-revalidate)

### LINKED ISSUE #2668: `mst-persist` changes expected "Reload Page" behaviour in Classifier (e.g. workflow changes not showing immediately)
## Functionality Bug/Quirk(?)

Package: `lib-classifier`

As a result of `mst-persist` in #2627, reloading the Classifier does not fetch any _recent changes_ to the Workflows.

**Example:**
- As a developer, let say I open the Classifier to work on Workflow 1234, with a task asking "How many CATS do you see?"
The Classifier will render WF1234 and the task properly. 🆗 
- Then, let's say I edit WF1234 (in another window) so task now asks "How many DOGS do you see?"
- If I reload the Classifier (or 'empty cache and hard reload' in Chrome), the task will continue to ask "How many CATS do you see?" ❌ 

**Explanation:**
This is, of course, because the WF data is stored in sessionStorage (which isn't wiped even by a hard reload), and the Classifier has no idea there's a newer version of the WF data on the server. (See FEM PR #2627)

Solution/workaround is simple: open a new tab, instead of just reloading. (Or clear your sessionStorage.)

That said, I can see how this can be very confusing, since the persist-in-localStorage is basically _overriding the expected behaviour of the Reload Page action._

I don't believe this will significantly affect volunteers (unless we have volunteers who never close their tabs) but it's worth keeping some eyes on this. Project owners might  be a bit more affected, as Delilah pointed out. (i.e. they'd have one tab open to edit the project, and another tab with the project which they'll keep refreshing, only to see no changes.)

### Status

Medium priority. We have a simple workaround, but that's not yet communicated with the volunteers.

Work with `mst-persist` is ongoing; tagging @eatyourgreens on the matter.

eatyourgreens:
Add the [`useSWR`](https://swr.vercel.app/) hook to the root `Classifier` and an `onWorkflowChange` effect which updates the store when workflow data changes in Panoptes.

Package:
lib-classifier

Closes #2668.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/Classifier.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.js
@@ -3,6 +3,7 @@ import { Paragraph } from 'grommet'
 import makeInspectable from 'mobx-devtools-mst'
 import { Provider } from 'mobx-react'
 import { persist } from 'mst-persist'
+import useSWR from 'swr'
 import PropTypes from 'prop-types'
 import React, { useEffect, useMemo, useState } from 'react'
 import zooTheme from '@zooniverse/grommet-theme'
@@ -95,6 +96,11 @@ export default function Classifier({
   })
 
   const [loaded, setLoaded] = useState(false)
+  const { data } = useSWR(`/workflows/${workflowID}`, client.panoptes.get)
+  let workflowData
+  if (data?.text) {
+    workflowData = data.text
+  }
 
   async function onMount() {
     if (cachePanoptesData) {
@@ -144,6 +150,16 @@ export default function Classifier({
       workflows.selectWorkflow(workflowID, subjectSetID, subjectID)
     }
   }, [loaded, subjectID, subjectSetID, workflowID])
+
+  useEffect(function onWorkflowChange() {
+    const { workflows, subjects } = classifierStore
+    if (loaded && workflowData) {
+      const [ workflowSnapshot ] = JSON.parse(workflowData).workflows
+      workflows.setResources([workflowSnapshot])
+      // TODO: the task area crashes without the following line. Why is that?
+      subjects.setActiveSubject(subjects.active?.id)
+    }
+  }, [loaded, workflowData])
 
   useEffect(function onAuthChange() {
     if (loaded) {
--- a/packages/lib-classifier/package.json
+++ b/packages/lib-classifier/package.json
@@ -40,6 +40,7 @@
     "react-i18next": "~11.15.1",
     "react-player": "~2.7.0",
     "react-resize-detector": "~7.0.0",
+    "swr": "~1.1.2",
     "valid-url": "~1.0.9"
   },
   "peerDependencies": {
--- a/packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
+++ b/packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
@@ -130,25 +130,9 @@ const SubjectStore = types
     }
 
     function afterAttach () {
-      createWorkflowObserver()
-      createClassificationChangeObserver()
+      addDisposer(self, autorun(_onWorkflowChange))
+      addDisposer(self, autorun(_onClassificationChange))
       addMiddleware(self, _addMiddleware)
-    }
-
-    function createWorkflowObserver () {
-      const workflowDisposer = autorun(
-        _onWorkflowChange,
-        { name: 'SubjectStore Workflow Observer autorun' }
-      )
-      addDisposer(self, workflowDisposer)
-    }
-
-    function createClassificationChangeObserver () {
-      const classificationDisposer = autorun(
-        _onClassificationChange,
-        { name: 'SubjectStore Classification Change Observer autorun' }
-      )
-      addDisposer(self, classificationDisposer)
     }
 
     async function _fetchPreviousSubjects(workflow, priority) {
--- a/packages/lib-classifier/src/store/WorkflowStepStore/WorkflowStepStore.js
+++ b/packages/lib-classifier/src/store/WorkflowStepStore/WorkflowStepStore.js
@@ -74,19 +74,16 @@ const WorkflowStepStore = types
     }
   }))
   .actions(self => {
-    function afterAttach () {
-      createWorkflowObserver()
+    function _onWorkflowChange() {
+      const workflow = tryReference(() => getRoot(self).workflows?.active)
+      if (workflow) {
+        self.reset()
+        setStepsAndTasks(workflow)
+      }
     }
 
-    function createWorkflowObserver () {
-      const workflowDisposer = autorun(() => {
-        const workflow = tryReference(() => getRoot(self).workflows?.active)
-        if (workflow) {
-          self.reset()
-          setStepsAndTasks(workflow)
-        }
-      }, { name: 'WorkflowStepStore Workflow Observer autorun' })
-      addDisposer(self, workflowDisposer)
+    function afterAttach () {
+      addDisposer(self, autorun(_onWorkflowChange))
     }
 
     function getNextStepKey () {
--- a/yarn.lock
+++ b/yarn.lock
@@ -16075,6 +16075,11 @@ svg-loaders-react@~2.0.1:
   resolved "https://registry.yarnpkg.com/svg-loaders-react/-/svg-loaders-react-2.0.1.tgz#67b50783b1dd0f9bb845a7248e46290ba828fc06"
   integrity sha512-7oOA29mMLPuGNZ5UkyprH/4RysYy6F19/P8cr5mbrWJeNseH01ih5YpynNFhgpvUaQ8h6vStMS8wd3IFELcFYA==
 
+swr@~1.1.2:
+  version "1.1.2"
+  resolved "https://registry.yarnpkg.com/swr/-/swr-1.1.2.tgz#9f3de2541931fccf03c48f322f1fc935a7551612"
+  integrity sha512-UsM0eo5T+kRPyWFZtWRx2XR5qzohs/LS4lDC0GCyLpCYFmsfTk28UCVDbOE9+KtoXY4FnwHYiF+ZYEU3hnJ1lQ==
+
 symbol-tree@^3.2.4:
   version "3.2.4"
   resolved "https://registry.yarnpkg.com/symbol-tree/-/symbol-tree-3.2.4.tgz#430637d248ba77e078883951fb9aa0eed7c63fa2"


eatyourgreens:
Some nice features of `swr`:
- it handles cache validation for us.
- workflows are automatically fetched, in the background, when they change in Panoptes.
- data refreshes are fired by [`onvisibilitychange`](https://developer.mozilla.org/en-US/docs/Web/API/Document/onvisibilitychange), so workflows update when you change active tab or wake a browser from sleep.


https://user-images.githubusercontent.com/59547/150809890-fc421345-488d-4d32-bdd5-8d750fe35584.mov

packages/lib-classifier/src/components/Classifier/Classifier.js
```diff
@@ -94,6 +95,11 @@ export default function Classifier({
   })
 
   const [loaded, setLoaded] = useState(false)
+  const { data } = useSWR(`/workflows/${workflowID}`, client.panoptes.get)
```
eatyourgreens, line 98: We could add an access token to this request, so that project owners can preview private workflows in the new classifier.

packages/lib-classifier/src/components/Classifier/Classifier.js
```diff
@@ -94,6 +95,11 @@ export default function Classifier({
...
+  let workflowData
+  if (data?.text) {
+    workflowData = data.text
```
eatyourgreens, line 101: `workflowData` here includes the full workflow resource, including `workflow.classifications_count` and `workflow.completeness`. I think that should be ok, as they aren't stored in the MobX data model, but it's worth double-checking that changes to the workflow counters don't trigger a full refresh of the classifier, via `useEffect`.

eatyourgreens: eg. someone else classifying, elsewhere in the world, and updating the total classification count shouldn't cause the workflow to reload in my browser and reset my work in progress.

packages/lib-classifier/src/components/Classifier/Classifier.js
```diff
@@ -137,10 +143,11 @@ export default function Classifier({
...
   useEffect(function onURLChange() {
     const { workflows } = classifierStore
-    if (loaded && workflowID) {
-      workflows.selectWorkflow(workflowID, subjectSetID, subjectID)
+    if (loaded && workflowData) {
+      const [ workflowSnapshot ] = JSON.parse(workflowData).workflows
+      classifierStore.refreshWorkflow(workflowSnapshot, subjectSetID, subjectID)
     }
-  }, [loaded, subjectID, subjectSetID, workflowID])
+  }, [loaded, subjectID, subjectSetID, workflowData])
```
eatyourgreens, line 150: `onURLChange` here runs whenever any part of `workflowData` changes. Maybe we should restrict that to only a subset of workflow properties.

shaunanoordin:
## Quick Thoughts

- Code read looks good 👍 If I'm understanding this correctly...
  - _Previously,_ the "fetch workflow pls" action was initiated on the `onURLChange()` side effect, i.e. when the code detects "ahhah, the classifier is fully loaded and I have a WF ID" 
  - _Now,_ the "fetch workflow pls" action is initiated on `const { data } = useSWR(`/workflows/${workflowID}`, client.panoptes.get)`, which would notably trigger the first time the component is rendered(? I think)
  - ❔ This does imply that the "fetch workflow pls" action can trigger _before_ the Classifier is "fully loaded" (i.e. before the onMount() registers onSubjectReset and onSubjectChange) but I don't think this will have any adverse effects on the Classifier store behaviour(? Question mark? (Probably not (I can't keep track of the auto-actions and observers in the store TBH)))
- [ ] **Note to self/TODO:** figure out what happens if we load the Classifier when workflowID isn't defined. Previously, the fetch (in onURLChange()) simply wouldn't trigger. With this PR... will useSWR() try to fetch from _panoptes/api/workflows/_ or maybe _panoptes/api/workflows/undefined_ ?
- **[swr](https://www.npmjs.com/package/swr)** looks like a very impressive solution, as it abstracts away a lot of the nitty-gritty of the caching mechanics. 👍 
  - I'm noting that `useSWR()` also returns an optional `error` value, which we're not using to detect errors.
  - (Not sure if this is worth me testing for errors though, and if so, to what extent. 🤔 )

shaunanoordin:
## Review Notes

I've discovered some bugs but I'm not 100% sure if they're specific to this PR, or something to do with `mst-persist`. I'll continue testing tomorrow, but here are my notes so far.

The following were all tested on `app-project` on localhost with macOS + Chrome, on test project https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022

New behaviour: Classifier will periodically fetch **workflow data** from the API, to ensure the WF is up to date. ✅ 
- the periodic "refreshing" fetch (as opposed to the initial data fetch) will trigger whenever the window/tab re-gains visibility and/or when the page is reloaded.
- if the fetched data has no changes, the Classifier's Tasks, etc remain the same. If there are differences, the Classifier's Tasks will update accordingly.
  - ⚠️ Untested: what happens if the user has an active Classification when this happens. **Shaun's TODO:** test this with something big like a Transcription Task

The new behaviour looks good 👍 I've tested it by editing WF X on the Project Builder in one Chrome tab, and classifying on WF X on another Chrome tab.

**Issue 1 (minor?):** the Classifier will sometimes make calls to `/api/workflows/undefined` (an invalid API URL)
- If you open a project's Classify page - e.g. https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022/classify - the Classifier will make an incorrect call to `panoptes-staging.zooniverse.org/api/workflows/undefined`
- This is due to `useSWR()` triggering a fetch even if WorkflowID is `undefined`
- From a volunteer perspective, this error isn't visible and doesn't affect the user experience. (Therefore, it's a minor issue.)
- From a backend perspective though, this means the Classifier is sending junk requests to the servers for no good reason. (Therefore, I'm adding a question mark to the _minor?_ issue label)

**Issue 2 (major?):** If I switch between two workflows _on the Project home page_, I can cause the Subject Set of Workflow A to appear while classifying Workflow B
- The trigger is very very weird. You need to start at the _Project home page,_ open WF A, go back to the home page, open WF B, back to the home page, then open WF A. 
- The issue doesn't trigger if you switch workflows from the `/classify` page 🤷 
- ⚠️ Again, I'm not sure if this is related to `swr` or `mst-persist`! Additional tests tomorrow.

Steps to replicate (`app-project`):
- In a new incognito window, open the test project home page: https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022
- click the Transformers WF
- click the project title to return to the home page
- click the Ultraman WF
- click the project title to return to the home page
- click the Transformers WF
  - Expected: a Transformers subject should be displayed alongside the Transformers task
  - Actual: an Ultraman subject is displayed alongside the Transformers task

image

_Above: that's an Ultraman, not a Transformer! After switching back and forth between the Transformers WF and the Ultraman WF, I'm able to confuse the Classifier to serve me Ultraman subjects with Transformers tasks._

packages/lib-classifier/src/store/RootStore.js
```diff
@@ -98,6 +98,12 @@ const RootStore = types
       onPatch(self, _onPatch)
     }
 
+    function refreshWorkflow(workflowSnapshot, subjectSetID, subjectID) {
+      self.workflows.setResources([workflowSnapshot])
+      self.workflows.selectWorkflow(workflowSnapshot.id, subjectSetID, subjectID)
```
eatyourgreens, line 103: I think this promise needs to resolve with a new active workflow before starting a new classification. Otherwise the subject queue might be advanced for the previous workflow.

eatyourgreens:
@shaunanoordin when you change workflow on the classify page, you apply a new `workflowID` prop to the mounted `Classifier` component. When you go to the home page and return, you unmount the `Classifier`, then mount it again with a new `workflowID` prop.

packages/lib-classifier/src/store/RootStore.js
eatyourgreens: @shaunanoordin this is why you're seeing stale subjects in the queue after a workflow change. `onSubjectAdvance()` is running before the workflow has changed.

eatyourgreens:
I added some logging to the classifier, then ran a production build locally and tracked what happens for a couple of workflows on my test project. Feel free to jump in if I've got anything wrong here, but this is what I think happens when the classifier starts up:
1. A new, empty store is created.
2. SWR requests the workflow from Panoptes
2. A snapshot is read asynchronously from session storage and applied to the store. The `loaded` flag blocks `useEffect` from running until the snapshot is loaded.
3. The hydrated workflow fires reactions in `store.workflowSteps` and `store.subjects`, which populate the classification tasks and subject queue. Requests are sent to Panoptes for the _stored workflow_.
4. Hydration finishes and `loaded` is set to true.
4. `onURLChange` (which might need to be renamed) runs and applies the SWR data to the workflows store.
5. At this point, two things might happen. Either:
    1. The SWR data is an update for the stored workflow. The active workflow in the model is refreshed with new values from Panoptes. The workflow reactions don't seem to fire in this case eg. if the task instructions update. I might need to look into this further.
    1. The SWR data is a completely new workflow. The active workflow changes to this new workflow, firing reactions which reset the workflow steps and the subjects store. Requests are sent to Panoptes for the _SWR workflow_.

shaunanoordin:
@eatyourgreens your [rundown of the process](https://github.com/zooniverse/front-end-monorepo/pull/2677#issuecomment-1021378585) is on point 👍 

One thing I missed during my initial read, which your explanation elucidated me to, was that even though the workflow data is now _fetched before onMount(),_ the WF data isn't registered into the store until _onMount() is complete_ (i.e. loaded = true). That makes sense 👌 

In other news:

### Debug Update

So I've done some additional tests, and turns out **Issue 2** (switching between workflows causes wrong Subjects to appear) is **currently extant on master.** (commit dfe976e, at least) I'll open an issue to track that bug, but at least we can rule out `swr` causing that problem.

(Update: I've also confirmed that the issue is still extant in this branch, at commit 3a267a1)

I can't think of any other tests I can throw at `swr` - the only question I have is, how do we want to handle Issue 1 (if you go to /classify, you'll pink /workflows/undefined) ? Full review incoming.

eatyourgreens:
If that workflow is on master and affecting live workflows, then does that mean classifications are being submitted for the wrong subjects on live projects?

Does #2676 fix it, by turning off session storage?

packages/lib-classifier/src/store/RootStore.js
shaunanoordin: Oooh I might need to trace `onSubjectAdvance()` then, as I'm honestly not 100% sure of the sequence of triggers here.

One thing to note though is that if `onSubjectAdvance()` is the problem, it might not just be contained to the call in `refreshWorkflow()`. [My recent tests](https://github.com/zooniverse/front-end-monorepo/pull/2677#issuecomment-1021612402) have shown that that the problem (Issue 2) is extant in _master_ as well.

shaunanoordin:
@eatyourgreens 

> Does 2676 fix it, by turning off session storage?

That'll be my next test. 👌  I'll wrap up this 2677 review first, then create an Issue for Issue 2, then post an update on whether 2676 takes care of the problem.

eatyourgreens:
If #2676 does fix the problem, it should be deployed ASAP. 

I think I see what might be happening here. When the store hydrates from a snapshot, reactions run which fetch subjects for the workflow in the snapshot.

eatyourgreens:
@mcbouslog is anyone on the Chicago team available to help here? I'm concerned that classifications may be affected on live projects.

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR adds a new "fetch-to-refresh" mechanism via `swr`. This new mechanism pairs with the recent `mst-persist`'s "save-store-in-sessionStorage" mechanism, and prevents "saved stores" from going stale and borking the "Reload Page" action.

- This _fully solves_ #2668
  - Fix: Reloading the page now properly triggers a workflow fetch
  - Improvement: if project owners have the Classifier open in one tab and the Project Builder in another, the edited workflow's changes will reflect quickly on the Classifier. (Tagging @goplayoutside3 as a minor FYI, since you brought up the issue of project owners) 
- ⚠️ NOTE: this PR does NOT solve issue #2670, as "wrong tutorials" still persist even with a page reload.

**New behaviour:** Classifier will periodically fetch workflow data from the API, to ensure the WF is up to date.
- the periodic "refreshing" fetch (as opposed to the initial data fetch) will trigger whenever the window/tab re-gains visibility and/or when the page is reloaded.
- if the fetched data has no changes, the Classifier's Tasks, etc remain the same. If there are differences, the Classifier's Tasks will update accordingly.

Functionality looks good!

### Testing

Tested on localhost with macOS + Chrome on [a random test project.](https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022/)

Testing steps 1: synced editing
- Open the Classify page for Workflow X in one tab/window
- Open the Project Builder to edit Workflow X in another
- Edit WF X in the project builder (e.g. change a question text)
- Switch the the Classify page tab.
- Observe that the changes are now reflected on the classifier

Testing steps 2: page reload
- Open the Classify page for Workflow X
- Reload the page.
- Observe (on the Chrome network tab) that data is being fetched from the Panoptes `/api/workflows/X` endpoint

All tests look good. Standard functionality (I can view the Project home page, I can select a WF to work on, and I can make+submit Classifications) look good.

Please see above posts for additional testing notes.

### Status

Approved 👍 I have some notes & caveats since this is a bit experimental, but this PR should be fine to go.

Key question:
- ❓ What do we plan to do when a user visits a `/classify` URL, e.g. `https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022/classify` ?
  - As noted in the additional testing notes, this causes the code to fetch `/api/workflows/undefined` (literally, because workflow ID is `undefined`)
  - It doesn't affect the UX, but is rather rude to the backend.

Key things to monitor:
- Keep an eye on big Transcription Projects. There _shouldn't be an instance_ of the "fetch-to-refresh" action causing the user's several-hours-worth-of-work to be reset, but best keep an ear out to be safe.

eatyourgreens:
I’m going to hold off on this until #2676 is merged and data caching is behind a flag.

shaunanoordin:
2676 just got merged! (Thanks Mark! 👍 ) I'm proceeding to test if issue #2690 is solved by disabling the sessionStorage.

shaunanoordin:
Confirmed, with the fix applied in 2676, Issue 2 (aka #2690 ) is no longer a problem! 👍 The lack of a saved store in sessionStorage does the trick.

eatyourgreens:
@shaunanoordin I've rewritten the `useEffect` hooks to differentiate between changes to the URL (`workflowID`) and changes to the workflow data.

eatyourgreens:
OK, let's try this out on staging. If it breaks live projects, I think it should be reverted.

Issue #2712, opened 2022-01-30
Classifier: remove the deprecated `inject` decorator


eatyourgreens:
## Package
lib-classifier

## Is your feature request related to a problem? Please describe.
The [`inject` decorator](https://github.com/mobxjs/mobx-react#provider-and-inject) is deprecated in `mobx-react`, and decorators are deprecated in MobX 6:
https://michel.codes/blogs/mobx6

## Describe the solution you'd like
The classifier has a couple of helpers that can be used to connect components to the store; a `useStores` hook and a `withStores` Higher Order Component. Both take an optional `storeMapper` function, which maps `classifierStore` to props.

```js
import React from 'react'
import { useStores } from '@helpers'
import MyViewComponent from './MyViewComponent'

function storeMapper(classifierStore) {
  const { workflows } = classifierStore
  return { workflows }
}

function MyConnectedComponent(props) {
  const { workflows } = useStores(storeMapper)
  return <MyViewComponent {...props} workflows={workflows} />
}
```

```js
import { withStores } from '@helpers'
import MyViewComponent from './MyViewComponent'

function storeMapper(classifierStore) {
  const {
    subjects: {
      active: subject
    }
  } = classifierStore

  return {
    subject
  }
}

export default withStores(MyViewComponent, storeMapper)
```

## Additional context
The NextJS apps probably still have old components that use `inject`. Those should be updated too.

eatyourgreens:
If we remove all the class decorators from our code, then I think we can also remove the babel plugin that parses them.

eatyourgreens:
All done now.

PR #2719, opened 2022-01-31
Remove @inject from subject viewers

eatyourgreens:
Remove `@inject` from `BarChartViewer`, `LightCurveViewer` and `SubjectGroupViewer`. Convert each from a class to a functional component with hooks. I've added a `useJSONData` hook to load JSON data for BCV and LCV subjects.

The enzyme tests were written to [test implementation details](https://kentcdodds.com/blog/testing-implementation-details), like internal class state or lifecycle methods. I've rewritten them to test the rendered subject viewer wrapper, and its props, either `onError` or `onReady`.

Test projects:
- LCV: https://localhost:8080/?project=nora-dot-eisner/planet-hunters-tess&env=production&demo=true
- Subject Groups: https://localhost:8080/?project=zookeeper/galaxy-zoo-weird-and-wonderful&env=production&demo=true

Package:
lib-classifier

Towards #2712.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SubjectGroupViewer/SubjectGroupViewerContainer.js
```diff
@@ -90,42 +90,51 @@ function storeMapper (stores) {
...
+  interactionMode = 'annotate',
+  isCurrentTaskValidForAnnotation,
+  loadingState = asyncStates.initialized,
+  onError = () => true,
+  onKeyDown = () => true,
+  onReady = () => true,
+  setOnPan = () => true,
+  setOnZoom = () => true,
+  subject = undefined
+}) {
```
eatyourgreens, line 113: The subject group viewer has a lot of props!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/BarChartViewer/BarChartViewerContainer.spec.js
```diff
@@ -35,106 +36,95 @@ const failSubject = Factory.build('subject', {
...
   describe('with an invalid subject', function () {
-    let cdmSpy
-    let onErrorSpy
     let nockScope
     let wrapper
     before(function () {
       sinon.stub(console, 'error')
-      cdmSpy = sinon.spy(BarChartViewerContainer.prototype, 'componentDidMount')
-      onErrorSpy = sinon.spy()
       nockScope = nock('http://localhost:8080')
```
eatyourgreens, line 55: In the project app tests, we mock the Panoptes API with nock. Maybe that would be better for the subject viewer tests too?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/BarChartViewer/BarChartViewerContainer.spec.js
```diff
@@ -35,106 +36,95 @@ const failSubject = Factory.build('subject', {
...
   describe('with a subject', function () {
     let nockScope
-    let cdmSpy
-    let cduSpy
     let wrapper
 
     before(function () {
-      cdmSpy = sinon.spy(BarChartViewerContainer.prototype, 'componentDidMount')
-      cduSpy = sinon.spy(BarChartViewerContainer.prototype, 'componentDidUpdate')
       nockScope = nock('http://localhost:8080')
```
eatyourgreens, line 128: Same comment about mocking `panoptes-staging.zooniverse.org` or `panoptes.zooniverse.org` here.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/BarChartViewer/BarChartViewerContainer.spec.js
```diff
@@ -143,60 +133,55 @@ describe('Component > BarChartViewerContainer', function () {
...
           subject={subject}
         />
       )
-
-      cdmSpy.returnValues[0].then(() => {
-        expect(onReadySpy).to.have.been.calledOnceWith({ target: wrapper.instance().viewer.current })
-      }).then(done, done)
     })
 
     it('should update component state when there is a new valid subject', function (done) {
```
eatyourgreens, line 164: This test isn't directly testing component state any more. It's testing the data props that are passed down to the subject viewer from the container.

eatyourgreens, line 164:
```suggestion
    it('should update subject data when there is a new valid subject', function (done) {
```

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
```diff
@@ -1,16 +1,17 @@
 import asyncStates from '@zooniverse/async-states'
-import * as d3 from 'd3'
+import { extent } from 'd3'
 import { zip } from 'lodash'
 import PropTypes from 'prop-types'
-import React, { Component } from 'react'
-import request from 'superagent'
```
shaunanoordin, line 6: Goodbye, `superagent`!

I was going to ask if we shouldn't also remove this from package.json, then I realised that DataImageViwer(?), SingleTextViewer(?), ScatterPlotViewer, and VariableStarViewer still uses the dependency, dangit. I didn't even realise those first two viewers existed.

eatyourgreens: `SIngleTextViewer` inspired me to write this PR, when I saw that it was still using outdated code from when we first started the monorepo.

packages/lib-classifier/src/helpers/useJSONData/README.md
```diff
@@ -0,0 +1,17 @@
+# useJSONData
+  
+A custom hook which loads JSON data for a Panoptes subject.
```
shaunanoordin, line 3: [minor] While this is true, the description is missing one important detail: `useJSONData` loads JSON data for a Subject, and assumes that Subject that has _exactly one JSON(-ish) URL data._

Specifically:
- A text file "technically" counts as a JSON URL (hence, _JSON-ish_ ) for this function, because the Panoptes CLI has (had?) issues differentiating JSON/TXT MIME types.
- If a Subject has multiple JSON(-ish) URLs, only the first is pulled by useJSONData()

This doesn't matter in current (2018 to 2022) LightCurveViewer (or BarChartViewer, I guess) scenarios, since project owners who use these Viewers know the precise Subject format required. However, I can imagine an edge case where useJSONData doesn't work as described, e.g. when we have a Subject that looks like...

```
subjectWithTextAndJSON.locations = {
  'text/plain': 'actual-boring-plain-text.txt',
  'application/json': 'json-object-for-data-viewer.json'
}
// In this case, the function attempts to load the plain text as JSON, fails at response.json(), and then skips the actual JSON object file.
```

I'm marking this as a minor thing right now (⭐ I think editing the README text to include the caveats above should suffice), because if a future theoretical SuperDataViewer needs to use this function to get a Subject with 10 JSON files and a dozen TXT files, we can adjust the code accordingly when it's less theoretical.

shaunanoordin:
Running manual tests now for SubjectGroupViewer and LightCurveViewer, brb

packages/lib-classifier/src/helpers/useJSONData/README.md
eatyourgreens: Really good point, but I think this can wait until we have real examples of subjects, rather than try to second guess what they might look like now.

shaunanoordin:
## Quick Review

I've encountered a problem with the LightCurveViewer, but I'm not sure what's the cause. I seem to be able to replicate it consistently.

**Issue 1:** when switching between Annotate and Move modes on the LCV, the zoom level resets (❗)

https://user-images.githubusercontent.com/13952701/153633279-217bf572-ae6e-4099-b7c9-a3bec45ba1f5.mov

Testing Setup:
- Test URL: https://local.zooniverse.org:8080/?env=production&project=nora-dot-eisner/planet-hunters-tess&demo=true (TESS Planet Hunters)
- Baseline: https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/classify/workflow/11235?demo=true (issue not detected on live)

Testing steps:
- Open the Classifier. You'll start in 'Annotate mode' (the pointing arrow icon)
- Click on the 'Move mode' button (four-direction arrow icon) on the Toolbar
- Use the mouse to zoom in (mouse wheel) and scroll to (mouse drag) look at an interesting position X on the Subject.
- Click on the 'Annotation mode' button on the Toolbar
  - Expected: you should STILL be zoomed in on the Interesting Position X on the Subject, so you can place an annotation mark precisely there
  - Actual: **the zoom level gets reset to 0**

Notes:
- There are small scenarios where this bug doesn't get triggered, but I haven't been able to fully document those cases yet. e.g. if you use keyboard (left/right, +/-) in Annotation mode to navigate, instead of switching to Move mode and using the mouse to navigate, then you might not notice the problem at all.
- I'll also need to double check on `master` to see if I'm missing something

Status: **major bug** ❗

Testing ongoing to pick up anything other issues before I circle around to see if I can figure out wtf is going on here. One thing to note is that the D3.js has its own "internal memory" for zoom/pan values (IIRC), so something is making the Viewer think it's refreshing and thus triggering the D3.js zoom reset.

eatyourgreens:
Isn't move mode controlled by a prop? Changing a component prop would certainly trigger a render, although I'm not sure why it would reset zoom.

eatyourgreens:
@shaunanoordin this code generates a new `dataPoints` array from `JSONdata` on each render. The LCV is set up to reset the chart if `dataPoints` changes. That might explain the bug.

shaunanoordin:
Ha! OK, looks like we've come to the same conclusion 👍 Should be an easy fix, we can just pass in the Subject ID to the LCV and use that to trigger `sameSubject?`

eatyourgreens:
@shaunanoordin already done! See the latest commit.

shaunanoordin:
I see your changes, Jim! 👍 I'll run tests now, brb!

shaunanoordin:
## Quick Review (Update)

LightCurveViewer is looking good! 👍 

- **[Issue 1](https://github.com/zooniverse/front-end-monorepo/pull/2719#issuecomment-1036416233)** has been solved
- I can view a valid LCV Subject
- I can switch to Move mode to navigate with a mouse (pan with mouse drag, zoom with mouse scroll)
- I can switch to Annotate mode to place an annotation mark using a mouse (mouse drag). 
  - In Annotate mode, I can move (mouse drag), edit (mouse drag edges of marks), and delete (click on 'X' button) existing marks
- I can navigate the LCV using keyboard keys (left/right keys to pan, +/- keys to zoom)
- I can submit Classifications
  - When I submit a Classification, the zoom levels reset when the next Subject loads.

(Cripes, I need to turn those 👆 into actual component tests)

shaunanoordin:
Yup, I can see that for `SubjectGroupViewerContainer`, the following state values:

```
const [panX, setPanX] = useState(0)
const [panY, setPanY] = useState(0)
const [zoom, setZoom] = useState(1)
```

...keep resetting to their default values for some reason. I haven't been able to trace why, unfortunately. This is now a Monday problem!

eatyourgreens:
I think this is what’s happening:
https://beta.reactjs.org/apis/usestate#updating-state-based-on-the-previous-state

eatyourgreens:
@shaunanoordin I've fixed those problems by using callbacks when state is changing continuously.

shaunanoordin:
> I think this is what’s happening:
> https://beta.reactjs.org/apis/usestate#updating-state-based-on-the-previous-state

Oh, that's new to me! I've never seen `setVal(val + 1)` with a function passed in, like `setVal(_val => _val + 1)`.

Thanks Jim! I'll wrap up testing later tonight

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR is primarily a refactor/code cleanup for the following Subject Viewer: `BarChartViewer`, `LightCurveViewer`, and `SubjectGroupViewer` (aka Grid Tool)

- The container components have been rewritten as functional component. 
- Notable code changes:
  - State-changing hooks that increment existing state values use this format: `setValue(_val => _val + 1)` and not this: `setValue(val + 1)`
  - LightCurveViewer now checks for Subject IDs to determine if it's looking at the same Subject.

Code changes look good, and testing for both LightCurveViewer and Grid Tool look great. I can't seem to find a live project that uses BarChartViewer, so I have to rely on Storybook for that.

### Testing

Tested on localhost with various projects, with macOS + Chrome.

Testing for **LightCurveViewer**
- Test URL: https://local.zooniverse.org:8080/?env=production&project=nora-dot-eisner/planet-hunters-tess&demo=true (TESS Planet Hunters)
- ✅ Full functionality checklist [noted in a previous comment](https://github.com/zooniverse/front-end-monorepo/pull/2719#issuecomment-1036445562)

Testing for **SubjectGroupViewer**
- Test URL: https://local.zooniverse.org:8080/?env=production&project=zookeeper/galaxy-zoo-weird-and-wonderful&demo=true (Galaxy Zoo Weird & Wonderful)
- ✅ Full functionality checklist [noted in a previous comment](https://github.com/zooniverse/front-end-monorepo/pull/2719#issuecomment-1036468783)

Testing for **BarChartViewer**
- Tested on Storybook
- ✅ Functionality checklist:
  - I'm able view the Subject displayed as a bar chart
  - I'm able to hover my mouse across different bars to see their numerical values.
  - I'm able to tab-navigate across different bars (and see their numerical values when I have focus)

### Status

LGTM 👍

PR #2755, opened 2022-02-02
Classifier: persist selected subject set ID

### LINKED ISSUE #2754: Failed to resolve reference [subject ID] to type...
## Package
- I think `lib-classifier`

## Describe the bug
1. from https://frontend.preview.zooniverse.org/projects/msalmon/hms-nhs-the-nautical-health-service?demo=true
2. select a workflow
3. select a subject set
4. refresh page
5. [bug] subject banner broken
6. refresh again (sometimes multiple times to get to...)
7. page show "Application error...", console states: `Error: [mobx-state-tree] Failed to resolve reference '44445025' to type '((SingleImageSubject | SingleTextSubject | SingleVideoSubject | SubjectResource) | SubjectGroup)' (from node: /subjects/active)`

## Screenshots
Application error more concerning, but for reference broken subject banner:
Screen Shot 2022-02-02 at 4 09 56 PM

eatyourgreens:
Remember `workflow.subjectSet`, which points to the current subject set, when refreshing a workflow from Panoptes.

- pass the `subjectSetID` prop to the store when hydrating a workflow.
- improved workflow data fetching: only fetch when `workflowID` is defined.
- add an undefined workflow option to the dev classifier, for testing and debugging.
- Don't clear the store if we accidentally click Classify but then return to the same workflow.

Package:
lib-classifier

Closes #2754.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/components/Classifier/Classifier.js
```diff
@@ -49,6 +49,8 @@ export default function Classifier({
   useEffect(function onWorkflowVersionChange() {
     const { workflows, subjects } = classifierStore
     if (workflowSnapshot) {
```
mcbouslog, line 51:
```suggestion
    if (workflowSnapshot && workflows?.active) {
```

If I don't edit as noted then I get `Uncaught TypeError: Cannot read properties of undefined (reading 'subjectSet')` when testing on https://local.zooniverse.org:8080/?project=nora-dot-eisner/planet-hunters-tess&env=production&demo=true

packages/lib-classifier/src/components/Classifier/Classifier.js
```diff
@@ -49,6 +49,8 @@ export default function Classifier({
...
+      const { subjectSet } = workflows.active
+      workflowSnapshot.subjectSet = subjectSet
```
mcbouslog, line 53:
```suggestion
      const subjectSet = workflows?.active?.subjectSet
      if (subjectSet) {
        workflowSnapshot.subjectSet = subjectSet
      }
```

Getting `Uncaught TypeError: Cannot read properties of undefined (reading 'subjectSet')` when testing on https://local.zooniverse.org:8080/?project=nora-dot-eisner/planet-hunters-tess&env=production&demo=true. 

The suggestion, or whatever similar change you see fit, will fix TESS while keeping the fix for HMS NHS.

mcbouslog (dismissed):
With small change this should be good to go

packages/lib-classifier/src/components/Classifier/Classifier.js
eatyourgreens: I've rewritten this so that the data flow for hydration is the same pattern used in NextJS:
component props => workflow data snapshot => classifier workflow store.

eatyourgreens:
@mcbouslog I've added a couple of other changes to this:
- the data fetching function should check for a workflow ID now, before fetching from Panoptes.
- the workflow picker in the dev classifier has a 'None' option, for testing and debugging.

shaunanoordin:
Just a note that I'm running a review in addition to Mark's. Testing on HMS NHS looks good so far (on `app-project`), at least in in respect to 2754, but I'm going to need to run tests on non-NHSHMS projects to complete the review, since these changes should be affecting all WF fetches.

(Note to self: WF ID is persisted, and now SSet ID is persisted - but will there be occasions where Subject ID needs to be persisted? Find examples)

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR fixes #2754 , which appears to be caused by the Subject Set ID not being persisted in localStorage along with the Workflow data

- So the problem is:
  - right now, when we load a "persistent storage"-enabled project like HMS NHS, the WF data is stored in localStorage. Cool.
  - However, if such a project _also_ allows Subject Set selection, the selected Subject Set isn't stored in persistent storage, which causes a mismatch.
- Solution is to store the Subject Set ID along with the WF data in persistent storage 
- **Additionally:** when we load URLs such as `https://local.zooniverse.org:3000/projects/darkeshard/test-project-2021/classify`, the Classifier no longer pings `/workflows/undefined` 👍 
  - i.e. the "fetch Workflow" action in the Classifier now only triggers when a WF ID is specified, excellent.

Code read looks fine, testing looks mostly good, given my testing.

### Testing

Tested on `app-project` on localhost, on a variety of projects, with macOS + Chrome

❗ NOTE: certain projects use **"persistent data"** stored in localStorage. (Stored per project, for the length of the session.) Be aware of different code behaviour when looking at any project page the first time vs all subsequent visits in that session.

Scenario 1: baseline
- Tested on a [basic test project](https://local.zooniverse.org:3000/projects/darkeshard/test-project-2021) with no persistent store, and multiple workflows, where you can't choose the Subject Set.
- Project home page + classify pages work normally. Classify page works as expected, I can make & submit classifications 👍 

Scenario 2: primary test, HMS NHS
- Tested on [HMS NHS](https://local.zooniverse.org:3000/projects/msalmon/hms-nhs-the-nautical-health-service/classify?env=production&demo=true) (has persistent storage)
- Issue 2754 can't be replicated 👌 
  - I can select a WF, select a Subject Set, and keep reloading the Classify page with no problems.

Scenario 3: additional test for Indexed Subject Sets + persisted stores
- Tested on the [RGBE Herbarium](https://local.zooniverse.org:3000/projects/emhaston/the-rbge-herbarium-exploring-gesneriaceae-the-african-violet-family/classify/workflow/19381/subject-set/100867/subject/71232372?env=production)
- Issue 2754 can't be replicated 👌 
  - I can select a WF, select a Subject Set, _select a Subject,_ and keep reloading the Classify page with no problems.

### Status

LGTM 👍

PR #2789, opened 2022-02-04
Add hydrated state to ClassifierContainer

### LINKED ISSUE #2610: Browser back button leads to already seen subject

mcbouslog:
Package:
- app-project
- lib-classifier

Closes #2610.

Describe your changes:
- in `app-project` ClassifyPage set `cachePanoptesData` per workflow `prioritzed` property
- in `lib-classifier` ClassifierContainer add a hydrated state to determine if and when store should be saved to browser session storage

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/ClassifierContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/ClassifierContainer.js
@@ -81,8 +81,17 @@ async function fetchWorkflow(workflowID) {
 }
 
 function useWorkflowSnapshot(workflowID) {
-  const { data } = useSWR(workflowID, fetchWorkflow)
-  return data ?? null
+  const [prevSnapshot, setPrevSnapshot] = useState(null)
+  const response = useSWR(workflowID, fetchWorkflow)
+  let data = null
+  if (workflowID) {
+    data = response.data
+  }
+  if (data && prevSnapshot?.id !== data.id) {
+    setPrevSnapshot(data)
+    return data
+  }
+  return prevSnapshot
 }
 
 export default function ClassifierContainer({
@@ -107,11 +116,12 @@ export default function ClassifierContainer({
     initialState: {}
   })
 
+  const [hydrated, setHydrated] = useState(false)
   const [loaded, setLoaded] = useState(false)
   const workflowSnapshot = useWorkflowSnapshot(workflowID)
 
-  async function onMount() {
-    if (cachePanoptesData) {
+  async function hydrateStore() {
+    if (cachePanoptesData && !hydrated) {
       try {
         const storageKey = `fem-classifier-${project.id}`
         await persist(storageKey, classifierStore, {
@@ -134,6 +144,7 @@ export default function ClassifierContainer({
           */
           subjects.setActiveSubject(subjects.active.id)
         }
+        setHydrated(true)
       } catch (error) {
         console.log('store snapshot error.')
         console.error(error)
@@ -149,8 +160,11 @@ export default function ClassifierContainer({
   }
 
   useEffect(() => {
-    onMount()
-  }, [])
+    console.log({ cachePanoptesData, hydrated, loaded })
+    if (!hydrated) {
+      hydrateStore()
+    }
+  }, [cachePanoptesData, hydrated])
 
   useEffect(function onAuthChange() {
     if (loaded) {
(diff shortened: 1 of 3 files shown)


eatyourgreens:
When I was testing this on Friday, I ran into store errors when the hydrated snapshot was out of sync with the workflow in the page URL. 

With the changes here, the classifier loads and renders first, then the store hydrates later. That makes it hard to follow the data flow. 

This fixes the bug, but I’d prefer a flow where the store is created and hydrated before the classifier mounts. I think hydrating after the store provider mounts will lead to subtle bugs, or store crashes.

eatyourgreens:
Where I think this might crash is if I re-enter the classifier after changing workflows. Imagine that I've classified on workflow A previously but the page URL has changed to workflow B eg. I went to the home page and chose a new workflow.
1. Classifier mounts and requests workflow B (from the workflow ID prop.)
2. Classifier renders with an empty store (THIS IS WHERE BEHAVIOUR HAS CHANGED.)
3. The empty store runs actions and reactions to load workflow B and its subjects.
4. Classifier hydrates with workflow A (from the previous stored session.)
5. API call for workflow B resolves and the workflows model is updated.

In this case, do we have to worry about the order that 3., 4. and 5. happen? Can hydration overwrite the selected workflow with stale state? Can the store hydrate with subjects for workflow A even though we're classifying on workflow B.

This is why I'd prefer a data flow where the store `Provider` isn't mounted and the classifier isn't rendered until after hydration is complete. In that case, the data flow is much more predictable:
1. Classifier mounts and requests workflow B (from the workflow ID prop.)
2. Classifier hydrates with workflow A (from the previous stored session.)
3. Classifier renders with a hydrated store (containing workflow A.)
4. Store reactions run to create a new classification from the hydrated data.
5. API call for workflow B resolves, and the store is updated. Subjects reset and we start a new queue.

packages/lib-classifier/src/components/Classifier/ClassifierContainer.js
```diff
@@ -149,8 +151,11 @@ export default function ClassifierContainer({
...
-    onMount()
-  }, [])
+    console.log({ cachePanoptesData, hydrated, loaded })
+    if (!hydrated) {
+      hydrateStore()
+    }
+  }, [cachePanoptesData, hydrated])
 
   useEffect(function onAuthChange() {
     if (loaded) {
```
eatyourgreens, line 161: `workflowID` is a required prop in the classifier so I think it should be checked here. 

Maybe keep track of the last workflow ID that was passed down from the URL. There’s a bug at the moment where the classifier renders with an undefined workflow if you go to `/classify`.

eatyourgreens:
I've updated this with extra debugging info, and tried to make sure the required `workflowID` prop is respected in the classifier. I'm still seeing it crash when I run it locally.

Davy Notebooks, showing an application error after changing workflows on the Classify page.

packages/lib-classifier/src/components/Classifier/ClassifierContainer.js
```diff
@@ -159,7 +173,7 @@ export default function ClassifierContainer({
   }, [loaded, authClient])
 
   try {
-    if (loaded) {
+    if (loaded && workflowSnapshot?.id) {
```
eatyourgreens, line 176: `workflowID` is a required prop in the classifier so I think it should be checked here. 

I've tried to keep track of the last workflow ID that was passed down from the URL, to fix a bug where the classifier renders with an undefined workflow ID if you go to `/classify`.

packages/lib-classifier/src/components/Classifier/ClassifierContainer.js
```diff
@@ -81,8 +81,17 @@ async function fetchWorkflow(workflowID) {
...
+  const response = useSWR(workflowID, fetchWorkflow)
+  let data = null
+  if (workflowID) {
+    data = response.data
+  }
+  if (data && prevSnapshot?.id !== data.id) {
+    setPrevSnapshot(data)
+    return data
+  }
+  return prevSnapshot
```
eatyourgreens, line 94: This beefs up `useWorkflowSnapshot` so that it returns the last workflow when `workflowID` is undefined.

eatyourgreens:
I’ve opened #2791 to set `cachePanoptesData` before the classifier mounts. Hopefully that will fix the problem.

PR #2790, opened 2022-02-05
Project: add allowedWorkflows to ClassifyPage

eatyourgreens:
Refactor the classify page to filter the available project workflows by workflow level, when workflow assignment is enabled. Don't allow a volunteer's chosen workflow (from the page URL) to be a higher level than their assigned workflow. Remove redundant state that was set by a `useEffect` hook.

I found this was useful advice on avoiding effects that set state:
https://beta.reactjs.org/apis/usestate#storing-information-from-previous-renders

Package:
app-project

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/src/screens/ClassifyPage/ClassifyPageContainer.js
+++ b/packages/app-project/src/screens/ClassifyPage/ClassifyPageContainer.js
@@ -12,32 +12,23 @@ function ClassifyPageContainer({
   ...props
 }) {
   const [subjectID, setSubjectID] = useState(subjectFromURL)
-  const [canLoadWorkflowFromUrl, setCanLoadWorkflowFromUrl] = useState(true)
-  let workflowFromUrl = workflows.find(workflow => workflow.id === workflowID)
+  const collectionsModal = createRef()
 
   let assignedWorkflow
+  let allowedWorkflows = workflows.slice()
+  let assignedWorkflowLevel = 1
   if (assignedWorkflowID) {
     assignedWorkflow = workflows.find(workflow => workflow.id === assignedWorkflowID)
+    assignedWorkflowLevel = parseInt(assignedWorkflow?.configuration.level, 10)
   }
-  const collectionsModal = createRef()
+  if (workflowAssignmentEnabled) {
+    allowedWorkflows = workflows.filter(workflow => workflow.configuration.level <= assignedWorkflowLevel)
+  }
+  const workflowFromUrl = allowedWorkflows.find(workflow => workflow.id === workflowID) ?? null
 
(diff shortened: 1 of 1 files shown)


mcbouslog (approved):
This LGTM. Tested locally with TESS, HMS NHS, and staging project [workflow-assignment-testing](https://local.zooniverse.org:3000/projects/srallen086/workflow-assignment-testing).

Should I test any other project with certain workflow settings or under other specific circumstances?

While testing I noticed a bug unrelated to these changes that I opened issue #2991 for - with a workflow assignment project a signed out user can access workflows above their assigned level (i.e. https://local.zooniverse.org:3000/projects/srallen086/workflow-assignment-testing/classify/workflow/2619).

eatyourgreens:
Good point. Allowed workflows default to all linked workflows, which isn’t right when you are logged out. 

I’ve refactored user auth in #2979, which could be a good point to check for user roles and permissions. 

I don’t think any other projects need testing. At the moment, you either have access to all project workflows, or a limited set filtered by individual project settings. 

Something I did notice here is that user project settings would be easier to work with if they recorded your level, rather than a specific workflow ID. I assume there’s some architectural decision behind using workflow IDs. If so, an ADR would be useful to document why we use workflow IDs instead of level numbers.

srallen:
:wave: Since I'm the only person that would conceivably know this information, I'm happy to share:

First, The original system did promotion outside of zooniverse and the id is a unique identifier that has meaning and can be correlated to an actual workflow. The levels (or better conceptualized as the order of access since level may imply difficulty) are essentially arbitrary and can change.

Second, the user is assigned the workflow id because it involves less data changes if the project changes their workflows.  Consider a scenario that actually happened with Gravity Spy:

Gravity Spy initially launched with five workflows with levels 1-5. Later on, they decided they wanted to add two new workflows, a new level 6 and one that replaces level 5. The preexisting level 5 now becomes level 7. All the users who were assigned to level 5 should now be on level 7. In order to make this change, we add a new workflow and set in its configuration that it is level 6, update the existing level 5 workflow to be level 7, then add a new workflow that is now level 5. No change has to happen with user data and the users who were on level 5 are now on level 7.

Now if we instead assigned by the arbitrary level label, then all the users on level 5 would have had their preferences updated to be assigned level 7. 

What ended up happening with Gravity Spy is that they decided to also update the glitch classes in 4-7, so users on level 5 got demoted to level 4, so we had data updating on the workflow and user preferences, but potentially, this could have just been an update just on three workflow configurations and the correct assignments and access would be maintained.

PR #2863, opened 2022-02-18
Classifier: Refactor and test useHydratedStore hook. Remove mst-persist

eatyourgreens:
- consolidate `useStore` and `useHydratedStore`, so that `useHydratedStore` replaces `useStore` to create the new store from an optional snapshot. 
- move the classifier hooks into `src/hooks`, aliased to `@hooks`.
- remove `mst-persist`.

This continues #2793, consolidating the hooks that are used to create the classifier's MST store. The Classifier now calls `useHydratedStore`, which replaces `useStore` to create a new store. 
```js
const classifierStore = useHydratedStore(storeOptions, cachePanoptesData, `fem-classifier-${project.id}`)
```
If the `cachePanoptesData` flag is set, the new store is created from a saved snapshot, using `RootStore.create(snapshot, storeOptions)`, rather than `applySnapshot(store, snapshot)`. This means that `lib-classifier` uses the same data flow, and logic, as we use in `app-project` to build the MST store in the browser.

I've also removed `mst-persist`, but I've copied its `onSnapshot` handler into the `useHydratedStore` hook.
https://github.com/agilgur5/mst-persist/blob/ae86acceb931ff82a8081fefb9a82f639d6da43a/src/index.ts#L48-L49

## Testing

Davy Notebooks has session storage enabled and can be used for testing single workflows in the project app:
https://local.zooniverse.org:3000/projects/humphrydavy/davy-notebooks-project/classify?env=production&demo=true

To test this out in the dev classifier, with all the indexing features enabled, change `cachePanoptesData` from `false` to `true` in `dev/App.js` and load in the RBGE project with a subject set and subject selected:
https://local.zooniverse.org:8080/?project=emhaston/the-rbge-herbarium-exploring-gesneriaceae-the-african-violet-family&workflow=19381&subjectSet=99948&subject=70656987&env=production

You can also try the RGBE project in a production build of the project app at:
https://local.zooniverse.org:3000/projects/emhaston/the-rbge-herbarium-exploring-gesneriaceae-the-african-violet-family/classify/workflow/19381/subject-set/99948/subject/70656991

Package:
lib-classifier

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


shaunanoordin:
Blargh! I keep running into this Unhandled Runtime Error when I try to load a Classify page for any Workflow X on a [normal-project](https://local.zooniverse.org:3000/projects/darkeshard/test-project-2021/classify/workflow/3579?demo=true) or a [project-with-cache](https://local.zooniverse.org:3000/projects/humphrydavy/davy-notebooks-project/classify/workflow/19684?env=production&demo=true)

image

`master` (b5e71f54c61b77e12fad9c7644735a0d12a0ded4) looks fine and non-Classify pages on this branch look fine, but honestly there's probably something borked in my settings or my setup. I'll try a full yarn panic and bootstrap next time, so check back with me on Thursday. Might need to check my node settings too, since I know my nvm keeps resetting to v14 due to some shenanigans with the aliasing.

shaunanoordin:
Confirming that I can now load the Classifier on this branch, so testing has resumed! 👍

shaunanoordin:
Wait, did I speak too soon? Tests were running fine for my random test projects, but now that I'm testing with Davy Notebooks I'm encountering errors. Here are my notes so far:

### Testing

Tested on localhost with macOS + Chrome 99. I'm using new Incognito Mode tabs on each test/test step to maintain Session Storage freshness.

Baseline tests: normal-projects
- Project: [random test project with multiple WFs to choose from](https://local.zooniverse.org:3000/projects/darkeshard/test-project-2021/?demo=true)
- As a volunteer, I can view the project home page
- I can view the Classify page of any WF X (and submit Classifications) by either 1. choosing from the home page, 2. choosing from the Classify link's WF menu, or 3. typing in the URL directly
- From the Classify page for WF X, I can switch to WF Y by either 1. going back to the home page and choosing WF Y, or 2. choosing WF Y from the Classify link's WF menu.
- As a dev, I can confirm that the Classifier store is NOT stored in Session Storage.

Feature tests: projects-with-cache
- Project: [Davy Notebooks (production)](https://local.zooniverse.org:3000/projects/humphrydavy/davy-notebooks-project?env=production&demo=true)
- ERROR: when attempting to choose any WF from the home page, e.g. https://local.zooniverse.org:3000/projects/humphrydavy/davy-notebooks-project/classify/workflow/20873?env=production&demo=true, I receive the error `TypeError: Cannot read properties of undefined (reading 'subjectViewer')`
  - I can confirm that this error doesn't exist on `master`, nor on zooniverse.org

Screenshot 2022-03-11 at 17 34 19

_Above: the error message. This is a fresh Incognito Mode tab, so there's no saved state in Session Storage._

shaunanoordin:
Update:

- The error triggers even on the `/classify` with no WF ID.
- The error triggers on HMS NHS too
- New URLs where I see the same issue:
  - https://local.zooniverse.org:3000/projects/humphrydavy/davy-notebooks-project/classify?env=production&demo=true
  - https://local.zooniverse.org:3000/projects/msalmon/hms-nhs-the-nautical-health-service/classify?env=production&demo=true

Good news, other projects seem fine, e.g. [Galaxy Zoo Weird and Wonderful](https://local.zooniverse.org:3000/projects/zookeeper/galaxy-zoo/classify/workflow/10582?env=production) and [TESS](https://local.zooniverse.org:3000/projects/nora-dot-eisner/planet-hunters-tess/classify/workflow/11235?env=production), so we can narrow down the debug to anything that uses the Panoptes cache (Classifier Store in Session Storage) option, and it's specific to this PR (again, `master` is fine)

eatyourgreens:
My mistake. `setLoaded(true)` was outside an `if` block when it should have been inside it. The full error was `classifierStore is undefined` which, yeah, would definitely be very bad for the classifier. It's fixed now.

I really need a good strategy for testing that top-level classifier container. React Testing Library cautions against [testing hooks in isolation](https://react-hooks-testing-library.com/), but I wonder if `useHydratedStore` should have tests anyway eg. to catch that `loaded` state was being set before the store had indeed loaded.

eatyourgreens:
I've written an ADR to go along with this:

https://github.com/zooniverse/front-end-monorepo/pull/2946

eatyourgreens:
This branch might fix a bug in when a subject is specified in the URL. Load this page in a browser, then refresh, and the store should crash. 

https://www.zooniverse.org/projects/emhaston/the-rbge-herbarium-exploring-gesneriaceae-the-african-violet-family/classify/workflow/19381/subject-set/99948/subject/70656987

This branch also seems to be broken since #2968 merged. I'll take a look at that.

packages/lib-classifier/src/hooks/useSessionStorage.js
```diff
@@ -0,0 +1,20 @@
...
+    storage.setItem(storageKey, data)
+  }
+  const snapshotDisposer = onSnapshot(store, _saveSnapshot)
+  return addDisposer(store, snapshotDisposer)
+}
+
+export default function useSessionStorage(enableStorage, store, storageKey) {
+  if (!!store && enableStorage) {
+    persist(storageKey, store, asyncSessionStorage)
+  }
```
eatyourgreens, line 18: Something to check: Does this need an additional check to make sure that the `onSnapshot` handler isn't added if it's already been added to the store? I'm not sure if this function can be run more than once.

packages/lib-classifier/src/hooks/useSessionStorage.js
```diff
@@ -0,0 +1,25 @@
...
+  function _saveSnapshot(snapshot) {
+    const data = JSON.stringify(snapshot)
+    storage.setItem(storageKey, data)
+  }
+  const snapshotDisposer = onSnapshot(store, _saveSnapshot)
+  return addDisposer(store, snapshotDisposer)
+}
+
+export default function useSessionStorage(enableStorage, store, storageKey) {
+  const [loaded, setLoaded] = useState(false)
```
eatyourgreens, line 16: This function is designed to stop running when this flips from `false` to `true`.

packages/lib-classifier/src/hooks/useSessionStorage.js
eatyourgreens: Turns out that `ClassifierContainer` renders more than once, running the hook multiple times.

eatyourgreens:
@shaunanoordin @goplayoutside3 I've added some tests to this, so that the `useHydratedStore` hook is tested when run both with and without a stored snapshot in session storage.

shaunanoordin (approved):
## PR Review (Update)

Following up on [the initial review](https://github.com/zooniverse/front-end-monorepo/pull/2863#pullrequestreview-903696490) and [first update](https://github.com/zooniverse/front-end-monorepo/pull/2863#issuecomment-1065334265)

I did new code read to catch up with the latest changes.

- We've got a latch in useSessionStorage, with the `loaded` state var ensuring the Classifier Store only gets an "onSnapshot" listener _one time._ (This was added after the last review, right?)
- There's a small existence check added in checkForProgress. Nice.
- I think the major Classifier-crashing bugs I was encountering before were fixed in 2968 (which fixed the undefined classifierStore at the onLoad hook), and this PR has been rebased on top of that?

I might be missing some other fixes stemming from other PRs, but long story short, code read looks good, and testing looks fine 👍 

### Testing

Tested on localhost with macOS + Chrome 99. I'm using new Incognito Mode tabs on each test/test step to maintain Session Storage freshness.

Baseline tests: normal-projects
- Project: [random test project with multiple WFs to choose from](https://local.zooniverse.org:3000/projects/darkeshard/test-project-2021/?demo=true)
- As a volunteer, I can view the project home page
- I can view the Classify page of any WF X (and submit Classifications) by either 1. choosing from the home page, 2. choosing from the Classify link's WF menu, or 3. typing in the URL directly
- From the Classify page for WF X, I can switch to WF Y by either 1. going back to the home page and choosing WF Y, or 2. choosing WF Y from the Classify link's WF menu.
- As a dev, I can confirm that the Classifier store is NOT stored in Session Storage.

Feature tests: projects-with-cache
- Project: [Davy Notebooks (production)](https://local.zooniverse.org:3000/projects/humphrydavy/davy-notebooks-project?env=production&demo=true)
- Same behaviour tests as above
- As a dev, I can confirm that the Classifier store IS stored in Session Storage.

### Status

I have no additional comments to make on the code, and my previous concerns with the crashing manual tests have been addressed. 👍 This PR is now LGTM, and I'll wait for @goplayoutside3 's notes on this PR

_(PS: Sorry for the Saturday review, my Friday evening work plans went sideways so I'm playing catch up.)_

goplayoutside3 (changes requested):
## PR Review

I can confirm the same UX testing as Shaun. I'm able to load the example projects.

I'm requesting changes because I think the bug fix I flagged should be in its own PR.

I also think it's confusing that `cachePanoptesData` and `enableStorage` are the same variable, but named differently across hooks and the Readme. I recommend picking one or even consistently naming it "enableSessionStorage" because that's the intended purpose.

Overall, it feels weird to have the check for `cachePanoptesData` in every ClassifierContainer hook instead of _one_ check in ClassifierContainer. For example, if `cachePanoptesData` is false, simply init the store from scratch, and if `cachePanoptesData` is true, then cascade all three custom hooks to handle the sessionStorage portion.

In constrast, this is the data flow in this PR:

`useHydratedStore` calls `useStore` under the hood. But `useStore` requires an `initialState`. `initialState` is returned from `useSavedSnapshot`. If `cachePanoptesData` is false, `initialState` is simply an empty object {}. Only then is `useStore` called to init `classifierStore`. Finally, `useSessionStorage` is called anyway regardless if we need to use sessionStorage.

I think that cascade of function calls is confusing for the goal of _optional_ sessionStorage. If it remains, I think it needs to be explicitly outlined in ClassifierContainer via readme or inline comments.

eatyourgreens:
Hooks can’t be conditional, I’m afraid, so wrapping them in a check for `cachePanoptesData` will throw a React error.

eatyourgreens:
@goplayoutside3 I've consolidated the hooks into a single hook. The hook code gets a bit more complicated, but that allows us to use conditional logic without breaking [the rules of hooks](https://reactjs.org/docs/hooks-rules.html).

goplayoutside3 (approved):
Code read looks good! I think the `hooks/Readme` is clearer and `useHydratedStore` is easier to understand. ADR 42 also accurately reflects this refactor.

I can confirm the UX tested earlier on this branch, and projects with prioritized workflow(s) have their classifier store in sessionStorage and regular project's classifier store do not.

PR #2882, opened 2022-02-22
Light Curve Viewer: stop Animation when new Subject loads

shaunanoordin:
## PR Overview

Follows #2719

Whenever a new Subject is loaded on the LightCurveViewer, an "animation" plays where the data points "swoosh in" from the top-left corner. This PR attempts to fix that.

https://user-images.githubusercontent.com/13952701/155182096-b01addd1-0b1a-4fd5-931e-90395bea4824.mov

_The issue in question. In the movie above, we cycle through 3 Subjects (by clicking the Done button, not picture). There is a significant animation playing on the 2nd and 3rd Subjects loading_

### Dev Notes

**Analysis Pt 1:**
- The "swooshing" animation is triggered by `LightCurveViewer.js`'s `updateDataPoint()`'s `.transition()`.
- This in turn is triggered by `drawChart()`, when it's called when shouldAnimate is true.
- This in turn turn is triggered by EITHER...
  - **Scenario 1** `<ReactResizeDetector>` detecting a resize (shouldAnimate is true in this case)
  - **Scenario 2** `componentDidUpdate()` detecting that the data has changed (a new Subject has come in).

SIDE BAR: curiously, for Scenario 2, the previous iteration of the componentDidUpdate() code (prior to PR 2719) actually looked like this...
```
// the old code =>
if (!sameSubject) { // if sameSubject is FALSE...
  this.drawChart(width, height, sameSubject)  // ...then animate if TRUE
}
```
...which meant it would logically NEVER animate when a new Subject came in.

Anyway, Scenario 2 is easy to fix, it shouldn't animate when a new Subject comes in.

Scenario 1 however is a different story...

**Analysis pt 2:**

- After PR 2719, it appears that the LCV viewer ALWAYS starts off as an un-rendered(??) before the data points start appearing.
- This means Scenario 1 (resize) is ALWAYS triggering when the data points come in/a new Subject loads.
- Even if we force `drawChart()`'s shouldAnimate to be false all the time no matter what (an easy fix), the LCV would still "blink" when a new Subject is loaded. There'll be no distracting animation, but there's a small period of time where it looks like there's nothing to show, just an empty chart.

### Status

WIP. I'm trying to figure out why there's always a resize when a Subject loads.

Currently marked as medium priority. It's affecting live TESS users (and very annoying), but not so much that it breaks the ability to classify.

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewer.js
@@ -84,9 +84,6 @@ class LightCurveViewer extends Component {
   }
 
   componentDidUpdate (prevProps) {
-    const subject = this.props.subjectID
-    const prevSubject = prevProps.subjectID
-    const sameSubject = (subject === prevSubject)
     const dataChanged = this.props.dataPoints !== prevProps.dataPoints
 
     const currentTaskKey = (this.props.currentTask && this.props.currentTask.taskKey) || ''
@@ -99,7 +96,7 @@ class LightCurveViewer extends Component {
       const container = this.svgContainer.current
       const height = container.offsetHeight || 0
       const width = container.offsetWidth || 0
-      this.drawChart(width, height, sameSubject)
+      this.drawChart(width, height, false)
     } else if (!sameTask) { // Triggers when changing between Workflow tasks.
       // TODO: load annotations when changing tasks.
       // If invalid task, blank out all annotaitons
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js
@@ -98,7 +98,7 @@ export function LightCurveViewerContainer({
     return { dataExtent, dataPoints }
   }, [JSONdata])
 
-  if (!subject.id || !dataPoints.length) {
+  if (!subject.id) {
     return null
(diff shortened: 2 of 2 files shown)


shaunanoordin:
## PR Update

OK, so here's an update on Scenario 1 - turns out that it was this line that was the problem:

```
if (!subject.id || !dataPoints.length) {
  return null
}
```

[Previously,](https://github.com/zooniverse/front-end-monorepo/blob/b038f15b6424f0c6b8ca95556f4048693326511a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/LightCurveViewer/LightCurveViewerContainer.js#L155-L157) the code looked like this: 

```
if (!subject.id) {
  return null
}
```

The LCV (container) will naturally cycle between having empty dataPoints and non-empty dataPoints when loading new Subject. If we DON'T render the `<LightCurveViewer>` (i.e. we instead render null) when we have empty data points, the LCV (component) will be "re-created" (rendered from scratch) at every Subject load, which in turn triggers the Resize event.

There is still a very slight animation (a slight expansion from centre) when a Subject loads, but this should be consistent with what had been happening previously. Give me a second to check out b038f15b6424f0c6b8ca95556f4048693326511a to confirm.

shaunanoordin:
## PR Update

OK, I've confirmed that the slight animation (expand from centre) is consistent with previous iterations of the LCV

https://user-images.githubusercontent.com/13952701/155343679-e73b6446-15c3-4bc2-89e7-4d7cd4a65662.mov

_Above: what happens when we cycle through 4 Subjects on commit b038f15. There is a slight animation where the new data points "expand" from the centre. This is much more noticeable on a large screen/window._

At the moment, I'm unable to remove animation completely, but at least I can make sure the major swooshing issue is solved.

### Status

Ready for review

eatyourgreens (approved):

PR #2888, opened 2022-02-23
Classifier: Filter prioritised subject queues

### LINKED ISSUE #2818: Engaging Crowds: duplicate pages being served for HMS NHS
## Package
lib-classifier

## Describe the bug
Volunteers are reporting, on Talk, that when they reach the end of a batch of 10 subjects, the last 3 subjects are used again.
https://www.zooniverse.org/projects/msalmon/hms-nhs-the-nautical-health-service/talk/4936/2331577?page=1&scrollToLastComment=true

## Expected behavior
New subjects added to the queue shouldn't include duplicates.

## Additional context
We request new subjects from Designator either 2 or 3 subjects before the end of the queue, so that you can classify while new subjects load in the background. When subjects are shown in order, this means the first 2 or 3 unclassified subjects from Designator could be duplicates of the last 2 or 3 subjects in your existing queue.

eatyourgreens:
Add `subjects.prioritized` for prioritised workflows. Check subject priority, for priority queues, and only add new subjects with priorities beyond the last subject priority in the queue.

EDIT: bypass that check for completed workflows, which fall back to showing subjects to a volunteer in a random order.

With this change, you should always see prioritised subjects in ascending priority order, starting at the last subject that you classified, without any repeats.

Test projects:
- Davy Notebooks (uses prioritised workflows): https://local.zooniverse.org:3000/projects/humphrydavy/davy-notebooks-project/classify?env=production&demo=true
- HMS NHS (uses grouped, prioritised workflows): https://local.zooniverse.org:3000/projects/msalmon/hms-nhs-the-nautical-health-service/classify?env=production&demo=true
- RBGE Herbarium (uses grouped, prioritised, indexed workflows): https://local.zooniverse.org:3000/projects/emhaston/the-rbge-herbarium-exploring-gesneriaceae-the-african-violet-family/classify/workflow/19381?env=production&demo=true

Package:
lib-classifier

Closes #2818.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
+++ b/packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
@@ -188,7 +188,8 @@ const SubjectStore = types
             const metadataPriority = subject.metadata['#priority'] ?? subject.metadata.priority
             // subject metadata in the API response are strings, not numbers.
             const priority = metadataPriority ? parseFloat(metadataPriority) : -1
-            notSeen = priority > self.last.priority
+            const lastPriority = self.last?.priority || -1
+            notSeen = priority > lastPriority
           }
           if (notSeen && !alreadyStored) {
             self.resources.put(subject)
--- a/packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
+++ b/packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
@@ -184,7 +184,9 @@ const SubjectStore = types
           const alreadyStored = self.resources.get(subject.id)
           // assume all subjects from Panoptes are unseen by default
           let notSeen = true
-          if (self.prioritized) {
+          // completed workflows fall back to random selection
+          const { user_has_finished_workflow } = subject
+          if (!user_has_finished_workflow && self.prioritized) {
             const metadataPriority = subject.metadata['#priority'] ?? subject.metadata.priority
             // subject metadata in the API response are strings, not numbers.
             const priority = metadataPriority ? parseFloat(metadataPriority) : -1
--- a/packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
+++ b/packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
@@ -80,6 +80,12 @@ const SubjectStore = types
         lastSubject = self.queue[self.queue.length - 1]
       }
       return lastSubject
+    },
+    /** are subjects sorted in priority order */
+    get prioritized() {
+      const { workflows } = getRoot(self)
+      const workflow = tryReference(() => workflows?.active)
+      return workflow?.prioritized
     }
   }))
 
@@ -176,7 +182,15 @@ const SubjectStore = types
       newSubjects.forEach(subject => {
         try {
           const alreadyStored = self.resources.get(subject.id)
-          if (!alreadyStored) {
+          // assume all subjects from Panoptes are unseen by default
+          let notSeen = true
+          if (self.prioritized) {
+            const metadataPriority = subject.metadata['#priority'] ?? subject.metadata.priority
+            // subject metadata in the API response are strings, not numbers.
+            const priority = metadataPriority ? parseFloat(metadataPriority) : -1
+            notSeen = priority > self.last.priority
+          }
+          if (notSeen && !alreadyStored) {
             self.resources.put(subject)
             self.queue.push(subject.id)
           }
--- a/packages/lib-classifier/src/store/SubjectStore/SubjectStore.spec.js
+++ b/packages/lib-classifier/src/store/SubjectStore/SubjectStore.spec.js
@@ -394,7 +394,7 @@ describe('Model > SubjectStore', function () {
     })
   })
 
-  describe('indexed workflows', function () {
+  describe('prioritised, grouped workflows', function () {
     let store
     let subjects = []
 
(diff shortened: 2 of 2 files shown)


packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
```diff
@@ -176,7 +182,15 @@ const SubjectStore = types
...
-          if (!alreadyStored) {
+          // assume all subjects from Panoptes are unseen by default
+          let notSeen = true
+          if (self.prioritized) {
+            const metadataPriority = subject.metadata['#priority'] ?? subject.metadata.priority
+            // subject metadata in the API response are strings, not numbers.
+            const priority = metadataPriority ? parseFloat(metadataPriority) : -1
+            notSeen = priority > self.last.priority
+          }
+          if (notSeen && !alreadyStored) {
```
shaunanoordin, line 193: The code read here looks good, generally. I have some questions:

- Is there any scenario when `self.last` is undefined? If yes, then `self.last.priority` is going to crash the classifier. However, if self.last should _never_ be undefined, then we should keep this as is if the classifier crashes, 
- If a Subject doesn't have a valid priority (e.g. maybe a project owner mistyped it as `subject.metadata.priority = 'page 1'`), it should never get added to the queue, correct?
  - Note: if the last item in a queue has an invalid priority, it's going to mess up the rest of the queue, as `priority > self.last.priorty` becomes `1234 > NaN` (EDIT: `1234 > undefined` is also possible) which will always evaluate to false
  - Not a problem for this PR per se, but it does mean everyone needs to be aware that it's the project owner's responsibility for setting up their Subjects properly, and there are no safety nets.

eatyourgreens: In either of those cases, `subject.priority` is broken anyway, I guess.
https://github.com/zooniverse/front-end-monorepo/blob/e1c866f28057fdc9bac09f00b8b28eeda0afe6d1/packages/lib-classifier/src/store/SubjectStore/Subject/Subject.js#L93-L99

eatyourgreens: Good question about `subjects.last`. It's undefined when the queue is empty, so when the classifier first loads.
https://github.com/zooniverse/front-end-monorepo/blob/e1c866f28057fdc9bac09f00b8b28eeda0afe6d1/packages/lib-classifier/src/store/SubjectStore/SubjectStore.js#L75-L83

eatyourgreens: I'm fixing that bug right now but here's something else I've found while testing. A workflow can have `workflow.prioritized` set, but use _randomised subject selection_ if that workflow is complete and all the subjects are retired.

Try it out here, with a prioritised, completed workflow from Davy Notebbooks:
https://localhost:8080/?project=humphrydavy/davy-notebooks-project&workflow=19689&env=production&demo=true

eatyourgreens: I've added a check for `user_has_finished_workflow` to cover that case.

shaunanoordin: Noted, let me re-run some tests for that case 👍

shaunanoordin:
Blargh sorry for the slow review, I keep running into WTFs.

Thought I should make notes in this PR what I'm looking at. Currently, looking at [a random Davy Notebooks WF](https://local.zooniverse.org:3000/projects/humphrydavy/davy-notebooks-project/classify/workflow/19684?env=production&demo=true) causes the classifier to crash. I had to check that the WF + Subject Set was configured properly ([yes](https://www.zooniverse.org/lab/9006/workflows/19684), I [think](https://www.zooniverse.org/lab/9006/subject-sets/98157)) 

Screenshot 2022-02-24 at 17 55 42

I need to see what's so malformed about Subject 68872262

EDIT: wait, let me re-bootstrap everything. It occurs to me that I'm testing lib-classifier changes on app-project

eatyourgreens:
If you need to debug Davy, the URL in the dev classifier is:
https://localhost:8080/?project=humphrydavy/davy-notebooks-project&workflow=19689&env=production&demo=true
(replace the workflow ID with the workflow that you're interested in.)

shaunanoordin:
Thank gods, it was just an error on my end. I re-fetched the branch and didn't do a bootstrap on the latest update. Manual tests are all passing now 👍

packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
```diff
@@ -176,7 +182,18 @@ const SubjectStore = types
       newSubjects.forEach(subject => {
         try {
           const alreadyStored = self.resources.get(subject.id)
-          if (!alreadyStored) {
+          // assume all subjects from Panoptes are unseen by default
+          let notSeen = true
+          // completed workflows fall back to random selection
+          const { user_has_finished_workflow } = subject
```
shaunanoordin, line 188: OK, quick sanity check. user_has_finished_workflow is taken straight from the API... so is there any scenario that a Subject says "user hasn't finished workflow?" but the user actually has JUST finished the WF (e.g. they're submitting the classification for the final Subject now!) so the API hasn't updated that boolean yet?

Probably? Maybe? But it's such a narrow edge case I'm having trouble coming up with a test for it.

packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
```diff
@@ -176,7 +182,18 @@ const SubjectStore = types
...
+          if (!user_has_finished_workflow && self.prioritized) {
+            const metadataPriority = subject.metadata['#priority'] ?? subject.metadata.priority
+            // subject metadata in the API response are strings, not numbers.
+            const priority = metadataPriority ? parseFloat(metadataPriority) : -1
+            const lastPriority = self.last?.priority || -1
```
shaunanoordin, line 193: Existential check for `.last`, solid 👍 I appreciate that we're setting fallback values for these priority valuables.

A small note is that parseFloat(metadataPriority) can still return `NaN`, but the logic still checks out OK for determining the correct value of `notSeen`

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR changes the behaviour of the Subject selection logic, but should ONLY affect specific projects (Davy Notebooks, HMS NHS, etc)

- Subject selection logic:
  - Previously, if the API returns a Subject that we've already stored in our MST stores, we didn't add it to the user's queue (since that'll duplicate the thing)
  - Now, we check if the Subject's already in the store, AND - _IF the workflow is prioritised (i.e. Subjects are arranged like pages in a book) AND _ - if the new Subjects have already been seen (e.g. from pages earlier in the book)
- Criteria to be affected by the change:
  - For projects with prioritised WFs, users should NOT see repeated Subjects (i.e. from "earlier in the book") after reaching the "end of the "book", so to speak.
  - Other projects should see no change in Subject selection behaviour

Code read looks good, and tests seem OK? I'll repeat one thing I've mentioned previously - that for Prioritised WFs to work. the project owners need to know they're responsible for correctly constructing their Subjects (i.e. don't put invalid values like `subject.metadata.priority = 'page 1'` or `= '-10'` or etc - but this is a known cost for these kind of projects.

### Testing

Tested on `app-project` on localhost, with macOS + Chrome.

- Baseline project: [random test project with no special attributes](https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022/classify/workflow/3581?demo=true)  👍 
  - I can select a workflow and start classifying Subjects
  - Subjects are served in a random order, repeating after I reach the end of the queue
- Davy Notebooks
  - I can select a WF and classify Subjects
  - I'm unable to determine if the Subjects are in "order" as I have trouble distinguishing one page from the other. But main functionality checks out, so OK
- HMS NHS
  - I can select a WF, select a Subject Set, and classify Subjects
  - I can view my "position in the book", e.g. "Subject 52 / 267"
  - Subjects are served randomly, but ordered.
  - e.g. I might start in Subject 52/267, and upon submitting, the next Subject is 53, then the next is Subject 54. 👍 
  - I have not tested this thoroughly by classifying all 267 Subjects to reach the end of the set, however. 🤷 
- RBGE Herbarium:
  - I can select a WF, select a Subject Set, select a Subject, and classify Subjects
  - I can view my "position in the book"
  - I can use Prev/Next buttons to navigate between Subjects.
  - Caveat: everything's already classified, so I always get the "Finished, what next?" popup, meaning testing is limited
  - **Update:** Wait, goddamit, RBGE Herbarium JUST disabled all their WFs since they're 100% complete, and now I can't test anything
  - If I'm at the final Subject in the Set, submitting will cause the "Already Finished" popup to appear. Cool, this means that I'm not getting "new" Subjects appended to the "end of the book"
    - ❓ However, if I close the popup, I can try to _submit again,_ which causes the Classifier to bork out with a hidden, console-only "derp you've got duplicate keys" error
      - EDIT: Sorry, I can't replicate this now due to the RBGE Herbarium disabling all WFs

### Status

In any case, testing looks as good as can be. The code read looks legit, the baseline projects look fine, and I can't think of anything that could bork this.

LGTM 👍

packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
eatyourgreens: I suppose if they were one subject away from the end, and this is the subject that completes the workflow, but I'm getting tired of adding guard clauses to the classifier, to be honest.

PR #2808, opened 2022-02-09
Classifier: useSubjectImage hook

eatyourgreens:
Refactor `SingleImageViewerContainer` and `MultiFrameViewerContainer` as functional components. Move the common image loading code into a `useSubjectImage` hook.

Package:
lib-classifier

# Review Checklist

## General

- [x] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [x] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2930, opened 2022-03-07
Add activeTool validation on frame change to MultiFrameViewerContainer

### LINKED ISSUE #2810: Corresponding with Quakers: Can't move on to next task
## Describe the bug
Bug report from Talk (https://www.zooniverse.org/projects/rachaelsking/corresponding-with-quakers/talk/4893/2336091): "Sometimes when I am done transcribing I try to exit the page by checking the "are all the lines gray" prompt and it will get stuck, I can click either 'no' or 'yes' and neither will work and bring me to the "next" prompt and I will be left unable to leave the page without knowing if my work is saved."

From volunteer:
> I notice this last time that it seems like one of the lines I have annotated is not closing properly, like I'll go through look the pages and one dialog will still be open ie. is still green, almost like there's an error in being able to close out a line that keeps me from properly exiting because it isn't finished.

Sounds like they can still interact with the Y/N buttons:
> When it happens again, it has been when I've annotated many pages of a single document, when I go to the "grey lines" prompt it doesn't matter if I press 'yes' or 'no' the 'next' button will not turn green.

Possibility of being more likely to happen when they're the first annotator?
> I have been able to finish several documents. I suspect has something to do with the length of time I'm on a document. I do snatches of annotation throughout a day. If I'm luck I'll get to do a full one in one sitting, but sometimes I'll come back and forth too it over several hours (I really like being able to get a full letter...it's very satisfying) I can't recall exactly, but I think it's more likely to happen if I'm the first one on a document. ie. I'm spending a lot of time trying to decipher the words, and less time just checking and confirming.

## To Reproduce
Haven't yet been able to reproduce, but sounds like it might be similar to what we were trying to fix with #2179 (invalid marks causing the Next button to be disabled)

Project 14978

## Expected behavior
My understanding is that the #2179 fix would delete extraneous marks, so perhaps something else is blocking the way to the next task?

## Device information
Firefox on Macbook

mcbouslog:
Package:
- lib-classifier

Closes #2810.

## Describe your changes:
- adds active tool validation on frame change to MultiFrameViewerContainer

<em>Helpful explanations that will make your reviewer happy:</em>
  - (confirm existing issue) https://www.zooniverse.org/projects/rachaelsking/corresponding-with-quakers/classify/workflow/20543?demo=true
  - (check fixed) https://local.zooniverse.org:8080/?project=rachaelsking/corresponding-with-quakers&workflow=20543&env=production&demo=true
  - (multi frame viewer without active tool to check negative side effects) https://local.zooniverse.org:8080/?project=1876&demo=true
  1. on a frame start to create transcription line, but don't click twice to finish the line (leaving line unfinished, one dot)
  2. go to different frame, make test transcription line, everything will look ok
  3. for Corresponding Quakers answer second task Y/N question
  4. on prod - can't click Next vs on local - can click Next
- Which storybook stories should be reviewed? - n/a
- Are there plans for follow up PR’s to further fix this bug or develop this feature? - maybe, this will hopefully fix most of related cases, but user can still be prevented from advancing with an incomplete line as further detailed in [this comment in the related issue ](https://github.com/zooniverse/front-end-monorepo/issues/2810#issuecomment-1062107792)

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.js
@@ -13,7 +13,6 @@ import locationValidator from '../../helpers/locationValidator'
 import SingleImageViewer from '../SingleImageViewer/SingleImageViewer'
 import SVGPanZoom from '../SVGComponents/SVGPanZoom'
 
-
 function storeMapper(store) {
   const {
     enableRotation,
@@ -25,7 +24,17 @@ function storeMapper(store) {
     setOnZoom
   } = store.subjectViewer
 
+  const { activeStepTasks } = store.workflowSteps
+
+  const [activeInteractionTask] = activeStepTasks.filter(
+    (task) => task.type === 'drawing' || task.type === 'transcription'
+  )
+  const {
+    activeTool
+  } = activeInteractionTask || {}
+
   return {
+    activeTool,
     enableRotation,
     frame,
     move,
@@ -48,7 +57,7 @@ class MultiFrameViewerContainer extends React.Component {
     this.setOnDrag = this.setOnDrag.bind(this)
 
     this.subjectImage = React.createRef()
-    
+
     this.state = {
       img: {}
     }
@@ -86,7 +95,15 @@ class MultiFrameViewerContainer extends React.Component {
   }
 
   onFrameChange(frame) {
-    const { setFrame } = this.props
+    const {
+      activeTool,
+      setFrame
+    } = this.props
+
+    if (activeTool?.marks?.size > 0) {
+      activeTool.validate()
+    }
+
     setFrame(frame)
   }
 
@@ -139,7 +156,7 @@ class MultiFrameViewerContainer extends React.Component {
       subject
     } = this.props
     const { img } = this.state
-    
+
     // If image hasn't been fully retrieved, use a placeholder
     const src = img?.src || 'https://static.zooniverse.org/www.zooniverse.org/assets/fe-project-subject-placeholder-800x600.png'  // Use this instead of https://www.zooniverse.org/assets/fe-project-subject-placeholder-800x600.png to save on network calls
     const naturalWidth = img?.naturalWidth || 800
@@ -204,6 +221,9 @@ class MultiFrameViewerContainer extends React.Component {
 }
 
 MultiFrameViewerContainer.propTypes = {
+  activeTool: PropTypes.shape({
+    validate: PropTypes.func
+  }),
   enableInteractionLayer: PropTypes.bool,
   enableRotation: PropTypes.func,
   frame: PropTypes.number,
@@ -219,6 +239,9 @@ MultiFrameViewerContainer.propTypes = {
 }
 
 MultiFrameViewerContainer.defaultProps = {
+  activeTool: {
+    validate: () => {}
+  },
   enableInteractionLayer: true,
   enableRotation: () => null,
   frame: 0,
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.spec.js
@@ -140,7 +140,30 @@ describe('Component > MultiFrameViewerContainer', function () {
         expect(image).to.have.lengthOf(1)
         expect(image.prop('xlinkHref')).to.equal('https://some.domain/image.jpg')
       })
-    })  
+    })
+
+    describe('with marks from an active drawing or transcription task tool', function () {
+      const validateSpy = sinon.spy()
+
+      // testMarks is a generic Map, does not represent actual marks from a task tool
+      const testMarks = new Map([
+        [1, 'one'],
+        [2, 'two'],
+        [3, 'three']
+      ])
+
+      // mock an active drawing or transcription task tool:
+      const activeTool = {
+        marks: testMarks,
+        validate: validateSpy
+      }
+
+      it('should validate active tool marks on frame change', function () {
+        wrapper.setProps({ activeTool })
+        wrapper.instance().onFrameChange(2)
+        expect(validateSpy).to.have.been.calledOnce()
+      })
+    })
   })
 
   describe('with an invalid subject', function () {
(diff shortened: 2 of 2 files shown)


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.js
```diff
@@ -86,7 +95,15 @@ class MultiFrameViewerContainer extends React.Component {
...
 
   onFrameChange(frame) {
-    const { setFrame } = this.props
+    const {
+      activeTool,
+      setFrame
+    } = this.props
+
+    if (activeTool?.marks?.size > 0) {
+      activeTool.validate()
```
eatyourgreens, line 104: What does `tool.validate()` do? It seems like there might be hidden side effects here.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.js
```diff
@@ -86,7 +95,15 @@ class MultiFrameViewerContainer extends React.Component {
   }
 
   onFrameChange(frame) {
-    const { setFrame } = this.props
+    const {
+      activeTool,
+      setFrame
+    } = this.props
+
+    if (activeTool?.marks?.size > 0) {
```
eatyourgreens, line 103:
```suggestion
    if (activeTool?.marks.size > 0) {
```

Marks are defined in the Tool model so you shouldn’t need to check for existence.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.spec.js
```diff
@@ -140,7 +140,30 @@ describe('Component > MultiFrameViewerContainer', function () {
...
+      const activeTool = {
+        marks: testMarks,
+        validate: validateSpy
+      }
+
+      it('should validate active tool marks on frame change', function () {
+        wrapper.setProps({ activeTool })
+        wrapper.instance().onFrameChange(2)
+        expect(validateSpy).to.have.been.calledOnce()
+      })
```
eatyourgreens, line 165: Not a blocking comment, but `wrapper.instance()` should be avoided in tests, since it's null for functional components.

This test doesn't tell us what validation does. If the expectation is that incomplete marks are deleted, then a better test would set up an incomplete translation line and expect it to be deleted on frame change. That would be a behaviour-driven test, whereas this tests how the code is implemented. Implementation can change, in refactors, without changing behaviour, which makes tests like this one brittle.

eatyourgreens (approved):
The bug fix looks good. I'm seeing stray marks deleted when I change frame on this branch. If it's quick, and easy, to test that invalid marks are deleted, then I'd ask for a test that checks that behavior, in order to avoid regressions. If not, then lets get this deployed ASAP but do take a look at #2808, which refactors this viewer as a functional component with hooks for subject data fetching.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.js
mcbouslog: I think per [new lines 32-34](https://github.com/zooniverse/front-end-monorepo/pull/2930/files#diff-ce0847accb83040e9a4dca672ec039b4ad28a27404b896384cd4aa8bcf861127R34) `activeTool` could be an empty object without a `marks` property, that was my reasoning.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.spec.js
mcbouslog: Yes I agree, I'll refactor this test better asap.

eatyourgreens: No hurry to do this right now, by the way.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.js
mcbouslog: it checks each mark's `isValid` view and deletes the mark if `isValid` returns `false` (i.e. mark is invalid). For the transcription tool that is 
```
get isValid() {
  return self.x1 !== self.x2 || self.y1 !== self.y2
},
```
I had the same feeling about hidden side effects, but I think the underlying issue of invalid marks needing to be deleted (i.e. `tool.validate()` on frame change) applies to all drawing tools, is ok. Alternatively I tested on the workflow with multiple frames and no drawing task/tool and looked ok.

eatyourgreens: Sounds good to me. Deleting marks is the kind of hidden side effect I was thinking of too, but I have no better suggestions how to make it obvious to a casual reader that this code does that.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.js
eatyourgreens: Oh good point. `activeTool` could be set to null if we aren't drawing, which would be caught by the first existence check but I don't know if that would break elsewhere. 🤷

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.spec.js
mcbouslog: @eatyourgreens I think this test is better now, not ideal (still using `wrapper.instance()`), but since #2808 refactors to functional component and relatedly refactors tests with mount then we can include removing `wrapper.instance()` in that PR? Sorry that #2808 will need to be refactored because of this PR, but I can help or will prioritize review.

eatyourgreens: Sounds good. I wouldn’t worry about breaking #2808. 

I think the equivalent fix, using hooks, would be something like
```js
useEffect(() => {
 activeTool?.validate()
}, [frame])
```

eatyourgreens: New test looks good. 👍🏻

PR #2947, opened 2022-03-11
app-project: Fix broken dark/light mode theme toggle

### LINKED ISSUE #2555: Dark theme cookies aren't applied instantly

eatyourgreens:
Remove server-side cookie handling and the `initialState.ui` page prop. Read cookies in the browser on app mount and apply to the store. This should remember your selected theme across projects, rather than using cached page props from the CDN.

To test this out, toggle the theme between light and dark, then check that it persists across page transitions, and also after refreshing the page.

Package:
app-project

Closes #2555.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging

### DIFF
--- a/packages/app-project/src/helpers/getDefaultPageProps/getDefaultPageProps.js
+++ b/packages/app-project/src/helpers/getDefaultPageProps/getDefaultPageProps.js
@@ -1,4 +1,3 @@
-import getCookie from '@helpers/getCookie'
 import getStaticPageProps from '@helpers/getStaticPageProps'
 
 const environment = process.env.APP_ENV
@@ -8,12 +7,7 @@ const HOSTS = {
   staging: 'https://frontend.preview.zooniverse.org'
 }
 
-export default async function getDefaultPageProps({ params, query, req }) {
-
-  // cookie is in the next.js context req object
-  const mode = getCookie(req, 'mode') || null
-  const dismissedAnnouncementBanner = getCookie(req, 'dismissedAnnouncementBanner') || null
-
+export default async function getDefaultPageProps({ params, query }) {
   const { props: staticProps } = await getStaticPageProps({ params, query })
   const { project, notFound, title, workflowID, workflows } = staticProps
   const host = HOSTS[environment] || 'https://localhost:3000'
@@ -21,11 +15,7 @@ export default async function getDefaultPageProps({ params, query, req }) {
     snapshot for store hydration in the browser
   */
   const initialState = {
-    project,
-    ui: {
-      dismissedAnnouncementBanner,
-      mode
-    }
+    project
   }
 
   const props = {
--- a/packages/app-project/pages/_app.js
+++ b/packages/app-project/pages/_app.js
@@ -39,6 +39,7 @@ function MyApp({ Component, pageProps }) {
 
   function onMount() {
     console.info(`Deployed commit is ${process.env.COMMIT_ID}`)
+    store.ui.readCookies()
     store.user.checkCurrent()
   }
   useEffect(onMount, [])
--- a/packages/app-project/src/helpers/getCookie/getCookie.js
+++ b/packages/app-project/src/helpers/getCookie/getCookie.js
@@ -1,8 +1,8 @@
 import cookie from 'cookie'
 
-export default function getCookie (req, name) {
-  if (req?.headers?.cookie) {
-    const parsedCookie = cookie.parse(req.headers.cookie)
+export default function getCookie (name) {
+  if (document?.cookie) {
+    const parsedCookie = cookie.parse(document.cookie)
     if (parsedCookie && parsedCookie[name]) {
       return parsedCookie[name]
     }
--- a/packages/app-project/src/helpers/getCookie/getCookie.spec.js
+++ b/packages/app-project/src/helpers/getCookie/getCookie.spec.js
@@ -3,6 +3,10 @@ import { expect } from 'chai'
 import getCookie from './getCookie'
 
 describe('Helper > getCookie', function () {
+  after(function () {
+    document.cookie = null
+  })
+
   it('should return an empty string if there is not a next.js context', function () {
     const value = getCookie()
     expect(value).to.be.a('string')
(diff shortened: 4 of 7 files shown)


goplayoutside3 (commented):
## PR Review

I'm able to toggle the dark/light mode. Refreshing the page remembers my theme mode preference, and that preference is consistent while navigating around FEM pages. ✔️ That preference is not saved across browser tabs.

Clarification question: The refactor of `readCookies()` in the UI Store includes `getCookie('dismissedAnnouncementBanner')`. Dismissing a banner is not remember after page refresh on this branch though. Is there a plan for implementing banner dismissal? Is it a relatively quick refactor that could be included here?

eatyourgreens:
I’ll double check but I think that’s the code that was already there. I’ve no plans to do any more work on it.

eatyourgreens:
I think this line does the same thing, on the live site?

https://github.com/zooniverse/front-end-monorepo/blob/f421b4dea4d0f19b9589709baf9138842d3b8fe6/packages/app-project/src/helpers/getDefaultPageProps/getDefaultPageProps.js#L15

goplayoutside3:
> I’ll double check but I think that’s the code that was already there. I’ve no plans to do any more work on it.

I'm referring to the `getCookies()` function that you added to the UI Store.

eatyourgreens:
Yeah, that’s the same function that ran on the server. The only change is that `request` is replaced with `document`. 

I’ve tried dismissing the announcement banner on my test project and the cookie does seem to be ignored. 
https://frontend.preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground/classify/workflow/3586/subject-set/4925?env=staging&demo=true


The tests for that cookie are passing, here and on the main branch, so I’m not sure where the problem is.

eatyourgreens:
I wonder if the announcement banner cookie has always been broken and we've never noticed?

eatyourgreens:
I can see that `dismissedProjectAnnouncement` session cookie in the browser, but it's definitely being ignored when I reload the page, or switch to a new tab.

https://frontend.preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground/classify/workflow/3586/subject-set/4974?env=staging
Screenshot of the browser storage console, showing that a 'dismissedProjectAnnouncementBanner' cookie is set for the session.

eatyourgreens:
Wait! Should the cookie name be `dismissedProjectAnnouncementBanner` or `dismissedAnnouncementBanner`? Is inconsistent naming the bug? If so, I think that's something that can be easily fixed in this PR.

eatyourgreens:
@goplayoutside3 I've added a failing test for the announcement banner cookie. How has the code gone for so long with this typo in it, not caught by any tests?

eatyourgreens:
@goplayoutside3 `ui.readCookies()` should be reading and parsing both cookies properly now. I think that the announcement banner cookie has been broken for a very long time.

eatyourgreens:
Roger pointed out, in https://github.com/zooniverse/front-end-monorepo/pull/1235, that the announcement banner doesn't need to be stored in global window state. The code might be easier to understand if the announcement banner component was responsible for both reading and writing the cookie, via `useEffect`. Hooks weren't around when the monorepo first started but we could use them now.

packages/app-project/src/helpers/getDefaultPageProps/getDefaultPageProps.js
```diff
@@ -8,24 +7,15 @@ const HOSTS = {
   staging: 'https://frontend.preview.zooniverse.org'
 }
 
-export default async function getDefaultPageProps({ params, query, req }) {
-
-  // cookie is in the next.js context req object
-  const mode = getCookie(req, 'mode') || null
-  const dismissedAnnouncementBanner = getCookie(req, 'dismissedAnnouncementBanner') || null
```
eatyourgreens, line 15: This has probably never worked, because the store property is called `dismissedProjectAnnouncementBanner`.

goplayoutside3:
I can confirm that banner dismissal and theme mode are remembered while navigating between FEM pages, but now refreshing the page reverts the theme to light mode and the banner reappears. Was that intentional with your latest changes?

eatyourgreens:
@goplayoutside3 this might be the problem. The toggle has created two mode cookies, on `.zooniverse.org` and `local.zooniverse.org`. I've got both 'light' and 'dark' set now, which causes unpredictable results.

I didn't touch the cookie writing code here, but it seems like it has bugs.

Screenshot of cookies for a local build of the project app, showing two mode cookies: one dark, one light.

eatyourgreens:
This gets weirder. If I inspect `document.cookie` in the console, it returns `'mode=light'` even though I can see a `dismissedProjectAnnouncementBanner` cookie for `local.zooniverse.org` under Storage.

eatyourgreens:
The banner cookie path is not set correctly. If I change it to `/projects/eatyourgreens/-project-testing-ground?env=staging`, I can read the cookie, but only on the home page. This must also have been broken ever since that code was first written.

eatyourgreens:
@goplayoutside3 I've done some debugging and the UI store sets the banner cookie path to `/${slug}` but the path to a project on our site is `/projects/${slug`, of course!

eatyourgreens:
Weird that that didn't fail any tests, either.

goplayoutside3 (approved):
Glad the dismissed banner could also be included here! Looks good. I'm able to keep my theme mode and banner dismissal preferences across FEM pages and during page refresh.

PR #2982, opened 2022-03-21
Drawing Tasks: track Active Tool changes

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Required for: #2883 

This PR changes some basic Drawing Task code, in response to @ErikOstlund's new Polygon drawing tool.

- New behaviour 1️⃣: on a Drawing Task, when a user switches the active Tool, any currently active Mark is _finished off_ and then _unselected._
- ~~❗ New behaviour 2️⃣ : DrawingTool's `endMoveMark()` no longer calls onFinish()~~ [(UPDATE: removed)](https://github.com/zooniverse/front-end-monorepo/pull/2982#issuecomment-1077875431)
  - _I'd appreciate some extra scrutiny on this. Please see Dev Notes._
- Minor: `onFinish()` now doesn't crash if 'event' isn't provided.

### Dev Notes

~~I have additional notes on behaviour 2️⃣: DrawingTool's `endMoveMark()` no longer calls onFinish()~~ [(UPDATE: removed)](https://github.com/zooniverse/front-end-monorepo/pull/2982#issuecomment-1077875431)

- ~~This change was requested due to two factors: 1. unlike other tools, the Polygon tool isn't finished _until the user explicitly says it's finished,_ and 2. any legit reason to _click on_ the Polygon _during the initial process creating it_ (e.g. drag to move) currently causes the Polygon to _accidentally finish._~~   
- ~~I'm actually not sure why `DrawingToolMarks.endMoveMark() ` originally has `onFinish()`, and that worries me.~~ UPDATE: it's to ensure that if you move a mark outside the bounds of a Subject's image, the mark gets deleted.
- ~~As far as I can tell, the behaviour it's trying to create is "hey, if a user moves a mark, make sure you tell the Interaction Layer that the creation process is complete."~~
- ~~However, most marks (Circle, Rectangle, Point) are always _finished during the initial creation process_ anyway.~~
- ~~I really need extra eyes on this to catch any scenarios that I missed. Transcription Tasks? Freehand Line tools?~~
- ~~@mcbouslog `DrawingToolMarks.endMoveMark()` was initially added in #1677 , do you remember why onFinish() (or onFinishWithRef() back then) is required in that part?~~ 

Bonus note 3️⃣ : I just realised that the new `InteractionLayer.onActiveToolChange()` logic could also be expanded to `onActiveMarkChange()`
- `onActiveMarkChange()` would just try to call `previousActiveMark.finish()` and delete it if it's invalid. (Assuming we somehow have access to previousActiveMark, erm...)
- This might make sense since, if you're in the middle of creating Mark A, then click on/select Mark B, then the Classifier should _try to finish_ Mark A and not just leave it hanging
- (onActiveMarkChange may or may not absorb the onDeleteMark effect.)
- This might be something for a separate PR though.

### Testing

Testing should be done on a Workflow with a Drawing Task that has multiple different Drawing Tools. (Note: the Polygon tool is still in dev (#2883), so we'll have to stick with Points, Rectangles, Circles, etc. Maybe Transcription Lines too?)

Test projects:
- (main testing) A Drawing WF with Circle, Rectangle, and Point tools: https://local.zooniverse.org:8080/?env=staging&project=1966&workflow=3596&demo=true
- (optional testing) A simple Transcription WF: https://local.zooniverse.org:8080/?env=staging&project=1967&workflow=3597&demo=true

Baseline tests: (aka all other existing drawing task behaviour shouldn't change)
- You should be able to draw **multiple** drawing marks, using **different** drawing tools, and submit.
- You should be able to **edit** and **delete** drawing marks 
- (optional) If you're testing Transcription Tasks, you should be able to place transcription lines (place two dots) and transcribe text (subtask popup), and then submit
  - ⚠️ KNOWN BUG/QUIRK: there's an existing bug with Transcription Tasks where you can place the first dot for Line A, then click on Line B to change the active mark, then Line A will just hang around being a _invalid mark._ (Expected: invalid marks should be deleted when it's not actively selected)
  - On master, if you click on Line A, it'll disappear (because endMoveMark() triggers onFinish() which realises the mark is invalid)
  - On this branch, the mark hangs around until explicitly deleted. I'm not too worried about this change in behaviour since that's a bigger bug that needs a different fix anyway, and might be addressed by 3️⃣ 

Feature tests:
- New behaviour 1️⃣:
  - You should be able to select the Rectangle tool and draw Rectangle A
  - Rectangle A should be the active mark (bold lines with visible delete button)
  - When you select Circle Tool, Rectangle A should NO longer be active.
  - You should be able to draw Circle B with no issues.

### Review Checklist

**General**

**Post-merging**

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner?
- [ ] Is the new feature working?

### Status

Ready for review!


eatyourgreens:
> I'm actually not sure why DrawingToolMarks.endMoveMark() originally has onFinish(), and that worries me.

I thought it was to check transcription lines, after editing/moving a line's endpoints, but according to the Git history, it was added for subtasks. Here's the PR that added it (it was originally called `onFinishWithRef` but refs were removed during a clean-up of the code.)
https://github.com/zooniverse/front-end-monorepo/pull/1677

shaunanoordin:
Tests need to be updated, brb

shaunanoordin:
OH! Oh, thank you, failing tests! I now see WHY you might want to call `onFinish()` after you finish moving a mark: **it's to check if the Mark is still within the visible Subject image's bounds**

- On `master`, if you create Rectangle on the image, then drag it OUTSIDE the bounds of the image for whatever reason, the Rectangle is deleted.
- This PR accidentally undoes that behaviour.
- I think this was introduced in #1357, and it's a legit use case.

Hmm, OK, let me reconsider this, because we now have two legit behaviours competing over the code: 1. if you move a mark out of bounds, it should be deleted, and 2. you should be able to move (or click simply interact with) a Polygon mark during its creation process without immediately calling onFinish().

shaunanoordin:
I'm turning this into a draft and un-requesting reviews while I consider behaviour 2️⃣ . If I can't figure out a solution by tomorrow, I might just re-open this PR with only behaviour 1️⃣. 🤔

shaunanoordin:
## PR Update

For the moment, I'm unable to figure out a good solution for the conflict between _"you should be able to move a Polygon even when you're creating it"_ and _"if you move a mark outside the bounds of the Subject image, delete the mark"_

Actions:
- For now, I've decided to **not** make any sweeping changes to the DrawingToolMarks, and have therefore reinstated the `onFinish()` at `endMoveMark()`

```
function endMoveMark(event) {
  if (event?.currentTarget && !isInBounds(event.currentTarget)) {
    deleteMark()
  } else {
    onFinish(event)  // I'm back!
  }
}
```

Consequences:
- @ErikOstlund 's Polygon tool will continue to have a bug where if you move a Polygon in the process of creating it, it finishes inadvertantly. This will be noted and solved at a later date.

### Status

This PR is now ready for review, with a slightly reduced scope.

shaunanoordin:
I'm closing this PR since there's no active development on it. This may be revisited again in the future.

PR #2940, opened 2022-03-10
Classifier: Add support for inactive (private) workflows

eatyourgreens:
- swap `project.links.workflows` for `project.links.active_workflows` when validating workflow IDs in `app-project`.
- add `usePanoptesAuth`, `usePanoptesUser` and  `useProjectRoles` hooks to the classifier.
- move `useStores` into the `hooks` directory.
- check project roles during workflow selection in the classifier. Allow owners, testers and collaborators to select any project workflow. Otherwise, restrict selection to a project's active workflows.

To test this PR, add an inactive workflow to a project, in the project builder, then load that workflow directly by ID in either the dev classifier or the project app. It should load, if you are logged in as a user that has permission to view all project workflows. If not, you should see the usual error in the browser console: `unable to load workflow X for project Y.`

Package:
app-project
lib-classifier

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2986, opened 2022-03-23
Classifier: add a loading state

### LINKED ISSUE #2954: Classifier: previously seen tutorial opens after logging in.
## Package
lib-classifier

## Describe the bug
On a project like PH TESS, where I've already seen the tutorial, the tutorial will still open if I log in after the `Classifier` store has been created.

## To Reproduce
1. While signed out of Panoptes, visit https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/classify/workflow/11235.
2. Dismiss the tutorial, which opens for all anonymous users.
3. Sign in.
4. The tutorial opens again, even if your project preferences mark it as seen.

## Expected behavior
The tutorial should only open if I haven't seen it before.

## Additional context
When you sign in, `upp.loadingState` should be reset, while your preferences load, so that `tutorial.hasNotBeenSeen` returns false here:
https://github.com/zooniverse/front-end-monorepo/blob/afddbb4683c5664eea33d6889257316d5f5b5f11/packages/lib-classifier/src/store/TutorialStore/Tutorial/Tutorial.js#L21-L23

If the UPP loading state is still set to 'success', from the initial user check when the classifier mounted, then `tutorial.hasNotBeenSeen` will be true while your preferences load, which will then trigger the tutorial to open.

eatyourgreens:
- Add a `loading` state to the classifier container. Defer rendering the `Classifier` until after the container is mounted and the store has initialised.
- Fix a small auth bug in the UPP store. Get an access token after checking the current user session, before fetching user preferences.

#2863 re-opened #2954 on staging, because the classifier store now loads synchronously (`loading` in `ClassifierContainer` is never false.) This PR should fix that by deferring the first render of the `Classifier` until after the classifier store is initialised (and the classifier UPP store has been reset.) To test it out, log out and back in on the Classify page of a project, so that you are signing in with an existing classifier store (and UPP loading state already set.)

Package:
lib-classifier

Closes #2954.

# Review Checklist

## General

## Components

## Apps

- [ ] Does it work on mobile?

## Publishing

## Post-merging


packages/lib-classifier/src/store/UserProjectPreferencesStore/UserProjectPreferencesStore.js
```diff
@@ -56,11 +56,10 @@ const UserProjectPreferencesStore = types
       const { authClient } = getRoot(self)
 
       try {
-        const authorization = yield getBearerToken(authClient)
```
eatyourgreens, line 59: I'm not sure if this can be reliably reproduced, but I was seeing this occasionally return an empty string, even when I was logged in.

eatyourgreens:
This is a quick fix for the loading state bug. #2940 adds classifier hooks to fetch user data, and that might also be a better long-term solution for fetching project preferences.

goplayoutside3 (approved):
## PR Review

Looks good! When I navigate to a project where I've already seen the tutorial such as TESS, sign-in, and the tutorial modal does not appear. When I visit a project I have not seen before with my user account, the tutorial modal appears.

Additional testing of a random project like i-fancy-cats:
- I can sign-in and sign-out
- Make a classification
- Load all FEM project pages

PR #2541, opened 2021-11-15
Use CDN domains for asset prefixes

eatyourgreens:
- use `www.zooniverse.org` and `frontend.preview.zooniverse.org` for asset prefixes in production and staging, respectively.
- move project images to `www.zooniverse.org/projects/assets/` so that we can route image URLs to the NextJS app.
- Add `robots.txt` to ban bots from `fe-project.zooniverse.org` and `fe-content-pages.zooniverse.org`

Package:
app-content-pages
app-project

Closes https://github.com/zooniverse/front-end-monorepo/issues/715
Closes https://github.com/zooniverse/front-end-monorepo/issues/2518

Depends on the nginx config being updated first, to route the asset prefix URLs to the NextJS apps.
https://github.com/zooniverse/static/pull/253

# Review Checklist

## General

- [ ] Are the tests passing locally and on Travis?
- [ ] Is the documentation up to date?

## Components
- [ ] Has a storybook story been created or updated?
- [ ] Is the component accessible? 
  - [ ] Can it be used with a screen reader? [BBC guide to testing with VoiceOver](https://bbc.github.io/accessibility-news-and-you/accessibility-and-testing-with-voiceover-os.html)
  - [ ] Can it be used from the keyboard? [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - [ ] Is it passing accessibility checks in the storybook?

## Apps

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `yarn panic && yarn bootstrap` or `docker-compose up --build` and app works as expected?

## Publishing

- [ ] Is the changelog updated?
- [ ] Are the dependencies updated for apps and libraries that are using the newly published library?

## Post-merging

- [ ] Did the app deploy to https://frontend.preview.zooniverse.org/projects/:project-name/:owner or https://frontend.preview.zooniverse.org/about?
- [ ] Is the new feature working or bug now fixed?
  - [ ] Is there a Talk or blog post written to announce the new feature(s)?
- [ ] Is the design working across browsers (Firefox, Chrome, Edge, Safari) and mobile?
  - [ ] Is this approved by our designer?
- [ ] Is this ready for production deployment?

PR #2994, opened 2022-03-29
Action: Staging Deploy

zwolf:
- Sets up a staging deploy via Github Action for FEM. A few idiosyncrasies force this to be a bit custom.
- Image builds will continue in both Hub and GHCR until the Jenkinsfile is removed.
- The contentful secrets need to be added to the repo (or org, if more widely used) by someone who knows what they are (I didn't find them in Passbolt).
- Not currently using shared workflows because of
    - build-args. Is this a requirement? Should the ci-cd version take build-args as a `with:` param?
    - Separation of `-staging` and `-production` repos in the container registry. Is that necessary? It’s deploying a tagged hash either way, this just necessitates rebuilding & pushing the image to the production repo at deploy. Also means this can’t use the shared ci-cd deploy_app workflow, but that might be true regardless since there’s two deployments in there and the “deploy necessary?” check wouldn’t account for that.

 

## How to Review

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

## New Feature

## Refactoring

## Maintenance


eatyourgreens:
There’s a GitHub Action that builds the NextJS apps during CI, which means the Contentful secrets must already be in this repo. 

We do need separate images for staging and production. I think because each has its own Sentry setup at build time, but I can’t remember off the top of my head. I do remember that we had a single image until we added Sentry logging.

eatyourgreens:
Here are the NextJS build secrets, for production at least. I _think_ staging uses the same Contentful tokens, just different asset prefixes. 

Actually, I should double check these asset prefixes, now that we serve assets via the CDN. This test workflow could be out-of-date. 


https://github.com/zooniverse/front-end-monorepo/blob/9530ecd8dd125a517f07101dd736689d3529c4c8/.github/workflows/ci-tests.yml#L58-L65

zwolf:
The contentful keys are repo secrets with the names I assumed in this PR they'd have, but they should also be in Passbolt. The asset prefixes were copied along with the rest of the env vars passed as build-args to docker in the Jenkinsfile.

eatyourgreens:
I remembered why we have a staging image and a production image: the static JS bundles that webpack builds during `next build` are different in staging and production.

.github/workflows/deploy_staging.yml
```diff
@@ -0,0 +1,67 @@
...
+        push: true
+        tags: |
+          ghcr.io/zooniverse/front-end-monorepo-staging:latest
+          ghcr.io/zooniverse/front-end-monorepo-staging:${{ github.sha }}
+        build-args: |
+          APP_ENV="staging"
+          COMMIT_ID=${{ github.sha }}
+          CONTENTFUL_ACCESS_TOKEN=${{ secrets.CONTENTFUL_ACCESS_TOKEN }}
+          CONTENTFUL_SPACE_ID=${{ secrets.CONTENTFUL_SPACE_ID }}
+          CONTENT_ASSET_PREFIX=https://fe-content-pages.preview.zooniverse.org
```
eatyourgreens, line 35:
```suggestion
          CONTENT_ASSET_PREFIX=https://frontend.preview.zooniverse.org/about
```

.github/workflows/deploy_staging.yml
```diff
@@ -0,0 +1,67 @@
...
+          SENTRY_CONTENT_DSN=https://1f0126a750244108be76957b989081e8@sentry.io/1492498
+          PROJECT_ASSET_PREFIX=https://fe-project.preview.zooniverse.org
```
eatyourgreens, line 37:
```suggestion
          PROJECT_ASSET_PREFIX=https://frontend.preview.zooniverse.org/projects
```

eatyourgreens (commented):
These are the correct values for the asset prefixes. I completely missed these in #2541.

eatyourgreens:
The Contentful API keys are in our Contentful account, under Settings > API Keys.

Screenshot of the Contentful Settings menu, including the API Keys option.

eatyourgreens (approved):
Looks good to me. I'm not sure if those asset prefixes are used to build the static JS bundles, but it can't hurt to use values that serve static assets via the CDN.

PR #3014, opened 2022-04-12
Classifier: test the modal tutorial

eatyourgreens:
Render the modal tutorial with and without `hasNotSeenTutorialBefore` set. Test that the close button closes the tutorial and updates user project preferences to mark the tutorial as seen.

## Package
lib-classifier

## General


eatyourgreens:
The `Modal` exported by `lib-react-components` is a bit of a tag soup of nested div's, without any accessible roles and names that we can use to find the popup itself. We should fix that, but here I've used the level 2 heading as a proxy for detecting whether the modal tutorial is open.

mcbouslog (approved):
This looks great, nice refactor to RTL!

PR #3017, opened 2022-04-12
Classifier: test text task with suggestions

eatyourgreens:
- upgrade `user-events` to 14 (13.5 is no longer maintained.)
- test the text task with suggestions when we click into it, both with and without an existing value. Suggestions should show when the input is empty.
- fix broken test setup. Enzyme must be set up after JSDOM.
- disable timeouts on slow-running tests, so we can run the `Choice` tests again.

Follow up to #3001, which fixed a bug in transcription projects but didn't add tests.

## Package
lib-classifier

## General

### DIFF
--- a/packages/lib-classifier/package.json
+++ b/packages/lib-classifier/package.json
@@ -10,8 +10,8 @@
     "dev": "webpack serve --config webpack.dev.js",
     "lint": "zoo-standard --fix | snazzy",
     "start": "PANOPTES_ENV=production npm run dev",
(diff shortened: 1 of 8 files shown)


eatyourgreens:
The tests here were passing until #3014 merged. 🤔 

`userEvent` uses `setTimeout`, so maybe there's a timeout somewhere in our tests that isn't cleared properly?
https://onestepcode.com/testing-library-user-event-with-fake-timers/

mcbouslog (approved):
This LGTM! Interesting discoveries here about RTL, userEvent, mocha, etc., thanks for refactoring

eatyourgreens:
The classifier tests hang without `—exit`, so that might be something to revisit.

Issue #2370, opened 2021-08-19
Storybook: Remove addon-knobs


eatyourgreens:
## Package
app-content-pages
app-project
lib-classifier
lib-react-components

Storybook 6.3 [deprecated `addon-knobs`](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-addon-knobs). We should replace knobs with [controls](https://storybook.js.org/docs/react/essentials/controls), in those stories that are using them, and remove `addon-knobs` from the package dependencies.

[Migration notes](https://github.com/storybookjs/storybook/blob/next/addons/controls/README.md#how-do-i-migrate-from-addon-knobs).

eatyourgreens:
Here's a quick search for [stories that use `withKnobs`](https://github.com/zooniverse/front-end-monorepo/search?q=withKnobs).

eatyourgreens:
All done, now that we’re on Storybook 7.

PR #3032, opened 2022-04-25
Migrate storybooks from `storiesOf` to `CSF`

eatyourgreens:
- [x] Migrate existing stories with `npx sb migrate storiesof-to-csf --glob="**/*.stories.js"`
- [x] Tidy up remaining stories by hand.

## Package
app-project
lib-classifier
lib-react-components

The best way to test this is probably to start the storybooks, with `yarn storybook`, and compare against the published storybooks at https://zooniverse.github.io/front-end-monorepo

### DIFF
--- a/packages/app-project/src/components/Announcements/components/FinishedAnnouncement/FinishedAnnouncement.stories.js
+++ b/packages/app-project/src/components/Announcements/components/FinishedAnnouncement/FinishedAnnouncement.stories.js
@@ -1,29 +1,37 @@
-import { withKnobs, text } from '@storybook/addon-knobs'
-import { storiesOf } from '@storybook/react'
-import zooTheme from '@zooniverse/grommet-theme'
-import { Grommet } from 'grommet'
-import { Provider } from 'mobx-react'
+import { withKnobs, text } from '@storybook/addon-knobs';
+import zooTheme from '@zooniverse/grommet-theme';
+import { Grommet } from 'grommet';
+import { Provider } from 'mobx-react';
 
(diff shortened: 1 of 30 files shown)


mcbouslog (approved):
This LGTM! This is great.

Noting we can remove the `storiesOf` and maybe some add-on imports in:
- lib-classifier
    - DataImageViewer
    - ScatterPlotViewer
    - VariableStarViewer
- lib-react-components
    - ZooHeader

I noticed sometimes in the Docs heading tab, the Show code button on stories doesn't work in app-project, but there's the Story tab at the bottom anyway so doesn't matter, might not even be related.

not related to this PR, but there's not a video in the SingleVideoViewer story, not sure if that's something @ErikOstlund can add though it's not a high priority.

eatyourgreens:
Here’s an example where the docs page says ‘No code available’ on the live storybook. 
https://zooniverse.github.io/front-end-monorepo/@zooniverse/fe-project/index.html?path=/docs/project-app-screens-project-home-recent-subjects--carousel-view

I’m not sure what’s happening there, but the docs page should replace the story-source addon long term.

eatyourgreens:
Knobs need to be replaced by controls, in those stories that use them. 
https://github.com/zooniverse/front-end-monorepo/issues/2370

Controls can be automatically generated for component props, so hopefully that won’t be a lot of work.

eatyourgreens:
If we want to take advantage of generated controls and docs in our stories, then we should move away from classes and write new components as functions with destructured default arguments.

PR #3034, opened 2022-04-26
Add subject ImageAndText and Text models, refactor SingleText model

mcbouslog:
## Package
- lib-classifier

Towards DigiLeap image and text viewer - [InVision link](https://projects.invisionapp.com/d/main/default/#/console/21704442/459400826/preview).
Splitting ImageAndTextSubject and ImageAndTextViewer into separate PRs for easier review.
These changes add a new subject model for a subject with a single image and single text location, which is the expected DigiLeap typical subject.
This type of subject will be viewed with an ImageAndTextViewer and will utilize the TextFromSubject task.
These changes also add a new shared text model for subjects with a text location, that requests the text location's content on subject load.
These changes refactor the SingleTextSubject to use the new shared text model.

## Describe your changes
- add ImageAndTextSubject
- refactor SingleTextSubject and ImageAndTextSubject with generic TextSubject

## How to Review
- What Zooniverse project should my reviewer use to review UX? no UX to review, this just adds/refactors subject models, but the following workflow has the related image and text subject (as well as other workflows in project with text subjects):
  - https://local.zooniverse.org:8080/?project=1952&workflow=3542&demo=true
- What user actions should my reviewer step through to review this PR? N/A
- Which storybook stories should be reviewed? N/A
- Are there plans for follow up PR’s to further fix this bug or develop this feature? yes - image and text viewer PR

# Checklist

## General

## General UX

## New Feature

## Refactoring


packages/lib-classifier/src/store/SubjectStore/shared/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './TextSubject.js'
```
eatyourgreens, line 1: Should these files move to `store/SubjectStore/TextSubject`, for consistency with the other subject models?

packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
```diff
@@ -6,6 +6,7 @@ import { getIndexedSubjects, subjectSelectionStrategy } from './helpers'
 import { filterByLabel, filters } from '../../components/Classifier/components/MetaTools/components/Metadata/components/MetadataModal'
 import ResourceStore from '@store/ResourceStore'
 import Subject from './Subject'
+import ImageAndTextSubject from './ImageAndTextSubject'
 import SingleImageSubject from './SingleImageSubject'
 import SingleTextSubject from './SingleTextSubject'
 import SingleVideoSubject from './SingleVideoSubject'
```
eatyourgreens, line 12: Not required for this PR, but with the number of subject models growing, we could move them into their own folder eg.
```js
import * as SubjectModels from './subjects'
```
or similar.

That would save some boilerplate code whenever a new subject type is added.

packages/lib-classifier/src/store/SubjectStore/ImageAndTextSubject/ImageAndTextSubject.spec.js
```diff
@@ -0,0 +1,154 @@
...
+  })
+
+  describe('with invalid subject locations', function () {
+    it('should throw an error', function () {
+      expect(() => ImageAndTextSubject.create(invalidLocationsSubjectSnapshot)).to.throw()
+    })
+  })
+
+  describe('with location request response that fails', function () {
+    let subject
```
eatyourgreens, line 78: `subject` is already defined in the upper scope. This line should be fine, since you're declaring it again with `let`, but using the same name in nested scopes can lead to bugs, in that the value of `subject` from the upper scope will be passed into this function. It can also make code harder to read.

packages/lib-classifier/src/store/SubjectStore/ImageAndTextSubject/ImageAndTextSubject.spec.js
```diff
@@ -0,0 +1,154 @@
...
+      store.workflows.setResources([workflowSnapshot])
+      store.workflows.setActive(workflowSnapshot.id)
+      store.subjects.setResources([subject])
+      store.subjects.setActive(subject.id)
+
+      await when(() => subject.contentLoadingState === asyncStates.error)
+    })
+
+    after(function () {
+      nock.cleanAll()
```
eatyourgreens, line 97: This won't break anything, but you only need to manually clean nock if you use `.persist()`. Without `persist`, your mocked responses are automatically cleaned after they've been requested.

packages/lib-classifier/src/store/SubjectStore/ImageAndTextSubject/ImageAndTextSubject.spec.js
```diff
@@ -0,0 +1,154 @@
...
+        .get('/failure.txt')
+        .reply(404)
+
+      subject = ImageAndTextSubject.create(failureSubjectSnapshot)
+
+      const store = mockStore()
+      store.workflows.setResources([workflowSnapshot])
+      store.workflows.setActive(workflowSnapshot.id)
+      store.subjects.setResources([subject])
+      store.subjects.setActive(subject.id)
```
eatyourgreens, line 91:
```suggestion
      const store = mockStore({
        subject: failureSubjectSnapshot,
        workflow: workflowSnapshot
      })
```

packages/lib-classifier/src/store/SubjectStore/ImageAndTextSubject/ImageAndTextSubject.spec.js
```diff
@@ -0,0 +1,154 @@
...
+    })
+  })
+
+  describe('Views > viewer', function () {
+    before(function () {
+      const store = mockStore()
+      store.workflows.setResources([workflowSnapshot])
+      store.workflows.setActive(workflowSnapshot.id)
+      store.subjects.setResources([subject])
+      store.subjects.setActive(subject.id)
```
eatyourgreens, line 147:
```suggestion
      const store = mockStore({
        subject: subjectSnapshot,
        workflow: workflowSnapshot
      })
```

packages/lib-classifier/src/store/SubjectStore/Subject/Subject.js
```diff
@@ -67,7 +66,16 @@ const Subject = types
...
+            if (counts.total === counts.images) {
+              viewer = subjectViewers.multiFrame
+            }
+            if (
+              counts.total === 2 &&
+              counts.images === 1 &&
+              counts.text === 1
+            ) {
+              viewer = subjectViewers.imageAndText
+            }
```
eatyourgreens, line 78: Can we reflect on the current model with `getType(self)` to determine which type of subject this is, and thus which viewer to use? That would save having the refinement logic in two places eg. this will break if the refinement is edited but the Subject model isn't updated.
https://mobx-state-tree.js.org/API/#gettype

packages/lib-classifier/src/store/SubjectStore/SubjectStore.spec.js
```diff
@@ -370,6 +371,34 @@ describe('Model > SubjectStore', function () {
...
+      subjects = await mockSubjectStore(imageAndTextSubjects)
+    })
+
+    it('should be valid subjects', function () {
+      const expectedSubject = ImageAndTextSubject.create(imageAndTextSubjects[1])
+      expect(subjects.resources.get(expectedSubject.id)).to.deep.equal(expectedSubject)
+    })
+
+    it('should be of the correct subject type', function () {
+      expect(getType(subjects.active).name).to.equal('ImageAndTextSubject')
```
eatyourgreens, line 398: This is the check that I think you can also use to determine which type of viewer to return for a subject.

packages/lib-classifier/src/store/SubjectStore/shared/TextSubject.js
```diff
@@ -0,0 +1,53 @@
...
+          error.status = response.status
+          throw error
+        }
+        self.content = yield response.text()
+        self.contentLoadingState = asyncStates.success
+      } catch (error) {
+        self.contentLoadingState = asyncStates.error
+        self.error = error
+      }
+    }
```
eatyourgreens, line 45: Would it be worth moving this into a hook in the text viewer? Similar to the hook that fetches images for the image viewer, or the hook that fetches JSON for the data viewers.

packages/lib-classifier/src/store/SubjectStore/shared/TextSubject.spec.js
```diff
@@ -0,0 +1,139 @@
...
+  })
+
+  describe('with an invalid subject location', function () {
+    it('should throw an error', function () {
+      expect(() => SingleTextSubject.create(invalidLocationsSubjectSnapshot)).to.throw()
+    })
+  })
+
+  describe('with location request response that fails', function () {
+    let subject
```
eatyourgreens, line 77: `subject` is already declared in the upper scope.

packages/lib-classifier/src/store/SubjectStore/shared/TextSubject.spec.js
```diff
@@ -0,0 +1,139 @@
...
+        .get('/failure.txt')
+        .reply(404)
+
+      subject = ImageAndTextSubject.create(failureSubjectSnapshot)
+
+      const store = mockStore()
+      store.workflows.setResources([workflowSnapshot])
+      store.workflows.setActive(workflowSnapshot.id)
+      store.subjects.setResources([subject])
+      store.subjects.setActive(subject.id)
```
eatyourgreens, line 90:
```suggestion
      const store = mockStore({
        subject: failureSubjectSnapshot,
        workflow: workflowSnapshot
      })
```

packages/lib-classifier/src/store/SubjectStore/shared/TextSubject.spec.js
```diff
@@ -0,0 +1,139 @@
...
+        .get('/success.txt')
+        .reply(200, 'This is test subject content')
+
+      subject = ImageAndTextSubject.create(imageAndTextSubjectSnapshot)
+
+      const store = mockStore()
+      store.workflows.setResources([workflowSnapshot])
+      store.workflows.setActive(workflowSnapshot.id)
+      store.subjects.setResources([subject])
+      store.subjects.setActive(subject.id)
```
eatyourgreens, line 122:
```suggestion
      const store = mockStore({
        subject: imageAndTextSubjectSnapshot,
        workflow: workflowSnapshot
      })
```

eatyourgreens (approved):
Nothing blocking, but I think the tests could be tidied up by removing redundant setup code that's already part of the `mockStore` helper.

Beware of overlapping variable declarations in nested function scopes. That can lead to bugs, since the value from the upper scope is passed into the nested scopes.

packages/lib-classifier/src/store/SubjectStore/Subject/Subject.js
eatyourgreens: Maybe
```js
if (getType(self) === ImageAndTextSubject) {
  viewer = subjectViewers.imageAndText
}
```

`ImageAndTextSubject` then guarantees that the conditions are met for number of images and number of text files.

packages/lib-classifier/src/store/SubjectStore/shared/index.js
mcbouslog: Yes I think that's a good idea, moved accordingly.

packages/lib-classifier/src/store/SubjectStore/SubjectStore.js
mcbouslog: I think that's a good idea, opened #3041 to track, it would definitely clean things up in a nice way and help going forward as more subjects are added

packages/lib-classifier/src/store/SubjectStore/Subject/Subject.js
mcbouslog: yes the duplicative refinement logic certainly not ideal, refactored accordingly and I think this is better

packages/lib-classifier/src/store/SubjectStore/shared/TextSubject.js
mcbouslog: Yes! I have this on my short list of DigiLeap follow-up todos/clean-up. I haven't thought through how/if the hook would be used with the `textFromSubject` task and not duplicate location requests, I don't think there'll be any issue and it's possible (to request location once for viewer and task), just haven't thought through it utilizing a hook yet. I think requesting the text location with a hook similar to the other locations is what's best though. I'll create an issue so it doesn't get lost in my notes.

PR #3040, opened 2022-04-28
Fix QuickTalk Authentication

shaunanoordin:
## PR Overview

Package: fix is _in_ `lib-classifier`, but fix is _for_ `app-project`

This PR fixes an issue where logged-in users couldn't post comments via QuickTalk on _app-project_

Issue:
- when the QuickTalk feature is tested on `lib-classifier`, logged-in users can post comments to the Talk (Subject discussion) thread quite fine.
- however, when the QuickTalk feature is tested on `app-project`, the "create a Talk post" action fails, with the error message saying the user isn't authorised to do so (error 401)

Analysis:
- Problem seems to stem from this line in [talk-client.js](https://github.com/zooniverse/panoptes-javascript-client/blob/7a627dff983b0e4fdf9dd0979330f239a4442e78/lib/talk-client.js#L11): `talkClient.headers = apiClient.headers`
- For some reason, in lib-classifier, `apiClient.headers.Authorization` exists if the user is logged in. In app-project, `apiClient.headers.Authorization`. Seems like there's a difference in the apiClient between lib-classifier and app-project.

Solution:
- In lack of a more elegant solution, this PR fixes the issue by simply force-updating `talkClient.header` with the correct header values, using the known bearerToken we've been tracking to create the Authorization field.
- [ ] 🔨 Room for Improvement 1: use find a way to use the `fixTalkClientAuth` logic by modifying `talkClient.beforeEachRequest` without making the whole damn thing crash. We can't keep jimmying `fixTalkClientAuth()` manually before each request.
- [ ] 🔨 Room for Improvement 2: this may require that the QuickTalkContainer itself be rewritten as a functional component to keep it in line with current FEM standards.

Notes:
- QuickTalk is a feature on the Classifier that only appears with the QuickTalk experimental_tool enabled.
- User has to be logged in to submit a comment, but can be logged out to view comments.
- Some UI improvements are incoming in a separate PR: 1. QuickTalk button (on lower right of screen) needs more visibility, 2. a message saying "You must be logged in to post a comment" needs to be added for better UX.

### How to Review

- Be sure to run `yarn bootstrap` before running `app-project`
- You require a project that 1. has QuickTalk enabled as an experimental tool, and 2. has the Talk boards setup.
- Example URL: https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022/classify/workflow/3581
- Login using a Zooniverse account.
- Click on the QuickTalk icon to open the QuickTalk menu.
- Post a comment.
- Expected: your comment should be posted to the Subject Discussion thread successfully. Your comment should be both immediately visible on the QuickTalk menu and visible on the Subject Discussion Talk page should you visit it.

### Status

Not sure if I should mark this ready for review or a work in progress (see 🔨 Room For Improvement) but comment are still welcome.

❗ UPDATE / WARNING: I just realised that on `frontend.preview.zooniverse.org`, **the talkClient will always ping production** (i.e. `talk.zooniverse.org`), **regardless of the ?env=staging** query. This make me think that the disconnect between the `apiClient` in lib-classifier and the `apiClient` in app-project is bigger than previously imagined. Hmm...


eatyourgreens:
`beforeEveryRequest` is a wrapper for `auth.checkBearerToken()` in the API client, so this should work. 
```js
import talkClient from ‘panoptes-client/lib/talk-client’

talkClient.beforeEveryRequest = () => authClient.checkBearerToken()
```

Alternatively, you should be able to replace `authClient.checkBearerToken()` with the `getBearerToken` helper. 

I’ve written a couple of classifier hooks to do this for you, but they would need QuickTalk to be written as a functional component.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
```diff
@@ -41,58 +41,60 @@ function storeMapper (stores) {
...
   Quick Fix: use authClient to check User resource within the QuickTalk component itself
   - see https://github.com/zooniverse/front-end-monorepo/discussions/2362
    */
   async checkUser () {
     const authClient = this.props?.authClient
     if (!authClient) return
-    
+
     const authorization = await getBearerToken(authClient)  // Check bearer token to ensure session hasn't timed out
     const user = await authClient.checkCurrent()
```
eatyourgreens, line 82: I think these two calls are out of order. You should check the current user session before you can get a bearer token. Otherwise, I’m pretty sure you’ll find that the user is logged in but `authorization` is an empty string, and you can’t post comments.

eatyourgreens (approved):
This works, but I think there’s a bug in your code which tries to get a token before you’ve logged a user in.

eatyourgreens:
I think this can close, now that we’ve rewritten `QuickTalk` to use the new API client.

PR #3046, opened 2022-05-04
Test the Classifier

eatyourgreens:
Add tests that render the classifier with a mock workflow and single image subject. Check the rendered classifier for the two task area tabs, the task answers and the SVG subject image. Test that locale changes are passed to the store, and that user roles are checked before loading a workflow.

## Package
lib-classifier

## General

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/Classifier.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.spec.js
@@ -46,7 +46,6 @@ describe('Components > Classifier', function () {
         <Classifier
           classifierStore={store}
           project={projectSnapshot}
-          subjectID={subject?.id}
           workflowID={workflowSnapshot?.id}
           workflowSnapshot={workflowSnapshot}
           workflowVersion={workflowSnapshot?.version}
@@ -107,7 +106,6 @@ describe('Components > Classifier', function () {
         <Classifier
           classifierStore={store}
           project={projectSnapshot}
-          subjectID={subject?.id}
           workflowID={workflowSnapshot?.id}
           workflowSnapshot={workflowSnapshot}
           workflowVersion={workflowSnapshot?.version}
@@ -172,7 +170,6 @@ describe('Components > Classifier', function () {
           classifierStore={store}
           locale='en'
           project={projectSnapshot}
-          subjectID={subject?.id}
           workflowID={workflowSnapshot?.id}
           workflowSnapshot={workflowSnapshot}
           workflowVersion={workflowSnapshot?.version}
@@ -187,7 +184,6 @@ describe('Components > Classifier', function () {
           classifierStore={store}
           locale='fr'
           project={projectSnapshot}
-          subjectID={subject?.id}
           workflowID={workflowSnapshot?.id}
           workflowSnapshot={workflowSnapshot}
           workflowVersion={workflowSnapshot?.version}
--- a/packages/lib-classifier/src/components/Classifier/Classifier.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.spec.js
@@ -15,7 +15,6 @@ import RootStore from '@store'
 import { ProjectFactory, SubjectFactory } from '@test/factories'
 import mockStore, { defaultAuthClient, defaultClient } from '@test/mockStore/mockStore'
 import branchingWorkflow from '@test/mockStore/branchingWorkflow'
-import stubPanoptesJs from '@test/stubPanoptesJs'
 import Classifier from './Classifier'
 
 describe('Components > Classifier', function () {
--- a/packages/lib-classifier/src/components/Classifier/Classifier.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.spec.js
@@ -12,14 +12,12 @@ import { Factory } from 'rosie'
 import sinon from 'sinon'
 
 import RootStore from '@store'
-import { ProjectFactory, SubjectFactory } from '@test/factories'
+import { ProjectFactory, SubjectFactory, TutorialFactory } from '@test/factories'
 import mockStore, { defaultAuthClient, defaultClient } from '@test/mockStore/mockStore'
 import branchingWorkflow from '@test/mockStore/branchingWorkflow'
 import Classifier from './Classifier'
 
 describe('Components > Classifier', function () {
-  let subjectImage, tabPanel, taskAnswers, taskTab, tutorialTab, workflow
-
   this.timeout(0)
 
   function withStore(store) {
@@ -35,6 +33,8 @@ describe('Components > Classifier', function () {
   }
 
   describe('while the subject is loading', function () {
+    let subjectImage, tabPanel, taskAnswers, taskTab, tutorialTab, workflow
+
     before(function () {
       const subject = SubjectFactory.build({ locations: [{ 'image/png': 'https://foo.bar/example.png' }] })
       const store = mockStore({ subject })
@@ -95,6 +95,8 @@ describe('Components > Classifier', function () {
   })
 
   describe('after the subject has loaded', function () {
+    let subjectImage, tabPanel, taskAnswers, taskTab, tutorialTab, workflow
+
     before(function () {
       const subject = SubjectFactory.build({ locations: [{ 'image/png': 'https://foo.bar/example.png' }] })
       const store = mockStore({ subject })
@@ -156,7 +158,7 @@ describe('Components > Classifier', function () {
   })
 
   describe('when the locale changes', function () {
-    let locale
+    let locale, subjectImage, tabPanel, taskAnswers, taskTab, tutorialTab, workflow
 
     before(function () {
       const subject = SubjectFactory.build({ locations: [{ 'image/png': 'https://foo.bar/example.png' }] })
@@ -238,6 +240,8 @@ describe('Components > Classifier', function () {
     const allowedRoles = ['owner', 'collaborator', 'tester']
     allowedRoles.forEach(function (role) {
       describe(role, function () {
+        let subjectImage, tabPanel, taskAnswers, taskTab, tutorialTab, workflow
+
         before(async function () {
           const roles = [role]
           const subjectSnapshot = SubjectFactory.build({ locations: [{ 'image/png': 'https://foo.bar/example.png' }] })
@@ -289,6 +293,7 @@ describe('Components > Classifier', function () {
             }
           )
           await when(() => store.subjects.loadingState === asyncStates.success)
+          workflow = store.workflows.active
           store.subjectViewer.onSubjectReady()
           taskTab = screen.getByRole('tab', { name: 'TaskArea.task'})
           tutorialTab = screen.getByRole('tab', { name: 'TaskArea.tutorial'})
@@ -338,6 +343,8 @@ describe('Components > Classifier', function () {
   })
 
   describe('without permission to view an inactive workflow', function () {
+    let subjectImage, tabPanel, taskAnswers, taskTab, tutorialTab, workflow
+
     before(async function () {
       const roles = []
       const subjectSnapshot = SubjectFactory.build({ locations: [{ 'image/png': 'https://foo.bar/example.png' }] })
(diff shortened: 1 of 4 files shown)


eatyourgreens:
Tests are failing because the subject viewer isn't rendering the SVG image. I'm not sure why. I think I've missed a step in the store setup which flags the subject as loaded and ready.

eatyourgreens:
Looks like this tests all of the Classifier code except for locale-switching and user roles.
https://coveralls.io/builds/48857309/source?filename=packages%2Flib-classifier%2Fsrc%2Fcomponents%2FClassifier%2FClassifier.js

eatyourgreens:
Updated to add some coverage for locale switching.
https://coveralls.io/builds/48868898/source?filename=packages%2Flib-classifier%2Fsrc%2Fcomponents%2FClassifier%2FClassifier.js

eatyourgreens:
I've added a couple of tests for inactive workflows, with and without the 'owner' role, to check user permissions. Those involve a lot of mocking to set up the Panoptes API, but seem to be working now. I've made some changes to `mockStore` to make it easier to set up mock clients both for Panoptes auth and the Panoptes API itself.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
```diff
@@ -0,0 +1,431 @@
...
+      })
+    })
+  })
+
+  describe('with permission to view an inactive workflow', function () {
+    const allowedRoles = ['owner', 'collaborator', 'tester']
+    allowedRoles.forEach(function (role) {
+      describe(role, function () {
+        before(async function () {
+          const roles = ['owner']
```
eatyourgreens, line 247:
```suggestion
          const roles = [role]
```

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
```diff
@@ -0,0 +1,431 @@
...
+
+      it('should be disabled', function () {
+        taskAnswers.forEach(radioButton => {
+          expect(radioButton.disabled).to.be.true()
+        })
+      })
+    })
+  })
+
+  describe('after the subject has loaded', function () {
```
goplayoutside3, line 99: Could the "after the subject has loaded" test be in the same test suite as "while the subject is loading"? I think using `rerender` similar to the locale unit test would be closer to how the Classifier behaves in real life. The Classifier component can mount without a subject, and then React re-renders it when `subjectID` prop changes - is that correct?

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
```diff
@@ -0,0 +1,431 @@
...
+          const client = { ...defaultClient, panoptes }
+          const store = RootStore.create({}, { authClient, client })
+          render(
+            <Classifier
+              classifierStore={store}
+              project={projectSnapshot}
+              workflowID={workflowSnapshot.id}
+              workflowSnapshot={workflowSnapshot}
+              workflowVersion={workflowSnapshot.version}
+            />,
```
goplayoutside3, line 291: In this test suite and the following one, I noticed that Classifier is rendered without a `subjectID`. I think it would be valuable to illustrate somewhere in lib-classifier what we expect Classifier to do when a `subjectID` exists and when it doesn't. That would probably mean testing the side effects of `onURLChange()`. Is that outside the scope here and instead tested in WorkflowStore?

goplayoutside3 (commented):
I think in general these tests look good! A couple things I think we should think about:

1. Each describe block in Classifier's spec file checks for the task tabs, subject image, and task answers. However, these components all belong to `<DefaultLayout />`. The DefaultLayout component is always going to be rendered in each test suite setup here. In other words, I wasn't able to make the unit tests for tabs, image, and answers fail by changing code in the Classifier that wouldn't cause the test environment to error. I think unit tests for disabling the task answers should be in TaskArea components, and instead we consider what scenarios can be tested related to Classifier's try/catch block in its render method.
 
2. A unit test could be added for `showTutorial` prop, and check for ModalTutorial.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
eatyourgreens: I think that’s right. I think Engaging Crowds projects mount the classifier before you’ve selected a subject to work on, then pass in a subject ID after it has mounted. 

You’re absolutely right about `rerender`. I’m sure there are closed issues because the classifier didn’t originally react properly to prop changes, and we don’t have tests for those.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
eatyourgreens: That’s a good point. I don’t think we’ve got any live projects that use `subjectID`. When it’s present, subject selection should show you exactly that subject ie. you can select which subject to work on. 

In the dev classifier, it’s useful to debug specific subjects, when you know the ID.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
eatyourgreens: Here's an example of a bug that was caused by a props change.
https://github.com/zooniverse/front-end-monorepo/issues/2140

eatyourgreens: The classifier itself won’t rerender when the subject finishes loading, but the store will update, which then triggers the active task to become enabled. 

The `subjectID` prop is passed down from the page URL, in projects that allow you to select or link to specific subjects.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
```diff
@@ -0,0 +1,431 @@
...
+          subjectID={subject?.id}
+          workflowID={workflowSnapshot?.id}
+          workflowSnapshot={workflowSnapshot}
+          workflowVersion={workflowSnapshot?.version}
+        />,
+        {
+          wrapper: withStore(store)
+        }
+      )
+      store.subjectViewer.onSubjectReady()
```
eatyourgreens, line 120: This line is the difference from the previous test. It mocks the image load event. Maybe there’s a better way to do that? I’m not sure.

eatyourgreens: I’m pretty sure this can be merged into the previous test suite, so we update the store then test the output, rather than running the same setup all over again.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
```diff
@@ -0,0 +1,430 @@
...
+      const store = mockStore({ subject })
+      const project = store.projects.active
+      const projectSnapshot = { ...getSnapshot(project) }
+      workflow = store.workflows.active
+      const workflowSnapshot = { ...getSnapshot(workflow) }
+      render(
+        <Classifier
+          classifierStore={store}
+          project={projectSnapshot}
+          subjectID={subject?.id}
```
eatyourgreens, line 110: Same comment about the subject ID here.

eatyourgreens (commented):
There's a couple of subject IDs that can be removed, since we aren't testing subject set selection or subject selection.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
eatyourgreens: I've updated all the tests to remove the subject IDs, as they aren't used by the test workflow.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
eatyourgreens: I spent some time working on this today but kept running into RTL’s `afterEach` cleanup. The screen is empty if I go back to examine it again, unless I run a new ‘render’, which is what my existing tests do.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
goplayoutside3: Really good to know. Non-blocking suggestion that we add a note in this test suite about why `onSubjectReady()` and `rerender` can't be used in this scenario.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
```diff
@@ -0,0 +1,507 @@
...
+        taskAnswers.forEach(radioButton => {
+          expect(radioButton.name).to.equal('T0')
+        })
+      })
+
+      it('should be disabled', function () {
+        taskAnswers.forEach(radioButton => {
+          expect(radioButton.disabled).to.be.true()
+        })
+      })
```
goplayoutside3, line 93: The added tutorial tests look good!

I want to go back to my previous point that many of Classifer's unit tests here check for the task tabs, subject image, and task answers, but those components belong to TaskArea. For example, whether the TaskComponent is disabled or not is controlled by Task and tested in its spec file:

https://github.com/zooniverse/front-end-monorepo/blob/880c355016987b40a74ac00fde336186a6c074a1/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.spec.js#L104-L107

Classifier.js has a lot of `useEffect()` functions that change parts of the classifier store. I like your previous comment that this spec file can be expanded to test various types of subjects and workflows, and then we could observe the changes to the classifier store in each scenario. Disabled TaskArea components are already tested in their spec files, so I'm leaning toward Classifier unit tests focused only on the changes we expect in the classifier store - meaning "before and after a subject loads" blocks could be removed altogether. What do you think?

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
eatyourgreens: This should use something like `fireEvent.load(subjectImage)`, rather than directly manipulating the store. We want to check that you can classify after the subject has loaded, so that we can catch bugs like the one that was fixed by #3250. Also, that bug doesn’t have an issue. That might be useful for tracking regressions.

eatyourgreens: `rerender` is used to simulate prop changes. Props don’t change when a subject loads so `rerender` isn’t really relevant here. Mentioning it might be confusing. 

I need to rewrite these tests to fire events after the classifier mounts.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
eatyourgreens: The `TaskArea` tests don’t test the task alongside a subject viewer. That’s why we were able to deploy production code with a bug that passed the task and subject viewer tests, but showed a disabled task after the subject had loaded (#3250.) Tests here, with the appropriate subject and task, would have caught that bug and failed. 

The passing `TaskArea` tests are an example of the false positives discussed by Kent Dodds in his article about [testing implementation details](https://kentcdodds.com/blog/testing-implementation-details#false-positives).

eatyourgreens: Re. The `useEffect` hooks, they update the store but the way to test store state, and still catch bugs, is to test the rendered HTML for a given store state. 

For example, if I click a link to navigate to a new workflow, the workflow ID change will fire an effect that updates the store, but I’d test that the subject image is the expected subject for that workflow, and the tasks shown are the expected tasks for that workflow. If my tests only check `store.workflows.active`, then I don’t have that confidence that the subject viewer and task area have updated to show the new workflow.

goplayoutside3: That makes sense 👍

goplayoutside3 (approved):

eatyourgreens:
Still to do: figure out a better way to detect the subject loading. SVG images might not fire a load event in all browsers.

packages/lib-classifier/src/components/Classifier/Classifier.spec.js
```diff
@@ -0,0 +1,507 @@
...
+    before(async function () {
+      const subject = SubjectFactory.build({ locations: [{ 'image/png': 'https://foo.bar/example.png' }] })
+      const steps = [
+        { content: "Hello" },
+        { content: "Step 2" }
+      ]
+      const workflowTutorial = TutorialFactory.build({ steps })
+      const store = mockStore({ subject })
+      store.tutorials.setResources([workflowTutorial])
+      store.tutorials.setActive(workflowTutorial.id)
```
goplayoutside3, line 446: @eatyourgreens as I'm working through lib-classifier's specs, and converting them to run with Vitest, I noticed there's widespread use of `setResources` and `setActive` related to mobx and mobx-state-tree. However, I can't find any documentation about where those two functions come from. Are they defined by `mobx`? `mobx-state-tree`? Any insight into why lib-classifier relies on them would be really helpful, particularly when debugging failed tests. For instance, `setResources` no longer works as intended in this "with showTutorial" block when run with Vitest.

eatyourgreens: They’re part of the resource models that Sarah created, I think.

eatyourgreens: https://github.com/zooniverse/front-end-monorepo/blob/9725ce6103078eb26fe3d09478e1ab0d356bf332/packages/lib-classifier/src/store/ResourceStore/ResourceStore.js#L76-L96

goplayoutside3: Ah ok, thank you! I was confused about how every store model had access to that method, but I see `setActive` now too.

eatyourgreens: I think I’ve still got a very old issue suggesting that the APIs for the classifier store models should be documented. At one point, I think we were going to add JSDoc descriptions to these actions.

eatyourgreens: I can't remember if there's a good reason that these lines don't use [`applySnapshot`](https://mobx-state-tree.js.org/concepts/snapshots), which would be the idiomatic MST way of adding `workflowTutorial` to the model.

goplayoutside3: Thanks! I'll keep that in mind.

PR #3048, opened 2022-05-04
Maintenance Week: QuickTalk tests

shaunanoordin:
## PR Overview

Part of: Maintenance Week

This PR adds tests for the QuickTalk feature, which have thus far been lacking proper tests.

- Most of the functionality tests are performed in `QuickTalk.spec.js`
- Basic tests have been added for the subcomponents
- `QuickTalkContainer.spec.js` meanwhile is missing more robust tests, due to issues with hooking it up to a mock store.
  - I think the problem is that this container might need to be rewritten as a functional component
  - For the record, this is what it needs to test for:
     - If store.project.experimental_tools contains 'quicktalk', then the QuickTalk button should render.
     - If store.project.experimental_tools doesn't contain 'quicktalk', then the QuickTalk button shouldn't render.
     - If there's no store or project, don't render.

### Status

I'm going to mark this as ready for review while I look into creating tests for other components, and rewriting the QuickTalkContainer into a functional component.

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
+++ b/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
@@ -41,7 +41,7 @@ function QuickTalk ({
   screenSize,
   expand = false,
   fixedPosition = true,
-  showBadge = true // HACK: Button.badge crashes tests AND storybook for an undetermined reason. // TODO: debug 
+  showBadge = true // HACK: Button.badge crashes tests AND storybook for an undetermined reason. // TODO: debug
 }) {
   const { t } = useTranslation('components')
   // TODO: figure out if/how the QuickTalk component should/could be displayed on mobile
@@ -89,6 +89,7 @@ function QuickTalk ({
         />
         <Button
           a11yTitle='Close comments panel.'
+          data-testid='quicktalk-close-button'
           icon={<Close size='small' />}
           onClick={() => setExpand(false)}
           plain
--- a/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
+++ b/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
@@ -69,9 +69,10 @@ function QuickTalk ({
 
   return (
     <QTPanel
+      a11yTitle='QuickTalk Comments Panel'
       elevation='medium'
+      role='dialog'
       background={{ dark: 'dark-3', light: 'light-3' }}
-      data-testid='quicktalk-panel'
     >
       <Box
         direction='row'
--- a/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
+++ b/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
@@ -50,7 +50,6 @@ function QuickTalk ({
   if (!subject) return null
 
   const [_expand, setExpand] = React.useState(expand)
-  const a11yTitle = `Subject has ${comments.length} comment(s). Click to expand.`
   const badge = (showBadge && comments.length > 0) ? comments.length : false
 
   const QTButton = (fixedPosition) ? FixedButton : Button
@@ -59,7 +58,7 @@ function QuickTalk ({
   if (!_expand) {
     return (
       <QTButton
-        a11yTitle={a11yTitle}
+        a11yTitle={t('QuickTalk.aria.openButton', { count: comments.length })}
         onClick={() => setExpand(true)}
         icon={<Chat />}
         badge={badge}
@@ -69,7 +68,7 @@ function QuickTalk ({
 
   return (
     <QTPanel
-      a11yTitle='QuickTalk Comments Panel'
+      a11yTitle={t('QuickTalk.aria.mainPanel')}
       elevation='medium'
       role='dialog'
       background={{ dark: 'dark-3', light: 'light-3' }}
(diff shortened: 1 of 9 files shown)


packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.spec.js
```diff
@@ -69,23 +68,24 @@ describe('Component > QuickTalk', function () {
...
-    
-    it('should expand when clicked', function () {
-      fireEvent.click(wrapper.queryByTestId('quicktalk-button'))
-      
-      expect(wrapper.queryByTestId('quicktalk-button')).to.not.exist()
-      expect(wrapper.queryByTestId('quicktalk-panel')).to.exist()
+
+    it('should expand when clicked', async function () {
+      const user = userEvent.setup({ delay: null })
+      await user.click(screen.queryByTestId('quicktalk-button'))
```
eatyourgreens, line 79:
```js
screen.queryByRole(‘button’, { name: ‘Close comments panel’ })
```

eatyourgreens: Getting buttons by role and label seems better than adding IDs.

shaunanoordin: Thanks Jim, I've updated these button queries with roles. I've also added role=dialog to the quick talk panel

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
```diff
@@ -70,9 +69,10 @@ function QuickTalk ({
 
   return (
     <QTPanel
+      a11yTitle='QuickTalk Comments Panel'
       elevation='medium'
+      role='dialog'
```
eatyourgreens, line 74: Dialogs have some specific requirements so be careful, when adding the `dialog` role, that your component is a valid dialog. The panel might also need `aria-modal`. 
https://www.w3.org/TR/wai-aria-practices-1.2/

packages/lib-classifier/src/components/Classifier/components/QuickTalk/components/Comment.spec.js
```diff
@@ -0,0 +1,55 @@
...
+  it('should render the comment', function () {
+    expect(screen.getByText('This is the first comment')).to.exist()
+  })
+
+  it('should render the author details', function () {
+    expect(screen.getByText('Zooniverse Tester')).to.exist()
+    expect(screen.getByText(/\(@\s*zootester\s*\)/)).to.exist()
+
+    expect(screen.getByAltText('Avatar for Zooniverse Tester')).to.exist()
+    expect(screen.getByAltText('Avatar for Zooniverse Tester')).to.have.property('src', 'https://static.zooniverse.org/www.zooniverse.org/assets/simple-avatar.png')
```
eatyourgreens, line 48: 👍🏻

Not blocking, but should the alt text be localised?

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
eatyourgreens: You’ll also need to make sure that the dialog is focused after it opens. 
https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-7

packages/lib-classifier/src/components/Classifier/components/QuickTalk/components/Comment.spec.js
shaunanoordin: All alt text and Aria labels should be localised, _ideally._

Last night I was struggling with getting the [i18n English translations to be injected into the tested components,](https://react.i18next.com/misc/testing) as without that working, I'd always end up with literally `<button role="button" aria-label="QuickTalk.aria.openButton" />` instead of `<button role="button" aria-label="Click to open Quick Talk" />`

Alternative would be to just use (literally) `screen.getByAltText('QuickTalk.aria.AvatarAltText')` but I honestly can't tell if querying by translation key instead of translated text is better or worse. (e.g. RetiredBanner.spec.js checks for the string "Banners.AlreadySeenBanner.bannerText" instead of the actual English text)

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
shaunanoordin: Noted. I did not realise using `role=dialog` would have a lot of additional functionality implications, but these will be addressed.

shaunanoordin:
Hang on, does rebasing a PR remove the PR Review Approval?

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
eatyourgreens: To be honest, a popup needs to implement those even if it doesn't explicitly declare `role=dialog` in the accessibility tree. After it opens, the reading focus should be at the first comment (or the top of the dialog) and Tab/shift-Tab should navigate you forward and backward through the interactive elements in the comments box.

eatyourgreens:
I haven't approved this yet.

eatyourgreens (approved):
Tests look good.

In terms of general UX, the Quick Talk dialog panel should be updated to behave like a dialog, so reading focus should jump to the open panel in a screenreader, and keyboard focus should jump to the open panel for anyone who is navigating the page with Tab/shift-Tab.

PR #3050, opened 2022-05-05
Run coverage reports on changes to main branch

eatyourgreens:
Split the `ci-tests` workflow into two workflows:
- `ci-build`, which builds the libraries, apps and storybooks after pushes to PRs.
- `coverage`, which runs the tests and generates a coverage report after pushes to PRs and merges to `master`. It can also be run manually from the Actions panel.

#2581 broke our coverage reports, so that the last build reported was 14 December 2021. This PR restores test and coverage reports on changes to `master`.
https://coveralls.io/github/zooniverse/front-end-monorepo

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -1,4 +1,4 @@
-name: Build and test
+name: test build
 
 on:
   pull_request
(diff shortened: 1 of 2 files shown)


goplayoutside3 (approved):
Looks good! 

I noticed that coveralls is automated to give a percentage report for every Panoptes-Front-End pull request. Should we do the same for FEM? If not, can we add the coveralls.io link to FEM's readme for easy reference?

eatyourgreens:
I can turn on comments for Coveralls, if it isn’t already on. 

It’s not at all obvious, but there’s a coveralls.io badge at the top of the Readme, which is also a link to the coverage report.

eatyourgreens:
PR comments should be turned on now, in https://coveralls.io/github/zooniverse/front-end-monorepo/settings

eatyourgreens:
Let’s see if this works.

PR #3114, opened 2022-05-07
Bump yarn to 1.22.18

eatyourgreens:
`yarn` is nagging that our version is out of date, so upgrade to latest allowed with `yarn policies set-version`.


shaunanoordin (approved):
## PR Review

This looks like a small update - I didn't have any issues running this on localhost and testing with e.g. lib-classifier. 👍

PR #3138, opened 2022-05-13
Classifier: RTL tests for TextTask

eatyourgreens:
Replace enzyme with React Testing Library for the text task. Test the task with and without suggestions.
- [x] test the annotation value after typing in the text box.
- [ ] test the annotation value after picking a suggestion.
- [ ] test the annotation value after applying a text modifier.

Breaking change: I've renamed `subTaskPreviousAnnotations` to `suggestions`, since that more accurately reflects what it does.

## Package
lib-classifier

Unit Test Conversion [Project Board](https://github.com/zooniverse/front-end-monorepo/projects/18)

## General


packages/lib-classifier/src/plugins/tasks/text/components/TextTask/TextTask.spec.js
```diff
@@ -1,247 +1,299 @@
...
-        expect(annotation.value).to.equal(expectedText)
+      textInput = screen.getByRole('textbox', {
+        name: task.instruction
       })
+      await user.type(textInput, 'Hello, this is some test text.')
+      await user.pointer([
+        {
+          target: textInput,
+          offset: 7,
+          keys: '[MouseLeft>]'
```
shaunanoordin, line 161: Oh wow, I did not realise [this](https://testing-library.com/docs/user-event/pointer#pressing-a-button-or-touching-the-screen) is how you action to define "mouse down (and then drag)"

Minor question: would you also need to add the "mouse up" action (`[/MouseLeft]`) after you move to offset 11?

shaunanoordin (approved):
## PR Review

Package: lib-classifier
Part of: Maintenance Week(s)

This PR updates the tests the Text Task. Looks good!

- I'm making a note of the fairly complex (for my test-writing skills) interactions are used in the "tagging text" section. This is very thorough.
- Minor question, out of curiosity: you've created a `withGrommet()` function to ensure the TextTask is wrapped in `<Grommet>`. Is there a reason for this? I noticed that some component tests are wrapped in Grommet (e.g. ConfirmModal), and I'm not sure if this has to do with the way user tests interacted with the component if it does/doesn't have the proper ZooTheme applied.

Changes LGTM 👍

eatyourgreens:
The `Grommet` wrapper is copied across from the Enzyme tests. I haven’t tried running the tests without it.

packages/lib-classifier/src/plugins/tasks/text/components/TextTask/TextTask.spec.js
eatyourgreens: Maybe. To be honest, I can’t remember if you need to explicitly add that here. It’s a good question.

PR #3163, opened 2022-05-19
Workaround Storybook 6.5 import bug

eatyourgreens:
Since Storybook 6.5.0, importing all of a module's named exports adds an extra `__namedExportsOrder` key, which then breaks the MST store in stories, where it expects each named import to be an MST model.

This PR works around the bug by filtering imports to check that each model is actually a model, using `type.isType`.

See https://github.com/storybookjs/storybook/issues/17587 but what you should notice on the main branch, at the moment, is that the classifier story fails for any story that uses `@test/mockStore`, because the store fails to build. This PR should fix that and get all the stories showing again.

## Package
lib-classifier

https://github.com/storybookjs/storybook/issues/17587

## How to Review
Build the classifier storybook with `yarn storybook` and check that the stories load without error. With Storybook 6.5.0, I'm seeing most of the subject viewer and task stories fail.

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/drawing/models/DrawingAnnotation.js
+++ b/packages/lib-classifier/src/plugins/tasks/drawing/models/DrawingAnnotation.js
@@ -3,7 +3,7 @@ import DrawingTask from './DrawingTask'
 import Annotation from '../../models/Annotation'
 import * as markTypes from '@plugins/drawingTools/models/marks'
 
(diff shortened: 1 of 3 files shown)


eatyourgreens:
The storybook bug should be fixed by https://github.com/storybookjs/storybook/pull/18284 in Storybook 6.5.4.

PR #3193, opened 2022-05-24
Classifier: Localise workflow task text

### LINKED ISSUE #3299: Text tasks don't observe the stored task for changes
## Package
lib-classifier

## Describe the bug
`DefaultTextTask` and `TextTaskWIthSuggestions` aren't wrapped in `observer`, so don't update when the task instruction changes eg. when changing task language.

## To Reproduce
Found while reviewing #3193, by opening one of the Every Name Counts workflows in the classifier, and changing the language from English to French or Spanish. The task language changes in the classifier store, but the displayed language remains English.

## Expected behavior
The task instruction should change when the page language changes.

## Additional context
`observer` was added to tasks in #1859.

From the MobX docs:
>observer only enhances the component you are decorating, not the components called by it. So usually all your components should be wrapped by observer. Don't worry, this is not inefficient. On the contrary, more observer components make rendering more efficient as updates become more fine-grained.

[Always read observables inside observer components](https://mobx.js.org/react-integration.html#always-read-observables-inside-observer-components).

eatyourgreens:
Following on from #3145, this localises `task.help`, `task.instruction` and `task.question` by converting them to views, which read their values from `task.strings.help`, `task.strings.instruction` and `task.strings.question` respectively.

I've tried to update as many tests and stories as possible to use translation strings, so this branch will be the base branch for all other work on workflow tasks.

Closes #3299.

## Package
lib-classifier

## How to Review
This dropdown task on staging has instructions in three languages.
https://localhost:8080/?project=908&workflow=3339

Every Name Counts, on production, has task instructions and questions in English and German. The subject content can be distressing, though.
https://localhost:8080/?project=arolsen-archives/every-name-counts&env=production&workflow=18126&language=de

https://user-images.githubusercontent.com/59547/170210370-923654f2-1064-4193-b01f-a9c2c48b320e.mov

# Checklist

## General

## General UX

## New Feature

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/Classifier.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.js
@@ -111,7 +111,7 @@ Classifier.propTypes = {
   subjectSetID: PropTypes.string,
   subjectID: PropTypes.string,
   workflowSnapshot: PropTypes.shape({
-    id: PropTypes.string
-  }),
-  workflowID: PropTypes.string.isRequired
+    id: PropTypes.string,
+    version: PropTypes.string
+  })
 }
--- a/packages/lib-classifier/src/components/Classifier/Classifier.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.spec.js
@@ -177,7 +177,7 @@ describe('Components > Classifier', function () {
   })
 
   describe('when the locale changes', function () {
-    let locale, subjectImage, tabPanel, taskAnswers, taskTab, tutorialTab, workflow
+    let locale, subjectImage, tabPanel, taskInstruction, taskAnswers, taskTab, tutorialTab, workflow
 
     before(async function () {
       sinon.replace(window, 'Image', class MockImage {
@@ -187,12 +187,43 @@ describe('Components > Classifier', function () {
           setTimeout(() => this.onload(), 500)
         }
       })
-      const subject = SubjectFactory.build({ locations: [{ 'image/png': 'https://foo.bar/example.png' }] })
-      const store = mockStore({ subject })
-      const project = store.projects.active
-      const projectSnapshot = { ...getSnapshot(project) }
-      workflow = store.workflows.active
-      const workflowSnapshot = { ...getSnapshot(workflow) }
+
+      const roles = []
+      const subjectSnapshot = SubjectFactory.build({ locations: [{ 'image/png': 'https://foo.bar/example.png' }] })
+
+      nock('https://panoptes-staging.zooniverse.org/api')
+          .persist()
+          .get('/field_guides')
+          .reply(200, { field_guides: [] })
+          .get('/project_preferences')
+          .query(true)
+          .reply(200, { project_preferences: [] })
+          .get('/project_roles')
+          .reply(200, { project_roles: [{ roles }]})
+          .get('/subjects/queued')
+          .query(true)
+          .reply(200, { subjects: [subjectSnapshot, ...Factory.buildList('subject', 9)] })
+          .post('/project_preferences')
+          .query(true)
+          .reply(200, { project_preferences: [] })
+
+      const checkBearerToken = sinon.stub().callsFake(() => Promise.resolve('mockAuth'))
+      const checkCurrent = sinon.stub().callsFake(() => Promise.resolve({ id: 123, login: 'mockUser' }))
+      const authClient = { ...defaultAuthClient, checkBearerToken, checkCurrent }
+      const client = { ...defaultClient, panoptes }
+
+      const workflowSnapshot = branchingWorkflow
+      workflowSnapshot.strings = workflowStrings
+
+      const projectSnapshot = ProjectFactory.build({
+        links: {
+          active_workflows: [workflowSnapshot.id],
+          workflows: [workflowSnapshot.id]
+        }
+      })
+
+      const store = RootStore.create({}, { authClient, client })
+
       const { rerender } = render(
         <Classifier
           classifierStore={store}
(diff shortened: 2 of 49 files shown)


eatyourgreens:
So many merge conflicts ☹️

goplayoutside3:
I think this is on the right track! I did find a bug though from the following actions:
1. I’m able to use the LocaleSwitcher to change https://local.zooniverse.org:3000/projects/eatyourgreens/-project-testing-ground/classify/workflow/3339 to French and back to English.
 
2. However, visiting https://local.zooniverse.org:3000/projects/arolsen-archives/every-name-counts/classify/workflow/17212?env=production (after adding `de` to NextJS config) and using the LocaleSwitcher does not always change the task instructions to German. Refreshing the page will solve the issue, and this bug seems intermittent. I’m able to reliably reproduce it by opening a fresh browser window (Firefox, Chrome, or Safari) and using the LocaleSwitcher once. Switching the locale multiple times eventually works as expected, but the first and second attempt to switch locale does not.
 
3. Storybook is looking good though! I only noticed that TextTask’s story with suggestions is not rendering `help`.

eatyourgreens:
I think #3239 might fix that.

eatyourgreens:
Oh, I missed out `task.strings` from the text task stories but that's fixed now.

On this branch, `workflow.strings` uses [`types.frozen`](https://mobx-state-tree.js.org/API/#frozen), which is assumed to be immutable. That might explain the task not updating when the language changes. The task panel also gets its props from `store.workflowSteps`, not `store.workflows`, so that could also explain the task not changing when the workflow changes in the store. #3239 tries to fix that by making `workflow.strings` observable, and updating the task strings when workflow tasks change.

goplayoutside3:
What's the best way to review this base branch and all of its branching PR's?

There are major bugs on this branch and I'm unsure if they're intended to be fixed by the other workflow translations PR's or if this one should be merged first.
1. When viewing https://local.zooniverse.org:3000/projects/arolsen-archives/every-name-counts/classify/workflow/17212?env=production, there are no task questions displayed.
2. Most of the Tasks storybook is broken on this PR.

eatyourgreens:
I can see task instructions for that workflow in the standalone classifier, in French at least. I haven't tried German.
Screenshot of the Buchenwald workflow for Every Name Counts, showing text inputs with French instructions.

eatyourgreens:
The storybook is fixed. #3239 changed the signature of `workflowSteps.setTaskStrings` so that was a pretty quick fix.

eatyourgreens:
I can see French text for Every Name Counts if I go directly to the French workflow:
https://local.zooniverse.org:3000/projects/fr/arolsen-archives/every-name-counts/classify/workflow/17212?env=production

I'm not sure why the standalone classifier works, but not the Classify page in the project app. I'm not really sure how to debug the `LocaleSwitcher`.

goplayoutside3:
> I'm not sure why the standalone classifier works, but not the Classify page in the project app. I'm not really sure how to debug the `LocaleSwitcher`.

I'm not sure that's the issue. I attached a screenshot of PFE where the workflow text areas have labels for "Birthday - Date" and those labels do not appear in any language in FEM on this branch.

PFE:
PFE

FEM:
FEM

goplayoutside3:
Edited my above comment because I meant to say that the dropdowns for Every Name Counts do not have their labels when this branch is run locally.

eatyourgreens:
That bug is hard to reproduce too. When I tried to record it, the language menu worked exactly as intended. I'm not really sure why there's such a big pause between selecting the language and seeing the page change (compared with the dev classifier, which is pretty quick.)


https://user-images.githubusercontent.com/59547/172955180-1c470dce-11b2-4738-b831-9696e2de0bc3.mov

eatyourgreens:
https://frontend.preview.zooniverse.org/projects/arolsen-archives/every-name-counts/classify/workflow/17212 doesn't have instructions for the dropdown tasks either, so they're broken on master.

eatyourgreens:
I inspected the workflow in the network panel, and the instructions for those select menus are empty strings.
Screenshot of the network panel, inspecting one of the dropdown tasks for the workflow. The task instructions are empty.

goplayoutside3:
Just following up on our slack discussion - the absence of titles above dropdown elements when running Every Name Counts in FEM is intentional. ADR [30](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-30.md) discusses a lot of things about the dropdown task built for FEM, but in short, those kind of Every Name Counts tasks are intended to be date pickers in FEM some day, so no need to address it here.

Before this PR is merged, I think it's important to understand why the other bug we've discussed happens. From me:
> ...using the LocaleSwitcher with Every Name Counts does not always change the task instructions to German. I’m able to reliably reproduce it by opening a fresh browser window (Firefox, Chrome, or Safari) and using the LocaleSwitcher once.

From you:
> That bug is hard to reproduce too. When I tried to record it, the language menu worked exactly as intended. I'm not really sure why there's such a big pause between selecting the language and seeing the page change (compared with the dev classifier, which is pretty quick.)

Although PH-TESS and the Dual Language dropdown [workflow](http://local.zooniverse.org:3000/projects/eatyourgreens/-project-testing-ground/classify/workflow/3339?env=staging) do not have an issue switching locales, I think it'd be valuable to try and create a test workflow that does reproduce the same bug as Every Name Counts, so it can be solved before more complicated project workflows are migrated to FEM.

eatyourgreens:
@goplayoutside3 have you got any ideas why changing the locale prop works in the standalone classifier, but the `LocaleSwitcher` component is buggy in the project app? I’m baffled.

goplayoutside3:
> @goplayoutside3 have you got any ideas why changing the locale prop works in the standalone classifier, but the `LocaleSwitcher` component is buggy in the project app? I’m baffled.

I don't off the top of my head, but I'll dig a little deeper next week!

eatyourgreens:
Same here. It’s a weird bug. I might start by writing some tests that simulate mounting the classifier with different locales, but the same global store.

eatyourgreens:
I think I've reproduced that bug. I've taken a screenshot in case I can't get it to trigger again. I also think that it's present on master. We only notice it on this branch because this branch displays `task.strings.instruction` in place of `task.instruction`.

In this screenshot, you can see that the page locale (NextJS) is `es`, the classifier locale and workflow translation (logged in the console) are also `es`, but the task is displayed in English.

Screenshot of Every Name Counts, showing English task text when the selected language is Español.

The tutorial is loaded in Spanish, so this bug is specific to the task translations in `store.workflowSteps`, which are not showing changes to `store.workflows.active.strings`.

Screenshot of Every Name Counts, showing Spanish tutorial text when the selected language is Español.

eatyourgreens:
Logging `workflowStrings`, in the classifier component, shows task text in the selected language, even though the task panel shows English text. The workflow store is updating as expected when the classifier locale changes, but the changes aren't being passed on to workflow steps and the task panel.

We also know, now, that the bug is in the `Classifier` component (specifically, the store) and not in the NextJS app.

eatyourgreens:
@goplayoutside3 if I switch to the tutorial tab, then back to the task tab, the task does re-render with the correct change. I'm not sure, but I think the bug could be in the task panel itself, in that it might not be re-rendering when the store changes.

eatyourgreens:
I've added a `Classifier` test that checks the task panel text after a language change. I expect it will fail because of this bug.

eatyourgreens:
@goplayoutside3 I think I've fixed it. Task components should be wrapped in `observer`. At some point, the `TextTask` was split into `DefaultTextTask` and `TextTaskWithSuggestions`. Wrapping both those components in `observer` seems to fix the bug: they re-render when the `task` prop changes.

eatyourgreens:
This explains the bug:

>observer only enhances the component you are decorating, not the components called by it. So usually all your components should be wrapped by observer. Don't worry, this is not inefficient. On the contrary, more observer components make rendering more efficient as updates become more fine-grained.

From https://mobx.js.org/react-integration.html#always-read-observables-inside-observer-components

goplayoutside3 (approved):
Looks good ! Locale switches in Every Name Counts work as expected and the task text is translated. Great explanations while debugging btw! Those points are super helpful for understanding the bug and how the store was the root of the cause.

Non blocking note - I noticed that Transcription Task and Simple Dropdown stories do not work with `isThereTaskHelp` - does it make sense to fix on this PR? Simple Dropdown’s `isThereTaskHelp` doesn’t work on the deployed version of FEM’s storybook either.

goplayoutside3:
> Should we open an issue about tasks being stored in both `workflow.tasks` and `step.tasks`, which makes it hard to track changes to tasks?

I do this adds to the confusion of tasks - is it preferable to store them in one place? For example, does `steps.tasks` become obsolete if tasks are only stored in `workflow.tasks`? If it's too late at this point in building the mobx classifier store, we could add some clear notes in the /store Readme about how the two relate.

eatyourgreens:
I think the stories without help might be stories where the mock help is an empty string. Should be an easy fix but it can go in a separate PR.

PR #3213, opened 2022-05-26
QuickTalk: Refactor to functional component

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Affects: experimental QuickTalk feature (only affects projects with experimental tool enabled)

This PR refactors the QuickTalkContainer into a _functional component._

- No actual functions or features have been changed (EDIT: except for the two minor changes below). Users should see no difference in the feature.
- Minor: error messages now use translated text.
- Minor: when QuickTalk panel opens, close button now automatically gets focus. (I thought I added this in 3048, but I must have dropped it, d'oh!)

### Testing

- Test on `lib-classifier` (not app-project) with a project with QuickTalk enabled, e.g. https://local.zooniverse.org:8080/?env=staging&project=darkeshard%2Ftest-project-2022&workflow=3581
- You should be able to see the QuickTalk button on the lower-right of the screen, with a number _if_ any comments already exist.
- If you click on the button, the QuickTalk panel should expand, and you should see any comments already made.
- If you're logged in, you should be able to post comments on the expanded panel.

Known UI issues: (collapsed) QuickTalk button should have a background so it doesn't blend into the page (z-index madness) ; a message saying "you should be logged in if you want to post a message" should appear on the expanded panel.

### Status

Ready for review! 👍

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
@@ -148,12 +148,12 @@ function QuickTalkContainer ({
 
       const authorization = await getBearerToken(authClient)  // Check bearer token to ensure session hasn't timed out
       const user = await authClient.checkCurrent()
-      if (!authorization || !user) throw('User not logged in')
+      if (!authorization || !user) throw new Error(t('QuickTalk.errors.noUser'))
 
       // First, get default board
       const boards = await talkClient.type('boards').get({ section, subject_default: true })
       const defaultBoard = boards && boards[0]
-      if (!defaultBoard) throw('A board for subject comments has not been setup for this project yet.')
+      if (!defaultBoard) throw new Error(t('QuickTalk.errors.noBoard'))
 
       // Next, attempt to find if the Subject already has a discussion attached to it.
       const discussions = await talkClient.type('discussions').get({
--- a/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
@@ -93,15 +93,8 @@ function QuickTalkContainer ({
   function fetchComments () {
     resetComments()
 
-    console.log('+++ DEBUG')
-    return
-
-    /*
-    const subject = this.props?.subject
     const project = subject?.project
-    if (!subject || !project) {
-      return
-    }
+    if (!subject || !project) return
 
     const section = 'project-' + project.id
     const query = {
@@ -113,20 +106,20 @@ function QuickTalkContainer ({
     }
 
     talkClient.type('comments').get(query)
-      .then (comments =>{
-        this.setState({ comments })
+      .then (allComments =>{
+        setComments(allComments)
 
         let author_ids = []
         let authors = {}
         let authorRoles = {}
 
-        author_ids = comments.map(comment => comment.user_id)
+        author_ids = allComments.map(comment => comment.user_id)
         author_ids = author_ids.filter((id, i) => author_ids.indexOf(id) === i)
 
         apiClient.type('users').get({ id: author_ids })
           .then(users => {
             users.forEach(user => authors[user.id] = user)
-            this.setState({ authors })
+            setAuthors(authors)
           })
 
         talkClient.type('roles')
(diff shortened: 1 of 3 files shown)


packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
```diff
@@ -174,21 +148,21 @@ class QuickTalkContainer extends React.Component {
 
       const authorization = await getBearerToken(authClient)  // Check bearer token to ensure session hasn't timed out
```
eatyourgreens, line 149: I think I’ve pointed this bug out before but this will return an empty string if you call it before you call `checkCurrent`. See #2986.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
```diff
@@ -198,11 +172,10 @@ class QuickTalkContainer extends React.Component {
         }
 
         await talkClient.type('comments').create(comment).save()
```
eatyourgreens, line 174: This could use `panoptes.post` to post comments with the new API client.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
```diff
@@ -101,24 +86,24 @@ class QuickTalkContainer extends React.Component {
...
         let author_ids = []
         let authors = {}
         let authorRoles = {}
 
-        author_ids = comments.map(comment => comment.user_id)
+        author_ids = allComments.map(comment => comment.user_id)
         author_ids = author_ids.filter((id, i) => author_ids.indexOf(id) === i)
-        
+
         apiClient.type('users').get({ id: author_ids })
```
eatyourgreens, line 101: This should use `panoptes.get` in the new API client.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
```diff
@@ -101,24 +86,24 @@ class QuickTalkContainer extends React.Component {
...
           .then(users => {
             users.forEach(user => authors[user.id] = user)
-            this.setState({ authors })
+            setAuthors(authors)
           })
-        
+
         talkClient.type('roles')
```
eatyourgreens, line 107: Are Talk roles different from Panoptes roles? You can use the new client to get roles. The `useProjectRoles` hook has an example.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
```diff
@@ -101,24 +86,24 @@ class QuickTalkContainer extends React.Component {
       page: 1,
       sort: 'created_at',  // PFE used '-created_at' to sort in reverse order, and I have no idea why.
     }
-        
+
     talkClient.type('comments').get(query)
```
eatyourgreens, line 90: The new API client could be used here too.

eatyourgreens:
There’s an auth bug in the existing code, which can be fixed here. 

All the API calls should use the new `panoptes` client but maybe do that in a separate PR? You shouldn’t need to import any code from the old `panoptes-client` and you aren’t doing anything here that isn’t being done with the new client elsewhere in the classifier.

eatyourgreens (approved):
Looks good but:
- there's an auth bug which blocks the first posted comment from being posted.
- all API requests should really be handled with `panoptes-js` not the old client.
- after I post a comment, keyboard focus goes back to the window, or parent document. I'm not sure. Either way, posting a comment from the keyboard works until after the comment is posted, at which point the Quick Talk panel loses keyboard focus.

eatyourgreens:
This might be useful for [managing keyboard focus](https://medium.com/swlh/improve-accessibility-in-your-react-app-by-managing-focus-in-mutable-content-4ddf4ed92186).

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
```diff
@@ -174,21 +148,21 @@ class QuickTalkContainer extends React.Component {
 
       const authorization = await getBearerToken(authClient)  // Check bearer token to ensure session hasn't timed out
       const user = await authClient.checkCurrent()
```
eatyourgreens, line 150: Another thought about auth: Panoptes user sessions are quite long-lived, so call this once, when the component mounts, in order to establish a session (if one exists.)

Then get a bearer token as and when you need access to perform a restricted action, like posting a comment.

shaunanoordin:
Thanks Jim! 👍 I'm merging this PR, with the following goals for the follow up PR:

- "First comment not posting" auth bug to be investigated
- apiClient and talkClient to be changed to use new panoptes-js client.
- (Address UI visual + accessibility improvements? Might put that in a third PR)

eatyourgreens:
This is the auth bug. 
https://github.com/zooniverse/front-end-monorepo/pull/3213#discussion_r882256896

PR #3216, opened 2022-05-26
Clean up Yarn

eatyourgreens:
Delete all the old `yarn` versions that have built up in `.yarn/releases`.


eatyourgreens:
Over 1 million deleted lines!

shaunanoordin (approved):
## PR Review

Changes look good - we're just yeeting some really old, _no-longer-used code_ from the repo. LGTM 👍

Minor side note, yarn 1.22.15 onwards uses `.cjs` instead of `.js`? And that's a _patch version,_ not even a minor version change? Hmm.

eatyourgreens:
`.cjs` is CommonJS, which is mostly replaced by ES Modules now.

PR #3244, opened 2022-06-01
QuickTalk: replace old Panoptes client with talkAPI

shaunanoordin:
## PR Overview

Package: `lib-classifier` (and a bit of `panoptes-js`)
Follows: #3240 (introduction of talkAPI to `@zooniverse/panoptes-js`), #3213 (refactor QuickTalk to functional component), and #3221 (fix issue with user auth)

This is part of QuickTalk's refactor & re-coding. Previously, QuickTalk used the old Panoptes API, which caused several issues including: 1. you couldn't post comments when running on app-project, because the client used on app-project and lib-classifier are different (and the latter isn't authenticated when the former is), and 2. an environment mismatch for the app-project client and the lib-classifier client.

With this PR, QuickTalk uses the new talkAPI introduced to `@zooniverse/panoptes-js` in 3240

- Features & functionality should remain the same, with one exception...
- ❗ Major fix: ...QuickTalk now works properly on app-project. A logged-in user should be able to post a Comment for a Subject.
- Refactors/recoding:
  - All instances of `panoptes-client` has been removed.
  - All GET-ing and POST-ing actions now use talkAPI and have been split into helper functions.
- QuickTalk continues to check for user + auth whenever: 1. a new Subject loads, or 2. a comment is posted.
- ❗ Bug fix, panoptes-js: `talkAPI.put()` and `talkAPI.post()` now passes the correct arguments to the `panoptes` client the talkAPI is wrapped around.

### Misc Dev Notes

- Helper functions currently don't have tests, as I'm unsure if they'd be useful when what we'd need to test for is the (shape/type of) responses from the Talk API.
- On the other hand, QuickTalkContainer is now a good candidate for expanded tests, but I still need to figure out an elegant way to shimmy in stubPanoptesJS(). 🤔 Explicitly passing in the client (e.g. `postTalkComment (text, discussion, user, authorization, /* this -> */talkAPI)`) got slightly messy.
- EDIT: I might need to try this anyway on a separate PR, messy or not.

### Testing

Basic testing: ensure QuickTalk continues to function as expected.

- Test on `lib-classifier` (not app-project) with a project with QuickTalk enabled, e.g. https://local.zooniverse.org:8080/?env=staging&project=darkeshard%2Ftest-project-2022&workflow=3581
- You should be able to see the QuickTalk button on the lower-right of the screen, with a number if any comments already exist.
- If you click on the button, the QuickTalk panel should expand, and you should see any comments already made.
- If you're logged in, you should be able to post comments on the expanded panel.
  - There is actually a very subtle but important difference between 1. posting a Comment on a Subject _with_ existing Comments, and 2. posting a Comment on a Subject _without_ existing Comments.
  - If you want to test posting a comment on a "fresh" Subject, try https://local.zooniverse.org:8080/?env=staging&project=darkeshard%2Ftest-project-2022&workflow=3582

Advanced testing: QuickTalk should also work on app-project.

- Same tests, but do it on `app-project` with a project with QuickTalk enabled, e.g. https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022/classify/workflow/3581

#### General

#### Refactoring

### Status

Ready for review!


eatyourgreens:
Code changes look good after a very quick initial read. I still need to checkout this branch and run it locally.

In terms of testing, I would mock the API with `nock`, rather than stub `panoptes-js`, and mount the component with the mocked API. I wouldn't worry about testing individual functions. These [notes on testing hooks](https://react-hooks-testing-library.com/#when-not-to-use-this-library) also apply here. Your data-fetching functions can be tested by testing the `QuickTalk` component and having it fetch (or post) mock data.

eatyourgreens (approved):
Noting that the URL is https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022/classify/workflow/3581?env=staging if you want to test with a production build of the project app.

This looks good. I can read comments, and post if I'm logged in. The posting UX is a bit clunky: there was a long post while I saw a 'posting comment' message. Since you already have the comment on the client, you could add it to the comments list while posting in the background for a smoother experience. I'd also recommend focusing the new comment after posting.

When I'm not logged in, the bottom of the comments list could have a call to action like 'Log in to join the conversation.' or similar.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
```diff
@@ -72,52 +78,34 @@ function QuickTalkContainer ({
     setUserId((authorization && user) ? user.id : undefined)
   }
 
-  function fetchComments () {
+  async function fetchComments () {
```
eatyourgreens, line 81: This function is an effect, so define it inside the `useEffect` hook. At the moment, you generate a new function on each component render.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
```diff
@@ -72,52 +78,34 @@ function QuickTalkContainer ({
     setUserId((authorization && user) ? user.id : undefined)
```
eatyourgreens, line 78:
```suggestion
    setUserId(user?.id)
```

`authorization` isn't needed here, since you aren't making an API request with an Auth header.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
eatyourgreens: Same comment for the `checkUser` function. Also, `useEffect` should watch for changes to any of the external dependencies of these functions, like `authClient` and `project`.

eatyourgreens (commented):
A couple of comments about tightening up the `useEffect` hook.

eatyourgreens:
I was also going to ask if `panoptes-client` should be removed from the classifier, but it doesn't seem to be in `package.json`. I guess it's being hoisted from one of the other packages in the monorepo.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js
shaunanoordin: 👍

shaunanoordin:
Thanks for the review, Jim!

> I was also going to ask if panoptes-client should be removed from the classifier, but it doesn't seem to be in package.json. I guess it's being hoisted from one of the other packages in the monorepo.

(1) I can see that `"panoptes-client": "~3.4.1",` _is_ still being used by `lib-classifier/package.json`, but it's only invoked in `lib-classifier/dev/components/App.js` to initialise oAuth on localhost.

(2) UI/UX issues are noted, and will be addressed in a separate PR.

(3) I'm not familiar enough with good Hook patterns to properly optimise the code as suggested, (my brain still thinks in terms of explicit steps,) but I am making a note of it for a future follow up. e.g. I can understand moving the checkUser() code into the useEffect() hook, but fetchComments() is invoked both on subject change and in postComment(), so I haven't figured out how that'd work.

(4) As mentioned, the QuickTalkContainer needs better tests. Thanks for the suggestion to use `nock` - that makes more sense to mock the Talk API server instead of messing about with the client.

I'm going ahead and merging this PR, then continuing on building up on it. The UI/UX follow up PR takes priority. 👌

eatyourgreens:
I find that invoking `React.StrictMode` can be very useful in finding bugs like the ones I’ve noted in the `useEffect` hook.

eatyourgreens:
Your effect hook should be written like this. However, it will run every time you create new `checkUser` and `fetchComments` functions, which is on every render at the moment. 

```js
useEffect(function () {
  fetchComments()
  checkUser()
  return () => { // cancel unfinished API requests here. }
}, [checkUsers, fetchComments])
```

eatyourgreens:
`fetchComments` could use `useMemo` or `useCallback` to stop it being generated on every render.

eatyourgreens:
I also just realised that this merged PR is going to hit the `/me` endpoint on each new subject. You should only request from that endpoint when a user session starts. 

The classifier actually checks `/me` when it mounts, so `checkUser` might be redundant here.

eatyourgreens:
Shouldn’t `postComments` call `setComments` rather than `fetchComments`? I missed that when I was reviewing this.

eatyourgreens:
The cleanup function for `fetchComments` should run `resetComments()` too. I missed that.

PR #3250, opened 2022-06-02
Fix ImageAndTextViewer not calling onReady or onError

mcbouslog:
## Package
- lib-classifier

## Describe your changes
- pass onError and onReady from SubjectViewer to ImageAndTextViewer to either SingleImageViewer or SingeTextViewer

## How to Review
  - fix - https://local.zooniverse.org:8080/?project=1952&workflow=3594&demo=true
  - bug - https://frontend.preview.zooniverse.org/projects/markb-panoptes/digileap-testing/classify/workflow/3594?env=staging
- What user actions should my reviewer step through to review this PR? try to edit text in textFromSubject task (text input)
- Which storybook stories should be reviewed? I think N/A, bug in task related to code missing in viewer, separate stories
- Are there plans for follow up PR’s to further fix this bug or develop this feature? yes! the viewers are inconsistently getting some subject viewer store props passed from the SubjectViewer component and from the individual viewer connectors. I'm going to create an issue to refactor for consistency. 

# Checklist

## General

## General UX

## Bug Fix
- [x] The bug is fixed

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ImageAndTextViewer/ImageAndTextViewer.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ImageAndTextViewer/ImageAndTextViewer.spec.js
@@ -2,6 +2,7 @@ import { expect } from 'chai'
 import { Grommet } from 'grommet'
 import { Provider } from 'mobx-react'
 import React from 'react'
+import sinon from 'sinon'
 import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import asyncStates from '@zooniverse/async-states'
@@ -172,6 +173,22 @@ describe('ImageAndTextViewer', function () {
       expect(screen.getByRole('button', { name: 'SlideTutorial.StepNavigation.previous' })).to.exist()
       expect(screen.getAllByRole('radio', { name: 'SlideTutorial.StepNavigation.go' })).to.have.lengthOf(2)
       expect(screen.getByRole('button', { name: 'SlideTutorial.StepNavigation.next' })).to.exist()
+    })
+
+    it('should call the onReady prop', function () {
+      const onReadySpy = sinon.spy()
+
+      render(
+        <ImageAndTextViewerConnector
+          loadingState={asyncStates.success}
+          onReady={onReadySpy}
+          subject={store.subjects.active}
+        />, {
+          wrapper: withStore(store)
+        }
+      )
+
+      expect(onReadySpy).to.have.been.calledOnce()
     })
 
     describe('when the frame is changed', function () {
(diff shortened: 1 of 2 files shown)


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ImageAndTextViewer/ImageAndTextViewer.spec.js
```diff
@@ -174,6 +175,22 @@ describe('ImageAndTextViewer', function () {
       expect(screen.getByRole('button', { name: 'SlideTutorial.StepNavigation.next' })).to.exist()
     })
 
+    it('should call the onReady prop', function () {
```
mcbouslog, line 178: I tried to add a similar test for an image location first, but I couldn't figure out with [the `useSubjectImage` hook and other related loading code](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.js#L40-L53).

eatyourgreens: If your mock image URL returns 404, then `onError` will probably fire rather than `onReady`. 

I’m not sure if JSDOM supports image load events.

goplayoutside3 (approved):
Looks good to me, bug is fixed!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ImageAndTextViewer/ImageAndTextViewer.spec.js
mcbouslog: I tried variations on the mock subject image location, including a nonsense url, url that would error, and url that I thought would succeed, and none worked as I expected with `onReady` or `onError` to be called or not called. I think I'm missing something fundamental, maybe there's a completely different way to structure the tests to accomplish desired result, or maybe as you state JSDOM doesn't support image load events. I'll revisit (we'll have to when refactoring SingleImageViewer with RTL?), but will merge this for now to fix the bug.

eatyourgreens: This is for an older version of the events library, but [explicitly firing a load event](https://stackoverflow.com/questions/60481331/testing-img-onload-img-onerror-using-jest-and-react-testing-library) on the image might work: `fireEvent.load(image)`. The syntax probably needs to be updated for version 14 of the user events library.

eatyourgreens: I'd also guess that the image in the tests doesn't have any layout, since we aren't rendering it in a browser, so no `clientHeight`, `clientWidth` etc.

Issue #2233, opened 2021-06-07
Transcription Task modal is obscuring text being transcribed


snblickhan:
## Is your feature request related to a problem? Please describe.
Users on Beyond Borders are reporting (see this [Talk thread](https://www.zooniverse.org/projects/mainehistory/beyond-borders-transcribing-historic-maine-land-documents/talk/3952/2019110)) that the transcription task modal consistently obscures the text they're trying to transcribe. This was a similar report from ASM, though they note that it is nice that they can actually move the box around (an improvement from the beta version!). 

## Describe the solution you'd like
The modal should pop up *below* the placement of both dots. Whatever our current metric is for placement, we should increase it.

## Describe alternatives you've considered
None

## Additional context
This was a request from the ASM days, so it may be that whatever we put in place to address that wasn't enough. I think it's worse on projects with closely-spaced lines. For example, it's not that common on the Davy Notebooks Project, but I just tried a few pages on [Beyond Borders](https://www.zooniverse.org/projects/mainehistory/beyond-borders-transcribing-historic-maine-land-documents) and had to adjust every single modal because it consistently covered up the text I was trying to transcribe.

snblickhan:
Additional information/context from a volunteer (via DM):
>On the much belabored Transcription Box location in the Maine Land Document project, I've been trying to figure out how the placement is done by the software. Frankly, it makes no sense. Clearly, the box placement is related to the transcription line that it belongs to. However, there is no consistency whatsoever in that relationship that I can determine. It seems that, typically, the top of the box is below the line enough to recognize descenders and some punctuation - but only in the first few lines of a transcription (full page, 40-60 or more lines of text). It then shifts to just barely below the line and then could be creeping toward and past the line a pixel width per line or possibly be stable for a few lines before it moves up. At the bottom of the page, it often is covering half the line or more (on my monitor at least).

>Now, I'm aware that there are many intermediate factors betwen what is planned and what is seen in this type of graphic interface. And different monitors (and hopefully to a lesser extent browsers and OSs) play a part. BUT, the location should consistent relative to every transcription line and not variable as it is. Then, one could adjust to however it appears on their set up. Now, there is no possible adjustment that will work for an entire document other than moving the box away from the transcription line.

eatyourgreens:
[...]

PR #3320, opened 2022-06-15
Transcription workflows: adjust subtask position

eatyourgreens:
- Change the minimum popup height to 200px.
- Include the visible window in position calculations.
- When the popup is below the transcribed line, place the top left corner on the bottom of the transcribed line's bounding box, not in the middle of the line.
- When the transcribed line is towards the bottom of the screen, position the popup above the transcribed line, allowing an offset for the minimum popup height and the transcribed line's bounding box.

## Package
lib-classifier

Towards #2233. (I'm not sure if the changes here would close that issue for all possible transcription documents and workflows.)

## How to Review
https://localhost:8080/?project=mariaedgeworthletters/maria-edgeworth-letters&env=production

Try transcribing some lines towards the top or the bottom of the viewport. The popup shouldn't appear outside the visible viewport any more, and should be better positioned to avoid obscuring the written lines.

Screenshot of the transcription task, with a purple line at the bottom of the window and the text task popup positioned toward the centre of the visible window area.

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/SubTaskPopup/SubTaskPopup.js
@@ -11,7 +11,7 @@ import ConfirmModal from './components/ConfirmModal'
 import SaveButton from './components/SaveButton'
 
 const MIN_POPUP_WIDTH = 350
-const MIN_POPUP_HEIGHT = 100
+const MIN_POPUP_HEIGHT = 200
 
 /**
   A popup that renders activeMark.tasks for the active mark. Incomplete task annotations are confirmed, on save or close, for required tasks.
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/helpers/getDefaultPosition/getDefaultPosition.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/helpers/getDefaultPosition/getDefaultPosition.js
@@ -1,7 +1,11 @@
-const MIN_HEIGHT = 100
+const MIN_HEIGHT = 200
 const MIN_WIDTH = 350
 
 export default function getDefaultPosition (bounds = {}, minHeight = MIN_HEIGHT, minWidth = MIN_WIDTH) {
+  const viewport = {
+    height: window?.innerHeight,
+    width: window?.innerWidth
+  }
   // Calculate default position
   let x = 0
   let y = 0
@@ -19,8 +23,18 @@ export default function getDefaultPosition (bounds = {}, minHeight = MIN_HEIGHT,
     const markWidth = bounds.width || 0
     const markHeight = bounds.height || 0
 
-    x = markX + markWidth * 0.5
-    y = markY + markHeight * 0.5
+    const xOffset = markWidth * 0.5
+    let yOffset = markHeight
+
+    // Get the distance from the bottom of the mark to the bottom of the viewport.
+    const verticalSpace = viewport.height - (markY + markHeight)
+    // If the available space is too small, push the popup upwards.
+    if (verticalSpace < minHeight) {
+      yOffset = 0 - (minHeight + markHeight)
+    }
+
+    x = markX + xOffset
+    y = markY + yOffset
   }
 
   // Keep within bounds of the viewport


eatyourgreens:
I've assumed that `clientHeight` for the SVG line is roughly equivalent to the line height on the written page, but that seems to mostly work.

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`

This PR fixes an issue with the **vertical placement** of the Transcription Task's subtask popup.

- Baseline: for a project with a Transcription Task, after either (1) a volunteer places a line (click to place start point, then end point) OR (2) clicks on an existing/aggregated line, the **subtask popup** appears.
  - This popup (with a Text sub-Task) _should_ appear in a position that easily allows volunteers to view the text box and being/continue transcribing text.
  - At the moment however, the popup appears in such a way that its top-left corner is positioned at the mid-point between the start point and end point.
- The problem: if the transcribed line appears **near the bottom of the screen (viewport)**, then the subtask starts appearing "off screen"
    Screenshot 2022-06-15 at 21 28 09
    _Pictured: we created a transcription line at the bottom of the screen, causing the subtask popup to be "cut off at the bottom". master branch._
- The fix: PR fixes the issue by simply switching the position of the Subtask Popup to appear ABOVE the line, IF Subtask would appear off screen (below the viewport).
    Screenshot 2022-06-15 at 22 12 42
    _Pictured: same, we created a transcription line at the bottom of the screen, but the subtask popup appears above the line, meaning it's still in view of the user. This PR's branch._

Tests look good to me, specifically when it comes to constraining **the vertical position,** and the vertical position only.

⚠️ One thing to note is that this PR is reviewed with the following knowledge:
_Jim, via Slack: ... It’s more an improvement on the existing code, which can position the popup outside the visible window, rather than an attempt to fix all possible positioning scenarios._

I'll still document all the other _(incredibly unlikely and very niche)_ scenarios that are not covered by this PR in the Dev Notes (below), but those notes should _not_ count against this PR.

### Testing

Tested on lib-classifier on localhost with macOS +Chrome 102

- Testing URL: https://localhost:8080/?project=mariaedgeworthletters/maria-edgeworth-letters&env=production - a project/WF with one standard Transcription Task.
- I placed transcription lines on various places and observed the position where the Subtask Popup appears.
- Placed a _standard horizontal transcription line_ on the page, with the viewport positioned so that the line appears _near the top of the screen._
  - Subtask appears below the line, within the visible screen. 🆗 
- Placed a _standard horizontal transcription line_ on the page, with the viewport positioned so that the line appears _near the bottom of the screen._
  - Subtask appears above the line, within the visible screen. 🆗 
  - ❔ ~~The Subtask popup _can_ appear directly in front of (blocking) the green line in some cases (see picture above), but I don't think this is a notable issue.~~ _This only occurs at wonky zoom levels + screen sizes._
- Same as above, but RTL instead of LTR
  - Same results 🆗 

Looks good! 👍 

### Dev Notes

Here are a few scenarios that _aren't_ covered by this PR's fix. The following are NOT blocking comments - just read 'em and then file 'em away in your mind palace.

- Subtask doesn't compensate for limited horizontal space, e.g. on mobile (portrait) view
    Screenshot 2022-06-15 at 21 28 47
    _Above: window was narrowed horizontally until the layout of the classifier switched into mobile/portrait mode, then a standard horizontal line was placed. Theoretically it's possible to compensate the xOffset and popupWidth, but that's for another PR._
- Subtask popup position acts _slightly_ differently for _vertical lines_ placed near the bottom of the screen: specifically, for _tall_ vertical lines vs _short_ vertical lines.
  - For a short vertical line, the popup appears near the line:
      Screenshot 2022-06-15 at 22 07 48
  - For a tall vertical line (from bottom _up to_ 50% viewport height), there's a big gap between the line and the popup.
      Screenshot 2022-06-15 at 22 07 29
  - This is due to the yOffset calculating for markHeight: the taller the line, the higher the popup appears.
  - This is in practice a non-issue, though. Even if the position might be a bit odd, the popup is still vertically bound to the visible viewport.
  - This might only mildly bother somebody who's working with a really tall vertical monitor.
- Subtask position calculations don't work when viewport is too small. ...because of course. Why did you make your window so small? (I still tested it anyway.)

### Status

LGTM! 👍 Let's get this merged for the volunteers.

eatyourgreens:
A refinement could be to check the aspect ratio of the SVG mark’s bounding box, then have different behaviour for horizontal and vertical boxes.

eatyourgreens:
Another refinement would be better tests. The current tests check that the `x,y` coordinates are numbers but they could test a mocked box against the mocked window height and width. We have a helper for out-of-bounds shapes which might be useful here.

Issue #3210, opened 2022-05-25
Refactor multi-frame viewer


mcbouslog:
## Package
- lib-classifier

Refactor the multi-frame viewer to use the single image viewer (connector and all) similar to the image and text viewer to remove duplicate code and simplify image (and related viewer) state to fewer components.

Issue #3093, opened 2022-05-06
Hide Tools in ImageToolbar for Certain SubjectViewers


ErikOstlund:
## Package
`lib-classifier`

## Is your feature request related to a problem? Please describe.

For some SubjectViewers, e.g. the SingleVideoViewer, the tools in the ImageToolbar are not needed. This was discussed and confirmed with @mcbouslog 

- MoveButton
- ZoomInButton
- ZoomOutButton
- RotateButton
- FullscreenButton
- ResetButton

![Screen Shot 2022-05-06 at 11 19 09 AM](https://user-images.githubusercontent.com/9533007/167172657-23c21387-e685-407c-ab4b-d46f98102a13.png)

## Describe the solution you'd like

When certain SubjectViewers are loaded, the above listed tools should not be visible (should not render).

The FieldGuide should always be visible.

## Describe alternatives you've considered

We discussed having all tools listed above being visible but disabled. This may confuse volunteers and is not an ideal solution.

eatyourgreens:
PH TESS hides the Rotate button, so we could revisit how that was done. I think it’s a flag that’s set in the store.

eatyourgreens:
The solution might be as simple as making the default handlers null, then only rendering a button if it has a handler defined by the current subject viewer.

PR #3325, opened 2022-06-16
Fix invert button for multi-frame subjects and line color consistency

### LINKED ISSUE #3328: Invert button not working as expected
## Package
- lib-classifier

## Describe the bug
- invert button not working with multi-frame viewer
- invert button inverting drawing/transcription task line color (color should remain consistent)
- issues are related as underlying problem relates to where filter invert is conditionally added (i.e. parent SVG, or SVG image) and Safari compliance (Safari doesn't support filter invert on SVG image element)

## To Reproduce
Steps to reproduce multi-frame bug:
1. Enable invert button in InvertButtonContainer 
2. Click on https://local.zooniverse.org:8080/?project=humphrydavy%2Fdavy-notebooks-project&env=production&demo=true&workflow=18575
3. Click invert button
4. Bug - no image invert

Steps to reproduce line color bug:
1. Enable invert button in InvertButtonContainer 
2. Click on https://local.zooniverse.org:8080/?project=markb-panoptes%2Ffem-drawing-tools&demo=true&workflow=3344
3. Draw a few lines and points
4. Click invert button
5. Bug - lines change color

## Expected behavior
- invert button works with multi-frame viewer
- invert button keeps line color consistent
- invert button works in Safari

## Additional context
- [PFE SVGImage invert](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/svg-image.jsx)

mcbouslog:
## Package
- lib-classifier

- closes #3328 

## Describe your changes
- add invert status from subject viewer store to MultiFrameViewerContainer
- add SVGImage component with svg filter to invert colors, image applies/defines filter on svg image per invert property
- ^ similar to [PFE SVGImage](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/svg-image.jsx)

## How to Review
  - (multi-frame subject, transcription lines) - [⚠️ **CAUTION** ⚠️ live production project] - https://local.zooniverse.org:8080/?project=humphrydavy%2Fdavy-notebooks-project&env=production&demo=true&workflow=18575
  - (single image frame, and text, subject) - https://local.zooniverse.org:8080/?project=1952&workflow=3602&demo=true
  - if drawing/transcription task, draw a few lines, noting colro
  - click Invert button in ImageToolbar
  - image color should invert, line colors should remain the same
  - click Invert button in ImageToolbar
  - image color should revert to original/normal, line colors should remain the same
  - http://localhost:6006/?path=/story/image-toolbar-image-toolbar--with-multi-frame-viewer&args=dark:true&globals=locale:en;locales.en:English;locales.test:Test+Language
- Are there plans for follow up PR’s to further fix this bug or develop this feature? the MultiFrameViewer will be further refactored as per #3210 

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGImage/SVGImage.js
```diff
@@ -0,0 +1,54 @@
...
+  return (
+    <SubjectImage
+      aria-label={`Subject ${subjectID}`}
+      dragMove={move ? onDrag : () => null}
+      height={naturalHeight}
+      role='img'
+      style={invert ? { filter: 'url("#svg-invert-filter")' } : {}}
+      width={naturalWidth}
+      xlinkHref={src}
+    />
```
eatyourgreens, line 52: Wrapping this in either `<svg filter='invert(1)' />` or `<g filter='invert(1)' />` might be simpler than appending an [SVG `<filter>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/filter) to the document body.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGImage/SVGImage.js
```diff
@@ -0,0 +1,54 @@
...
+
+  const SubjectImage = move ? DraggableImage : 'image'
+
+  return (
+    <SubjectImage
+      aria-label={`Subject ${subjectID}`}
+      dragMove={move ? onDrag : () => null}
+      height={naturalHeight}
+      role='img'
+      style={invert ? { filter: 'url("#svg-invert-filter")' } : {}}
```
eatyourgreens, line 49:
```suggestion
      filter={invert ? 'url("#svg-invert-filter")' : undefined}
```
I think this is the correct syntax if you want this component to render SVG, rather than SVG/HTML. It depends whether you envisage this component ever being used in a pure SVG document.

eatyourgreens (commented):
Depending on whether it works in Safari, you could wrap the `<image>` element in a filtered element:
```svg
<g filter='invert(1)'>
  <image xlink:href='…' />
</g>
```

or even this (d3/visx does this):

```svg
<svg filter="invert(1)" xmlns="http://www.w3.org/2000/svg">
  <image " xlink:href="…"></image>
</svg>
```

It needs testing in Safari, but I think it's easier to read.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGImage/SVGImage.js
eatyourgreens: Never mind: I've tried this in Safari and the filter on the embedded `<svg>`  doesn't work. For some reason, it only works on the root element.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGImage/SVGImage.js
```diff
@@ -0,0 +1,54 @@
...
+  src,
+  subjectID
+}) {
+  useEffect(function onInvert () {
+    if (invert) {
+      if (!document.getElementById('svg-invert-filter')) {
+        document.body.insertAdjacentHTML('afterbegin', INVERT)
+      }
+    }
+  }, [invert])
```
eatyourgreens, line 39:
```suggestion
if (invert) {
  if (!document.getElementById('svg-invert-filter')) {
    document.body.insertAdjacentHTML('afterbegin', INVERT)
  }
}
```

The effect hook here [adds an extra render](https://beta.reactjs.org/apis/usestate#storing-information-from-previous-renders) that you probably don't need. Do you really need to check `invert` every time the button is toggled? It seems, to me, like you should add the SVG filter to the document on the first render, then skip this block on subsequent renders if the filter already exists.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGImage/SVGImage.js
eatyourgreens: I've tested this in Safari and it works.

eatyourgreens: I've tried this out in Safari, and it works.

eatyourgreens:
Looks good on Davy Notebooks. 👍 


https://user-images.githubusercontent.com/59547/175490623-84b967e6-cf3b-46c9-872d-18840fede663.mov

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewerContainer.spec.js
```diff
@@ -2,13 +2,14 @@ import { mount } from 'enzyme'
 import { Provider } from 'mobx-react'
 import sinon from 'sinon'
 import React from 'react'
+import asyncStates from '@zooniverse/async-states'
 
 import mockStore from '@test/mockStore'
-import { DraggableImage, MultiFrameViewerContainer } from './MultiFrameViewerContainer'
-import FrameCarousel from './FrameCarousel'
-import SingleImageViewer from '../SingleImageViewer/SingleImageViewer'
 import TranscriptionLineTool from '../../../../../../plugins/drawingTools/experimental/models/tools/TranscriptionLineTool/TranscriptionLineTool'
```
eatyourgreens, line 8:
```suggestion
import TranscriptionLineTool from '@plugins/drawingTools/experimental/models/tools/TranscriptionLineTool/TranscriptionLineTool'
```
The aliased path might make this easier to read.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGImage/SVGImage.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    render(
+      <SVGImage
+        naturalHeight={100}
+        naturalWidth={100}
+        src='https://some.domain/image.jpg'
+        subjectID='1234'
+      />
+    )
+
+    const image = screen.getByRole('img')
```
eatyourgreens, line 18:
```suggestion
    const image = screen.getByRole('img', { name: 'Subject 1234' })
```

I've been trying to get into the habit of querying by role and name, to check that an element is appropriately named in the accessibility tree.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGImage/SVGImage.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    expect(image).to.exist()
+    expect(image.getAttribute('xlink:href')).to.equal('https://some.domain/image.jpg')
+    expect(image.getAttribute('aria-label')).to.equal('Subject 1234')
```
eatyourgreens, line 21: If you add a name to the `getByRole` query then you can probably remove this check.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGImage/SVGImage.spec.js
```diff
@@ -0,0 +1,46 @@
...
+          src='https://some.domain/image.jpg'
+          subjectID='1234'
+        />
+      )
+
+      const filter = document.querySelector('filter')
+      expect(filter).to.exist()
+      expect(filter.getAttribute('id')).to.equal('svg-invert-filter')
+
+      const image = screen.getByRole('img')
```
eatyourgreens, line 41:
```suggestion
      const image = screen.getByRole('img', { name: 'Subject 1234' })
```

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/MultiFrameViewer/MultiFrameViewer.stories.js
```diff
@@ -57,151 +55,85 @@ const subjectWithMore = Factory.build('subject', {
...
-export const PanZoomMore = () => {
-  return (
-    <ViewerContext mode="light" subject={subjectWithMore} theme={zooTheme}>
-      <Box direction="row">
-        <DecoratedMultiFrameViewerContainer
-          loadingState={asyncStates.success}
-          enableInteractionLayer={false}
-          subject={subjectWithMore}
-        />
-        <ImageToolbar />
```
eatyourgreens, line 178: Isn't this needed in order to use the invert button in the storybook?

eatyourgreens (approved):
Looks good, and fixes the invert button (I tested Davy Notebooks in a production build of the project app.)

I have two big comments about the storybook. These should be addressed either here or in a new PR, before the storybook is deployed.
1. The invert button story doesn't have an on/off state when I click the button, even though it's a toggle control.
2. The image toolbar has been removed from the viewer stories, so we don't have any stories that demonstrate the invert button. Here are live stories that use the image toolbar.
https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/index.html?path=/story/subject-viewers-singleimageviewer--with-zoom-controls&globals=locale:en;locales.en:English;locales.test:Test+Language

https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/index.html?path=/story/subject-viewers-multiframeviewer--pan-zoom-more&globals=locale:en;locales.en:English;locales.test:Test+Language

mcbouslog:
> 

Thanks Jim!

1. - is there an example of a story that updates an arg based on a different arg function? I'm not sure how to implement without wrapping the component in the store and removing `active` as an arg. Noting that while they should also function as noted, the annotation and move button stories also don't toggle `active` onClick, but relatedly...
2. - there are two new stories:
  a. http://localhost:6006/?path=/story/image-toolbar-image-toolbar--with-single-image-viewer&globals=locale:en;locales.en:English;locales.test:Test+Language
  b. http://localhost:6006/?path=/story/image-toolbar-image-toolbar--with-multi-frame-viewer&globals=locale:en;locales.en:English;locales.test:Test+Language
  that have a single image and multi-frame image with the ImageToolbar component, including invert, and sets `active` state for the annotation, move, and invert buttons as expected

eatyourgreens:
Sorry, I missed those new stories when I looked through the storybook. They look good. Is the idea to add image toolbar stories for each new subject viewer? If you could do something like `import * as subjectViewers from 'components/SubjectViewers'` then the image toolbar stories could loop through all available viewers, generating a story for each one.

You're right that all the toggle button stories are broken. That's work for a separate PR, but they should each have an on/off state.

eatyourgreens:
Not related to this PR, but the focus state for the invert button obscures the on/off state. When the button has focus, you can’t really tell if it’s on or off. That would be fixed by using a focus ring on the focused button, rather than changing the fill colour.

mcbouslog:
> you could do something like `import * as subjectViewers from 'components/SubjectViewers'` then the image toolbar stories could loop through all available viewers, generating a story for each one.

I like this idea! I think could be included as part of the work towards #3093.

mcbouslog:
> Not related to this PR, but the focus state for the invert button obscures the on/off state. When the button has focus, you can’t really tell if it’s on or off. That would be fixed by using a focus ring on the focused button, rather than changing the fill colour.

I added a related card in https://github.com/zooniverse/front-end-monorepo/projects/17#card-83272996 so I don't forget to revisit this, nice catch and solution, thanks!

PR #3351, opened 2022-06-28
QuickTalk: Improve UI

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Affects: QuickTalk experimental feature

This PR attempts to improve the UI for the QuickTalk component.

The goals of this PR:
- Improve messaging for (expanded) QuickTalk panel
  - [x] If a user isn't logged in, a message now appears prompting them login if they want to post a comment. (Previously, there was nothing indicating what users should do)
  - [x] If a Subject has no comments, a message now appears informing volunteers. (Previously, a Subject with no comments just left a blank QuickTalk panel.)
      image
      _Above: the messages that appear when a Subject has no comments, and user isn't logged in._
  - [x] The "Post Comment" form now has additional text for when the form is ready for use ("Leave a note about this subject") and when the post is successful ("Comment posted!"). These two new messages come before/after the "Posting comment..." message, and as always, an error message overrides this.
    - Accessibility: status messages are now marked with role="status"
- [x] Solve the issue of the (collapsed) QuickTalk button not being visually distinguishable on 
    image
    _Above: the QuickTalk button (on lower right) merges visually with the content a large Task area._
  - [x] OPTION 1: provide the QuickTalk button some background to make it a "solid visual" item. (UPDATE: see [below](https://github.com/zooniverse/front-end-monorepo/pull/3351#issuecomment-1169156164))
  - OPTION 2: make QuickTalk a **tab,** alongside Task and Tutorial in the TaskArea. (might spin off this idea into its own experimental PR) (UPDATE: this is a separate PR)

Not addressed in this PR (probably):
- Advanced keyboard navigation.
- Paging, for Subjects with 10+ comments.

⚠️ TO INVESTIGATE:
- On `lib-classifier`, when a user logs out, the "Post Comment" textbox doesn't disappear immediately.
  - Is there a similar issue on `app-project`?
  - User CAN'T post while logged out though, so that's safe. ("User is not logged in" error message will appear, despite the textbox being there.)

### Status

~~WIP~~

Ready for review.
- [Update 1](https://github.com/zooniverse/front-end-monorepo/pull/3351#issuecomment-1169156164): QT button now has a solid visual design
- [Update 2](https://github.com/zooniverse/front-end-monorepo/pull/3351#issuecomment-1169228936): layout and keyboard input improved
- [Update 3](https://github.com/zooniverse/front-end-monorepo/pull/3351#issuecomment-1169261656): testing recommendations

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
+++ b/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
@@ -105,7 +105,12 @@ function QuickTalk ({
             plain
           />
         </Box>
-        <Box overflow='auto' tabIndex='0'>
+        <Box
+          aria-label={t('QuickTalk.panelContent')}
+          overflow={{ vertical: 'auto', horizontal: 'hidden' }}
+          role='group'
+          tabIndex='0'
+        >
           <Box flex={false} pad={{ bottom: 'small', left: 'small', right: 'small' }}>
             {comments.length > 0 && (
               <UnorderedList as='ul' flex={false}>
--- a/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
+++ b/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import { withResponsiveContext } from '@zooniverse/react-components'
-import { Anchor, Box, Button } from 'grommet'
+import { Anchor, Box, Button, Paragraph } from 'grommet'
 import { Chat, Close } from 'grommet-icons'
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
@@ -117,6 +117,14 @@ function QuickTalk ({
               postCommentStatus={postCommentStatus}
               postCommentStatusMessage={postCommentStatusMessage}
             />
+          )}
+          {!userId && (
+            <Box
+              background={{ dark: 'dark-1', light: 'light-1' }}
+              margin={{ horizontal: 'none', 'vertical': 'xsmall' }}
+            >
+              <Paragraph textAlign="center">{t('QuickTalk.loginToPost')}</Paragraph>
+            </Box>
           )}
         </Box>
       </Box>
--- a/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
+++ b/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import { withResponsiveContext } from '@zooniverse/react-components'
-import { Anchor, Box, Button, Keyboard, Paragraph } from 'grommet'
-import { Chat, Close } from 'grommet-icons'
+import { Anchor, Box, Button, Heading, Keyboard, Paragraph } from 'grommet'
+import { Chat, Close, FormNextLink } from 'grommet-icons'
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
 import asyncStates from '@zooniverse/async-states'
@@ -88,15 +88,11 @@ function QuickTalk ({
           direction='row'
           flex={false}
           justify='between'
-          pad='small'
+          pad={{ vertical: 'none', horizontal: 'small'}}
         >
-          <Anchor
-            a11yTitle={t('QuickTalk.aria.goToTalk')}
-            label={t('QuickTalk.headerLink')}
-            href={subject.talkURL}
-            target='_blank'
-            icon={<Chat />}
-          />
+          <Heading level='4' margin={{ top: 'small', bottom: 'none' }} pad='none'>
+            {t('QuickTalk.aria.panelHeading')}
+          </Heading>
           <Button
             a11yTitle={t('QuickTalk.aria.closeButton')}
             autoFocus={true}
@@ -106,7 +102,7 @@ function QuickTalk ({
           />
         </Box>
         <Box
-          aria-label={t('QuickTalk.panelContent')}
+          aria-label={t('QuickTalk.aria.panelContent')}
           overflow={{ vertical: 'auto', horizontal: 'hidden' }}
           role='group'
           tabIndex='0'
@@ -164,6 +160,17 @@ function QuickTalk ({
               </Paragraph>
             </Box>
           )}
+          <Anchor
+            alignSelf='center'
+            label={t('QuickTalk.aria.goToTalk')}
+            href={subject.talkURL}
+            icon={<FormNextLink size='small' />}
+            margin='xsmall'
+            target='_blank'
+            rel='nofollow noopener noreferrer'
+            reverse={true}
+            size='xsmall'
+          />
         </Box>
       </QTPanel>
     </Keyboard>
--- a/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
+++ b/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useRef } from 'react'
 import { withResponsiveContext } from '@zooniverse/react-components'
 import { Anchor, Box, Button, Heading, Keyboard, Paragraph } from 'grommet'
 import { Chat, Close, FormNextLink } from 'grommet-icons'
@@ -48,9 +48,22 @@ function QuickTalk ({
   fixedPosition = true,
   showBadge = true // HACK: Button.badge crashes tests AND storybook for an undetermined reason. // TODO: debug
 }) {
+  const panelContent = useRef()
   const { t } = useTranslation('components')
   // TODO: figure out if/how the QuickTalk component should/could be displayed on mobile
   // if (screenSize === 'small') return null
+
+  useEffect(function scrollToLatestComment () {
+    if (postCommentStatus === asyncStates.success) {
+      // Comments appear at the bottom, so scroll to the bottom
+      const newYCoord = panelContent.current?.scrollHeight
+      panelContent.current?.scrollTo({
+        top: newYCoord,
+        left: 0,
+        behavior: 'smooth'
+      })
+    }
+  }, [ postCommentStatus ])
 
   if (!subject) return null
 
@@ -102,6 +115,7 @@ function QuickTalk ({
           />
         </Box>
         <Box
+          ref={panelContent}
           aria-label={t('QuickTalk.aria.panelContent')}
           overflow={{ vertical: 'auto', horizontal: 'hidden' }}
           role='group'
--- a/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
+++ b/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import { withResponsiveContext } from '@zooniverse/react-components'
-import { Anchor, Box, Button, Paragraph } from 'grommet'
+import { Anchor, Box, Button, Keyboard, Paragraph } from 'grommet'
 import { Chat, Close } from 'grommet-icons'
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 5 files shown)


shaunanoordin:
## PR Update

- (Collapsed) QuickTalk button is now placed in a _visually solid and slightly elevated_ container
    image
    _Above: QT button no longer "smushes" into background visuals._
    image
    _Above: QT button with notifications_

shaunanoordin:
## PR Update

- Keyboard navigation: the content portion of the QT panel (where all the comments are) can now receive keyboard focus, allowing users to scroll up/down using up/down arrow keys.
  - ⚠️ The content portion does NOT automatically receive focus when the QT panel opens, however. Instead, the Close button gets focus, because that's the only thing on the panel that can always receive autoFocus. (PostForm doesn't count because a user might not be logged in.)
  - I'm having trouble with this one, so I may leave it for another PR.
- Keyboard navigation: the Esc key now closes the expanded QT panel. Woohoo!
  - ⚠️ ...unless the user has the PostForm's text area in focus. Hmm. Might need to address this in another PR, since I'm not 100% if it's a good idea to override the 'Esc' key when the user is typing text. (It might matter if e.g. the user is pressing Esc to disable autocomplete but instead accidentally closes the QuickTalk panel.)
- Layout improvement: the Post Form (if you're logged in, or the "please login" message box if you aren't) is now always visible. Previously, volunteers would have had to scroll all the way down to the bottom of the content portion to find the "submit comment" form.
    image
    _Above: PostForm is now always visible on the QT panel. QT content has focus, allowing users to scroll up/down using the keyboard_

### Status

Ready for review!

shaunanoordin:
## PR Update

### Testing Recommendations

**Testing Setup**

You'll need to test this PR on a project with QuickTalk enabled.

- Example on `lib-classifier`: a WF with lots of Ultramen waiting for your comments - https://local.zooniverse.org:8080/?env=staging&project=darkeshard%2Ftest-project-2022&workflow=3581 
- Example on `lib-classifier`: a WF with a "tall" Task area, to test the visibility of the QT button - https://local.zooniverse.org:8080/?env=staging&project=darkeshard%2Ftest-project-2022&workflow=3597
- Example on `app-project`: https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022/classify/workflow/3581?env=staging&demo=true

**Testing Behaviour**

This is primarily a UI/UX update, so much of this testing involves asking the question: "Does the QuickTalk component make sense?"

- Main Goal: QuickTalk component should allow users to interact with the Talk/discussion of a Subject on the Classifier
- When collapsed, the QuickTalk _button_ should be distinguishable, easy to access, and not block/be blocked by other elements.
  - The QT button should expand into the QT panel when clicked on.
  - You should be able to access & open the QT button via keyboard.
  - The QT button should have numbers indicating the number of comments the Subject has.
- when expanded, the QuickTalk _panel_ should display comments from the Subject's Talk data.
  - ⚠️ Current Limitation: it only shows the first 10 comments at max.
  - Clicking on the "Subject Discussion" title will link the user to the Talk page. (EDIT: is this visually obvious?)
  - You should be able to scroll the comments via (1) mouse scroll, (2) scroll bar, or (3) keyboard up/down _IF the content area has focus_
  - If the user is logged in, the "Post Comment" form should be visible.
  - If the user isn't logged in, there should be a prompt suggesting they log in.
  - The QT panel should collapse into the QT button when either (1) the close X button is clicked, or (2) the Esc key is pressed
    - ⚠️ Current Limitation: the Esc key doesn't close the QT panel if the user is typing a comment (i.e. has focus on the text box)

### Design Notes

@seanmiller26 I'm tagging you in, in case you have any design feedback on this. Sorry, I should have tagged you in the first design iteration!

Here are some quick notes:

- The QuickTalk component is built as an experiment to add Talk discussions [(example)](https://frontend.preview.zooniverse.org/projects/darkeshard/test-project-2022/talk/subjects/136478) into the Classifier
- This PR is the second design iteration. If you want to see how it looks like currently, try [this](https://frontend.preview.zooniverse.org/projects/darkeshard/test-project-2022/classify/workflow/3581?env=staging)

Here are some visual states of the QuickTalk component:

image

_Above: QuickTalk button (collapsed form) with 9+ comments for this Subject_

image

_Above: QuickTalk panel (expanded form), with 10 comments, and the user logged in_

image

_Above: QuickTalk button (collapsed form) with 0 comments for this Subject_

image

_Above: QuickTalk panel (expanded form) with 0 comments, and the user NOT logged in_

Slightly longer note:
- There is an idea to replace the UI design altogether, from "button fixed to the bottom-right of the browser window" to "additional Tab section in the Task Area, next to the "Task" and "Tutorial" tabs.
- Rationale: because the TaskArea-Tab idea is cleaner than a fixed element floating around the screen. (e.g. the QT button isn't flush to the right of the screen because it'll interfere with/be interfered by the "Switch to Dark Mode" button)
- I plan to spin up an experiment for this, but if you have thoughts or better ideas, please let me know.

eatyourgreens (commented):
A couple of quick comments about tabbing through the box, particularly if you're listening to a screenreader and can't see the `QuickTalk` box.

Is the popup meant to be modal? At the moment, it looks like you can tab out of the box to the underlying page.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
```diff
@@ -52,75 +57,111 @@ function QuickTalk ({
...
+          />
+          <Button
+            a11yTitle={t('QuickTalk.aria.closeButton')}
+            autoFocus={true}
+            icon={<Close size='small' />}
+            onClick={() => setExpand(false)}
+            plain
+          />
+        </Box>
+        <Box overflow='auto' tabIndex='0'>
```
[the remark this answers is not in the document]
eatyourgreens: This is really useful: [Scrollable overflow containers](https://www.a11yproject.com/posts/how-to-use-the-tabindex-attribute/#scrollable-overflow-containers).

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
```diff
@@ -52,75 +57,111 @@ function QuickTalk ({
...
-              const author = authors[comment.user_id]
-              const roles = authorRoles[comment.user_id]
+        <Box
+          direction='row'
+          flex={false}
+          justify='between'
+          pad='small'
+        >
+          <Anchor
+            a11yTitle={t('QuickTalk.aria.goToTalk')}
```
eatyourgreens, line 94: Why is this here when the link already has a text label?

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
```diff
@@ -52,75 +57,111 @@ function QuickTalk ({
...
+            label={t('QuickTalk.headerLink')}
```
eatyourgreens, line 95: This looks like a modal heading, saying 'Subject discussion', when the box opens. I'd recommend removing the link and making that the actual title bar for the popup. Having the title bar of the popup be a link to another page is very confusing UX, and not something I've seen before. I don't think volunteers will understand or expect that they can click this text in order to go to Talk.

eatyourgreens (commented):
A couple of other comments. The title bar of the popup is a link, which is confusing when I'm using this in practice.

eatyourgreens:
Logging out looks good to me. The comments box vanishes when I log out on the test workflow.


Uploading Screen Recording 2022-07-01 at 09.39.26.mov…

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
```diff
@@ -52,75 +57,111 @@ function QuickTalk ({
...
+          {!userId && (
+            <Box
+              background={{ dark: 'dark-1', light: 'light-1' }}
+              margin={{ horizontal: 'none', 'vertical': 'xsmall' }}
+              pad='xsmall'
+            >
+              <Paragraph textAlign='center'>
+                {t('QuickTalk.loginToPost')}
+              </Paragraph>
+            </Box>
```
eatyourgreens, line 160:
```suggestion
            <Paragraph
              background={{ dark: 'dark-1', light: 'light-1' }}
              margin={{ horizontal: 'none', 'vertical': 'xsmall' }}
              pad='xsmall'
              textAlign='center'
            >
              {t('QuickTalk.loginToPost')}
            </Paragraph>
```

Try this out before committing it, but I think you can use those styling properties directly on the paragraph.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
```diff
@@ -52,75 +57,111 @@ function QuickTalk ({
...
             />
           )}
+          {!userId && (
+            <Box
+              background={{ dark: 'dark-1', light: 'light-1' }}
+              margin={{ horizontal: 'none', 'vertical': 'xsmall' }}
+              pad='xsmall'
+            >
+              <Paragraph textAlign='center'>
+                {t('QuickTalk.loginToPost')}
```
eatyourgreens, line 158: When I first saw this, I clicked on the word 'login' in order to login. A link to the login form might be useful, but I realise that we can't add NextJS links to the classifier library.

eatyourgreens (commented):
Changes generally look good, but there's a breaking change on this branch. When I post a comment, my new comment does not scroll into view.


https://user-images.githubusercontent.com/59547/176861495-f2f84549-a04c-4893-bc85-d39ae0bc1288.mov

I'm going to hold off on approval because I'm not sure if that needs to be fixed.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/components/PostForm.js
```diff
@@ -49,18 +44,22 @@ function PostForm ({
...
   return (
     <Box
       background={{ dark: 'dark-1', light: 'light-1' }}
       flex={false}
-      pad='small'
+      pad='xsmall'
     >
       {statusText && (
-        <Text>{statusText}</Text>
+        <Text role='status'>{statusText}</Text>
```
eatyourgreens, line 62:
```suggestion
        <Text role='status' aria-live="polite">{statusText}</Text>
```
[MDN recommends](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#roles_with_implicit_live_region_attributes) explicitly setting politeness for status roles.

eatyourgreens: Having said that, [these screenreader tests](https://www.digitala11y.com/status-role/) say the implicit role is well-supported so maybe that's needed for older screen readers? 🤷‍♂️

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
eatyourgreens: If you remove the link, and make this a heading, then autofocus for the close button also makes more sense.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/components/PostForm.js
shaunanoordin: Oh that's definitely something I missed

The resources I was reading on [W3](https://www.w3.org/TR/wai-aria-1.1/#status) and the [_other_ MDN page](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role) both said that `role=status` elements have `aria-live=polite`

But the [MDN page you shared](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#roles_with_implicit_live_region_attributes) does mention the explicit aria-live=polite is to "To maximize compatibility", so you're probably right that it's needed for older screen readers.

I'll commit the suggestion, thanks Jim 👍

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
shaunanoordin: I remember running into this problem earlier - I couldn't get the Paragraph to accept certain style props.

With the original `<Box styles='...'><Paragraph ... /></Box>`:
image

With the suggested `<Paragraph styles='...' />`:
image

Anyway, check this out, I replaced my login bits with:
```
<Box pad='large' background='red'>
  This is a Box with pad='large' background='red'
</Box>
<Paragraph pad='large' background='red'>
  This is a Paragraph with the same props
</Paragraph>
```

And I got this:
image

I thought a Paragraph was supposed to inherit all the basic props of a Box?? Or was this an earlier version of Grommet that I was using? (On Classrooms?) In any case, thank you for the suggestion.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
shaunanoordin: The separation of user auth from app-project and lib-classifier is haunting me yet again 😞

shaunanoordin:
Thanks Jim! OK, here are my upcoming actions in response to the feedback:

- [ ] **[Blocking]** Newest comment should always receive focus when successfully posted.
  - The cause is simple: in the original design, when a user posts a comment, the comments (content) panel _has its view scrolled to the bottom already,_ (because that's where the post form is) so the new comment always pops into view.
  - Plan: programmatically scroll the scrollable div to the bottom upon successful post?
  - Before I do that, I need to properly digest [the article you shared.](https://www.a11yproject.com/posts/how-to-use-the-tabindex-attribute/#scrollable-overflow-containers)
- [ ]  Rework Header.
  - "Subject Discussion" header shouldn't be a link to the discussion page.
  - Plan: have a discrete `<a>` link that says  "go to Subject Discussion Talk page" or similar. Make the "Subject Discussion" header just text or just remove it altogether. Maybe a header bit explaining "this is the Subject Discussion component!" isn't necessary.
- [ ] Rework initial keyboard focus ❓ 
  - Plan: ???
  - I might need to punt this to another PR because I've been having issues getting keyboard focus to work the way I want it.
  - Again, the alternative Tab design I'm cooking up may make this moot.

Not addressing in this PR:
- QuickTalk UI is meant to be a non-modal dialog in this design.
- I'm already working on an alternate design version where QuickTalk is the third tab in the TaskArea (aka the Tab design), so that may make the modal option moot.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
```diff
@@ -52,75 +57,111 @@ function QuickTalk ({
...
+          <Anchor
+            a11yTitle={t('QuickTalk.aria.goToTalk')}
+            label={t('QuickTalk.headerLink')}
+            href={subject.talkURL}
+            target='_blank'
+            icon={<Chat />}
+          />
+          <Button
+            a11yTitle={t('QuickTalk.aria.closeButton')}
+            autoFocus={true}
```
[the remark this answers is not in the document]
shaunanoordin: Here's the meandering & confused tale of how I got here:

- I designed the QuickTalk as a non-modal dialog. (Why I didn't go for modal is a separate tale.)
- I tried adding `role=dialog` thinking _"ahhah, it's a separate content from the rest of the UI, so role=dialog should help people understand this component!"_
- You helpfully [pointed out](https://github.com/zooniverse/front-end-monorepo/pull/3048#pullrequestreview-971240072) that a dialog actually needs additional behaviour, namely (1) the dialog should receive focus when it opens, and (2) that focus should go to the [first comment](https://github.com/zooniverse/front-end-monorepo/pull/3048#discussion_r871253530) (or at least top) of the dialog 
- I couldn't focus on the first comment, since there may be scenarios where the panel opens with 0 comments.
- I tried making the "Subject Discussions" header/link the first focused item, but I couldn't get the element.to.focus() since the element wasn't rendered when the "open dialog" function triggered. (NOTE: I will have to try this again now that I know a bit more about [callback refs](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs))
- I found that Grommet's `<Anchor autoFocus={true} />` doesn't work, so that's out. (Variations such as `autofocus={true}`, `autofocus`, and `autoFocus` also failed to register the intended property on the Grommet element.)
- The only element/Grommet component I could give focus to using the autofocus prop is a Button, and so...

...that's how & why the Close button got focus when the dialog pops up. As usual, there's probably a simpler solution to this that I'm overlooking, but man, this so far has been a trip.

- Side note 1: `<Box overflow='auto' tabIndex='0' autofocus={true} / autoFocus={true} / autofocus / autoFocus>`, for the content box which contains all the comments, also failed to gain focus. I really wanted this to get focus, since this is the actual _content_ of the dialog.
- Side note 2: Curiously, I see `<Box autoFocus />` used in SlideTutorial, but I think what's getting autofocus when the SlideTutorial modal appears is a hidden `<a>`nchor, not the intended Box. 🤷 Side _side_ note: I think the SlideTutorial is missing a `<Box tabindex=0>` that should allow a user to scroll through vertically-tall tutorial content.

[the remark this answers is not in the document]
eatyourgreens: TBH the close button should be focused, since it’s the first focusable element in the popup. 

I think the ARIA advice for dialogs mentions that too.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
eatyourgreens: I thought Paragraph accepted the same styling props as Box, but obviously not.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
eatyourgreens: Yeah. This could be a button that fires a callback but it’s fine as it is. 

NB. In the project app, I think you can just append `?login=true` to the current page location. Obviously that won’t work in the dev classifier.

eatyourgreens:
[`Element.scrollIntoView`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) might be useful. 

Alternatively, there’s the design pattern used by Twitter where the comment box is at the top and comments are listed most recent first. That makes the most recent comment obvious, but reading a conversation in reverse order can be hard. Assuming that people read the previous comments before adding their own. Anecdotally, that didn’t really happen with the comment box on OWD subjects. People added hashtags and maybe one or two words as comments.

eatyourgreens:
NB. the first focusable element in a popup should autofocus. The reason for that is that screen reader users will read forward from the focused element. If the content box is focused, they may not even realise that the close button is there. 

Autofocus in the tutorial is very confusing for blind users, based on the feedback we had from one totally blind volunteer who struggled to use both the tutorial and the field guide. I’d be cautious about looking to those as examples.

packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalk.js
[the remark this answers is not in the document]
eatyourgreens: Big takeaway from that is that whenever you add a `tabindex`, you should also add an accessible name (either `aria-label` or `aria-labelledby`.)

shaunanoordin:
## PR Update

Feedback items addressed!

- 🎉 Newest Comment now scrolls into view!
- Header rework: "Subject Discussion" is now just a heading and not a link ; the link to the Subject Discussion has been moved (see image below)
- Keyboard focus: the X Close button is now the first interactive element in the QT, and has focus when the panel expands. (i.e. no behaviour change, but the layout & context surrounding the button has.)

image
_Above: the latest version of the expanded QuickTalk panel. The link to the Talk page has now been moved to after the Post form, and given better context, i.e. as an alternative option for viewing the Subject Discussion._

### Dev Notes

> Alternatively, there’s the design pattern used by Twitter where the comment box is at the top and comments are listed most recent first. 
- As Twitter user I am embarrassed that I didn't crib this.
- Funny story, "sorting by most recent first" was actually one of my initial prototypes, because Subject Discussion on Talk DOES sort by most recent first. [(Example)](https://www.zooniverse.org/projects/darkeshard/test-project-2022/talk/subjects/136475?env=staging) ...but I thought it was weird, because the Post form is at the _bottom._ 🤨 
- So I declared _"Ahhah! I shall sort it oldest first, then keep the Post form at the bottom, and surely this won't bite me in the ass when there are more than 10 comments and I haven't addressed paging yet!"_
- ...and this is why I should always ping @seanmiller26 before I get creative with the UI.

> NB. the first focusable element in a popup should autofocus. The reason for that is that screen reader users will read forward from the focused element. If the content box is focused, they may not even realise that the close button is there.
- This is very useful context, I did not think of the "forward only" view that screen readers would have.

### Status

Ready for review!

shaunanoordin:
Thanks Jim! 👍 

> Looks much better. One small comment after using it from the keyboard: where does focus go after I post a comment then close the box with Esc? It should probably go back to the Quick Talk button, but that isn't needed for this PR.

I just tested this - on macOS + Chrome, when you close the expanded QuickTalk panel with the Esc key, the focus reverts to the `body` element (same as having a component disappear on you), so that's not ideal. The next Tab key press does navigate to the QuickTalk button (instead of resetting to the start of the page), curiously.

Should be a straightforward fix in the next UI PR 👌

PR #3372, opened 2022-06-30
UI Fix: allow `<Tabs>` to have more than 2 `<Tab>`s in a row

shaunanoordin:
## PR Overview

This PR fixes an issue where the `<Tabs>` component incorrectly assigns all its children to have `flex-bases: 50%`, NO MATTER the number of child `<Tab>`s.

Expected: `<Tabs>` should display all `<Tab>`s in a row, if there's space.
Screenshot 2022-06-30 at 18 10 47

Actual: `<Tabs>` only displays up to two `<Tab>`s in a row.
Screenshot 2022-06-30 at 18 11 11

### Dev Notes

Some terminology notes:
```
<Tabs>  // This is the parent/container
  <Tab title='cat'>Meow!</Tab>  // This is the first child, with a Tab HEADER saying "cat", and a Tab BODY (or Tab Panel) saying "Meow!" 
  <Tab title='dog'>Bark!</Tab>  // ...second child
  <Tab title='shaun'>Fart!</Tab>  // ...etc
<Tabs>
```

So here's what's causing the problem:
- `<Tabs>` tries to assign each child `<Tab>` an equal flex-basis ("default size") value, i.e. `flex: 1 1 ${100 / props.children.length}%;`
- e.g. if there are 3 children, each should get `flex: 1 1 33.333%`
- Problem is, the `props.children` doesn't actually represent the number of `<Tab>` children.
- Instead, `props.children` always has two elements: the _collection_ of Tab HEADERS, and a single Tab BODY (of the first Tab, I think.)

So, to correctly calculate the number of child `<Tab>`s, we need to look at the number of children _in the collection of Tab Headers._ i.e., `props.children[0].props.children`

### How to Review

- Boot up `lib-react-components` on `master` and run `yarn storybook`
- Take a look at [Tabs > Light Theme](http://localhost:6007/?path=/story/components-tabs--light-theme-default)
- On `master`, there should be only two Tab(s) on that story - Foo and Bar - so it looks fine.
- Now, try editing the story so that it has more than two Tab(s), e.g. Foo, Bar, Zap, Kaboom
- You should see the problem: there are only up to two Tab(s) in a row.
- Now, switch to this `fix-tabs-flex-basis` branch, and you should see the problem solved.

### Checklist for Reviewer: Bug Fix

- [ ] The bug is fixed
- ~~Unit tests are added or updated~~

### Status

Ready for review! This will affect a future QuickTalk UI improvement PR. I'll link that PR once it's up.

### DIFF
--- a/packages/lib-react-components/src/Tabs/theme.js
+++ b/packages/lib-react-components/src/Tabs/theme.js
@@ -28,10 +28,14 @@ const theme = {
       const borderColor = getBorderColor(props)
       const activeTabHeaderColor = getActiveTabHeaderColor(props)
       const hoverBackgroundColor = getHoverBackground(props)
+      const tabChildren = props.children?.[0]?.props?.children
+      const flexBasis = tabChildren
+        ? `${100 / tabChildren.length}%`
+        : 'auto'
       return css`
         button[role="tab"] {
           color: ${tabHeaderColor};
-          flex: 1 1 ${100 / props.children.length}%;
+          flex: 1 1 ${flexBasis};
           > div {
             border-bottom: 1px solid ${borderColor};
             border-right: 1px solid ${borderColor};
@@ -78,7 +82,7 @@ const theme = {
       }
     },
     panel: {
-      extend: props => { 
+      extend: props => {
         const backgroundColor = getBackgroundColor(props)
         const borderColor = getBorderColor(props)
         return css`
--- a/packages/lib-react-components/src/Tabs/Tabs.stories.js
+++ b/packages/lib-react-components/src/Tabs/Tabs.stories.js
@@ -30,8 +30,10 @@ export const LightThemeDefault = () => (
     themeMode="light"
   >
     <Tabs>
-      <Tab title="foo">Foo</Tab>
-      <Tab title="bar">Bar</Tab>
+      <Tab title="one">One</Tab>
+      <Tab title="two">Two</Tab>
+      <Tab title="three">Three</Tab>
+      <Tab title="four">Four</Tab>
     </Tabs>
   </Grommet>
 );


packages/lib-react-components/src/Tabs/theme.js
```diff
@@ -28,10 +28,14 @@ const theme = {
       const borderColor = getBorderColor(props)
       const activeTabHeaderColor = getActiveTabHeaderColor(props)
       const hoverBackgroundColor = getHoverBackground(props)
+      const tabChildren = props.children?.[0]?.props?.children
+      const flexBasis = tabChildren
+        ? `${100 / tabChildren.length}%`
```
eatyourgreens, line 33: I'm confused. Why doesn't `props.children.length` work here?

eatyourgreens: This looks a bit weird on phone screens, where flexbox moves the fourth tab down under the first three.

The Tabs component with four tabs, in the storybook but set to iPhone 5. The fourth tab is rendered underneath the other three.

eatyourgreens (approved):
This fixes the tab layout on desktop screens, but probably needs more work for tablets and phones.

eatyourgreens (approved):
This fixes the tab layout on desktop screens, but probably needs more work for tablets and phones.

packages/lib-react-components/src/Tabs/theme.js
eatyourgreens: How does this look with the default Grommet tab styles?

shaunanoordin: > I'm confused. Why doesn't `props.children.length` work here?

Turns out `props.children` always has exactly two children: the _tabs header,_ and the _tabs body_ (of the currently selected tab). So previously, whether you had 1 child tab or 100 child tab(s), each will always get allocated half the available flex space, i.e. `flex: 1 1 50%`

I'm assuming the _intent_ of `flex: 1 1 ${100 / props.children.length}%` was to ensure each child tab has an equal amount of flex-basis. (So with 3 tabs, each should get 33% base size.)

This does cause issues if we expect Tabs to have a lot of children with limited screen width, as you saw with the phone view.

> This looks a bit weird on phone screens, where flexbox moves the fourth tab down under the first three.
> How does this look with the default Grommet tab styles?

Grommet actually plays its much smarter by having [scrolling when there are too many child tab(s)](https://storybook.grommet.io/?path=/story/controls-tabs-responsive--responsive).

image

Not sure what we did to override this.

PR #3399, opened 2022-07-06
Update staging deploys to build-push-action/v3

eatyourgreens:
- Upgrade to `build-push-action/v3`.
- remove checkout, which isn't used by the build.

This is the build workflow that I've been using to deploy #2531 so I thought I'd update staging to use the same workflow.

- #2531 uses this workflow to deploy `fe-static.zooniverse.org` for testing.


zwolf (approved):
This needs to be updated to use the shared ci-cd workflow, which itself needs to be updated to use build-push-action@v3. That workflow, however uses `context: .` and requires the checkout to build.

This is fine for a v3 test but this repo needs some ci/cd attention I haven't paid it yet.

eatyourgreens:
I think one of the changes in v3 is that checkout is built into the action now.

PR #3345, opened 2022-06-24
Use internal API service names for SSR

eatyourgreens:
Use internal service names to connect directly to Panoptes when the project app is deployed to our kubernetes cluster.
- Add a `getServerSideAPIHost` helper, which will return the service name for a given deployment and API environment. It falls back to the default host in local development.
- Use `getServerSideAPIHost` to set an optional API host in all server-side data-fetching functions.
- Add Sentry logging to data-fetching functions
- Add custom request headers to SSR data fetching
- Update `panoptes-js` to support custom HTTP headers.
- Test default page props against a mocked API in mocked deployment environments.

This commit is included in #2531 but this PR makes the API client changes easier to see on their own. These changes will also work with either dynamic page builds or static page builds.

## Package
app-project
lib-panoptes-js

## Linked Issue and/or Talk Post
- included in #2531. 
- towards #3341.
- towards #3342.

## How to Review
The changes here are pulled out of #2531. When the project app is deployed to kubernetes (`APP_ENV` set to 'staging' or 'production'), server-side data fetching functions will use internal service names for the API: `http://panoptes-staging-app/api` or `http://panoptes-production-app/api` with the `X-Forwarded-Proto: https` header set.

This can't be tested locally, because development will continue to use `https://panoptes-staging.zooniverse.org/api` and  `https://www.zooniverse.org/api` for the Panoptes API, but the code here is deployed to https://fe-static.zooniverse.org eg. https://fe-static.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess

I've also added tests which run `getDefaultPageProps` against a mocked Panoptes API for both staging and production deploys.

# Checklist
_PR Creator - Please cater the checklist to fit the review needed for your code changes._
_PR Reviewer - Use the checklist during your review. Each point should be checkmarked or discussed before PR approval._

## General
- [ ] Tests are passing locally and on Github
- [ ] Documentation is up to date and changelog has been updated if appropriate
- [ ] You can `yarn panic && yarn bootstrap` or `docker-compose up --build` and FEM works as expected
- [ ] FEM works in all major desktop browsers: Firefox, Chrome, Edge, Safari (Use Browserstack account as needed)
- [ ] FEM works in a mobile browser

## General UX
Example Staging Project: [i-fancy-cats](https://local.zooniverse.org:3000/projects/brooke/i-fancy-cats)
- [ ] All pages of a FEM project load: Home Page, Classify Page, and About Pages
- [ ] Can submit a classification
[...]

PR #3400, opened 2022-07-06
getServerSideAPIHost: set default API env

eatyourgreens:
Set the default API environment to 'production', which matches our default environment for deploys.

I've also updated the tests to test against a mock API for the default `env`, rather than explicitly setting the API environment.

- #3345 sets the API host to use when the app is deployed, but the default API environment is undefined, which corresponds to `www.zooniverse.org`.


camallen (approved):
LGTM - this can be overridden using the `?env=staging` query param in the URL.

Issue #3336, opened 2022-06-21
Page zoom shrinks the subject viewer


eatyourgreens:
## Package
lib-classifier

## Describe the bug
Zooming into the Classify page, on any project, increases the text size (as expected) but shrinks the subject viewer. Zooming out shrinks the text but enlarges the subject.

https://user-images.githubusercontent.com/59547/174794439-4bef79f2-bd19-475e-8773-a9f1e22f510d.mov

## To Reproduce
1. Go to the Classify page on any project.
2. Use page zoom to zoom in.

## Expected behavior
The subject viewer should be enlarged, more like this screen recording.

https://user-images.githubusercontent.com/59547/174795413-19c200c2-d48f-4ed5-800e-f4c3b9548f97.mov

### Desktop (please complete the following information):
 - OS: MacOS 11.6.6
 - Browser: Firefox
 - Version: 101

## Additional context
The default layout sets the subject size to [`8fr`](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Basic_Concepts_of_Grid_Layout#the_fr_unit), which will shrink to fit remaining space after the task panel has enlarged.
https://github.com/zooniverse/front-end-monorepo/blob/a1af82bf7da492fa3b004fec31e37cb221242dbf/packages/lib-classifier/src/components/Classifier/components/Layout/components/DefaultLayout/DefaultLayout.js#L23-L25

PR #3415, opened 2022-07-08
Classifier: responsive layout for tablets

### LINKED ISSUE #3410: Classifier: layout isn't responsive on medium widths
## UI Issue

Package: `lib-classifier`
Follows: #3337 
Noted on commit 9fedf9157926adfc3f599598f2124595718486bf (not deployed to production yet) with macOS + Chrome 103 / FF 102

The Classifier is no longer responsive on "medium" viewport widths.

Expected: Task Area should always be visible in viewport, for medium widths. Subject image should shrink accordingly
Screenshot 2022-07-07 at 18 43 40

Actual: Task Area is cut off at medium widths. Subject image does not shrink.
Screenshot 2022-07-07 at 18 43 21

**Steps to Reproduce**

- Boot up `lib-classifier`
- Go to https://local.zooniverse.org:8080/?env=staging
- Gradually shrink the width of your browser window's width and observe the various responsive states.
- Wide format (> 1280px I think?) works fine, narrow/mobile format (< 700px) works wine, but medium format (anything in between) has no responsive layout.

### Dev Notes

I've stepped into the TARDIS to figure out when this bug was introduced, and can confirm that this is a result of the changes to Classifier's DefaultLayout in #3337.

- I don't think the loss of responsiveness is intended, as 3337 was supposed to fix the Subject's visual size when the browser zoom is activated
- A quick fix is to roll back 3337, but that would reopen #3336

### Status

Major UI issue, I think, but we have some time to take care of it.

This UI issue was noticed while I was giving @seanmiller26 a demo on QuickTalk on my localhost. As of 7 Jul 18:58 BST, this isn't out in the wild yet, so isn't an urgent problem yet, but should be fixed before the next deploy.

eatyourgreens:
First pass at a responsive layout for tablet screens:
- Subject fills available space on larger screens.
- Fixed task area width of 25.333rem (405px) on larger screens.
- Proportional columns in the ratio 45.333/25.333 on medium screens.
- Single column layout on small screens (less than 700px.)

Pixel widths are calculated assuming a root font size of 16px in the browser. Users may have customised this to fit their own needs eg. enlarged text for poor eyesight.

## Package
lib-classifier

- closes #3410.

## How to Review
Try out the project Classify page at a variety of viewport widths, representing different devices. Also try out the Classify page at a variety of page zoom levels (maybe up to 200%?)

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/Layout/components/DefaultLayout/DefaultLayout.js
+++ b/packages/lib-classifier/src/components/Classifier/components/Layout/components/DefaultLayout/DefaultLayout.js
@@ -12,17 +12,16 @@ import QuickTalk from '@components/Classifier/components/QuickTalk'
 import SubjectViewer from '@components/Classifier/components/SubjectViewer'
 import TaskArea from '@components/Classifier/components/TaskArea'
 
-const subjectSize = '45.333rem'
-const taskSize = '25.333rem'
-
 const ContainerGrid = styled.div`
   display: grid;
   grid-gap: 30px;
   grid-template-areas: "viewer task";
+  grid-template-columns: auto 25.333rem;
   position: relative;
 
-  @media (min-width: 701px) {
-    grid-template-columns: minmax(${subjectSize}, auto) ${taskSize};
+  @media screen and (max-width: 1160px) {
+    grid-gap: 1.75rem;
+    grid-template-columns: 45.333fr 25.333fr;
   }
 
   @media screen and (max-width: 700px) {


mcbouslog (commented):
At width 1015px (or between widths 995-1095 visually, but I think up to 1111) the tasks and theme toggle hang to the right, kind of off screen.

I think the media query might need be 1111px, based on (from left to right, in px):
- page padding left: 20
- project title: 18
- title-viewer grid gap: 20
- subject min: 725
- subject-task grid gap: 30
- task min: 240
- viewer-toggle grid gap: 20
- theme toggle: 18
- page padding right: 20
= total: 1111

Setting the query at 1111 makes the subject slightly smaller from 100%-110%, which is not ideal, but makes the subject larger at 125%. On mouse over in Chrome the zoom + and - buttons show "Make Text Larger" or "Make Text Smaller" so I think the slightly smaller subject but slightly larger text at 110% is ok / good / expected.

eatyourgreens:
Good point. Something I dislike here is that the classifier layout is controlled by the styling of the project app’s Classify page. The classifier is styled only for use in the project app, and any styling change in project page layout would also change the breakpoint here. 

I think a better layout would be styled by the parent page somehow, so that the classifier can be used in different contexts.

packages/lib-classifier/src/components/Classifier/components/Layout/components/DefaultLayout/DefaultLayout.js
```diff
@@ -12,20 +12,20 @@ import QuickTalk from '@components/Classifier/components/QuickTalk'
...
 
 const ContainerGrid = styled.div`
   display: grid;
   grid-gap: 30px;
   grid-template-areas: "viewer task";
   position: relative;
 
-  @media (min-width: 701px) {
-    grid-template-columns: minmax(${subjectSize}, auto) ${taskSize};
+  @media (min-width: 995px) {
```
mcbouslog, line 24:
```suggestion
  @media (min-width: 1111px) {
```

mcbouslog (approved):
100% agree regarding "the classifier layout is controlled by the styling of the project app’s Classify page", though for now I think will be good to get these changes in.

eatyourgreens:
I'm not sure about the suggested changes. They'd switch the Classify page to the single column view at screen widths < 1110px. I think a two column layout, but with a smaller subject, would be more appropriate on medium size screens.

eatyourgreens:
For example, change the minimum subject size to `25.333rem` so the smallest possible subject is a 50/50 split with the task area on smaller screens.

packages/lib-classifier/src/components/Classifier/components/Layout/components/DefaultLayout/DefaultLayout.js
```diff
@@ -12,20 +12,20 @@ import QuickTalk from '@components/Classifier/components/QuickTalk'
...
+    grid-template-columns: ${subjectSize} ${taskSize};
   }
 
-  @media screen and (max-width: 700px) {
+  @media screen and (max-width: 994px) {
```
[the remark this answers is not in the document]
mcbouslog: Suggestions not ideal, resolving conversation

packages/lib-classifier/src/components/Classifier/components/Layout/components/DefaultLayout/DefaultLayout.js
```diff
@@ -12,20 +12,20 @@ import QuickTalk from '@components/Classifier/components/QuickTalk'
 import SubjectViewer from '@components/Classifier/components/SubjectViewer'
 import TaskArea from '@components/Classifier/components/TaskArea'
 
-const subjectSize = '45.333rem'
-const taskSize = '25.333rem'
+const subjectSize = 'minmax(45.333rem, auto)'
```
mcbouslog, line 15:
```suggestion
const largeSubjectSize = 'minmax(45.333rem, auto)'
const mediumSubjectSize = 'minmax(25.333rem, auto)'
```

packages/lib-classifier/src/components/Classifier/components/Layout/components/DefaultLayout/DefaultLayout.js
```diff
@@ -12,20 +12,20 @@ import QuickTalk from '@components/Classifier/components/QuickTalk'
...
 const ContainerGrid = styled.div`
   display: grid;
   grid-gap: 30px;
   grid-template-areas: "viewer task";
   position: relative;
 
-  @media (min-width: 701px) {
-    grid-template-columns: minmax(${subjectSize}, auto) ${taskSize};
+  @media (min-width: 995px) {
+    grid-template-columns: ${subjectSize} ${taskSize};
```
mcbouslog, line 25:
```suggestion
  @media (min-width: 1111px) {
    grid-template-columns: ${largeSubjectSize} ${taskSize};
  }

  @media (min-width: 701px) and (max-width: 1110px) {
    grid-template-columns: ${mediumSubjectSize} ${taskSize};
  }
```

packages/lib-classifier/src/components/Classifier/components/Layout/components/DefaultLayout/DefaultLayout.js
mcbouslog, line 28:
```suggestion
  @media screen and (max-width: 700px) {
```

mcbouslog (commented):
@eatyourgreens I agree two columns for medium width preferable, here are some suggested changes towards that, though open to other approaches

packages/lib-classifier/src/components/Classifier/components/Layout/components/DefaultLayout/DefaultLayout.js
eatyourgreens: I think I’m missing something but this looks the same as:
```suggestion
const subjectSize = 'minmax(25.333rem, auto)'
```

mcbouslog: I've dismissed my suggestions - 'minmax(25.333rem, auto)' - would be the better way to implement, stinks that also had problems.

I think increase legibility of both. Per the mouse over in Chrome the zoom + and - buttons show "Make Text Larger" or "Make Text Smaller", which to me implies increasing both is ok/good, but I'm not sure.

eatyourgreens:
I've gone back and rewritten the media queries. The main change is proportional columns on medium viewports. It might re-open #3336, since zooming the page doesn't enlarge subjects any more. For example, you can't enlarge the handwriting to make a transcription page easier to read. That was the main use-case for larger subject images.

- Minimum subject width of 45.333rem (725px) on larger screens.
- Fixed task area width of 25.333rem (405px) on larger screens.
- Proportional columns in the ratio 45.333/25.333 on medium screens.
- Single column layout on small screens (less than 700px.)

eatyourgreens:
The two column layout looks a bit odd on an iPad Mini in portrait orientation but we can iterate on the breakpoints in a separate PR.

Screenshot of the project Classify page on an iPad Mini in protrait orientation. The subject and task appear side-by-side, slightly squashed.

shaunanoordin (approved):
## PR Review

Package: `lib-classifier`, but tested on app-project

This PR changes how the Classifier's responsive layout changes on different viewport sizes. We're interested in ensuring that the Classifier looks OK on three distinct "categories" of viewport _widths:_

- Small viewport widths (up to 700px) for mobile devices. This should use the single-column layout.
- Medium viewport widths (up to 1160px). This should use the two-column layout, and the **Task Area should shrink to fit available space.**
- Large viewport widths (above 1160px). This should also use the two-column layout, and the **Subject should expand to fit available space.**

### Testing

For the "Classifier looks OK" criteria, we're making some base assumptions for what a "normal" Classifier scenario would look like. The assumptions are:
- The project's Subjects will be portrait-oriented **images** with reasonable width:height ratios (e.g. 400x800px / 1:2, not some tall & thin 400x800000px / 1:2000)
- The project's Tasks will be of reasonable length (e.g. Question with 5 Answers, or a Text Task) (i.e. its vertical height of Task Area doesn't exceed viewport height by a ridiculous amount)
- A user's viewport height will at least be 700px tall in landscape view. (i.e. users aren't looking at websites through their letterbox slot)

Testing Steps:
- [Test URL](https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022/classify/workflow/3581?env=staging) will be on `app-project`, with a test project reflecting the assumptions above.
- A simple visual verification is all that's required.
- As long as a user can (1) view the Subject and (2) view the Task Area (either next to the Subject, or by scrolling down a bit), the test passes.

Tested on macOS (Macbook) with Chrome 103, using device emulation:
- custom responsive size, from 300x700 to 700x700: small layout looks OK 👍 
- responsive size, from 600x700 to 1000x700: medium layout looks OK 👍
  - Task Area resizes reasonably 
- responsive size, from 1200x700 to 4000x700: large layout looks OK 👍 
  - Subject image resizes reasonably, though obviously I'd be genuinely surprised if somebody has a 4000x700 screen. 
- emulating iPad Air (820x1180), both landscape & portrait: looks OK 👍 
- emulating Pixel 5 (851x393), both landscape & portrait: looks OK 👍 

Tested on Android (Sony Xperia) with Chrome (412x960 resolution):
- On portrait, Classifier uses small layout, and looks OK
- On landscape, Classifier uses medium layout, and looks OK

(Special thanks to @goplayoutside3 for pointing out mobile testing for localhost!)

### Status

LGTM 👍 Through all my testing, I couldn't any _(reasonable)_ dimensions in which the responsive layout becomes unusable or broken.

PR #3446, opened 2022-07-19
Drawing tools: edit freehand lines

eatyourgreens:
- Add an `editing` state to freehand lines.
- Double-click to start editing, then click a second time to select the area to cut.
- Drag the open drag handle to draw. Editing automatically ends when you reach the target point.
- ~~Closed shapes aren't fully supported at the moment.~~
- Undo isn't supported at the moment, to cancel accidental cutting of line segments.

https://user-images.githubusercontent.com/59547/179710672-1f6fb58c-a9bc-4479-b837-449dc9059cea.mov

## Package
lib-classifier

## How to Review
Freehand lines should be editable in any project that uses them, or in the Freehand Line story in the classifier storybook.

This PR doesn't include undo/redo, but a future PR could wrap `line.points` in an undo manager, so that we can undo and redo changes to a line. [UndoManager](https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mst-middlewares/README.md#undomanager) is an MST middleware that records patches to a model, which we've used elsewhere in the classifier to handle the 'Next' and 'Back' buttons.

# Checklist

## General

## General UX

## New Feature

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/experimental/components/FreehandLine/FreehandLine.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/experimental/components/FreehandLine/FreehandLine.js
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react'
+import React, { useState } from 'react'
 import PropTypes from 'prop-types'
 import { observer } from 'mobx-react'
 import styled from 'styled-components'
@@ -27,7 +27,11 @@ function createPoint(event) {
   const { clientX, clientY } = event
   // SVG 2 uses DOMPoint
   if (window.DOMPointReadOnly) {
-    return new DOMPointReadOnly(clientX, clientY)
+    const svgPoint = new DOMPointReadOnly(clientX, clientY)
+    const { x, y } = svgPoint.matrixTransform
+      ? svgPoint.matrixTransform(event.target?.getScreenCTM().inverse())
+      : svgPoint
+    return { x, y }
   }
   // jsdom doesn't support SVG
   return {
@@ -38,7 +42,6 @@ function createPoint(event) {
 
 function FreehandLine({ active, mark, onFinish, scale }) {
   const { path, initialPoint, lastPoint, finished, isClosed } = mark
-  const lineRef = useRef()
   const [editing, setEditing] = useState(false)
 
   const dragPoint = !mark.isCloseToStart && mark.dragPoint
@@ -56,10 +59,7 @@ function FreehandLine({ active, mark, onFinish, scale }) {
   }
 
   function onDoubleClick(event) {
-    const svgPoint = createPoint(event)
-    const { x, y } = svgPoint.matrixTransform
-      ? svgPoint.matrixTransform(lineRef.current?.getScreenCTM().inverse())
-      : svgPoint
+    const { x, y } = createPoint(event)
     mark.setDragPoint(mark.selectPoint({ x, y }))
     setEditing(true)
   }
@@ -68,10 +68,7 @@ function FreehandLine({ active, mark, onFinish, scale }) {
     if (!editing) {
       return true
     }
-    const svgPoint = createPoint(event)
-    const { x, y } = svgPoint.matrixTransform
-      ? svgPoint.matrixTransform(lineRef.current?.getScreenCTM().inverse())
-      : svgPoint
+    const { x, y } = createPoint(event)
     mark.cutSegment(mark.selectPoint({ x, y }))
   }
 
@@ -83,7 +80,6 @@ function FreehandLine({ active, mark, onFinish, scale }) {
   return (
     <StyledGroup
       className={ editing ? 'editing' : undefined}
-      ref={lineRef}
       onPointerUp={active ? onFinish : undefined}
     >
       {active && !editing && !isClosed && (
--- a/packages/lib-classifier/src/plugins/drawingTools/experimental/components/FreehandLine/FreehandLine.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/experimental/components/FreehandLine/FreehandLine.js
@@ -40,6 +40,12 @@ function createPoint(event) {
   }
 }
 
+function cancelEvent(event) {
+  event.preventDefault()
+  event.stopPropagation()
+  return false
+}
+
 function FreehandLine({ active, mark, onFinish, scale }) {
   const { path, initialPoint, lastPoint, finished, isClosed } = mark
   const [editing, setEditing] = useState(false)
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Now with support for clipping a closed loop.

https://user-images.githubusercontent.com/59547/179753840-bd7cb9ad-fb34-46a9-8913-a1c34c20800c.mov

eatyourgreens:
Added guide lines to show the path that's been cut.


https://user-images.githubusercontent.com/59547/179782605-619a9d9a-c8f8-4487-a5c4-0f1d44d13164.mov

eatyourgreens:
The interaction design here is very unforgiving of errors. I think it would be easier to use if the interactions went something like this:
1. Click (or double-click) and drag to place and move the start point.
2. Click and drag to place the end point.
3. Click (or double-click) the selected line segment in order to remove it.

Ideally, all these actions should be undoable (preferably with a gesture on tablets and touchscreens.)

If I start editing by mistake, how should I cancel? 'Esc' might work, but isn't available when drawing on a tablet with a stylus.

shaunanoordin:
I'm making a point to review this whole PR by the end of today, since I've unfortunately been ignoring it for the past 2 weeks+ . Bootstrapping, and go!

shaunanoordin (changes requested):
## PR Review

Package: `lib-classifier`
Part of: "Correct-A-Cell" project

This PR modifies the (experimental) **Freehand Drawing Line** tool, allowing users to edit _a selected segment_ of the freehand line (FHL).

- Original intent of the tool is this: if Caesar/any ML service provides a predefined freehand line shape, then allow the user to "correct" that line shape to better match the Subject image (usually the membrane of a biological cell).
- New behaviour: user can edit segments of the FHL mark (aka "correction feature")
  - user indicates the start of the segment by _double clicking_ on an existing FHL mark (closed or otherwise), then clicking at the end of the segment.
  - user can then use the "continue drawing" feature/behaviour of the FHL to correct the segment

This PR adds a significant new feature, which functions under _certain conditions,_ but there are two major issues.
- ❗ [blocking] the correction feature does not work on **Chrome 104.**
- ⚠️ [not-blocking] the correction feature acts very strangely if it's interrupted in the middle of the correction process (e.g. user accidentally clicking elsewhere to create a new line, or changing tools)

More details in dev notes. (EDIT: I was so focused on testing that I forgot to do a code read - I'll do that on the next pass.)

### Dev Notes

Full details of new behaviour:

1. (initial) let's say a user creates a new FHL mark. (The screenshots below use closed shapes, but open shapes/unclosed lines work similarly.)
     Screenshot 2022-08-15 at 18 40 40
2. (editing) user **double-clicks** on a part of the FHL to place the _"START of the segment to be replaced"._ The FHL is now in editing mode.
    Screenshot 2022-08-15 at 18 40 52
3. (editing) user **clicks** on another part of the FHL to place the _"END of the segment to be replaced"_ 
    Screenshot 2022-08-15 at 18 41 00
4. (editing) user can now **drag** the "start" marker to the "end" marker. The dragging (drawing) process _draws a new line segment..._ 
5. (completion) ...and when the user reaches the "end" marker, the "correction" process is complete, and the new line segment replaces the previously marked one.  
    Screenshot 2022-08-15 at 18 41 13

With this in mind, let's examine the issues mentioned earlier.

**Issue 1:** the correction feature does not work on Chrome.

Details:
- Tested on macOS + Chrome 104, standard desktop mode, with a mouse.
- Double clicking on a FHL mark doesn't do anything.
- In fact, replacing the path.onClick handler on `FreehandLine.js` with a simple console.log indicates that _no click events_ are triggering

Analysis:
- The problem lies with `draggable.js`'s `dragStart()`, which runs `setPointerCapture()`
- This only affects Chrome. Firefox and Safari aren't affected.
- I _think_ that when a user **clicks** on the FHL path...
  - user's mouse goes down, and path's "pointer down" triggers (if it had any).
  - event bubbles up to draggable, whose "pointer down" starts the pointer capture.
  - user's mouse goes up, and path's "pointer up" triggers (if it had any) then its "on click" triggers.
    - however, on Chrome, the pointer event is captured, so the "on click" handler gets ignored.
  - event bubbles up to draggable, whose "pointer up" stops the pointer capture.

Solutions?:
- add onPointerDown and onPointerUp to the FHL's `path`, which capture the pointer events _before_ the parent's `draggable` does the same?  
- After all, when there's a draggable child inside a draggable parent, the child's click/drag/pointer events all trigger properly, so I assume that's because the child captured the pointer before the parent could.

⚠️ Even if we have a solution, the fundamental discrepancy between Chrome and Firefox/Safari's behaviours is making me a bit wary. I might need to run this test again on BrowserStack to confirm it's not just an issue specific to my machine. (And if it isn't, what else is borked?)

**Issue 2:** the correction feature acts very strangely if interrupted.

Details:
- The correction feature works perfectly if the user follows steps 1 to 5 (see notes above) of the new behaviour.
- If the users encounters an interruption during steps 2 to 4 (the "editing" stage), the mark becomes "unstable"
- An interruption is any of the following:
  - User clicks on the canvas, triggering the "draw new FHL" action (very easy to accidentally do)
  - User changes the active Drawing Tool
  - User selects another existing mark
- "Unstable" can mean any of the following:
  - FHL mark stays in editing mode, even though this isn't a valid state to finish on
  - FHL starts "shrinking" due to confusion over what's the "current" selected segment
  - FHL is broken into multiple segments (difficult to replicate)
- **(special case)** If a user has already placed the Start and End marks for the segment, and then _proceeds to click on the FHL mark again,_ the FHL mark gets very confused as to which bits are selected, especially if you _alternately_ click on the sections before the Start/after the End mark.

Solutions?:
- Mark should listen for possible interruption scenarios (e.g. mark losing focus) and trigger a "cancel" process.
- (space left blank for future ideas) (still thinking about the special case)

### Status

I think the "doesn't work in Chrome" issue is a major one that needs to be addressed before this PR can be approved. ❗ This is required for baseline functionality.

The "correction feature acts strangely if interrupted" issue, while also significant, is something that can be handled in a separate PR. 👌 I'm aware that 1. this is an experimental tool anyhoo, and 2. that [we've already acknowledged the UI/UX limitations at this stage.](https://github.com/zooniverse/front-end-monorepo/pull/3446#issuecomment-1192407906)

> The interaction design here is very unforgiving of errors.
> If I start editing by mistake, how should I cancel? 'Esc' might work, but isn't available when drawing on a tablet with a stylus.

brb! Going to do a proper code read now and study how that "special case" interruption is actually happening.

eatyourgreens:
Here it is in Chrome 104 with slightly better event handling. I'll push these changes up.

https://user-images.githubusercontent.com/59547/184857368-4bb18f05-43c0-4aac-89aa-9bfdcfee7b51.mov

shaunanoordin:
Update acknowledged; I'm re-doing the code read, tests, and review now. 👍

shaunanoordin (approved):
## PR Review (Update)

Follows [previous review](https://github.com/zooniverse/front-end-monorepo/pull/3446#pullrequestreview-1073048576)

The latest update to this PR to fix the major issue where the Edit Freehand Line functionality wasn't working on Chrome.

- Fix here was to cancel the default event (see: `cancelEvent()` in `components/FreehandLine`)
  - Cancelling the default event stops the event from bubbling up to the Draggable wrapper, thus preventing the Draggable from running _pointer capture._

Code read looks good, and tests look good - there are no further blocking issues.

### Testing

Tested on localhost with macOS + Chrome with [a test project/workflow with Freehand Line drawing tool](https://local.zooniverse.org:8080/?env=staging&project=darkeshard%2Ftest-project-2022&workflow=3596)

Testing steps:
- Load the test workflow in lib-classifier, choose the Freehand Line tool
- Create a Freehand Line mark (drag & draw with a mouse)
  - Testing variations: (a) create a closed shape, (b) create an open shape, (c) select an existing FHL mark
- Once the FHL mark is selected, _double click_ anywhere on the _line_ to mark the "start of the segment to be edited/replaced" with a "start point"
- _Click_ anywhere on the lines to mark the "end of the segment" with an "end point"
- _Drag_ the start point to the end point, in a new path.
- Observe that the "segment to be edited/replaced" is now replaced with the new path

Note: the other issue, _"the correction feature acts very strangely if interrupted",_ still exists, but is not a blocker for this PR.

### Status

LGTM 👍 

This PR works as intended and meets the requirements to pass (i.e. the new functionality works and doesn't break anything that's already working). A follow up PR will be needed to address the _"the correction feature acts very strangely if interrupted"_ issue, but that's a tomorrow problem.

Let's get this merged!

PR #3559, opened 2022-08-19
Freehand line: cancel edits on deselect

### LINKED ISSUE #3558: Freehand Line: Edit/Correction feature acts strangely if interrupted
## UI/UX Bug on Experimental Feature

Follows #3446, specifically **issue 2** in [this review](https://github.com/zooniverse/front-end-monorepo/pull/3446#pullrequestreview-1073048576)

Context: in PR 3446, we modified the (experimental) Freehand Drawing Line tool, allowing users to edit a selected segment of the freehand line (FHL).

Problem: **the correction feature acts very strangely if interrupted.**

Details:
- The edit/correction works perfectly if user does all the following: 1. create FHL mark, 2. double click on FHL to place start point, 3. click on FHL to place end point, 4. drag start point draw new segment... 5. ...drag to end point to complete/confirm new segment
- If the users encounters an interruption during steps 2 to 4 (the "editing" stage), the mark becomes "unstable"
- An interruption is any of the following:
  - User clicks on the canvas, triggering the "draw new FHL" action (very easy to accidentally do)
  - User changes the active Drawing Tool
  - User selects another existing mark
- "Unstable" can mean any of the following:
  - FHL mark stays in editing mode, even though this isn't a valid state to finish on
  - FHL starts "shrinking" due to confusion over what's the "current" selected segment
  - FHL is broken into multiple segments (difficult to replicate)
- **(special case)** If a user has already placed the Start and End marks for the segment, and then proceeds to _click on the FHL mark again,_ the FHL mark gets very confused as to which bits are selected, especially if you _alternately_ click on the sections before the Start/after the End mark.

### Status

Low to medium priority, because while this definitely a problem that borks UI/UX and the data submitted, (1) the FHL still works if the user follows exact procedures, and (2) the FHL is still an experimental tool.

Low urgency since no active project (i.e. Correct-A-Cell) is using FHL's edit/correction feature at the moment.

Automatically assigned to Jim as a follow up to 3446, but please un-assign if you have other things on your place 👌

eatyourgreens:
- copy the current line path to `mark.originalPoints` before cutting a freehand line.
- listen to changes in the `active` prop and restore the original line when `active` is false (when the line loses focus.)

## Package
lib-classifier

- closes #3558.

## How to Review
Draw a freehand line, start to edit  it then try selecting another mark, or switching to a different drawing tool. On this branch, switching away from a line should cancel editing and restore the original line.

https://user-images.githubusercontent.com/59547/185629689-98efd437-37cf-4e61-8b82-02097e6ebdb6.mov

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/experimental/components/FreehandLine/FreehandLine.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/experimental/components/FreehandLine/FreehandLine.js
@@ -56,6 +56,12 @@ function FreehandLine({ active, mark, onFinish, scale }) {
     (point, index) => index === 0 ? `M ${point.x},${point.y}` : `L ${point.x},${point.y}`
   ).join(' ')
 
+  if (clippedPath && !active) {
+    mark.setCoordinates(mark.originalPath)
+    mark.setClipPath([])
+    cancelEditing()
+  }
+
   if (editing && !dragPoint) {
     cancelEditing()
   }
--- a/packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/FreehandLine/FreehandLine.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/experimental/models/marks/FreehandLine/FreehandLine.js
@@ -107,6 +107,7 @@ const FreehandLineModel = types
   .volatile(self => ({
     clipPath: types.array(SingleCoord),
     dragPoint: types.maybeNull(SingleCoord),
+    originalPath: types.array(SingleCoord),
     targetPoint: types.maybeNull(SingleCoord)
   }))
   .actions((self) => ({
@@ -122,7 +123,7 @@ const FreehandLineModel = types
     },
 
     setCoordinates(points) {
-      self.points = points
+      self.points = points.map(({ x, y }) => ({ x, y }))
     },
 
     move() {
@@ -160,6 +161,7 @@ const FreehandLineModel = types
       const deleteCount = targetIndex - dragIndex - 1
       const distFromEnd = self.points.length - targetIndex - 1
       const spansStartPoint = self.isClosed && (distFromEnd + dragIndex) < deleteCount
+      self.originalPath = self.points.map(({ x, y }) => ({ x, y }))
       if (!spansStartPoint) {
         /*
         Segment lies entirely within the line path, so splice points


shaunanoordin (commented):
## PR Review

Package: `lib-classifier`
Affects: experimental Freehand Line drawing tool
Fixes #3558 

This PR fixes (or _partially fixes,_ see below ⚠️ ) an issue, where, if

- Video 1 below shows the problem in action. Attempting to edit a FHL mark (by placing an edit segment start/end markers) and then interrupting the process (by clicking elsewhere on the image) does NOT undo the in-the-process-of-editing state of the mark. The mark appears "broken".
- Video 2 (0:08 to 0:16) shows the fix in action. Interrupting the edit process now simply resets the FHL mark. 👍  
- ⚠️ However, Video 2 (0:16 to 0:20) also shows an exception to the fix. If one start/end marker is placed **before** the 'origin point' of a **closed shape,** and the other marker is placed **after** the 'origin' point, then an interruption will cause the FHL mark to appear "broken".

**Video 1: the problem to be solved**
https://user-images.githubusercontent.com/13952701/187720349-791837af-252c-4afc-9fa3-4c2b2c58029c.mov

**Video 2: the solution, as of bc215ba**
https://user-images.githubusercontent.com/13952701/187720465-27e901ae-b161-4331-9489-6b7fa6dc39a2.mov

### Status

Code read looks good (storing the previous path in originalPath makes perfect sense) and the fix works in _most_ cases, but I think we'll want to figure out that _"exception to the fix"_ before this gets merged.

eatyourgreens:
In that second case, the start and end points of the line are trimmed, rather than spliced, so that's probably why the edits are lost. The start and endpoints of the path array change.

eatyourgreens:
@shaunanoordin that's fixed now.

shaunanoordin (approved):
## PR Review (Update)

Follows [initial review](https://github.com/zooniverse/front-end-monorepo/pull/3559#pullrequestreview-1092144044)

The recent changes to the branch fixes the "exception to the fix" issue found in the earlier review.

- New code read looks good. (And far more streamlined than the previous version, if I remember correctly?)
- Functionality checks out: if I place the start _and_ end "edit markers" on a FHL mark, then interrupt the edit process (by de-selecting the mark), then the edit process is cancelled and the FHL mark returns to its original shape 👍 
  - This is true even if I place the start and end markers before/after the "origin point" of the FHL mark, which was the cause of the "exception to the fix" issue. 👍 

This looks really good! I found some additional minor quirks (see dev notes), but they're not issues as far as I'm concerned.

### Dev Notes

_mac OS + Chrome on localhost:_
https://user-images.githubusercontent.com/13952701/188203734-997618a5-8615-444f-9ff9-a7f757fec171.mov

Minor quirk 1: (0:06 to 0:09 in the video)
- If you place _only the start marker_ then interrupt, then when you return to the FHL mark, that start marker is still there. (Instead of having no marker, which would be the case if the edit state is fully reset.)
- Impact is minimal - there's no wonkiness in the data being submitted, the UI visuals are fine, and the UI behaviour is only slightly off if the user was expecting the edit state to fully reset.

Minor quirk 2: (exactly 0:16)
- When a closed-shape FHL mark is in edit mode, it loses the "closing segment" visual.
- At exactly 0:16, if you look at the top tip of the triangle (the FHL mark's origin point), there's a gap that appears as soon as we enter editing mode.
- This is due to the FreehandLineModel's getPath() logic not adding that final `Z` any time there's a DragPoint.
- It's a minor cosmetic issue, AFAIC.

Minor quirk 3: (0:17 to 0:18)
- If you place the start/end markers near the origin point, the markers get "displaced" from where you actually click.
- e.g. if you _intend_ to place the end marker at point N (which is near the origin point), and you click at point N, then the end marker instead appears at point N+1
- This seems minor enough to me to consider it a quirk, but I've got a mental flag here to keep an eye out to see if test users agree or consider this an issue. 

### Status

LGTM! Let's get this merged. 👌

eatyourgreens:
Try modifying your local branch to add that closing `Z` even while editing. The result is interesting.

PR #3568, opened 2022-08-25
fix modal-subtask buttons on touch devices 

MaysaB:
Staging branch URL: 
https://github.com/zooniverse/front-end-monorepo/tree/fix-3344

## Package
lib-classifier
lib-react-components

## Linked Issue and/or Talk Post
Fixes #3344 

## Changes
Problem -  the subtask modal buttons on 'Save and Close' and 'X' do not work on mobile devices. Pointer actions outside of task area for pop up is being overidden and being replaced by drag actions.

Analysis - The 'Save and Close' button, as well as the 'Close' (x) button were missing the property that the task area has (className="subtaskpopup-element-that-ignores-drag-actions"). This property doesn't allow the pointer actions to be overridden.

Solution - add the property to both the Save and Close button in a <Box> element and also to the <CloseButton> element so that they ignore the drag actions (but only on those elements.)

See screen recording below. 
View on an iPhone12Pro 390x844 at 125%

https://user-images.githubusercontent.com/49612167/186196257-f3a9338d-8460-4691-9768-a18a667aa10b.mov

## How to Review
_Helpful explanations that will make your reviewer happy:_

- Zooniverse Projects to review UX

This is production env: 
(https://local.zooniverse.org:3000/projects/maceb/drawing-project/classify/workflow/3614?reload=0&workflow=3614)

This is staging env.
https://local.zooniverse.org:8080/?env=staging&project=maceb%2Fdrawing-project&workflow=3614

To view the solved bug:

1. Go to the staging environment
2. Create a new project, add a drawing tool task, add a subtask to the drawing tool task, ensure you've added a subject. 
3. Go to classify. Draw something on your subject. Click on the drawing to bring up the subtask. 
4. Go to developer tools and change the dimensions to AirPad or iPhone12.
5. Close the subtask using the 'save and close'.
6. Click the drawing again, close the subtask using the (x) on the top right hand corner of the component.

 **Ensure you have done yarn build on both lib-classifier and lib-react-components, then run yarn dev on app-project to run production env**

To test on mobile, Safari v.15.6:

1. Find your laptop’s IP address (on MacOS you can hold option + click on wifi symbol)
2. Run app-project locally
3. Connect your mobile device to the same wifi network as your laptop
4. Visit the project url on your mobile device, for example: https://your.ip.address.here:3000/projects/brooke/i-fancy-cats or https:/your.ip.address.here:3000/projects/maceb/drawing-project/classify/workflow/3614?reload=0&workflow=3614

[...]

PR #3570, opened 2022-08-25
Fix SubTaskPopup tests

eatyourgreens:
- add a `subtaskpopup-task` CSS class to task components.
- use that class to find tasks in the tests.

This merges into #3568, and should fix the broken tests in that PR.


MaysaB (approved):
Thank you for this Jim, definitely shouldn't have removed the CSS Class from box component. I want to look at the code in more detail but I've ran this locally and all the tests have passed, LGTM!

eatyourgreens:
I think the tests were making the mistake that only tasks could be styled with `.subtaskpopup-element-that-ignores-drag-actions` and each task component would be styled with that class. That essentially prevents you from adding that class to any other elements inside the popup.

PR #3586, opened 2022-08-31
Fix NextJS asset prefixes

### LINKED ISSUE #3587: Content Pages: can't use filters / incorrect asset deployment URL
## Functionality Issue

Package: `app-content-pages`
Pages: https://zooniverse.org/about/team and https://zooniverse.org/about/publications
Noted as of 31 Aug deployment

The problem:
- If you open our About > Team or About > Publications page, you'll get the page content, but the page isn't interactive.
- Clicking on the filters/sections (e.g. "Oxford" under About > Team) does nothing.

The analysis:
-  None of the JavaScript files (which would add the interactivity) are being pulled correctly 
    Screenshot 2022-09-01 at 01 00 13
- Here we see the page trying to get `https://fe-content-pages.zooniverse.org/_next/static/chunks/webpack-743de9e125f46c7a.js` (which uses the incorrect `fe-content-pages.zooniverse.org` base URL) when it should be calling `https://www.zooniverse.org/about/_next/static/chunks/webpack-743de9e125f46c7a.js` (note the `www.zooniverse.org/about` base URL)
- This must be an issue with the app-content-pages deployment that we're only noticing now.

The solution:
- PR fix ready: #3586

### Status

Problem identified and a solution incoming. I do believe this is a low-priority problem that can wait until next Wednesday's deploy, though.

Tagging @lcjohnso and @snblickhan as an FYI, and just in case either of you think the About pages need to be patched on the live servers ASAP as a hotfix.

eatyourgreens:
Update the production NextJS apps to fetch assets via `www.zooniverse.org`.

Since #2541, NextJS assets should be fetched from `www.zooniverse.org` in production and `frontend.preview.zooniverse.org` on staging.

Closes #3587.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,9 +38,9 @@ pipeline {
         COMMIT_ID = "${GIT_COMMIT}"
         CONTENTFUL_ACCESS_TOKEN = credentials('contentful-access-token')
         CONTENTFUL_SPACE_ID = credentials('contentful-space-ID')
-        CONTENT_ASSET_PREFIX = "https://fe-content-pages.zooniverse.org"
+        CONTENT_ASSET_PREFIX = "https://www.zooniverse.org/about"
         SENTRY_CONTENT_DSN = 'https://1f0126a750244108be76957b989081e8@sentry.io/1492498'
-        PROJECT_ASSET_PREFIX = "https://fe-project.zooniverse.org"
+        PROJECT_ASSET_PREFIX = "https://www.zooniverse.org/projects"
         SENTRY_PROJECT_DSN = 'https://2a50683835694829b4bc3cccc9adcc1b@sentry.io/1492691'
       }
 


shaunanoordin:
Do we need to worry about the `fe-content-pages.zooniverse.org` URL prefix in [`ci-build.yml`](https://github.com/zooniverse/front-end-monorepo/blob/907e6b53279caa73fca69152cd7e22b5c00f19cc/.github/workflows/ci-build.yml#L62) as well?

(And/or the `https://fe-project.zooniverse.org` URL prefix [in the same file?](https://github.com/zooniverse/front-end-monorepo/blob/907e6b53279caa73fca69152cd7e22b5c00f19cc/.github/workflows/ci-build.yml#L36))

eatyourgreens:
The CI builds aren't deployed, so they can really set the asset prefixes to any URL.

mcbouslog (approved):
Based on the conversations in Slack and review of existing deploy configs (note [preview](https://github.com/zooniverse/front-end-monorepo/blob/master/.github/workflows/deploy_staging.yml#L28-L36) is set to these code changes) I'm comfortable approving these code changes. Noting we can't test these changes until FEM is deployed to production and there will be additional follow-up work to refactor the production env variables.

mcbouslog:
**Copy-pasting a few related comments from Slack I found helpful as reference:**

The Jenkins build produces images with the wrong asset prefix, including pre-built static pages for Publications and Team. Those pages have the wrong asset prefix and will expire in 3600s. Once they expire, they will be rebuilt in the running container, which has the correct asset prefix.
The project app, on the other hand, dynamically builds the page HTML on every page request for every volunteer. While this is horribly inefficient, doesn’t scale up as volunteer numbers go up, makes multiple redundant calls to the Panoptes API etc. etc. it does mean that those builds always use the container environment for the running NextJS app.

TLDR: the initial Docker build has a different environment from the container environment (Jenkinsfile env vs. kubernetes template env.) For the first hour after a deploy, we see content pages that were built in that incorrect environment.

**And plans for further refactors:**

We didn’t used to be able to put secrets in the build arg step because it was in the Jenkinsfile, so they had to go in the k8s template. The Jenkinsfile’s being deprecated, so it’s a matter of figuring out what vars need to exist in GHA runners in addition to our own k8s deploys, then removing the remaining duplicates between the two.

I think the only reason they might be needed at build time is that the content pages app builds pages during the yarn build step. We could update that app so that pages are built on demand instead.
There could also be env vars that Webpack uses during the build step. I'm not sure.

PR #3671, opened 2022-09-17
Replace url-parse with URL

eatyourgreens:
Replace the [`url-parse`](https://www.npmjs.com/package/url-parse) package with the [native URL interface](https://developer.mozilla.org/en-US/docs/Web/API/URL), which is supported in Node and in browsers. Using the native interface should make both NextJS apps more secure. 

From `url-parse`'s Readme:
>`url-parse` was created in 2014 when the WHATWG URL API was not available in Node.js and the URL interface was supported only in some browsers. Today this is no longer true. The URL interface is available in all supported Node.js release lines and basically all browsers. Consider using it for better security and accuracy.

## Package
app-content-pages
app-project
lib-panoptes-js

## How to Review
Sign in and register should work in both NextJS apps. The project app should still load projects based on the slug in the page URL. 

# Checklist

## General

## General UX

## Refactoring

### DIFF
--- a/packages/app-content-pages/src/shared/components/AuthModal/AuthModalContainer.js
+++ b/packages/app-content-pages/src/shared/components/AuthModal/AuthModalContainer.js
@@ -2,18 +2,17 @@ import { inject, observer } from 'mobx-react'
 import { withRouter } from 'next/router'
 import { func, shape, string } from 'prop-types'
 import { Component } from 'react'
-import Url from 'url-parse'
 
 import AuthModal from './AuthModal'
 
 class AuthModalContainer extends Component {
-  constructor () {
+  constructor() {
     super()
     this.closeModal = this.closeModal.bind(this)
     this.onActive = this.onActive.bind(this)
   }
 
-  onActive (activeIndex) {
+  onActive(activeIndex) {
     const url = this.getUrlObject()
 
     if (activeIndex === 0) {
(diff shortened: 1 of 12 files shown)


packages/app-content-pages/src/shared/components/ZooHeaderWrapper/ZooHeaderWrapperContainer.js
```diff
@@ -4,72 +4,94 @@ import { withRouter } from 'next/router'
...
   }
 
-  redirect (urlObject) {
-    const { pathname, push } = this.props.router
-    const urlString = urlObject.toString().substr(urlObject.origin.length)
-    push(pathname, urlString, { shallow: true })
+  redirect(urlObject) {
+    const { push } = this.props.router
+    const urlString = urlObject.toString()
+    push(urlString, urlString, { shallow: true })
```
eatyourgreens, line 41: I’m not 100% sure, but I think you can pass URL objects to `router.push` without converting them to strings. That would simplify the code.

mcbouslog (commented):
Code changes look good, but when I run https://local.zooniverse.org:3000/projects/brooke/i-fancy-cats locally (on Chrome) it renders "window is not defined" and there are a few Uncaught Errors, the first of which is `Uncaught Error: Target container is not a DOM element.`.

mcbouslog:
> Code changes look good, but when I run https://local.zooniverse.org:3000/projects/brooke/i-fancy-cats locally (on Chrome) it renders "window is not defined" and there are a few Uncaught Errors, the first of which is `Uncaught Error: Target container is not a DOM element.`.


[This Sentry issue](https://sentry.io/organizations/zooniverse-27/issues/3605940210/?environment=production&environment=%22staging%22&environment=development&environment=staging&project=1492691) is me, related to error noted.

eatyourgreens:
I’ll stage this to kubernetes so that we can take a look. I assume the page just crashes in Node?

eatyourgreens:
Oh, I see the error in the diff. `AuthModalContainer` is missing a couple of lines in the project app.

packages/app-content-pages/src/shared/components/AuthModal/AuthModalContainer.js
```diff
@@ -27,53 +26,49 @@ class AuthModalContainer extends Component {
...
 
     return activeIndex
   }
 
-  getUrlObject () {
-    const { asPath } = this.props.router
-    return new Url(asPath, true)
+  getUrlObject() {
+    const isBrowser = typeof window !== 'undefined'
+    if (isBrowser) {
```
eatyourgreens, line 52: These changes also need to be added to the project app.

eatyourgreens:
Ok, this branch is deployed here (with the error):
https://fe-project-branch.preview.zooniverse.org/projects/brooke/i-fancy-cats?env=staging

I'm going to push a change to the `AuthModalContainer` code and deploy again.

eatyourgreens:
Re-deployed to `fe-project-branch` but it's still erroring because `url.searchParams` is undefined on the server.
https://fe-project-branch.preview.zooniverse.org/projects/marshexplorer/marsh-explorer

eatyourgreens:
@mcbouslog I'd made changes to the auth modal in the content pages app but not copied those same changes to the project app. At the moment, the Zooniverse page header has to be edited in both places 😒 . Everything should be working now eg. https://fe-project-branch.preview.zooniverse.org/projects/marshexplorer/marsh-explorer

eatyourgreens:
I'm also not sure why `AuthModalContainer` is rendered server-side, since you can only log in or register in the browser.

mcbouslog (approved):
Looks good! Tested locally, tested on https://fe-project-branch.preview.zooniverse.org/projects/marshexplorer/marsh-explorer, ran app-content locally and confirmed expected sign in and register modals appear 👍 .

PR #3704, opened 2022-09-27
Maintenance: replace deprecated Grommet 'tag' prop with 'as'

shaunanoordin:
## PR Overview

Packages: `app-project`, `lib-classifier`, `lib-react-components`

This PR is a minor maintenance refactor. The `tag` prop for Boxes (and by extension, all components descended from it) in Grommet is deprecated; the correct prop to use is `as`.

### Status

Ready for review, low priority.

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/dataVisAnnotation/components/DataVisAnnotationTask.js
+++ b/packages/lib-classifier/src/plugins/tasks/dataVisAnnotation/components/DataVisAnnotationTask.js
@@ -32,7 +32,7 @@ function DataVisAnnotationTask (props) {
 
   return (
     <Box>
-      <StyledText size='small' tag='legend'>
+      <StyledText as='legend' size='small'>
         <Markdownz>
           {task.instruction}
         </Markdownz>
--- a/packages/lib-react-components/src/ZooHeader/ZooHeader.js
+++ b/packages/lib-react-components/src/ZooHeader/ZooHeader.js
@@ -67,6 +67,7 @@ export default function ZooHeader({
 
   return (
     <StyledHeader
+      as='header'
       background='black'
       direction='row'
       fill='horizontal'
@@ -74,7 +75,6 @@ export default function ZooHeader({
       pad='none'
       responsive={false}
       role='presentation'
-      tag='header'
       {...props}
     >
       <Box
--- a/packages/app-project/src/shared/components/Stat/Stat.js
+++ b/packages/app-project/src/shared/components/Stat/Stat.js
@@ -7,16 +7,16 @@ function Stat ({ className, label, value }) {
   return (
     <div className={className}>
       <Text
+        as='div'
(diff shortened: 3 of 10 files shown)


packages/lib-react-components/src/ZooHeader/ZooHeader.js
```diff
@@ -67,14 +67,14 @@ export default function ZooHeader({
...
   return (
     <StyledHeader
+      as='header'
       background='black'
       direction='row'
       fill='horizontal'
       justify='between'
       pad='none'
       responsive={false}
       role='presentation'
```
eatyourgreens, line 77: Why has this header got a 'presentation' role?

packages/lib-classifier/src/plugins/tasks/dataVisAnnotation/components/DataVisAnnotationTask.js
```diff
@@ -32,7 +32,7 @@ function DataVisAnnotationTask (props) {
 
   return (
     <Box>
-      <StyledText size='small' tag='legend'>
+      <StyledText as='legend' size='small'>
```
eatyourgreens, line 35: All these `<legend>` tags have invalid content (#2578.) `<legend>` can only take a single string as content: no headings, paragraphs, lists, links etc.

eatyourgreens (approved):
Looks good, but reading the code highlighted a couple of other concerns around invalid or inaccessible HTML in the output.

packages/lib-react-components/src/ZooHeader/ZooHeader.js
eatyourgreens: Aha, [Using Navigation Landmarks](https://accessibility.blog.gov.uk/2016/05/27/using-navigation-landmarks/): a page should only have one `header`, which is a direct child of `body`. I assume this gives the page more than one header, in which case maybe remove the `header` here, or restructure the page so that it only has one header, as per the advice in Léonie Watson's post.

packages/lib-classifier/src/plugins/tasks/dataVisAnnotation/components/DataVisAnnotationTask.js
eatyourgreens: `legend` must be a child of `fieldset` too. These legends aren’t part of fieldsets.

packages/lib-react-components/src/ZooHeader/ZooHeader.js
shaunanoordin: Oh gods, we have **six** `header`s on a project home page like Planet Hunters TESS, only one has role=presentation.

I dove a bit into understanding how `header`s should used (issue: #3710) and am in a bit of a rabbit hole; I think my next step is to actually learn to properly use my Mac VoiceOver so I can understand how people with screenreaders experience our website, instead of just trying my best to understand the HTML & ARIA specs.

PR #3731, opened 2022-10-03
Maintenance Week: add tests for RotateHandle

shaunanoordin:
## PR Overview

Package: `lib-classifier`

Adds React Testing Library tests for the RotateHandle component

- RotateHandle is a mostly presentational component. (Functionality tests should be in draggable and the components that use the draggable RotateHandle.)
- Test IDs added for querying. (I based this on Rectangle.spec.js's pattern, since I can't find a good consensus online on what kind of ARIA role this should take on.)

Ready for review

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/components/RotateHandle/RotateHandle.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/components/RotateHandle/RotateHandle.js
@@ -9,7 +9,11 @@ const RotateHandle = forwardRef(function RotateHandle(
   const transform = `translate(${x}, ${y}) scale(${1 / scale})`
 
   return (
-    <g ref={ref} transform={transform}>
+    <g
+      data-testid='rotate-handle'
+      ref={ref}
+      transform={transform}
+    >
       <svg
         xmlns='http://www.w3.org/2000/svg'
         x='0'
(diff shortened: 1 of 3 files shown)


eatyourgreens (approved):
Looks good. Instead of test IDs, you can probably use `document.querySelector(‘g’)` and `document.querySelector(‘g svg’)`.

packages/lib-classifier/src/plugins/drawingTools/components/RotateHandle/RotateHandle.js
```diff
@@ -9,7 +9,11 @@ const RotateHandle = forwardRef(function RotateHandle(
   const transform = `translate(${x}, ${y}) scale(${1 / scale})`
 
   return (
-    <g ref={ref} transform={transform}>
+    <g
+      ref={ref}
+      transform={transform}
+      data-testid='rotate-handle'
+    >
```
eatyourgreens, line 16: You can add alt text to SVGs by giving them `role=“img”` and an `aria-label`.

packages/lib-classifier/src/plugins/drawingTools/components/RotateHandle/RotateHandle.js
```diff
@@ -9,7 +9,11 @@ const RotateHandle = forwardRef(function RotateHandle(
...
       <svg
         xmlns='http://www.w3.org/2000/svg'
```
eatyourgreens, line 18: You can add alt text to SVGs by giving them `role=“img”` and an `aria-label`.

eatyourgreens: If you do do that, I recommend [Contextually Marking up accessible images and SVGs](https://www.scottohara.me/blog/2019/05/22/contextual-images-svgs-and-a11y.html) by Scott O’Hara.

shaunanoordin:
👍 Thanks for the feedback. I've removed the testIDs in favour of document queries, and added an aria-label for the inner `<svg>`

PR #3716, opened 2022-09-29
Classifier: Refactor survey task Chooser to conditionally show filter components

mcbouslog:
## Package
- lib-classifier

## Linked Issue and/or Talk Post
Closes #3689 , towards other accessibility issues

## Describe your changes
- conditionally render survey task filter components (FilterStatus and ClearFilters) if there are filters
- update tests with react-testing-library

## How to Review
_Helpful explanations that will make your reviewer happy:_
- What Zooniverse project should my reviewer use to review UX?
  - with filters - https://local.zooniverse.org:8080/?project=markb-panoptes%2Fsurvey-task-test-project&workflow=2434
  - without filters - https://local.zooniverse.org:8080/?project=markb-panoptes%2Fsurvey-task-test-project&workflow=3627
- What user actions should my reviewer step through to review this PR?
  - confirm filters components are shown or not shown per survey task with or without filters
- Which storybook stories should be reviewed? n/a
- Are there plans for follow up PR’s to further fix this bug or develop this feature? yes, see linked Milestone

# Checklist
_PR Creator - Please cater the checklist to fit the review needed for your code changes._
_PR Reviewer - Use the checklist during your review. Each point should be checkmarked or discussed before PR approval._

## General
- [ ] Tests are passing locally and on Github
- [ ] Documentation is up to date and changelog has been updated if appropriate
- [ ] You can `yarn panic && yarn bootstrap` or `docker-compose up --build` and FEM works as expected
- [ ] FEM works in all major desktop browsers: Firefox, Chrome, Edge, Safari (Use Browserstack account as needed)
- [ ] FEM works in a mobile browser

## General UX
Example Staging Project: [i-fancy-cats](https://local.zooniverse.org:3000/projects/brooke/i-fancy-cats)
- [ ] All pages of a FEM project load: Home Page, Classify Page, and About Pages
- [ ] Can submit a classification
- [ ] Can sign-in and sign-out
- [ ] The component is accessible
  - Can be used with a screen reader [BBC guide to VoiceOver](https://bbc.github.io/accessibility-news-and-you/assistive-technology/testing-steps/voiceover-mac.html)
  - Can be used from the keyboard [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - It is passing accessibility checks in the storybook

## Bug Fix
- [ ] The PR creator has listed user actions to use when testing if bug is fixed
- [ ] The bug is fixed
- [ ] Unit tests are added or updated

PR #3735, opened 2022-10-05
Classifier: Refactor Chooser and child components for focus and accessibility improvements

### LINKED ISSUE #2312: Convert survey task Choices to ul and ChoiceButton to li
Convert survey task components Choices to `ul` and ChoiceButton to `li` for improved accessibility.

Originally noted in #1991.

Noting per MDN's [menu](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu) and [menuitem](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menuitem) documentation, those options are not well supported or deprecated, respectively.

### LINKED ISSUE #3298: Survey task steals keyboard focus
## Package
lib-classifier

## Describe the bug
The choice buttons in the survey task use `autofocus`, leading to a couple of serious usability and accessibility problems:
1. The filter component is skipped, meaning that users of screen readers or screen magnifiers might not even know that the list of choices can be filtered.
2. When the popup tutorial is open, keyboard focus is underneath the modal layer.
Screenshot of Wildcam Darién. The modal tutorial is open but the Ñeque option in the Task panel has keyboard focus.

## To Reproduce
I think this can be reproduced on any survey task project. You may need to be logged out, in order to force the tutorial to open.

## Expected behavior
When the tutorial is open, it should have focus. It's already coded to do this, so avoid having autofocus anywhere else in the classifier when the tutorial is open.

If the Task panel must be focused (eg. when starting a new classification), keyboard focus should go to the first focusable element in the panel, which is the filter control. In that case, I'm not sure how a user that relies on a screen reader, or a screen magnifier, would know that the subject has changed but I can see how focusing the filter control would make classification easier for a keyboard user.

## Additional context
- [WCAG 2.4.3: Focus order.](https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html)
- [WCAG 3.2.1: On focus.](https://www.w3.org/WAI/WCAG21/Understanding/on-focus.html)

Also see #2455, where autofocus skips survey task choice descriptions for screen reader users.

mcbouslog:
## Package
- lib-classifier

Closes #2312 , closes #3298 

Does not address additional accessibility concerns, test conversions, and storybook refactors further detailed in the [Survey Task: Pathfinder Projects milestone](https://github.com/zooniverse/front-end-monorepo/milestone/38).

## Describe your changes
- [x] refactor Choices and ChoiceButton to `ul` and `li`s, respectively
- [x] remove autoFocus from survey task
- [x] fix Choices/ChoiceButton stealing focus with modal (i.e. Tutorial) open
- [x] refactor ChoiceButton focus styling
- [x] refactor FilterStatus tests with react-testing-library
- [x] refactor Choices tests with react-testing-library
- [x] refactor ChoiceButton tests with react-testing-library

## Issues identified in this PR to be addressed in subsequent PRs:
- (#3816) refactor SurveyTaskContainer and SurveyTask tests with behavioral tests, including:
  - ChoiceButton on 'Backspace,' on 'Enter,' and on 'Space'
  - any other implementation tests (i.e. no "...should call..." tests)
- (#3816) add full survey task story, refactor/delete other stories
- (#3745) check filter button accessibility (icon buttons accessible to voice commands?)
- (#3824) Choice button interactions are inconsistent
- (#3826) thumbnails reloading and related Choices layout reflow
- (#3831) Choice closes on 'Escape' keydown

## How to Review
  - note with tutorial modal showing the survey task choices aren't stealing focus
  - http://localhost:6006/?path=/story/tasks-surveytask-chooser--default&globals=locale:en;locales.en:English;locales.test:Test+Language
- Are there plans for follow up PR’s to further fix this bug or develop this feature? yes, see milestone

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

## Refactoring


eatyourgreens:
When the subject changes, it might be a better UX for blind users if the focus shifts to the new subject. I’m thinking of survey tasks with audio subjects. 

If the filter is auto focused, you have to shift-tab back to the player controls, which is annoying. 

At the moment, I’m thinking maybe just remove autofocus altogether from survey tasks.

mcbouslog:
> At the moment, I’m thinking maybe just remove autofocus altogether from survey tasks.

I agree! Autofocus with all tasks doesn't work on initial classifier load anyway, I think only works on next step. I'll remove.

eatyourgreens:
One other accessibility question, which may or may not be relevant here. We have had contact, in the past, with a volunteer who had a severe motor disability, such that they were unable to use the keyboard or a mouse much. They dictated their classifications with Dragon Naturally speaking eg. by saying 'Click wildebeest', 'Click identify' and so on.

All the buttons here have clear text labels, so that should still work, but the filter controls now have icon buttons so I'm not sure how accessible they are to voice commands. How would a voice user know how to select a filter?

Not a blocking comment, but something to keep in mind.

mcbouslog:
Per #3716 - a To Do for this PR - add a Chooser test for choices remaining after a user has selected a filter. The refactored filter buttons included in these changes will hopefully make the noted test straightforward.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
```diff
@@ -26,10 +26,10 @@ function ChoiceButton({
 
   const choiceButton = useRef(null)
   const handleClick = useCallback(() => {
-    onChoose(choiceId)
+    !disabled && onChoose(choiceId)
```
eatyourgreens, line 29: Disabled buttons can’t fire click events, so I’m curious as to how this gets triggered.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
```diff
@@ -26,10 +26,10 @@ function ChoiceButton({
...
   }, [choiceId, onChoose])
   const handleKeyDown = useCallback((event) => {
-    onKeyDown(choiceId, event)
+    !disabled && onKeyDown(choiceId, event)
```
eatyourgreens, line 32: Similarly, disabled buttons can’t respond to the keyboard either. They’re removed from the tab order in the DOM.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
mcbouslog: By adding the `as='li` I think it's taking away some expected button behavior, including `disabled` functionality. For the related tests I added this line (and the similar line in `handleKeyDown`).

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
```diff
@@ -51,7 +51,9 @@ function ChoiceButton({
 
   return (
     <Button
+      a11yTitle={choiceLabel}
       ref={choiceButton}
+      as='li'
```
eatyourgreens, line 56: List items aren’t focusable or clickable. The list of buttons should be marked up as a list, but the buttons should be inside the list items.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
eatyourgreens: You shouldn’t use `as=‘li’` on buttons! A list of buttons should look like this in HTML. A screen reader will announce this as a list with two items. Each item is a button. 

```html
<ul>
  <li><button>First Button</button></li>
  <li><button>Second Button</button></li>
</ul>
```

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/Choices.js
```diff
@@ -99,8 +110,13 @@ export function Choices({
...
+        const hasFocus = index === focusIndex
+        let tabIndex = -1
+        if (focusIndex === null && index === 0) {
+          tabIndex = 0
+        } else if (index === focusIndex) {
+          tabIndex = 0
+        }
 
         return (
           <ChoiceButton
```
eatyourgreens, line 122: Now that `StyledGrid` is an unordered list, the individual buttons should be wrapped in `li`. Best to keep the `ul` and `li` elements together in the code, so I would put the list items here.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
eatyourgreens: Since the list can be navigated with the arrow keys, it behaves like a menu, so it might be appropriate to mark it up as a menu:

```html
<ul role=“menu”>
  <li role=“menuitem”>
    <button>First Button</button>
  </li>
  <li role=“menuitem”>
    <button>Second Button</button>
  </li>
</ul>
```

mcbouslog: Great suggestions, refactored accordingly!

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/CharacteristicsFilter/FilterStatus/FilterStatus.spec.js
```diff
@@ -1,91 +1,100 @@
...
   before(function () {
-    wrapper = mount(
-      <FilterStatus
-        task={task}
-      />, {
-        wrappingComponent: Grommet,
-        wrappingComponentProps: { theme: zooTheme }
-      }
-    )
+    scrollToMock = sinon.stub(window, 'scrollTo')
```
eatyourgreens, line 25: I think `window.scrollTo` might already be mocked in the global test setup. If not, can you add it, so that it’s available for all tests?

eatyourgreens: You can probably copy the project app test setup. 
https://github.com/zooniverse/front-end-monorepo/blob/feb82afe5f19f269f688c8fb8e3a38fd596016fb/packages/app-project/test/create-global-document.js#L32

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/Choices.js
```diff
@@ -93,37 +95,46 @@ export function Choices({
...
-            disabled={disabled}
-            hasFocus={hasFocus}
-            onChoose={onChoose}
-            onKeyDown={handleKeyDown}
-            selected={selected}
-            src={src}
-            tabIndex={tabIndex}
-            thumbnailSize={thumbnailSize}
-          />
+            role='menuitem'
```
eatyourgreens, line 116:
```suggestion
            role='menuitemcheckbox'
            aria-checked={ selected ? ‘true’ : ‘false’ }
```

When you add a role, you should also check which states apply to that role, and add them if necessary. 

A checkbox might be good here because many items can be selected at a time.

eatyourgreens (commented):
A checked state might help tell which menu items are selected.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/Choices.js
eatyourgreens: After reading the MDN docs for checkbox menu items, I’m wondering if this role and state should be set on the button. 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitemcheckbox_role

eatyourgreens: Yeah, the HTML in the example here has the `menuitem` roles on the clickable links. That was my mistake. Menu items should be interactive elements like links, buttons and checkboxes. 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role

mcbouslog: Great suggestions, thank you!
Per the linked ARIA docs I added `role='presentation'` to the `li`s.

eatyourgreens:
This is looking good. Just noting that I've rebased now that #3782 has merged, so you may need to pull down a fresh copy of this branch from GitHub.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/Choices.js
```diff
@@ -35,18 +37,18 @@ export function Choices({
...
       const lastSelectedIndex = filteredChoiceIds.indexOf(selectedChoiceIds[selectedChoiceIds.length - 1])
       if (lastSelectedIndex > -1) {
         setFocusIndex(lastSelectedIndex)
       } else {
-        setFocusIndex(0)
+        setFocusIndex(null)
       }
     } else {
-      setFocusIndex(0)
+      setFocusIndex(null)
```
eatyourgreens, line 51: When the focus index is null, where does keyboard focus go?

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
```diff
@@ -52,7 +52,10 @@ function ChoiceButton({
   return (
     <Button
       ref={choiceButton}
+      aria-checked={selected}
```
eatyourgreens, line 55:
```suggestion
      aria-checked={selected ? 'true' : 'false'}
```
ARIA states are always strings. There aren't any boolean attributes in ARIA.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.spec.js
```diff
@@ -1,128 +1,107 @@
...
+            choiceId='RDVRK'
+            choiceLabel='Aardvark'
+            disabled
+            hasFocus
+            onKeyDown={onKeyDownSpy}
+            tabIndex={0}
+          />
+        </Grommet>
+      )
+      await userEvent.keyboard('{enter}')
```
eatyourgreens, line 103: I don't think this `userEvent` is pressing 'Enter' on the choice button. You might need to specify a target element for the event.

eatyourgreens: You should probably also update your user event tests from v13 to v14, and use [`userEvent.setup()`](https://testing-library.com/docs/user-event/setup/) to create a user.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.stories.js
```diff
@@ -45,18 +55,22 @@ const Template = ({
...
+    <Box
+      width={getColumnWidth(thumbnailSize)}
+    >
+      <ChoiceButton
+        choiceId={choiceId}
+        choiceLabel={choiceLabel}
+        disabled={disabled}
+        hasFocus={hasFocus}
+        onChoose={() => console.log(choiceId)}
+        onKeyDown={({ key }) => console.log(choiceId, key)}
```
eatyourgreens, line 67: These should use [Storybook actions](https://storybook.js.org/docs/react/essentials/actions) rather than console logs.

eatyourgreens (commented):
Some small changes. I'd strongly recommend upgrading your user event tests to the latest version. The version 13 API is deprecated, but still works for convenience.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/Choices.js
mcbouslog: It appears to go back to the `body` when I test on app-project https://local.zooniverse.org:3000/projects/markb-panoptes/survey-task-test-project/classify/workflow/2434 or lib-classifier https://local.zooniverse.org:8080/?project=1634&workflow=2434 , then when I tab on app-project keyboard focus goes to the Filter button.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.stories.js
mcbouslog: yes! nice catch.
I still plan to rewrite all the stories and create a full survey task story, opened #3816 to track.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/Choices.js
eatyourgreens: Is that a problem?

eatyourgreens (approved):
Changes look good. I’m not sure about focus being reset to the document body when you’re classifying with the keyboard and remove a choice. Will that cause problems? If I’ve removed a selected menu item, but that menu item is active, I’d expect it to stay active and focused, just change state from checked to unchecked. 

When the Choice panel closes, I’d expect focus to return to the active menu button, even after pressing cancel.

eatyourgreens:
When I'm classifying on the test workflow from the keyboard:
- the filter buttons are almost impossible to use but I assume that's covered by #3746.
- the focus indicator for the choice menu is a background colour, not the default focus ring, for the active button. That's really easy to confuse with a selected choice, and volunteers have commented about this being confusing on the old survey task. Our interaction design should be consistent throughout and try to use the default focus indicator.

eatyourgreens:
When a choice is open, the button interactions are inconsistent.
- Not This and Identify both respond to Enter and Space.
- Question radio and checkbox answers only respond to Space.

eatyourgreens:
If I close a choice by pressing Enter or Space on 'Not This', the menu component loses focus so I can't carry on and select another choice by pressing the up arrow or down arrow.

This is a serious bug and I'd prefer to see this one fixed before this PR is merged, if possible. The active choice should retain focus, even if it wasn't checked with the Identify button.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTaskContainer.js
```diff
@@ -134,10 +147,12 @@ class SurveyTaskContainer extends React.Component {
         answers={answers}
         disabled={disabled}
         filters={filters}
+        focusedChoice={focusedChoice}
         handleAnswers={this.handleAnswers.bind(this)}
         handleChoice={this.handleChoice.bind(this)}
         handleDelete={this.handleDelete.bind(this)}
         handleFilter={this.handleFilter.bind(this)}
+        handleFocusedChoice={this.handleFocusedChoice.bind(this)}
```
eatyourgreens, line 155: I missed this before, but you shouldn’t run `.bind` on every render. It creates a new, bound function each time it runs. You should bind your event handlers once, in the component constructor, like this:

```js
this.handleEvent = this.handleEvent.bind(this)
```

You can then safely reference `this.handleEvent` in your component.

All the event handlers here need to be updated.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/Choices.js
```diff
@@ -28,30 +28,18 @@ const defaultTheme = {
...
   filteredChoiceIds = [],
+  focusedChoiceId = '',
   handleDelete = () => {},
+  handleFocusedChoice = () => {},
   onChoose = () => true,
   selectedChoiceIds = [],
   task,
   theme = defaultTheme
 }) {
-  const [focusIndex, setFocusIndex] = useState(null)
```
eatyourgreens: I think you could have kept this to track state while the choices menu is open, but change the initial state from `null` to `focusedChoiceId` when the menu mounts.

eatyourgreens: If you were to do that, then the Choice component should store the current choice when you press Not This or Identify, which then becomes the initial value here. Otherwise, this component wouldn’t need to change.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/Choices.spec.js
```diff
@@ -39,78 +39,82 @@ describe('Component > Choices', function () {
...
-    expect(choiceButtons[1]).to.equal(document.activeElement)
-    await user.keyboard('{arrowup}')
-    expect(choiceButtons[0]).to.equal(document.activeElement)
+    await user.keyboard('{delete}')
+    expect(handleDeleteSpy).to.have.been.calledOnceWith('RDVRK')
   })
 
-  it('should call handleDelete with choice ID on choice button backspace keyDown', async function () {
-    const handleDeleteSpy = sinon.spy()
+  it('should call handleFocusedChoice with expected choice ID on arrowDown keyDown', async function () {
```
eatyourgreens, line 80: Testing implementation like this is very brittle since it blocks you from refactoring your code. I know I’ve said this before, but this is an important lesson to learn. If you test the behaviour of your components then you can refactor safely, knowing that the behavioural tests will still pass if the refactor works.

eatyourgreens: Behavioural tests for the menu here would be tests like:

- it should delete the selected choice on Backspace. 
- it should change the selected choice on ArrowUp. 
- it should change the selected choice on ArrowDown. 

Those behaviours won’t change when you refactor, even if you change the internal details of how state is stored and handled.

eatyourgreens (approved):
The menu focus bug is fixed. 👍 

There's a weird glitch when the choice panel closes: the menu thumbnails reload (and layout reflows), even though they're already loaded. See this video. Incidentally, the video also shows the confusing focus state early on. Around 15s it looks like both Eland and Crane are active (Eland is actually focused.)


https://user-images.githubusercontent.com/59547/198227277-c6480323-f04a-4e8e-97d1-323bb7d5381a.mov

In the video above, I have to tab all the way through the `Choice` component in order to cancel my choice with 'Not This', which is very annoying. 'Esc' is a pretty standard convention for closing popups that have opened by mistake. Can the `Choice` component cancel and return to the menu when 'Esc' is pressed?

You've now got the `SubTask` component managing the state of the `Chooser` menu. It might be confusing to have one component managing the state of another component like that.

Event handlers in class components should be bound once, when the component mounts, so that needs to be fixed. You can follow this example here, from the project app.
https://github.com/zooniverse/front-end-monorepo/blob/92b7aff8c8ef10fe4904debf1e95434809026d3b/packages/app-project/src/shared/components/CollectionsModal/CollectionsModalContainer.js#L28-L45

mcbouslog:
I think there might be some confusion on the Choice and Chooser component renderings, [code here in the SurveyTask component](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.js#L26-L45) - depending on `selectedChoice` the Choice component renders or the Chooser component renders. I think that's causing the noted "menu thumbnails reload", that's what's preventing Escape or similar from closing the Choice component, and that's causing `focusedChoiceId` state if set/kept in Choices to not persist when a Choice is opened and then closed (either from Not This or Identify), as Chooser (and child Choices) is completely re-rendering after a Choice is closed. These are all still issues that need to be addressed, but maybe that help explains why these issues are happening.

eatyourgreens:
I’m confused by that comment. None of those problems are caused by the `Chooser` component unmounting and mounting again.

eatyourgreens:
I think the content reflow is a bug in the Media component, which is completely refreshing images even when they’re in the browser cache.

eatyourgreens:
From a quick read of the code yesterday, the `Choice` component could call `handleDelete(choiceId)` on ‘Esc’. That doesn’t seem hard. That could be handled either in the `Choice` component or in `SurveyTask`.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/Choices.spec.js
mcbouslog: I refactored the ArrowUp and ArrowDown tests in Choices accordingly, will refactor Backspace, Enter, and Space in #3816 .

eatyourgreens:
I've opened #3825 to update the `Media` component for thumbnail images. I'm not sure if it will fix the layout reflow when the `Chooser` unmounts and mounts again.

mcbouslog:
Thank you @eatyourgreens for your reviews!

I've documented issues identified in the PR that will be addressed with subsequent work in the initial comment with links to related GitHub issues. Please let me know if I've missed any!

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/Choices.js
mcbouslog: I think I've refactored according to suggestions. There is now a `previousChoice` state in SurveyTaskContainer and `focusIndex` state in Choices. I was confused as I thought you were suggesting to move what was `focusedChoiceId` state from SurveyTaskContainer to Choices, but I think that interpretation of your suggestion is incorrect. Hopefully the current refactor better reflects your suggestion.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/Choices.js
```diff
@@ -93,41 +85,52 @@ export function Choices({
...
-            onKeyDown={handleKeyDown}
-            selected={selected}
-            src={src}
-            tabIndex={tabIndex}
-            thumbnailSize={thumbnailSize}
-          />
+            role="presentation"
+          >
+            <ChoiceButton
+              choiceId={choiceId}
```
eatyourgreens, line 109: In terms of organising the code, I would declare the menu item roles and states here, as optional props for `ChoiceButton`, so that all the menu states are declared in the same component. It’s not a big deal, but choice button can’t be rendered in isolation otherwise eg. if you have a `ChoiceButton` story at the moment, you’re rendering a menu item in the browser without a wrapping menu. 

Of course, you could also update `ChoiceButton` stories and tests to make sure that the button is always wrapped in a menu. 

It’s the equivalent of writing a list item outside of a list, which is invalid HTML. 

```html
<li>this is invalid HTML</li>
```

If you aren’t rendering isolated menu items in stories and tests, you can ignore this comment.

mcbouslog: Yes that makes sense!

PR #3749, opened 2022-10-10
Maintenance Week: add/update tests for Tabs and Tab(s)

shaunanoordin:
## PR Overview

Package: lib-classifier

This PR Adds React Testing Library tests for the Tabs component, and updates the Tab (singular) component to use Enzyme tests.

- `Tab` no longer tests to see if it's specifically generating a "SpacedText" element. As discussed on Slack, the RTL method isn't interested with the "hows" of the implementation.
- The Tabs Story has been updated to demonstrate the use of more complex titles.

Ready for review.


packages/lib-react-components/src/Tab/Tab.spec.js
```diff
@@ -1,42 +1,72 @@
...
-    const tabTitle = wrapper.find(TITLE)
-    expect(tabTitle).to.have.lengthOf(1)
+  describe('with complex props', function () {
+    it('should should pass through the `title` prop if it\'s a component', function () {
+      const ComplexTitle = () => (
+        <div>
+          <Key size='small' />
+          &nbsp;
+          <b data-testid="b-tag">Special</b>
+        </div>
```
eatyourgreens, line 54:
```suggestion
        <span>
          <Key size='small' />
          &nbsp;
          <b data-testid="b-tag">Special</b>
        </span>
```

I suspect you want the tab title to render inline, in which case it should use a text `<span>` rather than `<div>`.

Also, I suspect that the text label for this tab is 'Key Special' but that you intend the tab name to be 'Special'. You can add a test to get the tab by name and check that.

packages/lib-react-components/src/Tab/Tab.spec.js
```diff
@@ -1,42 +1,72 @@
...
+      render(
+        <Grommet theme={zooTheme} >
+          <Tabs>
+            <Tab title={<ComplexTitle />}>A tab with a rather complex title (tab header)</Tab>
+            <Tab title='bananas'>A banana is a long fruit.</Tab>
+          </Tabs>
+        </Grommet>
+      )
+
+      const keySvg = screen.getByLabelText('Key')
```
eatyourgreens, line 66:
```suggestion
      const complexTab = screen.getByLabelText('Key Special')
      const keySvg = screen.getByLabelText('Key')
```
Maybe add a query that checks the name of the tab in screen readers? I'm not convinced that 'Key Special' is an accessible name, to be honest.

eatyourgreens (commented):
Tests look good but I'm not convinced of the accessibility of the complex title example. Complex titles should have accessible text alternatives.

packages/lib-react-components/src/Tab/Tab.spec.js
eatyourgreens: OOI, if any HTML markup is allowed inside tab titles, what happens if you pass in an ordered or unordered list? We should probably restrict the content to `span` and other inline elements, like the content model for `<legend>` or `<label>`.

eatyourgreens: Grommet tabs are implemented as `<button>` tags. `div`, `ul`, `p`, headings etc. are invalid content for HTML buttons.

shaunanoordin: `<span>` makes sense. Making this change now.

shaunanoordin:
## PR Update

The Tab/Tabs tests and Story have been updated with more accessible examples.

- The previous `🔑 Special` title (read as "Key Special") is now a slightly more sensible(?) `🔑 Members Only` (read as "Restricted Section: Members Only"
- Complex title examples now use `<span>` instead of `<div>` in the examples, but there's no mechanism preventing devs from putting wacky `<ul>`s in their implementation.

PR #3733, opened 2022-10-04
Maintenance Week: add tests for ZooniverseLogo, ZooniverseLogoType

shaunanoordin:
## PR Overview

Package: `lib-react-components`
Part of: Maintenance Week Oct 2022

Adds React Testing Library tests for the ZooniverseLogo and ZooniverseLogotype components

PR #3732, opened 2022-10-04
ZooFooter React Testing Library tests

eatyourgreens:
- add tests for the Zooniverse logotype and tag line.
- delete Enzyme tests from all subcomponents.

## Package
lib-react-components

## How to Review
_Helpful explanations that will make your reviewer happy:_
- What Zooniverse project should my reviewer use to review UX?
- What user actions should my reviewer step through to review this PR?
- Which storybook stories should be reviewed?
- Are there plans for follow up PR’s to further fix this bug or develop this feature?

# Checklist
_PR Creator - Please cater the checklist to fit the review needed for your code changes._
_PR Reviewer - Use the checklist during your review. Each point should be checkmarked or discussed before PR approval._

## General
- [ ] Tests are passing locally and on Github
- [ ] Documentation is up to date and changelog has been updated if appropriate
- [ ] You can `yarn panic && yarn bootstrap` or `docker-compose up --build` and FEM works as expected
- [ ] FEM works in all major desktop browsers: Firefox, Chrome, Edge, Safari (Use Browserstack account as needed)
- [ ] FEM works in a mobile browser

## Refactoring
- [ ] The PR creator has described the reason for refactoring
- [ ] The refactored component(s) continue to work as expected

PR #3750, opened 2022-10-10
Maintenance Week: add tests for ZooniverseLogo

shaunanoordin:
## PR Overview

Replaces #3733 

This PR adds React Testing Library tests for the ZooniverseLogo component.

- Also, the "Zooniverse Logo" title now has translations hooked up.
- No tests will be added for ZooniverseTypeLogo, since these were removed in #3732

Ready for review.

### DIFF
--- a/packages/lib-react-components/src/ZooniverseLogo/ZooniverseLogo.spec.js
+++ b/packages/lib-react-components/src/ZooniverseLogo/ZooniverseLogo.spec.js
@@ -1,5 +1,5 @@
-import { mount } from 'enzyme'
 import React from 'react'
+import { render, screen } from '@testing-library/react'
 
 import ZooniverseLogo from './ZooniverseLogo'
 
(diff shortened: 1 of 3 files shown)


packages/lib-react-components/src/ZooniverseLogo/ZooniverseLogo.spec.js
```diff
@@ -8,37 +8,36 @@ describe('ZooniverseLogo', function () {
...
+        <ZooniverseLogo id={ID} size={SIZE} />
+      )
+    })
+
+    it('should render without crashing', function () {
+      expect(screen).to.be.ok()
+    })
+
+    it('should set the height and width from the `size` prop', function () {
+      const zooLogo = screen.getByLabelText('Zooniverse Logo')  // Gets the SVG
```
shaunanoordin, line 23: I'm slightly baffled as to why this is rendering "Zooniverse Logo" instead of "ZooniverseLogo.title" in the tests, but it seems to be a difference in how the lib-classifier's components/tests are set up vs lib-react-components's components/tests. (It's that `import '../translations/i18n'` bit setting up the fallback language, innit?)

eatyourgreens: #3761 might change this.

eatyourgreens (approved):

PR #3771, opened 2022-10-12
Add projects to Sentry production releases

eatyourgreens:
The `projects` key is required in order for the action to run.


mcbouslog (approved):
Looks good!

Iteration on #3764 .

PR #3784, opened 2022-10-17
GitHub Actions: Deploy production-release

eatyourgreens:
- add a `deploy_production` workflow, which builds and deploys a production image on pushes to the `production-release` tag.
- delete the Jenkinsfile.


eatyourgreens:
There's probably a very good reason that we don't re-use the Docker staging build with the same commit hash, but I can't remember what it is.

.github/workflows/deploy_production.yml
```diff
@@ -0,0 +1,64 @@
...
+      - production-release
+  workflow_dispatch:
+
+jobs:
+  build_and_push_image:
+    name: Build Production Image
+    runs-on: ubuntu-latest
+    steps:
+    - name: Login to GitHub Container Registry
+      uses: docker/login-action@v1
```
camallen, line 15: maybe switch to v2 of the login action? 

```suggestion
      uses: docker/login-action@v2
```

.github/workflows/deploy_production.yml
```diff
@@ -0,0 +1,64 @@
...
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
```
camallen, line 20: I think you want to use the buildx system in docker (the new one)
```suggestion
    - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@v2

```

camallen (approved):
I think this is good any. Any errors can be iterated on to fix it.

camallen:
> There's probably a very good reason that we don't re-use the Docker staging build with the same commit hash, but I can't remember what it is.

I think historically the env had different values baked into it at build time so it was hard to pull the same image and configure at boot / runtime to use the prod system values. That may have changed and could be worth iterating on but short term as a jenkins replacement i think this is ok as we isolate the images to their own container namespace in GHCR

.github/workflows/deploy_production.yml
eatyourgreens: Sure. I'll update #2531 to try this out with the static deploy.

eatyourgreens:
Yeah, I think the webpack production and staging builds, in NextJS, output different JS bundles and we've had problems, in the past, where scripts 404'ed in a production page because they weren't found on a staging image.

.github/workflows/deploy_production.yml
eatyourgreens: I've started a build on #2531 with BuildX enabled. If that works, I'll add it here too.

zwolf:
Does this action do anything specific to FEM that can't be handled by the shared build/push and deploy_app workflows in ci-cd? If not, why not use them?

eatyourgreens:
I don't know. I copied the `deploy_staging.yml` workflow, but I can take a look at the shared workflows.

eatyourgreens:
If we can use the shared workflows, we should probably update `deploy_branch.yml`, `deploy_staging.yml` and `deploy_production.yml` in a separate PR. They're all pretty much based around the same template.

PR #3842, opened 2022-11-03
Refactor survey task container, tests, and storybooks

### LINKED ISSUE #3745: Survey task filter buttons missing focus styling
## Package
- lib-classifier

## Describe the bug
The survey task filter buttons within the Filter drop button are missing any focused state styling.

## To Reproduce
1. Go to https://frontend.preview.zooniverse.org/projects/markb-panoptes/survey-task-test-project/classify/workflow/2434?env=staging
2. Click on the Filter button
3. Press tab to focus a filter button
4. Note there's no focus styling

## Expected behavior
The survey task filter buttons within the Filter drop button have focused state styling.

### LINKED ISSUE #3816: Refactor survey task container and storybooks
## Package
- lib-classifier

To dos:
- [x] refactor survey task container and component to functional component
- [x] create story for entire survey task
- [ ] refactor remaining stories for latest storybook or remove stories

mcbouslog:
## Package
- lib-classifier

- closes #3745 
- closes #3816 

## Describe your changes
- [x] refactors SurveyTaskContainer from class to functional component
- [x] combines SurveyTaskContainer and SurveyTask components
- [x] add survey task storybook
- [x] refactors any remaining survey task tests with react-testing-library
- [x] removes implementation tests throughout survey task components
- [x] add integration tests to SurveyTask tests

PFE staging survey task for comparison - https://master.pfe-preview.zooniverse.org/projects/markb-panoptes/survey-task-test-project/classify?reload=0&workflow=2434&env=staging

## How to Review
  - http://localhost:6006/?path=/story/tasks-survey--default&globals=locale:en;locales.en:English;locales.test:Test+Language
- Are there plans for follow up PR’s to further fix this bug or develop this feature? yes, see https://github.com/zooniverse/front-end-monorepo/milestone/38

# Checklist

## General

## General UX

## Refactoring

## Post-merge


packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.stories.js
```diff
@@ -0,0 +1,56 @@
...
+    </Box>
+  )
+}
+
+export const Default = Template.bind({})
+Default.args = {
+  task
+}
+
+const taskWithoutCharacteristics = Object.assign({}, task, { characteristics: {} })
```
eatyourgreens, line 52:
```suggestion
const taskWithoutCharacteristics = { …task, characteristics: {} }
```

This might be easier to read with a spread operator.

mcbouslog:
The noted line changes are a lot! 👎  I'd be happy to split this up somehow, completely open to suggestion.

The majority of changes are the deletion of enzyme and related implementation tests.
SurveyTask.js, SurveyTask.spec.js, and SurveyTask.stories.js are the core changes.

I've tabbed through what I think are common actions and tried with VoiceOver, but there are a lot of pieces and actions within a survey task, apologies if and please let me know if I've missed anything.

The integration tests in SurveyTask.spec.js take a long time to process. A users actions within a survey task are complex and multi-stepped, so to some degree I think the complexity is inherent, but I'm sure there are refactors that could improve the process time. I'd like to avoid query edits like `getByText` instead of `getByRole`, but this might be a use case where that is preferred.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/CharacteristicsFilter/components/FilterButton.js
```diff
@@ -55,6 +55,7 @@ export default function FilterButton (props) {
       />
       {checked && (
         <CloseButton
+          aria-label={`Remove ${valueLabel} filter`}
```
mcbouslog, line 58: Should this have a translation string? 
If so, I'm thinking it can be added in a subsequent PR, but I'm not sure.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -1,51 +1,566 @@
...
+        // the following answers "How many?" with "1" and "What behavior do you see?" with "Eating"
+        
+        // tabbing (x3) to the "How many?" question, selecting the "1" answer with space key, tabbing (x4) to the "What behavior do you see?" question, selecting the "Eating" answer with space key
+        await user.keyboard('[Tab][Tab][Tab][Space][Tab][Tab][Tab][Tab][Space]')
+        // confirm the Identify button is enabled, now that required questions are answered
+        identifyButton = getByRole('button', { name: 'SurveyTask.Choice.identify' })
+        // confirm the Identify button is enabled, now that required questions answered
+        expect(identifyButton.disabled).to.be.false()
+      })
     })
```
mcbouslog, line 564: I think I need to add a test that deletes an existing identified choice. Realized thanks to https://coveralls.io/builds/54306361.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.js
```diff
@@ -1,24 +1,91 @@
...
+  const [previousChoice, setPreviousChoice] = React.useState('')
+  const [selectedChoice, setSelectedChoice] = React.useState('')
+
+  useEffect(() => {
+    return () => {
+      if (annotation._choiceInProgress) {
+        annotation.setChoiceInProgress(false)
+      }
+    }
+  })
```
eatyourgreens, line 25: This sets `annotation._choiceInProgress` to false on every change of props or state, if it's true. You could replace it with the following if block, but I'm not really sure what this block is doing.

```suggestion
  if (annotation._choiceInProgress) {
    annotation.setChoiceInProgress(false)
  }
```

eatyourgreens: It seems like `annotation._choiceInProgress` can never be true. What is it used for?

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.js
```diff
@@ -1,24 +1,91 @@
...
+      delete newFilters[characteristicId]
+    } else {
+      newFilters = {}
+    }
+
+    setFilters(newFilters)
+  }
+
+  function handleIdentify () {
+    const parsedFilters = JSON.parse(JSON.stringify(filters))
```
eatyourgreens, line 75: Seems kind of slow to repeatedly parse the filters if they've already been parsed. Why does this convert them to a string then parse them again for each identification?

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.js
```diff
@@ -72,4 +130,4 @@ SurveyTask.propTypes = {
   }).isRequired
 }
```
eatyourgreens, line 131: Maybe document the task props with docgen comments?

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.js
mcbouslog: Great catch! `choiceInProgress` is used to prevent clicking Done (and related) while a choice is selected. Here's [a link to the survey task model](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/plugins/tasks/survey/models/SurveyTask.js#L58) that uses `choiceInProgress` within the `isComplete` view.

I just added a commit with a related failing test and a commit to edit the survey task component useEffect with a dependency on the task prop to fix the failing test.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
mcbouslog: Added in [[b880fa0](https://github.com/zooniverse/front-end-monorepo/pull/3842/commits/b880fa00990b557f216c9933d400db6d842354cf)](https://github.com/zooniverse/front-end-monorepo/pull/3842/commits/b880fa00990b557f216c9933d400db6d842354cf)

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/CharacteristicsFilter/components/FilterButton.js
eatyourgreens: My test for this: what would a French volunteer expect this button to say? Here, I’d expect the label to be French.

eatyourgreens: We do have survey projects in Spanish, if you want to test translations. 

Chimp & See is probably available in German too.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.js
eatyourgreens: Thanks. That makes sense.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.js
mcbouslog: Great, question, I'm not sure. I tried searching for a past explanation, and the best I can tell is that I copied it [from PFE](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/survey/index.jsx#L110) and maybe it was creating a copy of the `filters` object for saving to the annotation value, though I'm not sure why. However, `answers` works similarly without any such copying, and I tried filtering on different characteristics, making multiple identifications, changing filters, and I see no reason for the JSON methods (everything works as expected without the JSON methods).

I've removed the `JSON.parse` and `JSON.stringify` in 35fbca34a3523776ef20c6fc40d28312ec63e5ad.

eatyourgreens: Deja vu!
https://github.com/zooniverse/Panoptes-Front-End/pull/5076/files#r234709160

eatyourgreens: You’re right, it’s cloning the `filters` object.

eatyourgreens: If the `filters` prop does need to be cloned, that can be done during the render, wrapped in a `useMemo` hook to memo-ise the copied object.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.js
mcbouslog: I attempted to, could not figure out why they weren't showing, and it turned out it was because I was importing the survey task via plugins and to generate docgen comments I needed to import the survey task component colocated with story 🤦‍♂️ , so this change fixed it 👍 :
```
import SurveyTask from '@plugins/tasks/survey'

import SurveyTask from './SurveyTask'
```

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.js
mcbouslog: I can't think of how the `filters` object would have a value that references an object, a filter characteristic can only be only be one thing, i.e. color can only be red, so I think the deep clone (with JSON methods) is not necessary.

After testing things out, now I also don't think copying the `filters` object is necessary at all. I created a workflow with two survey tasks - https://local.zooniverse.org:8080/?project=1634&workflow=3719 - and I don't see any issues with filters and answers from one task to the other and the annotations look as expected.

So I think it's ok not to copy the `filters` object before adding it to the `annotation.value` (how code is currently), but please let me know if you can think of anything I'm missing.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/CharacteristicsFilter/components/FilterButton.js
mcbouslog: Yes I agree, thanks for comments and references!
It was a small change so I added with [551b04d](https://github.com/zooniverse/front-end-monorepo/pull/3842/commits/551b04d2b41f57554c82261bf6eaf075e01cb71b).
I'll follow [the related translations docs/steps](https://github.com/zooniverse/how-to-zooniverse/blob/master/Translations/lokalise.md#pulling-new-dictionary-keys-into-lokalise-from-github) after merge.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.js
eatyourgreens: Seems good to me, and we can catch any changes with tests that check the annotation value after an identification.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/CharacteristicsFilter/components/FilterButton.js
mcbouslog: Fixing related tests atm...

eatyourgreens (approved):
New tests are passing and the storybook looks good. Now that you can use the full task in the storybook, I must admit that I found the filters hard to use. I couldn't figure out how to apply them and return to the list of choices.

I see what you mean about the tests running slowly. The `user-event` library does try to simulate real usage, so an interaction that takes a couple of seconds in a browser should take a couple of seconds to run. Having said that, a couple of tests took more than 10s to run for me. I think maybe merge this PR, then iterate on test speed later.

The tests for clicking the filters are warning about a memory leak in the characteristics section, so that might be worth looking into later.

```
with user keystrokes
      ✔ should remove a previously identified choice with delete key (8736ms)
      ✔ should remove a previously identified choice with backspace key (8065ms)
      when the Filter button is keyed with Enter
        ✔ should show characteristic filter sections (3619ms)
        when filters are keyed
          ✔ should show the filter button with a remove filter button (2302ms)
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    at /Users/odonnell/zooniverse/front-end-monorepo/node_modules/grommet/components/RadioButtonGroup/RadioButtonGroup.js:20:23
    at div
    at I (/Users/odonnell/zooniverse/front-end-monorepo/node_modules/styled-components/dist/styled-components.cjs.js:1:19189)
    at /Users/odonnell/zooniverse/front-end-monorepo/node_modules/grommet/components/Box/Box.js:21:24
    at CharacteristicSection (/Users/odonnell/zooniverse/front-end-monorepo/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/CharacteristicsFilter/Characteristics/components/CharacteristicSection.js:18:34)
    at div
    at I (/Users/odonnell/zooniverse/front-end-monorepo/node_modules/styled-components/dist/styled-components.cjs.js:1:19189)
    at /Users/odonnell/zooniverse/front-end-monorepo/node_modules/grommet/components/Box/Box.js:21:24
    at observerComponent (/Users/odonnell/zooniverse/front-end-monorepo/node_modules/mobx-react-lite/dist/mobxreactlite.cjs.development.js:445:12)
    at div
```

EDIT: according to that stack trace, the memory leak warning is coming from the `onFilter` callback in `CharacteristicsSection`, when it’s called by the radio button group’s `onChange` handler. It’s called from both `RadioButtonGroup` and `FilterButton` so that might be a bug, if it’s running twice.

eatyourgreens:
The storybook does show a styling bug in the filters. If you look at the icons here, you can see that they are cropped horizontally. Possibly they're styled with `cover` when they should use `contain`? Something like that, maybe?

Screenshot of the 'Like' filters. Some of the silhouettes of the animals have their heads and tails cropped.

mcbouslog:
I'm stuck on the memory leak warning. I feel like the following tests are almost the same other than that the second tests `userEvent`s are with `keyboard`:

test 1 (passing, no memory leak):
```
it('should show the choices that match the filter', async function () {
    const user = userEvent.setup({ delay: null })
    const { getAllByRole, getByRole, getByTestId } = render(<DefaultStory />)
    const filterButton = getByRole('button', { name: 'SurveyTask.CharacteristicsFilter.filter' })
    await user.click(filterButton)
    const redFilterButton = getByTestId('filter-CLR-RD')
    expect(redFilterButton).to.be.ok()

    // the following user events filter choices by the color red and then close the characteristics filters, returning to the list of choices.

    await user.click(redFilterButton)
    await user.click(filterButton)
    const choiceButtons = getAllByRole('menuitemcheckbox')
    expect(choiceButtons.length).to.equal(3)
    expect(choiceButtons[0]).to.have.text('Aardvark')
    expect(choiceButtons[1]).to.have.text('Kudu')
    expect(choiceButtons[2]).to.have.text('Fire')
  })
  ```
  
  test 2 (passing, but with memory leak):
  ```
  it('should show the choices that match the filter', async function () {
    const user = userEvent.setup({ delay: null })
    const { getAllByRole, getByRole } = render(<DefaultStory />)
    const filterButton = getByRole('button', { name: 'SurveyTask.CharacteristicsFilter.filter' })
    // tabbing to and opening the Filter button
    await user.keyboard('[Tab][Enter]')
    // tabbing to the pattern section that contains the solid filter and selecting the solid filter with space key
    await user.keyboard('[Tab][Tab][Space]')
    // close the filters
    await user.click(filterButton)
    // confirming that the choices are filtered by the solid filter
    const choiceButtons = getAllByRole('menuitemcheckbox')
    expect(choiceButtons.length).to.equal(3)
    expect(choiceButtons[0]).to.have.text('Aardvark')
    expect(choiceButtons[1]).to.have.text('Elephant')
    expect(choiceButtons[2]).to.have.text('Kudu')
  })
  ```
  
The only time `onFilter` is called within the FilterButton is when the Close button is clicked to remove a filter, and includes `event.preventDefault()` and `event.stopPropagation()` with the intent to stop the `onFilter` callback from bubbling up to the the CharacteristicsSection radiogroup onChange. I think the `onFilter` in the FilterButton Close button isn't (shouldn't be) called in the noted test above.

There are somewhat related testing-library userEvent issues with radio inputs:
- testing-library / user-event/issues/1048
- testing-library / user-event/issues/786

There are 3 `userEvent.keyboard` related tests that are causing the memory leak warning:
- should show the choices that match the filter
- should remove the filter on remove filter button keypress (within Characteristics)
- should remove the filter on remove filter button keypress (within FilterStatus)

Or maybe this is related to the state of the FilterStatus Grommet DropButton which is set from the Chooser component, but why it's an issue for the keyboard tests on not the click tests makes me think this is not what's causing the memory leak.

I'm not sure what to do about them. At the moment, I lean towards removing the tests and opening a related issue to revisit and re-add?

**Comment by eatyourgreens** [comment]
The stack trace says that state is being set on an unmounted component in the radio button group. My guess would be that the `onChange` handler is running after the component has unmounted.

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    at /Users/odonnell/zooniverse/front-end-monorepo/node_modules/grommet/components/RadioButtonGroup/RadioButtonGroup.js:20:23
```

If memory is leaking during the tests, that will slow the tests down but fixing it isn't urgent (in my opinion.) The question is, have the tests identified a bug in the filters when they're run in a browser?

eatyourgreens:
>The only time onFilter is called within the FilterButton is when the Close button is clicked to remove a filter, and includes event.preventDefault() and event.stopPropagation() with the intent to stop the onFilter callback from bubbling up to the the CharacteristicsSection radiogroup onChange. I think the onFilter in the FilterButton Close button isn't (shouldn't be) called in the noted test above.

There’s a couple of possible problems here:
- Radio inputs fire change events, not click events. 
- `FilterButton` has clickable behaviour (the close button) nested inside each filter button. You can’t nest clickable behaviour like that. For example, you couldn’t nest a navigation link inside a button, or nest a button inside another button. That’s invalid HTML and should be caught by a validation check. 

Neither of these are problems with the changes here, which can be merged. I think the new tests have surfaced a bug that wasn’t being caught before. 

TBH, I think the close button is being misused here. The logic for the selected filter can all be handled by the radio button group’s change handler. 

This PR should stay focused on the tests and stories, with another PR to fix the filter bug.

eatyourgreens:
There's another usability bug, not related to the changes here but obvious now that the task story exists.

Filters can't be deselected from the keyboard. Tab into a filter, change the active radio button with the arrow keys then try to toggle the new value by pressing enter. I can't find a key that will turn the selected value off. Radio buttons can't be unchecked, except by selecting a different button, so this is expected behaviour for radio buttons. However, there's a passing test that says the Enter key removes the active filter.

eatyourgreens:
I've opened https://github.com/zooniverse/front-end-monorepo/issues/3919 for the broken filter button labels.

eatyourgreens:
I opened #3918 to fix the filter icons and refactor the callbacks to use `useCallback`. It still has the memory leak warning. When you delete a filter from the `FilterStatus` component, a couple of things happen:
- The stored task updates to remove the filter.
- The radio button group, which is connected to the stored task, unmounts, because the `Drop` component closes.

I think the error is happening because `FilterStatus` indirectly controls the state of the radio button group, via the stored task. I would guess that the Grommet component has a `setValue` call which is trying to run after the filters close, because the selected filter has changed.

mcbouslog:
Agreed regarding the Close button misuse, thank you for opening #3919 !

I think [the comment](https://github.com/zooniverse/front-end-monorepo/pull/3842#issuecomment-1328941910) noting how `FilterStatus` indirectly controls the state of the radio button group is spot on. I'll open a related issue noting the tests with the memory leaks. I tried refactoring RadioButtonGroup with `defaultValue`, but it didn't fix the memory leak, from a quick initial try anyway.

I think I'll merge this as is for now, to get it into staging for further testing, and to iterate on with smaller PRs.

Issue #3919, opened 2022-11-26
Survey task filter radio button labels use nested buttons


eatyourgreens:
## Package
lib-classifier

## Describe the bug
The label for each filter radio button includes a close button when the radio button is checked. `<label>` tags can only contain inline text and images. Nested `<button>` tags are invalid and raise an accessibility issue. With one clickable element nested inside another clickable element, the label's behaviour is not obvious when clicked.
https://github.com/zooniverse/front-end-monorepo/blob/699ef772c3a25a0bed9b51072898f0b4ad557eed/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/CharacteristicsFilter/components/FilterButton.js#L56-L65

The radio button's label is also confusing, since it contains the close button label.
<img width="707" alt="Radio button label saying: antelope/deer Remove antelope/deer filter." src="https://user-images.githubusercontent.com/59547/204101081-a5dc5305-80e5-42e1-a0f0-eb57e5ce0ce8.png">

## To Reproduce
Select a filter then inspect the radio button label in the DOM.

## Expected behavior
Labels should use valid HTML and not generate accessibility errors.

mcbouslog:
I think the close / remove button should be removed.

With the desired UI of selecting or deselecting one item, and only ever one item at a time, from a group of items, what's the best implementation?

Maybe a checkbox group with that only allows one item to be checked?
Maybe a radio button group with onClick that unchecks checked inputs?
Something else?

eatyourgreens:
>Maybe a checkbox group with that only allows one item to be checked?

This is basically re-inventing radio buttons, albeit with the option to turn one off. Checkboxes have a completely different keyboard interaction from radio buttons, so the design would have to clarify, visually, that the interaction is different.

eatyourgreens:
We could also allow multiple selection. See, for example, the filtered search on https://www.johnlewis.com/search?search-term=coffee

eatyourgreens:
Radio buttons are a good choice when only one answer can be selected but always provide an extra option eg. Male/Female/Prefer not to answer.

eatyourgreens:
The [GDS guidelines for radio buttons](https://design-system.service.gov.uk/components/radios/) are worth reading.

mcbouslog:
[...]

PR #3933, opened 2022-11-30
Classifier: Refactor survey task tests

mcbouslog:
## Package
- lib-classifier

## Describe your changes
- refactor survey task tests for when choice with required questions selected
- refactor some (not all, left at least one test with `getByRole` for related component) tests with `getByText` instead of `getByRole`
- improves `SurveyTask.spec.js` tests runtime by approximately 4 seconds

## How to Review

# Checklist

## General

## Refactoring

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
+++ b/packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
@@ -178,7 +178,7 @@ describe('SurveyTask', function () {
           const fireChoiceButton = screen.getByLabelText('Fire')
           // select the Fire choice
           await user.click(fireChoiceButton)
-          const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Fire')
+          const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
           // identify the Fire choice
           await user.click(identifyButton)
           // confirm the remaining choices are the 3 choices that match the red filter
@@ -496,7 +496,7 @@ describe('SurveyTask', function () {
         expect(choiceHeading).to.be.ok()
         // tabbing to the "Not this" button
         await user.keyboard('[Tab][Tab]')
-        const notThisButton = screen.getByLabelText('SurveyTask.Choice.notThis Nothing here')
+        const notThisButton = screen.getByLabelText('SurveyTask.Choice.notThis')
         expect(notThisButton).to.equal(document.activeElement)
         // pressing Enter to close the choice (Nothing here)
         await user.keyboard('[Enter]')
@@ -520,7 +520,7 @@ describe('SurveyTask', function () {
         expect(choiceHeading).to.be.ok()
         // tabbing to the "Identify" button
         await user.keyboard('[Tab][Tab][Tab]')
-        const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Nothing here')
+        const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
         expect(identifyButton).to.equal(document.activeElement)
         // pressing Enter to identify and close the choice (Nothing here)
         await user.keyboard('[Enter]')
@@ -541,7 +541,7 @@ describe('SurveyTask', function () {
         render(<NoFiltersStory />)
         // tabbing to the first choice (Aardvark) and pressing Enter to open the choice (Aardvark)
         await user.keyboard('[Tab][Enter]')
-        let identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Aardvark')
+        let identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
         // confirm the Identify button is disabled, pending required questions answered
         expect(identifyButton.disabled).to.be.true()
         // the required questions for Aardvark are "How many?" and "What behavior do you see?"
@@ -550,7 +550,7 @@ describe('SurveyTask', function () {
         // tabbing (x3) to the "How many?" question, selecting the "1" answer with space key, tabbing (x4) to the "What behavior do you see?" question, selecting the "Eating" answer with space key
         await user.keyboard('[Tab][Tab][Tab][Space][Tab][Tab][Tab][Tab][Space]')
         // confirm the Identify button is enabled, now that required questions are answered
-        identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Aardvark')
+        identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
         // confirm the Identify button is enabled, now that required questions answered
         expect(identifyButton.disabled).to.be.false()
       })
@@ -561,7 +561,7 @@ describe('SurveyTask', function () {
       render(<NoFiltersStory />)
       let choiceButton = screen.getByLabelText('Fire')
       await user.click(choiceButton)
-      const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Fire')
+      const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
       // identify choice (Fire) and close choice (Fire) component
       await user.click(identifyButton)
       // confirm choices showing
@@ -588,7 +588,7 @@ describe('SurveyTask', function () {
       render(<NoFiltersStory />)
       let choiceButton = screen.getByLabelText('Fire')
       await user.click(choiceButton)
-      const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Fire')
+      const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
       // identify choice (Fire) and close choice (Fire) component
       await user.click(identifyButton)
       // confirm choices showing
--- a/packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
+++ b/packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
@@ -343,25 +343,6 @@ describe('SurveyTask', function () {
         expect(fireChoiceButton.getAttribute('aria-checked')).to.equal('true')
       })
 
-      it('should disable the Identify button until required questions are answered', async function () {
-        const user = userEvent.setup({ delay: null })
-        const { getByRole } = render(<DefaultStory />)
-        const choiceButton = getByRole('menuitemcheckbox', { name: 'Aardvark' })
-        await user.click(choiceButton)
-        let identifyButton = getByRole('button', { name: 'SurveyTask.Choice.identify' })
-        // confirm the Identify button is disabled, pending required questions answered
-        expect(identifyButton.disabled).to.be.true()
-        // the required questions for Aardvark are "How many?" and "What behavior do you see?"
-        // the following answers "How many?" with "3" and "What behavior do you see?" with "Eating"
-        const howManyInput = getByRole('radio', { name: "3" })
-        await user.click(howManyInput)
-        const whatBehaviorsInput = getByRole('checkbox', { name: "Eating" })
-        await user.click(whatBehaviorsInput)
-        identifyButton = getByRole('button', { name: 'SurveyTask.Choice.identify' })
-        // confirm the Identify button is enabled, now that required questions answered
-        expect(identifyButton.disabled).to.be.false()
-      })
-
       it('should disable "Done & Talk" and "Done" buttons', async function () {
         const user = userEvent.setup({ delay: null })
         const { getByRole } = render(<DefaultStory />)
--- a/packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
+++ b/packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
@@ -67,9 +67,9 @@ describe('SurveyTask', function () {
       const NoFiltersStory = composeStory(NoFilters, Meta)
       
       it('should not show a filter button', function () {
-        const { queryByRole } = render(<NoFiltersStory />)
+        const { queryByText } = render(<NoFiltersStory />)
         
-        const filterButton = queryByRole('button', { name: 'SurveyTask.CharacteristicsFilter.filter' })
+        const filterButton = queryByText('SurveyTask.CharacteristicsFilter.filter')
         
         expect(filterButton).to.not.exist()
       })
@@ -113,9 +113,9 @@ describe('SurveyTask', function () {
     describe('when the Filter button is clicked', function () {
       it('should show characteristic filter sections', async function () {
         const user = userEvent.setup({ delay: null })
-        const { getAllByRole, getByRole } = render(<DefaultStory />)
+        const { getAllByRole, getByText } = render(<DefaultStory />)
         // filterButton is the Filter button above the choices
-        const filterButton = getByRole('button', { name: 'SurveyTask.CharacteristicsFilter.filter' })
+        const filterButton = getByText('SurveyTask.CharacteristicsFilter.filter')
         await user.click(filterButton)
         // the filterSections are the characteristic filter sections, i.e. the sections for "Like", "Pattern", and "Color" for the mock task
         const filterSections = getAllByRole('radiogroup')
@@ -126,8 +126,8 @@ describe('SurveyTask', function () {
       describe('when filters are clicked', function () {
         it('should show the filter button with a remove filter button', async function () {
           const user = userEvent.setup({ delay: null })
-          const { getByRole, getByTestId } = render(<DefaultStory />)
-          const filterButton = getByRole('button', { name: 'SurveyTask.CharacteristicsFilter.filter' })
+          const { getByTestId, getByText } = render(<DefaultStory />)
+          const filterButton = getByText('SurveyTask.CharacteristicsFilter.filter')
           await user.click(filterButton)
           // the stripesFilterButton is the button to filter choices by "stripes". Stripes is a specific value of the "Pattern" characteristic.
           const stripesFilterButton = getByTestId('filter-PTTRN-STRPS')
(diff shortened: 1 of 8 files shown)


packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -214,8 +210,8 @@ describe('SurveyTask', function () {
 
         it('should remove the filter on remove filter button click (within FilterStatus)', async function () {
           const user = userEvent.setup({ delay: null })
-          const { getAllByRole, getByRole, getByTestId } = render(<DefaultStory />)
-          const filterButton = getByRole('button', { name: 'SurveyTask.CharacteristicsFilter.filter' })
+          const { getAllByRole, getByTestId, getByText } = render(<DefaultStory />)
+          const filterButton = getByText('SurveyTask.CharacteristicsFilter.filter')
```
eatyourgreens, line 214: Shouldn’t this be `screen.getByLabelText`?

eatyourgreens (commented):
Is there a reason that you’re not using `screen` to run queries? 

Buttons and other labelled elements should use `getByLabelText` rather than `getByText`.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
eatyourgreens: I think that applies to all the buttons in the tests.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -264,8 +260,8 @@ describe('SurveyTask', function () {
 
         it('should remove filters on Clear Filters button click (within Chooser)', async function () {
           const user = userEvent.setup({ delay: null })
-          const { getAllByRole, getByRole, getByTestId } = render(<DefaultStory />)
-          const filterButton = getByRole('button', { name: 'SurveyTask.CharacteristicsFilter.filter' })
+          const { getAllByRole, getByTestId, getByText } = render(<DefaultStory />)
+          const filterButton = getByText('SurveyTask.CharacteristicsFilter.filter')
           await user.click(filterButton)
           const cowHorseFilterButton = getByTestId('filter-LK-CWHRS')
```
eatyourgreens, line 266: This could use `screen.getByLabelText` too, since each icon should also have a text label.

eatyourgreens:
A couple of other thoughts:

- `within` can help to narrow the scope of a query. 
- you also have APIs like `querySelector` and `querySelectorAll` to find DOM nodes for testing. 
- you can use `beforeEach` to reduce repetition in test blocks.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/CharacteristicsFilter/ClearFilters/ClearFilters.js
```diff
@@ -27,6 +27,7 @@ export default function ClearFilters (props) {
         {t('SurveyTask.CharacteristicsFilter.showing', { showing: showingChoices, total: totalChoices })}
       </SpacedText>
       <PlainButton
+        a11yTitle={t('SurveyTask.CharacteristicsFilter.clearFilters')}
```
eatyourgreens, line 30: This button has a text label already. 
```suggestion
```

packages/lib-classifier/src/plugins/tasks/survey/components/components/Choice/Choice.js
```diff
@@ -93,11 +93,13 @@ export default function Choice({
         pad={{ top: 'small' }}
       >
         <Button
+          a11yTitle={`${t('SurveyTask.Choice.notThis')} ${choice.label}`}
```
eatyourgreens, line 96: Duplicated label. 
```suggestion
```

packages/lib-classifier/src/plugins/tasks/survey/components/components/Choice/Choice.js
```diff
@@ -93,11 +93,13 @@ export default function Choice({
...
           fill='horizontal'
           label={t('SurveyTask.Choice.notThis')}
           onClick={() => handleDelete(choiceId)}
         />
         <PrimaryButton
+          a11yTitle={`${t('SurveyTask.Choice.identify')} ${choice.label}`}
```
eatyourgreens, line 102: Duplicated label. 
```suggestion
```

eatyourgreens (changes requested):
Don’t use ARIA labels to duplicate button labels.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -15,19 +15,20 @@ describe('SurveyTask', function () {
...
+        const filterButton = screen.getByLabelText('SurveyTask.CharacteristicsFilter.filter')
         
         expect(filterButton).to.be.ok()
       })
 
       it('should show the choices', function () {
-        const { getAllByRole } = render(<DefaultStory />)
+        render(<DefaultStory />)
         
+        const choicesMenu = screen.getByTestId('choices-menu')
```
eatyourgreens, line 29: I think there's only one menu, so this can be:
```suggestion
        const choicesMenu = document.querySelector([role=menu])
```

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -114,174 +116,158 @@ describe('SurveyTask', function () {
...
+          render(<DefaultStory />)
+          // click the Filter button above choices to open the characteristics filters
+          filterButton = screen.getByLabelText('SurveyTask.CharacteristicsFilter.filter')
           await user.click(filterButton)
+        })
+        
+        it('should show the filter button with a remove filter button', async function () {
           // the stripesFilterButton is the button to filter choices by "stripes". Stripes is a specific value of the "Pattern" characteristic.
-          const stripesFilterButton = getByTestId('filter-PTTRN-STRPS')
+          const stripesFilterButton = screen.getByTestId('filter-PTTRN-STRPS')
```
eatyourgreens, line 143: Maybe get these buttons by their label text, since they all have text labels.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -114,174 +116,158 @@ describe('SurveyTask', function () {
...
-          const fireChoiceButton = getByRole('menuitemcheckbox', { name: 'Fire' })
-          expect(fireChoiceButton).to.be.ok()
-
+          const fireChoiceButton = screen.getByLabelText('Fire')
           // select the Fire choice
           await user.click(fireChoiceButton)
-          const identifyButton = getByRole('button', { name: 'SurveyTask.Choice.identify' })
-          expect(identifyButton).to.be.ok()
-
+          const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Fire')
```
eatyourgreens, line 182:
```suggestion
          const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
```
Why has the choice been added to the button label?

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -500,82 +465,84 @@ describe('SurveyTask', function () {
...
         await user.keyboard('[Tab]')
         await user.keyboard('[ArrowUp]')
         await user.keyboard('[Enter]')
-        let choiceHeading = getByRole('heading', { name: 'Nothing here' })
+        let choiceHeading = screen.getByRole('heading', { name: 'Nothing here' })
         expect(choiceHeading).to.be.ok()
         // tabbing to the "Not this" button
         await user.keyboard('[Tab][Tab]')
-        const notThisButton = getByRole('button', { name: 'SurveyTask.Choice.notThis' })
+        const notThisButton = screen.getByLabelText('SurveyTask.Choice.notThis Nothing here')
```
eatyourgreens, line 500:
```suggestion
        const notThisButton = screen.getByLabelText('SurveyTask.Choice.notThis')
```

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -500,82 +465,84 @@ describe('SurveyTask', function () {
...
         await user.keyboard('[Tab]')
         await user.keyboard('[ArrowUp]')
         await user.keyboard('[Enter]')
-        let choiceHeading = getByRole('heading', { name: 'Nothing here' })
+        let choiceHeading = screen.getByRole('heading', { name: 'Nothing here' })
         expect(choiceHeading).to.be.ok()
         // tabbing to the "Identify" button
         await user.keyboard('[Tab][Tab][Tab]')
-        const identifyButton = getByRole('button', { name: 'SurveyTask.Choice.identify' })
+        const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Nothing here')
```
eatyourgreens, line 524:
```suggestion
        const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
```

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -500,82 +465,84 @@ describe('SurveyTask', function () {
...
       })
 
       it('should disable the Identify button until required questions are answered', async function () {
         const user = userEvent.setup({ delay: null })
-        const { getByRole } = render(<NoFiltersStory />)
+        render(<NoFiltersStory />)
         // tabbing to the first choice (Aardvark) and pressing Enter to open the choice (Aardvark)
         await user.keyboard('[Tab][Enter]')
-        let identifyButton = getByRole('button', { name: 'SurveyTask.Choice.identify' })
+        let identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Aardvark')
```
eatyourgreens, line 545:
```suggestion
        let identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
```

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -584,26 +551,27 @@ describe('SurveyTask', function () {
         // tabbing (x3) to the "How many?" question, selecting the "1" answer with space key, tabbing (x4) to the "What behavior do you see?" question, selecting the "Eating" answer with space key
         await user.keyboard('[Tab][Tab][Tab][Space][Tab][Tab][Tab][Tab][Space]')
         // confirm the Identify button is enabled, now that required questions are answered
-        identifyButton = getByRole('button', { name: 'SurveyTask.Choice.identify' })
+        identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Aardvark')
```
eatyourgreens, line 554:
```suggestion
        identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
```

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -584,26 +551,27 @@ describe('SurveyTask', function () {
...
 
     it('should remove a previously identified choice with delete key', async function () {
       const user = userEvent.setup({ delay: null })
-      const { getAllByRole, getByRole } = render(<NoFiltersStory />)
-      let choiceButton = getByRole('menuitemcheckbox', { name: 'Fire' })
+      render(<NoFiltersStory />)
+      let choiceButton = screen.getByLabelText('Fire')
       await user.click(choiceButton)
-      const identifyButton = getByRole('button', { name: 'SurveyTask.Choice.identify' })
+      const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Fire')
```
eatyourgreens, line 565:
```suggestion
      const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
```

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
```diff
@@ -612,24 +580,25 @@ describe('SurveyTask', function () {
...
 
     it('should remove a previously identified choice with backspace key', async function () {
       const user = userEvent.setup({ delay: null })
-      const { getAllByRole, getByRole } = render(<NoFiltersStory />)
-      let choiceButton = getByRole('menuitemcheckbox', { name: 'Fire' })
+      render(<NoFiltersStory />)
+      let choiceButton = screen.getByLabelText('Fire')
       await user.click(choiceButton)
-      const identifyButton = getByRole('button', { name: 'SurveyTask.Choice.identify' })
+      const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify Fire')
```
eatyourgreens, line 592:
```suggestion
      const identifyButton = screen.getByLabelText('SurveyTask.Choice.identify')
```

eatyourgreens (changes requested):
Why has the Identify button label been changed to 'Identify Aardark', 'Identify Fire' etc? It should be 'Identify' for any choice.

packages/lib-classifier/src/plugins/tasks/survey/components/SurveyTask.spec.js
eatyourgreens: `querySelector` might be faster. I'm not sure.

mcbouslog:
lib-react-components and Grommet buttons (Primary, Button, etc.) that define the label as a property on the component are not found with `getByLabelText`. If the noted buttons have an `a11yTitle` they are found with `getByLabelText`.

Components that are found with `getByLabelText`:
- [ChoiceButton](https://github.com/zooniverse/front-end-monorepo/blob/c8bc958cd755c91cfc62f50229c6415e7242b0e2/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js#L58-L59)
- [FilterStatus](https://github.com/zooniverse/front-end-monorepo/blob/c8bc958cd755c91cfc62f50229c6415e7242b0e2/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/CharacteristicsFilter/FilterStatus/FilterStatus.js#L69-L70)

Components with defined labels that aren't found with `getByLabelText`:
- [ClearFilters](https://github.com/zooniverse/front-end-monorepo/blob/c8bc958cd755c91cfc62f50229c6415e7242b0e2/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/CharacteristicsFilter/ClearFilters/ClearFilters.js#L29-L34)
- (within [Choice](https://github.com/zooniverse/front-end-monorepo/blob/c8bc958cd755c91cfc62f50229c6415e7242b0e2/packages/lib-classifier/src/plugins/tasks/survey/components/components/Choice/Choice.js#L95-L105)) Not this
- (within [Choice](https://github.com/zooniverse/front-end-monorepo/blob/c8bc958cd755c91cfc62f50229c6415e7242b0e2/packages/lib-classifier/src/plugins/tasks/survey/components/components/Choice/Choice.js#L95-L105)) Identify

eatyourgreens:
Is that because you're looking for eg. `SurveyTask.Choice.identify` but the label is 'Identify'?

eatyourgreens:
No wait, the docs says this:
>The example above does NOT find the input node for label text broken up by elements. 
https://testing-library.com/docs/queries/bylabeltext#name

So those buttons could be failing because they use HTML markup in their labels. [ChoiceButton](https://github.com/zooniverse/front-end-monorepo/blob/c8bc958cd755c91cfc62f50229c6415e7242b0e2/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js#L58-L59) here, for example, has markup in its label, because it uses images.

`ChoiceButton` there, by the way, should not have an ARIA label. A button should either have a text label:
```html
<button>Text label</button>
```
or an ARIA label: 
```html
<button aria-label="Text label"><img /></button>
```
but never both together.

To be honest, in that second case the button label should be image alt text.
```html
<button><img alt="Text label" /></button>
```

mcbouslog:
For buttons with HTML markup in the labels, preventing use of `getByLabelText`, what do you think is the preferred query? I think ideally it'd be 'getByRole('button', { name: 'Aardvark' })`, but that query takes awhile, so maybe for these tests `getByText`?

eatyourgreens:
This query will find all menu item checkboxes inside the first menu on the page. 

```js
querySelector([role=“menu”]).querySelectorAll([role=“menuitem”])
```

If the checkboxes have values, you can further refine the query with `[value=“Aardvark”]`. Otherwise, you could pluck the first item from the returned `NodeList`, or filter the node list by inspecting the text of each element to find a particular button. 

Stepping back from specific tests, is the run time being spent rendering the survey task, or running queries? If it’s rendering excessively, you can cache rendered nodes in variables and reuse them. Have a look at the `Classifier` tests, which do this because they are also slow. The `Classifier` tests are also split into two files, so consider that too.

eatyourgreens (approved):

eatyourgreens:
Reading the source code, label text queries ignore buttons.
https://github.com/testing-library/dom-testing-library/blob/edffb7c5ec2e4afd7f6bedf842c669ddbfb73297/src/queries/label-text.ts#L25

eatyourgreens:
Also noting here that `ChoiceButton`, linked above, wraps a `div` with `button`, which is invalid HTML.

eatyourgreens:
This is running a lot faster now on my laptop, though the excessive rendering is still making it quite slow.

I've rebased and added a commit that renders a story once, in a `before()` block, then tests the output DOM of that story in separate `it()` blocks. You can use that as an example to speed up the other test blocks. There are a lot of tests where you run the same render multiple times. You could run it once, store the output in variables then test each of those variables in `it()` blocks.

I also fixed a bug in the 'Clear FIlters' button, which had the wrong label. text Your tests had caught the bug, but you'd updated the tests to pass by using the broken label, rather than fixing the button so that it passed with the expected label.

eatyourgreens:
CI isn't timing out, so I'm happy for this to be merged as-is.

eatyourgreens:
>The filters can be removed by tabbing to the close/remove button within the filter and pressing space or enter. 

This is a bug. See #3919. Tab on each radio button should take you out of the radio button group and on to the next focusable element on the page.

eatyourgreens:
FWIW, I'm toggling the selected radio button with the cursor keys, then trying Space or Enter to clear the selected value.

PR #3966, opened 2022-12-06
Remove query-string

eatyourgreens:
Remove `query-string` from the dev classifier.

Replace [`query-string`](https://www.npmjs.com/package/query-string) with the native [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) interface.

## Package
lib-classifier

## How to Review
Run the standalone classifier on https://localhost:8080 with `yarn dev`.

# Checklist

## Maintenance

### DIFF
--- a/packages/lib-classifier/dev/index.js
+++ b/packages/lib-classifier/dev/index.js
@@ -1,12 +1,15 @@
-import queryString from 'query-string'
-import React from 'react'
 import ReactDOM from 'react-dom'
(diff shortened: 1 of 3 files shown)


mcbouslog (approved):
`lib-classifier` https://local.zooniverse.org:8080/ continues to work as expected, confirmed staging and production projects/workflows load and query params update as expected 👍

Issue #1302, opened 2019-11-13
Support audio / video examples in survey task


rogerhutchings:
cc #1286 

For projects like SoNYC, we should be able to give audio or video examples for possible selections in the survey task.

mcbouslog:
Staging workflow with audio and video media in (first two) choices - https://frontend.preview.zooniverse.org/projects/markb-panoptes/survey-task-test-project/classify/workflow/3503?env=staging

mcbouslog:
From https://github.com/zooniverse/front-end-monorepo/pull/3989#issuecomment-1344795450 - good suggestions regarding audio/video:
- can’t embed audio or video in a button label, would need an image icon or something
- maybe an audio survey task would have playable sound clips with the menu button under each player

PR #3989, opened 2022-12-09
Refactor survey task choice button labels

eatyourgreens:
Refactor choice button labels to use the Grommet `Image` component and fix invalid HTML inside the `<button>` element.

## Package
lib-classifier

## How to Review
You can check the changes here with the Survey Task story in the storybook. It shouldn't have changed, except that the layout shift when choices first load should now be fixed.

# Checklist

## General

## Refactoring

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
+++ b/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
@@ -69,7 +69,7 @@ function ChoiceButton({
         >
           {thumbnailSize !== 'none' && src &&
             <Image
-              alt={`${choiceLabel} media`}
+              alt=''
               height='fill'
               margin={{ right: '1ch' }}
               src={thumbnailSrc}
--- a/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
+++ b/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
@@ -1,11 +1,12 @@
 import {
   Box,
   Button,
+  Image,
   Text
 } from 'grommet'
 import PropTypes from 'prop-types'
 import { useCallback, useEffect, useRef } from 'react';
-import { Media, withThemeContext } from '@zooniverse/react-components'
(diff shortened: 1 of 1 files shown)


packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
```diff
@@ -59,17 +62,18 @@ function ChoiceButton({
...
         <Box
+          as='span'
           direction='row'
           fill
           align='center'
         >
           {thumbnailSize !== 'none' && src &&
-            <Media
+            <Image
               alt={`${choiceLabel} media`}
```
eatyourgreens, line 72: This text will get added to the button label, so it will say something like 'Wildebeest media Wildebeest.'.

eatyourgreens: Here are the choice button labels for the mock task in the storybook. Easy enough to fix.

The Firefox accessibility inspector open to list the choice menu labels in the storybook.

eatyourgreens: The tests must be missing this bug too.

eatyourgreens:
I removed the `.skip` on this test block, but feel free to add it back if it makes the tests unstable again.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/components/Choices/components/ChoiceButton.js
mcbouslog: Nice catch! I don't have an opinion of which of the two solution.

mcbouslog (approved):
This looks great. I compared storybooks local vs deployed and looks good.

Noting if there's a project that'd like to use audio or video as part of the choice button media the ChoiceButton would have to be refactored (maybe back to use Media), but I'm not aware of any such project and other styling refactors would be required so we can revisit if need be. Related to #1302.

eatyourgreens:
You can’t embed audio or video in a button label so those would need an image icon or something.

eatyourgreens:
Maybe an audio survey task would have playable sound clips with the menu button under each player? Definitely something to revisit.

PR #4107, opened 2023-01-10
Maintenance Week: add tests for Tooltip

shaunanoordin:
## PR Overview

Part of: Maintenance Week

This PR adds React Testing Library tests for the Tooltip component.

- Also, lib-react-component's test setup now sets a stub for `global.cancelAnimationFrame`, which is required for [Tippy,](https://www.npmjs.com/package/tippy.js/v/6.2.1) which is used by Tooltip.

Ready for review


packages/lib-react-components/src/Tooltip/Tooltip.spec.js
```diff
@@ -1,41 +1,31 @@
...
-    expect(renderedLabel.props().arrow).to.be.false()
-    expect(renderedLabel.props().animation).to.equal('zoomIn')
-    expect(props.placement).to.equal('bottom')
-    expect(props.trigger).to.equal('focus')
+  it('should show the tooltip when the pointer hovers over the element', async function () {
+    const user = userEvent.setup({ delay: null })
+
+    await user.hover(screen.queryByText('Help Menu'))
+    expect(screen.queryByText('Click this button to open the help menu')).to.exist()
   })
```
eatyourgreens, line 30: It should also show the tooltip when the button has focus. Can you add a test for that?

eatyourgreens (approved):
Looks good. Is it worth adding a test for the case where you tab to the button?

You could also add a test for the case where the button is clicked. I assume that closes the tooltip, but I'm not sure.

shaunanoordin:
## PR Update

Thanks to Jim for test recommendations:
- the Tooltip appears when the user tab-navigates to the element.
- the Tooltip also appears when the element gains focus _for whatever other reason._ (e.g. programmatically)

Tests have been added for both scenarios 👍

shaunanoordin:
## PR Update Update

Almost forgot about this!

- Tooltip DOES disappear when the element (button) is clicked. Tests have been added for this as well 👍

PR #4122, opened 2023-01-12
Maintenance Week: add tests for Tooltip's sub-components

shaunanoordin:
## PR Overview

Follows #4107 
Package: lib-react-components
Part of: Maintenance Week

This PR adds React Testing Library tests for Tooltip's sub-components. They're pretty basic, honestly.

Ready for review


eatyourgreens (approved):
LGTM. I don't think we need detailed tests for subcomponents like this either. If we have stories for a component, I think it's enough to write a test suite that imports the stories then runs tests on them.

PR #4123, opened 2023-01-12
Maintenance Week: add tests & story for Zooniverse Logo & Logotype

shaunanoordin:
## PR Overview

Follows #3750 
Package: lib-react-components
Part of: Maintenance Week

This PR adds React Testing Library tests for Zooniverse Logotype, and stories for both Zooniverse Logo & Logotype.

Ready for review

### DIFF
--- a/packages/lib-react-components/src/ZooniverseLogo/ZooniverseLogo.stories.js
+++ b/packages/lib-react-components/src/ZooniverseLogo/ZooniverseLogo.stories.js
@@ -8,7 +8,7 @@ export default {
   title: 'Components/ZooniverseLogo',
   component: ZooniverseLogo,
   args: {
-    id: 'The Zooniverse',
+    id: 'the-zooniverse',
     size: '200px',
   },
   parameters: {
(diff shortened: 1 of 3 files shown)


packages/lib-react-components/src/ZooniverseLogo/ZooniverseLogo.stories.js
```diff
@@ -0,0 +1,29 @@
...
+import { Box, Button, Grommet } from 'grommet'
+
+import readme from './README.md'
+import ZooniverseLogo from './ZooniverseLogo'
+
+export default {
+  title: 'Components/ZooniverseLogo',
+  component: ZooniverseLogo,
+  args: {
+    id: 'The Zooniverse',
```
eatyourgreens, line 11: Should this be a HTML element `id`, in which case 'The Zooniverse' is not a valid identifier.

eatyourgreens (approved):
LGTM. `id` might need to be [a valid identifier](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) so double check that.

eatyourgreens (changes requested):
I checked and those IDs should be DOM element identifiers, so no spaces allowed!

shaunanoordin:
Thanks! 👍 For both stories, the `id: 'The Zooniverse'` has now been changed to `id: 'the-zooniverse'`

eatyourgreens (approved):

PR #4124, opened 2023-01-12
Refactor draggable tests and fix draggable console warning

mcbouslog:

## Package
- lib-classifier

## Describe your changes
- [x] fix draggable console warning "Warning: Received `false` for a non-boolean attribute `dragging`."
- [ ] refactor tests to remove `enzyme`

## How to Review

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

## Refactoring

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/components/draggable/draggable.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/components/draggable/draggable.js
@@ -121,7 +121,7 @@ function draggable(WrappedComponent) {
           <WrappedComponent
             ref={this.wrappedComponent}
             {...rest}
-            dragging={dragging}
+            dragging={dragging.toString()}
           >
             {children}
           </WrappedComponent>


packages/lib-classifier/src/plugins/drawingTools/components/draggable/draggable.js
```diff
@@ -121,7 +121,7 @@ function draggable(WrappedComponent) {
           <WrappedComponent
             ref={this.wrappedComponent}
             {...rest}
-            dragging={dragging}
+            dragging={dragging.toString()}
```
eatyourgreens, line 124: This is going to break everywhere else in the code the rely on dragging being either `true` or `false`. Te reason for that is that `'false'` is truthy in JS, so `if (dragging)` will now be true throughout the code even when we're not dragging.

This needs to be changed back and a different solution found.

mcbouslog: Yes I was worried it would break many things as well! I quickly tried the single image viewer (I Fancy Cats) and the LCV (TESS) and dragging mode and annotation mode appeared to work as expected. Is there a specific viewer you think would most like break or user actions I'm missing?

eatyourgreens: I think the solution we need is that the `Draggable` HOC should use a different prop name, or the `dragging` React prop should not be passed down to HTML elements as an element attribute.

mcbouslog: Ah yes I see a new warning as expected with these changes - "Warning: Failed prop type: Invalid prop `dragging` of type `string` supplied to `ForwardRef(observerComponent)`, expected `boolean`.)" when dragging marks.

eatyourgreens: @mcbouslog it will probably break drawing marks styling here:
https://github.com/zooniverse/front-end-monorepo/blob/103fc15dd0428a843d7c1a85a93b4951c9edb67e/packages/lib-classifier/src/plugins/drawingTools/components/Mark/Mark.js#L20-L26

I did a quick GitHub searching for `dragging`, and it looks like drag handles use it too. Maybe individual drawing tools also.

Issue #4136, opened 2023-01-16
Tests logs


mcbouslog:
The CI step for "Coverage report / Run tests" or `yarn test:ci` logs more than can be displayed in the related action log.

This issue for documenting efforts towards improving the tests logs.

As of Jan. 13, 2023, per my local machine lib-classifier tests logged 10,642 lines.

- [ ] [this line](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/subjects/Subject/Subject.spec.js#L60) in Subject.spec.js logs roughly 3,790
- [ ] I think [these tests](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/DoneAndTalkButton/DoneAndTalkButtonConnector.spec.js) in the DoneAndTalkButtonConnector log roughly 882 lines
- [add others here...]

mcbouslog:
PR #4237 refactors Subject.spec.js

PR #4141, opened 2023-01-16
Classifier: Subject.spec - remove store subjects reset

mcbouslog:
## Package
- lib-classifier

- towards #4136 

## Describe your changes
- removes `rootStore.subjects.reset()` in "Subject with a transcription workflow" tests, in related `before` function
- the "with a transcription workflow" tests are causing many lines of "Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree." to be logged
- `panoptes-js` is stubbed to return mock subjects and a new store is created in the lines before this deleted line, I'm not sure why subjects are then reset, so I think the line can safely be deleted
- affected tests continue to pass

## How to Review

# Checklist

## General

## Refactoring


packages/lib-classifier/src/store/subjects/Subject/Subject.spec.js
```diff
@@ -57,7 +57,6 @@ describe('Model > Subject', function () {
         get: sinon.stub().callsFake(() => Promise.resolve({ body: { tutorials: [] } }))
       }
       rootStore = mockStore({ workflow: workflowSnapshot, client })
-      rootStore.subjects.reset()
```
eatyourgreens, line 60: This empties the queue, so the call to `populateQueue` can be removed too. Having said that, I'm not sure why I added these in the first place. Could be a workaround for a bug in the mock store.

packages/lib-classifier/src/store/subjects/Subject/Subject.spec.js
```diff
@@ -57,7 +57,6 @@ describe('Model > Subject', function () {
...
       client.caesar.request.resetHistory()
       await rootStore.subjects.populateQueue()
```
eatyourgreens, line 61: I think this can be changed to a promise that waits for the subject queue to be ready, using `when()`.

packages/lib-classifier/src/store/subjects/Subject/Subject.spec.js
eatyourgreens: Here's an example from the Classifier tests. You can probably adapt it for this test.
https://github.com/zooniverse/front-end-monorepo/blob/1cb860105e320dab4e18b889b800d44416c9feef/packages/lib-classifier/src/components/Classifier/Classifier.workflows.spec.js#L130

eatyourgreens (approved):
Looks good to me. I can't remember why the tests are set up like this, so I'm assuming it was a workaround for a bug in the mock store.

packages/lib-classifier/src/store/subjects/Subject/Subject.spec.js
eatyourgreens: Here's an example from the Classifier tests. You can probably adapt it for this test.
https://github.com/zooniverse/front-end-monorepo/blob/1cb860105e320dab4e18b889b800d44416c9feef/packages/lib-classifier/src/components/Classifier/Classifier.workflows.spec.js#L130

mcbouslog:
Closing in favor of #4237

PR #4238, opened 2023-02-06
Edit next app asset prefixes to bypass CDN on production, go through CDN on staging

mcbouslog:
## Package
- app-project
- app-content-pages

## Linked Issue and/or Talk Post
- iteration on #4225 

## Describe your changes
- while the intermittent CDN issue is ongoing, we'll _temporarily_ route production traffic to bypass the CDN (https://fe-project.zooniverse.org bypasses the CDN), and we think avoid the javascript loading issue
- while the intermittent CDN issue is ongoing, we'll use staging (https://frontend.preview.zooniverse.org/) with the CDN, so we or outside support can test/recreate the ongoing javascript loading issues

## How to Review
_Helpful explanations that will make your reviewer happy:_
- What Zooniverse project should my reviewer use to review UX?
- What user actions should my reviewer step through to review this PR?
- Which storybook stories should be reviewed?
- Are there plans for follow up PR’s to further fix this bug or develop this feature?

# Checklist
_PR Creator - Please cater the checklist to fit the review needed for your code changes._
_PR Reviewer - Use the checklist during your review. Each point should be checkmarked or discussed before PR approval._

## General
- [x] Tests are passing locally and on Github
- [ ] Documentation is up to date and changelog has been updated if appropriate
- [x] You can `yarn panic && yarn bootstrap` or `docker-compose up --build` and FEM works as expected
- [x] FEM works in all major desktop browsers: Firefox, Chrome, Edge, Safari (Use Browserstack account as needed)
- [ ] FEM works in a mobile browser

## General UX
Example Staging Project: [i-fancy-cats](https://local.zooniverse.org:3000/projects/brooke/i-fancy-cats)
- [x] All pages of a FEM project load: Home Page, Classify Page, and About Pages
- [x] Can submit a classification
- [x] Can sign-in and sign-out
- [ ] The component is accessible
  - Can be used with a screen reader [BBC guide to VoiceOver](https://bbc.github.io/accessibility-news-and-you/assistive-technology/testing-steps/voiceover-mac.html)
  - Can be used from the keyboard [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - It is passing accessibility checks in the storybook

## Refactoring
- [x] The PR creator has described the reason for refactoring
- [x] The refactored component(s) continue to work as expected

PR #4240, opened 2023-02-07
Remove redundant asset prefixes

eatyourgreens:
[`assetPrefix`](https://nextjs.org/docs/api-reference/next.config.js/cdn-support-with-asset-prefix) shouldn't be used as a replacement for `basePath`. It should only be set when static assets are hosted on a different domain than the NextJS app.

### DIFF
--- a/packages/app-content-pages/next.config.js
+++ b/packages/app-content-pages/next.config.js
@@ -6,9 +6,6 @@ const path = require('path')
 const withSourceMaps = require('@zeit/next-source-maps')()
 
 const assetPrefixes = {
-  development: '/about',
-  branch: 'https://fe-project-branch.preview.zooniverse.org/about',
-  static: 'https://fe-static.zooniverse.org/about',
   production: 'https://fe-content-pages.zooniverse.org/about/'
 }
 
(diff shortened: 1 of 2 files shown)


packages/app-content-pages/next.config.js
```diff
@@ -5,13 +5,7 @@ const Dotenv = require('dotenv-webpack')
 const path = require('path')
 const withSourceMaps = require('@zeit/next-source-maps')()
 
-const assetPrefixes = {
-  development: '/about',
-  branch: 'https://fe-project-branch.preview.zooniverse.org/about',
-  staging: 'https://frontend.preview.zooniverse.org/about',
-  static: 'https://fe-static.zooniverse.org/about',
-  production : 'https://www.zooniverse.org/about'
```
mcbouslog, line 13: Noting that depending on how this rebases after #4238 merges, and if we're in bypass CDN mode, this line should be `production : 'https://fe-project.zooniverse.org/about'` until the related issue is resolved.

mcbouslog (approved):
LGTM :shipit:

PR #4326, opened 2023-03-01
Add ADR 47 for highlighter task

mcbouslog:
## Package
- lib-classifier

- [Figma interactive prototype of the highlighter task](https://www.figma.com/proto/miwXeAuA8iqBOD4E8ofOXR/Digi-Leap?scaling=min-zoom&page-id=0%3A1&starting-point-node-id=102%3A10&node-id=102%3A10)
- main PFE highlighter task [issue](https://github.com/zooniverse/Panoptes-Front-End/issues/3458) and [PR](https://github.com/zooniverse/Panoptes-Front-End/pull/3738)

## Describe your changes
- adds ADR 47 for highlighter task

## How to Review
  - [PFE highlighter task](https://www.zooniverse.org/projects/markbouslog/digileap-testing-production/classify?reload=1&workflow=23429)
  - highlight text, select label
- Are there plans for follow up PR’s to further fix this bug or develop this feature? yes, develop highlighter task for FEM

# Checklist

## General


docs/arch/adr-47.md
```diff
@@ -0,0 +1,69 @@
...
+To use the highlighter task a volunteer selects a label from the task area, highlights the related text in the text subject viewer, then selects a different label from the task area to highlight additional text in the text subject viewer or completes the task. Highlighted text can be unhighlighted.
+
+The highlighter task label(s) description and color is defined by researchers.
+
+The following is an example of a highlighter task annotation with labels "Scientific Name", "Collector Name", "Habitat", and "Location":
+```json
+{
+  "value": [
+    {
+      "labelInformation": {
```
mcbouslog, line 20: This annotation example is based on the existing PFE highlighter task annotation. Should this be snake case? `taskType` is camel case, so maybe it's ok this is also camel case.

docs/arch/adr-47.md
```diff
@@ -0,0 +1,69 @@
...
+  "taskType": "highlighter"
+}
+```
+
+## Decision
+Create a highlighter task in FEM.
+
+## Consequences
+Create an experimental task, the highlighter task, in FEM.
+If a workflow uses the highlighter task, add event listeners to the text subject viewer.
```
mcbouslog, line 66: Can the [interaction layer](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer) be used or is a simpler solution preferable? 

I think adding event listeners to the text subject viewer is only necessary if the user actions are select label then highlight text, and adding event listeners is not necessary if the user actions are highlight text then select label.

docs/arch/adr-47.md
```diff
@@ -0,0 +1,69 @@
...
+
+1 Mar 2023
+
+## Context
+To support OCR verification (the DigiLeap project), with subjects that consist of an image and OCR output as text file, we will create a highlighter task. The highlighter task will allow volunteers to highlight text in the OCR output with a label or labels defined by researchers.
+
+The FEM highlighter task is an iteration on the [PFE highlighter task](https://github.com/zooniverse/Panoptes-Front-End/tree/master/app/classifier/tasks/highlighter).
+
+## Features
+To use the highlighter task a volunteer selects a label from the task area, highlights the related text in the text subject viewer, then selects a different label from the task area to highlight additional text in the text subject viewer or completes the task. Highlighted text can be unhighlighted.
```
mcbouslog, line 11: Is this possible and preferable? As noted in this [PFE PR 3458 comment](https://github.com/zooniverse/Panoptes-Front-End/issues/3458#issuecomment-279364964) having a user select a label then highlight text might not be the technically or accessibly preferable solution.

I'll try to create some storybooks to investigate.

mcbouslog: Noting with the PFE highlighter task a user highlights text then selects the related label.

docs/arch/adr-47.md
mcbouslog: Likely a bad idea, but instead of the text subject viewer as is, maybe this task would work with a text subject viewer that is a text(area) input, the value of which can't be changed, but pointer and key events to highlight and unhighlight the contents of would be easier to work with within.

docs/arch/adr-47.md
```diff
@@ -0,0 +1,69 @@
...
+          "color": "#DCC6E0",
+          "label": "Location"
+      },
+      "start": 105,
+      "end": 206,
+      "text": "Located at Sheep Rock along Air\nBellows Gap Rd. (SR 1130) ca. 0.1 mi prior to Pruitt Rd. on the left."
+    }
+  ],
+  "task": "T0",
+  "taskType": "highlighter"
```
mcbouslog, line 57: I don't see if/how the PFE highlighter task used `taskType`, is that unique to FEM?
If `taskType` is unique to FEM is it ok to use `highlighter` even though the PFE highlighter task precedes this task?
I could use something else, like `highlighter-fem`.

docs/arch/adr-47.md
eatyourgreens: I could be wrong but isn’t the `InteractionLayer` designed to handle pointer events on interactive SVGs?

eatyourgreens: >maybe this task would work with a text subject viewer that is a text(area) input,

A couple of questions:
- how does the current highlighter task work? I know that it was originally built to work with keyboard, touch and pointer inputs, so it would be a good starting point. 
- have you considered using `contentEditable`?

docs/arch/adr-47.md
eatyourgreens: You’ll need some sort of interaction, eg. pressing a button, to indicate that selection is complete. 

Is it worth adding a link to the [Selection API](https://developer.mozilla.org/en-US/docs/Web/API/Selection) somewhere in this ADR? I assume that’s the API that you’ll be using.

eatyourgreens: For what it’s worth, the interactions involved in text selection are:
1. Select an initial range of text in the document (using a pointer, touch or the keyboard.)
2. Adjust the extent of the initial range. On desktop, this can be done from the keyboard. On touch, you’d use the little handles at the beginning and end of the range. 
3. Do something with the selection eg. Press a button to highlight it.

docs/arch/adr-47.md
eatyourgreens: This depends on how you plan to aggregate the annotations, I think.

docs/arch/adr-47.md
eatyourgreens: Which is preferred by Caesar?

I’d continue with the existing annotation format, unless there’s a pressing reason to change it. 

The task labels will be localised, so I’d assume that aggregation uses the project primary language.

docs/arch/adr-47.md
eatyourgreens: A concrete example of interaction design for text selection would be making text bold in a rich text editor.

eatyourgreens: Would it be helpful to think of the task here as something like the toolbar in a word processor or rich text editor?

docs/arch/adr-47.md
mcbouslog: > If the existing highlighter task meets the requirements, I’d port the code across and iterate from that starting point. That would at least get an initial prototype up in a couple of days or so.

Excellent suggestion and what I'm going to do! ✔️

mcbouslog: I'm having trouble using [the keyboard (Mac OS)](https://support.apple.com/en-us/HT201236) with the [PFE highlighter task](https://master.pfe-preview.zooniverse.org/projects/eatyourgreens/text-highlighter-demo/classify). My expectation was that I'd be able to tab into the text viewer and select text with arrow keys or the other Document Shortcuts linked? I'm not sure if that's right though.

eatyourgreens: As a sighted user, I use the shortcuts after double clicking to select a word eg. Select a word then adjust the selection by a few characters if it isn’t quite right. 

You might need to ask a blind user how to select text from the keyboard with VoiceOver or NVDA. There might be instructions online.

docs/arch/adr-47.md
eatyourgreens: Also: we never used this tool in a live project, so there's no chance of breaking existing classifications if you use `highlighter`.

`highlighter-fem` seems redundant, to me, given that classifications are versioned now, so you can distinguish classifications from the new classifier by looking at the `version` field.

eatyourgreens (approved):
Looks good to me. Maybe add a section about how these annotations will be aggregated by Caesar. Other tasks (single question, drawing, dropdown menu) use array indexes to track which choice was made, but this task is using colours and labels.

eatyourgreens (approved):

mcbouslog:
A note on translations and the highlighter task annotation - the task highlighter button label text is defined by translated strings - `task.strings.get(`highlighterLabels.${index}.label`)` - in [this code here](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/plugins/tasks/experimental/highlighter/components/HighlighterTask.js#L169), which works with different translations.

When an annotation is created the label information is pulled from - `const labelInformation = getSnapshot(task.highlighterLabels[labelIndex])` - in [this code](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/plugins/tasks/experimental/highlighter/components/HighlighterTask.js#L123).

After creating and reviewing a few classifications made in different languages the annotation `labelInformation.label` value was consistently the same, I think in the project's primary language.

I think this makes sense because the task information saved in the app (what `getSnapshot` will return) is what's defined in the workflow response, which will define workflow/tasks strings per the primary language, and won't change per translation changes. 

So I think the annotation `labelInformation.label` value is good (can be aggregated on across different translations), but I could use a sanity check if this all makes sense or I'm missing something.

PR #4331, opened 2023-03-03
Refactor getLayout with named exports

eatyourgreens:
Replace the `layoutsMap` object with named exports from `Layout/components`.

## Package
lib-classifier

## How to Review
Layout selection should continue to work as usual. Transcription subjects should expand to fill the available space in the window. Other subjects should expand up to 100rem (1600px.)

## General

## Refactoring


goplayoutside3 (commented):
This looks good! Everything is working as expected when I load transcription and non-transcription projects.

Just want to doublecheck though - Is this line doing anything in the Workflow store model once the layouts map is removed?
https://github.com/zooniverse/front-end-monorepo/blob/dfca9fb28e6bc442002d2f5a4cf091bae28683c3/packages/lib-classifier/src/store/WorkflowStore/Workflow/Workflow.js#L66

eatyourgreens:
Oh, good question. That should load `layouts.default`, which is the default export. I can’t remember if the tests check that case.

eatyourgreens:
@goplayoutside3 there's now an explicit test for the default case.

goplayoutside3 (approved):
Looks good!

PR #4333, opened 2023-03-06
lib-classifier: Add survey task instruction

### LINKED ISSUE #2437: Survey Task: Add an explanatory headline question to the survey task?
## Expected behavior

Really more like desired behaviour (maybe something to consider for the new FEM), but it would be good if a question could be specified for the survey task. Something like:
> "What type of Xs are present?"
It's not always entirely obvious what the task is. Specific context here is a project classifying knitting patterns and the survey is of garment types. However, there are several other tasks in the workflow and it's not entirely obvious what interaction is required from the user.

I recognise that a project tutorial might help, but not everyone reads or follows the tutorial!

## Current behavior
There is no opportunity to enter a question for the survey task.

## Steps to replicate
N/A

## Additional information

- **Operating system:** Any
- **Browser:** Any

mcbouslog:
## Package
- lib-classifier

- closes #2437 
- linked to https://github.com/zooniverse/Panoptes-Front-End/pull/6460

## Describe your changes
- conditionally show survey task instruction
- update stories, tests, and mock task accordingly

## How to Review
  - survey task with instruction - https://local.zooniverse.org:8080/?project=1634&workflow=3729
  - survey task without instruction - https://local.zooniverse.org:8080/?project=1634&workflow=2434
  - note task instruction
  - http://localhost:6006/?path=/story/tasks-survey--default&globals=locale:en;locales.en:English;locales.test:Test+Language

# Checklist

## General

## General UX

## New Feature

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
@@ -60,7 +60,12 @@ function Task ({
   }
 
   return (
-    <Box key={annotation.id} basis='auto'>
+    <Box
+      key={annotation.id}
+      as='fieldset'
+      basis='auto'
+      style={{ border: 'none' }}
+    >
       <TaskComponent
         {...props}
         autoFocus={autoFocus}
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
@@ -60,12 +60,7 @@ function Task ({
   }
 
   return (
-    <Box
-      key={annotation.id}
-      as='fieldset'
-      basis='auto'
-      style={{ border: 'none' }}
-    >
+    <Box key={annotation.id} basis='auto'>
       <TaskComponent
         {...props}
         autoFocus={autoFocus}
--- a/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/Chooser.js
+++ b/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/Chooser.js
@@ -20,6 +20,18 @@ const StyledText = styled(Text)`
   }
 `
 
+const components = {
+  a: Text,
+  h1: Text,
+  h2: Text,
+  h3: Text,
+  h4: Text,
+  h5: Text,
+  h6: Text,
+  p: Text,
+  span: Text
+}
+
 function Chooser ({
   disabled = false,
   filters = {},
@@ -54,7 +66,7 @@ function Chooser ({
     >
       {task.instruction 
         ? <StyledText as='legend' size='small'>
-            <Markdownz>
+            <Markdownz components={components}>
               {task.instruction}
             </Markdownz>
           </StyledText>
--- a/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/Chooser.js
+++ b/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/Chooser.js
@@ -48,7 +48,10 @@ function Chooser ({
   const filteredChoiceIds = getFilteredChoiceIds(filters, task)
   
   return (
-    <Box>
+    <Box
+      as={task.instruction ? 'fieldset' : 'div'}
+      style={{ border: 'none' }}
+    >
       {task.instruction 
         ? <StyledText as='legend' size='small'>
             <Markdownz>
--- a/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/Chooser.js
+++ b/packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/Chooser.js
@@ -1,12 +1,19 @@
-import { Box } from 'grommet'
+import { Box, Text } from 'grommet'
 import { observer } from 'mobx-react'
 import PropTypes from 'prop-types'
 import { useState } from 'react';
+import styled from 'styled-components'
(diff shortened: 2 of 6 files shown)


packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/Chooser.js
```diff
@@ -32,11 +39,26 @@ function Chooser ({
...
+    p: StyledText,
+    span: Text
+  }
+
   const showFilters = Object.keys(task.characteristics).length > 0
   const filteredChoiceIds = getFilteredChoiceIds(filters, task)
   
   return (
     <Box>
+      {task.instruction
```
eatyourgreens, line 59: Should this be the label for the choices menu? If so, it’ll need to be marked up as a legend or label, depending on how you structure the task in the DOM. 

At the moment, I think it’s a generic paragraph, not linked to the task inputs at all.

mcbouslog: Yes good point! I've refactored as legend, similar to the single, multiple, drawing, and dropdown tasks.

eatyourgreens: Cool. `legend` must be a child of `fieldset` so you might need to add that too.

mcbouslog: Ah yes, though it appears the other tasks noted are also [missing `fieldset`](https://github.com/zooniverse/front-end-monorepo/pull/3704/files#r981680120). I think adding `as='fieldset'` to the `Box` in Task.js makes sense. I'll add a commit accordingly.

mcbouslog:
Holding on merging https://github.com/zooniverse/Panoptes-Front-End/pull/6460 until this is merged and deployed. @eatyourgreens I think this is good to go with the `fieldset` change, but let me know if not or I'm missing anything.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
```diff
@@ -60,7 +60,12 @@ function Task ({
   }
 
   return (
-    <Box key={annotation.id} basis='auto'>
+    <Box
+      key={annotation.id}
+      as='fieldset'
```
eatyourgreens, line 65: This will mark up all tasks as fieldsets, not just survey tasks. That's likely to be a problem for some tasks eg. #2578 the transcription task instructions aren't valid fieldset legends.

eatyourgreens (commented):
I'm not sure about marking up all tasks as fieldsets.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.js
eatyourgreens: I'm thinking that `fieldset` must have `legend`, so the legend should also be defined in `Task.js`. Otherwise, it's too easy to add a new task without a legend.

Alternatively, move the `fieldset` down to the survey task itself, so that individual tasks are responsible for how they are marked up.

packages/lib-classifier/src/plugins/tasks/survey/components/components/Chooser/Chooser.js
mcbouslog: As suggested I've moved `fieldset` within the survey task Chooser component, close to the related `legend`.

eatyourgreens:
Neither of the test links work, so I can't check this in a screenreader. Those workflows aren't in the list of active workflows.

Screenshot of the dev classifier, listing all the active workflows for the test project.

eatyourgreens:
This is producing invalid HTML. The storybook story has `<p>` nested inside `<legend>` but only inline content is allowed in legends. They must be short, single strings so no paragraph breaks, lists, links, headings etc. The same rules as form labels, essentially.

<img width="206" alt="Screenshot of the DOM inspector showing a <p> nested in a <legend> for the example survey task." src="https://github.com/zooniverse/front-end-monorepo/assets/59547/6d613999-86f1-4894-9d60-96218058e664">

mcbouslog:
Sorry about the bad test project links, I've updated the initial comment, and and will note as follows:
- survey task with instruction - https://local.zooniverse.org:8080/?project=1634&workflow=3729
- survey task without instruction - https://local.zooniverse.org:8080/?project=1634&workflow=2434

eatyourgreens:
>I updated the Markdownz component the task instructions are within to have a components prop of inline (Grommet Text) components. That sets the child element of the legend to a span. Let me know if there's a better solution or if you have something else in mind. Nice catch and thanks for the clarification (that only inline content should be within legend element)!

Marking up a form in HTML is a skill every web developer should know, to be honest. #2578 has an example of a valid `fieldset`, and an invalid one from one of our projects. MDN has the content model for every HTML element too, including [`<legend>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend#technical_summary).

eatyourgreens (approved):
VoiceOver announces the default story as a group, titled 'Select the animals you see in the image.' 👍 

The filter button announces the group name twice, which is weird but not blocking. Each of the menu items announces its legend once, after the item name eg. 'Aardvark, menu item, Please select an animal - TEST INSTRUCTION.'

VoiceOver dialog in Safari, showing that the filter button announces its fieldset legend twice, after the button name.

One question, outside the scope of this PR, but are the survey task instructions localised in Panoptes? I think the `task.instruction` key is automatically added to translations for workflows, but I'm not 100% sure.

mcbouslog:
Good question about the translation, had to check, and confirmed `task.instruction` is included in translations.

inst_translation

PR #4371, opened 2023-03-17
Media previews for JSON data

### LINKED ISSUE #4328: Subject previews don't support JSON URLs
## Package
app-project

Subject previews on the project home page (Recent Talk comments) and classify page (Recent classifications) don't support subject locations with JSON URLs.

Given a URL that returns a JSON-encoded string, we should either be able to generate an image of the data, or generate an SVG rendering of the data. The SVG could be embedded directly in the DOM, or loaded in as an SVG image.

## Additional context
Projects like PH-TESS and SuperWASP Black Hole Hunters currently upload an image of the light curve as the first location for each subject, which then displays as the preview image for that subject. Dynamically generating preview images like this may not be feasible on a large scale for VRO projects, unless we offer our own service to render JSON-encoded strings as images.

eatyourgreens:
- add a `Data` component to handle `application/json` in the shared `Media` component.
- reflect on the MST type of a loaded data file to determine which viewer should be used.
- add a `DataSeriesPlot` component, which is a modified version of the `ScatterPlot` component that's used to show JSON data charts in the classifier.
- add a story and a test for `application/json` URLs.
- bump `@zooniverse/react-components` to 1.5.
- add data subjects to the project app storybook.

SVG previews of supernova light curves in the Recent Classifications section of the classifier.

SVG previews of supernova light curves in the Zooniverse Talk section of the project home page.

## Package
app-project
lib-react-components

- closes #4328.

## How to Review
See also https://github.com/zooniverse/Panoptes-Front-End/pull/6491

I have a test workflow with JSON subjects:
https://pr-6491.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground/classify?reload=0&workflow=3742&env=staging

If you classify a few subjects on that workflow, you should find them showing up in your Recents on the project classify page, on this branch. Subjects with Talk comments will also show up on the project's home page. There's one data subject with a comment at the moment (see the screenshot above.)

# Checklist

## General

## General UX

## New Feature


packages/lib-react-components/src/Media/Media.js
```diff
@@ -1,27 +1,32 @@
 import mime from 'mime/lite'
-import Audio from './components/Audio'
-import ThumbnailImage from './components/ThumbnailImage'
-import Video from './components/Video'
+import * as Viewers from './components'
```
goplayoutside3, line 2: I'm sometimes wary of using `import *` in case it interrupts tree-shaking, but here it's okay because every Viewer is being used in the Media.js file. Is that correct?

packages/lib-react-components/src/Media/components/Data/components/DataSeriesPlot/DataSeriesPlot.js
goplayoutside3, line 1: Would unit tests be valuable for this component?

packages/lib-react-components/src/Media/components/Data/components/DataSeriesPlot/helpers/getDataSeriesColor/getDataSeriesColor.js
```diff
@@ -0,0 +1,38 @@
...
+  return validateColor(chosenColor)
+}
+
+export default function getDataSeriesColor({
+    seriesOptions: seriesOptions = {},
+    seriesIndex: seriesIndex = 0,
+    themeColors: themeColors = {},
+    defaultColors: defaultColors = [],
+    highlighted: highlighted = false
+} = {}) {
```
goplayoutside3, line 31:
```suggestion
export default function getDataSeriesColor({
  seriesOptions = {},
  seriesIndex = 0,
  themeColors = {},
  defaultColors = [],
  highlighted = false
} = {}) {
```

packages/lib-react-components/test/setup.js
```diff
@@ -37,6 +37,7 @@ class ResizeObserver {
 
 window.ResizeObserver = ResizeObserver
 window.scrollTo = () => true
+window.cancelAnimationFrame = () => true
```
goplayoutside3, line 40: What component test is this required for?

packages/lib-react-components/src/Media/components/Data/components/DataSeriesPlot/DataSeriesPlot.js
```diff
@@ -0,0 +1,288 @@
...
+      >
+        <rect
+          borderColor={(dark) ? colors['light-5'] : colors['dark-5']}
+          fill={(dark) ? colors['light-3'] : colors['neutral-6']}
+          height={plotHeight}
+          left={leftPosition}
+          top={topPosition}
+          underlayParameters={underlayParameters}
+          width={plotWidth}
+        />
```
goplayoutside3, line 206: The `borderColor` and `underlayParameters` props throw console warnings because < rect > is  a DOM element. I recommend renaming to not clutter the console in a dev environment.

Screenshot 2023-03-24 at 7 48 02 AM
Screenshot 2023-03-24 at 7 47 46 AM

goplayoutside3 (commented):
Looking good! I reviewed with your test project:
https://local.zooniverse.org:3000/projects/eatyourgreens/-project-testing-ground/classify/workflow/3742, however just wanted to double check that the actual classifier data plot in FEM is not supposed to load for that workflow. Is that correct?

I was able to make two classifications while signed in by loading that workflow in PFE, and then viewing the workflow in FEM. The two data subjects showed up in my RecentSubjects component. I also see the data subject with a recent Talk comment on your project's home page.

packages/lib-react-components/src/Media/Media.js
eatyourgreens: Yep. The alternative would be to remove the static import and use `await import(Viewers[type])` at runtime instead.

packages/lib-react-components/src/Media/components/Data/components/DataSeriesPlot/helpers/getDataSeriesColor/getDataSeriesColor.js
eatyourgreens: I copied these helpers across from the classifier, without paying much attention to the code. They'll need changing in the classifier too.

packages/lib-react-components/test/setup.js
eatyourgreens: One of the `@visx` libraries, but I can't remember which one. We could figure it out by commenting this out and seeing what breaks.

packages/lib-react-components/src/Media/components/Data/components/DataSeriesPlot/helpers/getZoomBackgroundColor/getZoomBackgroundColor.spec.js
eatyourgreens, line 1: Looks like this didn't copy across properly from the classifier.

eatyourgreens, line 1: This file is empty in the classifier too. Weird.
https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/helpers/getZoomBackgroundColor/getZoomBackgroundColor.spec.js

eatyourgreens:
>Looking good! I reviewed with your test project:
https://local.zooniverse.org:3000/projects/eatyourgreens/-project-testing-ground/classify/workflow/3742, however just wanted to double check that the actual classifier data plot in FEM is not supposed to load for that workflow. Is that correct?

Yep. There's no JSON subject viewer set up for that workflow at the moment, so you have to classify a few subjects with the old classifier (linked in the PR description.)

packages/lib-react-components/test/setup.js
eatyourgreens: I think this might have been for `@visx/axis`, which I took out of this PR after removing axes from the preview charts. TippyJS uses `cancelAnimationFrame` too.

eatyourgreens:
If I draw a line, then delete it, it's removed from the final classification so this looks good. The array code could maybe be improved by using native array functions, rather than loops with `Array.push`, to reduce the number of operations per render.

Helen's feedback needs to be addressed. Otherwise, I think this is in good shape now.

packages/lib-react-components/src/Media/components/Data/components/DataSeriesPlot/DataSeriesPlot.js
eatyourgreens: Same problem with the tests for the `Media` component stories. The `Data` story is fine in the browser, but won't render an `<svg>` element in JSDOM. I worked around that with a test ID on the container `<div>`, but that isn't ideal.

packages/lib-react-components/src/Media/Media.js
eatyourgreens: Side note: dynamic subject viewer import at runtime would dramatically decrease the classifier page weight for transcription projects, which are currently downloading the PH-TESS light curve viewer, plus all of d3, even though they don't use any of that code.

goplayoutside3 (approved):

packages/lib-react-components/src/Media/components/Data/components/DataSeriesPlot/DataSeriesPlot.js
eatyourgreens: Reminder to myself to take a look at how `ScatterPlotViewer` is tested in the classifier, to see if that has any clues as to how to get this to work.

PR #4436, opened 2023-04-06
Classifier: change Annotate Button icon to "edit/pencil"

### LINKED ISSUE #4435: Design: Classifier's Annotate Button should look like a pencil ("edit" icon)
## UI Enhancement

Package: `lib-classifier`
Context: idea brought up during front-end call today

On the Classifier's Image Toolbar, the Annotate Button (which indicates that users can make drawing annotations on the Subject) looks like this:

image

But maybe it should look like this?

image
_(Source: Grommet "Edit" icon)_

### Dev Notes

I can very easily swap [AnnotateButton's](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/AnnotateButton) [Pointer Icon](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/AnnotateButton/PointerIcon.js) , but I need the appropriate SVG.

@seanmiller26 do you have a pencil SVG of size 32x32 we can use? I'm not sure why we're not using [Grommet Icons](https://icons.grommet.io/?s=edit), nor for that matter, where the original SVGs for these FEM buttons ([eg1](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/ZoomInButton/ZoomInIcon.js) [eg2](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/components/ResetButton/ResetIcon.js)) came from. 🤷 

### Status

Low priority, but should be a minor effort for something - actually, if you think about it - makes a lot of sense.

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Affects: Classifier -> Image Toolbar -> Annotate Button
Closes #4435 

This PR changes the Annotate Button's "arrow/pointer" icon with a "pencil/edit" icon, which we believe makes the purpose of "Annotate mode" more obvious. 

Old:

image

New:

image and image

### Dev Notes

See 4435 for notes on how the icons in the ImageToolbar use _fill_ instead of _stroke_ to paint the icons' lines.

### Status

Ready for review!


shaunanoordin:
For additional context, here's what the icon looks like on a /classify page (i.e. viewed as a whole on app-project), e.g. https://local.zooniverse.org:3000/projects/darkeshard/test-project-2021/classify/workflow/3519:

image

goplayoutside3:
Looks good to me! Will wait for Sean's design approval.

NB suggestion to exclude the Enzyme unit test in this PR as we're trying to get rid of those. In fact, none of the Icon components in the image toolbar really need a unit test. They're simply svg strings.

seanmiller26 (approved):
Looks great! This should help with any previous user confusion on what the difference was between the cursor and pan icons.

eatyourgreens:
Not a blocker, but what does the pencil button mean on workflows that only have question tasks (or survey tasks?)

seanmiller26:
@eatyourgreens we discussed at the most recent front-end meeting that we would only show this pencil/edit icon for any drawing task. We would NOT show it for survey, question tasks, etc. The top-most tool would be pan. There will be a future PR, I'm sure, which describes this more in-depth.

We may want to have both of these changes take place at the same time to avoid any future confusion.

eatyourgreens:
👍🏻

It came to mind because I’d just been looking at this workflow: https://frontend.preview.zooniverse.org/projects/zooniverse/gravity-spy/classify/workflow/1610

shaunanoordin:
## PR Update

Thanks all! Here's an update for this PR:

- ❗ This PR should only be merged once PR #4454 is addressed
  - Based on [Sean's comments,](https://github.com/zooniverse/front-end-monorepo/pull/4436#issuecomment-1500738185) we want to make sure we address the _"Classifier's Annotate Button should only appear for Drawing Workflows"_ behaviour _at the same time_ as this PR, so we have a coherent UX
- Future cleanup: we should remove the unnecessary (and very _Enzyme-ish_ ) tests from the "Icon" components
  - See [Delilah's comments](https://github.com/zooniverse/front-end-monorepo/pull/4436#issuecomment-1499557363)

seanmiller26:
This PR is reliant on changing the 

> "Classifier's Annotate Button should only appear for Drawing Workflows" 

So if that's what we need to do until then, then ok by me.

goplayoutside3:
@seanmiller26 I've got this on my radar to merge, I'm just waiting to be sure https://github.com/zooniverse/front-end-monorepo/pull/5994 is wrapped up.

I'll keep my assignment and do the merging 👍

goplayoutside3 (approved):

PR #4468, opened 2023-04-13
Fix button icon sizing

eatyourgreens:
Wrap nested Grommet icons inside an SVG element. Nested SVG elements inherit their size and position from their nearest parent SVG. https://www.sarasoueidan.com/blog/nesting-svgs/#nesting-svgs

Screenshot of the freehand line tools in Firefox 111. The icons for the line controls are sized normally.


kieftrav (approved):
As discussed, this is clunky but does what we need. Good find and LGTM!

PR #4592, opened 2023-05-09
Classifier: fix offset in VisXZoom's "zoom-layer"

### LINKED ISSUE #4593: Classifier: ScatterPlotViewer (VisXZoom)'s "zoom-layer" has a wonky offset
## PR Overview

Package: lib-classifier
Affects: ScatterPlotViewer (and theoretically any component that would use VisXZoom)
Follows: #4482's review

At the moment, the "zoom layer" (the invisible rectangle that's supposed to cover a scatterplot chart, and listen for pan/zoom actions from the mouse) has an odd offset.

_Screenshot: (before fix, [SuperWASP Blackhole on lib-classifier](https://localhost:8080/?env=production&project=hughdickinson/superwasp-black-hole-hunters&workflow=17386)) : The visible "zoom layer", marked by a focus-outline on Chrome, is offset to the bottom-right. This means if a user's mouse is on the far left or top edge of the chart, they can't use the mouse to pan & zoom._
Screenshot 2023-05-08 at 21 32 44

Expected behaviour:

_Screenshot ([ScatterPlotViewer -> X Range Selection on Storybook](http://localhost:6006/?path=/story/subject-viewers-scatterplotviewer--x-range-selection&globals=locale:en;locales.en:English;locales.test:Test+Language)) : The "zoom layer" should perfectly cover the area/"viewport" set for the visible data points._
Screenshot 2023-05-09 at 00 46 14

Note: the Zoom Layer should be fixed with respect to the view of the Subject Viewer itself (i.e. pan & zoom shouldn't move it around), so users always have a consistent layout to interact with.

### Dev Notes

Q: Does this bug affect live projects?
A: Yes, [SuperWASP: Black Hole Hunters,](https://www.zooniverse.org/projects/hughdickinson/superwasp-black-hole-hunters) which is an FEM project that uses the ScatterPlotViewer. Not sure what else might be affected.

Q: Does this bug appear in multiple browsers? 
A: Yes, tested on macOS + Chrome 112 and FF 112, though only really noticeable in Chrome due to the focus-outline.

Q: What are the conditions for this bug trigger?
A: The Classifier has to display a ScatterPlotViewer with pan/zoom enabled.

### Status

Relatively low urgency. It's a definite UI/UX issue, but it was only noticed during a separate PR review, so the issue doesn't seem to have inconvenienced volunteers yet.

Fix in progress: #4592

Please see that PR for full technical details.

shaunanoordin:
## PR Overview

Package: lib-classifier
Affects: ScatterPlotViewer (and theoretically any component that would use VisXZoom)
Follows: #4482's review
Fixes #4593 

At the moment, the "zoom layer" (the invisible rectangle that's supposed to cover a scatterplot chart, and listen for pan/zoom actions from the mouse) has an odd offset.

_Screenshot: (before fix, [SuperWASP Blackhole on lib-classifier](https://localhost:8080/?env=production&project=hughdickinson/superwasp-black-hole-hunters&workflow=17386)) : The visible "zoom layer", marked by a focus-outline on Chrome, is offset to the bottom-right. This means if a user's mouse is on the far left or top edge of the chart, they can't use the mouse to pan & zoom._
Screenshot 2023-05-08 at 21 32 44

This fix removes that offset.

_Screenshot (after fix, [ScatterPlotViewer -> X Range Selection on Storybook](http://localhost:6006/?path=/story/subject-viewers-scatterplotviewer--x-range-selection&globals=locale:en;locales.en:English;locales.test:Test+Language)) : Now, the "zoom layer" is correctly covering all the visible data points._
Screenshot 2023-05-09 at 00 46 14

### Dev Notes

From the DOM, I can see...
```
<g class="chartContent" transform="translate(60, 10)">
  <rect class="chartBackground" left="60" top="10">
  <g> // 1st data point
  <g> // 2nd data point
  (etc)
  <rect data-testId="zoom-layer" transform="translate(60, 10)">
```

The chartContent parent has an x-offset of 60px to compensate for the width of the y-axis, and a y-offset of 10px (for aesthetic reasons, I think?). These offsets are then replicated by the "zoom layer" rect, which unfortunately introduced and extra unwanted offset of wonkiness. (Side note: the chartBackground's [top & left are invalid and ignored.](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/rect#attributes))

Digging into the code, I see that the unnecessary `left` and `top` are added in `VisXZoom.js`:
```
left = 60 // padding to compensate for the width of the y-axis's markers
top = 10 // padding to compensate for the height of the (bottom?) y-axis's markers
<Zoom left={left} top={top} ...>  // Hey, left and top are already used here
  <ZoomingComponent ...>
    <ZoomEventLayer
      left={left} top={top}  // Yeah, now you're just offsetting an element whose parent already has the same offset.
    >
```

Removing top/left from `<ZoomEventLayer>` seems like the most sensible thing to do.

UPDATE: I'm now not 100% sure if `<Zoom/>` even uses [left & top,](https://airbnb.io/visx/docs/zoom) so maybe I can remove those two params from `VisXZoom.js` altogether. 🤷 Ping me if you think I should.

### Status

Ready for review. I'd be particularly interested in knowing of any scenarios that I might have missed, e.g. _"actually, a current/future component WILL require left/top offsets to be passed to the VisXZoom, so you can't just delete those params",_ etc

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VisXZoom/VisXZoom.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VisXZoom/VisXZoom.js
@@ -123,7 +123,6 @@ function VisXZoom({
 
 
   const ZoomingComponent = zoomingComponent
-  console.log('+++ ZoomingComponent', ZoomingComponent)
   return (
     <Zoom
       constrain={constrain}
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VisXZoom/VisXZoom.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VisXZoom/VisXZoom.js
@@ -123,6 +123,7 @@ function VisXZoom({
 
 
   const ZoomingComponent = zoomingComponent
+  console.log('+++ ZoomingComponent', ZoomingComponent)
   return (
     <Zoom
       constrain={constrain}
@@ -148,7 +149,6 @@ function VisXZoom({
             <ZoomEventLayer
               focusable
               height={height}
-              left={left}
               onDoubleClick={onDoubleClick}
               onKeyDown={onKeyDown}
               onPointerDown={panning ? _zoom.dragStart : DEFAULT_HANDLER}
@@ -159,7 +159,6 @@ function VisXZoom({
               onWheel={onWheel}
               panning={panning}
               tabIndex={0}
-              top={top}
               width={width}
             />
           </ZoomingComponent>
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VisXZoom/VisXZoom.spec.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/VisXZoom/VisXZoom.spec.js
@@ -177,21 +177,17 @@ describe('Component > VisXZoom', function () {
         expect(wrapper.find(StubComponent).find(ZoomEventLayer)).to.have.lengthOf(1)
       })
 
-      it('should set the height, width, and left and top positions by props', function () {
+      it('should set the height and width by props', function () {
         const wrapper = mount(
           <VisXZoom
             data={mockData}
-            left={20}
             height={height}
             width={width}
-            top={40}
             zoomingComponent={StubComponent}
           />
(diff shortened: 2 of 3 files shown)


shaunanoordin:
What do you mean, tests are failing? Let's see...

shaunanoordin:
OK, tests fixed, but that VisXZoom.spec.js is still running on Enzyme, which is a TODO for another day.

eatyourgreens (approved):

PR #4526, opened 2023-04-27
Fetch FreehandLine Tool reductions from Caesar POC.

kieftrav:
_Please request review from `@zooniverse/frontend` team or an individual member of that team._ 

## Describe your changes
Implement the hook to pull any machine-made freehand lines from Caesar.

## How to Review
https://localhost:8080/?project=kieftrav/freehand-line-test

Run branch locally to see it pull in a test Caesar reduction. Testing and Storybook stories are a WIP.

## General
- [ ] Tests are passing locally and on Github
- [ ] Documentation is up to date and changelog has been updated if appropriate
- [ ] You can `yarn panic && yarn bootstrap` or `docker-compose up --build` and FEM works as expected
- [ ] FEM works in all major desktop browsers: Firefox, Chrome, Edge, Safari (Use Browserstack account as needed)
- [ ] FEM works in a mobile browser

## General UX
Example Staging Project: [freehand-line-test](https://local.zooniverse.org:3000/projects/kieftrav/freehand-line-test)
- [ ] All pages of a FEM project load: Home Page, Classify Page, and About Pages
- [ ] Can submit a classification
- [ ] Can sign-in and sign-out
- [ ] The component is accessible
  - Can be used with a screen reader [BBC guide to VoiceOver](https://bbc.github.io/accessibility-news-and-you/assistive-technology/testing-steps/voiceover-mac.html)
  - Can be used from the keyboard [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - It is passing accessibility checks in the storybook

## New Feature
- [ ] The PR creator has listed user actions to use when testing the new feature
- [ ] Unit tests are included for the new feature
- [ ] A storybook story has been created or updated
  - Example of SlideTutorial [component](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.js) with docgen comments, and its [story](https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/index.html?path=/docs/other-slidetutorial--default)

PR #4607, opened 2023-05-11
Fix broken Caesar hooks

eatyourgreens:
- remove early return before `useCaesarReductions`.
- add type checking before using `caesarReductions`.
- don't override existing classification `marks` in `useFreehandLineReductions`.

This should fix a couple of breaking changes in #4526.


eatyourgreens:
#4607 should fix the broken drawing tasks. Transcription tests are still failing on that branch, though.

kieftrav (approved):

PR #4639, opened 2023-05-17
Add the git ref to page metadata

eatyourgreens:
Add the git ref name (eg. `master`, `production-release` or a branch name) to project builds. Add it to page metadata as `zooniverse:deployed_ref` so that we can easily check which branch we are using for a branch deploy.

Page metadata in the DOM inspector, showing that the deployed ref is 'tutorial-sizing'.

## Package
app-project 

## How to Review
Run a kubernetes deploy and check the page metadata in the DOM inspector.


mcbouslog (approved):
Great idea, thanks for making the changes!

PR #4648, opened 2023-05-19
JSONData ADR

eatyourgreens:
An Architecture Decision Record for the latest updates to JSON data subjects. It follows on from [ADR 28](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-28.md), ie. the consequences from ADR 28 form the context for this ADR. 

### DIFF
--- a/docs/arch/README.md
+++ b/docs/arch/README.md
@@ -47,3 +47,4 @@
 - [ADR 45: Custom Video Controls](adr-45.md)
 - [ADR 46: Flipbook Viewer](adr-46.md)
 - [ADR 47: Highlighter Task](adr-47.md)
+- [ADR 48: JSONData subject location type and viewer](adr-48.md)
--- a/docs/arch/adr-48.md
+++ b/docs/arch/adr-48.md
@@ -1,0 +1,50 @@
+# ADR 48: JSONData subject location type and viewer
+
+8th June 2023
+
+## Context
+
+Since [ADR 28](adr-28.md) there are a number of JSON data subject viewers, each with its own loading and initialisation code, and its own custom view for rendering a chart as SVG:
+- Bar Chart Viewer.
+- Data Image Viewer.
+- Light Curve Viewer.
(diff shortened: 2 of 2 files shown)


eatyourgreens:
Subject previews in the NextJS app, and Talk, collections and recents in PFE only work for charts if the MIME type is `application/json`.

If the MIME type of the JSON data is `text/plain`, then you still need to generate a fallback image for each chart and add it as the first location for each subject. Otherwise, subject previews will use the text viewer. 

I should add that to this ADR.

shaunanoordin (approved):
## PR Review

Package: documentation

This PR adds a succinct explanation for our handling of JSON data in Subjects.

LGTM. We just need to change that XX to a proper ADR number and we're good to merge. 👍

eatyourgreens:
>If the MIME type of the JSON data is text/plain, then you still need to generate a fallback image for each chart and add it as the first location for each subject. Otherwise, subject previews in PFE will use the text viewer, and the new Media component will simply show nothing.

I still need to add that to the Consequences section. I'll merge once that's done and the number's updated.

eatyourgreens:
@shaunanoordin here's my updated Consequences for this ADR:

>- We don't need individual data loading code for each JSON data subject viewer. Old connector and container components, and their corresponding Enzyme tests, can be deleted.
>- Workflow configuration doesn't need to be set by hand when a subject location's MIME type is `application/json`.
>- Preview SVG's are automatically generated for `application/json` subject locations.
>- Planet Hunters TESS, which uses the `text/plain` MIME type, does still need to have its viewer configured manually.
>- When a JSON subject uses the `text/plain` MIME type, the project team still needs to add a preview image to each subject. The preview **must** be the first location for each subject and **must** use an `image/*` MIME type.
>- Manual workflow configuration is replaced by a single `subject_viewer: 'jsonData'` setting. Old settings for individual viewers are kept for backwards compatibility, but deprecated.

PR #4799, opened 2023-06-12
lint report for missing translations

NilsJacobsen:
## Description

I installed `@inlang/cli` to lint the language resources to see where we have missing Translations or where we have no `en` reference anymore. As I am German, I was looking for the status of the german translations and there are still `118` missing. I started updating them.

## Further reading

I added the cli to your project so everybody can check the status of the messages by executing `npx inlang lint`. Also `machine translate` and `open editor` are available see [docs](https://inlang.com/documentation/apps/inlang-cli)


NilsJacobsen:
Hey @goplayoutside3 thanks for the fast reply:) 
I wanted to provide you with a better way of handling translations to improve your lint report. I would say the inlang/cli is a no brainer for devs and they can coexist with Lokalise (Disclaimer: I also contribut on inlang). It's only a short config. 

But for sure I leave it too you:) Have a nice one:)

PR #4801, opened 2023-06-13
Organisations: small fixes

eatyourgreens:
- add `organization` to the store during SSR, and pass a snapshot to the browser.
- return null when organisations aren't found eg. they exist but are not public.
- model translation strings as a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) of nullable strings.

## Package
app-project

## How to Review
Try this out with projects that are part of public organisations, eg. Snapshot Serengeti or Notes from Nature, and projects that are part of private organisations, eg. Galaxy Zoo or Gravity Spy.

# Checklist

## General

## General UX

## Refactoring

### DIFF
--- a/packages/app-project/src/helpers/fetchOrganization/fetchOrganization.js
+++ b/packages/app-project/src/helpers/fetchOrganization/fetchOrganization.js
@@ -18,6 +18,7 @@ async function fetchOrganizationData(organizationID, env) {
   } catch (error) {
     logToSentry(error)
     console.log('Error loading organization:', error)
+    return null
   }
 }
 
--- a/packages/app-project/stores/Organization.js
+++ b/packages/app-project/stores/Organization.js
@@ -2,6 +2,8 @@ import { types } from 'mobx-state-tree'
 
 import numberString from './types/numberString'
 
+const TranslationStrings = types.map(types.maybeNull(types.string))
+
 const Organization = types
   .model('Organization', {
     id: types.maybeNull(numberString),
@@ -9,12 +11,12 @@ const Organization = types
     listed: types.optional(types.boolean, false),
     primary_language: types.optional(types.string, 'en'),
     slug: types.optional(types.string, ''),
-    strings: types.frozen({})
+    strings: TranslationStrings
   })
 
   .views(self => ({
     get title () {
-      return self.strings?.title
+      return self.strings.get('title')
     },
   }))
 
--- a/packages/app-project/stores/Organization.js
+++ b/packages/app-project/stores/Organization.js
@@ -5,9 +5,11 @@ import numberString from './types/numberString'
 const Organization = types
   .model('Organization', {
     id: types.maybeNull(numberString),
+    display_name: types.optional(types.string, ''),
+    listed: types.optional(types.boolean, false),
     primary_language: types.optional(types.string, 'en'),
     slug: types.optional(types.string, ''),
-    strings: types.frozen({}),
+    strings: types.frozen({})
   })
 
   .views(self => ({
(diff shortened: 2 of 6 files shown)


packages/app-project/src/helpers/fetchOrganization/fetchOrganization.js
```diff
@@ -18,6 +18,7 @@ async function fetchOrganizationData(organizationID, env) {
   } catch (error) {
     logToSentry(error)
     console.log('Error loading organization:', error)
+    return null
```
eatyourgreens, line 21: There's an ESLint rule to catch inconsistent returns from functions. It would have caught this, but I don't think it's set on the project app.

eatyourgreens: https://github.com/zooniverse/front-end-monorepo/blob/24b9fb49ebe2c825e21dc8c593e7df301e5e08ba/packages/tools-standard/options.js#L24-L26

packages/app-project/stores/Organization.js
```diff
@@ -5,9 +5,11 @@ import numberString from './types/numberString'
 const Organization = types
   .model('Organization', {
     id: types.maybeNull(numberString),
+    display_name: types.optional(types.string, ''),
```
eatyourgreens, line 8: Do we use display names? The tests expect them but they've been replaced by translation strings.

packages/app-project/stores/Organization.js
```diff
@@ -5,9 +5,11 @@ import numberString from './types/numberString'
...
+    listed: types.optional(types.boolean, false),
     primary_language: types.optional(types.string, 'en'),
     slug: types.optional(types.string, ''),
```
eatyourgreens, line 11: This shouldn't be optional. All organisations must have a slug. However, tests fail if the slug is required.

eatyourgreens: Same comment for the project model too.

packages/app-project/stores/Organization.js
```diff
@@ -5,9 +5,11 @@ import numberString from './types/numberString'
...
-    strings: types.frozen({}),
+    strings: types.frozen({})
```
eatyourgreens, line 12: Translation strings should really be a map of strings: `types.map(types.string)`. I think there are examples of this in the classifier.

packages/app-project/stores/Organization.js
eatyourgreens: The app breaks if you make slugs required. 🤔

packages/app-project/stores/Organization.js
eatyourgreens: This is a small enough change that it can go into this PR too.

packages/app-project/stores/Organization.js
eatyourgreens: Yeah, they only seem to be here so that the tests pass.

mcbouslog (approved):
LGTM! Thank you for the quick fixes.

I tested on the following projects with listed org:
- https://local.zooniverse.org:3000/projects/markb-panoptes/uwworgtest-chicago-wildlife
- https://local.zooniverse.org:3000/projects/fr/markbouslog/digileap-testing-production/classify/workflow/23429?env=production&demo=true
Not listed org:
- https://local.zooniverse.org:3000/projects/zookeeper/galaxy-zoo?env=production&demo=true
No org:
- https://local.zooniverse.org:3000/projects/brooke/i-fancy-cats

PR #4803, opened 2023-06-13
Refactor Panoptes authentication in all packages 

### LINKED ISSUE #4766: Reflect on JWT payload for authentication
Reflect on the payload of the Panoptes JWT to get the user, rather than spamming the API at `/api/me`.

## Package
- app-content-pages
- app-project
- lib-classifier 
 

## Related issues

- https://github.com/zooniverse/sugar/issues/47
- https://github.com/zooniverse/front-end-monorepo/pull/4697

### LINKED ISSUE #4845: Classifier user is null when logged in
## Package
lib-classifier 
## Describe the bug
When the `Classifier` component mounts, `usePanoptesUser` returns null, even if you are signed in on the Classify page. 

https://github.com/zooniverse/front-end-monorepo/blob/a592e95080d4f269fda753290f5a8478b2537339/packages/lib-classifier/src/components/Classifier/Classifier.js#L25

## To Reproduce
Either run the classifier in a debugger, stepping through the code and examining the `user` variable, or log `user` to the console. Do this in the Classify page while logged in. 

## Expected behavior
`usePanopteUser` should be null while you are signing in (on page load) but should return the signed-in user once you are logged in. Requests for workflows etc. should be deferred until you are signed in (see #4875.)

## Additional context
Since #4697 we can get the logged-in user immediately, from the Panoptes bearer token. 

`auth.checkCurrent()` should run once, when the Classify page loads. Thereafter, the current Panoptes session should be checked with `auth.checkBearerToken()` (see #4766.)

### LINKED ISSUE #4875: Classifier: Subject Queue for logged-in users sometimes fetched twice
## Behaviour Bug

Package: `lib-classifier`
First noted on PR #4848's [review](https://github.com/zooniverse/front-end-monorepo/pull/4848#pullrequestreview-1489300490) (20 Jun)
Confirmed to exist at least from 78c9baf (1 Jun) to 18d65f3 (21 Jun)

If you're already logged in AND start by opening the Classify page, there's a chance the Classifier will send 2x requests to /queued (with the correct header auth, etc) instead of 1x request.

Here's how you trigger the problem:
- Ensure you're logged in as a user (e.g. darkeshard), and ensure you can monitor network connections
- Open the Classify page of a test project [(example)](https://frontend.preview.zooniverse.org/projects/darkeshard/test-project-2021/classify/workflow/20333)
- Observe the number of times /queued is called as soon as the Classify page is loaded
  - Expectation: the /queued endpoint should be hit once (with the user's auth in the headers)
  - Actual: the /queued endpoint is _sometimes_ hit _twice_ (with the correct auth)
- ⚠️ NOTE: keep refreshing the page to see the results, there's a level of inconsistency to triggering this problem.

_Screenshot: the page loads, and the /queued endpoint is hit twice instead of once_

image

### Dev Notes

Initial notes:

- The problem is very easy to replicate on localhost development build (as Jim pointed out, this is likely due to React Strict Mode trying to find bugs due to impure renders)
- The problem is harder (more inconsistent?) to replicate - but still possible - on frontend.preview (which uses a production build?)
  - e.g. TESS on frontend.preview and www very very rarely triggers double fetches ; while a rando test project on frontend.preview [(example)](https://frontend.preview.zooniverse.org/projects/darkeshard/test-project-2022/classify/workflow/22164) will trigger double fetches 50% of the time.
  - UPDATE: take my "very very rarely triggers" with a pinch of salt. Jim has confirmed that he managed to trigger the bug on Gravity Spy and Planet Hunters TESS. (See [Slack](https://zooniverse.slack.com/archives/CANKLB50E/p1687364601026659?thread_ts=1687363524.775759&cid=CANKLB50E)) 
- localhost testing shows that the bug exists as far back as 78c9baf at least (that's 1 June), as shown by local testing.

Updates:

- Q: do the number of subjects available from /queued affect how likely the bug is to trigger?
  - A: I can't confirm if this is a factor.
  - Early on (on local dev), my test projects that triggered the double fetch only had < 10 subjects available to the workflow, and they triggered the bug more-or-less consistently. Meanwhile, an actual project like Planet Hunters TESS (also on local dev) almost never triggered. (About 10 refreshes for each test) 
  - However, I've since upped the number of subjects on my test project's workflows (> 20 subjects) and can still see the double-fetch bug happening.
  - And then I took a look at Davy Notebooks...

- Q: what's with Davy Notebooks?
  - A: Davy Notebooks is weird. (This is on production btw, so that's the www URL at the latest 18d65f3 commit.)
  - Let's start by logging in and opening the [root Classify page](https://www.zooniverse.org/projects/humphrydavy/davy-notebooks-project/classify) (no WF selected)
      image 
  - If I _switch to_ Workflow 22114 (92% complete, which has >= 10 Subjects), the /queued endpoint is hit exactly once and returns 10 subjects.
      image
  - If I _switch to_ Workflow 22814 (97% complete, which as exactly 1 Subject), the /queued endpoint is hit _twice_ and returns 1 subject.
      image
  - If I then _refresh_ the page to reload WF 22814... the queued endpoint is never hit. What?
      image
  - Is this cached? Auto-hydrated from the server side? etc etc

- Q: does the number of Subjects served by the /endpoint affect how likely the bug is to trigger?
  - A: Can you give more specific scenarios? 

- OK, sure. How about... Q: does the project being "finished" (e.g. Planet Hunters TESS) make the bug less likely to trigger? After all, if a project is finished, the /queued endpoint will always serve the same 10 Subjects, so there's some consistency there.
  - A: uuuhhh

- Q: does the user affect how likely the bug is to trigger? e.g. if `darkeshard` has already seen so many Subjects in the test project, perhaps that reduces the pool of Subjects the /endpoint has to offer him?
  - A: uuuhhh

### Status

Requires further investigation

### LINKED ISSUE #4946: Classifier unmounts and remounts with a logged-in user
## Package
app-project

## Describe the bug
Related to #4875.

If the Classify page loads with an existing user session, `appLoadingState`  goes through the following sequence of values:
1. 'success' (from the existing user session.)
2. 'loading' (while project preferences reload.)
3. 'success' (after project preferences have loaded.)

This means that `ClassifierWrapper` here mounts, unmounts then mounts again.
https://github.com/zooniverse/front-end-monorepo/blob/30b77b0db75475157904e8447cef8460492e459b/packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapper.js#L93-L99

This leads to classifier initialisation, including requests for queued subjects, running twice  on page load and any annotations in progress being lost. 

## To Reproduce
Load a project's Classify page while you have a stale user session cached in browser session storage.

You can also reproduce this bug by leaving the Classify page open in a tab, going to another tab, then returning to the original tab after the cache has become stale (about 60s, I think.)

## Expected behavior
Background refresh should update the user and project preferences without unmounting the classifier.

### LINKED ISSUE #4975: Multiple project_roles requests
## Package
- lib-classifier

## Describe the bug
It appears that the same `project_roles` request is made every few seconds on a project classify page.

## To Reproduce
1. Go to https://www.zooniverse.org/projects/humphrydavy/davy-notebooks-project/classify/workflow/22812?demo=true
2. Observe the dev console network requests
3. Watch for `project_roles` requests

## Expected behavior
One `project_roles` request is made per project classify page load.

eatyourgreens:
Refactor `usePanoptesUser` to get the current user from the stored Panoptes JWT (#4766.) Fall back to `auth.checkCurrent()`, which gets a new session and token from the Panoptes API, if we don't have a stored token.

Refactor the classifier to authenticate in `ClassifierContainer`, waiting until sign-in is complete and all user data has loaded before rendering the `Classifier` component (#4845 and #4875.) Move all the workflow validation up to `ClassifierContainer`, so that we validate the workflow ID (from the URL) when the user has loaded, then pass down null to `Classifier` if they don't have permission to view that workflow.

Test `ClassifierContainer`, checking that:
- inactive workflows only load when you have permission to view them.
- the mock API for `/subjects/queued` is only called once.

Fix a bug, in the project app, where updating the project user in the project app's store will unmount and mount the classifier (#4946.)

Fix a bug where a `project_roles` request is continually made (closes #4975).

The `Classifier` component uses MobX observable data, but isn't wrapped in `observer`, so I've also fixed that here.

## Package
- app-content-pages
- app-project
- lib-classifier

- Closes #4766.
- Closes #4845.
- Closes #4875.
- Closes #4946.
- Closes #4975.
- https://github.com/zooniverse/panoptes-javascript-client/issues/53 

## How to Review
This PR reflects on the Panoptes JSON Web Token to get the logged-in user for the project app, the content pages app and the classifier, which should be faster than pinging the API for the current user session and won’t ask Panoptes OAuth to generate a new token (https://github.com/zooniverse/panoptes-javascript-client/issues/53.) If there's no stored token yet, it falls back to `auth.checkCurrent()`, which is currently used for all user checks.

You should be able to use any project, logged-in or not, and continue to use Admin Mode without any changes to the app's behaviour.

On page load, there should only be one request to `/api/me` to get the user resource. 

Subjects should only be fetched once, with your access token, after you log in (#4875.)

Apps will still lose your stored session when they unload eg. when going from PFE to FEM on the same origin, or when going from the project app to the content pages app. That can be tested in staging, on the `https://frontend.preview.zooniverse.org` origin, after this PR merges. 

# Checklist

## General

## General UX

## Refactoring

### DIFF
--- a/packages/app-project/stores/User/User.spec.js
+++ b/packages/app-project/stores/User/User.spec.js
@@ -1,21 +1,80 @@
 import asyncStates from '@zooniverse/async-states'
+import * as client from '@zooniverse/panoptes-js'
 import { expect } from 'chai'
+import { when } from 'mobx'
 import { getSnapshot } from 'mobx-state-tree'
+import nock from 'nock'
 import sinon from 'sinon'
 
 import Store from '@stores/Store'
 import initStore from '@stores/initStore'
-import placeholderEnv from '@stores/helpers/placeholderEnv'
-
-const user = {
-  display_name: 'Jean-Luc Picard',
-  id: '1',
-  login: 'zootester1'
-}
 
 describe('stores > User', function () {
-  let rootStore = Store.create({}, placeholderEnv)
-  let userStore = rootStore.user
+  const user = {
+    display_name: 'Jean-Luc Picard',
+    id: '1',
+    login: 'zootester1'
+  }
+  let rootStore
+  let userStore
+
+  beforeEach(function () {
+    nock('https://panoptes-staging.zooniverse.org/api')
+    .persist()
+    .get('/users/1/recents')
+    .query(true)
+    .reply(200, { recents: [] })
+    .get('/users/2/recents')
+    .query(true)
+    .reply(200, { recents: [] })
+    .get('/collections')
+    .query(true)
+    .reply(200, { collections: [] })
+
+    nock('https://panoptes-staging.zooniverse.org/api')
+    .get('/project_preferences?project_id=1&user_id=1&http_cache=true')
+    .reply(200, {
+      project_preferences: [
+        { activity_count: 23 }
+      ]
+    })
+    .get('/project_preferences?project_id=1&user_id=1&http_cache=true')
+    .reply(200, {
+      project_preferences: [
+        { activity_count: 25 }
+      ]
+    })
+    .get('/project_preferences?project_id=1&user_id=2&http_cache=true')
+    .reply(200, {
+      project_preferences: [
+        { activity_count: 27 }
+      ]
+    })
+
+    nock('https://talk-staging.zooniverse.org')
+    .persist()
+    .get('/conversations')
+    .query(true)
+    .reply(200, { conversations: [] })
+    .get('/notifications')
+    .query(true)
+    .reply(200, { notifications: [] })
+
+    nock('https://graphql-stats.zooniverse.org')
+    .persist()
+    .post('/graphql')
+    .reply(200, { data: { statsCount: [] }})
+
+    rootStore = Store.create({ project: {
+      id: '1',
+      loadingState: asyncStates.success
+    }}, { client })
+    userStore = rootStore.user
+  })
+
+  afterEach(function () {
+    nock.cleanAll()
+  })
 
   it('should exist', function () {
     expect(userStore).to.be.ok()
@@ -33,15 +92,48 @@ describe('stores > User', function () {
     expect(userStore.display_name).to.equal(user.display_name)
   })
 
-  it('should clear the user', function () {
+  it('should clear the user', async function () {
+    userStore.set(user)
     expect(userStore.id).to.equal(user.id)
     expect(userStore.login).to.equal(user.login)
     expect(userStore.display_name).to.equal(user.display_name)
+    await when(() => userStore.loadingState === asyncStates.success)
 
     userStore.clear()
 
     expect(userStore.id).to.be.null()
     expect(userStore.login).to.be.null()
     expect(userStore.display_name).to.be.null()
+  })
+
+  describe('with an existing user session', function () {
+    
+    it('should refresh project preferences for the same user', async function () {
+      userStore.set(user)
+      const { personalization } = userStore
+
+      expect(userStore.id).to.equal(user.id)
+      await when(() => personalization.projectPreferences.loadingState === asyncStates.success)
+
+      userStore.set(user)
+      expect(userStore.personalization.projectPreferences.loadingState).to.equal(asyncStates.success)
+      expect(rootStore.appLoadingState).to.equal(asyncStates.success)
+    })
+
+    it('should load project preferences for new users', async function () {
+      userStore.set(user)
+      const { personalization } = userStore
+
+      expect(userStore.id).to.equal(user.id)
+      await when(() => personalization.projectPreferences.loadingState === asyncStates.success)
+
+      userStore.set({
+        display_name: 'Worf',
+        id: '2',
+        login: 'zootester2'
+      })
+      expect(userStore.personalization.projectPreferences.loadingState).to.equal(asyncStates.loading)
+      expect(rootStore.appLoadingState).to.equal(asyncStates.loading)
+    })
   })
 })
--- a/packages/app-project/stores/User/User.spec.js
+++ b/packages/app-project/stores/User/User.spec.js
@@ -2,12 +2,9 @@ import asyncStates from '@zooniverse/async-states'
 import * as client from '@zooniverse/panoptes-js'
 import { expect } from 'chai'
 import { when } from 'mobx'
-import { getSnapshot } from 'mobx-state-tree'
 import nock from 'nock'
-import sinon from 'sinon'
 
 import Store from '@stores/Store'
-import initStore from '@stores/initStore'
 
 describe('stores > User', function () {
   const user = {
--- a/packages/lib-classifier/src/components/Classifier/Classifier.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.js
@@ -9,13 +9,12 @@ import Layout from './components/Layout'
 import ModalTutorial from './components/ModalTutorial'
 
 function Classifier({
-  canPreviewWorkflows = false,
   locale,
   onError = () => true,
   showTutorial = false,
   subjectID,
   subjectSetID,
-  workflowSnapshot,
+  workflowSnapshot = null,
 }) {
   const classifierStore = useStores()
   const { workflows, projects } = classifierStore
@@ -57,9 +56,9 @@ function Classifier({
     if (workflowID) {
       console.log('starting new subject queue', { workflowID, subjectSetID, subjectID })
       workflows.setResources([workflowSnapshot])
-      workflows.selectWorkflow(workflowID, subjectSetID, subjectID, canPreviewWorkflows)
+      workflows.selectWorkflow(workflowID, subjectSetID, subjectID)
     }
-  }, [canPreviewWorkflows, subjectID, subjectSetID, workflowID, workflows])
+  }, [subjectID, subjectSetID, workflowID, workflows])
 
   useEffect(function onWorkflowStringsChange() {
     if (workflowStrings) {
@@ -86,7 +85,6 @@ function Classifier({
 }
 
 Classifier.propTypes = {
-  canPreviewWorkflows: PropTypes.bool,
   locale: PropTypes.string,
   onError: PropTypes.func,
   showTutorial: PropTypes.bool,
--- a/packages/lib-classifier/src/components/Classifier/Classifier.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.js
@@ -17,8 +17,7 @@ function Classifier({
   workflowSnapshot = null,
 }) {
   const classifierStore = useStores()
-  const { workflows, projects } = classifierStore
-  const project = projects.active
+  const { workflows } = classifierStore
   const workflowID = workflowSnapshot?.id
   const workflowStrings = workflowSnapshot?.strings
   let workflowVersionChanged = false
--- a/packages/lib-classifier/src/components/Classifier/Classifier.js
+++ b/packages/lib-classifier/src/components/Classifier/Classifier.js
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types'
 import { useEffect } from 'react';
 import i18n from '../../translations/i18n'
 
-import { usePanoptesUser, useProjectPreferences, useProjectRoles } from '@hooks'
+import { useProjectPreferences, useProjectRoles } from '@hooks'
 import Layout from './components/Layout'
 import ModalTutorial from './components/ModalTutorial'
 
@@ -16,14 +16,14 @@ export default function Classifier({
   showTutorial = false,
   subjectID,
   subjectSetID,
+  userID,
   workflowSnapshot,
 }) {
 
   const project = classifierStore.projects.active
   const workflowID = workflowSnapshot?.id
   const workflowStrings = workflowSnapshot?.strings
-  const user = usePanoptesUser()
-  const projectRoles = useProjectRoles(project?.id, user?.id)
+  const projectRoles = useProjectRoles(project?.id, userID)
   let workflowVersionChanged = false
 
   if (workflowSnapshot) {
@@ -49,7 +49,7 @@ export default function Classifier({
     }
   }
 
-  const upp = useProjectPreferences(project?.id, user?.id)
+  const upp = useProjectPreferences(project?.id, userID)
 
   const uppLoading = upp === undefined
   const { userProjectPreferences } = classifierStore
@@ -68,7 +68,7 @@ export default function Classifier({
     }
   }
 
-  const canPreviewWorkflows = adminMode && user?.admin ||
+  const canPreviewWorkflows = adminMode ||
     projectRoles.indexOf('owner') > -1 ||
     projectRoles.indexOf('collaborator') > -1 ||
     projectRoles.indexOf('tester') > -1
@@ -122,8 +122,10 @@ Classifier.propTypes = {
   showTutorial: PropTypes.bool,
   subjectSetID: PropTypes.string,
   subjectID: PropTypes.string,
+  userID: PropTypes.string,
   workflowSnapshot: PropTypes.shape({
     id: PropTypes.string,
+    strings: PropTypes.object,
     version: PropTypes.string
   })
 }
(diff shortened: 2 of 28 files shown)


eatyourgreens:
Adding the 'bug' label to this because it fixes a small bug in the classifier. When you are already logged in, on the Classify page, `usePanoptesUser` in the classifier returns null when the `Classifier` component first loads.

eatyourgreens:
You can try out a production build of this branch here:
https://fe-project-branch.preview.zooniverse.org/projects/zooniverse/gravity-spy

goplayoutside3:
In addition to the above comment, I clicked on https://fe-project-branch.preview.zooniverse.org/projects/zooniverse/gravity-spy, clicked on the first workflow and immediately saw 2 queue requests in the network tab

eatyourgreens:
Yeah, I traced that bug to a hook that’s running twice, so this PR won’t fix it. I’ll remove it. 

https://github.com/zooniverse/front-end-monorepo/issues/4875#issuecomment-1607787426

goplayoutside3:
Is #4845 closed in this PR too? I logged the `user` from `usePanoptesUser` in the Classifier component, and when I navigate to Planet Hunters TESS's classify page, `user` is `null` while some of the other functions in the component are running. Should I be looking for something else?

Screenshot 2023-06-27 at 1 59 34 PM

eatyourgreens:
Hmmm, that sounds like it shouldn’t happen. My reading of the auth client is that you should be logged in almost immediately once you have an access token. 

When I load the Gravity Spy Level 1 workflow, while logged in, the classifier loads and shows a subject almost immediately. 
https://fe-project-branch.preview.zooniverse.org/projects/zooniverse/gravity-spy/classify/workflow/1610

I’m not sure if it’s faster than staging. I’d expect it to be faster than if the classifier gets the user from `/api/me` when it loads. 
https://frontend.preview.zooniverse.org/projects/zooniverse/gravity-spy/classify/workflow/1610

eatyourgreens:
I thought that `auth.checkCurrent()` calls `/api/me` each time it runs, but that turns out not to be true. See https://github.com/zooniverse/panoptes-javascript-client/issues/207.

eatyourgreens:
I've rebased this, so that it's still ahead of the latest `production-release` and rewritten user auth in the classifier. The Panoptes user is now checked in `ClassifierContainer` and `usePanoptesUser` returns a `loading` flag while the API is being checked. The `Classifier` component isn't rendered until loading is complete. I think that should fix #4875.

eatyourgreens:
I can't reproduce either #4875 or #4946 on this branch, so I'm marking those as fixed. Checked with Gravity Spy: Level 1: https://fe-project-branch.preview.zooniverse.org/projects/zooniverse/gravity-spy/classify/workflow/1610

`ClassifierContainer` waits for sign-in to complete (either with an existing valid token or from the Panoptes API) before mounting the classifier, so #4845 should be fixed too.

packages/lib-classifier/src/components/Classifier/Classifier.js
```diff
@@ -81,22 +53,20 @@ export default function Classifier({
...
 
   useEffect(function onURLChange() {
-    const { workflows } = classifierStore
     if (workflowID) {
       console.log('starting new subject queue', { workflowID, subjectSetID, subjectID })
       workflows.setResources([workflowSnapshot])
-      workflows.selectWorkflow(workflowID, subjectSetID, subjectID, canPreviewWorkflows)
+      workflows.selectWorkflow(workflowID, subjectSetID, subjectID)
     }
-  }, [canPreviewWorkflows, subjectID, subjectSetID, workflowID])
```
eatyourgreens, line 90: The `WorkflowStore` model needs updating, to either remove `canPreviewWorkflows` or set the default to `true`. Otherwise, previewing inactive workflows won't work.

eatyourgreens: The tests that were removed from `Classifier` need to be moved up to `ClassifierContainer` too.

eatyourgreens:
Tests are passing now, so this ready for review.

packages/app-project/stores/User/User.spec.js
```diff
@@ -1,21 +1,80 @@
 import asyncStates from '@zooniverse/async-states'
+import * as client from '@zooniverse/panoptes-js'
 import { expect } from 'chai'
+import { when } from 'mobx'
 import { getSnapshot } from 'mobx-state-tree'
+import nock from 'nock'
 import sinon from 'sinon'
 
 import Store from '@stores/Store'
 import initStore from '@stores/initStore'
```
goplayoutside3, line 10:
```suggestion
import { when } from 'mobx'
import nock from 'nock'

import Store from '@stores/Store'
```

packages/lib-classifier/src/components/Classifier/Classifier.js
```diff
@@ -1,33 +1,29 @@
...
   subjectID,
   subjectSetID,
-  workflowSnapshot,
+  workflowSnapshot = null,
 }) {
-
-  const project = classifierStore.projects.active
+  const classifierStore = useStores()
+  const { workflows, projects } = classifierStore
+  const project = projects.active
```
goplayoutside3, line 21:
```suggestion
  const { workflows } = classifierStore
```

packages/lib-classifier/src/components/Classifier/ClassifierContainer.js
```diff
@@ -1,3 +1,4 @@
+import asyncStates from '@zooniverse/async-states'
```
goplayoutside3, line 1:
```suggestion
```

packages/lib-classifier/src/components/Classifier/ClassifierContainer.js
```diff
@@ -143,11 +170,15 @@ ClassifierContainer.propTypes = {
...
   onSubjectReset: PropTypes.func,
   onToggleFavourite: PropTypes.func,
   project: PropTypes.shape({
     id: PropTypes.string.isRequired
   }).isRequired,
   showTutorial: PropTypes.bool,
-  theme: PropTypes.object
+  subjectID: PropTypes.string,
+  subjectSetID: PropTypes.string,
+  theme: PropTypes.object,
```
goplayoutside3, line 182:
```suggestion
```

`mode` is no longer a prop either.

packages/lib-classifier/src/components/Classifier/ClassifierContainer.spec.js
```diff
@@ -0,0 +1,715 @@
...
+          authClient={authClient}
+          project={projectSnapshot}
+          workflowID={workflowSnapshot.id}
+        />,
+        {
+          wrapper: withGrommet()
+        }
+      )
+      const taskTab = await screen.findByRole('tab', { name: 'TaskArea.task'})
+      const tutorialTab = screen.queryByRole('tab', { name: 'TaskArea.tutorial'})
```
goplayoutside3, line 110:
```suggestion
```

These two are fetched eight times in the spec file, but not used. Is that intentional or might be left over from moving tests around?

packages/lib-classifier/src/components/Classifier/hooks/usePanoptesUserSession.js
```diff
@@ -0,0 +1,19 @@
+import {
+  usePanoptesUser,
+  useProjectPreferences,
+  useProjectRoles
+} from '@hooks'
+
+export default function usePanoptesUserSession({ authClient, projectID }) {
+  const { user, loading: userLoading } = usePanoptesUser(authClient)
```
goplayoutside3, line 8: Non-blocking: Because `userProjectPreferences()` and `useProjectRoles()` return an `isLoading` variable, would it make sense for `usePanoptesUserSession()` and `usePanoptesUser()` to also return `isLoading` rather than "loading" for consistency? Or perhaps the other way around changing the former two to `loading`.

packages/lib-classifier/src/hooks/usePanoptesAuth.js
```diff
@@ -3,8 +3,7 @@ import { useEffect, useState } from 'react'
 import { useStores } from '@hooks'
```
goplayoutside3, line 3:
```suggestion
```

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
```diff
@@ -39,7 +39,7 @@ const SubjectViewer = types
 
   .views(self => ({
     get disableImageToolbar () {
-      const subject = getRoot(self).subjects.active
+      const subject = tryReference(() => getRoot(self).subjects?.active)
```
goplayoutside3, line 42: Question: Just for my understanding, what does the change here implement?

goplayoutside3 (approved):
Looks good! I re-tested each of the Issues tagged here. The dev classifier runs as expected, as do the Zooniverse content pages and classifier storybook.

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
eatyourgreens: `subjects.active` is a reference (it's actually the subject ID.) References should be accessed via `tryReference` to avoid errors. 

https://mobx-state-tree.js.org/API/#tryreference

eatyourgreens: The mock store errored on this line, because `subjects.active` isn't always a valid reference.

packages/lib-classifier/src/components/Classifier/ClassifierContainer.spec.js
eatyourgreens: They're testing that the tabs have loaded, but I think that might be redundant.

eatyourgreens: Removing them per your suggestion will fail the tests, but there might be a better way to wait for the workflow to have loaded and the first subject to be ready.

packages/lib-classifier/src/components/Classifier/ClassifierContainer.spec.js
```diff
@@ -0,0 +1,715 @@
...
+      const tabPanel = screen.queryByRole('tabpanel', { name: '1 Tab Contents'})
```
eatyourgreens, line 111:
```suggestion
      const tabPanel = await screen.findByRole('tabpanel', { name: '1 Tab Contents'})
```

Maybe change this to wait for the first task to be ready?

packages/lib-classifier/src/components/Classifier/ClassifierContainer.spec.js
goplayoutside3: A note sounds good too! On code read I was just confused as to why they're grabbed from the screen, but then not used in any assertions

packages/lib-classifier/src/hooks/usePanoptesUser.js
```diff
@@ -20,5 +42,5 @@ export default function usePanoptesUser() {
     }
   }, [authClient])
 
-  return user
+  return { user, loading }
```
eatyourgreens, line 45:
```suggestion
  return { user, isLoading: loading }
```
Change this for consistency with `useSWR`.

packages/lib-classifier/src/components/Classifier/ClassifierContainer.spec.js
eatyourgreens: Isn't that true of the `Classifier` tests too?

eatyourgreens: Reading back through these tests, I don't test that `/subjects/queued` is only called once (#4875.) I'll fix that too.

eatyourgreens: All fixed now. I added a second mock for `/subjects/queued`, and expect it not to be called.

packages/lib-classifier/src/components/Classifier/hooks/usePanoptesUserSession.js
eatyourgreens: I've updated `usePanoptesUser` to use a similar API to [the `useSWR` API.](https://swr.vercel.app/docs/api)

PR #4894, opened 2023-06-27
Implement Custom 404 page for project pages

kieftrav:

Toward: https://github.com/zooniverse/front-end-monorepo/issues/4257

## Describe your changes
- Update default 404 page to incorporate changes
- Create custom 404 component to handle "project not found" errors
- Incorporates rotating background images and translations

## How to Review
- General 404: https://local.zooniverse.org:3000/projects/test404
- Missing Classification Workflow: https://local.zooniverse.org:3000/projects/kieftrav/freehand-line-test/classify/workflow/403
- Existing Classification Workflow: https://local.zooniverse.org:3000/projects/kieftrav/freehand-line-test/classify/workflow/3691?env=staging

** Still working on test and various edge cases / permutations checking. Fully working POC

# Checklist

## General

## General UX

## Post-merge

### DIFF
--- a/packages/app-project/pages/404.js
+++ b/packages/app-project/pages/404.js
@@ -1,14 +1,17 @@
+import { ThemeContext } from 'grommet'
 import { ZooHeaderWrapper } from '@components'
 import { ZooFooter } from '@zooniverse/react-components'
 import { Box, Image, Heading, Text, Paragraph, Anchor } from 'grommet'
 import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
 import { useTranslation } from 'next-i18next'
+import theme404 from './../src/helpers/404theme.js'
 import styled from 'styled-components'
 
 
 const ContainerBox = styled(Box)`
   position: relative;
   color: white;
+  background-color: #000;
   z-index: 1;
 `
 
@@ -47,30 +50,26 @@ export default function Error404({
   const { t } = useTranslation('components')
 
   return (
-    <>
+    <ThemeContext.Extend value={theme404}>
       <ZooHeaderWrapper locale={locale} />
       <ContainerBox
         width="100%"
-        height="50vh"
+        height="80vh"
         alignContent="center"
-        background="#1e1e1e"
         justify="center"
-        style={{ backgroundColor: '#000' }}
       >
         <Overlay
           background={`url("${backgroundURL}")`}
           width="100%"
-          height="50vh"
+          height="80vh"
           justify="center"
-          style={{ opacity: '50%' }}
         />
-
         <PageContent
           alignContent="center"
           justify="center"
           textAlign="center"
         >
-          <Paragraph textAlign="center">
+          <Paragraph textAlign="center" style={{ margin: 0 }}>
             <Image
               id="404-logo"
               fit="contain"
@@ -80,31 +79,37 @@ export default function Error404({
               src={`/projects/assets/logoWhite404.png`}
             />
           </Paragraph>
-          <Heading level="2" textAlign="center" style={{ maxWidth: '100%' }}>{t('404.heading')}</Heading>
+          <Heading
+            level="1"
+            textAlign="center"
+            margin={{ top: '20px', bottom: '20px' }}
+            style={{ maxWidth: '100%' }}
+          >{t('404.heading')}</Heading>
           <Paragraph textAlign="center">
-            {t('404.message')}
-            <br />
-            <Anchor
-              id="404-to-project-home"
-              color="white"
-              weight="bold"
-              href={`/`}
-              label={t('404.returnHome')}
-            />
-            <br />
-            {t('404.or')}
-            <br />
-            <Anchor
-              id="404-to-project-home"
-              color="white"
-              weight="bold"
-              href={`/projects`}
-              label={t('404.findNewProject')}
-            />
+            <Text style={{ fontStyle: 'italic', display: 'block', marginBottom: '30px' }}>{t('404.message')}</Text>
+            <Text style={{ display: 'block', marginTop: '30px', marginBottom: '15px' }}>• <Anchor
+                id="404-to-project-home"
+                color="white"
+                weight="bold"
+                href={`/`}
+                label={t('404.returnHome')}
+                style={{ textDecoration: 'underline' }}
+              />
+            </Text>
+            <Text style={{ display: 'block', marginTop: '15px' }}>•&nbsp;
+              <Anchor
+                id="404-to-project-home"
+                color="white"
+                weight="bold"
+                href={`/projects`}
+                label={t('404.findNewProject')}
+                style={{ textDecoration: 'underline' }}
+              />
+            </Text>
           </Paragraph>
         </PageContent>
       </ContainerBox>
       <ZooFooter locale={locale} />
-    </>
+    </ThemeContext.Extend>
   )
 }
--- a/packages/app-project/public/locales/en/components.json
+++ b/packages/app-project/public/locales/en/components.json
@@ -1,10 +1,9 @@
 {
   "404": {
-    "heading": "Nothing here",
+    "heading": "Nothing here.",
     "message": "The page you're looking for no longer exists.",
-    "returnHome": "Return to the home page",
-    "or": "or",
-    "findNewProject": "find a new project to explore"
+    "returnHome": "Return to the homepage",
+    "findNewProject": "Find a new project to explore"
   },
   "Announcements": {
     "AuthenticationInvitation": {
--- a/packages/app-project/src/helpers/404theme.js
+++ b/packages/app-project/src/helpers/404theme.js
@@ -1,0 +1,65 @@
+const theme = {
+  heading: {
+    level: {
+      1: {
+        small: {
+          size: '37.5px',
+        },
+        medium: {
+          size: '48px',
+        },
+        large: {
+          size: '80px',
+        },
+        xlarge: {
+          size: '100px',
+        }
+      },
+      2: {
+        small: {
+          size: '18px',
+        },
+        medium: {
+          size: '24px',
+        },
+        large: {
+          size: '37.5px',
+        },
+        xlarge: {
+          size: '48px',
+        }
+      },
+    },
+  },
+  // we're treating paragraph and text the same typographically
+  paragraph: {
+    small: {
+      size: '14px',
+    },
+    medium: {
+      size: '16px',
+    },
+    large: {
+      size: '18px',
+    },
+    xlarge: {
+      size: '20px',
+    }
+  },
+  text: {
+    small: {
+      size: '14px',
+    },
+    medium: {
+      size: '16px',
+    },
+    large: {
+      size: '18px',
+    },
+    xlarge: {
+      size: '20px',
+    }
+  }
+}
+
+export default theme


packages/app-project/pages/[panoptesEnv]/[owner]/[project]/about/education.js
```diff
@@ -12,12 +12,12 @@ export async function getStaticProps({ locale, params }) {
   const pageTitle = page?.strings?.title ?? 'Education'
 
   return {
-    notFound,
```
Comment (eatyourgreens) [inline packages/app-project/pages/[panoptesEnv]/[owner]/[project]/about/education.js:15]: If you remove `notFound`, you’ll need to manually set the response code to 404. 

Otherwise, I think you’re serving a regular project page, with a 200 response, but the content says 404. Search engines will index that as a regular page, unless the status code is 404.

Comment (eatyourgreens) [reply packages/app-project/pages/[panoptesEnv]/[owner]/[project]/about/education.js:15]: You might be able to set the response code by throwing an error in the `App` component. I think that’s how we send 500 Server Error when Page components error.

eatyourgreens:
We’ve already got `public/assets` for static images. I think `https://www.zooniverse.org/images` was already in use, but I don’t remember. 

That URL is proxied through to the NextJS app here, so probably best not to change it. 
https://github.com/zooniverse/static/blob/f251e3d9e08eaab1d597dafa7babec65056e3e30/nginx-fem-staging-redirects.conf#L6-L13

That URL is also handled by the project app middleware here:
https://github.com/zooniverse/front-end-monorepo/blob/ad4ef6062b6deeb23382f7eea32f01753f45aa86/packages/app-project/middleware.js#L13-L18

packages/app-project/pages/404.js
```diff
@@ -0,0 +1,74 @@
...
+  z-index: -2;
+  
+  div.image {
+    position: fixed;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+
+    background-image: url("/images/background${randomImage}.png");
```
eatyourgreens, line 26:
```suggestion
    background-image: url("/projects/assets/background${randomImage}.png");
```


Related to my comments about our proxy setup for the NextJS apps. `https://www.zooniverse.org/Images` will try to load the image from PFE. 

Static images for the project app are in `public/assets`, and the URL is `/projects/assets`.

packages/app-project/pages/404.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+`
+
+export default function Error404({ project404Fragment = '' }) {
+  return (
+    <Container>
+      <div className="image"></div>
+      <div className="message">
+        <h1><img src="/images/logoWhite404.png" alt="404" aria-label="404" /></h1>
```
eatyourgreens, line 69:
```suggestion
        <h1><img src="/projects/assets/logoWhite404.png" alt="404" /></h1>
```

packages/app-project/server/server.js
```diff
@@ -31,6 +34,18 @@
   server.use(Sentry.Handlers.requestHandler())
   setLogging(server)
 
+  server.get('/images/:fileName', (req, res) => {
+    const [fileName, fileType] = req.params.fileName.split('.');
+    const filePath = path.join(process.cwd(), `/public/images/${fileName}.${fileType}`);
+    try {
+      const imageBuffer = fs.readFileSync(filePath);
```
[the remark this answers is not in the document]
eatyourgreens: This actually opens us up to directory traversal attacks, via the URL in the browser location bar. It would be interesting to see if this allows an end user to display `next.config.js` in their browser, by editing an image URL to point to the config file. 

Since NextJS already has built-in support for static files in the `public` directory, you don’t need to write your own custom handler for images.

eatyourgreens:
Regarding the responses to send back from missing pages, I made a list of potential responses in this discussion. 

https://github.com/zooniverse/front-end-monorepo/discussions/2446

For this PR, it’s definitely worth considering how the changes affect search engines and SEO.

packages/app-project/server/server.js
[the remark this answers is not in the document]
kieftrav: @eatyourgreens - good catch on this. Was able to reuse the /projects/assets URL.

packages/app-project/pages/[panoptesEnv]/[owner]/[project]/about/education.js
Comment (kieftrav) [reply packages/app-project/pages/[panoptesEnv]/[owner]/[project]/about/education.js:13]: This took a lot of time to debug but the way NextJS operates is that returning the notFound value forces it to display only the 404 page... Which means no header bar, no current state, nothing being passed along. Therefore, to get app-like behavior (essentially the nav bar at the top) we have to return a 200 status code. To mitigate the search engines indexing the pseudo-404 page I've added a meta tag that tells robots (like search engines) to not index the page - which should result in no issues related to search engine indexing and SEO.

packages/app-project/pages/404.js
kieftrav: Incorporated

packages/app-project/pages/404.js
kieftrav: Incorporated

kieftrav:
@eatyourgreens - gonna bring the 404-vs-200 status code + SEO implications  up to the top-level conversation (had address within a thread above).. Next.js is very strict about 404 page generation and requires it to be a compiled-on-server-load file. The impact is there cannot be any state-oriented display (no header bar, no dynamic URL's, etc). The way I found to mitigate this is to have the custom 404 page return a status code of 200 and, when it's a 404 page, include the meta tag that prevents search engines from indexing the page.

This should maintain the current status quo in that pages that should be indexed will be indexed and the new pseudo-404 pages that return a status code of 200 will have the meta tag that says not to index, therefore not affecting SEO and indexed pseudo-404 pages.

eatyourgreens:
The server must be configured to send a 404 response I'm afraid. With NextJS that means returning `notFound: true` or throwing an error and setting the status to 404 on the error page here:

https://github.com/zooniverse/front-end-monorepo/blob/128aa8776f01288c26a26a88d75b1de024883e42/packages/app-project/pages/_app.js#L99

Interestingly, what you've done here is allow the server to build a page for any project URL, which means it can be DDoS'ed by sending it requests for pages on projects that don't exist. It will build pages for every one of those projects and cache the resulting HTML. 

That's why NextJS has a static 404 page by default. Requests that 404 send back the same page, rather than building a new page for every non-existent URL.

goplayoutside3:
Based on the discussion so far, I have a proposed path forward and Jim and Travis let me know what you think! I have additional code comments on some css styling, but would like to know your plan on the following ideas first:

1. My recommendation is to separate handling of the root 404 page into its own PR (or reduce this PR), get that page merged and deployed, and then continue design/dev discussion on possible UX handling of project-specific 404 behavior. I agree with Jim’s point that in the goal to create a project-specific UX for 404 pages, we’re adding a somewhat hacky solution to NextJS’s default behavior and end up building pages that immediately become stale. I do think rendering helpful links for volunteers who end up on a non-existent page will be immensely helpful to them, but we can be strategic about messaging plus sending them back to Zooniverse’s homepage if needed.

2. In the root 404 page, can the ZooHeader and ZooFooter be included? I ran this branch locally, and when I hit the link to https://local.zooniverse.org:3000/projects/test404, user information is fetched even if project information is not.

eatyourgreens:
Sounds good. 

One other note about the changes here. The project slug is never validated, as far as I can tell, so a slug like `zoonivers/gravity-spy` will build new project pages with that slug. As noted above, that makes it super easy to DDoS the NextJS app, by simply sending it requests for non-existent projects until the server runs out of disk space. 

If the project slug isn't valid, you definitely want to return `notFound: true` or throw a custom error so that our custom error page handles the response.

kieftrav:
Thanks @goplayoutside3 for the suggested path forward. I'll get this split into two PR's.

In the root 404 page, the ZooHeader and ZooFooter cannot be included because they incorporate user/app state. The page is compiled on server start and cannot inject any state of any kind. I'm sure we could refactor the header and footer to have a stateless version - this is particularly true of the footer - but it's not doable in its current form. 

Essentially the 404 page that Next.js expects has no javascript that runs on the client side - it's basically a compiled HTML+CSS page.

eatyourgreens:
There's also the option of passing `statusCode` and `title` as page props here, to build a more dynamic error page for individual projects.
https://github.com/zooniverse/front-end-monorepo/blob/77e3b71a0dd92ceaa7f80667e527175e32c3ee48/packages/app-project/pages/_app.js#L72-L79

eatyourgreens:
You can also use `getStaticProps` in `pages/404.js`, so builds aren’t totally static. 
https://nextjs.org/docs/pages/building-your-application/routing/custom-error#404-page

eatyourgreens:
We do get requests like `/projects/wp/admin.php`, from people looking for security vulnerabilities, or just typos in the project slug eg. `/projects/zooinverse/gravity-spy`. In those cases, your link to the project home page will link to `/projects/undefined` so do validate the project before generating a project-specific error page.

eatyourgreens:
Here's a very quick page with header and footer that I hacked together on the `base-404` branch. This is only five minutes work, so needs more work to fix the CSS (for example, none of the links in the header are clickable), but it shows that it is possible to insert the Zooniverse header and footer into a static page that's built at build time. 

I moved `randomImage` into the page body so that it runs in the browser on each render. Depends whether you want a different image for each visitor, I think.

```js
import { ZooHeader, ZooFooter } from '@zooniverse/react-components'

function 404Page() {
  let randomImage = Math.round(Math.random() * 8) + 1 // 1-9
  return (
    <>
      <ZooHeader />
      <PageBody image={randomImage} />
      <ZooFooter />
    </>
  )
}
```

Our team and publications pages are also static builds at build time, so you can also look at those for examples of how to insert the header and footer into a static page.

If we want sign in to be available, then I think the error page will need to be wrapped in a context provider so that the user store actions are available in the browser. Personally, I think I'd be ok with Sign In and Register not being shown on the error page.


https://github.com/zooniverse/front-end-monorepo/assets/59547/80b01a4c-8fe1-4bde-b796-52f96117d666

eatyourgreens:
If you look in the browser console while the 404 page loads, you can see the current user session being fetched from Panoptes, so user state is available in the browser.

Console logs for the static 404 page, showing a user session being fetched from the Panoptes API.

kieftrav:
Thanks @eatyourgreens for the quick prototyping. I had tried that early on and in my local testing I've been using Safari and doesn't render either header or footer component and instead throws a bunch of console errors. As discussed separately, I'll do some sleuthing on why there is a browser difference after checking out your prototype branch.

Screenshot 2023-08-10 at 8 11 12 AM

eatyourgreens:
Here's #5148 in Safari 16:

The new 404 page, with a header and footer, in Safari 16.

eatyourgreens:
>I'm sure we could refactor the header and footer to have a stateless version

As a quick fix, maybe edit `ZooHeader` so that it only shows 'Sign In' and 'Register' if the corresponding callbacks are passed in. The publish that as a new patch version of `@zooniverse/react-components`.

eatyourgreens:
Here’s another URL from our logs that can be used for testing. 
https://zooniverse-27.sentry.io/issues/4387092521/?alert_rule_id=752481&alert_type=issue&project=1492691&referrer=issue_alert-slack

packages/app-project/src/components/Custom404/Custom404.js
```diff
@@ -0,0 +1,58 @@
...
+      <Text>{t('404.pageDoesntExist')}</Text>
+      <Paragraph size="medium">
+        <Text>{t('404.returnToProject')}</Text>
+        <Anchor
+          id="404-to-project-home"
+          color="white"
+          weight="bold"
+          href={`/projects/${projectSlug}`}
+          label={` ${t('404.projectHomePage')} `}
+        />
```
eatyourgreens, line 26: This link will need to use `next/link` in a production app.

packages/app-project/pages/_app.js
```diff
@@ -78,13 +79,18 @@ function MyApp({ Component, pageProps }) {
       )
     }
 
+    if (showCustom404) {
```
eatyourgreens, line 82: You can pass `statusCode=404` in the page props, and take advantage of the existing error block (lines 73 – 80) to show a custom 404 page for individual projects.

https://github.com/zooniverse/front-end-monorepo/blob/677c528b6a2279e71c1241fbab57673d2c5838cd/packages/app-project/pages/_app.js#L73-L80

eatyourgreens: So, when a project doesn't exist, return `notFound: true` to show the default 404 page for the app.

When a project does exist, but a project page doesn't exist for that project, return something like this to use a custom 404 error page for a specific project. The project will be passed down in page props, so you can then pass it into the error page as a prop.

```js
{
  notFound: false,
  props: {
    statusCode: 404,
    //… other page props here
  }
}
```

seanmiller26:
Noting some styling changes to fix the hierarchy of copy on the page: 

Screen Shot 2023-08-25 at 11 35 52 AM

Screen Shot 2023-08-25 at 11 36 35 AM


Font styling changes:

- [ ] "The page you're looking for" - Medium italic font
- [ ] "homepage" - one word
- [ ] The two links separated to account for translations issues
- [ ] Delete "or"
- [ ] Font sizing should be increased to the next default size for all lines. 
          
It may be worth a discussion of our default Grommet styling as I am unsure on how we decided on these values and think it would be helpful for the team to get some clarification.

- [ ] Background color of #000000 and the random image opacity set to 50%

Layout changes:
See above for correct spacing

kieftrav:
@seanmiller26 - deploying a branch build now with all the changes. The only thing that hasn't been integrated is a bumping of the text sizing as that's a little unclear to me how we could do that. Will chat with @goplayoutside3 about it next week when she's back in office.

seanmiller26:
@kieftrav OK to wait for the text discussion with Delilah, also happy to be a part of that conversation. 

See below for further tweaks:

Screen Shot 2023-09-01 at 10 57 45 AM

- [ ] Reduce bottom margin of "Nothing Here" to 20
- [ ] Tighten up the spacing b/w 404 logo and the next line
- When inspecting, I was confused where this extra space is coming from. There is no margin or padding defined. Maybe you can you clarify what's happening there?

eatyourgreens:
I figured out what's causing the styling problems with the live 404 page. It uses invalid HTML. Specifically, it nests a heading inside a paragraph. I'll push a PR to rewrite the page with valid HTML.

eatyourgreens:
- https://github.com/zooniverse/front-end-monorepo/pull/5274

kieftrav:
@seanmiller26 @eatyourgreens - Jim, I was able to merge in and incorporate your changes. Good catch on the invalid DOM. I also fixed the spacing issue between the 404 header image and the "Nothing here." statement as its container is a paragraph and has a margin attached to it (the paragraph was used to center it). All that said, your changes are incorporated and should inspect properly at this point. 

Next step is fixing font size which would be great to do with @goplayoutside3

goplayoutside3:
Documenting what we just discussed via Slack ~

1. Here's a link to the object in our lib-grommet-theme that determines the font size of heading components. The "small", "medium", etc variables belong to an optional `size` prop that can be added to a Heading component in addition to its `level`:
 https://github.com/zooniverse/front-end-monorepo/blob/e54655a2d39a191fa87e86d40ccc8b642e75a67f/packages/lib-grommet-theme/src/index.js#L309-L334

2. Rather than modify the font sizes in the global theme object in this particular PR, here's an example of creating a theme object specific to one component in the FEM codebase. I recommend the separate theme object strategy for now, and then addressing changes to the global Grommet theme object for Headings in a separate PR.

    Theme file: https://github.com/zooniverse/front-end-monorepo/blob/e54655a2d39a191fa87e86d40ccc8b642e75a67f/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/components/theme.js

    https://github.com/zooniverse/front-end-monorepo/blob/e54655a2d39a191fa87e86d40ccc8b642e75a67f/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/components/FlipbookControls.js#L212

seanmiller26:
Alright, the Figma file has been adjusted to be consistent with site-wide standards which we can implement and conform to at a future time. It's a good idea to theme this page specifically as it does have a few unique outliers such as the italic font. [Figma](https://www.figma.com/file/ZuYDEMK56dlnvasBAa5jNB/404-Pages?type=design&node-id=101-70&mode=design)

I took some time to update our design system. Take a look at the typography pages for further guidance on the heading sizes: [Design System](https://www.figma.com/file/HUWCyrjkwgPsGKLXhLGb21/Design-System?type=design&node-id=0-1&mode=design&t=JDfdrAumfGMYDzdF-0)

Let me know if anything else is needed

kieftrav:
@seanmiller26 - got the font sizing implemented from the updated Design System. Thank you. A couple things that I'd appreciate some clarity on:

- What are the breakpoint page sizes between S|M|L|XL?
- What is the line height for each of the variations in S|M|L|XL between H1|H2|Body text?
- Is there any assumed padding / margin for H1|H2|Body text?

seanmiller26:
Hey Travis, if possible I think its ok to use the Grommet defaults for line height and padding. I do think it’s worth having a wider discussion with the team before any of these values are defined. Same with the breakpoint sizes for the time being. The staging looks good as I change the browser window sizing.
But for this PR, ok to push forward and wrap up the 404 page with default values.

packages/app-project/helpers/404theme.js
```diff
@@ -0,0 +1,65 @@
+const theme = {
```
goplayoutside3, line 1: Please move this file into /src/helpers to keep things consolidated.

packages/app-project/helpers/404theme.js
```diff
@@ -0,0 +1,65 @@
...
+          size: '48px',
+        },
+        large: {
+          size: '80px',
+        },
+        xlarge: {
+          size: '100px',
+        }
+      },
+      2: {
```
goplayoutside3, line 18: Is there a Heading level=2 used in the 404 pages?

packages/app-project/pages/404.js
```diff
@@ -47,7 +49,7 @@ export default function Error404({
   const { t } = useTranslation('components')
 
   return (
-    <>
+    <ThemeContext.Extend value={theme404}>
       <ZooHeaderWrapper locale={locale} />
       <ContainerBox
```
goplayoutside3, line 54: I can't directly comment on it, but this ContainerBox has a `background` prop that's being overwritten by `style={{ backgroundColor }}`. I recommend removing `background` if the color is declared in the style object instead.

goplayoutside3: Also the ContainerBox and its Overlay are set at 50vh. The Figma designs look like 100vh - ZooHeader height though. Is the difference from the designs intentional?

goplayoutside3 (commented):
Looks like the custom theme is working well! I have some questions and suggestions before this PR can be merged though. I also recommend editing the title of this PR so when it's squashed and merged to master the title does not inaccurately reflect that code changes.

packages/app-project/helpers/404theme.js
kieftrav: No - I implemented the full set of styles that are possible to create from the Figma file.

packages/app-project/pages/404.js
kieftrav: I can't remember when it was discussed yet the reason it's something smaller than the whole viewport is so that the footer is visible. When it was originally posed Sean didn't create a follow up item to change it. I think we can ask @seanmiller26 to clarify.. Specifically, how much of the viewport should the 404 background image take up compared to the ZooHeader and ZooFooter?

- XL page it takes up the whole viewport minus the header
- L page it takes up ~60%
- M page takes up whole viewport
- S page takes up whole viewport

packages/app-project/pages/404.js
```diff
@@ -1,14 +1,17 @@
...
 import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
 import { useTranslation } from 'next-i18next'
+import theme404 from './../src/helpers/404theme.js'
 import styled from 'styled-components'
 
 
 const ContainerBox = styled(Box)`
   position: relative;
   color: white;
+	background-color: #000;
```
goplayoutside3, line 14: ContainerBox still has a `background` prop. It appears here and is set to '#1e1e1e'.
Screenshot 2023-09-22 at 10 40 19 AM

By moving `background-color` to the styled-component function, this element still has two background colors defined and '#1e1e1e' is being overwritten by '#000':
Screenshot 2023-09-22 at 10 39 21 AM

Apologies if you're still intending to address my initial comment about the `background` prop, but hope this better explains why the `background` prop should be removed from ContainerBox in favor of defining `background-color` as black once.

packages/app-project/pages/404.js
seanmiller26: @kieftrav Unfortunately Figma doesn't support relative units like vh, vw, vmin, vmax, etc, so calculating all of that manually is a bit of a chore. If I can ever specify these values in the future I'll be sure to do that.

Not sure what discussion mentioned the footer being visible, and I don't disagree, but let's go ahead and try to match the Figma design a little closer. Can we try a 80vh to split the difference?

goplayoutside3 (approved):
Looks good!

seanmiller26 (approved):

PR #5014, opened 2023-07-14
Stronger checking for classifier data fetching 

### LINKED ISSUE #5015: Intermittent bug: empty labels for workflow tasks
## Package
lib-classifier 

## Describe the bug
The classifier sometimes loads without any text in the task labels. 

I Fancy Cats, without any text for the question or its answers.
## To Reproduce
I think you basically have to load and reload the classifier until this triggers. 

## Expected behavior
Workflow tasks should be labelled. 

## Additional context
This code checks for `workflowSnapshot` but doesn't check `workflowSnapshot.strings`, so the classifier might render before the workflow strings have been set.
https://github.com/zooniverse/front-end-monorepo/blob/3847048f66f7e59bf1574adc5fbafa0e900891c1/packages/lib-classifier/src/components/Classifier/ClassifierContainer.js#L133-L139

eatyourgreens:
Set the current `userID` to undefined while the user is loading. Check that in both `useProjectPreferences` and `useProjectRoles` so that the possible return values are:
- undefined (resource is loading.)
- null (user isn't logged in.)
- a Panoptes resource (user is logged in.)

Wait until workflow translations have loaded before rendering the classifier (closes #5015.)
 
## Package
lib-classifier

## How to Review
I checked this in the dev classifier. With these changes, you shouldn't see any requests made for either project preferences or project roles until the user is authenticated, and set to either null or a Panoptes user resource.

You shouldn't see GET requests like this, where the user ID is `false` because the user session hasn't authenticated yet.
https://panoptes-staging.zooniverse.org/api/project_preferences?project_id=335&user_id=false&http_cache=true

# Checklist

## General

## General UX

## Refactoring

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/hooks/usePanoptesUserSession.js
+++ b/packages/lib-classifier/src/components/Classifier/hooks/usePanoptesUserSession.js
@@ -6,7 +6,7 @@ import {
 
 export default function usePanoptesUserSession({ authClient, projectID }) {
   const { data: user, isLoading: userLoading } = usePanoptesUser(authClient)
-  const userID = !userLoading && user?.id
+  const userID = userLoading ? undefined : (user?.id || null)
   const { data: upp } = useProjectPreferences({ authClient, projectID, userID })
   const { data: projectRoles } = useProjectRoles({ authClient, projectID, userID })
   const userHasLoaded = userID ?
--- a/packages/lib-classifier/src/components/Classifier/ClassifierContainer.js
+++ b/packages/lib-classifier/src/components/Classifier/ClassifierContainer.js
@@ -136,7 +136,9 @@ export default function ClassifierContainer({
   - the workflow has loaded.
   - the project has been added to the store.
   */
-  const classifierIsReady = userHasLoaded && !!workflowSnapshot && !!classifierStore.projects.active
+  const workflowIsReady = !!workflowSnapshot?.strings
+  const projectIsReady = !!classifierStore.projects.active
+  const classifierIsReady = userHasLoaded && workflowIsReady && projectIsReady
   try {
     if (classifierIsReady) {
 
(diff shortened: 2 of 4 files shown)


packages/lib-classifier/src/components/Classifier/hooks/usePanoptesUserSession.js
```diff
@@ -6,7 +6,7 @@ import {
 
 export default function usePanoptesUserSession({ authClient, projectID }) {
   const { data: user, isLoading: userLoading } = usePanoptesUser(authClient)
-  const userID = !userLoading && user?.id
+  const userID = userLoading ? undefined : user?.id
```
mcbouslog, line 9: I'm not sure I'm thinking about this correctly, but I think 
- if the user is loading we want `userID` to be undefined ✔️ , 
- if the user isn't loading then there are two possible scenarios
  - we have a user in which case we want `userID` to be the user.id ✔️  
  - we want `userID` to be null ✖️ (for an anonymous / logged out user) which I think the following change would do...

```suggestion
  const userID = userLoading ? undefined : user?.id || null
```

eatyourgreens: Ha! I just pushed exactly this change, then saw your comment.

eatyourgreens: But yeah, you’re exactly right. We need to be able to distinguish between a value that’s falsy because it’s still loading and a value that’s falsy because you aren’t actually logged in.

eatyourgreens: This change gets the tests passing again for anonymous users. Without it, you never get past ‘Loading…’.

mcbouslog (approved):
This LGTM!
Tested signed out and signed in. Noted the expected number of `project_preferences` requests. Couldn't replicate empty translations strings issue, but didn't note instance of when testing locally.

PR #5051, opened 2023-07-24
Use constant callbacks in the dev classifier

eatyourgreens:
The dev classifier stubbed callback functions are constants, but the `App` component generates new `onAddToCollection` and `onCompleteClassification` callback functions on each render. This declares them once, so that they are only set once.

## Package
lib-classifier

## How to Review
This is a small change, but you shouldn't see console logs warning that the dev classifier's callbacks have been reset on every render eg. when toggling the theme from dark to light.

# Checklist

## Bug Fix
- [ ] The bug is fixed


mcbouslog (approved):
LGTM! :shipit:

PR #5082, opened 2023-07-28
Refactor daily classification counts

### LINKED ISSUE #5081: Daily classifications count is lost when you go to Talk
https://github.com/zooniverse/front-end-monorepo/blob/2092b81a59e211b6240a47beab8e56d948e8b5c7/packages/app-project/stores/User/UserPersonalization/UserPersonalization.js#L15-L17

Your daily session count is stored in volatile storage here, which means that it isn't saved in the `panoptes-user` snapshot when you leave the Classify page. As a result, it starts again from 0 each time that you return.

One subtle side effect of this bug is that `sessionCount % 5` will always be 0 here, meaning that you'll be nagged about workflow promotion on every subject, if you classify using Done & Talk rather than Done.
https://github.com/zooniverse/front-end-monorepo/blob/2092b81a59e211b6240a47beab8e56d948e8b5c7/packages/app-project/stores/User/UserPersonalization/UserPersonalization.js#L38-L40

`sessionCount` is also used to calculate your total classification count here, leading to the impression that classifications have been lost when the volatile state resets to 0.
https://github.com/zooniverse/front-end-monorepo/blob/2092b81a59e211b6240a47beab8e56d948e8b5c7/packages/app-project/stores/User/UserPersonalization/UserPersonalization.js#L42-L45

eatyourgreens:
- store `user.personalization.sessionCount` in user snapshots, so that it can persist across page visits (closes #5081.)
- only run the graphQL query for daily stats when we're fetching a new user from Panoptes. Existing users (including users stored in browser session storage) will already have their stats stored in the personalisation models.
- increment daily counts separately from session counts, since open tabs can persist across calendar days.
- increment project activity count separately from session counts, to avoid double counting of classifications.
- add `user.personalization.todaysStats` and `user.personalization.todaysCount` for convenience.
- add `stats.increment()` as a convenience method to increment daily counts.

## Package
app-project

- #5081.
- [bug concerning vanishing classifications statistics / resetting total classifications counter](https://www.zooniverse.org/talk/17/3011540) (Zooniverse Talk.)

## How to Review
On a workflow like Gravity Spy Level 1, you should be able to make a few quick classifications, while logged in, leave the page and come back to see your work still recorded.

Logged out, your counts should be recorded while you're on the page, but won't be persisted if you leave the page.

Daily classification statistics for a Gravity Spy volunteer.

# Checklist

## General

## General UX

## Bug Fix
- [x] The bug is fixed

## Refactoring

### DIFF
--- a/packages/app-project/stores/User/UserPersonalization/UserPersonalization.js
+++ b/packages/app-project/stores/User/UserPersonalization/UserPersonalization.js
@@ -10,24 +10,12 @@ const UserPersonalization = types
   .model('UserPersonalization', {
     notifications: types.optional(Notifications, {}),
     projectPreferences: types.optional(UserProjectPreferences, {}),
+    sessionCount: types.optional(types.number, 0),
     stats: types.optional(YourStats, {})
   })
-  .volatile(self => ({
-    sessionCount: 0
-  }))
   .views(self => ({
     get counts() {
-      const todaysDate = new Date()
-      let todaysCount = 0
-      try {
-        if (self.stats.thisWeek.length === 7) {
-          const todaysStats = self.stats.thisWeek.find(stat => stat.dayNumber === todaysDate.getDay())
-          todaysCount = todaysStats.count
-        }
-      } catch (error) {
-        todaysCount = 0
-      }
-      const today = todaysCount + self.sessionCount
+      const today = self.todaysCount
 
       return {
         today,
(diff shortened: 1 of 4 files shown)


mcbouslog (approved):
LGTM!

Confirmed stats increment as expected logged in using Done, Done and Talk, and logged out.

PR #5210, opened 2023-08-22
YourStats: Remove enzyme and convert tests to RTL.

kieftrav:
## Package
app-project

## Describe your changes
Converted tests for the YourStats from Enzyme to RTL. 

## How to Review
[RecentSubjects Story](http://localhost:9001/?path=/story/project-app-screens-classify-your-stats--default)

## General

## Refactoring


packages/app-project/src/screens/ClassifyPage/components/YourStats/YourStats.spec.js
```diff
@@ -1,15 +1,41 @@
...
+describe('Component > YourStats', function () {
+  beforeEach(function () {
+		const YourStatsStory = composeStory(YourStats, Meta)
+		render(<YourStatsStory />)
+	})
 
-let wrapper
+	it('should have a heading', function () {
+    expect(screen.findByText(`Your ${YourStatsStoreMock.project.display_name} Statistics`)).to.be.ok()
+  })
```
shaunanoordin, line 14: **Note 1:** I don't think `findByText()` can be used this way.

findByText() returns a _Promise,_ not the _element_ you're looking for. [(docs)](https://testing-library.com/docs/queries/about/#types-of-queries). So if you replaced the text with "NONSENSE TEXT", then this happens...

```
expect(screen.findByText('NONSENSE TEXT')).to.be.ok()
=> expect(Promise).to.be.ok()
=> a Promise is an object, therefore it's OK
=> test passes
``` 

So this test passes no matter what text you're trying to find. 😱 

**Note 1a:** Also! The current RTL setup _doesn't actually render the translations,..

So if you run `screen.debug()`, you'll see that the heading isn't rendered as `Your Stats Mock Project Statistics` (as you'd see on Storybook), but instead as `Classify.YourStats.title` (i.e. it renders the translation key)

**Suggested Solution**

Option 1: if you need to use findBy...(), make sure to use await.

For example,

```
it('should have a heading', async function () {
  const heading = await screen.findByText('Classify.YourStats.title')
  expect(heading).to.be.ok()
})
```

⭐  Option 2: the simpler way is to use getBy...() _(recommended!)_

getByText() will either find the component you're looking for, or throw an error if _either 0 or > 1 elements_ are found. [(docs)](https://testing-library.com/docs/queries/about/#types-of-queries) For example,

```
it('should have a heading', function () {
  expect(screen.getByText(`Classify.YourStats.title`)).to.be.ok()
})
```

Easy!

packages/app-project/src/screens/ClassifyPage/components/YourStats/YourStats.spec.js
```diff
@@ -1,15 +1,41 @@
...
+	it('should have today\'s classification count ', function () {
+    expect(screen.getAllByText(`${YourStatsStoreMock.user.personalization.counts.today}`).length).to.equal(2)
+  })
+
+	it('should have this week\'s classification total ', function () {
+    expect(screen.queryByText(`${YourStatsStoreMock.user.personalization.counts.total}`)).to.be.ok()
+  })
+
+	it('should have a chart heading', function () {
+    expect(screen.findByText(`${YourStatsStoreMock.project.display_name} Daily Classification Count`)).to.be.ok()
```
shaunanoordin, line 25: See Note 1 - I think findByText() should be replaced by getByText()

Also see Note 1a - I think the text we need to look for is "Classify.YourStats.DailyClassificationsChart.title"

packages/app-project/src/screens/ClassifyPage/components/YourStats/YourStats.spec.js
shaunanoordin: Addendum: when we're looking for exactly one element, then in most cases getByText() works pretty much the same as queryByText(). So using either one of them should be fine, AFAIK.

findByText() however, since it returns a Promise, seems more situational - if you look at the other unit tests in `app-project`, it's findBy...() calls are _only_ used in conjunction with the more complex event or user input-driven tests. (e.g. DropdownNav, LocaleSwitcher, etc)

packages/app-project/src/screens/ClassifyPage/components/YourStats/YourStats.spec.js
```diff
@@ -1,15 +1,41 @@
...
-describe('Component > YourStats', function () {
-  before(function () {
-    wrapper = shallow(<YourStats />)
+	it('should have today\'s classification count ', function () {
+    expect(screen.getAllByText(`${YourStatsStoreMock.user.personalization.counts.today}`).length).to.equal(2)
+  })
+
+	it('should have this week\'s classification total ', function () {
+    expect(screen.queryByText(`${YourStatsStoreMock.user.personalization.counts.total}`)).to.be.ok()
+  })
```
shaunanoordin, line 22: **Note 2:** this test actually fails for me with the error message, "TestingLibraryElementError: Found multiple elements with the text: 100"

This seems pretty straightforward - the text that we're looking for is _too generic._ The 100 can describe either the classifications total (which is what we're looking for), or a number on the y-axis. 🤷 

image

I think a better implementation is to be _very specific_ about what we're supposed to be looking at. i.e. "Find the text next to the text 'Classifications Total', and check that the text says '100'."

However, as to _how_ you'd create such a specific query... erm... gimme a second. Processing, processing...

shaunanoordin: Additionally: simply replacing `expect(getByText('100')).to.be.ok()` with `expect(getAllByText('100')).to.equal(2)` isn't a great solution either, as it's just looking around the rendered screen for the existence of the number 100, without care of the context of how/where that number is used.

Anyway, back to processing a suggested solution...

eatyourgreens: `within` is your friend in situations like these, or chaining RTL queries from a previous query result.

eatyourgreens: Eg. 

https://github.com/zooniverse/front-end-monorepo/blob/ee66328323a5fb37245f027fad8154bedbe77876/packages/lib-classifier/src/components/Classifier/Classifier.spec.js#L88

shaunanoordin: **Suggested Solution**

OK, so we need to be very specific in our query to find that "100 Classifications Total" element.

Option 1: find the text "Classifications Total", then find the "100"

From studying the HTML, we know that the "Classifications Total" text sits in its own div, and the number "100" exists in the div juuust before it. So...

```
it('should have this week\'s classification total ', function () {
  const elem = screen.getByText('Classify.YourStats.totalCount')
  expect(elem.previousElementSibling.textContent).to.equal('100')
})
```

Option 2: add classnames to the stats, so we can easily distinguish the element via code

In YourStats.js, add some descriptive class names:
```
<Stat
  className='your-stats_total'
  label={t('Classify.YourStats.totalCount')}
  value={counts.total}
/>
```

In YourStats.spec.js, look for the element with the classname, then use ['within'](https://testing-library.com/docs/dom-testing-library/api-within/) to query _inside that element_
```
import { within } from '@testing-library/dom'

...

it('should have this week\'s classification total ', function () {
  const elem = document.getElementsByClassName('your-stats_total')[0]
  expect(within(elem).getByText('100')).to.be.ok()
})
```

packages/app-project/src/screens/ClassifyPage/components/YourStats/YourStats.spec.js
```diff
@@ -1,15 +1,41 @@
...
-let wrapper
+	it('should have a heading', function () {
+    expect(screen.findByText(`Your ${YourStatsStoreMock.project.display_name} Statistics`)).to.be.ok()
+  })
 
-describe('Component > YourStats', function () {
-  before(function () {
-    wrapper = shallow(<YourStats />)
+	it('should have today\'s classification count ', function () {
+    expect(screen.getAllByText(`${YourStatsStoreMock.user.personalization.counts.today}`).length).to.equal(2)
```
shaunanoordin, line 17: See Note 2. Whatever solution is implemented to check for "Classifications Total" should also apply to "Classifications Today"

packages/app-project/src/screens/ProjectAboutPage/ProjectAboutPage.spec.js
shaunanoordin, line 1: Changes to this file seem fine, though it kinda goes a wee bit beyond _(what I thought was this PR's intended scope of)_ "YourStats unit test updates".

packages/app-project/src/screens/ProjectAboutPage/components/AboutDropdownNav/AboutDropdownNav.spec.js
shaunanoordin, line 1: Changes to this file seem fine, though it sorta goes a tiny bit beyond _(what I thought was this PR's intended scope of)_ "YourStats unit test updates".

packages/app-project/src/screens/ClassifyPage/components/YourStats/YourStats.spec.js
kieftrav: Thanks for the insight on how Promises are handled within RTL.. I assumed that they were resolved before evaluated :/

Fixed using the translation keys.

packages/app-project/src/screens/ClassifyPage/components/YourStats/YourStats.spec.js
kieftrav: This was a bit more complicated because the `Stat` component does not allow setting data attrbiutes... I ended up going with the class route wherein the classes are named `test-classify-your-stats-today-count` and `test-classify-your-stats-total-count` to indicate that they are used for a testing instead of stylistic purpose.

I'm not a big fan of going too deep into complex querying the internal structure of a component (and in this case a sub-component) for the same reasons RTL doesn't have a getByClass() method... Which is its focused on implementation structure. That all said, I think this is a reasonable path with which to move forward!

packages/app-project/src/screens/ProjectAboutPage/ProjectAboutPage.spec.js
kieftrav: So... This was fun. Once I created the tests for the YourStats these specific tests started failing for not obvious reason. Yet, my hunch is that it has to do with parallelization and rendering timing. This is why the addition of `waitFor` fixed it... It encapsulated the following code in a retry process after the `fireEvent` and created the timing space for it to work again. Yay for parellelization?

packages/app-project/src/screens/ProjectAboutPage/components/AboutDropdownNav/AboutDropdownNav.spec.js
kieftrav: Same reasoning as the `ProjectAboutPage.spec.js` above

shaunanoordin (approved):
## PR Review

package: app-project

This PR updates the unit tests for the YourStats component to use React Testing Library. Goodbye, Enzyme!

Tests overall look good (it's sufficient to check that the component isn't crashing), though I do have some strong suggestions to improve this PR:

- [ ] Test coverage: the "should have today's classification count" (SHTCC) test should check for the value "10"
  - Rationale: currently, the SHTCC test only checks _if the element renders._ A more thorough test would be to check _if the element renders the correct value,_ as provided by the input parameters.
  - Earlier review notes [(Shaun's)](https://github.com/zooniverse/front-end-monorepo/pull/5210#discussion_r1303428597) [(Jim's)](https://github.com/zooniverse/front-end-monorepo/pull/5210#discussion_r1303419816) noted how this could be done using `within`
  - Caveat: I understand that, [as you noted,](https://github.com/zooniverse/front-end-monorepo/pull/5210#discussion_r1303560133), this is some _complex querying into the internal structure of the component_ but I think the complexity tradeoff is worth it to ensure we check for functionality. Otherwise, some clown (me, most likely) might accidentally muck the code so that input=`{daily_count: '5'}` results in output = "Keep up the good work! You've completed 🤡🤡🤡🤡🤡 classifications today"
- [ ] Test coverage: the "should have this week's classification total" test should check for the value "100"
  - Ditto ☝️ 
- [ ] Coding style: instead of using `className` (e.g. className="test-classify-your-stats-today-count"), use `data-testid` instead.
  - Rationale: see below

### Dev Notes

(Following our conversation on Slack)

_Travis:_
Any thoughts on using `data-testid`  instead of adding arbitrary classes when their sole use is for tests? Generally it's discouraged by RTL but it seems to make sense in our case to avoid accessibility concerns

_Shaun:_ 
In short: when in doubt just use `data-testid`

In long, my personal preferences is to...
- Use `data-testid` if we _only_ want to identify components for tests.
- Use `className` if a CSS class name could _also_ help describe the component in the HTML structure

For example, in Community Catalog, I use CSS class names so that the rendered HTML elements share the same names as their component names.
https://github.com/zooniverse/community-catalog/blob/master/src/components/SubjectActionsPanel/SubjectActionsPanel.js#L31

Note that this is _not_ an accessibility thing, just something I like to do since it helps me debug code.

### Status

LGTM, but please consider those suggestions above before merging. ☝️

kieftrav:
@shaunanoordin - thanks for the comprehensive review of the PR review :)

I agree that it's preferred to use `data-testid`. Because of the `Stat` component structure, it does not do anything with [data-testid] prop whereas it does utlize the `className` prop. I think this opens a broader question with component refactoring which I'll make note of for future dev conversations: what does a testable component - especially when nesting/layering components - look like?

PR #5216, opened 2023-08-23
SubjectPreview: Remove enzyme and convert tests to RTL

kieftrav:
## Package
app-project

## Describe your changes
Converted tests for the SubjectPreview from Enzyme to RTL. Includes changes for CollectionsButton and TalkLink.

## How to Review
[SubjectPreview Story](http://localhost:9001/?path=/story/project-app-shared-subject-preview--plain)
[CollectionsButton Story](http://localhost:9001/?path=/story/project-app-shared-collections-button--collections-button)
[TalkLink Story](http://localhost:9001/?path=/story/project-app-shared-talk-link--talk-link)

## General

## Refactoring

### DIFF
--- a/packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.js
+++ b/packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.js
@@ -10,7 +10,7 @@ const defaultSubject = {
   toggleFavorite: () => false,
   locations: []
 }
-function SubjectPreview ({
+function SubjectPreview({
   height = '250px',
   isLoggedIn = false,
   placeholder,
@@ -28,6 +28,7 @@ function SubjectPreview ({
       fill
     >
       <Anchor
+        data-testid={`subject-preview-link-${subject.id}`}
         href={href}
       >
         <Box
@@ -37,6 +38,7 @@ function SubjectPreview ({
           width={width}
         >
           <Media
+            data-testid={`subject-image-${subject.id}`}
             alt={`subject ${subject.id}`}
             height={700}
             placeholder={placeholder}
--- a/packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.mock.js
+++ b/packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.mock.js
@@ -1,0 +1,57 @@
+import Store from '@stores/Store'
+
+export const SubjectPreviewState = {
+  isLoggedIn: true,
+  slug: 'zooniverse/snapshot-serengeti',
+  subjectId: '123'
+}
+
+export const PlainSubjectMock = {
+  favorite: false,
+  id: SubjectPreviewState.subjectId,
+  locations: [
+    {
+      'image/jpeg':
+        'https://panoptes-uploads.zooniverse.org/production/subject_location/48972f7b-8a4d-4f98-a85b-ed3578db75f0.jpeg'
+    }
+  ]
+}
+
+export const TranscriptionSubjectMock = {
+  favorite: false,
+  id: SubjectPreviewState.subjectId,
+  locations: [
+    {
+      'image/jpeg':
+        'https://panoptes-uploads.zooniverse.org/production/subject_location/fb2c57e2-96f7-49b1-9bd4-9bdc73d847f8.jpeg'
+    }
+  ]
+}
+
+export const VideoSubjectMock = {
+  favorite: false,
+  id: SubjectPreviewState.subjectId,
+  locations: [
+    {
+      'video/mp4':
+        'https://panoptes-uploads.zooniverse.org/production/subject_location/279b23de-b1e8-4a1c-90c4-2d25bbee787d.mp4'
+    }
+  ]
+}
+
+const snapshot = {
+  project: {
+    strings: {
+      display_name: 'Snapshot Serengeti',
+    }
+  },
+  user: {
+    collections: {
+      collections: [],
+    }
+  }
+}
+
+export const StoreMock = Store.create(snapshot)
+
+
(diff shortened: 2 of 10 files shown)


packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.js
```diff
@@ -5,84 +5,87 @@ import { FavouritesButton, Media } from '@zooniverse/react-components'
...
-        />
-      </Box>
-    </Box>
-  )
+	return (
+		<Box
+			fill
+		>
+			<Anchor
+				role="subject-preview-link"
```
eatyourgreens, line 31: This isn't a valid ARIA role, which is why the builds are failing. 

I'd recommend not changing the announced role for links, as that will break them in the accessibility tree.

eatyourgreens (changes requested):

packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.js
```diff
@@ -5,84 +5,87 @@ import { FavouritesButton, Media } from '@zooniverse/react-components'
...
+				href={href}
+			>
+				<Box
+					align='center'
+					height={height}
+					overflow='hidden'
+					width={width}
+					role="shit"
```
eatyourgreens, line 39: Not sure what this is doing.

packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.js
```diff
@@ -5,84 +5,87 @@ import { FavouritesButton, Media } from '@zooniverse/react-components'
...
+				>
+					<Media
+						role="subject-preview-media"
```
eatyourgreens, line 42: Another invalid role.

eatyourgreens (changes requested):
These roles need to be removed too.

packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.js
kieftrav: How about using the Title attribute instead? Specifically, I'm wanting to make sure _this particular link_ has the href we're looking for from a testing-perspective.

packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.js
kieftrav: De-escalating frustration... lol. Removed!

packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.js
kieftrav: Same as above.. How about using the Title attribute to specify _this particular_ media element. There's also a quirk with the tests where it won't find a Media element by it's alt tag. Here's a snippet from `screen.debug()` in the `SubjectPreview.spec.js` file.

`
<a
          class="StyledAnchor-sc-1rp7lwl-0 epna-Df"
          href="/projects/zooniverse/snapshot-serengeti/talk/subjects/123"
          title="subject 123 preview link"
        >
          <div
            class="StyledBox-sc-13pk1d4-0 cnzmCK"
          >
            <div
              class="StyledBox-sc-13pk1d4-0 RwhTp"
              controls=""
              title="subject 123 image"
            />
            <noscript />
          </div>
        </a>
`

packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.js
eatyourgreens: Don't use `title`, particularly on links. It can be confusing in a screen reader and will appear as a tooltip for sighted users. 

https://www.tpgi.com/using-the-html-title-attribute-updated/

eatyourgreens: A CSS class can be used as a selector for a particular class of links.

kieftrav: kk - RTL doesn't like class detection. Another alternative is to use the `data-testid` attribute to specifically find an element. Would that me a reasonable compromise?

packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.js
kieftrav: As per above comment regarding not using the title... How about using `data-testid` instead for this?

packages/app-project/src/shared/components/SubjectPreview/SubjectPreview.js
eatyourgreens: Here's an example from the transcription task, which is tested with RTL, to find SVG groups that represent transcribed lines.

https://github.com/zooniverse/front-end-monorepo/blob/ee66328323a5fb37245f027fad8154bedbe77876/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/components/TranscribedLines/TranscribedLines.spec.js#L213-L216

eatyourgreens: The native DOM APIs will always be faster than RTL's queries, so I tend  to use the native APIs where they're suitable. Queries like `getByRole` are wrappers for `document.querySelectorAll`, if you delve into the RTL query code.

eatyourgreens: `data-test-id` works too. CSS classes are good when you want to apply a semantic description to a whole class of elements eg. a list of subject preview links on the home page.

eatyourgreens: You could also apply a class to the preview card itself, then query for `.subject-preview a` to find the link inside a card.

kieftrav: Cool - because this is used for testing purposes only I'll opt for the `data-testid` attribute. If this changes in the future it makes sense to turn it into a class / some other query selector.

eatyourgreens (approved):
Looks good. 

There are a lot of whitespace changes in these PRs. The monorepo uses 2 spaces for tabs but  I think your IDE is set up for 4.

PR #5231, opened 2023-08-24
SubjectSetPicker: Remove enzyme and convert tests to RTL

kieftrav:
## Package
app-project

## Describe your changes
Converted tests for the SubjectSetPicker from Enzyme to RTL. 

## How to Review
[SubjectSetPicker Story](http://localhost:9001/?path=/story/project-app-shared-subject-set-picker--default)

## General

## Refactoring


packages/app-project/src/shared/components/SubjectSetPicker/SubjectSetPicker.js
```diff
@@ -88,18 +88,20 @@ function SubjectSetPicker ({
...
-                passHref
-              >
-                <Anchor>
-                  <SubjectSetCard
-                    {...subjectSet}
-                    completeness={panoptesCompleteness}  /* This will override subjectSet.completeness */
-                  />
-                </Anchor>
-              </Link>
+              <Box data-testid={`test-subject-set-link-${subjectSet.id}`} key={subjectSet.id}>
```
eatyourgreens, line 91: Why are these wrapped in a flexbox now?

packages/app-project/src/shared/components/SubjectSetPicker/SubjectSetPicker.spec.js
```diff
@@ -1,101 +1,45 @@
...
-      // ]
-      expect(cards.at(0).props().display_name).to.equal('Transformers Series')
-      expect(cards.at(0).props().completeness).to.equal(0)
-
-      expect(cards.at(1).props().display_name).to.equal('Overwatch Series')
-      expect(cards.at(1).props().completeness).to.equal(0.67)
-
-      expect(cards.at(2).props().display_name).to.equal('Ultraman Series')
-      expect(cards.at(2).props().completeness).to.equal(1)
-    })
```
eatyourgreens, line 87: This test should be present in the RTL tests too. We want to confirm that completed sets are moved to the end of the list.

packages/app-project/src/shared/components/SubjectSetPicker/SubjectSetPicker.stories.js
```diff
@@ -1,42 +1,34 @@
...
   args: {
-    workflow: mockWorkflow
+    workflow: SubjectSetPickerMock
   }
 }
 
-export function Default({ workflow }) {
-  return (
+export const Default = ({ workflow }) => (
+  <Grommet theme={zooTheme}>
```
eatyourgreens, line 15: I think all our stories are wrapped in Grommet by default.

eatyourgreens (changes requested):
There's a test for completed subject sets that seems to be missing. Completed sets are moved to the end of the list and we should test that.

packages/app-project/src/shared/components/SubjectSetPicker/SubjectSetPicker.js
kieftrav: There's no way to add a `className` or a `data-testid` to the `Link` that wraps the SubjectSetCard component. Open to other suggestions so that I can grab that component without too much fuss / assumption about DOM-level implementation.

eatyourgreens: `Link` is an abstract component. The HTML link is rendered by the Grommet `Anchor`. Grommet components will pass DOM attributes through to the HTML element so you should be able to modify the link anchor.

kieftrav: Gotcha - moved to `<Anchor>` element

packages/app-project/src/shared/components/SubjectSetPicker/SubjectSetPicker.spec.js
kieftrav: Re-added

packages/app-project/src/shared/components/SubjectSetPicker/SubjectSetPicker.stories.js
kieftrav: Moved from the Story to the Spec

packages/app-project/src/shared/components/SubjectSetPicker/SubjectSetPicker.js
eatyourgreens: Here's the docs for `next/link` as a wrapper for links:
https://nextjs.org/docs/pages/api-reference/components/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag

eatyourgreens (approved):

PR #5241, opened 2023-08-28
MetaToolsButton: Remove enzyme and convert to RTL

kieftrav:
## Package
lib-react-components

## Describe your changes
Converted tests for the MetaToolsButton from Enzyme to RTL. 

## How to Review
[MetaToolsButton Story](http://localhost:9001/?path=/story/components-metatoolsbutton--button)

## General

## Refactoring

### DIFF
--- a/packages/lib-react-components/src/MetaToolsButton/MetaToolsButton.js
+++ b/packages/lib-react-components/src/MetaToolsButton/MetaToolsButton.js
@@ -17,16 +17,17 @@ export const StyledPlainButton = styled(PlainButton)`
 function MetaToolsButton (props) {
   const { disabled, icon, onClick, text } = props
 
-  return (
-    <StyledPlainButton
-      disabled={disabled}
-      icon={icon}
-      labelSize='small'
-      text={text}
-      onClick={onClick}
-      {...props}
-    />
-  )
+	return (
+		<StyledPlainButton
+			data-testid="test-meta-tools-button"
+			disabled={disabled}
+			icon={icon}
+			labelSize='small'
+			text={text}
+			onClick={onClick}
+			{...props}
+		/>
+	)
 }
 
 MetaToolsButton.propTypes = {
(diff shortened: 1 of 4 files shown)


packages/lib-react-components/src/MetaToolsButton/MetaToolsButton.js
```diff
@@ -17,16 +17,21 @@ export const StyledPlainButton = styled(PlainButton)`
...
+      <StyledPlainButton
+        data-testid="test-meta-tools-button"
+        disabled={disabled}
+        icon={icon}
+        labelSize='small'
+        text={text}
+        onClick={onClick}
+        {...props}
+      />
+    )
```
shaunanoordin, line 31: Question: why is this render wrapped in a `try`? I don't think our other components run try-catches in their render, so I'm assuming this is for a debug?

Also, if we ever have a try-catch in the main render method (and the _try_ returns elements to render), then I assume the _catch_ would also need to return _something,_ even if it's `return null`

packages/lib-react-components/src/MetaToolsButton/MetaToolsButton.js
```diff
@@ -17,16 +17,21 @@ export const StyledPlainButton = styled(PlainButton)`
...
+  } catch(err) {
+    console.log('goddamn error', err)
```
shaunanoordin, line 33: lol OK maybe this is a debug thing. Don't forget to remove this before this PR is ready to roll.

shaunanoordin (approved):
## PR Review

This PR updates the `MetaToolsButton` component's Story and unit tests, replacing the latter's Enzyme code with React Testing Library.

Changes look pretty straightforward, and the test & Story seem to work as expected. I'm giving this a 👍 ...

❗ **...however,** the try/catch in MetaToolsButton needs to be removed before merge, since I assume it's a "I was testing something on localhost" thing. If my assumption is wrong and it's _not_ a temporary test thing, please let me know and I'll re-review.

PR #5254, opened 2023-08-29
CloseButton: Remove enzyme and convert to RTL

kieftrav:
## Package
lib-react-components

## Describe your changes
Converted tests for the CloseButton from Enzyme to RTL. 

## How to Review
[CloseButton Story](http://localhost:9001/?path=/story/components-closebutton--default)

## General

## Refactoring

### DIFF
--- a/packages/lib-react-components/src/CloseButton/CloseButton.spec.js
+++ b/packages/lib-react-components/src/CloseButton/CloseButton.spec.js
@@ -1,33 +1,23 @@
-import { shallow } from 'enzyme'
-import sinon from 'sinon'
-import CloseButton from './CloseButton'
-import { expect } from 'chai'
+import { render, screen } from '@testing-library/react'
+import { composeStory } from '@storybook/react'
+import Meta, { Default, Light, Disabled } from './CloseButton.stories.js'
 
-describe('<CloseButton />', function () {
-  let wrapper
-  before(function () {
-    wrapper = shallow(<CloseButton closeFn={sinon.spy()} />)
+describe('Component > CloseButton', function () {
+  it('renders the default button', function () {
+    const DefaultStory = composeStory(Default, Meta)
+    render(<DefaultStory />)
+    expect(screen.getByRole('button').hasAttribute('disabled')).to.be.false()
   })
 
-  it('renders without crashing', function () {
-    expect(wrapper).to.be.ok()
+  it('renders the light button', function () {
+    const LightStory = composeStory(Light, Meta)
+    render(<LightStory />)
+    expect(screen.getByRole('button').hasAttribute('disabled')).to.be.false()
   })
 
-  it('calls on the closeFn prop on click', function () {
-    wrapper.simulate('click')
-    expect(wrapper.props().onClick).to.have.been.calledOnce()
-  })
-
-  it('should not pass along a href prop', function () {
-    wrapper.setProps({ href: 'www.google.com' })
-    expect(wrapper.props().href).to.be.undefined()
-  })
-
-  describe('with a color prop', function () {
-    it('should set the icon colour', function () {
-      const colouredButton = shallow(<CloseButton color='neutral-6' closeFn={sinon.spy()} />)
-      const icon = colouredButton.prop('icon')
-      expect(icon.props.color).to.equal('neutral-6')
-    })
+  it('renders the disabled button', function () {
+    const DisabledStory = composeStory(Disabled, Meta)
+    render(<DisabledStory />)
+    expect(screen.getByRole('button').hasAttribute('disabled')).to.be.true()
   })
 })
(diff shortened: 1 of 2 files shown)


packages/lib-react-components/src/CloseButton/CloseButton.spec.js
```diff
@@ -1,33 +1,23 @@
...
 
-describe('<CloseButton />', function () {
-  let wrapper
-  before(function () {
-    wrapper = shallow(<CloseButton closeFn={sinon.spy()} />)
+describe('Component > CloseButton', function () {
+  it('renders the default button', function () {
+    const DefaultStory = composeStory(Default, Meta)
+    render(<DefaultStory />)
+    expect(screen.getByRole('button', { disabled: false })).to.be.ok()
```
shaunanoordin, line 9: ⚠️ Ahhah, we've got a small problem here - `disabled: false/true` isn't actually a proper option for getRole()

Check out the `options?` listed in [RTL's ByRole documentation](https://testing-library.com/docs/queries/byrole) -
there's `hidden`, `name`, `description`, etc, but you can't select by disabled state.

You can test this yourself by changing `disabled: false` to `disabled: true`, and the `expect()` will _still_ say, yep, that's definitely `ok()`. This means that getByRole() is ignoring the invalid "disabled" option.

shaunanoordin: One option is to do check the `disabled` attribute separately.

For example, we can replace...
```
expect(screen.getByRole('button', { disabled: false })).to.be.ok()
```

With...
```
const button = screen.getByRole('button')
expect(button).to.be.ok()
expect(button.hasAttribute('disabled')).to.be.false()
```

packages/lib-react-components/src/CloseButton/CloseButton.spec.js
```diff
@@ -1,33 +1,23 @@
...
-  describe('with a color prop', function () {
-    it('should set the icon colour', function () {
-      const colouredButton = shallow(<CloseButton color='neutral-6' closeFn={sinon.spy()} />)
-      const icon = colouredButton.prop('icon')
-      expect(icon.props.color).to.equal('neutral-6')
-    })
+  it('renders the disabled button', function () {
+    const DisabledStory = composeStory(Disabled, Meta)
+    render(<DisabledStory />)
+    expect(screen.getByRole('button', { disabled: true })).to.be.ok()
```
shaunanoordin, line 21: Same comment as above: `disabled` is not a valid option for `getByRole()`

packages/lib-react-components/src/CloseButton/CloseButton.spec.js
```diff
@@ -1,33 +1,23 @@
...
+    render(<DefaultStory />)
+    expect(screen.getByRole('button', { disabled: false })).to.be.ok()
   })
 
-  it('renders without crashing', function () {
-    expect(wrapper).to.be.ok()
+  it('renders the light button', function () {
+    const LightStory = composeStory(Light, Meta)
+    render(<LightStory />)
+    expect(screen.getByRole('button', { disabled: false })).to.be.ok()
```
shaunanoordin, line 15: Same comment as above: `disabled` is not a valid option for `getByRole()`

shaunanoordin (changes requested):
## PR Review

This PR makes updates to the CloseButton component's tests and Storybook story.

Code changes _mostly_ look good - the tests are testing basic rendering, and the storybook shows relevant use cases - but there's a small problem with how getByRole() is used.

### Status

Changes requested:
- [ ] [] remove `{ disabled: ... }` argument in `screen.getByRole('button', { disabled: false })`
  - Reason: "disabled" [isn't a valid option for getByRole](https://testing-library.com/docs/queries/byrole), and gives people reading the test a false assurance.
- [ ] [optional]: instead of trying to select a disabled/not-disabled button, use hasAttribute, e.g. `expect(myExampleButton.hasAttribute('disabled')).to.be.false()`, to determine if a button is disabled/not-disabled.

Other than that small thing, everything else looks good. 👌

packages/lib-react-components/src/CloseButton/CloseButton.spec.js
```diff
@@ -1,33 +1,23 @@
...
-import { expect } from 'chai'
+import { render, screen } from '@testing-library/react'
+import { composeStory } from '@storybook/react'
+import Meta, { Default, Light, Disabled } from './CloseButton.stories.js'
 
-describe('<CloseButton />', function () {
-  let wrapper
-  before(function () {
-    wrapper = shallow(<CloseButton closeFn={sinon.spy()} />)
+describe('Component > CloseButton', function () {
```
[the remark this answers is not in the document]
kieftrav: I'm not sure if there's any actual standard naming here. Probably something to talk about at some point

shaunanoordin (approved):
## PR Review (Update)

I'm sorry I've been sitting on this one's re-review for so long! I wasn't paying attention to the updates, and didn't realise that the feedback about changing `getByRole()`'s use of `disabled` has already been addressed. 👍 

LGTM, let's get this PR merged!

PR #5327, opened 2023-09-13
Fix missing padding in the logged-out user menu

eatyourgreens:
PR #5278 accidentally removed the padding from the signed-out user navigation menu in `ZooHeader`. This restores it.

This also fixes an invalid button label on the user account dropdown menu.

## Package
- lib-react-components

## How to Review
Compare the signed-out user navigation menu with the live web site.

# Checklist

## Bug Fix
- [x] The bug is fixed

### DIFF
--- a/packages/lib-react-components/src/ZooHeader/ZooHeader.js
+++ b/packages/lib-react-components/src/ZooHeader/ZooHeader.js
@@ -32,6 +32,7 @@ export const StyledLogoAnchor = styled(Anchor)`
 `
 
 const defaultHandler = () => true
+const signedOutUserNavPadding = { horizontal: 'medium', vertical: 'small' }
 
 export default function ZooHeader({
   breakpoint = 960,
@@ -72,6 +73,7 @@ export default function ZooHeader({
     t('ZooHeader.mainHeaderNavListLabels.build')
   ]
 
+  const userNavigationPadding = Object.keys(user).length === 0 ? signedOutUserNavPadding : undefined
   return (
     <StyledHeader
       ref={ref}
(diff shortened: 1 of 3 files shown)


goplayoutside3 (changes requested):
Please run this branch locally and double check. On loading it in Chrome, the padding of the header is way too large: 
Screenshot 2023-09-13 at 2 57 22 PM

eatyourgreens:
Oh, that’s not good. I did check it locally before opening the PR, both logged in and logged out.

eatyourgreens:
Here's the top bar in Firefox:
Screenshot of the Zooniverse site navigation in Firefox, with normal padding.

And Chrome:
Screenshot of the Zooniverse site navigation in Firefox, with padding that's double what it should be.

I'll take a look to see why the padding is double what it should be in Chrome. I don't see anything obvious in the changes here that would do that. The only difference I see in the HTML is that the signed-out menu uses buttons, but the signed-in menu uses links.

eatyourgreens:
My fix is to apply padding only when you're logged out but it seems odd that the height of that box changes depending on whether it contains links or buttons.

eatyourgreens:
I figured it out: the user account dropdown menu button has internal padding which sets the height of the whole top bar. Note that `div` inside `button` here is invalid HTML, so that may cause problems too.

Zooniverse site navigation, with a signed-in user and the user account button highlighted in the DOM inspector, showing the padding around the button label.

eatyourgreens:
I've fixed the invalid HTML for the button label, replacing the `div` with `span`. I left the padding on the button itself. I wasn't sure if we want to have the padding inside or outside the clickable area. It's currently inside the clickable area, so I've left that alone.

mcbouslog (approved):
Confirmed styling looks good on Chrome, Safari, and Firefox 👍

Issue #3779, opened 2022-10-14
Pure ES Modules break the Mocha tests


eatyourgreens:
WHAT?? The tests for `app-project` and `lib-classifier` are failing on:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/shaun/projects/front-end-monorepo/node_modules/d3/src/index.js from /Users/shaun/projects/front-end-monorepo/packages/app-project/src/shared/components/ProjectStatistics/components/CompletionBar/CompletionBar.js not supported.
Instead change the require of index.js in /Users/shaun/projects/front-end-monorepo/packages/app-project/src/shared/components/ProjectStatistics/components/CompletionBar/CompletionBar.js to a dynamic import() which is available in all CommonJS modules.
```

and

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/shaun/projects/front-end-monorepo/node_modules/d3/src/index.js from /Users/shaun/projects/front-end-monorepo/packages/lib-classifier/src/components/Classifier/components/Feedback/components/Graph2dRangeFeedback.js not supported.
Instead change the require of index.js in /Users/shaun/projects/front-end-monorepo/packages/lib-classifier/src/components/Classifier/components/Feedback/components/Graph2dRangeFeedback.js to a dynamic import() which is available in all CommonJS modules.
```

and etc

OK, so upon review:
- _D3 now ships as pure ES modules and requires Node.js 12 or higher._ - this IS a problem.

Additional reading:
- [The awkward valley to ESM: Node.js, Victory, and D3](https://formidable.com/blog/2022/victory-esm/) - 15 Jun 2022, formidable.com - TL;DR: as more libraries (e.g. d3.js) drop CommonJS support and go 100% ESM (ECMAScript Modules), a lot of upstream/downstream dependencies that use CommonJS get borked. (ESM setups can read ESM + CommonJS modules, but CommonJS setups can't read ESModules.) Awkward transpiling workarounds are required.
- [Importing Issues - Error [ERR_REQUIRE_ESM]: Must use import to load ES Module](https://github.com/d3/d3-array/issues/218) - 29 Jun 2021, D3 repo - TL;DR: [lol good luck](https://github.com/d3/d3-array/issues/218#issuecomment-870600490)
[...]

PR #5352, opened 2023-09-17
Markdownz: replace remark with markdownz

### LINKED ISSUE #372: Migrate Markdownz's editor and help
**Package** lib-react-components

The rest of Markdownz should be migrated and updated in the process.

**To Dos**

- [ ] Example

### LINKED ISSUE #2607: Upgrade Remark
## Package
lib-react-components

`remark`, the markdown parser used by our `Markdownz` component, is on version `12.0.1`. The latest released version is `14.0.2`.

Upgrade to 14.

## Additional context
We get the occasional security warning about outdated dependencies of `remark` (eg. `trim`.) Upgrading might improve package security.

Some notes on version compatibility:
 - 11 is compatible with our `Markdownz` component. #2609 upgrades us to Remark 11 without issues.
 - 12 introduced breaking changes to the `remark` API. We'll need to update Markdownz, and add `remark-footnotes`, before we can upgrade.
 - 13 rewrote the markdown parser and requires plugins to be updated to the new parser API. Our fork of `remark-ping` is written for the old parser and will need to be rewritten.
 - 14 is a pure ESM module. Our test setup errors if you import a pure module.
 
`remark-react` is also not supported any more, and the recommendation is to migrate to `rehype-react`.

### LINKED ISSUE #5439: Markdownz: paragraph text rendered as subscript
There aren't any subscripts on the Galaxy Zoo Research page:

https://www.zooniverse.org/projects/zookeeper/galaxy-zoo/about/research

However, Remark renders the end of this paragraph as subscript, because of the `~` symbol in the redshift estimates. 

Paragraph text rendered as subscript text on the Galaxy Zoo Research page.

_Originally posted by @eatyourgreens in https://github.com/zooniverse/front-end-monorepo/pull/5352#issuecomment-1746146768_

### LINKED ISSUE #5440: Markdownz: Embedded video overflows its container and doesn't play
## Package
_Choose from the list:_
- [ ] app-content-pages
- [ ] app-project
- [ ] lib-classifier
- [ ] lib-panoptes-js
- [x] lib-react-components
- [ ] unknown

## Describe the bug
Embedded video does not play in a markdown page.

https://github.com/zooniverse/front-end-monorepo/assets/59547/9c96c973-f2c0-4530-a1ed-3c696d19bbb2

## To Reproduce
Play the embedded video on this page:
https://frontend.preview.zooniverse.org/projects/goplayoutside3/fem-test-project/about/research

## Expected behavior
Embedded video should play, like the same embedded video here on PFE:
https://www.zooniverse.org/projects/goplayoutside3/fem-test-project/about/research

Tested with Firefox 118 on MacOS 11.7.10

eatyourgreens:
Replace `remark` with [`markdownz`](https://www.npmjs.com/package/markdownz), to parse Zooniverse markdown to HTML, and [`rehype`](https://www.npmjs.com/package/rehype/v/11.0.0), to render the parsed HTML as a React component tree.

Bump `@zooniverse/react-components` to 1.8.0.

Our `markdown-it` plugins rely on Node modules. I've added fallbacks to the dev classifier and project Webpack builds, where they require them.

~~Rehype 12+ is only available as ESM, so I've installed Rehype v11.~~ Rehype is bundled with `markdownz`.

- ~~includes #5341.~~
- closes #372.
- closes #2607 but potentially opens another issue, in that Rehype 12+ is also only available as ESM (#3779.)
- closes https://github.com/zooniverse/markdownz/issues/102.
- fixes #5439.
- fixes #5440.

## Package
- app-content-pages
- app-project
- app-root
- lib-classifier
- lib-react-components

## How to Review
I've updated the Markdownz Grid Example story to add various examples of Zooniverse-flavoured markdown.
http://localhost:6007/?path=/story/components-markdownz--grid-example

I've also deployed this to kubernetes, so that we can check various production projects:
- https://fe-project-branch.preview.zooniverse.org/projects/brooke/i-fancy-cats?env=staging (subject metadata includes URLs.)
- https://fe-project-branch.preview.zooniverse.org/projects/zookeeper/galaxy-zoo
- https://fe-project-branch.preview.zooniverse.org/projects/zooniverse/gravity-spy
- https://fe-project-branch.preview.zooniverse.org/projects/zooniverse/snapshot-serengeti
- https://fe-project-branch.preview.zooniverse.org/projects/humphrydavy/davy-notebooks-project/classify/workflow/24833
- https://fe-project-branch.preview.zooniverse.org/projects/penguintom79/penguin-watch/about/faq (an example of table-of-contents links, using fragment identifiers.)
- https://fe-project-branch.preview.zooniverse.org/projects/sassydumbledore/chimp-and-see/classify/workflow/18435 (embedded YouTube videos in the field guide.)
- https://fe-project-branch.preview.zooniverse.org/projects/sassydumbledore/chimp-and-see/about/research (markdown tables used to layout images next to text.)
- https://fe-project-branch.preview.zooniverse.org/projects/rsengar/pulsar-seekers/about/research (embedded YouTube video about the project.)

# Checklist

## General

## General UX

## Refactoring


packages/lib-react-components/src/Markdownz/Markdownz.spec.js
```diff
@@ -16,16 +12,16 @@ import { markdown } from './helpers/testExamples'
 describe('<Markdownz />', function () {
   it('parses markdown to jsx', function () {
     const { container } = render(<Markdownz>{markdown}</Markdownz>)
-    expect(container.querySelectorAll('a[href]')).to.have.lengthOf(12)
-    expect(container.querySelectorAll('p')).to.have.lengthOf(17)
+    expect(container.querySelectorAll('a[href]')).to.have.lengthOf(10)
+    expect(container.querySelectorAll('p')).to.have.lengthOf(18)
```
eatyourgreens, line 16: These changes get the tests passing, but I can't figure out, from the example markdown, what the expected values really should be.

eatyourgreens:
Had to install Rehype 11, because Rehype 12+ is only available as ESM.

eatyourgreens:
Rehype parses a HTML fragment to a full page, wrapped in `<html>`, which results in an invalid DOM.

<img width="431" alt="<html> nested inside <div> on the Galaxy Zoo Research page." src="https://github.com/zooniverse/front-end-monorepo/assets/59547/16079f93-8042-4fc8-b52f-427e82cdd815">

eatyourgreens:
`markdown-it` was last updated a year ago. That doesn’t look good. 
https://github.com/markdown-it/markdown-it/issues/948

goplayoutside3:
Re: Rehype 12+ is only available as ES - 

I didn't realize that we'd run into the same headache of ESM-only package by switching to rehype so thank you for noting that right away! I think I'm hesitant about replacing remark with rehype because it seems like a temporary solution and also hesitant that markdown-it hasn't been updated recently (as you noted). What are your thoughts on FEM markdown strategy going forward? What's your opinion on switching to `markdownz` to address the Github Issues linked, versus sticking it out with Remark for now?

eatyourgreens:
>What's your opinion on switching to markdownz to address the Github Issues linked, versus sticking it out with Remark for now?

Depends on what stakeholders, which would be project teams here, expect from Zooniverse-flavoured markdown. If they expect their markdown to continue to work without changes, then `markdownz` will continue to parse it to the same HTML that they're already accustomed to.

With Remark, the input markdown changes and the output HTML is different, which should be summarised in [ADR 11](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-11.md). So project teams will have to edit their existing markdown for the new parser, and we will need to document the required changes for them.

My personal preference would be to leave Zooniverse-falvaoured markdown unchanged if possible, so use `markdownz` to convert markdown to HTML.

NB. Remark uses Rehype to convert its output HTML to a React component tree, and `remark-react` has been replaced by `rehype-react`. Whether we use Remark or `markdownz` to parse markdown to HTML, we'll still need Rehype to map the output to Grommet components.

Re. ESM: that's a wider issue that affects all our dependencies. This monorepo was built on CommonJS, at a time when CommonJS was being dropped in favour of ES modules. For example, several of the packages in the monorepo are written as CommonJS modules. These packages are all written with `require` and `module.exports`:
- lib-async-states
- lib-grommet-theme
- lib-panoptes-js

We need to be able to build and test in a world where not all of our dependencies are CommonJS any more. Whichever markdown parser we choose to use, it's not likely to be available as a CommonJS module forever.

Bun has a good summary of the current state of [Node module resolution](https://bun.sh/docs/runtime/modules).

eatyourgreens:
@shaunanoordin has also written some useful notes about ES modules on Community Catalogue: https://github.com/zooniverse/community-catalog/issues/165

eatyourgreens:
Also worth bearing in mind that the monorepo was originally written for Node 8, and packages haven’t been updated in line with changes in subsequent versions of Node, particularly support for `import/export`.

eatyourgreens:
I changed the name of this since it’s changing the markdown-to-HTML conversion, not the HTML-to-React conversion.

goplayoutside3:
Thanks! All of the info about ESM and markdownz is helpful. I think my preference would also be to make migration from PFE to FEM as easy as possible for project owners while also not needing to install or write Remark plugins to handle custom features of Zooniverse-flavoured markdown. 

I still want to do some manual comparison of live FEM projects to this branch because I also want to consider the project owners who already wrote all of their project content in compatibility with Remark. There shouldn't be major editing needed from project owners in the process of switching to `markdownz` in FEM. Will continue to review this week!

eatyourgreens:
>I still want to do some manual comparison of live FEM projects to this branch because I also want to consider the project owners who already wrote all of their project content in compatibility with Remark. There shouldn't be major editing needed from project owners in the process of switching to markdownz in FEM. Will continue to review this week!

I think we'll be OK here, because the editor in the project builder uses `markdownz`. Project teams would have had to manually edit their markdown to make it compatible with the FEM implementation of Remark, without any way to preview it.

eatyourgreens:
Rebasing this to keep it ahead of the main branch (and production release.)

eatyourgreens:
>also hesitant that markdown-it hasn't been updated recently (as you noted).

`markdown-it` just released 13.0.2. New version of `markdownz` to follow shortly.
https://github.com/zooniverse/markdownz/pull/199

eatyourgreens:
Useful conversation here about the maintenance status of `markdown-it`. It is still supported. Looks like we can expect 13.1, with ESM support, and 14 (a full ES module.)
https://github.com/markdown-it/markdown-it/issues/948#issuecomment-1740664308

goplayoutside3:
There are several incorrect displays of markdown on this branch compared to FEM staging.
1. Subscript citation displays as normal text in the [Galaxy Zoo Research page](https://fe-project-branch.preview.zooniverse.org/projects/zookeeper/galaxy-zoo/about/research) in the section: Where do the images comes from? There is supposed to be a subscript citation ([frontend.preview](https://frontend.preview.zooniverse.org/projects/zookeeper/galaxy-zoo/about/research)).
2. The links to Zooniverse usernames are broken. See the [Gravity Spy team page](https://fe-project-branch.preview.zooniverse.org/projects/zooniverse/gravity-spy/about/team) where each team member has a link to their Zooniverse username in the markdown. The link looks like normal text on this branch, but an actual link on [frontend.preview](https://frontend.preview.zooniverse.org/projects/zooniverse/gravity-spy/about/team).
3. Video files in markdown extend beyond their container bounds. Example: https://fe-project-branch.preview.zooniverse.org/projects/goplayoutside3/fem-test-project/about/research
4. MarkdownzTable component in the Markdownz storybook file does not look like a table. Please run lib-react-components storybook locally and take a look. The too-large video example is also evident in the Markdownz storybook.

Please do some manual comparison of live FEM projects to this branch because there shouldn't be major editing needed from project owners. Their already existing markdown should display without changes.

eatyourgreens:
There aren't any subscripts on the Galaxy Zoo Research page:

https://www.zooniverse.org/projects/zookeeper/galaxy-zoo/about/research

There is, however, a massive bug in the Remark version on staging. 😳

This needs fixing before we can move the project over. 

Paragraph text rendered as subscript text on the Galaxy Zoo Research page.

eatyourgreens:
Re. usernames, I think Markdownz expects a space before the @ character. Something like `(@cplb)` won't auto-link if you post it on Talk either, because of the brackets.

eatyourgreens:
You need to compare markdown rendering here with markdown rendering on the live site.

Gravity Spy is an example of a project where they've updated their markdown to work with Remark. That's where the friction will be: new-style syntax that doesn't work in the old library. 

Please do open an issue on https://github.com/zooniverse/markdownz about mentions requiring a space (or even better, a PR!)

eatyourgreens:
Video works in markdownz. 

https://www.zooniverse.org/projects/goplayoutside3/fem-test-project/about/research

I'm not sure what's going on on FEM staging. Is there missing CSS styling for video?

eatyourgreens:
Video doesn't play for me on FEM staging. I've opened https://github.com/zooniverse/front-end-monorepo/issues/5440.

eatyourgreens:
Overflowing video is fixed here:
https://fe-project-branch.preview.zooniverse.org/projects/goplayoutside3/fem-test-project/about/research


https://github.com/zooniverse/front-end-monorepo/assets/59547/6e3bdb02-04f3-441f-8ce0-04960d04f64c

eatyourgreens:
Rebased and updated to include `app-root`.

eatyourgreens:
I've released [markdownz 8.3.0](https://github.com/zooniverse/markdownz/releases/tag/v8.3.0) with some security fixes, and updated this branch.

eatyourgreens:
Are there any project pages with examples of links that use `+tab+` to open in a new tab?

eatyourgreens:
Chimp & See use embedded YouTube videos in their field guide.
https://www.zooniverse.org/projects/sassydumbledore/chimp-and-see/classify

I'll publish 8.3.2 with support for embedded iframes.

goplayoutside3:
Tags look better! A few things to follow-up on:

1. Re: MarkdownzTable - I'm looking at the overall format of the Default Markdownz story, not how a table element is rendered. The [production Markdownz storybook](https://zooniverse.github.io/front-end-monorepo/@zooniverse/react-components/index.html?path=/story/components-markdownz--default) displays three column headers (content, markdown, and rendered). However, on this branch, the three-column formatting is lost. Project managers like Sam use the Markdownz story to easily reference how elements display in FEM, so please restore the column formatting for their benefit.

2. This change from Remark to `markdownz` needs an ADR at some point - either included within this PR or written directly afterward.

3. The Chimp&See field guide videos do display now. However, the youtube videos are displayed wider than the field guide modal's min-width of 550px, and therefore push the "close modal" button off screen because the modal is forced to be wider than 550px. Why do images conform to the modal’s dimensions, but videos do not?

4. External links:
    - Links that previously opened in a new tab (example [here](https://fe-project-branch.preview.zooniverse.org/projects/goplayoutside3/fem-test-project/about/research)), no longer do so on this branch because `remark-external-links` plugin adds `target=_blank` and `nofollow, noopener, noreferrer` to all absolute urls, whereas I think all project teams that used PFE relied on adding `+tab+` to links. Is that why you asked for examples of `+tab+` in a project?
    - More relevant documentation:
        - [ADR 11](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-11.md) mentions how “+tab+ in front of the url will no longer work because of the sanitization that remark-react does”.
        - This PR [4019](https://github.com/zooniverse/front-end-monorepo/pull/4019) removed +tab+ from the metadata modal in the classifier because that modal uses markdown.
        - Linking the relevant Issue [102](https://github.com/zooniverse/markdownz/issues/102), which I think you’re already aware of.
    - Final thoughts: Whether we decide that FEM's handling of markdown force-opens external websites in a new tab or not, any new markdown behavior introduced by this PR needs to be communicated immediately to live FEM project teams.

packages/lib-react-components/package.json
```diff
@@ -4,6 +4,7 @@
   "license": "Apache-2.0",
   "author": "Zooniverse <contact@zooniverse.org> (https://www.zooniverse.org/)",
   "version": "1.6.4",
+  "version": "1.7.0",
```
goplayoutside3, line 7: This package.json incorrectly has two `version` properties.

eatyourgreens:
>The Chimp&See field guide videos do display now. However, the youtube videos are displayed wider than the field guide modal's min-width of 550px, and therefore push the "close modal" button off screen because the modal is forced to be wider than 550px. Why do images conform to the modal’s dimensions, but videos do not?

YouTube videos are fixed width, I think. They use a custom syntax, which doesn't specify dimensions eg. `@[youtube](dQw4w9WgXcQ)`. If you inspect the DOM, you'll see they're embedded using an iframe.

eatyourgreens:
Your [example project page on staging](https://frontend.preview.zooniverse.org/projects/goplayoutside3/fem-test-project/about/research) has an accessibility failure. It opens a link in a new tab without warning, in advance, that clicking the link will open a new tab. See ['Opening new windows and tabs from a link only when necessary'](https://www.w3.org/WAI/WCAG21/Techniques/general/G200).

If all external links open in new tabs, we're going to need some kind of warning to avoid a level A failure for [change of context on focus](https://www.w3.org/WAI/WCAG21/Understanding/on-focus).

eatyourgreens:
>Re: MarkdownzTable - I'm looking at the overall format of the Default Markdownz story, not how a table element is rendered. The [production Markdownz storybook](https://zooniverse.github.io/front-end-monorepo/@zooniverse/react-components/index.html?path=/story/components-markdownz--default) displays three column headers (content, markdown, and rendered). However, on this branch, the three-column formatting is lost. Project managers like Sam use the Markdownz story to easily reference how elements display in FEM, so please restore the column formatting for their benefit.

No problem. I found that the three column table doesn't really work on smaller screens, but I have no strong opinion. This is what I've got on this branch, with a wide screen (1878px.)

The Markdownz story in the storybook, laid out as a three-column table on a wide screen.

eatyourgreens:
> ADR 11 mentions how “+tab+ in front of the url will no longer work because of the sanitization that remark-react does”.

This puzzles me for a couple of reasons:
- `remark-react` is a Remark wrapper for `rehype-react`. We're still using the same Rehype HTML-to-component mappings in this PR. I haven't changed the mapping for links. 
- I don't think `remark-react` sanitises URLs, so I'm not sure what that means.

packages/lib-react-components/CHANGELOG.md
```diff
@@ -13,6 +14,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 - restore module field in `package.json`.
+=======
+## [1.7.0] 2023-10-04
+
+### Changed
+- `Markdownz`: replace `remark` with `markdownz` and `rehype`.
+>>>>>>> 3901ee6f (Markdownz: replace remark with rehype)
```
eatyourgreens, line 22: Accidentally left in from a merge conflict.

eatyourgreens:
I've updated subject metadata so that it opens links in a new tab without using markdown. I'm not sure why it wasn't doing that already, to be honest.

eatyourgreens:
Also worth noting that since [markdownz 8.3.0](https://github.com/zooniverse/markdownz/releases/tag/v8.3.0), all output HTML is now sanitised against XSS. Remark itself doesn't sanitise by default, and is using outdated packages, so we should be mindful of security here. See [Remark's security notes](https://github.com/remarkjs/remark#security).

eatyourgreens:
No problems here with the Chimp & See videos. The field guide wraps to the size of the content, without overflow. I think I might have posted a screenshot of this before.

Screenshot of the field guide for Chimp & See, with an embedded YouTube video. The field guide modal wraps the video content without overflowing horizontal scroll bars.

eatyourgreens:
I'm seeing a flash of raw markdown when the page first loads, so I think there's a bug in `markdownz` when it's run in Node.


https://github.com/zooniverse/front-end-monorepo/assets/59547/76ccc9cf-d84d-401b-9d92-39128bd65a15

eatyourgreens:
Updated the examples to include a Chimp & See page that's using markdown tables.

packages/lib-react-components/src/Markdownz/Markdownz.stories.js
```diff
@@ -23,25 +75,38 @@ export default {
...
+    </Markdownz>
+  </>
 )
 
 export const WithCustomComponents = () => (
-  <Markdownz components={{ tr: TableRowWithBorder }}>
-    {markdownExample}
-  </Markdownz>
+  <>
+    <MarkdownExamplesTable components={{ tr: TableRowWithBorder }} />
```
goplayoutside3, line 98: Please run lib-react-components storybook locally and double check. I do not see TableRowWithBorder` applied as a custom component.

packages/lib-react-components/src/Markdownz/helpers/storybookExamples.js
```diff
@@ -0,0 +1,152 @@
...
+    content: '#tiger'
+  },
+  {
+    label: 'Zooniverse Subject mention (only works in project context)',
+    content: '^S1234'
+  },
+  {
+    label: 'Link inside a link',
+    content: '[foo @username bar](http://example.com)'
+  },
```
goplayoutside3, line 83: This rendered example does not match the deployed version. The markdown string is the same though, so please run the storybook locally and take a look at what might be happening:

This branch:
Screenshot 2023-10-24 at 12 34 19 PM


Deployed [version](https://zooniverse.github.io/front-end-monorepo/@zooniverse/react-components/index.html?path=/story/components-markdownz--default):
Screenshot 2023-10-24 at 12 34 26 PM

packages/lib-react-components/src/Markdownz/Markdownz.stories.js
```diff
@@ -23,25 +75,38 @@ export default {
...
+    <Markdownz>
+      {markdownExample}
+    </Markdownz>
+  </>
 )
 
 export const GridExample = () => (
```
goplayoutside3, line 105: In this story, there's a video string: `alt`, however the width constraint no longer works on this branch. Adding a width constraint to a markdown video was a bonus feature added to FEM, and is not supported in PFE. 

Examples:
- FEM: https://frontend.preview.zooniverse.org/projects/goplayoutside3/fem-test-project/about/research
- PFE: https://www.zooniverse.org/projects/goplayoutside3/fem-test-project/about/research

Thoughts on if we should keep the width constraint for videos in FEM? I don't know of any projects currently using it.

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/Metadata/components/MetadataModal/MetadataModal.js
```diff
@@ -27,10 +27,13 @@ const DatumWrapper = styled(Text)`
 `
 
 export function formatValue (value) {
+  if (value?.startsWith('http')) {
```
goplayoutside3, line 30: 🚨 This does not work. Please run [Mark’s test project](https://local.zooniverse.org:3000/projects/markbouslog/test-project-mb/classify/workflow/22300?env=production) locally and double check. I get an error thrown by the MetadataModal component. Suggestion to write a unit test for this change too. 

Here's a link to Mark's project [workflow on staging](https://frontend.preview.zooniverse.org/projects/markbouslog/test-project-mb/classify/workflow/22300) where metadata links open in a new tab.

Screenshot 2023-10-24 at 1 11 54 PM

goplayoutside3 (commented):
The storybook columns look much better - thanks! I left comments on specific code files, and here are the big pictures responses:
1. Re: 
> Your [example project page on staging](https://frontend.preview.zooniverse.org/projects/goplayoutside3/fem-test-project/about/research) has an accessibility failure. It opens a link in a new tab without warning...
- Noting that this accessibility failure is the current FEM markdown behavior, and it is not specific to my example project. TESS example: the link in the first paragraph of their [Research](https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/about/research) page.
- I most definitely agree we should give some kind of warning when a link is going to open a new tab. However, this PR is a change to FEM so FEM project owners are going to notice the differences. How do we mitigate changing FEM’s markdown behavior for external links? I think we can just communicate to all FEM project owners they need to use `+tab+`. Thoughts?
- Should we consider a design change where all FEM markdown links that contain `+tab+` should also get an "external link icon"? (discussion could be moved to Slack #design-thoughts)
___
2. Re:
> “No problems here with the Chimp & See videos”

Is this screenshot of the tutorial modal positioning when you first opened it? Or did you drag the modal to this position in order to see the full video element?
___
3. Re:
> Security and "Also worth noting that since markdownz 8.3.0, all output HTML is now sanitised against XSS"

- https://github.com/remarkjs/remark#security recommends use of `rehype-sanitize`. Should we be doing so?
- Related to security, [ADR 11](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-11.md) mentions that `markdownz` uses `dangerouslySetInnerHTML`, and indeed it does in [markdown.jsx](https://github.com/zooniverse/markdownz/blob/22dc7ecfbb8424967e8f94ef0b0d6ee9e712d6f0/src/components/markdown.jsx#L33). Is inserting `markdownz` into FEM a regression in security?

eatyourgreens:
This PR doesn't import the `Markdownz` component or use it anywhere.

packages/lib-react-components/src/Markdownz/Markdownz.spec.js
```diff
@@ -3,29 +3,25 @@ import { render, screen } from '@testing-library/react'
...
-  RESTRICTED_USERNAMES,
-  buildResourceURL,
-  renderMedia,
-  replaceImageString,
-  shouldResourceBeLinkable
+  renderMedia
 } from './Markdownz'
 import { markdown } from './helpers/testExamples'
 
 // TO DO: Add back working snapshots to test the overall HTML output
```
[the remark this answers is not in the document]
eatyourgreens: What does 'snapshots' mean here? HTML output from each of the examples? I think the storybook does a pretty good job of demonstrating our custom markdown but I'm not familiar enough with Zooniverse projects to say how project teams or volunteers are using it. So it's hard for me to comment on how it should work, which means that it's then difficult for me to define testable behaviour. 

I think this requires us having some example markdown snippets coupled with the exact HTML that we expect from those snippets.

packages/lib-react-components/src/Markdownz/Markdownz.stories.js
eatyourgreens: I see borders on each row in the markdown table, as expected.

eatyourgreens: That's the table in the last example, by the way. The table that's used to lay out the page is written in JSX, not markdown, so won't change.

packages/lib-react-components/src/Markdownz/helpers/storybookExamples.js
eatyourgreens: This is expected for `markdownz`. If you try it in Talk, or in a PFE project page, it won't work either. 

I guess the remark version fixed this.

eatyourgreens: The `markdownz` code for mentions is a regex that's looking for strings preceded by @. 
https://github.com/zooniverse/markdownz/blob/22dc7ecfbb8424967e8f94ef0b0d6ee9e712d6f0/src/lib/default-transformer.js#L4

packages/lib-react-components/src/Markdownz/Markdownz.stories.js
eatyourgreens: 🤷🏻‍♂️. It's not supported by the [html5-embed plugin ](https://www.npmjs.com/package/markdown-it-html5-embed)but we can write our own plugins if project teams want this. Should `<video>` use height and width attributes, or should the height and width be styled with CSS?

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/Metadata/components/MetadataModal/MetadataModal.js
eatyourgreens: That's odd. `value` must be a string for the key value pairs in subject metadata, which is serialised as a string by the API, or possibly null if the value is empty. 

This works on every subject I've tested it with. It should work in the storybook too. I thought it already had tests. I'll take a look. 

Either way, metadata values shouldn't be parsed as markdown here. URLs should be rendered as links in JSX.

eatyourgreens: Do you happen to have the subject metadata that's breaking the link here?

packages/lib-react-components/src/Markdownz/Markdownz.spec.js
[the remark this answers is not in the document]
eatyourgreens: At the moment, I'm figuring out how our markdown is expected to work by both reading the markdown help in the project builder, and reading through the list of installed `markdown-it` plugins. 

https://github.com/zooniverse/markdownz/blob/22dc7ecfbb8424967e8f94ef0b0d6ee9e712d6f0/src/lib/utils.js#L27-L38

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/Metadata/components/MetadataModal/MetadataModal.js
eatyourgreens: Did you take a look at the tests? They render the mock subject from the storybook. The metadata includes a URL, and it's rendering without errors. 

I'm really curious to know how a subject has broken this, because we should add the failing case to the storybook. 

https://github.com/zooniverse/front-end-monorepo/blob/dbc9aca77ea8f103d041cff6c3018ab57c2e78be/packages/lib-classifier/src/components/Classifier/components/MetaTools/components/Metadata/components/MetadataModal/MetadataModal.stories.js#L8-L13

eatyourgreens:
> I think we can just communicate to all FEM project owners they need to use +tab+. Thoughts?

Wasn't one of the reasons for this PR that we tell people to use `+tab+` for external links, but that isn't working in the monorepo. Hence https://github.com/zooniverse/markdownz/issues/102. I think I'm misunderstanding the problem here.

Whatever we've been telling project owners, we need to be clear about how Zooniverse flavoured markdown works. 

This is going to be really confusing if different teams have received different advice about how to write markdown in the project builder.

eatyourgreens:
> Is inserting markdownz into FEM a regression in security?

No, it's an improvement in security because `utils.getHTML(markdown)` now sanitises the output HTML string with DOMPurify, before it's passed to Rehype to be converted to a React component tree and inserted into the page. 

I think `remark-react` had HTML sanitisation too, but that's deprecated now, hence this PR. 

Would `rehype-sanitise` strip out anything that we use eg. embedded iframes for the Chimp & See YouTube videos?

We can always see if it's possible for C&S to upload their videos to Panoptes, rather than hosting them on YouTube.

eatyourgreens:
> Should we consider a design change where all FEM markdown links that contain +tab+ should also get an "external link icon"? (discussion could be moved to Slack #design-thoughts)

Have a read of this excellent piece by Michelle Barker. My only question is how would a blind volunteer be made aware that the browsing context is about to move to a new window or tab. That said, this seems better than nothing. 

https://css-irl.info/styling-external-links-with-attribute-selectors/

eatyourgreens:
> Is this screenshot of the tutorial modal positioning when you first opened it? Or did you drag the modal to this position in order to see the full video element?

I'm not sure. I've been testing on an extra wide screen too, so probably not representative of the average volunteer. 

Those YouTube iframes are fixed width, though, which is a vote against using them. Unless we override their default styling. 

Is there a big demand for YouTube videos in projects? C&S was the only project I found that used them.

Manatee Chat have video examples, but I think they upload them to `panoptes-uploads`.

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/Metadata/components/MetadataModal/MetadataModal.js
eatyourgreens: Bug confirmed on this branch here. 
https://fe-project-branch.preview.zooniverse.org/projects/markbouslog/test-project-mb/classify/workflow/22300

Nothing looks unusual about the subject metadata, though.

eatyourgreens: Could be that the tests here are passing when they should be failing. However, if that were the case I’d expect the storybook to crash too.

eatyourgreens:
>Related to security, [ADR 11](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-11.md) mentions that markdownz uses dangerouslySetInnerHTML, and indeed it does in [markdown.jsx](https://github.com/zooniverse/markdownz/blob/22dc7ecfbb8424967e8f94ef0b0d6ee9e712d6f0/src/components/markdown.jsx#L33).

That’s why we’ve built a new `Markdownz` component in the monorepo, which doesn’t use `dangerouslySetInnerHTML`. You won’t see that used anywhere in this PR. 

`markdownz` now sanitises the string that’s passed to `dangerouslySetInnerHTML`, which protects its own `Markdownz` component against XSS injection. 
- https://github.com/zooniverse/markdownz/pull/201

packages/lib-classifier/src/components/Classifier/components/MetaTools/components/Metadata/components/MetadataModal/MetadataModal.js
eatyourgreens: I've found the bug. I Fancy Cats subjects have numbers in their metadata. The mock subject in the tests only has string metadata.

The metadata changes are big enough that I'm going to move them out into their own PR.
Metadata for an I Fancy Cats subject, which is a mixture of strings and numbers.

eatyourgreens: I've cherry-picked these changes into https://github.com/zooniverse/front-end-monorepo/pull/5525, so that they can be merged while this is still pending review.

packages/lib-react-components/src/Markdownz/Markdownz.stories.js
eatyourgreens: The markdown table example from the storybook, but each row has a custom bottom border.

This is in Firefox, by the way, but it shouldn't vary at all between browsers.

goplayoutside3: Is it not the borders on the entire Markdownz layout table?

Screenshot 2023-10-25 at 6 56 52 PM

goplayoutside3: ^ my screenshot is from the [deployed](https://zooniverse.github.io/front-end-monorepo/@zooniverse/react-components/index.html?path=/story/components-markdownz--with-custom-components) storybook

goplayoutside3: So to be clear, the difference between the "old" Markdownz storybook examples and the "new" examples is that the new examples are isolated? For example, the footnotes do not get bumped all the way down to the bottom of the story anymore. The footnotes just exist is their own square. 

I think this is fine! I was confused on how to compare changes to the story examples

packages/lib-react-components/src/Markdownz/helpers/storybookExamples.js
goplayoutside3: Sounds good thanks! This isn’t in the PFE markdown guide, so I’m fine with just using `markdownz`'s handling. I honestly can't remember why the link inside a link example was added to FEM's storybook.

packages/lib-react-components/src/Markdownz/Markdownz.stories.js
goplayoutside3: I’m going to say it’s fine to **not** support video “resize in markdown” in FEM, and Zooniverse can recommend to teams that they upload example videos via panoptes rather than a service like youtube (see note below about communication to Zooniverse PMs and FEM project teams).

For example, I looked at the Manatee Chat video examples in their tutorial and field guide in [FEM](https://local.zooniverse.org:3000/projects/cetalingua/manatee-chat/classify/workflow/24994?env=production), and sizing looks great 👍

goplayoutside3:
> markdownz now sanitises the string that’s passed to dangerouslySetInnerHTML.

Awesome! Thank you for the clarification

> how would a blind volunteer be made aware that the browsing context is about to move to a new window or tab.

I'm going to raise this discussion with us plus Sean in this `markdownz` [Issue](https://github.com/zooniverse/markdownz/issues/102). I'll document some of what we've discussed over there!

> Re: youtube videos. Is there a big demand for YouTube videos in projects? C&S was the only project I found that used them.

I think recommending video upload via panoptes rather than a service like youtube would benefit project owners and volunteers (see note below about communication to Zooniverse PMs and FEM project teams).

> Wasn't one of the reasons for this PR that we already tell people to use `+tab+` for external links, but that isn't working in the monorepo?

Yes, but unfortunately Zooniverse PMs have already communicated to at least some FEM project teams that `+tab+` does not currently work in FEM (Slack thread [here](https://zooniverse.slack.com/archives/C01TYAR0V43/p1661374884281589?thread_ts=1661374475.093859&cid=C01TYAR0V43)). I definitely agree it’s confusing if different teams have received different advice about how to write markdown in the project builder. This goes for all markdown elements such as recommended video sources, not just external links.

When this PR is merged, happy to help document the exact resulting changes in markdown behavior for FEM projects (even though those “changes” are just re-aligning FEM markdown processing with PFE’s lab). The documentation process will also need an ADR and explanation to Zooniverse PMs probably via Slack.

packages/lib-react-components/src/Markdownz/Markdownz.stories.js
eatyourgreens: You’re right that this is a confusing story. I’ll move the table to the top of the examples, so that it’s  more obvious what’s changed. 

I forgot to note that I replaced the original story with individual markdown examples. I found that the current story is hard to debug. If there’s a problem anywhere in the markdown, the whole story crashes. Isolated examples made debugging easier.

packages/lib-react-components/src/Markdownz/helpers/storybookExamples.js
eatyourgreens: It is an odd example. I don’t think I’ve seen anyone put an @-mention inside a link either. 

I would guess that hashtags and subject mentions can’t be placed inside links either.

packages/lib-react-components/src/Markdownz/Markdownz.stories.js
eatyourgreens: Sounds good. The Remark implementation writes video and audio as HTML `` tags, with height and width, then maps images to a React video component if `src` is a video URL.

eatyourgreens:
> Yes, but unfortunately Zooniverse PMs have already communicated to at least some FEM project teams that +tab+ does not currently work in FEM.

Ah, right, that makes sense now!

eatyourgreens:
I found a couple of other bugs with links. My test markdown is `[Team page]('./team?env=staging')`.
1. Remark strips the query string from URLs, so links (internal and external) can't use query params? That needs checking with fully-qualified URLs too.
2. Remark and Rehype both render links as Grommet `Anchor` components, but they should use `next/link` in Next.js apps. That should be easily fixed by passing in a custom `components` object from the Next.js apps.

eatyourgreens:
About pages are also generating 500 errors instead of showing the 404 page.

goplayoutside3:
Checking in here before I do one final review - are you planning to add an ADR as mentioned in FE standup?

I imagine this could be merged early next week, and deployed during normal FEM deploy with any changes to markdown behavior communicated to Zooniverse PMs.

eatyourgreens:
Yep, I plan to write an ADR. It can go in a separate PR if I don't get to it on Monday.

eatyourgreens:
I've released `markdownz@9.0.0`, which includes Rehype, so that Rehype doesn't have to be installed separately here.
https://github.com/zooniverse/markdownz/releases/tag/v9.0.0

eatyourgreens:
PFE staging (https://master.pfe-preview.zooniverse.org) and this branch are now both on markdownz 9.1.0.

PFE uses the component version:

```jsx
import { Markdown } from 'markdownz';

return (
  <Markdownz>
    Some example markdown, with [a link](https://www.zooniverse.org).
  </Markdownz>
);
```

This branch uses the hooks version, with custom component mappings. Other than that, they should both be the same.

```jsx
import { useMarkdownz } from 'markdownz';

const html = useMarkdownz({
  components: customComponentsObject,
  content: 'Some example markdown, with [a link](https://www.zooniverse.org).'
});

return <>{html}</>;
```

eatyourgreens:
The new pulsars project uses a YouTube video:
https://master.pfe-preview.zooniverse.org/projects/rsengar/pulsar-seekers/about/research

goplayoutside3:
The links to example projects are helpful - thank you! `markdownz` updates are looking good, and I think this PR is good to go 👍 👍 👍 

Because of the change to external links (requiring `+tab+` to open new browser tab), I propose this gets merged directly AFTER this week's FEM deploy. There are definitely changes to markdown behavior here that live FEM project teams should be aware of. For example, there are links in the Davy Notebooks banner ([this branch](https://fe-project-branch.preview.zooniverse.org/projects/humphrydavy/davy-notebooks-project/classify), [production](https://www.zooniverse.org/projects/humphrydavy/davy-notebooks-project/classify)) and the Davy Notebooks team + volunteers are accustomed to those opening in a new tab, especially from the classify page.

Let's notify the #project-building channel after this gets merged tomorrow about changes to markdown behavior going live on next FEM deploy (November 7). That'll give PMs a week or so to communicate to project teams.

goplayoutside3 (approved):

PR #5386, opened 2023-09-22
Fix SVG subject scaling

### LINKED ISSUE #5291: Transcription Task: gray lines blending into subject background
## FEM 
Transcription Task

On the Davy Notebooks Project, certain subjects have the same background color as the greyed-out lines, creating problems where people don't realize a page is complete and start re-transcribing things. The current workaround is to use the color inversion tool, which clearly shows the grey lines.

I think we need to consider a new color for completed lines that won't blend into certain paper colors (brighter?).

## Additional context
Grey lines blending in:
Screenshot 2023-09-05 at 9 43 59 AM

Same page, shown in inverted color:
Screenshot 2023-09-05 at 9 44 05 AM

### LINKED ISSUE #5384: Image scale is initially set to 1 for SVG subject images
https://github.com/zooniverse/front-end-monorepo/blob/691020eac40a2b48b0701666235c043c705f2d1b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js#L57-L63

`clientWidth` here is being returned as the full width of the subject image in pixels, not the width of its container in the browser. This results in SVG lines and shapes being rendered much smaller than they should be (see #5291.)

eatyourgreens:
- [Reset the `SVGPanZoom` component state](https://react.dev/reference/react/useState#resetting-state-with-a-key) when its image size changes, so that the default viewbox always matches the current image height and width.
- allow `SVGImage` to be passed a ref. Use that ref to report subject image dimensions in classifications. 
- fix `TranscriptionLine` scaling.

## Package
- lib-classifier

- fixes #5384.
- closes #5291.

## How to Review
This bug was first reported on Davy Notebooks and Beyond Borders, because it's most obvious when previous transcription lines are shown on top of an image. However, it affects all the image subject viewers.

When a subject loads:
- the correct client dimensions should be recorded in the classification.
- previous transcription lines should be scaled correctly relative to the viewport, not shown too small (ie. not drawn at the same pixel scale as the full-size subject image.)

Screenshot of a page from Davy Notebooks, in a reasonably wide browser window, showing the previously transcribed lines scaled to the correct size relative to the browser viewport.

With the line-width scaled to the subject size:
Screenshot of a page from Davy Notebooks, in a reasonably wide browser window, showing the previously transcribed lines scaled to the correct size relative to the browser viewport and the subject image size.

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
@@ -86,11 +86,9 @@ function SVGPanZoom({
 
   function onDrag(event, difference) {
     setViewBox((prevViewBox) => {
-      const newViewBox = {
-        ...prevViewBox,
-        x: prevViewBox.x - difference.x / 1.5,
-        y: prevViewBox.y - difference.y / 1.5
-      }
+      const newViewBox = { ...prevViewBox }
+      newViewBox.x -= difference.x / 1.5
+      newViewBox.y -= difference.y / 1.5
       return newViewBox
     })
   }
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
@@ -1,10 +1,23 @@
 import PropTypes from 'prop-types'
-import { cloneElement, useEffect, useState } from 'react'
+import { cloneElement, useEffect, useRef, useState } from 'react'
+import styled from 'styled-components'
+
+const FullWidthDiv = styled.div`
+  width: 100%;
+`
+
+function imageScale(imgRef, naturalWidth) {
+  const { width: clientWidth, height: clientHeight } = imgRef?.current
+    ? imgRef.current.getBoundingClientRect()
+    : {}
+  const scale = clientWidth / naturalWidth
+  return !Number.isNaN(scale) ? scale : 1
+}
 
 const DEFAULT_HANDLER = () => true
 function SVGPanZoom({
+  imgRef = null,
   children,
-  img,
   limitSubjectHeight = false,
   maxZoom = 2,
   minZoom = 1,
@@ -38,8 +51,7 @@ function SVGPanZoom({
   }
 
   useEffect(function onZoomChange() {
-    const newViewBox = scaledViewBox(zoom)
-    setViewBox(newViewBox)
+    scaleViewBox(zoom)
   }, [zoom])
 
   useEffect(() => {
@@ -50,27 +62,26 @@ function SVGPanZoom({
   }, [zooming, src])
 
   useEffect(() => {
+    setViewBox({
+      x: 0,
+      y: 0,
+      height: naturalHeight,
+      width: naturalWidth
+    })
     setZoom(1)
-    setViewBox(defaultViewBox)
-  }, [src])
+  }, [naturalWidth, naturalHeight])
 
-  function imageScale(img) {
-    const { width: clientWidth, height: clientHeight } = img
-      ? img.getBoundingClientRect()
-      : {}
-    const scale = clientWidth / naturalWidth
-    return !Number.isNaN(scale) ? scale : 1
-  }
-
-  function scaledViewBox(scale) {
+  function scaleViewBox(scale) {
     const viewBoxScale = 1 / scale
-    const xCentre = viewBox.x + viewBox.width / 2
-    const yCentre = viewBox.y + viewBox.height / 2
     const width = parseInt(naturalWidth * viewBoxScale, 10)
     const height = parseInt(naturalHeight * viewBoxScale, 10)
-    const x = xCentre - width / 2
-    const y = yCentre - height / 2
-    return { x, y, width, height }
+    setViewBox((prevViewBox) => {
+      const xCentre = prevViewBox.x + prevViewBox.width / 2
+      const yCentre = prevViewBox.y + prevViewBox.height / 2
+      const x = xCentre - width / 2
+      const y = yCentre - height / 2
+      return { x, y, width, height }
+    })
   }
 
   function onDrag(event, difference) {
@@ -104,28 +115,29 @@ function SVGPanZoom({
         return
       }
       case 'zoomto': {
-        setZoom(1)
         setViewBox({
           x: 0,
           y: 0,
-          width: naturalWidth,
-          height: naturalHeight
+          height: naturalHeight,
+          width: naturalWidth
         })
+        setZoom(1)
+        return
       }
     }
   }
 
   const { x, y, width, height } = viewBox
-  const scale = imageScale(img)
+  const scale = imageScale(imgRef, naturalWidth)
 
   return (
-    <div style={{ width: '100%' }}>
+    <FullWidthDiv>
       {cloneElement(children, {
         scale,
         viewBox: `${x} ${y} ${width} ${height}`,
         svgMaxHeight: limitSubjectHeight ? `min(${naturalHeight}px, 90vh)` : null
       })}
-    </div>
+    </FullWidthDiv>
   )
 }
 
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.js
@@ -78,7 +78,8 @@ const FlipbookViewer = ({
   return (
     <Box>
       <SVGPanZoom
-        img={subjectImage.current}
+        key={viewerLocation.url}
+        imgRef={subjectImage}
         limitSubjectHeight={limitSubjectHeight}
         maxZoom={5}
         minZoom={0.1}
@@ -97,8 +98,9 @@ const FlipbookViewer = ({
           rotate={rotation}
           width={img.naturalWidth}
         >
-          <g ref={subjectImage} role='tabpanel' id='flipbook-tab-panel'>
+          <g role='tabpanel' id='flipbook-tab-panel'>
             <SVGImage
+              ref={subjectImage}
               invert={invert}
               move={move}
               naturalHeight={img.naturalHeight}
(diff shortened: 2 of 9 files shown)


packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
```diff
@@ -104,6 +104,7 @@ const SubjectViewer = types
 
       onSubjectReady (event) {
         const { target = {} } = event || {}
+        console.log(event)
```
eatyourgreens, line 107: Stray console log

eatyourgreens:
On this branch, previously transcribed lines are scaled correctly but I'm still seeing the wrong client dimensions in a demo classification for a Davy Notebooks page.

Client dimensions of 3135px x 3560px in a demo classification for Davy Notebooks.

eatyourgreens:
>On this branch, previously transcribed lines are scaled correctly but I'm still seeing the wrong client dimensions in a demo classification for a Davy Notebooks page.

I think his might be fixed by [resetting `SVGPanZoom` component state](https://react.dev/reference/react/useState#resetting-state-with-a-key) when the image URL changes.

eatyourgreens:
The bug is fixed by resetting the [SVG viewBox](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox) when the image URL changes, so I think it is caused by the placeholder image, which sets the initial image size to 800px by 600px. The subject loads into a viewBox of `0 0 800 600` rather than the dimensions of the subject image.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
```diff
@@ -104,28 +115,29 @@ function SVGPanZoom({
...
     }
   }
 
   const { x, y, width, height } = viewBox
-  const scale = imageScale(img)
+  const scale = imageScale(imgRef, naturalWidth)
 
   return (
-    <div style={{ width: '100%' }}>
+    <FullWidthDiv>
```
goplayoutside3, line 134: Out of curiosity, does moving the `width` property to styled-components from the style object have an impact on the UI or is it a dx preference? I ask because styled-components has been brought up in FE discussion as a tech debt, and want to make sure I understand whether it's preferable to keep using styled-components' `styled()` function versus defining css properties directly in the style object?

goplayoutside3 (changes requested):
I think this PR unintentionally breaks the keyboard pan/zoom interaction for most subject viewers. For example, when I load the classifier's storybook on this branch, I can't pan or zoom via the arrow keys for the SingleImageViewer, FlipbookViewer, MultiFrameViewer, ImageAndTextViewer, SeparateFramesViewer, etc. 

I also ran i-fancy-cats locally and keyboard pan/zoom is broken for its subject viewer.

eatyourgreens:
That's weird, because I specifically tested that and it's working fine for me. What should I do to reproduce the problem?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
eatyourgreens: The advantage being that the arguments to `styled` are a CSS string, so easily moved to a stylesheet if necessary.

eatyourgreens:
D'oh! I was testing with the pan/zoom option selected in the image toolbar, which does work.

eatyourgreens:
Here's a quick screen recording of pan/zoom from the keyboard on my laptop. It works when the image toolbar has focus, but I'm not sure about the subject image.


https://github.com/zooniverse/front-end-monorepo/assets/59547/2c21d412-2365-46ec-9032-3b4e6a41238f

eatyourgreens:
Elements decorated with `draggable()` can't be focused when you click on them, so `DraggableImage` can't be selected by clicking on it. That's an accessibility failure so I'll fix that here too.

eatyourgreens:
Keyboard focus for draggable components  doesn't have a quick fix, so I've reverted the changes to `SVGImage` instead. The regular SVG `<image>`, that it uses by default, can be selected by clicking on it.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
goplayoutside3: Thanks!

goplayoutside3:
Pan/zoom with the keyboard is looking good now!

I'm confused about the transcription lines though. Is there supposed to be a noticeable difference in the display of transcription lines on a subject image, such as in Davy Notebooks? I do see the new drop shadow when invert is false, however I don't see any difference in the size of the transcription lines. For example, here is subject 91027036 of workflow 24837 for comparison. My laptop display is 1512 x 982 with browser full screen, and I can't see any difference in size of transcription lines.

1. Production: https://www.zooniverse.org/projects/humphrydavy/davy-notebooks-project/classify/workflow/24837/subject/91027036?env=production&demo=true
Screenshot 2023-09-26 at 4 19 07 PM

2. This branch: https://local.zooniverse.org:3000/projects/humphrydavy/davy-notebooks-project/classify/workflow/24837/subject/91027036?env=production&demo=true
Screenshot 2023-09-26 at 4 19 29 PM

Is there a specific case where I can replicate previous transcription lines should not scaling correctly relative to the viewport?

goplayoutside3:
Oh wait! I see the smaller tooltip icon at the end of each transcription line when a subject first loads on production. Toggling the invert setting is what corrects it to a large, more easily clickable size.

I think a follow-up question then is, do we want to increase the thickness of the lines too? On this branch even if the point on the end of each line is larger, the thickness of the line is still the same size as production.

eatyourgreens:
It looks like the width of transcription lines is set here. `GRAB_STROKE_WIDTH` is a constant defined for all the drawing tools. 

Should be easy enough to open a PR to change those values, then see how that affects the drawing tools in general (if we change the grab width, which is the clickable size.)

My advice, based on experience, would be to keep this PR focused on the subject dimensions bug, rather than letting the scope creep.

eatyourgreens:
My mistake. The grab width is defined just for transcription lines, so that makes it easier to change. 
https://github.com/zooniverse/front-end-monorepo/blob/54fe0cc3b757958c1d6c098a10b8af98f5fd7643/packages/lib-classifier/src/plugins/drawingTools/experimental/components/TranscriptionLine/helpers/constants.js#L2

eatyourgreens:
OK, #5414 will update those values. I’m editing it on my phone but the battery is dying.

eatyourgreens:
Weird that transcription lines aren't using the default drawing styles defined here:

https://github.com/zooniverse/front-end-monorepo/blob/e6cdcf2532d1b70b8306de914fa3a182bb416ea4/packages/lib-classifier/src/plugins/drawingTools/components/Mark/Mark.js#L7-L8

goplayoutside3:
A separate PR for transcription line width sounds good 👍 

However, I tested more FEM projects with various subject viewers, and this PR introduces a bug for the flipbook viewer. Each subject image flashes (maybe it's an animation?) when it becomes the active frame rather than the previously preloaded/smooth transition. The change in UX is most noticeable with `autoplay` enabled in the flipbook:
1. [Daily Minor Planet ](https://local.zooniverse.org:3000/projects/fulsdavid/the-daily-minor-planet/classify?env=production&demo=true)
2. Flipbook autoplay [storybook](http://localhost:6006/?path=/story/subject-viewers-flipbookviewer--autoplay)

https://github.com/zooniverse/front-end-monorepo/assets/23665803/ccb1ede0-0fe3-45af-b0cd-0dd656a33547

eatyourgreens:
I think that’s the image unmounting when the URL changes, now that images are keyed by URL. 🤔

eatyourgreens:
Flipbook images are expected to be the same size, so maybe the key should be the dimensions, not the URL.

eatyourgreens:
@goplayoutside3 keying the zoom component by image height and width seems to fix that problem.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SVGComponents/SVGPanZoom/SVGPanZoom.js
eatyourgreens: In terms of performance, definitely prefer CSS classes over inline styles on individual elements.

Good:
```jsx
const StyledDiv = styled.div`
  .styledThing {
    color: red;
  }
`

<StyledDiv>
  <p className='styledThing'>Hello</p>
  <p className='styledThing'>World</p>
</StyledDiv>
```

Bad:
```jsx
<StyledDiv>
  <p style={{ color: 'red' }}>Hello</p>
  <p style={{ color: 'red' }}>World</p>
</StyledDiv>
```

eatyourgreens:
I've added the missing scale factor to transcription lines too, so `strokewidth = { 2 / scale}` rather than `strokewidth={2}`.

snblickhan:
> Oh wait! I see the smaller tooltip icon at the end of each transcription line when a subject first loads on production. Toggling the invert setting is what corrects it to a large, more easily clickable size.
> 
> I think a follow-up question then is, do we want to increase the thickness of the lines too? On this branch even if the point on the end of each line is larger, the thickness of the line is still the same size as production.

I would prefer to divorce the convo around line thickness from the drop shadow, since it isn't clear if we actually want (or need) to increase visible line thickness. My preference would be to increase click area only, and then increase visible thickness only if we get feedback that it's necessary.

eatyourgreens:
No problem. I've dropped the last commit from this branch.

goplayoutside3 (approved):
The looping in the flipbook viewer looks much better thanks!

I still see a flash of animation when the frame changes in MultiFrameViewer. Not blocking because that viewer doesn't have a looping feature, but I'm curious why changing the key to dimensions didn't remove the flash of animation for MultiFrameViewer too. You can see the difference in behavior by navigating between frames of the Transcription test project:

Local: https://local.zooniverse.org:3000/projects/blicksam/transcription-task-testing/classify/workflow/13898?env=production&demo=true

Staging: https://frontend.preview.zooniverse.org/projects/blicksam/transcription-task-testing/classify/workflow/13898?demo=true

eatyourgreens:
There's also a missing key in the `FrameCarousel` component, according to error messages in the dev console, but that's an easy fix.

eatyourgreens:
>I still see a flash of animation when the frame changes in MultiFrameViewer. Not blocking because that viewer doesn't have a looping feature, but I'm curious why changing the key to dimensions didn't remove the flash of animation for MultiFrameViewer too. You can see the difference in behavior by navigating between frames of the Transcription test project:

Each of those images is a different size. Here are the keys, logged to the console in strict mode.

Image keys for a multi-frame transcription subject, logged to the console. Each frame has a different key.

PR #5410, opened 2023-09-26
update window location when subject changes

eatyourgreens:
Run `router.replace` inside `onSubjectChange`, so that it's only called once when the classifier subject changes. Update `onSubjectChange`'s dependencies, so that it uses the correct subject ID from the window location.


packages/app-project/src/screens/ClassifyPage/components/ClassifierWrapper/ClassifierWrapper.js
```diff
@@ -81,22 +77,21 @@ export default function ClassifierWrapper({
...
   */
   const onSubjectChange = useCallback((subject) => {
-    setClassifierSubjectID(subject?.id)
-  }, [])
+    if (subjectFromURL) {
+      const subjectPageURL = `${baseURL}/subject/${subject.id}`
+      /* True when the classifier subject changed, but new id is not yet reflected in the url */
+      if (subjectFromURL !== subject.id) {
+        const href = addQueryParams(subjectPageURL)
+        router.replace(href, href, { shallow: true })
```
eatyourgreens, line 91: This could also be `router.push`. Depends how we want the browser Back button to behave.

goplayoutside3 (approved):
Looks good! router.replace instead of push sounds good to me 👍

PR #5430, opened 2023-10-02
ZRC: restore the ES module entry point

eatyourgreens:
- Restore the `module` entry point to `@zooniverse/react-components`. It was removed in #5325.
- Bump the version to 1.6.3.

## Package
- lib-react-components


eatyourgreens:
Content pages build and run without errors.

Projects build and run. I can browse and submit classifications.

PR #5500, opened 2023-10-19
[Security] Ignore post-install scripts

eatyourgreens:
Don't run post-install scripts during Docker builds. This is a vector for supply chain attacks, and there have been a few cases where they have been exploited to distribute malware via NPM.

## How to Review
A Docker build should still run without errors, even though post-install scripts don't run during the build stage. 

Unfortunately, we don't run a Docker build during CI, so that won't be tested by our test builds. 

CI tests and CI builds in Node should still run successfully without post-install scripts. 

# Checklist

## General

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -19,7 +19,7 @@ jobs:
          node-version: 'lts/hydrogen'
          cache: 'yarn'
 
-      - run: yarn install --production=false --frozen-lockfile
+      - run: yarn install --production=false --frozen-lockfile --ignore-scripts
       - run: yarn workspace @zooniverse/react-components build:es6
       - run: yarn workspace @zooniverse/classifier build:es6
       - run: yarn workspace @zooniverse/fe-project build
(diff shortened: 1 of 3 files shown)


mcbouslog (approved):
LGTM, nice catch and refactor! Since app-content-pages is loading as expected I'm likely doing something wrong on my end with app-project (see notes below).

When I run `docker-compose up --build` for app-content-pages `http://local.zooniverse.org:3000/about/team` loads as expected 👍 .

When I run `docker-compose up --build` for app-project
- the storybook fails with Exit code 127
- when I attempt to load a project home page I get the Zooniverse 404 page. is this expected?

eatyourgreens:
That's odd. I've only ever run docker builds from the root directory, which builds and runs both apps together (one on port 3000, the other on port 3001.)

mcbouslog:
> That's odd. I've only ever run docker builds from the root directory, which builds and runs both apps together (one on port 3000, the other on port 3001.)

Yes! Thank you. Running docker from the root directory app-project also works as expected 👍 .

eatyourgreens:
No errors building an image and deploying it to staging, so that sounds like a one-off thing. NB. `docker compose up` in the `app-project` directory runs `yarn dev`, not `yarn start`, so the image would use Panoptes staging. Production projects would 404.

PR #5577, opened 2023-11-01
Fix #5575, allow sticky sidebar in About page to overflow/scroll

### LINKED ISSUE #5575: About Page: sticky sidebar doesn't show full nav list when viewport's too short
## Minor UI Bug

Affects: `app-content-pages` (About -> Team and About -> Publications page)
Follows #5552 

The _newly-sticky_ sidebar on the About -> Team and About -> Publications page has an issue when the height of the browser viewport is shorter than the nav list.

- Expected: as a user viewing the page, I should be able to view all the items in the nav list, scrolling if necessary.
- Actual: the sticky nav list is "stuck" showing only the top of the list.

Screenshot: example of Publications page's nav list being shorter than the viewport. It only shows 8.5 items instead of 11, and attempting to scroll the nav list will instead scroll the content on the right.

Screenshot 2023-11-01 at 17 28 15

### Suggested Solution

For the StickySidebar component, adding `max-height: 100vh; overflow: auto` should allow users to scroll the nav list.

Screenshot 2023-11-01 at 17 27 43

### Status

Minor UI issue. I don't think anyone will realistically have a window shorter than the nav list, but fix should also be pretty simple, I think.

shaunanoordin:
## PR Overview

package: `app-content-pages`
Affects: About -> Team, About -> Publications
Fixes #5575
Follows #5552

This PR fixes a small UI issue where, on extremely vertically short viewports _(OR when the browser is zoomed in, OR the browser has increased font sizes),_ the nav list in the sticky sidebar isn't fully displayed and can't be scrolled

Solution is a very straightforward `max-height: 100vh` and `overflow: auto`

Design flaw/tradeoff:
- Curiously, with the existing sticky behaviour, if you scroll aaallll the way down to the end of the page, the sidebar will eventually show the final items in the nav list. (screenshot on left) With this PR, the sidebar needs to be manually scrolled. (screenshot on right) 

image  image 

### Testing

Tested with macOS + Chrome 119

- Open the About -> Publications page (since it has the longest nav list): https://local.zooniverse.org:3000/about/publications
- then choose one of the following scenarios to ensure the height of the sidebar exceeds the height of the browser viewport:
  - Narrow viewport: resize browser window to something like 740x360 (Samsung Galaxy S8+ sideways)
  - Page Zoom: Cmd + Plus(+) sign to zoom in
  - Font Size: modify your browser settings to display larger fonts (for Chrome, increase _minimum_ font size in advanced settings.) 
- scroll down the page (see screenshot below)
- observe that the sidebar exhibits "sticky" behaviour (i.e as you scroll down, the sidebar doesn't "scroll off beyond the top"
- 🆕 observe that you can scroll the sidebar (independently of the page content) to view the whole nav list. (Hint: first item in the 11-item list is "All", the final item is "Data")

image

### Status

Ready for review. Low priority, but may have some impact on users who rely on zoomed-in views.

Discard if flaw/tradeoff isn't worth it.

### DIFF
--- a/packages/app-content-pages/src/screens/Publications/Publications.js
+++ b/packages/app-content-pages/src/screens/Publications/Publications.js
@@ -19,7 +19,9 @@ const Relative = styled.aside`
 `
 
 const StickySidebar = styled(Sidebar)`
-  position: sticky;
+  max-height: 100vh;
+  overflow: auto;
+  position: sticky;  
   top: 0;
 `
 
--- a/packages/app-content-pages/src/screens/Teams/Teams.js
+++ b/packages/app-content-pages/src/screens/Teams/Teams.js
@@ -16,7 +16,9 @@ const Relative = styled.aside`
 `
 
 const StickySidebar = styled(Sidebar)`
-  position: sticky;
+  max-height: 100vh;
+  overflow: auto;
+  position: sticky;  
   top: 0;
 `
 


goplayoutside3 (approved):
Looks good!

PR #5666, opened 2023-11-20
lib-user: add ProfileHeader tests with mocha, with type commonjs

mcbouslog:
## Package
- lib-user

## Describe your changes
- refactor lib-user package from type module to type commonjs for consistent mocha setup with rest of monorepo
- add ProfileHeader tests

## How to Review

# Checklist

## General

## General UX

## New Feature

### DIFF
--- a/packages/lib-user/package.json
+++ b/packages/lib-user/package.json
@@ -5,6 +5,7 @@
   "author": "Zooniverse <contact@zooniverse.org> (https://www.zooniverse.org/)",
   "version": "0.0.0",
   "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
   "exports": {
     ".": {
       "import": "./dist/esm/index.js",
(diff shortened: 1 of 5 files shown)


packages/lib-user/package.json
```diff
@@ -16,7 +17,6 @@
     }
   },
   "sideEffects": false,
-  "type": "module",
```
eatyourgreens, line 19: The Node docs recommend that you explicitly set `"type": "commonjs"` here, rather than relying on the default. I'm also wary of building a new library with an outdated module system.

eatyourgreens: >This flag currently defaults to "commonjs", but it may change in the future to default to "module". For this reason it is best to be explicit wherever possible; in particular, package authors should always include the ["type"](https://nodejs.org/docs/latest-v20.x/api/packages.html#type) field in their package.json files, even in packages where all sources are CommonJS. Being explicit about the type of the package will future-proof the package in case the default type of Node.js ever changes, and it will also make things easier for build tools and loaders to determine how the files in the package should be interpreted.

https://nodejs.org/docs/latest-v20.x/api/packages.html#determining-module-system

PR #5843, opened 2024-01-12
lib-user: Add Layout

mcbouslog:
## Package
- lib-user
- app-root

## Describe your changes
- adds Layout component per [related Figma designs](https://www.figma.com/file/qbqbmR3t5XV6eKcpuRj7mG/Group-Stats?type=design&node-id=457-5858&mode=dev) to lib-user
- adds @zooniverse/user dependency to app-root
- adds UserStats component to app-root/users/[login]/stats page.js

## How to Review
  - run `yarn dev` in app-root
  - go to https://local.zooniverse.org:3000/users/[login]/stats
  - adjust browser window width, note how different widths compare to linked Figma designs

# Checklist

## General

## New Feature


goplayoutside3 (commented):
Visually this is looking good! A couple of questions:
1. When I run app-root in dev mode, or `yarn build` app-root locally, I get an error that `styled-components` is not a function. The error is generated by calling `styled()` in UserLayout > PageContainer. Do you see the same error when run locally? The error can be closed to see the rendered page, but looks like this:
     Screenshot 2024-01-16 at 11 57 43 AM
     I tried the nextConfig strategy outlined in the Next.js [docs](https://nextjs.org/docs/app/building-your-application/styling/css-in-js#styled-components), but no luck on removing the error. I think it's a result of app-root being setup as `type:module` with ESM-only, and we're on `styled-components` v5. I didn't get as far as converting all of app-root to CJS (server.js is using ESM import/export), but wondering if you investigated this error at all.

2. What are the advantages or disadvantages of building a styled layout for user stats and group stats pages in app-root, versus including the layout within the UserStats component to be imported from lib-user into app-root? I ask because currently there's a "loading" moment before the stats page layout is rendered in between ZooHeader and ZooFooter. I think that loading UX could occur regardless of if the layout styling is in app-root or lib-user, but wondering if you've thought about that strategy yet (and might be a solution to app-root being ESM).

mcbouslog:
I did not realize styled-components was causing the noted error! That's frustrating, but sounds like it's not unexpected and can be revisited (if necessary) with related upgrades.

I tried refactoring the layout to lib-user as suggested and I think it works well and makes sense 👍 ! It adds lib-user as a dependency to app-root, and imports UserStats as a component for app-root/users/stats. I don't love the users/stats page.js with `use client`, but that page.js be substantially refactored to pass the authClient, Next link/routing, etc. so maybe ok for now.

packages/app-root/src/app/users/[login]/page.js
```diff
@@ -0,0 +1,7 @@
+export default function UserPage({ params }) {
+  return (
+    <div>
+      <p>This is {params.login}'s profile page.</p>
```
Comment (mcbouslog) [inline packages/app-root/src/app/users/[login]/page.js:4]: Open to suggestion about the dynamic segment name, related options:
1. currently it's `login` per the panoptes user resource property `login` (maybe should be `user_login`)
2. in [PFE the similar route](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/router.jsx#L151) uses `:profile_name`
3. another option is `username` or other similar variations

packages/app-root/src/app/users/[login]/stats/page.js
```diff
@@ -0,0 +1,9 @@
+'use client'
```
Comment (goplayoutside3) [inline packages/app-root/src/app/users/[login]/stats/page.js:1]: You mentioned not loving this implementation of `'use client'`, but I agree it's definitely fine to do for now! Our frontend stack is dependent on internal and external libraries with client-side React features, so marking an entire page as a client component isn't the end of the world at all. Also, a user's stats page is very dynamic data on top of the fact we don't have server-side authentication to handle display of stats on the server. Just wanted to encourage you not to dwell too much on this `use client`. It's difficult to utilize server component without rebuilding Zooniverse from the ground up 💁‍♀️

packages/app-root/src/app/users/[login]/page.js
Comment (goplayoutside3) [inline packages/app-root/src/app/users/[login]/page.js:4]: `eslint` doesn't like the apostrophe in this paragraph, and throws an error during `yarn build` of app-root locally. I suggest using `&apos;` instead.

packages/lib-user/src/components/shared/Layout/Layout.js
```diff
@@ -0,0 +1,151 @@
...
+          `
+    }
+  }
+`
+
+const PageBody = styled(Box)`
+  grid-column: 3 / 4;
+  grid-row: 2 / 4;
+  border-radius: 8px 8px 0 0;
+  padding-bottom: 2rem;
```
goplayoutside3, line 67: I suggest using `px` for CSS like padding. You don't really want padding to increase in size if someone's browser font size is set to `1rem = 40px`. See https://ashleemboyer.com/blog/why-you-should-use-px-units-for-margin-padding-and-other-spacing-techniques for more details.

packages/lib-user/src/components/shared/Layout/Layout.js
```diff
@@ -0,0 +1,151 @@
...
+          dark: 'dark-3',
+          light: 'neutral-6'
+        }}
+        border={{
+          color: {
+            dark: 'dark-2',
+            light: 'light-5'
+          },
+          side: 'left'
+        }}
```
goplayoutside3, line 120: Why does PageLeftColumn have a border-left? This renders a grey border on the very edge of a user's browser.

Same thing for PageRightColumn. It's rendering a border on the right edge of a user's browser. I think the border props can be removed altogether.

packages/lib-user/src/components/shared/Layout/Layout.js
```diff
@@ -0,0 +1,151 @@
...
+const PageLeftColumn = styled(Box)`
+  position: relative;
+  grid-column: 2 / 3;
+  grid-row: 3 / 4;
+
+  &::after {
+    content: '';
+    position: absolute;
+    top: 0;
+    right: -1px;
```
goplayoutside3, line 38:
```suggestion
    right: 0;
```

This `1px` difference is noticeable when zooming in:
Screenshot 2024-01-16 at 11 31 12 AM

packages/lib-user/src/components/shared/Layout/Layout.js
```diff
@@ -0,0 +1,151 @@
...
+
+const PageRightColumn = styled(Box)`
+  position: relative;
+  grid-column: 4 / 5;
+  grid-row: 3 / 4;
+
+  &::before {
+    content: '';
+    position: absolute;
+    top: 0;
```
goplayoutside3, line 78:
```suggestion
    top: 0;
    left: 0;
```

Just to be certain.

goplayoutside3 (commented):
Looking good! I left some minor comments for styling strategy, and I have one question for @seanmiller26:

In the About Zooniverse Pages and the Project About Pages, the content has a white background (`#ffffff`), and the grey page container is `light-1` (`#eff2f5`). In this User Stats Pages PR, the page container AND the page content are both `neutral-6` (`#ffffff`).

In FEM there's a precedent of defining an app's grey container with the Grommet styles below ([app-project](https://github.com/zooniverse/front-end-monorepo/blob/ecd4c20bb0d7fd2f61d5d7aa4cb744a41dc16abb/packages/app-project/pages/_app.js#L121-L125), [app-content-pages](https://github.com/zooniverse/front-end-monorepo/blob/ecd4c20bb0d7fd2f61d5d7aa4cb744a41dc16abb/packages/app-content-pages/src/shared/components/PageLayout/layout.js#L75-L79)). 
```
   background={{
      dark: 'dark-1',
      light: 'light-1'
   }}
```
Our Storybook is configured the same way. Do you want User Stats and Group Stats pages to follow the same pattern?

seanmiller26:
> In the About Zooniverse Pages and the Project About Pages, the content has a white background (#ffffff), and the grey page container is light-1 (#eff2f5). In this User Stats Pages PR, the page container AND the page content are both neutral-6 (#ffffff). Do you want User Stats and Group Stats pages to follow the same pattern?

Yes, please follow the precedent for this styling. #eff2f5 background (areas wider than 1440) #ffffff page container, and for the User Stats #ffffff for the actual component cards. @mcbouslog we can discuss dark mode options when you get to that point.

packages/app-root/src/app/users/[login]/stats/page.js
Comment (mcbouslog) [reply packages/app-root/src/app/users/[login]/stats/page.js:1]: Thanks that's helpful to hear!

packages/lib-user/src/components/shared/Layout/Layout.js
mcbouslog: The borders are related to the border and box-shadow noted in the design for the large view (related [Figma link](https://www.figma.com/file/qbqbmR3t5XV6eKcpuRj7mG/Group-Stats?node-id=603%3A7093&mode=dev)) that separates the white from grey sections. I couldn't implement the box-shadow for a few reasons, but did add the border. I've refactored the border to only show with a larger screen (when there are grey sections to delineate between).

seanmiller26: @goplayoutside3 Do you have any tips related to this shadow from your work on the Content pages styling?

mcbouslog: This layout is a little different than the Content pages styling in how the grid is setup, so the body appears to overlap the heading. In this layout the body is a child of the entire page container that defines the grid, compared to the Content pages where I think the body is a child of the white container.

Anyway, I think I have the box-shadow working, will push up changes in a minute...

goplayoutside3 (approved):
> I think the areas wider than 1440px are the desired grey (#eff2f5 / light-1)?

Yes you're right! That's my bad. I rarely look at pages on an ultra-wide screen so I forgot that these pages have somewhat of a double-elevation layout. 

👍 Looks great!

PR #5894, opened 2024-02-01
lib-user: Add Select component

mcbouslog:
## Package
- lib-user

## Describe your changes
- adds select input using Grommet Select per [related design](https://www.figma.com/file/qbqbmR3t5XV6eKcpuRj7mG/Group-Stats?type=design&node-id=24-547&mode=design&t=fayBN7HeeWQ3EL57-4)
- add related story and tests

## How to Review
  - review related story
  - http://localhost:6008/?path=/story/components-shared-select--date-ranges
- Are there plans for follow up PR’s to further fix this bug or develop this feature? yes, there might be refactors when implemented

# Checklist

## General

## New Feature


packages/lib-user/src/components/shared/Select/Select.js
```diff
@@ -0,0 +1,65 @@
...
+      >
+        <GrommetSelect
+          a11yTitle={name}
+          id={id}
+          name={name}
+          labelKey='label'
+          onChange={({ option }) => handleSelect(option)}
+          options={options}
+          size='medium'
+          style={{ textAlign: 'center' }}
```
mcbouslog, line 41: Not sure how I feel about this, but creating a styled-component just to add this one line of css didn't seem ideal either. I tried to add to the custom theme without luck.

kieftrav: I think it's ok. We've been making a lot of spacing css inline and when I did the SimpleDropdown project which also uses the Grommet Select I noticed a lot of limitations on customization / stylability.

packages/lib-user/src/components/shared/Select/Select.spec.js
```diff
@@ -0,0 +1,34 @@
...
+  it('should show the options', async function() {
+    render(<DefaultStory />)
+
+    const inputEl = screen.getByRole('textbox', { value: 'LAST 7 DAYS' })
+    await user.click(inputEl)
+
+    const options = screen.getAllByRole('option')
+
+    expect(options.length).to.equal(7)
+    expect(screen.getByText('LAST 7 DAYS')).to.be.ok()
```
kieftrav, line 26: I'd throw the options into a shared file for the stories and specs to share for confirmation instead of having it hardcoded twice with an unknown dependency between them.

kieftrav (approved):
Code looks good yet the styling isn't very close to the Figma file. Having spent a ton of work with Grommet Select I understand quite a bit the limitations there. Some things to explore (if you haven't already)

- What happens to the dropdown list when the option is really long? Does the dropdown width stay the same?
- Can you customize the scrollbar to match Figma better?
- Can you move the dropdown arrow a couple pixels to the left to better align with the center of the border-radius?
- Is there supposed to be a drop-shadow? Unclear from the Figma.
- Dark mode should probably invert the select box to white

kieftrav:
Here's some screenshots I took as well:

Screenshot 2024-02-06 at 8 37 52 AM
Screenshot 2024-02-06 at 8 50 40 AM
Screenshot 2024-02-06 at 9 05 37 AM

mcbouslog:
Thank you @kieftrav , I made a few a notes as follows, but will have to revisit some as well: 

> 
> * What happens to the dropdown list when the option is really long? Does the dropdown width stay the same? - great question, I'll follow-up with Sean, but I think it'll be ok if the button stays within a max width so as not to throw the larger section layout off, but the options expand the drop area? I'll revisit this shortly when I implement this select with projects that can have very long project names.
> * Can you customize the scrollbar to match Figma better? - this was tough, will come back to. I noticed the new /about Disciplines dropdown nav uses the standard scroll as well, so I think this will be ok for now
> * Can you move the dropdown arrow a couple pixels to the left to better align with the center of the border-radius? yes! ✅ 
> * Is there supposed to be a drop-shadow? Unclear from the Figma. - in [the UserStats page design](https://www.figma.com/file/qbqbmR3t5XV6eKcpuRj7mG/Group-Stats?type=design&node-id=82-491&mode=design&t=F7Xo9LERmYBJkN4p-0) the related select has a slight drop-shadow I based this implementation off of
> * Dark mode should probably invert the select box to white - I'll follow-up with Sean on dark mode and additional option styling

mcbouslog:
I think I noticed the Storybook zoom caused a lot of jankiness with the drop contents, while the browser zoom was ok, not sure what that's about, but noting.

mcbouslog:
Noting I added the brand background color styling to connect the bottom of the button to the drop content when the select is open.
screenshotOfSelect

PR #5919, opened 2024-02-15
fix: Store all drawn marks in the classification

eatyourgreens:
After creating a new mark, add the new mark to the previous annotation value and store this as a new annotation. Should fix a bug raised in this comment: https://github.com/zooniverse/front-end-monorepo/pull/5913#issuecomment-1945219415

You can try this out on How Did We Get Here, in the dev classifier. Draw on both frames of a subject, then check whether the drawn marks persist, in the classification, on the second step of the workflow.
https://local.zooniverse.org:8080/?project=communitiesandcrowds/how-did-we-get-here&env=production

## Expected behaviour

When I create a new mark, `classification.annotations` should be updated to include all the marks that I've drawn for the current task.

## Actual behaviour

When I draw a mark, the new drawing task annotation only contains mark IDs from the active frame. Mark IDs from other frames are silently dropped from the classification.

The marks themselves are still stored on their respective drawing tools, but the references to them are removed from the task annotation.

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
@@ -104,7 +104,7 @@ function InteractionLayer({
     mark.setSubTaskVisibility(false)
     // Add a time value for tools that care about time. For most tools, this value is ignored.
     mark.setVideoTime(timeStamp, duration)
-    const markIDs = marks.map((mark) => mark.id)
+    const markIDs = annotation.value?.map((mark) => mark.id)
     annotation.update([...markIDs, mark.id])
   }
 


eatyourgreens:
@kieftrav @goplayoutside3 I don't have permission to request reviews or add labels to Zooniverse PRs, so I'm pinging you to take a look at this. There might be an urgent issue, affecting live transcription projects, where work is being lost.

I think it's worth double checking whether classifications are broken for live transcription projects, when you draw on more than one page of a subject. It might be that lines are only being saved for the current, active frame. I'm not 100% certain about that, because the code that serialises a classification (before POSTing it to the API) may actually loop through all drawing tools and marks, and serialise them to the saved classification.

From a quick re-read of the code, the saved classifier snapshot only includes mark IDs that were added to the transcription task annotation, so I'm concerned that volunteer work might have been lost.

https://github.com/zooniverse/front-end-monorepo/blob/6b421edae1e92322bf3676cc558612e886983321/packages/lib-classifier/src/store/Classification/Classification.js#L22-L30

kieftrav:
@eatyourgreens @goplayoutside3 - this 1 line does fix the issue.. I want to be clear that _we are not losing any data upon classification submission_.

eatyourgreens:
Excellent. If this is just a rendering bug, then that's less urgent.

kieftrav:
@eatyourgreens - 100% a rendering bug. I couldn't figure out where `annotation.values['markID', 'markID']` was getting modified to only include the current frame's marks. That was the line that did it.

eatyourgreens:
Drawing task annotations are slightly different from other task annotations. They use `types.safeReference` to hold references to marks. Essentially, a drawing task annotation is an array of mark IDs. A side effect of this is that they're updated differently from other tasks, which hold a direct value rather than a reference. 

I think this could be automated by having the drawing annotation model observe `self.task.marks` and update `self.value` whenever that changes. That would remove the manual update, which could be prone to breakage.

eatyourgreens:
`self.task` is a task key in the annotation model (the name is confusing, a legacy of older code.) It can be resolved to an actual task with something like this line.

https://github.com/zooniverse/front-end-monorepo/blob/6b421edae1e92322bf3676cc558612e886983321/packages/lib-classifier/src/plugins/tasks/drawing/models/DrawingAnnotation.js#L17

Once you have the actual task, you can observe `actualTask.marks` for changes, as it's a MobX observable array. Then automate updating the stored annotation value whenever a mark is added or deleted.

eatyourgreens:
I've opened #5923 with those changes to the drawing annotation models. `autorun` might be more robust than manually updating annotations whenever a drawing tool changes, but it's a more involved change than the one-line fix here.

eatyourgreens:
Closing this in favour of #5923.

PR #5963, opened 2024-03-14
Update YourStats to callout to Eras instead of Zoo-stats-graphql 

yuenmichelle1:
## Package
- [x] app-project

N/A

## Describe your changes

Update YourStats to callout to Eras (eras.zooniverse.org) instead of Zoo-stats-graphql. Thanks @mcbouslog  for walking through this with me and additional support. 

## How to Review

# Checklist

## General

## General UX

## Refactoring


packages/app-project/stores/User/UserPersonalization/YourStats/YourStats.spec.js
```diff
@@ -61,42 +50,49 @@ describe('Stores > YourStats', function () {
...
+
+      statsNockScope = nock('https://eras-staging.zooniverse.org')
+      .get(`/classifications/users/${user.id}`)
+      .query({
+        'period': 'day',
+        'project_id': project.id
+      })
+      .reply(200, {
+        data: MOCK_DAILY_COUNTS
+      })
```
eatyourgreens, line 83: Nock mocks the Node `http` module, which means [it can't intercept `fetch` requests](https://stackoverflow.com/a/74656123). My experience with testing `fetch` in Node (see https://github.com/zooniverse/sugar/pull/170) is that I've had to mock responses with `undici`, which is the package that Node 18+ uses to make `fetch` requests.

Nock 14 adds [`fetch` support](https://github.com/nock/nock/pull/2580), but is still in beta.

mcbouslog: Thank you for noting @eatyourgreens ! Sorry all, it has taken me a long time to respond.

Noting my understanding for correction/comment:

Since we don't want tests to trigger actual requests to APIs (and spam the APIs), we stub or mock the code so the tests don't trigger actual requests. Often with FEM apps/libraries requests are made with panoptes-js or the panoptes-client (PJC) which `nock` can mock. The noted libraries/clients make http requests, which can be mocked with nock.

The new/recommended/preferred way to make API requests in JavaScript is with fetch, which we're starting to use in places, like with these changes, lib-user work, and other places. The lib-user work uses hooks, with swr, and doesn't really test the hooks (it should, but that's a different issue). `nock` doesn't work with fetch, though `undici` does. Maybe there are other libraries to help mock fetch requests as well? 

Importantly, the browser uses fetch and node (where the tests run) uses, if installed, node-fetch. So are these tests, running in node, actually making any requests? How can I tell? Once upon a time I used Wireshark, but I'm not sure that's relevant; there's probably an easier way.

So is the next step for these code changes?
1. merge changes knowing the tests will make API requests - I don't like this option
2. refactor tests to use `undici` or different library - doesn't look straightforward, but certainly doable
3. remove relevant tests - since stores are connected, this might mean removing a lot of tests? regardless, I don't like this option
4. refactor PJC ERAS/stats client with auth (similar to talkClient) and use that instead of fetch - I'd prefer to use fetch, I'd prefer not to use the older client more, but if we do this then the ERAS/stats client (from PJC) could be used elsewhere with auth. I don't love this option, but I don't hate it
5. I'm sure there are other options I'm not thinking of?

So it looks like probably 2 or maybe 4 is the next step? @yuenmichelle1 - I'll plan to run with this from here unless you really want to, as this might be new territory for the frontend.

eatyourgreens: >The new/recommended/preferred way to make API requests in JavaScript is with fetch, which we're starting to use in places, like with these changes, lib-user work, and other places. The lib-user work uses hooks, with swr, and doesn't really test the hooks (it should, but that's a different issue). nock doesn't work with fetch, though undici does. Maybe there are other libraries to help mock fetch requests as well?

`fetch` has been in browsers since 2016, I think. It's been in Node since Node 18, a couple of years ago. `fetch` in Node 18+ is implemented with `undici`. `node-fetch` is obsolete now.

>So are these tests, running in node, actually making any requests?

Good question. I'm surprised that they pass, because `nock` doesn't support `fetch` so I wouldn't expect the test requests to receive the mocked response.

mcbouslog: Noting that option 4 to refactor PJC ERAS client similar to the PJC talkClient is not as straightforward as I was hoping. The ERAS response is different than a panoptes or talkAPI response in ways which would require changing [these lines of code](https://github.com/zooniverse/panoptes-javascript-client/blob/main/lib/json-api-client/index.js#L76-L113). It's certainly doable, but since PJC is not the future anything more than minimal effort is questionable.

eatyourgreens: `json-api-client`, in PJC, can only be used with APIs that follow the [JSON:API](https://jsonapi.org/) spec. FWIW, I'd advise against modifying it to support APIs that don't follow that spec.

mcbouslog: 🤦‍♂️ Yes editing the json-api-client outside of the JSON:API spec would have been a bad idea! Thank you for noting.

mcbouslog (approved):
LGTM! I confirmed at https://local.zooniverse.org:3000/projects/zooniverse/gravity-spy/classify/workflow/1610?env=production classifying and YourStats works as expected 👍 🎉 

I'll plan to merge this before tomorrow's deploy (later this PM or early tomorrow AM).

PR #5994, opened 2024-03-21
refactor: sync annotate visibility when workflow changes

eatyourgreens:
Add `workflowSteps.hasAnnotateTask` to the `SubjectViewerStore` model as
a view. Add an `autorun` reaction which syncs `subjectViewer.annotateVisibility`
when `subjectViewer.hasAnnotateTask` changes.

## Package
lib-classifier

# Checklist

## General

## General UX

## Refactoring

### DIFF
--- a/packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
+++ b/packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
@@ -19,7 +19,6 @@ const SubjectViewer = types
     move: types.optional(types.boolean, false),
     rotationEnabled: types.optional(types.boolean, false),
     rotation: types.optional(types.number, 0),
-    showAnnotate: types.optional(types.boolean, true),
     separateFramesView: types.optional(types.boolean, false)
   })
 
@@ -47,7 +46,11 @@ const SubjectViewer = types
       }
       return false
     },
-    
+ 
+    get hasAnnotateTask () {
+      return getRoot(self)?.workflowSteps.hasAnnotateTask
+    },
+
     get interactionMode () {
       // Default interaction mode is 'annotate'
       return (!self.annotate && self.move) ? 'move' : 'annotate'
@@ -67,20 +70,17 @@ const SubjectViewer = types
     }
 
     return {
-      afterRootInitialize () {
-        // afterRootInitialize() enables listening on store.workflowSteps because they are both now initialized
-        reaction(
-          () => getRoot(self)?.workflowSteps.hasAnnotateTask,
-          (val) => {
-            self.setAnnotateVisibility(val)
-          }
-        )
-        
-        // reactions handle any changes to state, not the initial state
-        self.setAnnotateVisibility(getRoot(self)?.workflowSteps.hasAnnotateTask)
-      },
-      
       afterAttach () {
+        function _syncAnnotateVisibility() {
+          // Make sure the right button is active in the ImageToolbar
+          self.setAnnotateVisibility(self.hasAnnotateTask)
+          if (self.hasAnnotateTask) {
+            self.enableAnnotate()
+          } else {
+            self.enableMove()
+          }
+        }
+        addDisposer(self, autorun(_syncAnnotateVisibility))
         createSubjectObserver()
       },
 
--- a/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/ImageToolbar.js
+++ b/packages/lib-classifier/src/components/Classifier/components/ImageToolbar/ImageToolbar.js
@@ -14,12 +14,12 @@ import ZoomOutButton from './components/ZoomOutButton'
 import { useKeyZoom } from '@hooks'
 
 function storeMapper(classifierStore) {
-  return { showAnnotateButton: classifierStore.subjectViewer.showAnnotate }
+  return { hasAnnotateTask: classifierStore.subjectViewer.hasAnnotateTask }
 }
 
 // Generalized ...props here are css rules from the page layout
 function ImageToolbar (props) {
-  const { showAnnotateButton } = useStores(storeMapper)
+  const { hasAnnotateTask } = useStores(storeMapper)
   const { onKeyZoom } = useKeyZoom()
   
   return (
@@ -44,7 +44,7 @@ function ImageToolbar (props) {
         fill
         pad='clamp(8px, 15%, 10px)'
       >
-        {showAnnotateButton && <AnnotateButton />}
+        {hasAnnotateTask && <AnnotateButton />}
         <MoveButton />
         <ZoomInButton />
         <ZoomOutButton />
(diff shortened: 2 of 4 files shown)


eatyourgreens:
@kieftrav this is #5932 rewritten with `autorun`. If you check the console logs, yoou should see that the reaction runs once when it's created, then only runs again whenever `hasAnnotateTask` changes. You shouldn't see `autorun` running excessively, as it's only observing the `hasAnnotateTask` view.

Hopefully this is a helpful example of how to use MobX `autorun` with MobX State Tree.

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
```diff
@@ -67,20 +71,12 @@ const SubjectViewer = types
...
-        )
-        
-        // reactions handle any changes to state, not the initial state
-        self.setAnnotateVisibility(getRoot(self)?.workflowSteps.hasAnnotateTask)
-      },
-      
       afterAttach () {
+        function _syncAnnotateVisibility() {
+          console.log('syncing annotate visibility', self.hasAnnotateTask)
+          self.setAnnotateVisibility(self.hasAnnotateTask)
```
eatyourgreens, line 77: `self.annotateVisibility` is kind of redundant here, as it mirrors the value of `self.hasAnnotateTask`.

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
```diff
@@ -47,7 +47,11 @@ const SubjectViewer = types
       }
       return false
     },
-    
+ 
+    get hasAnnotateTask () {
+      return getRoot(self)?.workflowSteps.hasAnnotateTask
+    },
```
eatyourgreens, line 53: Putting this into a view guarantees that its value won't be set until after the root store has initialised.

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
eatyourgreens: In fact, you could rename the `hasAnnotateTask` view to `annotateVisibility`, and remove this reaction altogether. BUt then, you wouldn't be able to set `subjectViewer.annotateVisibility` from outside the model, which might be important.

kieftrav:
@eatyourgreens - thanks for putting together this PR! I didn't quite understand the possibilities of having `hasAnnotateTask()` be a view on the store and that being the way to get the render details we want at render-time... This made all the background setup / syncing entirely unnecessary.

I tested this in one of my projects that has two workflows - one with drawing, one without - https://localhost:8081/?project=kieftrav%2Fannotate-or-not&workflow=3782.

Curious if you have other thoughts/insights related to MobX that could clean this up further.

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
```diff
@@ -47,43 +46,28 @@ const SubjectViewer = types
...
+ 
+    get hasAnnotateTask () {
+      const canAnnotate = getRoot(self)?.workflowSteps.hasAnnotateTask
+
+      // Make sure the right button is active in the ImageToolbar
+      if (canAnnotate) {
+        self.enableAnnotate()
+      } else {
+        self.enableMove()
+      }
```
eatyourgreens, line 58: It's very bad practice to use actions inside a getter like this. Views should not change state, just return derived state that can then be observed/reacted to.

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
```diff
@@ -47,43 +46,28 @@ const SubjectViewer = types
...
-    function createSubjectObserver () {
-      const subjectDisposer = autorun(() => {
-        const validSubjectReference = isValidReference(() => getRoot(self).subjects.active)
-        if (validSubjectReference) {
-          const subject = getRoot(self).subjects.active
-          self.resetSubject(subject)
-        }
-      }, { name: 'SubjectViewerStore Subject Observer autorun' })
-      addDisposer(self, subjectDisposer)
-    }
```
eatyourgreens, line 67: I'm pretty sure you need this reaction. It resets the viewer when the active subject changes.

eatyourgreens (commented):
A couple of comments. I think you've removed an action that was being used by the subject viewer, and MST views should never change model state. Changing state should only happen inside actions.

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
eatyourgreens: Resetting the subject like this might also mean that this model is managing per-subject state that is better off stored in individual subjects. However, rewriting the data model is outside the scope of this PR. 

Whenever you see `getRoot(self)` in a model, that’s a code smell implying that the data model isn’t quite right. 

Nodes in the tree should prefer to inherit state from their parents, and rarely interact with siblings.

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
eatyourgreens: As an alternative to `autorun`, you could also handle side effects in your React components, ie. pass `hasAnnotateTask` into the image toolbar as an observable prop, and react to changes in React. 

https://mobx.js.org/the-gist-of-mobx.html#3-create-derivations-that-automatically-respond-to-state-changes

kieftrav:
@eatyourgreens / @goplayoutside3 just a heads up that I've had other priorities this last week but intend to return to this either end of this week / next week.

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
kieftrav: I've changed the code to incorporate this approach.

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
kieftrav: I've re-added this reaction.

kieftrav:
@eatyourgreens - I've made the changes to incorporate your feedback and believe this PR follows MobX's practices. I'll test this on a deployed branch and plan to merge this in if it works unless there are other suggestions to incorporate

kieftrav (approved):
Everything LGTM at this point and ready to merge.
- Tested locally using the "kieftrav/annotate-or-not" project + workflows
- Deployed to [FE-Branch](https://fe-project-branch.preview.zooniverse.org/projects/kieftrav/annotate-or-not?env=staging) with everything looking good 
- Fixes issue with separate frames in addition to the refactor.

PR #5998, opened 2024-03-24
fix: ensure upp.settings.workflow_id is a string

### LINKED ISSUE #5972: Workflow Assignment / Leveling Up: expect settings.workflow_id can be int or string
### Current Behavior
The Panoptes API allows `settings.workflow_id` from UserProjectPreferences, which is used for workflow assignment as part of the leveling up feature, to be written as a string or a integer type. While internal assignment of these values by the FE, Panoptes, and Caesar typically use the string -- and hence the type assumed by the FE was string for its current implementation -- project teams using the Python Client (via [ProjectPreferences.save_settings()](https://panoptes-python-client.readthedocs.io/en/latest/panoptes_client.html#panoptes_client.project_preferences.ProjectPreferences.save_settings) function) tend to assign integer values.

If a user has an integer type `settings.workflow_id` value, the FE throws an error (`Error: [mobx-state-tree] No matching type for union ({ workflow_id: (string | undefined?) } | undefined?)`) and the setting is ignored; the user is granted default workflow accessibility / as if they have no assigned workflow -- they cannot access higher workflows.

To reproduce:
- Login as zootester1 (user_id = )
- Visit https://frontend.preview.zooniverse.org/projects/lcjohnso/leveling-up-testbed, where zootester1 is assigned `workflow_id : 22843` (Level 3 on the project)
- Check console for error, confirm that user only has access to Level 1

### Desired Behavior
Either string or int are allowed types for `settings.workflow_id` value.

### Relevant Code
It appears settings.workflow_id is read from UserProjectPreferences here:
https://github.com/zooniverse/front-end-monorepo/blob/85bcd542e65fa06c7a8f5635ef56c854558286f3/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js#L37-L38

eatyourgreens:
Handle the case where `upp.settings.workflow_id` is a number
by converting it to a string before storing it.
- add tests to specify the behaviour of UPP settings.
- use the `numberString` type for Panoptes workflow IDs.
- fix the bug where `workflow_id` errors if it is a number.

## Package
- app-project
- fixes #5972

## How to Review
See #5972 for a test case, I think.

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
+++ b/packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
@@ -5,16 +5,26 @@ import asyncStates from '@zooniverse/async-states'
 import logToSentry from '@helpers/logger/logToSentry.js'
 import numberString from '@stores/types/numberString'
 
-const Preferences = types
+export const Preferences = types
   .model('Preferences', {
     minicourses: types.maybe(types.frozen()),
     selected_workflow: types.maybe(types.string),
     tutorials_completed_at: types.maybe(types.frozen())
   })
 
-const Settings = types
+export const Settings = types
   .model('Settings', {
-    workflow_id: types.maybe(types.string)
+    workflow_id: types.maybe(numberString)
+  })
+  .preProcessSnapshot(snapshot => {
+    if (!snapshot) {
+      return snapshot
+    }
+    const normalisedSnapshot = {
+      ...snapshot,
+      workflow_id: snapshot?.workflow_id?.toString()
+    }
+    return normalisedSnapshot
   })
 
 const UserProjectPreferences = types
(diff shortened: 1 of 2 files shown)


packages/app-project/stores/User/UserPersonalization/UserProjectPreferences/UserProjectPreferences.js
```diff
@@ -5,16 +5,26 @@ import asyncStates from '@zooniverse/async-states'
 import logToSentry from '@helpers/logger/logToSentry.js'
 import numberString from '@stores/types/numberString'
 
-const Preferences = types
+export const Preferences = types
   .model('Preferences', {
     minicourses: types.maybe(types.frozen()),
     selected_workflow: types.maybe(types.string),
```
eatyourgreens, line 11: I think this is also a workflow ID, in which case it should be `numberString` too.

goplayoutside3 (approved):
Looks good! Thanks Jim. I confirmed that while signed in to `zootester`, I have access to all three workflows at https://localhost.zooniverse.org:3000/projects/lcjohnso/leveling-up-testbed?env=production.

The expected workflows are locked in Gravity Spy too: https://localhost.zooniverse.org:3000/projects/zooniverse/gravity-spy?env=production

PR #6127, opened 2024-06-10
fix(project): update collections search asynchronously

eatyourgreens:
Make the collections search, in the popup collections modal, asynchronous by using a ref to store the current search text. This fixes a bug where search results were not correctly synchronised to the stored search term (in component state.)

If you typed 4 characters into the search box, all your collections were shown, because the component didn't wait for new search results before updating the dropdown menu. Now, it waits for new search results from the Panoptes API before updating.

I've also refactored the minimum required search text length (for full-text search) into `MIN_SEARCH_LENGTH`, which makes it easier to manage. 3 is the minimum required by Postgres, but fuzzy text searching can give weird results for 4 or 5-character strings, so increasing it might make the component easier to use. If you increase the value, I'd suggest also increasing the same value  in https://github.com/zooniverse/Panoptes-Front-End/pull/7124, so that the search behaviour matches in PFE.

## Package
- app-project

 - Fixes an issue where typing exactly 4 characters displays all your collections (see #6120.)

## How to Review
Open the collections modal and try any search that is exactly 4 characters. The list of collections now only shows results returned from the Panoptes full-text-search API.

Note that the API response can still miss collection names that include the search term.

https://github.com/zooniverse/front-end-monorepo/assets/59547/d40ca6da-4e73-40ed-8223-903ac8bcd5fc

# Checklist

## General

## General UX

## Bug Fix
- [x] The bug is fixed


packages/app-project/src/shared/components/CollectionsModal/components/SelectCollection/SelectCollection.js
```diff
@@ -4,6 +4,12 @@ import PropTypes from 'prop-types'
 import { useTranslation } from 'next-i18next'
 import { useState } from 'react'
 
+/*
+  Tune this value to determine when a search term
+  is long enough to use Panoptes full-text search.
+*/
+const MIN_SEARCH_LENGTH = 3;
```
eatyourgreens, line 11: Increasing this might make the search results more intuitive.

packages/app-project/src/shared/components/CollectionsModal/components/SelectCollection/SelectCollection.js
```diff
@@ -13,7 +19,7 @@ function SelectCollection ({
   selected = {},
   userID = ''
 }) {
-  const [searchText, setSearchText] = useState('')
+  const searchText = useRef('')
```
eatyourgreens, line 22: This is the fix for all your collections being shown when you type exactly 4 characters. It stops the component from rerendering immediately whenever `searchText` changes.

packages/app-project/src/shared/components/CollectionsModal/components/SelectCollection/SelectCollection.js
```diff
@@ -49,7 +53,12 @@ function SelectCollection ({
     return `${collection.display_name} (${collection.links.owner.display_name})`
   }
 
-  const options = ignorePanoptesFullTextSearch ? collections.filter(collectionNameFilter) : collections
+  /*
+    If the search text is long enough, use fuzzy full-text search. Otherwise, filter collections by display name.
+  */
+  const options = searchText.current.length > MIN_SEARCH_LENGTH
```
mcbouslog, line 59: I don't think this is updating as expected; when I first type 1-3 characters I'd expect it to filter based on display name but the options aren't changing. I think this is because `search.current` is updating in `onTextChange` but it's not triggering a re-render or for whatever reason it's not changing `options` as expected. I'm tinkering and hope to have a solution shortly.

eatyourgreens: You're right. I've pushed a commit that reverts back to `useState`. When I used a ref, the component wouldn't rerender because the returned collections list wasn't changing. The fix is to wait for the search to complete before updating component state, so that the options list updates when the search results are ready.

packages/app-project/src/shared/components/CollectionsModal/components/SelectCollection/SelectCollection.js
```diff
@@ -25,21 +31,20 @@ function SelectCollection ({
   For shorter strings, we request all your collections then filter the display names.
   */
 
-  function onTextChange(text) {
+  async function onTextChange(text) {
     const search = text.trim()
-    onSearch({
+    await onSearch({
```
eatyourgreens, line 36: The fix for the bug is to wait on the API request before updating state.

packages/app-project/src/shared/components/CollectionsModal/components/SelectCollection/SelectCollection.js
eatyourgreens: This component is old enough that React hooks (and functional components) didn't exist when I first wrote it. A better implementation now might be to search collections with `react-query` or `swr`.

mcbouslog: Cool thanks for the quick update! I was messing around similarly, though didn't have it all put together.

mcbouslog (approved):
The collections search is working more as expected with these changes 👍 .

The panoptes full-text search yields some unexpected results, as noted in this comment - https://github.com/zooniverse/front-end-monorepo/issues/6120#issuecomment-2158882470 - but that's unrelated to these changes and the specific issue of seeing all the collections with 4 characters.

Thanks for the fix!

PR #6192, opened 2024-08-03
fix(lib-react-components): AnimatedNumber displays async values as 0

### LINKED ISSUE #6190: Project Stats: Project Stats numbers (Classifications Today / Classifications Total) don't show on initial Next.js page fetch
## Inconsistent UI Bug?

Package: `app-project` and/or `lib-react-components`
Confirmed as of commit: 
Also see: #5974 (previous fix for AnimatedNumbers), #5971 (similar issue)
Possibly related Talk issue: https://www.zooniverse.org/talk/17/3402644 ?

In certain cases _when the Next.JS server first fetches a page,_ it's possible for the Project Stats ("Classifications Today" and "Classifications Total" count on the Classify page) to not display accurate numbers, despite fetching the correct data.

Replicating this issue is a faff, but can be done on localhost:
- Tested using macOS + Chrome 126 _non-Incognito,_ targeting app-project, with a logged-in session saved. (i.e. when I open the target URL, I'm already logged in with an account that has already made 5+ classifications today.)
- Start a FRESH instance of the local dev server (`yarn dev`)
- Open a new window in Chrome.
- Open target URL: https://local.zooniverse.org:3000/projects/fulsdavid/the-daily-minor-planet/classify/workflow/22354?env=production (Daily Minor Planets)
  - A hydration warning should appear.
  - Scroll down to the Project Stats; it should display inconsistent numbers (see screenshot below)
- Refresh the page. Project stats should be normal.

_Screenshot: Project Stats show 0 Classifications Today and 0 Classifications Total, but the bar chart indicates otherwise._

image

### Extended Testing

<details>
<summary>Asides from the issue seemingly stemming from the "initial Next.JS page fetch", additional testing shows that the Project Stats otherwise works fine.</summary>

- Tested with macOS + Chrome 126 on https://www.zooniverse.org/projects/fulsdavid/the-daily-minor-planet/classify/workflow/22354 and [local](https://local.zooniverse.org:3000/projects/fulsdavid/the-daily-minor-planet/classify/workflow/22354?env=production) (app-project, master)
- I'm able to load the FEM Classify page (while already logged in) and see the initial Classifications Today and Total counts.
While logged in, I can make classifications, and both the Classifications Today and Total counts update as expected. (About 5 classifications per test)
- I can log out, and the counts will reset to 0
- I can login to the same account, or to a different account, and the Classifications Today and Total counts update as expected.
</details>

### Dev Notes

Thoughts 1: I don't have enough expertise on the Next.JS infrastructure to understand why/how the fetch process would be a problem. Is it a hydration thing?

Even if it is, I can't tell if it's anything user-facing (i.e. if it affects the non-dev servers), and how I'd confirm/check that.

Thoughts 2: could it be possible that in certain scenarios, the animation of the `<AnimatedNumbers>` used in the project stats is interrupted, causing the numbers to "freeze" at 0 (for small counts) or appear to "lag" (for larger counts)?

I'm starting to think if it's worth removing **AnimatedNumbers** altogether, if it's a potential cause of problems.

Thoughts 3: this may issue or may not be related to the [Talk post that initiated this investigation](https://www.zooniverse.org/talk/17/3402644), since that volunteer said that despite refreshing their page, the lag still persisted.

> But since then it has started to fall behind the 'classifications today' counter, until it was lagging behind by 55 classifications. I tried refreshing the page, clearing my cache, waiting for an hour and refreshing again, none of which made any difference

But then again:

> Logging out of my account and back in again did seem to do something, though - the total classifications counter is now off by just two classifications. Better, but still not quite what I wanted.

### Status

~~Unknown priority.~~ Impact level unclear, but some volunteers are understandably upset when this occurs. Replicability/consistency of the issue is dubious since it relies on some Next.JS infrastructure shenanigans that I can only consistently replicate on my localhost.

[UPDATE (1 Aug):](https://github.com/zooniverse/front-end-monorepo/issues/6190#issuecomment-2264002520) now set to **low priority.**

eatyourgreens:
- Refactor `AnimatedNumber` to add missing `value` and `duration` effect dependencies. Declare `animateValue` as a pure function outside the component, so that it doesn’t generate a new function on every render.  Refactor the `useEffect` hook to minimise side effects.
- Check that the current and initial values are different before animating. Update the displayed value when animation is complete, so that a render always returns the latest value, not 0. Animate numbers with a delay when the displayed value is 0 and prefers reduced motion is off. Otherwise, display changes immediately eg. after submitting a classification and incrementing your counts.
- Add a `DeferredAnimation` story, which reproduces #6190.
- Add a status live region to the `YourStats` component, so that changes to your classification counts are announced in screenreaders.
- Add missing tests for the `AnimatedNumber` component.

## Package
- lib-react-components

- fixes #6190.
- fixes https://github.com/zooniverse/front-end-monorepo/issues/5971.

## How to Review
The behaviour of `AnimatedNumber` isn't changed here, so the stories in the storybook should be unchanged. Values are animated with a delay when the component first loads, then updated without a delay thereafter. If a value loads asynchronously, eg. from an external API, the first animation waits for the value to be non-zero before running. You can use the new `DeferredAnimation` story to both reproduce that bug and verify the fix. If an animation is interrupted by a user action (eg. submitting a classification) then the animation completes before updating to show the new value. You can check that with the `UpdateTheValue` story (and its tests.)

Bootstrap the monorepo and run a production build of the project app to test this out in a Next.js app. The original bug triggers when your stats load from Zooniverse offscreen, so delete any user stats that are cached in session storage (to force user preferences to load from Zooniverse) and wait until stats and preferences have loaded before scrolling down to the Your Stats widget. The bug won't trigger if Your Stats is already in the viewport when stats load, or if stats and preferences are loaded from session storage.

`yarn dev` runs the Next.js app in strict mode, which breaks the intersection observer that handles animations. ~~I didn't investigate that, but~~ strict mode runs `useEffect` twice so it seems there's a bug in the `useEffect` hook, in `AnimatedNumber`, when it is run more than once (~~which isn't fixed here.~~) In dev mode, your personal classification stats are always displayed as 0 (see #6190.) I’ve added a check for the case where the component might try to animate from 0 to 0, when the displayed value is 0 but `value` hasn’t loaded yet. That should fix the stuck counter in both strict mode and in production. I've also added a story that reproduces #6190 by rendering a deferred value. 

# Checklist

## General

## General UX

## Bug Fix
- [x] The bug is fixed

### DIFF
--- a/packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
+++ b/packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
@@ -14,6 +14,63 @@ export { prefersReducedMotion }
 function formatValue(num) {
   return format(',d')(num)
 }
+
+/**
+ * Use d3 to animate `element` from `initialValue` to `value` over `duration` in milliseconds.
+ * @param {HTMLElement} element 
+ * @param {number} initialValue 
+ * @param {number} value
+ * @param {number} duration 
+ * @returns a Promise that resolves once the animation is complete.
+ */
+function animateValue(element, initialValue, value, duration) {
+  if (value === initialValue) {
+    return
+  }
+  return new Promise((resolve) => {
+    select(element)
+      .data([value])
+      .transition()
+      .duration(duration)
+      .textTween(() => {
+        const interpolator = interpolate(initialValue, value)
+        return t => {
+          const interpolatedValue = interpolator(t)
+          if (interpolatedValue === value) {
+            resolve() // animation complete!
+          }
+          const niceValue = formatValue(interpolatedValue)
+          return niceValue
+        }
+      })
+  })
+}
+
+/**
+ * Use d3 to tween `element` from `initialValue` to `value`.
+ * @param {HTMLElement} element 
+ * @param {number} initialValue
+ * @param {number} value 
+ * @returns a Promise that resolves once the animation is complete.
+ */
+function lessAnimation(element, initialValue, value) {
+  if (value === initialValue) {
+    return
+  }
+  return new Promise((resolve) => {
+    select(element)
+      .data([value])
+      .transition()
+      .duration(0)
+      .textTween(() => {
+        return () => {
+          resolve() // animation complete!
+          return formatValue(value)
+        }
+      })
+  })
+}
+
 function AnimatedNumber({ duration = 1000, value }) {
   const numRef = useRef(null)
   const initialValueRef = useRef(0)
@@ -23,61 +80,23 @@ function AnimatedNumber({ duration = 1000, value }) {
   useEffect(() => {
     // If we already animated the number once, don't observe intersection
     // This could happen if the value prop updates, but page did not refresh
-    if (animated) return
+    if (animated || value === initialValue) return
 
     const numElement = numRef.current
 
-    function animateValue() {
-      if (value === initialValue) {
-        return
-      }
-      select(numElement)
-        .data([value])
-        .transition()
-        .duration(duration)
-        .textTween(() => {
-          const interpolator = interpolate(initialValue, value)
-          return t => {
-            const interpolatedValue = interpolator(t)
-            if (interpolatedValue === value) {
-              setAnimated(true) // animation complete!
-              initialValueRef.current = value
-            }
-            const niceValue = formatValue(interpolatedValue)
-            return niceValue
-          }
-        })
-    }
-  
-    function lessAnimation() {
-      if (value === initialValue) {
-        return
-      }
-      select(numElement)
-      .data([value])
-      .transition()
-      .duration(0)
-      .textTween(() => {
-        return () => {
-          setAnimated(true) // animation complete!
-          initialValueRef.current = value
-          return formatValue(value)
-        }
-      })
-    }
-
-    const intersectionObserver = new window.IntersectionObserver(entries => {
+    const intersectionObserver = new window.IntersectionObserver(async entries => {
       // If intersectionRatio is 0, the target is out of view and we do not need to do anything.
       if (entries[0].intersectionRatio <= 0) return
 
       // Once target element is in viewport, animate it then unobserve
       if (!prefersReducedMotion() && !animated) {
-        animateValue()
-        intersectionObserver.unobserve(numElement)
+        await animateValue(numElement, initialValue, value, duration)
       } else {
-        lessAnimation()
-        intersectionObserver.unobserve(numElement)
+        await lessAnimation(numElement, initialValue, value)
       }
+      setAnimated(true)
+      initialValueRef.current = value
+      intersectionObserver.unobserve(numElement)
     })
 
     intersectionObserver.observe(numElement)
--- a/packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
+++ b/packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
@@ -27,74 +27,47 @@ function animateValue(element, initialValue, value, duration) {
   if (value === initialValue) {
     return
   }
-  return new Promise((resolve) => {
-    select(element)
-      .data([value])
-      .transition()
-      .duration(duration)
-      .textTween(() => {
-        const interpolator = interpolate(initialValue, value)
-        return t => {
-          const interpolatedValue = interpolator(t)
-          if (interpolatedValue === value) {
-            resolve() // animation complete!
-          }
-          const niceValue = formatValue(interpolatedValue)
-          return niceValue
-        }
-      })
-  })
+  const interpolator = interpolate(initialValue, value)
+  return select(element)
+    .data([value])
+    .transition()
+    .duration(duration)
+    .textTween(() => t => formatValue(interpolator(t)))
+    .end()
 }
 
-/**
- * Use d3 to tween `element` from `initialValue` to `value`.
- * @param {HTMLElement} element 
- * @param {number} initialValue
- * @param {number} value 
- * @returns a Promise that resolves once the animation is complete.
- */
-function lessAnimation(element, initialValue, value) {
-  if (value === initialValue) {
-    return
-  }
-  return new Promise((resolve) => {
-    select(element)
-      .data([value])
-      .transition()
-      .duration(0)
-      .textTween(() => {
-        return () => {
-          resolve() // animation complete!
-          return formatValue(value)
-        }
-      })
-  })
-}
+const initialValue = 0
 
-function AnimatedNumber({ duration = 1000, value }) {
+function AnimatedNumber({ duration = 1000, value = 0 }) {
   const numRef = useRef(null)
-  const initialValueRef = useRef(0)
-  const initialValue = initialValueRef.current
   const [animated, setAnimated] = useState(false)
+  // render the value directly after the animation is complete.
+  const displayedValue = animated ? value : initialValue
 
   useEffect(() => {
-    // If we already animated the number once, don't observe intersection
-    // This could happen if the value prop updates, but page did not refresh
-    if (animated || value === initialValue) return
-
+    /*
+     * We're animating the DOM node content outside of the React render cycle
+     * so use node.textContent to check if we've already rendered the value.
+     */
     const numElement = numRef.current
+    if (animated) return // only run the intersection observer once.
+    if (formatValue(value) === numElement.textContent) return // nothing to animate yet.
 
     const intersectionObserver = new window.IntersectionObserver(async entries => {
       // If intersectionRatio is 0, the target is out of view and we do not need to do anything.
       if (entries[0].intersectionRatio <= 0) return
 
-      // Once target element is in viewport, animate it then unobserve
-      const animationFunction =
-        !prefersReducedMotion() && !animated ? animateValue : lessAnimation
-      await animationFunction(numElement, initialValue, value, duration)
-      setAnimated(true)
-      initialValueRef.current = value
+      /*
+       * Once target element is in viewport, animate it then unobserve.
+       * Only run the full animation if prefers reduced motion is false
+       * and d3 hasn't already started an animation.
+       */
+      const animationInProgress = numElement.textContent !== formatValue(initialValue)
+      const animationDuration =
+        !prefersReducedMotion() && !animationInProgress ? duration : 0
       intersectionObserver.unobserve(numElement)
+      await animateValue(numElement, initialValue, value, animationDuration)
+      setAnimated(true)
     })
 
     intersectionObserver.observe(numElement)
@@ -102,9 +75,9 @@ function AnimatedNumber({ duration = 1000, value }) {
     return () => {
       intersectionObserver.disconnect()
     }
-  }, [numRef.current, animated, value, duration])
+  }, [initialValue, value, duration])
 
-  return <span ref={numRef}>{!animated ? initialValue : formatValue(value)}</span>
+  return <span ref={numRef}>{formatValue(displayedValue)}</span>
 }
 
 AnimatedNumber.propTypes = {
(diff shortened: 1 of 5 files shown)


packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
```diff
@@ -75,7 +78,7 @@ function AnimatedNumber({ duration = 1000, value }) {
     return () => {
       intersectionObserver.disconnect()
     }
-  }, [numRef.current, animated])
+  }, [numRef.current, animated, value, duration])
```
eatyourgreens, line 81: React refs aren’t reactive values, so I don’t think `numRef` needs to be declared as a dependency here.

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
```diff
@@ -4,58 +4,67 @@ import { interpolate } from '@visx/vendor/d3-interpolate'
...
     // If we already animated the number once, don't observe intersection
     // This could happen if the value prop updates, but page did not refresh
     if (animated) return
 
     const numElement = numRef.current
 
+    function animateValue() {
+      if (value === initialValue) {
+        return
+      }
```
eatyourgreens, line 32: This should fix #6190 by holding off the animation until `value`  has loaded from Zooniverse APIs.

eatyourgreens:
With the changes here, you should see your stats show up correctly in dev mode. Screen recording from desktop Firefox.


https://github.com/user-attachments/assets/3f576637-f0b1-4cc8-b63d-3720f9232765

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
```diff
@@ -4,78 +4,106 @@ import { interpolate } from '@visx/vendor/d3-interpolate'
...
+  const numRef = useRef(null)
+  const initialValueRef = useRef(0)
+  const initialValue = initialValueRef.current
+  const [animated, setAnimated] = useState(false)
 
   useEffect(() => {
     // If we already animated the number once, don't observe intersection
     // This could happen if the value prop updates, but page did not refresh
-    if (animated) return
+    if (animated || value === initialValue) return
```
eatyourgreens, line 82: One bug in the original code is that it would create the animation callback even if `initialValue` and `value` are both 0. I've fixed that by storing the previous value of `value` in `initialValue`, once the animation completes. The name of `initialValue` might need to be changed, for clarity. I'm not sure.

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
```diff
@@ -4,78 +4,104 @@ import { interpolate } from '@visx/vendor/d3-interpolate'
...
-      } else {
-        lessAnimation()
-        intersectionObserver.unobserve(numElement)
-      }
+      const animationFunction =
+        !prefersReducedMotion && !animated ? animateValue : lessAnimation
+      await animationFunction(numElement, initialValue, value, duration)
+      setAnimated(true)
+      initialValueRef.current = value
+      intersectionObserver.unobserve(numElement)
```
eatyourgreens, line 96: Statements in JavaScript have side effects, but expressions don't, so I changed the logic to use a ternary expression rather than a if statement. See [Rethinking the JavaScript Ternary Operator](https://jrsinclair.com/articles/2021/rethinking-the-javascript-ternary-operator/) by James Sinclair.

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.stories.js
```diff
@@ -44,3 +45,13 @@ export const ScrollDown = () => {
...
+
+export const DeferredAnimation = () => {
+  const [value, setValue] = useState(0)
+  setTimeout(() => setValue(700000000), 2000)
+  return (
+    <Box pad={{ vertical: '120vh' }}>
+      <AnimatedNumber duration={4000} value={value} />
+    </Box>
+  )
+}
```
eatyourgreens, line 57: If you checkout the first commit on this branch, you can use this story to reproduce the original bug. On the main branch, this story always displays 0.

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.stories.js
```diff
@@ -2,6 +2,7 @@ import { Box } from 'grommet'
 import { useState } from 'react'
 
 import AnimatedNumber from './AnimatedNumber'
+import { set } from 'lodash'
```
eatyourgreens, line 5: Stray import from Copilot?

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
eatyourgreens: `value` can still continue to increment as we fetch fresh classification statistics from the API, or as local values increment, so `animated` still needs to be checked, if we only want to animate once.

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
eatyourgreens: The side effect, in the original code, was animating the content of a DOM node outside of the React render cycle. I guess that's still happening here, but the `await` might be easier to read.

eatyourgreens:
The bug that’s fixed happened because #5947 and #5974 broke the Rules of Hooks. It could have been caught by linting with the React hooks rules for ESLint. 
https://github.com/facebook/react/blob/main/packages/eslint-plugin-react-hooks/README.md

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
eatyourgreens: Renamed `initialValue` to `displayedValue`, since it is the value that's shown to a volunteer.

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
```diff
@@ -4,80 +4,94 @@ import { interpolate } from '@visx/vendor/d3-interpolate'
...
+       * Once target element is in viewport, animate it then unobserve.
+       * Only run the full animation if prefers reduced motion is false
+       * and the displayed value is 0.
+       */
+      const isFirstTime = numElement.textContent === '0'
+      const animationFunction =
+        !prefersReducedMotion() && isFirstTime ? animateValue : lessAnimation
+      intersectionObserver.unobserve(numElement)
+      await animationFunction(numElement, displayedValue, value, duration)
+      setDisplayedValue(value)
```
eatyourgreens, line 84: The text content of `numElement` is `formatValue(value)` when the animation promise resolves. We need to update the value that React will render on the next render, even though it's exactly the same.

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
```diff
@@ -4,80 +4,94 @@ import { interpolate } from '@visx/vendor/d3-interpolate'
...
+  if (value === initialValue) {
+    return
   }
+  const interpolator = interpolate(initialValue, value)
+  return select(element)
+    .data([value])
+    .transition()
+    .duration(duration)
+    .textTween(() => t => formatValue(interpolator(t)))
+    .end()
```
eatyourgreens, line 36: `transition().end()` returns a Promise that resolves when `transition()` completes, and rejects if the animation is interrupted. That's super useful.
https://observablehq.com/@d3/transition-end

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
```diff
@@ -4,80 +4,94 @@ import { interpolate } from '@visx/vendor/d3-interpolate'
...
-    .textTween(() => {
-      return () => {
-        setAnimated(true) // animation complete!
-        return formatValue(value)
-      }
-    })
-  }
+    .textTween(() => () => formatValue(value))
+    .end()
+}
```
eatyourgreens, line 56: `lessAnimation` is essentially `animateValue` with `duration` set to 0.

eatyourgreens: Replaced it with `animateValue(element, initialValue, value, 0)`.

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
```diff
@@ -4,80 +4,80 @@ import { interpolate } from '@visx/vendor/d3-interpolate'
...
-      } else {
-        lessAnimation()
-        intersectionObserver.unobserve(numElement)
-      }
+      /*
+       * Once target element is in viewport, animate it then unobserve.
+       * Only run the full animation if prefers reduced motion is false
+       * and the displayed value is 0.
+       */
+      const isZero = numElement.textContent === formatValue(initialValue)
```
eatyourgreens, line 64: The `UpdateTheValue` tests uncovered a race condition where d3 had inserted '10' into `numElement` but `animated` was false, so the animation ran once from 0 to 10, then again from 0 to 11. Testing the value that's been inserted by d3 guards against this and runs the test as expected.

eatyourgreens: You can reproduce that bug in the Storybook, if you click the Update button while the animation is running.

eatyourgreens: I've changed the behaviour of the component so that an animation will always run to completion, even if `value` changes while the animation is running.

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.js
```diff
@@ -4,80 +4,82 @@ import { interpolate } from '@visx/vendor/d3-interpolate'
...
     intersectionObserver.observe(numElement)
 
     return () => {
       intersectionObserver.disconnect()
     }
-  }, [numRef.current, animated])
+  }, [initialValue, value, duration])
 
-  return <span ref={numRef}>{!animated ? initialValue : formatValue(value)}</span>
+  return <span ref={numRef}>{formatValue(displayedValue)}</span>
```
eatyourgreens, line 82: Note that if an animation is running, d3 will override `displayedValue` here with the output from the `textTween` function. If the `value` prop changes while d3 is animating the number, you want to wait for that animation to complete before updating `displayedValue`.

eatyourgreens:
I couldn't figure out how to mock scrolling an offscreen element into view with tests that run in Node, so the tests only test the case where the animated element is in the viewport at all times. You can use the storybook to test offscreen rendering in a browser.

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.spec.js
```diff
@@ -0,0 +1,187 @@
...
+        })
+        const Story = composeStory(Stories.DeferredAnimation, Meta)
+        render(<Story />)
+        expect(prefersReducedMotion()).to.be.true()
+        const number = document.querySelector('span')
+        expect(number.textContent).to.equal('0')
+        // wait 2000ms for the value prop to change
+        clock.tick(2000)
+        clock.restore()
+        expect(number.textContent).to.equal('0')
```
mcbouslog, line 98: This is on me, but to confirm my understanding it's still `'0'` after 2 seconds because it's animating, then it's the new value (line 100) after the animation has concluded, even though the new value was passed to the component at 2 seconds (per DeferredAnimation story).

mcbouslog (approved):
This LGTM! Great tests, thank you for those! Inline comments very helpful.

Confirmed fix locally with lib-react-components storybook and with a project in app-project.

packages/lib-react-components/src/AnimatedNumber/AnimatedNumber.spec.js
eatyourgreens: It’s 0 here because it’s still waiting to start the animation. The `value` changes at 2s, which starts the animation. Previously, the animation started while the value was 0, and d3 would insert a 0 into the element after a 1s wait, overwriting the actual value. 

In the storybook, you can see that by scrolling the animated number into view before `value` changes. Previously, the displayed value stayed at 0, but now it starts animating once the prop changes.

PR #6194, opened 2024-08-06
DRAFT: ADR 62 - Volumetric Point Tool

kieftrav:
## Package

lib-classifier

## Describe your changes

Distillation of Volumetric Viewer + Volumetric Point Tool conversation at ZTM. No effect on any code in the codebase so minimal review requirements from a deployment perspective.

### DIFF
--- a/docs/arch/adr-61.md
+++ b/docs/arch/adr-61.md
@@ -1,0 +1,52 @@
+# ADR 61: Volumetric Point Tool
+
+## Status
+
+Proposed: 2024-08-06
+
+## Context
+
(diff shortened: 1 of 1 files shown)


goplayoutside3 (changes requested):
I'd like to get #6193 approved first, but jumping over from reviewing that ADR, this ADR is focused on the Tool needed to classify a 3D subject. In the Zooniverse classifier, the tool, annotation, and task are more closely related to the classification object eventually sent to panoptes, and should be discussed in detail here. Yes, a volunteer will be using the tool on top of the 3D viewer, but discussion of the models can be isolated from explanations about UI decisions.

Here are some guiding questions for this ADR to be a helpful reference in the future:
- Are you creating a new Subject model in the classifier, or using existing data subject models?
- Are you creating a new Annotation model or do the existing models in the MobX Store work?
- Does this new tool require a new task type? How does that type fit into the project builder?
- What parts of the existing drawing tools in FEM's classifier suffice, or are you creating an entirely new tool?

There's a lot to consider about viewer + tool + annotation + classification + task type, but I want to avoid incomplete discussions across multiple ADRs.

PR #6276, opened 2024-09-10
fix(classifier): proportional height for subject images

### LINKED ISSUE #6275: Science Scribbler: subject viewer changes size during classification
FWIW: I tried to see whether this was specific to the Transcription Task, or if there were other weird Firefox-specific 'page jump' issues. Here's an example from Science Scribbler (note change in image size between marking + question task). NB I don't think this is related (it happens on Chrome, too), but wanted to share just in case, since it seems like an odd thing to happen (not sure why subject image size would change mid-classification):

https://github.com/user-attachments/assets/64440326-f251-480c-8d8d-e5c0120680e3

_Originally posted by @snblickhan in https://github.com/zooniverse/front-end-monorepo/issues/6271#issuecomment-2334725408_

eatyourgreens:
When 'limit subject height' is enabled, style subject images with a proportional height of 90% of the viewport height, up to a maximum size of the original image height.

- closes #6275.

## How to Review
Subject height should now be fixed during all workflow tasks, but never larger than the viewport.

https://local.zooniverse.org:3000/projects/chloezycheng/science-scribbler-synapse-safari/classify/workflow/26732

https://github.com/user-attachments/assets/bb7f1df2-0ba5-487b-bed6-02bce9637440

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.js
@@ -54,7 +54,7 @@ function SingleImageViewer({
       >
         <PlaceholderSVG
           focusable
-          maxHeight={svgMaxHeight}
+          height={svgMaxHeight}
           maxWidth={limitSubjectHeight ? `${width}px` : '100%'}
           onKeyDown={onKeyDown}
           tabIndex={0}


eatyourgreens:
A fixed height subject breaks the layout on small screens, so this won't work on a phone.

goplayoutside3 (commented):
Unfortunately changing from `max-height` to `height` results in a bug where the viewer no longer adjusts to viewports < 1000px. The subject image must adjust as the viewport width adjusts. See https://local.zooniverse.org:3000/projects/fr/fulsdavid/the-daily-minor-planet/classify/workflow/22354?env=production&demo=true as an example where subject images are always square images. 

 🚨 Here's a screenshot of my phone's Safari browser with this branch run locally. The image is squished:  
Daily Minor Planet on a mobile device

eatyourgreens:
I'm not sure. CSS `height` must always be specified alongside `max-height`, otherwise the browser calculates a height, resulting in subjects that change height as you move through a workflow (#6275.) I'd suggest fixing the invalid CSS styles, but I'm not sure what the height should be set to on a phone.

eatyourgreens:
I'd suggest using `maxHeight` on smaller screens but that introduces #6061, which makes Squirrel Mapper much harder on a phone.

eatyourgreens:
I played around with this a bit more tonight but couldn't get it to work on tablet-sized screens.

PR #6299, opened 2024-09-12
PanoptesJS: remove superagent

shaunanoordin:
## PR Overview

Package: panoptes-js

This PR attempts to remove `superagent` from the PanoptesJS client

- Reason: `superagent` is a dependency that we don't technically need anymore, since the standard `fetch` API works fine, while maintaining superagent can be a bit of a faff.
- Observation: superagent() is _only_ used in the get(), post(), put(), and del() (i.e. delete) functions of `panoptes.js`, so replacing the dependency should be very easy.
  - _(Spoiler: it wasn't)_
- Priority: low priority, _for now._ Removing superagent is a "nice to have" item that will eventually escalate to "gotta get it done" once the maintenance cost gets too high in the future (or the library stops getting updates, whichever comes first).

### Dev Notes

As of 12 Sep 2024, this PR is in an experimental stage. The biggest lessons we've learnt are:
1. it's actually easy to replace the `superagent()` calls in `panoptes.js`, as initially observed.
2. however, there are MULTIPLE places in the FEM code that expects _a very specific superagent-shaped response_ from `panoptes.get()/panoptes.put()/panoptes.post()/panoptes.del()`, 

<details>
<summary>Comparison of the default response data we get from superagent vs the default response data we get from fetch().</summary>

Superagent response:

Screenshot 2024-09-12 at 15 56 50

Fetch API response:

Screenshot 2024-09-12 at 16 00 18
</details>

So in practice, we either need to...
1. accept the cost of going through the hundred(s?) instances in the FEM code ([and any other project that uses PanoptesJS](https://github.com/zooniverse/community-catalog)) that expects a superagent response and make changes there, OR
2. create a transformer/adapter that transform Fetch API responses into superagent-compatible responses.

### Status

Experimental. Put on pause as of 12 Sep 2024 while higher priority items in PanoptesJS is addressed.

As of commit [96096e: panoptes.get experimental: replace superagent with fetch](https://github.com/zooniverse/front-end-monorepo/commit/96096edae89583059e950d7a8a3d97497360fee5), the get() function has superagent() successfully replaced with fetch(), with a very rudimentary fetch-response-to-superagent-response transformer in place. You can run `lib-classifier > yarn dev` and [load a local test project,](https://local.zooniverse.org:8080/?env=staging&project=darkeshard%2Ftest-project-2022&workflow=3581) and the data will be fetched fine _for this instance._

### DIFF
--- a/packages/lib-panoptes-js/src/panoptes.js
+++ b/packages/lib-panoptes-js/src/panoptes.js
@@ -40,6 +40,7 @@ function parseHeaders(headers = {}) {
   return httpHeaders
 }
 
+// Note: if it makes the code more readable, we can merge getQueryParams into getQueryString.
 function getQueryParams (query) {
   const defaultParams = { admin: checkForAdminFlag(), http_cache: true }
 
@@ -50,6 +51,18 @@ function getQueryParams (query) {
   } else {
     return defaultParams
   }
+}
+
+/*
+Converts a query object (e.g. { foo: 'bar', img: 'http://foo.bar/baz.jpg' })
+into a query string (e.g. "foo=bar&img=http%3A%2F%2Ffoo.bar%")
+ */
+function getQueryString (query) {
+  const queryParams = getQueryParams(query)
+  let queryString = Object.entries(queryParams)
+    .map(([key, val]) => `${encodeURIComponent(key)}=${encodeURIComponent(val)}`)
+    .join('&')
+  return queryString.length > 0 ? `?${queryString}` : ''
 }
 
 // TODO: Consider how to integrate a GraphQL option
--- a/packages/lib-panoptes-js/src/panoptes.js
+++ b/packages/lib-panoptes-js/src/panoptes.js
@@ -57,12 +57,19 @@ function getQueryParams (query) {
 Converts a query object (e.g. { foo: 'bar', img: 'http://foo.bar/baz.jpg' })
 into a query string (e.g. "foo=bar&img=http%3A%2F%2Ffoo.bar%")
  */
-function getQueryString (query) {
+function getQueryString (query, endpoint = '') {
   const queryParams = getQueryParams(query)
   let queryString = Object.entries(queryParams)
+    .filter(([key, val]) => val !== undefined)
     .map(([key, val]) => `${encodeURIComponent(key)}=${encodeURIComponent(val)}`)
     .join('&')
-  return queryString.length > 0 ? `?${queryString}` : ''
+  
+  if (queryString.length > 0) {
+    return endpoint.includes('?')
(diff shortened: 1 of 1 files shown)


packages/lib-panoptes-js/src/panoptes.js
```diff
@@ -52,22 +53,49 @@ function getQueryParams (query) {
...
+    .map(([key, val]) => `${encodeURIComponent(key)}=${encodeURIComponent(val)}`)
+    .join('&')
+  
+  if (queryString.length > 0) {
+    return endpoint.includes('?')
+      ? `&${queryString}`  // If endpoint already has a query '?', prepend the query string with a '&'.
+      : `?${queryString}`  //  Otherwise, prepend with a '?'.
+  }
+  return ''
+}
```
eatyourgreens, line 73: You can use [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) to do this in all browsers and in Node too.

eatyourgreens:
```js
const searchParams = new URLSearchParams(query)
return searchParams.toString()
```

eatyourgreens:
Might be a good idea to upgrade the code to ES6 too. At the moment, it’s written for old versions of Node that are no longer supported.

packages/lib-panoptes-js/src/panoptes.js
```diff
@@ -52,22 +53,49 @@ function getQueryParams (query) {
...
-    .set('Accept', 'application/vnd.api+json; version=1')
-
-  if (headers){
-    request.set(parseHeaders(headers))
-  }
-  const queryParams = getQueryParams(query)
-
-  return request.query(queryParams).then(response => response)
+  const queryString = getQueryString(query, endpoint)
+  const fetchHeaders = Object.assign({
```
eatyourgreens, line 82: `Object.assign` has been replaced by the rest spread operator, which is much faster, in modern JavaScript.

packages/lib-panoptes-js/src/panoptes.js
```diff
@@ -52,22 +53,49 @@ function getQueryParams (query) {
...
+    'Content-Type': 'application/json',
+    'Accept': 'application/vnd.api+json; version=1',
+  }, parseHeaders(headers))
```
eatyourgreens, line 85:
```suggestion
  const fetchHeaders = {
    'Content-Type': 'application/json',
    'Accept': 'application/vnd.api+json; version=1',
    ...parseHeaders(headers))
  }
```

packages/lib-panoptes-js/src/panoptes.js
eatyourgreens: `Object.assign` should only appear in code if you need to support IE.

PR #6491, opened 2024-11-23
fix(lib-classifier): canvas context for SVG drawing tools

### LINKED ISSUE #6490: Drawing is broken for zoomed-in images in Firefox
## Package
_Examples_
- lib-classifier

## Describe the bug
Dragging a drawn mark, in Firefox, doesn't correctly map the cursor position to the enlarged SVG image. It looks like the cursor position is mapped to the original (not zoomed-in) viewport.

https://github.com/user-attachments/assets/0eb80883-8c49-4700-bd00-6f6bc88fca17

## To Reproduce

In Firefox, zoom in, place a point mark, then drag the point to a new position.

Alternatively, zoom in, draw a line, then drag one of the end points to a new position.

## Expected behavior
Drawing works as expected in Chrome and Safari. The dragged element should follow the pointer position.

## Additional context
Dragging interactions get the current transformation matrix from a `svg` element stored in `SVGContext` here:
https://github.com/zooniverse/front-end-monorepo/blob/1862adebef4ddfde8d071669c99038bf3343383c/packages/lib-classifier/src/plugins/drawingTools/components/draggable/draggable.js#L51

Creating a mark here gets the CTM from the `DrawingCanvas` component in the `InteractionLayer` component.
https://github.com/zooniverse/front-end-monorepo/blob/1862adebef4ddfde8d071669c99038bf3343383c/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js#L88

I don't think those are referencing the same DOM node, so that could account for differences in behaviour between creating a mark and dragging a mark. They probably should be using the same DOM node and transformation matrix.

eatyourgreens:
Make sure that `SVGContext.canvas` is a reference to the SVG `rect` element that's used as a drawing canvas, so that creating new drawing marks and editing existing marks both use the same screen CTM when calculating the pointer position in SVG space.

This fixes a bug in Firefox where the `draggable` decorator doesn't correctly track the active pointer, in the context of a zoomed SVG image.

## Package
- lib-classifier

- closes #6490.

## How to Review
On I Fancy Cats, you should be able to draw while zoomed in, without seeing any of the mismatched positioning described in #6490. You can try it out by drawing an ellipse while zoomed in, or by placing a point then dragging the point to a new position.

Try it out in Firefox and Chrome, while zoomed in and also after rotating the image. Pointer interactions should be correctly mapped to the SVG image now, regardless of any external transformation.

While drawing, in any browser, you should be able to pan and zoom with the keyboard too.

The stroke width of drawn marks should also stay the same as you zoom in. Lines shouldn’t be thicker at increased magnification. 

# Checklist

## General

## General UX

## Bug Fix
- [x] The bug is fixed

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
@@ -4,6 +4,7 @@ import { useContext, useRef, useState } from 'react';
 import styled, { css } from 'styled-components'
 
 import SVGContext from '@plugins/drawingTools/shared/SVGContext'
+import { convertEvent } from '@plugins/drawingTools/components/draggable/draggable'
 import DrawingToolMarks from './components/DrawingToolMarks'
 import TranscribedLines from './components/TranscribedLines'
 import SubTaskPopup from './components/SubTaskPopup'
@@ -24,38 +25,6 @@ const DrawingCanvas = styled('rect')`
 function cancelEvent(event) {
   event.preventDefault()
   event.stopPropagation()
-}
-
-function createPoint(event) {
-  const { clientX, clientY } = event
-  // SVG 2 uses DOMPoint
-  if (window.DOMPointReadOnly) {
-    return new DOMPointReadOnly(clientX, clientY)
-  }
-  // jsdom doesn't support SVG
-  return {
-    x: clientX,
-    y: clientY
-  }
-}
-
-function getEventOffset(event, canvas) {
-  const svgPoint = createPoint(event)
-  const svgEventOffset = svgPoint.matrixTransform
-    ? svgPoint.matrixTransform(canvas.getScreenCTM().inverse())
-    : svgPoint
-  return svgEventOffset
-}
-
-function convertEvent(event, canvas) {
-  const svgEventOffset = getEventOffset(event, canvas)
-  const svgCoordinateEvent = {
-    type: event.type,
-    x: svgEventOffset.x,
-    y: svgEventOffset.y
-  }
-
-  return svgCoordinateEvent
 }
 
 function InteractionLayer({
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
@@ -1,7 +1,9 @@
 import cuid from 'cuid'
 import PropTypes from 'prop-types'
-import { useRef, useState } from 'react';
+import { useContext, useRef, useState } from 'react';
 import styled, { css } from 'styled-components'
+
+import SVGContext from '@plugins/drawingTools/shared/SVGContext'
 import DrawingToolMarks from './components/DrawingToolMarks'
 import TranscribedLines from './components/TranscribedLines'
 import SubTaskPopup from './components/SubTaskPopup'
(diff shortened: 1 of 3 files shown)


packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -24,6 +26,38 @@ function cancelEvent(event) {
...
+function convertEvent(event, canvas) {
+  const svgEventOffset = getEventOffset(event, canvas)
+  const svgCoordinateEvent = {
+    type: event.type,
+    x: svgEventOffset.x,
+    y: svgEventOffset.y
+  }
+
+  return svgCoordinateEvent
+}
```
eatyourgreens, line 59: This change isn't necessary for the fix, but I thought it made the code easier to follow if it uses the same SVG transformation functions here as are used in the `draggable` decorator.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -43,7 +77,9 @@ function InteractionLayer({
   duration
 }) {
   const [creating, setCreating] = useState(false)
-  const canvas = useRef()
+  const svgContext = useContext(SVGContext)
+  const canvasRef = useRef()
+  svgContext.canvas = canvasRef.current
```
eatyourgreens, line 82: I'm not 100% sure if this is the best way to update a reference within a context object, but it works in both Chrome and Firefox.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
```diff
@@ -98,8 +98,9 @@ function InteractionLayer({
       frame: (multiImageCloneMarkers) ? 0 : frame,
       toolIndex: activeToolIndex
     })
+    console.log(canvasRef.current)
```
eatyourgreens, line 101: Oops.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewer.js
```diff
@@ -68,7 +66,6 @@ function SingleImageViewer({
             transform={transform}
           >
             <SVGImageCanvas
-              ref={canvasLayer}
```
eatyourgreens, line 71: This ref was originally used for dragging interactions, which caused the Firefox bug.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayer.js
eatyourgreens: In fact, the `InteractionLayer` component can import these functions, rather than defining them twice.

eatyourgreens:
Oh, I think that might be because the draggable image is dragged in a different context to draggable drawing marks. Marks should get the CTM from the `rect` that defines the drawing canvas, but the image is rendered separately.

eatyourgreens:
> However, if I remove the changes to SingleImageViewer I think mouse dragging/panning works and Firefox zoomed in drawn marks dragging works as expected.

I tried this quickly, but it breaks the drawing tools. Dragging the image also doesn't work as expected with the CTM that's used by the drawing tools. I'm not sure why the image and drawing tools are using different CTMs.

eatyourgreens:
I'll push my changes up, since they're easy to revert and this branch is already broken anyway.

eatyourgreens:
Since pushing my latest commit, I'm not seeing those problems any more in Firefox, so maybe I've fixed the problem? It's still bothering me a bit that the image and drawn marks are using different transformation matrixes. It seems like both should have the same transformations applied to them.

eatyourgreens:
Thinking about this some more, [`currentTarget`](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget) might be a more robust way to get the DOM node that is currently handling pointer events. It would simplify the interfaces for draggable behaviour too.

mcbouslog (approved):
Confirmed this fixes #6490 , tested in Firefox (confirming bug in Firefox with - https://frontend.preview.zooniverse.org/projects/brooke/i-fancy-cats/classify/workflow/693?env=staging&demo=true , confirming fix locally with https://local.zooniverse.org:3000/projects/brooke/i-fancy-cats/classify/workflow/693?env=staging&demo=true) , Chrome, and Safari. Other drawing and pan and zoom functionality working as expected.

PR #6511, opened 2024-12-03
VV - Implement Classifier Tool/Task

kieftrav:
## Package
- lib-classifier

## Describe your changes
Create the Volumetric Tool/Task and properly wire it into the Classifier. Also creates the basic structure for the Annotation, but that will be refactored/fleshed out in the follow up PR. Specifically, there is a missing `VolumetricAnnotation.spec.js` file that is missing and will be added in the next PR (along with the incorporation of the Volumetric Viewer's annotation model).

## How to Review
- All tests should pass
- Volumetric Task should load in local [Storybook](http://localhost:6006/?path=/story/tasks-volumetric--default)
- Loading up the Classifier locally with my staging [Volumetric Viewer](https://localhost:8080/?project=kieftrav/volumetric-viewer&env=staging) task should load up the TaskArea properly
- Design review will be completed in a later PR when Sean's styles are incorporated fully
- README should be descriptive and helpful

## General

## General UX

## New Feature

## Post-merge


packages/lib-classifier/src/plugins/tasks/components/TaskInput/components/TaskInputLabel/TaskInputLabel.js
```diff
@@ -15,6 +15,7 @@ const StyledTaskInputLabelWrapper = styled.span`
 `
 
 const StyledText = styled(Text)`
+  align-content: center;
   display: block;
```
goplayoutside3, line 19: This is `display: block` not flex, so `align-content` won't apply. Did you mean to include this change?

packages/lib-classifier/src/plugins/tasks/experimental/volumetric/components/VolumetricTask.js
```diff
@@ -0,0 +1,90 @@
...
+function VolumetricTask({ task }) {
+  return (
+    <Box>
+      <StyledInstructionText as="legend" size="small">
+        <Markdownz>{task.instruction}</Markdownz>
+      </StyledInstructionText>
+
+      <TaskInput
+        checked={true}
+        disabled={false}
```
goplayoutside3, line 52: The ReadMe for this components says "The volumetric task is disabled until the subject's JSON content is loaded". You've hardcoded `disabled` as `false` here. How is the task disabled while JSON content is loading? Did you mean to add that feature in a separate PR?

packages/lib-classifier/src/plugins/tasks/experimental/volumetric/components/VolumetricTask.js
```diff
@@ -0,0 +1,90 @@
...
+
+function VolumetricTask({ task }) {
+  return (
+    <Box>
+      <StyledInstructionText as="legend" size="small">
+        <Markdownz>{task.instruction}</Markdownz>
+      </StyledInstructionText>
+
+      <TaskInput
+        checked={true}
```
goplayoutside3, line 51: Is this hardcoded as `true` because there's only ever one active tool in any volumetric project? If that's the case please include a brief code comment here because `checked` and `index` props for TaskInput were originally designed for multi-tool drawing tasks.

packages/lib-classifier/src/plugins/tasks/experimental/volumetric/components/VolumetricTask.spec.js
```diff
@@ -0,0 +1,24 @@
...
+
+    it("should show the label text", function () {
+      render(<DefaultStory />)
+      expect(screen.getByText("3D Viewer")).to.be.ok()
+    })
+
+    it("should show the classification count", function () {
+      render(<DefaultStory />)
+      expect(screen.getByText("InputStatus.drawn")).to.be.ok()
+    })
```
goplayoutside3, line 22: This works for now, but `InputStatus.drawn` also includes the variable `count`. A more robust test would look for `count = 3` (because that's what you've hardcoded it as for now), and when the annotations feature is added for volumetric subjects, then you'd pass `annotations` to the VolumetricTask stories and look for the appropriate count number used in the stories.

packages/lib-classifier/src/plugins/tasks/experimental/volumetric/models/VolumetricTask.js
```diff
@@ -0,0 +1,23 @@
+import cuid from "cuid"
```
goplayoutside3, line 1: A comment just for awareness that if you'e going to continue using `cuid` it is a deprecated third-party package as documented in https://github.com/zooniverse/front-end-monorepo/issues/6493. All of the other task models still use it so this is just an fyi.

packages/lib-classifier/src/store/Project/Project.js
```diff
@@ -26,7 +26,7 @@ const Project = types
     },
 
     get isVolumetricViewer() {
-      return self.experimental_tools.includes('volumetricViewer')
+      return self.experimental_tools.includes('volumetricProject')
```
goplayoutside3, line 29: I noticed the label on your staging project's [admin page](https://www.zooniverse.org/admin/project_status/kieftrav/volumetric-viewer?env=staging) is still `VolumetricViewer`. Are there still updates needed in the PFE codebase to align it with `VolumetricProject`?

goplayoutside3 (commented):
I can run lib-classifier locally or app-project locally and see a viewer + task at https://local.zooniverse.org:3000/kieftrav/volumetric-viewer 👍 

Two blocking things:
1. I'm unable to run lib-classifier's Storybook on this branch. Every story errors with `Error: Cannot find module 'ajv-formats'`. Is that the case for you?
2. The handling of `disabled` in VolumetricTask is hardcoded and therefore contradicts  its Readme statement about disabling the task UI until the subject's json content is loaded.

packages/lib-classifier/src/plugins/tasks/components/TaskInput/components/TaskInputLabel/TaskInputLabel.js
kieftrav: So, when I revert this the alignment looks off. My understanding is that the `align-content` helps with the current element in the context of it's parent `flex` attribute. Could be wrong, but was trying to fix the visual issue you see in the screenshots below. Happy to revert if it should be handled in another way.

missing-align-content
da5-9165-166432b02712)

packages/lib-classifier/src/plugins/tasks/experimental/volumetric/components/VolumetricTask.js
kieftrav: Yes - that's why. Added the comment.

packages/lib-classifier/src/plugins/tasks/experimental/volumetric/components/VolumetricTask.spec.js
kieftrav: Yup - agreed. That'll come in the next PR.

packages/lib-classifier/src/plugins/tasks/experimental/volumetric/models/VolumetricTask.js
kieftrav: Yup - thank you for the cross-reference visibility

packages/lib-classifier/src/plugins/tasks/experimental/volumetric/components/VolumetricTask.js
kieftrav: I updated the README and code... I meant to disable the task until the Subject is loaded. This has been fixed/added.

kieftrav:
> I can run lib-classifier locally or app-project locally and see a viewer + task at https://local.zooniverse.org:3000/kieftrav/volumetric-viewer 👍
> 
> Two blocking things:
> 
> 1. I'm unable to run lib-classifier's Storybook on this branch. Every story errors with `Error: Cannot find module 'ajv-formats'`. Is that the case for you?

No - I did a `yarn panic && yarn bootstrap` and it worked just fine to run `yarn storybook` within the `lib-classifier` directory... I did a quick search on the `ajv-formats` on [NPM](https://www.npmjs.com/package/ajv-formats/v/1.3.1) and it looks like URL/date formatting/parsing... Cruft from some other work maybe?

> 2. The handling of `disabled` in VolumetricTask is hardcoded and therefore contradicts  its Readme statement about disabling the task UI until the subject's json content is loaded.

Good catch. I added in a check for the `subjectReadyState` to get the Subject's loading state. This will also be the basis for the Annotation details that'll be pulled into the next PR.

packages/lib-classifier/src/plugins/tasks/components/TaskInput/components/TaskInputLabel/TaskInputLabel.js
goplayoutside3: Ah I see. Through PR history it looks like the `display: block` was copy pasted from when the label text was a `span` instead of a Grommet Text. This alignment change is fine to keep in this PR, but I recommend removing `display: block` to avoid confusion during code read. TBH it doesn't even need the `margin: 1em 0` either. That was probably an attempt to vertically align the text awhile back, now solved by `align-content` instead.

packages/lib-classifier/src/plugins/tasks/experimental/volumetric/components/VolumetricTask.js
```diff
@@ -0,0 +1,100 @@
...
+  > svg {
+    height: 1.5em;
+    vertical-align: bottom;
+    width: 1.5em;
+  }
+`
+
+function VolumetricTask({ task }) {
+  const stores = useContext(MobXProviderContext)
+  const subjectReadyState = stores.classifierStore?.subjectViewer?.loadingState ?? asyncStates.loading
```
goplayoutside3, line 47: `disabled` is already available as a prop from Task
https://github.com/zooniverse/front-end-monorepo/blob/0a09c9de0186319e2823334a15eb7e0335932c90/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js#L56

No need to hook into context, you should use the `disabled` prop already provided. Examples are in SurveyTask, SimpleDropdownTask, TextFromSubjectTask, SingleChoiceTask, etc.

goplayoutside3, line 47:
```suggestion
function VolumetricTask({ disabled = false, task }) {
```

`disabled` is already available as a prop from Tasks
https://github.com/zooniverse/front-end-monorepo/blob/0a09c9de0186319e2823334a15eb7e0335932c90/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/Tasks.js#L56

No need to hook into context, you should use the `disabled` prop already provided. Examples are in SurveyTask, SimpleDropdownTask, TextFromSubjectTask, SingleChoiceTask, etc.

(Sorry if this shows up as a dupe notification, I just wanted to re-comment with code changes suggested)

goplayoutside3 (approved):
Storybook looks good. See my two recent comments for non-blocking code clean up.

packages/lib-classifier/src/plugins/tasks/experimental/volumetric/components/VolumetricTask.js
kieftrav: Got this change incorporated as well. Thanks!

PR #6581, opened 2024-12-18
refactor: image loading handlers

eatyourgreens:
Some changes to clean up image loading code, and speed up subject image loading.
- lib-classifier: remove `useEffect` from `useSubjectImage`, so that image loading code runs on image load, rather than waiting for the next React render cycle.
- lib-react-components: when there's no delay for `useProgressiveImage`, run the `onLoad` handler immediately, rather than waiting for the next event cycle.
- refactor `onLoad` event handlers to use `load` events, rather than `Image` objects. This brings the event handlers in line with the DOM standard.

## How to Review

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

## New Feature

## Refactoring

## Maintenance

## Post-merge


packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
```diff
@@ -115,15 +115,15 @@ const SubjectViewer = types
...
+      onSubjectReady (event, frameIndex = 0) {
         const { target = {} } = event || {}
         const {
           clientHeight = 0,
           clientWidth = 0,
           naturalHeight = 0,
           naturalWidth = 0
         } = target || {}
-        self.dimensions.push({ clientHeight, clientWidth, naturalHeight, naturalWidth })
+        self.dimensions[frameIndex] = { clientHeight, clientWidth, naturalHeight, naturalWidth }
```
mcbouslog, line 126: This is a great solution! 😎

mcbouslog (changes requested):
This is great, thank you @eatyourgreens !

The image loading changes work great, tested different size images and across browsers 🚀 .

As noted, this also edits a classification's `metadata.subject_dimensions` similar to PFE. Currently in FEM, a subject using the multi-frame or separate frame and flipbook viewers would create multiple (potentially duplicative) entries in the `metadata.subject_dimensions` array regardless of subject location / frame selected. These changes create one entry in the array per location / frame. 

This creates the scenario with the separate frames / flipbook viewer where historically in FEM and only FEM (not in PFE) if I started in flipbook, went to separate frames, then back to flipbook the different client dimensions would all be included in classification metadata, but now only the last client dimensions viewed per frame will be included in classification metadata, so in the noted scenario there'd be no record of the separate frame client dimensions. I think this is ok and there isn't anyone that'll notice, but noting nonetheless.

Lastly, and the only open item - I don't think an image and text subject (https://local.zooniverse.org:8080/?project=1693&workflow=3430&demo=true) and maybe if a subject has image and JSON (not sure exists) the `metadata.subject_dimensions` is being recorded correctly. If I click Done on the text frame it records the dimensions just for the text location. I think this is because the `useSubjectText` hook is not sending the frame through the to SubjectViewerStore `onSubjectReady`. The `useSubjectJSON` hook might need a similar update.

packages/lib-classifier/src/store/SubjectViewerStore/SubjectViewerStore.js
eatyourgreens: I took it striaght from PFE!

eatyourgreens:
I think Zwicky's Stellar Sleuths had a mixture of image and JSON data. The HR diagram for each subject was included as an image. That subject viewer was an odd one, though: https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/index.html?path=/story/subject-viewers-variablestarviewer--default

eatyourgreens:
Planet Hunters TESS has subjects with one image location (for Talk) and one text location which is actually a string of JSON-encoded light curve data. That should actually be recorded in the classification as a single frame, with the dimensions of the light curve viewer. I haven't changed that here. The default behaviour is `frame = 0`, which should be the same as existing behaviour (an array with only one subjet dimensions object.)

eatyourgreens:
Black Hole Hunters subjects only have one JSON data location, so those shouldn't be affected by the changes here either.

That should be true of light curves from LSST too.

eatyourgreens:
I've added the frame index to the image-and-text viewer. That viewer doesn't record any dimensions for text files, so I've left that unchanged. It should record the dimensions of the image properly now.

eatyourgreens:
> As noted, this also edits a classification's metadata.subject_dimensions similar to PFE. Currently in FEM, a subject using the multi-frame or separate frame and flipbook viewers would create multiple (potentially duplicative) entries in the metadata.subject_dimensions array regardless of subject location / frame selected. These changes create one entry in the array per location / frame.

There’s another subtle bug in the monorepo code: `subject_dimensions` records the order in which images loaded, which may vary from client to client. If an image fails to load, that isn’t recorded at all (this used to happen for roughly 1% of images served from `panoptes-uploads.zooniverse.org`.) PFE records subject dimensions in the same order as the subject locations array, with a gap in the array if frames fail.

mcbouslog (approved):
Text related updates look good, thank you!

mcbouslog:
> There’s another subtle bug in the monorepo code: `subject_dimensions` records the order in which images loaded, which may vary from client to client. If an image fails to load, that isn’t recorded at all (this used to happen for roughly 1% of images served from `panoptes-uploads.zooniverse.org`.) PFE records subject dimensions in the same order as the subject locations array, with a gap in the array if a frame fails.

Interesting! So in FEM the `subject_dimensions` array would be `undefined` / empty at the index that corresponds to the failed frame/location.

eatyourgreens:
>Interesting! So in FEM the subject_dimensions array would be undefined / empty at the index that corresponds to the failed frame/location.

With this PR, yes (same as PFE.) Previously it just wouldn't push anything to the array if an image didn't load.

PR #6604, opened 2025-01-08
fix(lib-classifier): refresh expired bearer tokens

### LINKED ISSUE #6594: Unable to create a new discussion from Quick Talk
https://github.com/zooniverse/front-end-monorepo/blob/51f064ed7d49dd54275113563831ad5b43b14387/packages/lib-classifier/src/components/Classifier/components/QuickTalk/QuickTalkContainer.js#L47

`usePanoptesAuth` here doesn’t refresh expired tokens, leading to permissions errors if you’re logged in, but using a tab that’s been open for two hours or more.

eatyourgreens:
Fix `usePanoptesAuth` hook to refresh expired bearer tokens in the classifier. Fixes a bug where Quick Talk stops working once your bearer token expires.

## Package
- lib-classifier

- fixes #6594.

## How to Review
The original issue (#6594) happens if you leave the classifier open long enough for your bearer token to expire, so you might need to open the classifier in a tab and leave it for a couple of hours. With the changes here, an expired token should always be refreshed, as long as you don't allow your Panoptes refresh token to expire. I think refresh tokens are good for up to 2 weeks.

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/models/tools/Tool/Tool.spec.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/models/tools/Tool/Tool.spec.js
@@ -94,7 +94,7 @@ describe('Model > DrawingTools > Tool', function () {
             taskKey: 'multiple',
             type: 'multiple',
             answers: ['apples', 'oranges', 'pears'],
-            required: '',
+            required: false,
             strings: {
               question: 'which fruit?'
             }
@@ -103,7 +103,7 @@ describe('Model > DrawingTools > Tool', function () {
             taskKey: 'single',
             type: 'single',
             answers: ['one', 'two', 'three'],
-            required: '',
+            required: false,
             strings: {
               question: 'how many?'
             }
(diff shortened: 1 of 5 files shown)


packages/lib-classifier/src/hooks/usePanoptesAuth.js
```diff
@@ -1,17 +1,18 @@
...
     const token = await getBearerToken(authClient)
-    setAuthorization(token)
+    if (token !== authorization) {
+      setAuthorization(token)
+    }
   }
 
-  useEffect(function onUserChange() {
-    checkAuth()
-  }, [authClient, userID])
```
eatyourgreens, line 14: The original bug happens because this hook stores your bearer token in component state and reuses it forever, even after it has expired. It should be checked (by running `checkAuth()`) every time that it's used.

eatyourgreens: General tip for reviewing React components: whenever you see `useEffect` it has probably introduced a subtle bug.

PR #6669, opened 2025-02-06
SelectCollection: make collectionNameFilter() and collectionLabel() more robust

shaunanoordin:
## PR Overview

Package: `app-project`
Affects: `<SelectCollection />`

This PR updates the SelectCollection component's collectionNameFilter() and collectionLabel() functions to be more robust, with respect to unexpected data.

### Dev Notes

This issue was brought in Dependabot PR #6667 , where a seemingly innocent bump from Grommet 2.41.0 to 2.45.1 caused the `CollectionModal` components to suddenly fail. The issue can be traced to SelectCollection's collectionLabel() not receiving its 'collection' data in a proper format - the function's a bit brittle in that sense.

⚠️ HOWEVER, I'm not sure why Grommet 2.45.1 (or its predecessors) suddenly caused the tests to fail. Something's changing the parameters of how `<SelectCollection>` is rendered... or something. I'm going to run some manual Storybook before/after comparisons to make sure nothing's too broken.

### Testing

Automated tests should pass. Uh, there's no manual test for this, really.

### Status

Ready for review.

Low priority 😴 since no new functionality is added, but I'll sound an alarm if it turns out that grommet bump is way more borked than we expect.

Once this is merged, Dependabot PR #6667 (or whatever new version replaces it) should be re-reviewed.

### DIFF
--- a/packages/app-project/src/shared/components/CollectionsModal/components/SelectCollection/SelectCollection.js
+++ b/packages/app-project/src/shared/components/CollectionsModal/components/SelectCollection/SelectCollection.js
@@ -47,6 +47,21 @@ function SelectCollection ({
     return displayNameLowerCase.includes(searchText.toLowerCase())
   }
 
+  /*
+  Determines the display label for the specific collection.
+  Returns "Example Collection" if that collection belongs to the current user,
+  otherwise returns "Example Collection (Another User's Name)"
+
+  NOTE: sometime between Grommet > 2.41.0 and <= 2.45.1, the behaviour of
+  <Select> changed. Now, when <Select> initially renders and _nothing_ is
+  selected, labelKey is called with an empty object {} argument. Prior to this
+  change, collectionLabel() was only called when there's a valid collection
+  object to render. (e.g. when clicking on Select to display list of options.)
+  
+  As a result, collectionLabel() now handles "no selection" (collection === {})
+  without crashing, returning a meaningless "undefined (undefined)". The
+  <Select> still shows blank ("") when there's no selection, though. Go figure.
+   */
   function collectionLabel(collection) {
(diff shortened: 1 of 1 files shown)


shaunanoordin:
Shaun, for your own reference, this is how the Collections Modal should look:

image

shaunanoordin:
Update: local testing show that, if I apply this code patch on top of 6667's dependency bump, the Collection Modal will display fine on Storybook. 👍

goplayoutside3 (approved):
Alrighty through more sleuthing and a brief false alarm about mobx-state-tree, we've narrowed down the effects of Grommet's upgrade to the Select component's `labelKey` handler `collectionLabel()`. In the [diff](https://github.com/grommet/grommet/compare/v2.41.0...v2.45.1) between our current Grommet version and v2.45.0, Grommet made an enhancement to Select so that Select's that use a `labelKey` handler correctly display the selected value with aria-labels. I'll attach a screenshot of the specific diff in this comment for future reference.

The Grommet change affected `collectionLabel()` as the `labelKey` handler because in our collections modal we're intentionally setting the default selected value as blank. Therefore, `collectionLabel()` was passed an empty argument and couldn't find `links.owner` on first render of the modal.

Screenshot 2025-02-11 at 10 31 55 AM

shaunanoordin:
[f41f10a](https://github.com/zooniverse/front-end-monorepo/pull/6669/commits/f41f10ae3be20832e485ddd5afcc9639621a66a5) has been added with notes from our debug session, so future devs know what's going on. Thanks for looking into this, Delilah! 👍

shaunanoordin:
Haha the tests for lib-classifier magically failed after I added a comment into the code. Alright, let's re-run that.

PR #6685, opened 2025-02-13
refactor(drawing-tools): drag handle sizing

eatyourgreens:
- Remove `defaultProps` from functional components.
- Refactor transcription lines to use translation and scale transforms for the line handles.
- Style drag handles and point marks with `non-scaling-size`.
- Fixes a small issue, introduced by #6066, where drag handles aren't rendered in the correct proportion with respect to marks.

Here are screenshots of the sizing issue:

Main branch (drag handles are too small):
A green rectangle drawn on top of a small subject image. The drag handles at the corners of the rectangle are too small, not much bigger than the line stroke width.

This branch (drag handles are correctly proportioned):
A green rectangle drawn on top of a small subject image. The drag handles at the corners of the rectangle are correctly proportioned with respect to the line stroke width.

Test cases in the dev classifier:
- point, line, and rectangle tools on a relatively small image: https://localhost:8080/?project=908&workflow=3370
- transcription lines on a relatively large image: https://localhost:8080/?project=mainehistory/beyond-borders-transcribing-historic-maine-land-documents&env=production

## Package
lib-classifier

# Checklist

## General

## General UX

## Refactoring

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/components/DragHandle/DragHandle.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/components/DragHandle/DragHandle.js
@@ -2,6 +2,7 @@ import { forwardRef } from 'react';
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import draggable from '../draggable'
+import { STROKE_WIDTH } from '../Mark/Mark';
 
 const StyledCircle = styled('circle')`
   stroke-width: 2;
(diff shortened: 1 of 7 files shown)


packages/lib-classifier/src/plugins/drawingTools/components/DragHandle/DragHandle.js
```diff
@@ -2,26 +2,26 @@ import { forwardRef } from 'react';
...
+import { STROKE_WIDTH } from '../Mark/Mark';
 
 const StyledCircle = styled('circle')`
   stroke-width: 2;
   &:hover {
     cursor: move;
   }
 `
-const RADIUS = screen.width > 900 ? 4 : 10
+const RADIUS = screen.width > 900 ? 3 * STROKE_WIDTH : 5 * STROKE_WIDTH
```
eatyourgreens, line 13: I thought it would be safer to make the drag handle size a multiple of the base stroke width here. If the stroke width changes, the default handle size will change proportionately.

eatyourgreens: Yeah, 4px seems a bit small to me, so I bumped it to 6.

eatyourgreens:
By the way, CI is failing because coveralls.io had a major outage yesterday and they're still recovering. https://status.coveralls.io/ says use this workaround to avoid blocking CI pipelines:
>`fail-on-error: false` if using Coveralls GitHub Action.

eatyourgreens:
>Per your suggestion I had been trying non-scaling-size, but was missing the related transform that makes it work.

The spec says that `non-scaling-size` ignores any scaling transforms applied from the element's parent, so I'm confused that I had to scale circles down to the correct radius for a given image size.

PR #6689, opened 2025-02-13
refactor(drawing-tools): revert non-scaling-size

eatyourgreens:
[`vector-effect: non-scaling-size`](https://developer.mozilla.org/en-US/docs/Web/CSS/vector-effect) isn't implemented in browsers, so this reverts it back to `non-scaling-stroke` in drag handles, point marks, and transcription lines.

Test cases in the dev classifier:
- point, line, and rectangle tools on a relatively small image: https://localhost:8080/?project=908&workflow=3370
- transcription lines on a relatively large image: https://localhost:8080/?project=mainehistory/beyond-borders-transcribing-historic-maine-land-documents&env=production

This PR shouldn't change how drawn marks are rendered at all. Points and drag handles should remain the same size and width, regardless of image zoom level.

Here's a quick codepen which draws two identical SVG circles, one with 'non-scaling-stroke' and one with 'non-scaling-size'. Stroke scaling is only ignored in the first. The SVG spec says that it should be ignored in both.
https://codepen.io/eatyourgreens/pen/ZYEYqEv

If you resize the SVG image, by resizing the browser window, you'll see that the righthand circle gets thicker or thinner, but the lefthand circle remains the same.

## Package
- lib-classifier

# Checklist

## General

## General UX

## Refactoring

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/components/DragHandle/DragHandle.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/components/DragHandle/DragHandle.js
@@ -34,8 +34,8 @@ const DragHandle = forwardRef(function DragHandle(
 
   return (
     <g ref={ref} transform={transform} data-testid={testid} >
-      <StyledCircle r={radius} {...styleProps} vectorEffect={'non-scaling-size'} />
-      <StyledCircle r={2 * radius} fill='transparent' stroke='transparent' vectorEffect={'non-scaling-size'} />
+      <StyledCircle r={radius} {...styleProps} vectorEffect={'non-scaling-stroke'} />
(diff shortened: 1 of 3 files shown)


eatyourgreens:
MDN: [`vector-effect` is widely supported in all browsers since 2020.](https://developer.mozilla.org/en-US/docs/Web/CSS/vector-effect)

Can I Use: [`non-scaling-stroke` is widely supported](https://caniuse.com/vector-effect). "Other values for the vector-effect attribute/property are currently at risk of being removed from the specification as they are not being developed by browser vendors."

😠

mcbouslog (approved):
LGTM, thanks for quick edit 🚀 !

mcbouslog:
Iteration on #6685

PR #6707, opened 2025-02-22
fix(a11y): disabled task popup keyboard accessibility

### LINKED ISSUE #2300: 'Choose another subject' is inaccessible with a screenreader, or with keyboard navigation
## Package
lib-classifier

## Describe the bug
The announcement that a subject is finished, and the options to choose another subject, are hidden away at the bottom of the page (after the footer.)

Here you can see VoiceOver reading out the task instruction when I enter the task area, although a sighted user gets a message that the subject is retired and not available.
The task area in VoiceOver. The visible text says that the subject is finished, but VoiceOver is reading out the task instruction.

## To Reproduce
1. Go to https://frontend.preview.zooniverse.org/projects/bogden/scarlets-and-blues/classify/workflow/17076/subject-set/96235/subject/63413209?finished=true
2. Read through the page in VoiceOver (or the screenreader of your choice.)
3. Alternatively, tab through the page to the task area.

## Expected behavior
When I enter the task area In a screenreader, I should hear "This subject has enough classifications, so any additional classifications submitted will not be used in its analysis. You can still view this subject for reference." and then options to choose a new subject.

When I tab through the page, I should tab into that popup when I reach the task area.

## Additional context
Adding content to the DOM at the end of the document is a common method for implementing modals. It kind of breaks for popups that aren't modal, because the rest of the page is still accessible via the keyboard. Ideally, the content of a targeted layer like this one would be inserted into the DOM at the start of the target element.

This same issue affects the content of the field guide, which should be accessible while also continuing to classify.

eatyourgreens:
Remove `createPortal` from the `DisabledTaskPopup` component. Render it inline in the task area, wrapped in a styled `dialog` element.

Use the `inert` DOM property to disable the task area when the task area is disabled and the subject is finished.

Tested with Chrome, Firefox and Safari/VoiceOver on MacOS 14.7.3, but it should also work with NVDA and JAWS on Windows. 

## Package
- lib-classifier

- closes #2300 (which has been open for almost 4 years!)

## How to Review
Navigate the page with the keyboard (Chrome/Firefox) or with VoiceOver (Safari.) The disabled task popup should now be the only content in the task area, when the subject is finished (and the workflow uses indexed subjects.) When you dismiss the popup, focus should continue from the top of the tabbed task area.

Keyboard navigation in Chrome:

https://github.com/user-attachments/assets/efaa5649-6369-4cb4-af05-2453eb2d1b50

VoiceOver/Safari:

https://github.com/user-attachments/assets/6b2cc96d-40b2-4111-bfc4-6bdb18489174

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed


goplayoutside3 (approved):
Works as the videos demonstrate, thank you!

Just a quick double check - when I run this PR locally and visit a finished subject url (for example [PRINT](https://local.zooniverse.org:3000/projects/printmigrationnetwork/print/classify/workflow/20199/subject-set/123000/subject/102039006?env=production&demo=true)), if I'm focused on the last button in the DisabledTaskPopup labeled 'dismiss', tabbing again takes me to the first Tab in the task area tablist even though that tablist is not clickable. I'm not against this UX, but wanted to check it's the intended behavior.

eatyourgreens:
That might be a bug. The tabs and the tutorial aren’t disabled when the dialog is open, either here or in production. Only the task tab is inert.

eatyourgreens:
I think the whole component should be inert when the dialog is open, otherwise you can still use the tabs, even though they’re visually hidden.

eatyourgreens:
@goplayoutside3 that should be fixed now. I've made the whole task area inert when the popup is open.

eatyourgreens:
Spotted a bug in dark mode, so I'll push a fix for that too.

The popup in dark mode, showing a large white border around the dialog. The background should be transparent, not white.

goplayoutside3:
Thanks! Agreed the use of "with" wrappers is a blocker in some components of FEM. I hope to continue slowly refactoring them in favor of hooks instead.

PR #6753, opened 2025-03-11
lib-user: Refactor group stats data export

### LINKED ISSUE #6363: User group CSV export ("Export all stats") doesn't work with a group of over 800 members
## Package
- lib-user

## Describe the bug
The CSV export ("Export all stats") doesn't work with a group of over 800 members. The related [request for all users in a group](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-user/src/components/Contributors/Contributors.js#L104-L109) exceeds nginx limits, as it includes the user IDs for all group members.

Note user group `2533715` as an example.

## Expected behavior
The CSV export ("Export all stats") should process as expected with all groups.

mcbouslog:
## Package
- lib-user

- closes #6363 

## Describe your changes
- refactor `handleGenerateExport` and `getAllUsers` to separate functions
- refactor `getAllUsers` with `CHUNK_SIZE` of 200 to batch requests of users to a character size that's broadly compatible (see Character Size Note)
- create ExportStats component to provide confirmation, loading, error, and success states UIs
- add progress indicator
- add file size estimate
- [x] design review and edits

Character Size Note
- previously groups with more than ~200 members exceeded request character size limitations (~2,084 per Chrome, ~4,000 - ~8,000 per nginx, depending on settings, to name a few)
- the request exceeded character size limitations because the user ids were included in the get request as a url param, i.e.  `https://www.zooniverse.org/api/users?page=1&id=**123456**%2C**789123**%2C**234567**%2C**891234**` (user ids surrounded by **'s)
- with groups of over roughly 200, this returns a 406 error

UX Changes Note
- with a very large group, like https://www.zooniverse.org/groups/2533715/contributors , after clicking the export stats link, it will take a long time to process the related requests and the CSV file may be as large as 24MB(!!)
- because many `/users` requests take a long time and the file size might be quite large...
- the confirm step (file size) and progress (many requests) component states are included in these changes
- the final success component state provides a link to click instead of auto-downloading

## How to Review
  - ⚠️ very large group ⚠️ (requires admin): https://local.zooniverse.org:3000/groups/2533715/contributors?env=production
  - medium group: [insert medium group link here]
  - small group: https://local.zooniverse.org:3000/groups/2620016/contributors?env=production
  - click the "Export all stats" link in the upper right of the group/contributors page
  - a few new ones for the different ExportStats states, starting with http://localhost:6008/?path=/story/components-contributors-exportstats--confirm-large-export

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

## Refactoring

## Post-merge


packages/lib-user/src/components/Contributors/helpers/getAllUsers.js
```diff
@@ -0,0 +1,55 @@
...
+          id: chunk.join(','),
+          page_size: 100 // per panoptes, maximum page_size is 100
+        }
+
+        const wrappedPromise = fetchPanoptesUsers(chunkQuery).then(result => {
+          setExportProgress(prevProgress => prevProgress + ((1 / totalChunks) * 100))
+          return result
+        })
+
+        userPromises.push(wrappedPromise)
```
mcbouslog, line 33: The wrapped promises are essentially the progress indicator. This is not exact. On very large groups the progress hangs around 90-98% as the very last few `/users` requests take awhile to return, and I'm not sure why. 

This makes me wonder - is a roughly accurate but not exact progress indicator that sometimes appears to but doesn't actually get stuck towards completion better or worse than a simple loading indicator? For most groups a progress status isn't necessary, but for the very large groups it feels like something is necessary.

packages/lib-user/src/components/Contributors/helpers/handleGenerateExport.js
```diff
@@ -0,0 +1,66 @@
...
+  setErrorMessage,
+  stats
+}) {
+  setExportStatus(asyncStates.loading)
+  setExportProgress(0)
+
+  try {
+    const allUsers = await getAllUsers({ memberIdsPerStats, setExportProgress })
+
+    const { filename, dataExportUrl } = await generateExport({
```
mcbouslog, line 23: I'm not sure without doing some research, but I presume there's way to calculate the actual size of the dataExport once it's generated. If so, the actual size could be added to the success state component.

packages/lib-user/src/components/Contributors/helpers/getAllUsers.js
goplayoutside3: I checked out this branch locally, and confirmed I can generated an export for the large group and the small group. In the large group, I can replicate the loading state freeze at 98%, and I noticed that the loading animation also freezes at that point.

What would cause the `<Loader />` animation to stop? Is that a point in the export process that can be hooked into to display an extra message like "Finishing up..."? Because the animation also froze, I could easily assume the export failed.

packages/lib-user/src/components/Contributors/helpers/handleGenerateExport.js
```diff
@@ -0,0 +1,102 @@
+import asyncStates from '@zooniverse/async-states'
+import { arrayOf, func, number, shape, string } from 'prop-types'
+
+import { getAllUsers } from './getAllUsers'
+
+async function generateExportWithWorker({ group, projects, stats, users }) {
+  return new Promise((resolve, reject) => {
+    const worker = new Worker(new URL('./generateExport.worker.js', import.meta.url))
```
mcbouslog, line 8: Creating the CSV was causing the UI to appear stuck at 98%. All the API requests had resolved, and the CSV generation started, which would cause the main browser JavaScript thread to hang. The only solution I could find was to put the CSV content creation on a separate thread with a [Web Worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API). I'm not super familiar with Web Workers, largely followed the examples linked in the MDN docs.

packages/lib-user/src/components/Contributors/components/ExportStats/ExportStats.js
```diff
@@ -0,0 +1,216 @@
...
+      <Layer>
+        <ExportBox>
+          <Text
+            textAlign='center'
+          >
+            {t('Contributors.ExportStats.generating')}
+          </Text>
+          <Button
+            label={t('Contributors.ExportStats.close')}
+            onClick={onClose}
```
mcbouslog, line 119: Since the browser might be hanging at this stage while the CSV content is created, maybe the Close button should be `disabled`? I lean towards the button _not_ be disabled, but open to suggestion.
```suggestion
            disabled
            onClick={onClose}
```

mcbouslog:
@seanmiller26 - apologies for the delay, here are some screenshots using the most basic/default styled elements:

contributors_export
ExportStats_confirm_large_export
ExportStats_confirm_small_export
ExportStats_loading
ExportStats_error
ExportStats_generating_file
ExportStats_success_download

packages/lib-user/src/components/Contributors/components/ExportStats/ExportStats.js
goplayoutside3: Will the `onClose` function cancel the csv generation?

I don't think the button should be disabled, but the `Contributors.ExportStats.generating` could offer guidance such as "Generating stats export file. Do not close this browser tab".

goplayoutside3 (approved):
Looks great, and I'm able to download the largest group export with the expected UI messages. Also appreciate the in-code comments for easy understanding!

Only NB suggestion would be to add guidance to the "generating" message mentioned below.

seanmiller26 (approved):
Including a status indication is a major improvement to what we have currently. I do agree with Delilah's comment about guidance so that users don't close the modal prematurely. Besides that, looks Good!

PR #6769, opened 2025-03-18
VV: fix bug that limits the max number of marks.

kieftrav:
## Package
- lib-subject-viewers

## Describe your changes
Fix color looping bug wherein creating a new mark with an annotationIndex greater than the number of colors generated should loop to the 1st color instead of erroring.

## How to Review
Load up the `lib-subject-viewers` storybook on production and this branch. Use "shift+click" to create at least 12 different marks on the 4x4x4 Story (any story works, this is the simplest example). This will fail in production and start re-looping through the colors. 

## General

## Bug Fix
- [ ] The bug is fixed


goplayoutside3 (approved):
Looks good!

PR #6773, opened 2025-03-19
Edit GitHub Actions workflow ymls to node v20.18 with .nvmrc

mcbouslog:
## Package
- GitHub Actions workflow ymls - ci-build, coverage, and deploy-storybook

## Linked Issue and/or Talk Post
- tests (that rely on cjs) fail with node >=20.19
- related node documentation: https://nodejs.org/en/blog/release/v20.19.0

## Describe your changes
- add `.nvmrc` specifying node v20.18
- update workflow ymls to reference `.nvmrc`
- update `package.json` with applicable node range
- update each package's `package.json` node range

## How to Review
_Helpful explanations that will make your reviewer happy:_
- What Zooniverse project should my reviewer use to review UX?
- What user actions should my reviewer step through to review this PR?
- Which storybook stories should be reviewed?
- Are there plans for follow up PR’s to further fix this bug or develop this feature?

# Checklist
_PR Creator - Please cater the checklist to fit the review needed for your code changes._
_PR Reviewer - Use the checklist during your review. Each point should be checkmarked or discussed before PR approval._

## General
- [ ] Tests are passing locally and on Github
- [ ] Documentation is up to date and changelog has been updated if appropriate
- [ ] You can `yarn panic && yarn bootstrap` or `docker-compose up --build` and FEM works as expected
- [ ] FEM works in all major desktop browsers: Firefox, Chrome, Edge, Safari (Use Browserstack account as needed)
- [ ] FEM works in a mobile browser

PR #6777, opened 2025-03-19
Node 20.19+ compatibility: mark some packages as CommonJS

shaunanoordin:
## PR Overview

This PR should hopefully fix the issue where our FEM tests are failing, by _explicitly_ marking packages as being _CommonJS_ packages (instead of the _new ESM default_ )

Context of the problem:
- We're using Node lts/iron to run our GitHub actions.[^1]
- lts/iron basically means Node 20.whatever [(link)](https://nodejs.org/en/blog/release/v20.9.0)
- _Recently,_[^2] that 20.x specifically meant Node 20.19...
- ...and Node 20.19 has a breaking change that made package ESM-types be default. [(link)](https://nodejs.org/en/blog/release/v20.19.0) (i.e. if "type" was missing in package.json, assume it's "type": "module")
- A lot of our packages didn't specify a type, and ran with the common assumption that it's a CommonJS package.
- As a result, a lot of our automated tests on GitHub went kaput. On local, we may not have noticed this, as we may have had different versions of node running (e.g. I'm running 20.10)

The solution:
- For each package that had its test crashed out, I can see that their package.json was missing an explicit "type" value.
- Solution was to add `"type": "commonjs"` to reinstate the _previous default behaviour_ of "treat this package as CommonJS".
- Credit to Jim for this solution. [(link)](https://github.com/zooniverse/front-end-monorepo/issues/3779#issuecomment-2737370217)

[^1]: Check our [/.github/workflow folder](https://github.com/zooniverse/front-end-monorepo/tree/42b42243a4875926ea6f450a50c8e87ac92a4a26/.github/workflows) various .yml files. For example, our [testing coverage file.](https://github.com/zooniverse/front-end-monorepo/blob/42b42243a4875926ea6f450a50c8e87ac92a4a26/.github/workflows/coverage.yml#L21)
[^2]: March 13 earliest, since that's when 20.19 was released. I'm writing this on 19 Mar.

Notes:
- Not all packages were updated with type=commonJS.
  - app-root explicitly has type=module
  - lib-grommet-theme has no explicit type, but doesn't run any tests, so this doesn't matter.
  - I will just shrug if you ask me about lib-async-states or lib-panoptes-js. 🤷 ~~These have tests and don't have explicit type=commonjs, yet aren't breaking.~~ Oh wait, they don't have **Storybook incorporated into their tests.**  That's the final ingredient in causing the breaking change!
- Longer-long term solution would be to make the packages actually ESM-compatible, but that's a story for another day.

### Status

Ready for review.

Delilah/Mark/Travis, if everything looks good, please go ahead and merge this. We'd also need to revert the short term fix in #6773 , I think.

### DIFF
--- a/packages/app-project/package.json
+++ b/packages/app-project/package.json
@@ -5,6 +5,7 @@
   "author": "Zooniverse <contact@zooniverse.org> (https://www.zooniverse.org/)",
   "version": "0.0.1",
   "main": "index.js",
+  "type": "commonjs",
   "scripts": {
     "build": "APP_ENV=${APP_ENV:-development} NODE_ENV=${NODE_ENV:-production} PANOPTES_ENV=${PANOPTES_ENV:-production} next build",
     "dev": "APP_ENV=${APP_ENV:-development} PANOPTES_ENV=${PANOPTES_ENV:-staging} node server/server.js",
--- a/packages/lib-classifier/package.json
+++ b/packages/lib-classifier/package.json
@@ -6,6 +6,7 @@
   "version": "0.0.1",
   "main": "dist/cjs/components/Classifier/index.js",
   "module": "dist/esm/components/Classifier/index.js",
+  "type": "commonjs",
   "repository": {
     "type": "git",
     "url": "https://github.com/zooniverse/front-end-monorepo.git",
--- a/packages/lib-content/package.json
+++ b/packages/lib-content/package.json
@@ -6,6 +6,7 @@
   "version": "0.0.1",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
+  "type": "commonjs",
   "exports": {
     ".": {
       "import": "./dist/esm/index.js",
--- a/packages/lib-react-components/package.json
+++ b/packages/lib-react-components/package.json
@@ -6,6 +6,7 @@
   "version": "1.13.0",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
+  "type": "commonjs",
   "exports": {
     ".": {
       "import": "./dist/esm/index.js",
--- a/packages/lib-subject-viewers/package.json
+++ b/packages/lib-subject-viewers/package.json
@@ -6,6 +6,7 @@
   "version": "0.1.0",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
+  "type": "commonjs",
   "exports": {
     ".": {
       "import": "./dist/esm/index.js",
--- a/packages/lib-user/package.json
+++ b/packages/lib-user/package.json
@@ -6,6 +6,7 @@
   "version": "0.1.0",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
+  "type": "commonjs",
   "exports": {
     ".": {
       "import": "./dist/esm/index.js",


shaunanoordin:
Whoah, what the heck? All the builds are failing?

image

shaunanoordin:
### Status Update

As of 18:21 GMT, I can see that while the tests are passing, the BUILDS are failing. Ironically, I should have tested for that. I'm removing this PR from reviewability.

eatyourgreens:
Other packages, like PanoptesJS, are pure CommonJS.

eatyourgreens:
A few years ago, I looked at Babel’s build scripts, which output two `package.json` files, one for the ESM build and one for CJS. I don’t know if they still do that, but it might be worth looking at.

shaunanoordin:
Trying to force our packages to use CommonJS, even if it's for a short term fix, is a dead end. This PR may be headed for the scrap pile very soon. However, I think I did find something else of interest.

## Investigation Update

Current hypothesis: **mocha can't handle JSX syntax,** _when run in an ESM package._

Investigation method:
- Isolate the minimum required setup to replicate the _Unexpected token '<'_ error we get when running yarn test.
  - So: 1. select the most independent package. 2. Narrow down test to only a single simple component with a simple .spec.js file. 3. dummy out code even further until the source of error is found.
- I selected `lib-react-components` and the `<AdminCheckbox>` component.
- package.json was modified so running `yarn test` in lib-react-component ONLY tests AdminCheckbox.
    <details>
      <summary>(updates to package.json)</summary>

      Added `"type": "module"`

      "test" script changed from
      `"test": "mocha --config ./test/.mocharc.json ./.storybook/specConfig.js \"./src/**.*.js\""`
      to
      `"test": "mocha --config ./test/.mocharc.json ./.storybook/specConfig.js \"./src/AdminCheckbox/AdminCheckbox.spec.js\""`
    </details>

Investigation results:
- _yarn test_ results in the same error. Good.
- AdminCheckbox.spec.js replaced with a very very simply dummy code that just goes `expect(true).equals(true)`
    <details><summary>(new AdminCheckbox.spec.js)</summary>

      ```
      // No imports, no nothing. Just this.
      describe('<AdminCheckbox />', function () {
        it('should render the label', function () {
          expect(true).equals(true)
        })
      })
      ```
    </details>
- _yarn test_ results in the same error. Good, now we can isolate AdminCheckbox.js. Problem must be in the mocha setup, or something.
- Long story short, I followed `.storybook/specConfig.js` to `.storybook/preview.js`, and proceed to slowly dummy out portions of the code until _yarn test_ successfully runs. The problem was determined to exist here:

```
// .storybook/preview.js

const decorators = [
  (Story, context) => {
    return (
      <StrictMode>  // HERE
        <Grommet
          background={background}
          theme={zooTheme}
          themeMode={context.globals.theme}
          full
        >
          <Story />
        </Grommet>
      </StrictMode>
    )
  }
]
```

- If I change that return to a simple `<div>`, _yarn test_ will still error out.
    <details><summary>(Example)</summary>

      ```
      const decorators = [
        (Story, context) => {
          return (
            <div>hello</div>  // Still causes errors
          )
        }
      ]
      ```
    </details>
- But if I change that return to null, _yarn test_ WILL work.
    <details><summary>(Example)</summary>

      ```
      const decorators = [
        (Story, context) => {
          return (
            null  // Suddenly, everything's OK.
          )
        }
      ]
      ```
    </details>
- This leads me to the hypothesis that `mocha` _doesn't know how to handle JSX, when the package is set to the ESM(module)-type._

Additional Testing To Strengthen Hypothesis:
- With preview.js's sole JSX syntax dummied out, I proceeded to un-dummy AdminCheckbox.spec.js
- True enough, any time AdminCheckbox.spec.js or AdminCheckbox.stories.js contains JSX, the _yarn test_ will fail.

### Follow Up Thoughts

🤔 In retrospect, it seems kinda obvious that the JSX syntax would be the problem. JSX isn't a native part of JS, and ESM is all about that native JS goodness.

- I suppose the next point of attack is finding out if there's something that can 
  - So far, using [@node-loader/babel](https://gist.github.com/smhg/9e75eee75818e09f99011e565340f45f) hasn't helped, hmm.
- I _thought_ lib-react-classifier's `@babel/register` (see lrc's .mocharc.json) would have handled the Babel transpiling to make JSX work, but apparently not?? (As far as I can tell, this was added when lib-react-components was "supposed" to be a CommonJS thing, and @babel/register is used to make CommonJS files able to understand the meaning of `import`.)

Complaints to and for myself:
- Shaun, how did you forget that you yourself struggled with not having JSX syntax when you made your own minimal web apps for hackday? Also the "Can't figure out the < blip blop" error should have been a dead giveaway.
- But to be fair to my Shaun-self, that error was reported in specConfig.js, which is the wrong place for the error.

eatyourgreens:
> Current hypothesis: mocha can't handle JSX syntax, when run in an ESM package.

There’s an open issue for this (JSX not being transpiled in tests): #3779. 

The problem, IIRC, is `babel/register`. See https://github.com/zooniverse/front-end-monorepo/issues/3779#issuecomment-2523821501.

eatyourgreens:
One problem with the monorepo is that both JS and JSX files use the `.js` extension whereas if you look at PFE, say, JSX files use `.jsx` and can be handled differently.

eatyourgreens:
This comment talks about the new module loader in Node 22, which has now been turned on in 20.19. 

https://github.com/zooniverse/front-end-monorepo/issues/3779#issuecomment-2523944605

eatyourgreens:
This excellent summary by @goplayoutside3 should be linked here too:
https://github.com/zooniverse/front-end-monorepo/issues/3779#issuecomment-2523753447

eatyourgreens:
I've tried removing the old CJS builds in #6780, but that still errors in the Mocha setup files (which are very old.)

>Exception during run: ReferenceError: require is not defined in ES module scope, you can use import instead
      at file:///home/runner/work/front-end-monorepo/front-end-monorepo/packages/lib-subject-viewers/test/setup.js:3:13

shaunanoordin:
No additional work is being done on this branch, so I'll be closing this for now. We'll revisit this if necessary.

PR #6796, opened 2025-03-26
fix(a11y): keyboard focus for dialogs

### LINKED ISSUE #6790: A11y: Esc and Tab have stopped working to navigate popup dialogs
## Package
- app-project
- lib-classifier

## Describe the bug
The Esc key no longer closes the Subject Info or Add To Collections popups. The Field Guide still closes on Esc.

When the field guide is open, Tab and Shift-Tab navigate the task area, so the content of the field guide is inaccessible from the keyboard. 

UPDATE: the underlying bug is that focus no longer shifts to an open dialog, after it opens. It stays on the triggering element (usually a button.) Tab works to navigate into modal popups, because everything else on the page is inert after they open. Tab doesn't work for the field guide, because it isn't modal (you can still annotate and classify while the field guide is open.)

## To Reproduce
https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/classify/workflow/11235

1. Click Field Guide.
1. Press Esc to close the popup.
1. Open the Field Guide and try to tab into it. Focus moves to the Task tab, not the open field guide.
1. Click Subject Info, Add To Collections, or Need Some Help With This Task.
1. Esc to close those popups doesn't work, unless you first press Tab to enter the dialog. 

## Expected behavior
When a dialog opens, focus should shift to the dialog. Esc should close the dialog and return focus to where it was before the dialog opened. Tab and Shift-Tab should navigate focusable content inside the open dialog. If the dialog is modal, the rest of the page should be inert while the dialog is open. If the dialog isn’t modal (eg. finished subjects or the field guide) tabbing out of the dialog should continue navigating the underlying page from the point at which you entered the dialog. 

## Device information
Latest Firefox and Chrome on MacOS 14.7.3.

## Additional context
The ARIA dialog pattern includes a list of expected keyboard interactions for modal dialogs.
https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/

Grommet’s `FocusedContainer` component should grab focus when a modal dialog opens. 
https://github.com/grommet/grommet/blob/master/src/js/components/FocusedContainer.js

Talk comment here:
https://www.zooniverse.org/talk/2354/3326062?comment=5945075&page=12

eatyourgreens:
A quick patch for broken keyboard focus in the `Modal` component. Manually focus the component root when it mounts, if the parent component has passed `trapFocus=true` to the `Modal` component.

Also, add the `dialog` role along with an accessible name for the dialog (using `aria-label`.)

Screenshot of the open field guide popup in Firefox dev tools, showing that it has the dialog role and the name 'Field Guide.'

## Package
- lib-react-components

- closes #6790.

## How to Review
Classifier dialogs should have keyboard focus now, when they open. In the accessibility tree (and in screenreaders), they should be announced as dialogs, with the component's title as the dialog name.

Since dialogs are now focused after opening, Esc should close them and Tab/Shift-Tab should navigate content inside an open dialog.

This staging workflow has some drawing tools with text and question tasks, so that you can test subtasks from the keyboard.
https://localhost:8080/?env=staging&project=eatyourgreens/-project-testing-ground&workflow=3370

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed


packages/lib-classifier/src/components/Classifier/components/TaskArea/components/DisabledTaskPopup/DisabledTaskPopup.js
```diff
@@ -61,6 +61,8 @@ export default function DisabledTaskPopup({
   return (
     <StyledDialog open={active} aria-label={t('TaskArea.DisabledTaskPopup.title')}>
       <Modal
+        role={undefined}
+        aria-label={undefined}
```
eatyourgreens, line 65: `Modal` here is being used to render content for a `<dialog>` element, so we have to unset its default dialog role and accName.

eatyourgreens:
I'm 90% sure that Grommet Layer used to handle focus automatically, so this PR might be patching around something that has changed recently in Grommet.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/DisabledTaskPopup/DisabledTaskPopup.js
eatyourgreens: It might be better to fix this in the `withLayer` helper, so that component authors are free to use the undecorated `Modal` component inside a native `<dialog>` if they wish.

eatyourgreens:
Popups are now announced as dialogs in VoiceOver, with a title (if provided.)

Subject metadata in VoiceOver, showing that the popup is annpunced as a dialog.

packages/lib-react-components/src/helpers/withLayer/withLayer.js
```diff
@@ -22,6 +23,8 @@ function withLayer (WrappedComponent) {
 
     return (
       <Layer
+        role='dialog'
+        aria-label={title}
```
eatyourgreens, line 27: This `Layer` can probably be replaced with a HTML `<dialog>`, now that dialogs are supported in all browsers.
https://caniuse.com/dialog

eatyourgreens:
>I'm 90% sure that Grommet Layer used to handle focus automatically, so this PR might be patching around something that has changed recently in Grommet.

I opened an issue on Grommet for this, though the changes here will fix the bug for Zooniverse React Components.

goplayoutside3 (approved):
Looks good, thank you!

PR #6799, opened 2025-03-26
Tutorial: fix navigation breaking when too many steps

### LINKED ISSUE #6798: Tutorial: navigation borks when there are too many steps
## UI Bug

Package: lib-classifier
Component: Tutorial (modal and non-modal versions)

When a Tutorial has too many steps, the "step navigation" (the dots indicating pages/steps in the tutorial) borks the UI.

_Screenshot: the "dots" in the "step navigation" (bottom of pic) trails off-screen to the right._

Image

### Dev Notes

- This isn't just a visual presentation issue; when the "step navigation" is too long, it actually _blocks_ the "Next tute step/page" button.
- There "threshold" of the number of pages/steps to trigger this issue _changes depending on screen width,_ so saying "keep tutorials under X steps/pages" won't always work. 🤔 
- The problem is in the `StepNavigation` component [(link)](https://github.com/zooniverse/front-end-monorepo/blob/208b9d5/packages/lib-classifier/src/components/Classifier/components/shared/StepNavigation/StepNavigation.js), in the CSS. Basically, the content is incorrectly overflowing the container, and causing an overlap of UI items.
    Image

### Non-Dev Notes

There's actually a valid reason for limiting the number of Tutorial steps, i.e. we can argue that projects shouldn't have too many Tutorial steps to begin with.

However, that's a separate discussion - at the moment, we have a UI failure that needs patching.

### Possible Solution

The most rudimentary solution is to add `overflow: auto` to StepNavigation to prevent the overlapping/overflowing UI issue.

Image

However, this does hide the dots behind a scrollable area, so once you get far enough into the tute, you won't see your "current page" unless you know that you need to scroll.

However _however,_ this rudimentary solution will allow the "Next Page" button to work at least, while we think of a better design solution.

### Status

Low priority. Solution proposed. (See #6799)

shaunanoordin:
## PR Overview

Package: lib-classifier
Closes #6798 

This PR provides a _rudimentary fix_ for the Tutorial system's StepNavigation, to prevent the Tutorial UI from completely breaking when a tutorial has too many steps.

Before this patch:

Image

After this patch:

Image

<details>
<summary>(Previous solution:)</summary>

Image

This solution was replaced [after this comment.](https://github.com/zooniverse/front-end-monorepo/pull/6799#issuecomment-2758633484)
</details>

### Status

Ready for review.

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/shared/StepNavigation/StepNavigation.js
+++ b/packages/lib-classifier/src/components/Classifier/components/shared/StepNavigation/StepNavigation.js
@@ -18,6 +18,7 @@ const StyledButton = styled(Button)`
 
 const StyledRadioButtonGroup = styled(RadioButtonGroup)`
   position: relative;
+  overflow: auto;
 
   > label {
     > span {
--- a/packages/lib-classifier/src/components/Classifier/components/shared/StepNavigation/StepNavigation.js
+++ b/packages/lib-classifier/src/components/Classifier/components/shared/StepNavigation/StepNavigation.js
@@ -16,9 +16,11 @@ const StyledButton = styled(Button)`
   }
 `
 
+// flex-wrap + gap are used to ensure if the Tutorial has too many steps, the 
 const StyledRadioButtonGroup = styled(RadioButtonGroup)`
   position: relative;
-  overflow: auto;
+  flex-wrap: wrap;
+  gap: 20px 0;
 
   > label {
     > span {


shaunanoordin:
## PR Update

Better idea from Sean: wrap excessive page buttons into a second(+) row(s).

Image

goplayoutside3 (approved):
Looks good! I used this project with overflow of tutorial steps to test: https://local.zooniverse.org:3000/projects/leinwc/island-critter-cam/classify/workflow/26051?env=production&demo=true

PR #6801, opened 2025-03-26
Implement ARIA recommendation for the Accordion pattern for the VV Planes view

kieftrav:
## Package
- lib-subject-viewers

## Describe your changes
Implement the [ARIA Accordion Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/) for the Plane View of the Volumetric Viewer

## How to Review
- Running Storybook locally with the Volumetric Viewer, click anywhere on the header to expand/collapse the planar view. - Using the keyboard, test that the three required keyboard interactions work as expected from the ARIA Accordion Pattern page (enter, space, tab, shift+tab). 
- The whole header should be highlighted when its element is focused through keyboard navigation. 

## General

### DIFF
--- a/packages/lib-subject-viewers/src/VolumetricViewer/components/Plane.js
+++ b/packages/lib-subject-viewers/src/VolumetricViewer/components/Plane.js
@@ -234,7 +234,7 @@ export const Plane = ({
     <StyledBox className={`plane-container plane-container-${dimension} ${expanded ? 'expanded' : 'collapsed'} no-select`} ref={containerRef}>
       <button
         aria-controls={`section-${planeId}`}
-        aria-expanded={expanded}
+        aria-expanded={expanded.toString()}
         aria-label={`Toggle ${viewer.dimensions[dimension]} Plane Visibility`}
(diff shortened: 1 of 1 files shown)


packages/lib-subject-viewers/src/VolumetricViewer/components/Plane.js
```diff
@@ -227,18 +232,31 @@ export const Plane = ({
 
   return (
     <StyledBox className={`plane-container plane-container-${dimension} ${expanded ? 'expanded' : 'collapsed'} no-select`} ref={containerRef}>
-      <Box className={`plane-title ${expanded ? 'expanded' : 'collapsed'}`}>
+      <button
+        aria-controls={`section-${planeId}`}
+        aria-expanded={expanded}
```
eatyourgreens, line 237:
```suggestion
        aria-expanded={expanded.toString()}
```

ARIA states are strings eg. `'true'|'false'`. I think that trips a lot of people up.

goplayoutside3 (approved):
Looks good!

PR #6803, opened 2025-03-26
Fix bug where the VV planar view wouldn't show an annotation if the histogram had removed its underlying color's voxel

kieftrav:
## Package
- lib-subject-viewers

## Describe your changes
When the Histogram controls are narrowed, all marks should remain visible regardless of if the underlying color has been removed from the histogram in the Plane view

## How to Review
Load up the Volumetric Viewer in the `lib-subject-viewer` Storybook. Annotate any voxel and then use the sliders for the histogram to hide voxels that are too dark / too light. It's possible to slide one slider all the way over to the other slider, making only marks visible (good for testing purposes). Before this change, marks would disappear in the Plane view.

## General

### DIFF
--- a/packages/lib-subject-viewers/src/VolumetricViewer/components/Plane.js
+++ b/packages/lib-subject-viewers/src/VolumetricViewer/components/Plane.js
@@ -169,10 +169,10 @@ export const Plane = ({
   }
 
   function drawPoint ({ context, point, x, y }) {
-    // Draw points that are not in threshold same color as background
-    if (viewer.isPointInThreshold({ point })) {
-      const annotationIndex = viewer.getPointAnnotationIndex({ point })
+    const annotationIndex = viewer.getPointAnnotationIndex({ point })
 
+    // Draw points that are not in threshold same color as background
+    if (viewer.isPointInThreshold({ point }) || annotationIndex !== -1) {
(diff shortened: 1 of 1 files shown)


goplayoutside3 (approved):
Looks good! One behavior to double check is -- do you intend for volunteers to use the historgram to hide all data except their annotations? Here's a screen recording where I made several annotations, but there are some light-colored patches still un-marked. Those un-marked patches still display at the far range of the histogram slider, so I'm wondering if that's intentional or not. I don't have a preference, but good to understand for tutorial wording and explanation of how the viewer works.

https://github.com/user-attachments/assets/e96de9a2-7161-40bf-93ac-9fba443e4bac

kieftrav:
> Looks good! One behavior to double check is -- do you intend for volunteers to use the historgram to hide all data except their annotations? Here's a screen recording where I made several annotations, but there are some light-colored patches still un-marked. Those un-marked patches still display at the far range of the histogram slider, so I'm wondering if that's intentional or not. I don't have a preference, but good to understand for tutorial wording and explanation of how the viewer works.
> 
>  histogram.mov

I don't actually think there's been an intentional decision around this. Will bring it up in our next meeting and a follow up PR if is relevant.

PR #6811, opened 2025-04-01
Fix LineControls bug with zoom refactor

kieftrav:
## Package
- lib-classifier

- https://github.com/zooniverse/front-end-monorepo/issues/6810

## Describe your changes
Fix reference for calculating where the LineControls component should layout. There is some undesired positioning due to the zoom/scaling factor being an integer instead of a float.

## How to Review
Follow same steps as issue. 

## General

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
@@ -20,7 +20,7 @@ const LineControls = forwardRef(function LineControls({
   ref
 ) {
   const scale = useScale()
-  let [activePosition, setActivePosition] = useState('tr')
+  let [activePosition, setActivePosition] = useState('tl')
   let [showDeleteButtons, setShowDeleteButtons] = useState(false)
   const { t } = useTranslation('plugins')
 
@@ -32,23 +32,20 @@ const LineControls = forwardRef(function LineControls({
   const OUTER_RADIUS = 40 / scale
   const INNER_RADIUS = 20 / scale
 
-  const { canvas } = useContext(SVGContext)
+  const { canvas, rotate } = useContext(SVGContext)
 
   // Get the transformation that's been applied to a parent element
-  const matrix = canvas.parentElement.parentElement.getAttribute('transform');
-  const regex = /\((.*?)\)/;
-  const match = matrix.match(regex);
-  const [hScale, hSkew, vSkew, vScale, hTranlate, vTranslate] = match[1].split(', ').map(s => parseInt(s, 10))
+  const { matrix } = canvas.parentElement.parentElement.transform.baseVal.consolidate()
+  const { a: hScale, d: vScale, e: hTranslate, f: vTranslate } = matrix
+  const LINE_CONTROL_INSET = 70;
 
-  const { width } = canvas.getBoundingClientRect()
-  const x = -hTranlate / scale
-  const y = -vTranslate / scale
-  const w = (width / scale) / scale;
+  const { width, height } = canvas.getBBox()
+  const x = (LINE_CONTROL_INSET - hTranslate) / hScale
+  const y = (LINE_CONTROL_INSET - vTranslate) / vScale
 
-  const p = (activePosition == 'tr')
-    ? { x: x + 50, y: y + 50 }
-    : { x: x + w - 50, y: y + 50 }
-
+  const p = (activePosition === 'tl')
+    ? { x, y }
+    : { x: x + ((width - (LINE_CONTROL_INSET * 2)) / hScale), y }
 
   function onEnterOrSpace(ev, func) {
     if (ev.keyCode === 13 || ev.keyCode === 32) {
@@ -58,7 +55,7 @@ const LineControls = forwardRef(function LineControls({
     }
   }
 
-  function deleteMark(event) {
+  function deleteMark() {
     setShowDeleteButtons(true)
   }
 
@@ -203,6 +200,7 @@ const LineControls = forwardRef(function LineControls({
       aria-label={t('lineControls.lineControls')}
       focusable='false'
       fill={FILL_COLOR}
+      transform={`rotate(${-rotate} ${width / 2} ${height / 2})`}
     >
       {buttons.map(({ label, path, action, icon }, index) => {
         const IconComponent = icon.type
--- a/packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
+++ b/packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
@@ -7,10 +7,9 @@ import { useTranslation } from '@translations/i18n'
 import { Pan, Radial, Redo, Trash, Undo, Checkmark, Close } from 'grommet-icons'
 import useScale from '../../../hooks/useScale'
 
-const StyledPath = styled('path')`
-  &:hover {
-    cursor: pointer;
-  }
+const StyledGroup = styled.g`
+  cursor: pointer;
+  pointer-events: all !important;
 `
 
 const LineControls = forwardRef(function LineControls({
@@ -33,11 +32,22 @@ const LineControls = forwardRef(function LineControls({
   const OUTER_RADIUS = 40 / scale
   const INNER_RADIUS = 20 / scale
 
-  const { viewBox, rotate, width, height } = useContext(SVGContext)
-  const [x, y, w, h] = viewBox.split(' ').map(n => parseInt(n, 10))
+  const { canvas } = useContext(SVGContext)
+
+  // Get the transformation that's been applied to a parent element
+  const matrix = canvas.parentElement.parentElement.getAttribute('transform');
+  const regex = /\((.*?)\)/;
+  const match = matrix.match(regex);
+  const [hScale, hSkew, vSkew, vScale, hTranlate, vTranslate] = match[1].split(', ').map(s => parseInt(s, 10))
+
+  const { width } = canvas.getBoundingClientRect()
+  const x = -hTranlate / scale
+  const y = -vTranslate / scale
+  const w = (width / scale) / scale;
+
   const p = (activePosition == 'tr')
-    ? { x: x + (50 / scale), y: y + (50 / scale) }
-    : { x: x + w - (50 / scale), y: y + (50 / scale) }
+    ? { x: x + 50, y: y + 50 }
+    : { x: x + w - 50, y: y + 50 }
 
 
   function onEnterOrSpace(ev, func) {
@@ -187,19 +197,18 @@ const LineControls = forwardRef(function LineControls({
   const buttons = (showDeleteButtons) ? deleteButtons : defaultButtons
 
   return (
-    <g
+    <StyledGroup
       ref={ref}
       role='group'
       aria-label={t('lineControls.lineControls')}
       focusable='false'
-      transform={`rotate(${-rotate} ${width / 2} ${height / 2})`}
       fill={FILL_COLOR}
     >
-      {buttons.map(({ fill, label, path, action, icon }, index) => {
+      {buttons.map(({ label, path, action, icon }, index) => {
         const IconComponent = icon.type
         return <g focusable='false' key={index}>
           <Tooltip label={label}>
-            <StyledPath
+            <path
               role='button'
               aria-label={label}
               focusable='true'
@@ -210,7 +219,7 @@ const LineControls = forwardRef(function LineControls({
               strokeWidth={STROKE_WIDTH}
               onPointerDown={action}
               onKeyDown={(ev) => { onEnterOrSpace(ev, action) }}
-            ></StyledPath>
+            ></path>
           </Tooltip>
           <svg
             height={icon.size}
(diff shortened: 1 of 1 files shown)


goplayoutside3 (commented):
I'm able to run this branch locally and draw on [Science Scribbler](https://local.zooniverse.org:3000/projects/chloezycheng/science-scribbler-synapse-safari/classify/workflow/28330?env=production&demo=true). However, the LineControls are positioned far away from the line I marked on the subject. Where is the position intended to be for drawn lines?

drawn mark

kieftrav:
> I'm able to run this branch locally and draw on [Science Scribbler](https://local.zooniverse.org:3000/projects/chloezycheng/science-scribbler-synapse-safari/classify/workflow/28330?env=production&demo=true). However, the LineControls are positioned far away from the line I marked on the subject. Where is the position intended to be for drawn lines?
> 
> drawn mark

It's supposed to sit statically in the top-left corner and have the whole thing visible. More like this screenshot:

Screenshot 2025-04-01 at 3 31 59 PM

goplayoutside3:
I see. My screenshot is from a 14inch Macbook, and the LineControls are partially hidden out of bounds of the viewer. Could you position them closer to drawn line? What do the controls look like on this branch on small screens like an ipad?

goplayoutside3:
Oh I think the difference between our screenshots is actually the classifier layout. The Science Scribbler workflow from which the original bug was reported is using CenteredLayout. I think just double check that LineControls are never out-of-bounds of the viewer in workflows with that layout, and then happy to re-review!

kieftrav:
> Oh I think the difference between our screenshots is actually the classifier layout. The Science Scribbler workflow from which the original bug was reported is using CenteredLayout. I think just double check that LineControls are never out-of-bounds of the viewer in workflows with that layout, and then happy to re-review!

@goplayoutside3 - how would I check if it's out of bounds? Do we have a hook/helper related to this?

goplayoutside3:
Run app-project on this branch locally and visit Science Scribbler per my comment above: https://local.zooniverse.org:3000/projects/chloezycheng/science-scribbler-synapse-safari/classify/workflow/28330?env=production&demo=true

Any mark drawn on that workflow displays the LineControls partially out-of-bounds of the subject viewer.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
```diff
@@ -33,11 +32,22 @@ const LineControls = forwardRef(function LineControls({
...
-  const [x, y, w, h] = viewBox.split(' ').map(n => parseInt(n, 10))
+  const { canvas } = useContext(SVGContext)
+
+  // Get the transformation that's been applied to a parent element
+  const matrix = canvas.parentElement.parentElement.getAttribute('transform');
+  const regex = /\((.*?)\)/;
+  const match = matrix.match(regex);
+  const [hScale, hSkew, vSkew, vScale, hTranlate, vTranslate] = match[1].split(', ').map(s => parseInt(s, 10))
+
+  const { width } = canvas.getBoundingClientRect()
```
eatyourgreens, line 43: Wouldn’t `canvas.getBBox()` make more sense here, assuming you want the canvas width in SVG coordinates?

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
```diff
@@ -33,11 +32,22 @@ const LineControls = forwardRef(function LineControls({
...
+  const x = -hTranlate / scale
+  const y = -vTranslate / scale
```
eatyourgreens, line 45: Sorry to interrupt the review, but these offsets are broken. The x and y offset now changes as you zoom in and out, because of the dependency on scale here. I recorded a short video in Firefox.


https://github.com/user-attachments/assets/bf769539-8f42-407b-a7e5-aebc17717cdd



Also, I don't remember how the positioning was set when we first set this up, but rotating the subject image now moves the controls outside the SVG viewport.

eatyourgreens: If I can make a suggestion, I would replace this positioning code with code that fixes the controls to either the top left or top right of the subject viewport. Here's a quick hack that I made to this branch on my lunch break, then ran locally.


https://github.com/user-attachments/assets/a1bdb099-3c00-4b83-9b7d-2c019d923396


https://github.com/user-attachments/assets/e18c7292-f08b-4e5b-bf9d-078413ad1b18

eatyourgreens:
```js
const { width } = canvas.getBBox()
  const offset = 50 / scale
  const x = activePosition === 'tr' ? offset - (hTranlate / hScale) : width - offset + (hTranlate / hScale)
  const y = offset - (vTranslate / vScale)

  const p = { x , y  }
```

eatyourgreens: When the image is rotated, you probably want to modify my simple code to include skew values from the transform matrix too.

eatyourgreens: You might also want to put the positioning logic into a memo-ised function. I see a lot of flickering of the line controls when I zoom in Firefox.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
```diff
@@ -33,11 +32,22 @@ const LineControls = forwardRef(function LineControls({
...
 
-  const { viewBox, rotate, width, height } = useContext(SVGContext)
-  const [x, y, w, h] = viewBox.split(' ').map(n => parseInt(n, 10))
+  const { canvas } = useContext(SVGContext)
+
+  // Get the transformation that's been applied to a parent element
+  const matrix = canvas.parentElement.parentElement.getAttribute('transform');
+  const regex = /\((.*?)\)/;
+  const match = matrix.match(regex);
+  const [hScale, hSkew, vSkew, vScale, hTranlate, vTranslate] = match[1].split(', ').map(s => parseInt(s, 10))
```
eatyourgreens, line 41: You might get better performance here if you get the matrix directly from the SVG DOM.

```suggestion
  // Get the transformation that's been applied to a parent element
  const { matrix } = canvas.parentElement.parentElement.transform.baseVal.consolidate()
  const { a: hScale, b: hSkew, c: vSkew, d: vScale, e: hTranslate, f: vTranslate } = matrix
```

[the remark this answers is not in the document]
eatyourgreens: Getting the matrix directly from the DOM would be better than trying to parse an attribute string here.
https://github.com/zooniverse/front-end-monorepo/pull/6811#discussion_r2024945586

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
eatyourgreens: If you want more compact code, this works to fix the line controls in place:

```js
  // Get the transformation that's been applied to a parent element
  const { matrix } = canvas.parentElement.parentElement.transform.baseVal.consolidate()
  const { a: hScale, b: hSkew, c: vSkew, d: vScale, e: hTranslate, f: vTranslate } = matrix

  const { width } = canvas.getBBox()
  const x = (70 - hTranslate) / hScale
  const y = (70 - vTranslate) / vScale

  const p = (activePosition == 'tr')
    ? { x: x, y: y }
    : { x: width - x, y: y }
```

kieftrav: @eatyourgreens - great work on this! That exactly fixed the issue. I couldn't figure out how to get the base matrix transformation without parsing the string (`transform.baseVal.consolidate()` being the key) as the transformation isn't applied to the base canvas.

I just had to re-add in some code to handle rotations and it works better than I remember it working (there was still some weird zoom issues on occasion, which I think might have been part of the larger zoom refactor cleanup).

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
```diff
@@ -33,11 +32,22 @@ const LineControls = forwardRef(function LineControls({
...
+
+  const { width } = canvas.getBoundingClientRect()
+  const x = -hTranlate / scale
+  const y = -vTranslate / scale
+  const w = (width / scale) / scale;
+
   const p = (activePosition == 'tr')
```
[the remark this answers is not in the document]
kieftrav: Yup - don't know how that got missed during the last review but have fixed it in this PR so that "tl" => top left.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
eatyourgreens: There might be a bug with mouse wheel zoom, which I think might be fixed by using `vScale` to scale x. For some reason, the mouse wheel didn’t seem to set `hScale` when I tried this with a mouse. 

Other than that, I didn’t see any problems with the positioning.

goplayoutside3 (approved):
Looks much better, thank you both. I'll echo what Jim said here:
> You might also want to put the positioning logic into a memo-ised function. I see a lot of flickering of the line controls when I zoom in Firefox.

I can replicate that behavior on Chrome, too. It's not blocking because this PR is intended to restore the ability to classify Science Scribbler, but good to keep in mind there's improvements available to LineControls.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
eatyourgreens: Consolidating a transform list, to get the matrix, is documented on MDN, but you really have to dig into the API documentation.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
```diff
@@ -33,12 +32,19 @@ const LineControls = forwardRef(function LineControls({
...
-    ? { x: x + (50 / scale), y: y + (50 / scale) }
-    : { x: x + w - (50 / scale), y: y + (50 / scale) }
+  const { canvas, rotate } = useContext(SVGContext)
+
+  // Get the transformation that's been applied to a parent element
+  const { matrix } = canvas.parentElement.parentElement.transform.baseVal.consolidate()
+  const { a: hScale, d: vScale, e: hTranslate, f: vTranslate } = matrix
+
+  const { width, height } = canvas.getBBox()
+  const x = (70 - hTranslate) / hScale
```
eatyourgreens, line 42: This might need checking with a mouse wheel. When I tried zooming with a real mouse, the line control position was wrong for the top right position.

eatyourgreens: Top left doesn't have problems with wheel zoom but top right does this (recorded very quickly with a MacBook trackpad.)

Personally, I wouldn't let this block a fix for a crashing bug, but I would mark it for fixing in the next patch release.


https://github.com/user-attachments/assets/dcec031d-ef7e-4b78-885c-8743fafd4f67

eatyourgreens: Oh, I see what the problem is. The toolbar buttons zoom in on the centre of the image, where the top right offset is `width - x`. The mouse wheel zooms in on the position of the mouse wheel, which has a different offset. That doesn't affect the top left position.

kieftrav: Good catch - there was an additional adjustment to get the LineControls when they're in the `tr` to maintain location as you zoom/pan.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.js
```diff
@@ -33,12 +32,20 @@ const LineControls = forwardRef(function LineControls({
...
+  const { a: hScale, d: vScale, e: hTranslate, f: vTranslate } = matrix
+  const LINE_CONTROL_INSET = 70;
+
+  const { width, height } = canvas.getBBox()
+  const x = (LINE_CONTROL_INSET - hTranslate) / hScale
+  const y = (LINE_CONTROL_INSET - vTranslate) / vScale
 
+  const p = (activePosition === 'tl')
+    ? { x, y }
+    : { x: x + ((width - (LINE_CONTROL_INSET * 2)) / hScale), y }
```
eatyourgreens, line 48: Maybe add a comment here to explain why you're subtracting the inset twice: once to get back to the right-hand edge of the image, then again to inset inside the image itself.

eatyourgreens:
>I can replicate that behavior on Chrome, too. It's not blocking because this PR is intended to restore the ability to classify Science Scribbler, but good to keep in mind there's improvements available to LineControls.

#6815 should fix the bug in the matrix calculations that was causing the controls to jump around during pan and zoom.

PR #6839, opened 2025-04-14
Add a mock OAuth API to All Projects stories

eatyourgreens:


packages/lib-user/src/components/shared/AllProjects/AllProjects.stories.js
```diff
@@ -14,18 +14,32 @@ export default {
...
+              users: [{
+                id: '1234',
+                login: 'testuser',
+                display_name: 'Test User',
+              }]
+            })
+          }
+          return HttpResponse.json({
+            error: 'Unauthorized'
+          }, { status: 401 })
```
eatyourgreens, line 38: Copilot suggested this. I'm not sure if it is the response that Panoptes OAuth sends when the bearer token doesn't match.

goplayoutside3 (approved):
Thank you for the working example!

PR #6852, opened 2025-04-21
fix: build lib-content with vite

eatyourgreens:
- explicitly set `@zooniverse/panoptes-js` to `type: commonjs`.
- update `@zooniverse/react-components` to destructure `panoptesJS` after importing the API client.
- add a `prepare` script to automatically build `lib-content` during the bootstrap script.


eatyourgreens:
`yarn bootstrap` fails on this branch, because the exported `content.js` file can't be used in purely server-side RSCs (such as the root app 404 page.)

eatyourgreens:
This discussion has some solutions for preserving directives in a Vite build. 
- https://github.com/vitejs/vite/discussions/15721

goplayoutside3 (approved):
Thank you for this, and the linked reading is really helpful! For now, I'm going to keep #6851 as a draft example of how to convert tests and the build process to Vite. I'm going to open a secondary experimental PR that refactors only Mocha --> Vitest, so we'll have an example of both scenarios, and can make an informed decision on how to move forward with FEM libraries.

PR #6859, opened 2025-04-24
build: transform ZRC imports in user and content libs

eatyourgreens:
Add `babel-plugin-transform-imports` to lib-user and lib-content, so that imports from `@zooniverse/react-components` can be transformed into tree-shakeable forms. Otherwise, every import from ZRC is importing the whole library.

```js
// source code
import { Modal } from '@zooniverse/react-components';

// transformed output
import Modal from '@zooniverse/react-components/Modal';
```

## Package
- lib-content
- lib-user

- #6780 
- #6851 

## How to Review
This doesn't change anything functionally. Output bundles might be a bit smaller, where they were importing the whole ZRC library before. The library output files, in `dist/esm`, should transform imports from ZRC so that they only import individual components, but that change should be invisible to the build overall.

# Checklist

## General

## General UX


eatyourgreens:
The root app already transforms imports from `@zooniverse/react-components`, but I'm not 100% certain if that transform is also applied to imports via dependencies.

https://github.com/zooniverse/front-end-monorepo/blob/da0e8e0ee88264661f95ffdb88f6149716ec6578/packages/app-root/next.config.mjs#L27-L32

eatyourgreens:
The user and content libs may need similar transformations.

```js
// this imports the whole content library.
import { CommunityContainer } from '@zooniverse/content';

// this imports only the CommunityContainer component.
import CommunityContainer from '@zooniverse/content/CommunityContainer';
```

goplayoutside3 (commented):
This PR increases the bundle size of lib-user and lib-content. It slightly decreases the First Load JS for app-root pages importing lib-content components (-4kb), and has no effect on app-root pages importing lib-user content. Let me know if you've other testing methods that document the effectiveness of these babel plugins.

### Before this PR:
- lib-content
    - esm: 260kb
    - cjs: 346kb
- lib-user
    - esm: 507kb
    - cjs: 1mb
app-root build before

 _______

### With the changes in this PR
- lib-content
    - esm: 261kb
    - cjs: 347kb
- lib-user
    - esm: 514kb
    - cjs: 1.1mb
app-root build after

eatyourgreens:
The bundle size actually decreases (422kB here vs 425kB without tree-shaking) but it's a pretty trivial change. When you do something like:

```
import { Modal } from `@zooniverse/react-components`
```

A bundler like webpack or rollup can't tree-shake that import, so your bundled output ends up containing the whole of `@zooniverse/react-components`.

Next.js has it's own [internal magic to deal with barrel files](https://nextjs.org/docs/app/guides/local-development#barrel-files), so maybe that's coming into play here.

That changes here enable tree-shaking of `@zooniverse/react-components`, so that's what I'd be looking for. This PR doesn't change the source code for either lib-content or lib-user, except to transform imports so that only specific modules are imported from ZRC. You shouldn't expect to see big changes in the `dist` directories for either library, only the transformed imports instead of barrel imports.

eatyourgreens:
> Let me know if you've other testing methods that document the effectiveness of these babel plugins.

Off the top of my head, you could take a look at `lib-classifier`, which has been using this technique to optimise build sizes for a few years now.

eatyourgreens:
if you want some background reading about optimising ES module imports:

- Next.js: [Check your imports](https://nextjs.org/docs/app/guides/local-development#3-check-your-imports)
- Vite: [Avoid barrel files](https://vite.dev/guide/performance#avoid-barrel-files)
- Webpack: [Tree shaking](https://webpack.js.org/guides/tree-shaking/)

PR #6862, opened 2025-04-29
VV added to Subject Preview

kieftrav:
## Package
- lib-react-components
- lib-subject-viewers

## Describe your changes
Modify Media component in `lib-react-components` to handle Volumetric data. This required a change to the subject file to add the data type to the JSON data. For example, subjects JSON looks like:

```
{
  "type": "volumetric",
  "data": "BASE64_STRING"
}
```

I also refactored the `lib-subject-viewers` ComponentViewer into two different views, one that's the default view and another that's a "preview" which renders only the volumetric region.

## How to Review
Testing this has multiple stages, primarily to 1) create a recent classification on a beta volumetric project, and 2) see the subject preview as expected.

1) `yarn panic && yarn bootstrap`
- You'll want to make sure `lib-react-components` develops its reference to `lib-subject-viewers`

2) Recent Classification Creation
- Load up `lib-classifier` locally, sign in, and classify the only subject on [this](https://localhost:8080/?project=kieftrav%2Fvolumetric-viewer&workflow=3831) workflow.

3) Load up `app-project` locally
- Visit the staging volumetric project [homepage](https://local.zooniverse.org:3000/projects/kieftrav/volumetric-viewer?env=staging). You should see your recent volumetric classification show up.
- Visit the staging volumetric project [classifier](https://local.zooniverse.org:3000/projects/kieftrav/volumetric-viewer/classify/workflow/3831?env=staging). You should see your recent volumetric classification show up.

4) Load up `app-root` locally
- Visit the logged-in [homepage](https://local.zooniverse.org:3000). You should see your recent volumetric classification show up

5) Load up `lib-subject-viewers` [Preview Storybook](http://localhost:6008/?path=/story/components-volumetricviewer--volume-64-x-64-x-64-preview)

6) Load up `lib-react-components` [Media => Volumetric Media Storybook](http://localhost:6007/?path=/story/components-media--volumetric-media)

## General

## General UX


packages/lib-subject-viewers/src/VolumetricViewer/components/ComponentViewerPreview.js
```diff
@@ -0,0 +1,23 @@
...
+  background-color: #000000;
+  position: relative;
+  width: 100%;
+  margin-top: -50px;
+`
+
+export const ComponentViewerPreview = ({ models }) => {
+  return (
+    <StyledBox className='volume-cube'>
+      <Cube {...models} />
```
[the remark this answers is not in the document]
goplayoutside3: Also take a look at how ComponentViewerPreview fits into the subject previews on your signed-in homepage. While it does render a cube, the parent component's aspect ratio is portrait-oriented so it looks like this:
 
Screenshot 2025-05-03 at 8 51 54 AM

goplayoutside3 (commented):
While this is an effective pattern (creating a separate component in lib-subject-viewers with only the Cube preview), I'm confused about the handling of Mobx in this PR. See my comment in VolumetricViewer for the specific line.

As noted in the description, this PR makes changes for handling subjects of `type: volumetric` instead of `type: application/json` in addition to the subject preview. I think this PR is missing the piece of handling `type: volumetric` in lib-classifier. If volumetric subjects previously had `type: application/json`, then the classifier store doesn't know of `type: volumetric` without defining that type in the store. Check out this file: [subjectLocation.js](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/subjects/Subject/SubjectLocation/SubjectLocation.js). Also take a look at all of the mock volumetric subjects in lib-classifier that still say `type: application/json`. Please consider separating the handling of subject type into a PR separate from the subject previews.

packages/lib-subject-viewers/src/VolumetricViewer/VolumetricViewer.js
```diff
@@ -11,13 +11,18 @@ import asyncStates from '@zooniverse/async-states'
 const DEFAULT_HANDLER = () => {}
 
 export default function VolumetricViewer ({
-  loadingState = asyncStates.initialized,
+  loadingState = asyncStates.success,
```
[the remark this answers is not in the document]
kieftrav: This is intended to be an inversion of responsibility such that if the component is utilized and has an external effect, the changing of this variable is managed effectively from the outside. If this isn't a requirement though (which it's not when we know subject is already loaded - which is the "what" behind the loadingState), then we should instead have its default value be that its successfully completed (i.e. not needed). The subject previews have the subject loaded before the component is even visible + passing subject data to the VV preview component. By changing it to `asyncStates.success`, we're removing the requirement to pass in a loadingState variable.

packages/lib-subject-viewers/src/VolumetricViewer/VolumetricViewer.js
```diff
@@ -11,13 +11,18 @@ import asyncStates from '@zooniverse/async-states'
...
   onReady = DEFAULT_HANDLER,
-  subject
+  subject,
+  view = 'default', // 'default' | 'preview'
 }) {
-  const { data, loading, error } = useVolumetricSubject({ onError, onReady, subject })
+  const { data, loading, error } = useVolumetricSubject({
+    onError,
+    onReady,
+    subject: JSON.parse(JSON.stringify(subject)) // doesn't work properly with MobX-structured subjects
```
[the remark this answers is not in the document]
kieftrav: That's not quite right - I'll address the `type: application/json` in another comment.. This is actually meant to de-obfuscate JSON objects sent as parameters via MobX and their use of observation of a proxy object. We don't care about observation at all - our data requirement is JSON. This defensively allows the VV to handle what kind of object gets passed in as a subject.

packages/lib-subject-viewers/src/VolumetricViewer/components/ComponentViewerPreview.js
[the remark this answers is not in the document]
kieftrav: Didn't notice that in my testing - I'll make sure to remove the OrbitControls (which is what automatically handles that click/drag). Great catch!

kieftrav:
> While this is an effective pattern (creating a separate component in lib-subject-viewers with only the Cube preview), I'm confused about the handling of Mobx in this PR. See my comment in VolumetricViewer.js for the specific line.
> 
> As noted in the description, this PR makes changes for handling subjects of `type: volumetric` instead of `type: application/json` in addition to the subject preview. I think this PR is missing the piece of handling `type: volumetric` in lib-classifier. If volumetric subjects previously had `type: application/json`, then the classifier store doesn't know of `type: volumetric` without defining that type in the store. Check out this file: [subjectLocation.js](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/subjects/Subject/SubjectLocation/SubjectLocation.js). Also take a look at all of the mock volumetric subjects in lib-classifier that still say `type: application/json`. Please consider separating the handling of subject type into a PR separate from the subject previews.

The subjects are actually still stored in MobX as `application/json` - that wasn't the change. The change is that the uploaded JSON data that constitutes a subject changed. It went from `<BASE64DATA>` to `{ type: 'volumetric', data: <BASE64DATA> }`. It didn't affect the MobX process of handling a subject generically - it's a modification in the JSON data that's parsed. This change enabled each of the subject previews to effectively use the JSON to determine the viewer client-side, and leverages the existing process of handling JSON subjects differently. There are two comments I'll make just after this comment that'll help illustrate what I did.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/VolumetricViewer/VolumetricViewerWrapper.js
```diff
@@ -4,7 +4,9 @@ import { MobXProviderContext } from 'mobx-react'
 import { useContext } from 'react'
 
 const DEFAULT_HANDLER = () => {}
-const VolumetricViewer = lazy(() => import('@zooniverse/subject-viewers/VolumetricViewer'))
+const VolumetricViewer = lazy(() => import('@zooniverse/subject-viewers/VolumetricViewer').then(module => {
+  return { default: module.default.Full }
+}))
```
goplayoutside3, line 9: While this lazy loading works, I want to ask about the syntax. In all other FEM packages, components like Full are surfaced as available default exports instead of chaining `.then()`. For example, from lib-subject-viewers you can still surface it like this in index.js:
`export { default as VolumetricViewerFull } from './VolumetricViewer/VolumetricViewerFull.js`. It removes the need for a barrel file like ` export default {Full, Preview, View}`.

Then, here in lib-classifier the import statement would be:
`const VolumetricViewerFull = lazy(() => import('@zooniverse/subject-viewers/VolumetricViewerFull')`

I don't think chaining `.then()` has an advantage here, and same thoughts about using it to get "Preview" in Media.

packages/lib-react-components/src/Media/Media.stories.js
```diff
@@ -28,6 +28,8 @@ const VIDEO_URL =
   'https://static.zooniverse.org/fem-assets/home-video.mp4'
 const TEXT_URL =
   'https://panoptes-uploads.zooniverse.org/subject_location/f5506d1c-a0e9-4aba-a418-6a6c46a7731a.txt'
+const VOLUMETRIC_URL = 
+  'https://panoptes-uploads-staging.zooniverse.org/subject_location/3c56e6a7-be27-4ab3-93a7-c41a3c867baf.json'
```
goplayoutside3, line 32: Double checking - is this subject structured with `{ type: 'volumetric'}`?

Is the subject url in `subjectMock.js`, which is used for all of the stories, structured as expected with `{ type: 'volumetric' }`? I noticed that file is not edited in this PR, so it could be another source of `useVolumetricSubject()` failing to find the url from `subject.locations` if the mocked subject object is structured differently than an actual subject passed to the viewer(s).

packages/lib-subject-viewers/src/VolumetricViewer/VolumetricFull.js
```diff
@@ -10,29 +11,52 @@ import asyncStates from '@zooniverse/async-states'
...
+    })
+    
+    setModelState(state)
+  }, [data])
 
   const isLoading = loadingState === asyncStates.initialized ||
     loadingState === asyncStates.loading ||
-    loading
+    loading || 
+    modelState === null
```
goplayoutside3, line 59: When I view the test volumetric project's classify page, the viewer is flashing in and out of its loading state. Here's a video using Brave Browser (like Chrome):

https://github.com/user-attachments/assets/74a5815c-d046-40a2-b8df-389152dbb1d1

The flashing is caused by the `isLoading` variable flipping back and forth between `true` and `false`, resulting in React re-rendering this viewer over and over. I added a `console.log(isLoading)` while running locally and reliably replicated the sporadic true/false behavior while the subject loads. This PR changes handling of `loadingState` and `modelState` and therefore how those contribute to `isLoading`, so that's where I'd start investigating. You might need to add some more subjects to this test workflow to test classifier loading behavior, and also review how `useEffect` contributes to re-renders of the viewer components.

goplayoutside3 (commented):
Thanks for splitting out the three different exports from lib-subject-viewers. I'm able to see subject previews on the staging project's homepage, classify page, and a subject preview on my homepage.

Unfortunately, the same flashing bug I noted before hasn't been solved. I was able to grab a video of the classifier behavior, which I'll attach to my comment in the viewer. Let me know if you'd like to chat about any of my comments!

kieftrav:
@goplayoutside3 - thanks for the second round review! I've fixed the crufty code, changed the way the VolumetricViewer exports its 3 different component views for better import statement naming, and updated the README to reflect these details.

packages/lib-subject-viewers/src/VolumetricViewer/components/ComponentView.js
goplayoutside3, line 1: What is this empty file for?

packages/lib-subject-viewers/src/VolumetricViewer/VolumetricView/index.js
goplayoutside3, line 1: I don't think this VolumetricView component is used for anything yet, is that correct? Okay to include it in this PR because it's likely useful for the new subject-level Talk page once built, but just wanted to confirm it's not intended for use somewhere right now?

goplayoutside3 (approved):
The previews look much better, and I tested primarily by running app-root locally to see my signed-in homepage and looking at https://fe-project-branch.preview.zooniverse.org/projects/kieftrav/volumetric-viewer/classify/workflow/3831?env=staging.

Couple of non-blocking things on this branch:
1. Noted in an earlier review -- the subject preview in the signed-in homepage's Recent Classifications section is inside a portrait-oriented 'card' component. The ComponentPreview cube is a landscape rectangle canvas. Please document in this PR if you intend to keep that UI when merging this PR, will address it later, or keep the UI for project launch.
2. Check out the VolumetricViewer _lib-classifier_ story: http://localhost:6006/?path=/story/subject-viewers-volumetricviewer--default. I think it's been broken for awhile, but hopefully there's a simple fix of its props that can be included in this PR.

PR #6962, opened 2025-07-15
Replace icons with the new IconActionButton from lib-react-components & add tooltips

kieftrav:
## Package
- [ ] lib-subject-viewers

## Describe your changes
Replace buttons in the VolumetricViewer with the new `IconActionButton`. Renamed `isActive` to `isDisabled` as that better encapsulates the desired behavior.

## How to Review
Run storybook for `lib-subject-viewers` to see the updated buttons.
- With no marks all buttons are disabled
- When a mark is created, all 3 buttons should become enabled
- When any button is clicked, all 3 buttons should become disabled. The only exception is `undo`, which will keep all buttons active until the mark itself has been deleted because you undid the first "click" of the mark

## Refactoring

### DIFF
--- a/packages/lib-subject-viewers/src/VolumetricViewer/components/MarkButtons.js
+++ b/packages/lib-subject-viewers/src/VolumetricViewer/components/MarkButtons.js
@@ -4,8 +4,6 @@ import DeleteIcon from './DeleteIcon'
 import styled from 'styled-components'
 import { IconActionButton } from '@zooniverse/react-components'
 
-console.log('IconActionButton', IconActionButton)
-
 const StyledMarkButtons = styled.div`
   display: flex;
   align-items: center;


seanmiller26:
Glad we are unifying styling where we can! The disabled state looks fine, and a user will know they cannot click them in that state. 

One minor thing I notice is the 'Add a New' icon looks different from the others. I'm wondering if we should take this opportunity to change it to the standard Grommet `Edit`, and if that'll help the styling.

kieftrav:
@seanmiller26 - replaced with Grommet's edit icon. The SVG like stroke is more inline with the other icons, which helps increase legibility. If you'd like to continue with the replacement, please give a thumbs up.

Screenshot 2025-07-17 at 7 06 16 AM

packages/lib-subject-viewers/src/VolumetricViewer/components/MarkButtons.js
```diff
@@ -1,60 +1,23 @@
 import { useEffect, useState } from 'react'
-import { Revert } from 'grommet-icons'
+import { Edit, Revert } from 'grommet-icons'
 import DeleteIcon from './DeleteIcon'
-import MarkCreateIcon from './MarkCreateIcon'
 import styled from 'styled-components'
+import { IconActionButton } from '@zooniverse/react-components'
+
+console.log('IconActionButton', IconActionButton)
```
mcbouslog:
```suggestion
```

PR #7017, opened 2025-09-05
fix(app-root): check user loading state with admin mode

eatyourgreens:
Check that the Panoptes user has loaded before checking admin status.

- This ports the changes from #7013 across to the root app, so that the `useAdminMode` hooks are in sync.

## How to Review

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

## New Feature

## Refactoring

## Maintenance

## Post-merge

### DIFF
--- a/packages/app-root/src/hooks/useAdminMode.js
+++ b/packages/app-root/src/hooks/useAdminMode.js
@@ -7,17 +7,21 @@ const adminBorderImage = 'repeating-linear-gradient(45deg,#000,#000 25px,#ff0 25
 
 export default function useAdminMode(user) {
   const [adminState, setAdminState] = useState(storedAdminFlag)
-  const adminMode = user?.admin && adminState
+  const userIsLoaded = user !== undefined
+  const userIsLoggedIn = !!user?.id
+  const userIsAdmin = userIsLoggedIn && user?.admin
+  const adminMode = userIsLoaded && userIsAdmin && adminState
 
   useEffect(function onUserChange() {
-    const isAdmin = user?.admin
-    if (isAdmin) {
-      const adminFlag = !!localStorage?.getItem('adminFlag')
-      setAdminState(adminFlag)
-    } else {
-      localStorage?.removeItem('adminFlag')
+    if(userIsLoaded) {
+      if (userIsAdmin) {
+        const adminFlag = !!localStorage?.getItem('adminFlag')
+        setAdminState(adminFlag)
+      } else {
+        localStorage?.removeItem('adminFlag')
+      }
     }
-  }, [user?.admin])
+  }, [userIsAdmin, userIsLoaded])
 
   useEffect(function onAdminChange() {
     if (adminMode) {


packages/app-root/src/hooks/useAdminMode.js
```diff
@@ -7,17 +7,21 @@ const adminBorderImage = 'repeating-linear-gradient(45deg,#000,#000 25px,#ff0 25
 
 export default function useAdminMode(user) {
   const [adminState, setAdminState] = useState(storedAdminFlag)
-  const adminMode = user?.admin && adminState
+  const userIsLoaded = user !== undefined
```
eatyourgreens: The project app has `user.loadingStatus` but that doesn't exist in the root app, so I'm assuming that the user is undefined while loading, then null if loaded but not logged in.

packages/app-root/src/hooks/useAdminMode.js
```diff
@@ -7,17 +7,21 @@ const adminBorderImage = 'repeating-linear-gradient(45deg,#000,#000 25px,#ff0 25
 
 export default function useAdminMode(user) {
```
eatyourgreens: `app-root` and `app-project` have different signatures here. ☹️ 

```js
// app-root
const { adminMode } = useAdminMode(user);

// app-project
const { adminMode } = useAdminMode();

Comment (eatyourgreens) [reply packages/app-root/src/hooks/useAdminMode.js]: Fixed in this PR.

File: packages/app-root/src/hooks/useAdminMode.js (hunk 1 above)
Comment (eatyourgreens) [reply packages/app-root/src/hooks/useAdminMode.js]: Fixed by importing `usePanoptesUser`, which gives you the user object and loading state.

File: packages/app-root/src/hooks/useAdminMode.js [hunk 3]
```diff
@@ -1,23 +1,29 @@
 import { useEffect, useState } from 'react'
+import { usePanoptesUser } from '@zooniverse/react-components/hooks'
 
 const isBrowser = typeof window !== 'undefined'
 const localStorage = isBrowser ? window.localStorage : null
 const storedAdminFlag = !!localStorage?.getItem('adminFlag')
 const adminBorderImage = 'repeating-linear-gradient(45deg,#000,#000 25px,#ff0 25px,#ff0 50px) 5'
 
-export default function useAdminMode(user) {
+export default function useAdminMode() {
```
shaunanoordin, line 9: Since useAdminMode() no longer requires `user`, we can remove said argument from the function call in https://github.com/zooniverse/front-end-monorepo/blob/db926bb/packages/app-root/src/components/PageContextProviders.js#L27

eatyourgreens: Done! This is the kind of detail that Copilot is good at catching.

eatyourgreens:
Here's a video of the bug in the root app. If you watch the storage panel, the `adminFlag` key gets deleted when the About page reloads (without a previously stored user object.) I think admin mode will turn off when you leave the root app and go elsewhere, as the admin state isn't persisted in local storage.

This is recorded on the main branch, with `user.admin` hardcoded to true in the browser (affter the user authenticates.)
https://github.com/user-attachments/assets/72331988-61e8-4be8-9b8b-e33f4fd2c146

packages/app-root/src/hooks/useAdminMode.js
shaunanoordin: Hey, are you calling _me_ a copilot??

I'm more of a backseat driver.

PR #7054, opened 2025-09-24
RotateRectangle: new resize behaviour

### LINKED ISSUE #7052: RotateRectangle drawing tool: change resizing behaviour
## UI Behaviour Change Request

Package: lib-classifier
Affects: Drawing Task -> RotateRectangle drawing tool

We want to change how the RotateRectangle drawing tool behaves, in regards to **resizing existing marks**.

Say we have a rectangle mark, with four corners ABCD, and origin/centre-coordinate O:

```
  A-------B
  |       |
  |   O   |
  |       |
  D-------C
```

- Currently: **resizing is anchored to origin O.** So if we drag C ↘️, it will automatically move A ↖️, B ↗️, and D ↙️. O will remain static.
- We want: **resizing to be anchored to the opposite corner.** So if we drag C ↘️, A will remain static, B will move ➡ and D will move ⬇️. O will move ↘️ too, because that's maths.
  - This is basically the same as how the standard Rectangle drawing tool works...
  - ...except that it must respect the rotation of the RotateRectangle.
- **Important: the annotation data shape must remain unchanged; only the UI behaviour should change.**

Or in video, here's what we have (i.e. current RotateRectangle):

https://github.com/user-attachments/assets/17806382-2863-456a-8411-c61c57cefd60

And here's what we want (i.e. Rectangle):

https://github.com/user-attachments/assets/3ea38934-5858-42f8-99c7-7e9675d09d68

This change was requested by @snblickhan on 23 Sep 2025, in response to _HCO exp 1_ testing feedback & in preparation for the upcoming _HCO 3.0 box collection project_.

### Testing

- Choose any workflow with a Drawing Task with 1x Rotate Rectangle and 1x Rectangle
- Example: https://local.zooniverse.org:8080/?env=production&project=blicksam/demonstration-project&workflow=29992 (lib-classifier) or https://frontend.preview.zooniverse.org/projects/blicksam/demonstration-project/classify/workflow/29992
- For the Rectangle mark:
  - Create a mark.
  - Drag a corner to resize the rectangle.
  - Observe how the other 3 corners move (or not) relative to the movement of the dragged corner.
- For the RotateRectangle mark:
  - Do the same as above, and compare differences.

### Dev Notes

The code that needs to be changed can be found in [packages/lib-classifier/src/plugins/drawingTools/components/RotateRectangle/RotateRectangle.jsx](https://github.com/zooniverse/front-end-monorepo/blob/4f4603f/packages/lib-classifier/src/plugins/drawingTools/components/RotateRectangle/RotateRectangle.jsx), e.g.:

```
{/* Original Bottom Right corner */}
{active && (
  <DragHandle
    x={x_right}
    y={y_bottom}
    dragMove={(e, d) =>
      onHandleDrag({
        x_left: x_left - rotateXY(d, angle).x,
        x_right: x_right + rotateXY(d, angle).x,
        y_top: y_top - rotateXY(d, angle).y,
        y_bottom: y_bottom + rotateXY(d, angle).y,
        angle: angle
      })
    }
  />
)}
```
 
Changing the values sent to onHandleDrag should sort the issue, but the problem is **maths be hard yo**. We can't just pin the values to `x_left: x_left, y_top: y_top` and hope that the top-left corner remains anchored (like how the [Rectangle mark](https://github.com/zooniverse/front-end-monorepo/blob/4f4603f/packages/lib-classifier/src/plugins/drawingTools/components/Rectangle/Rectangle.jsx) does it), because that doesn't work when the angle is anything other than 0º.

There is a solution to this, it's just going to require some maths homework.

Also see: [packages/lib-classifier/src/plugins/drawingTools/models/marks/RotateRectangle/RotateRectangle.js](https://github.com/zooniverse/front-end-monorepo/blob/4f4603f/packages/lib-classifier/src/plugins/drawingTools/models/marks/RotateRectangle/RotateRectangle.js)'s setCoordinates().

### Status

Investigation in progress. Solution requested before end of week (Fri 26 Sep 2025).

shaunanoordin:
## PR Overview

Package: lib-classifier
Affects: Drawing Task -> RotateRectangle drawing tool
Closes #7052 

This PR changes how the RotateRectangle marks behaves, when a user resizes it by moving/dragging one of its four corners (drag handles).

Say we have a rectangle mark, with four corners ABCD, and origin/centre-coordinate O:
```
  A-------B
  |       |
  |   O   |
  |       |
  D-------C
```

New behaviour:

- RotateRectangle: when resizing by dragging a corner, the action is _anchored_ to the _opposite corner._
  - So if we drag C ↘️, A will remain static, B will move ➡ and D will move ⬇️. O will move ↘️ too, by half the distance that C moves.
  - This is basically the same as how the standard Rectangle drawing tool works, except that it _respects the rotation_ of the RotateRectangle.
- **Important: the annotation data shape remains unchanged; only the UI behaviour changed.**

### Testing

What you're looking for is that the **resize behaviour of the RotateRectangle feels natural,** and that the behaviour is consistent no matter how rotated the rectangle _(or the Subject image!)_ is.

- Choose a workflow with a Drawing Task with 1x Rotate Rectangle and 1x Rectangle tool
  - Example: https://local.zooniverse.org:8080/?env=production&project=blicksam/demonstration-project&workflow=29992 (lib-classifier)
  - Compare with current baseline: https://frontend.preview.zooniverse.org/projects/blicksam/demonstration-project/classify/workflow/29992
- For the RotateRectangle too:
  - Select the tool.
  - Click & drag to create a mark.
  - Drag a corner to resize. Confirm that the opposite corner remains anchored in place while the other corners + origin/centre moves.
  - Rotate the mark.
  - Drag a corner to resize and confirm etc etc
  - Pan, zoom, and rotate _the Subject Image._
  - Drag a corner to resize and confirm etc etc
- For comparison, also try the Rectangle mark:
  - Create a mark.
  - Drag a corner to resize the rectangle. The feel between resizing a RotateRectangle mark and resizing a Rectangle mark should be similar.

### Status

Ready for review. Tagging @snblickhan for eyes on this. We want to get this merged soon, so it can get deployed next week.

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/components/RotateRectangle/RotateRectangle.jsx
+++ b/packages/lib-classifier/src/plugins/drawingTools/components/RotateRectangle/RotateRectangle.jsx
@@ -102,15 +102,13 @@ function RotateRectangle({
         <DragHandle
           x={x_left}
           y={y_top}
-          dragMove={(e, d) =>
-            onHandleDrag({
-              x_left: x_left + rotateXY(d, angle).x,
-              x_right: x_right - rotateXY(d, angle).x,
-              y_top: y_top + rotateXY(d, angle).y,
-              y_bottom: y_bottom - rotateXY(d, angle).y,
-              angle: angle
+          dragMove={(e, d) => {
+            mark.resizeByCorner({
+              dx: d.x,
+              dy: d.y,
+              corner: 'top left'
             })
-          }
+          }}
         />
       )}
 
@@ -119,28 +117,26 @@ function RotateRectangle({
         <DragHandle
           x={x_right}
           y={y_top}
-          dragMove={(e, d) =>
-            onHandleDrag({
-              x_left: x_left - rotateXY(d, angle).x,
-              x_right: x_right + rotateXY(d, angle).x,
-              y_top: y_top + rotateXY(d, angle).y,
-              y_bottom: y_bottom - rotateXY(d, angle).y,
-              angle: angle
+          dragMove={(e, d) => {
+            mark.resizeByCorner({
+              dx: d.x,
+              dy: d.y,
+              corner: 'top right'
             })
-          }
+          }}
         />
       )}
 
       {/* Original Bottom Right corner */}
       {active && (
         <DragHandle
-          fill='red'
           x={x_right}
           y={y_bottom}
           dragMove={(e, d) => {
             mark.resizeByCorner({
               dx: d.x,
-              dy: d.y
+              dy: d.y,
+              corner: 'bottom right'
             })
           }}
         />
@@ -151,15 +147,13 @@ function RotateRectangle({
         <DragHandle
           x={x_left}
           y={y_bottom}
-          dragMove={(e, d) =>
-            onHandleDrag({
-              x_left: x_left + rotateXY(d, angle).x,
-              x_right: x_right - rotateXY(d, angle).x,
-              y_top: y_top - rotateXY(d, angle).y,
-              y_bottom: y_bottom + rotateXY(d, angle).y,
-              angle: angle
+          dragMove={(e, d) => {
+            mark.resizeByCorner({
+              dx: d.x,
+              dy: d.y,
+              corner: 'bottom left'
             })
-          }
+          }}
         />
       )}
     </g>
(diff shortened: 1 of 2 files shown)


goplayoutside3 (approved):
Looks good! Confirmed that the test project's rotate rectangle tool resizes by anchoring the opposite corner, and the center point acts as a point to rotate around.

While testing I noticed that the rotate _handle_ becomes "sticky" to a mouse cursor even after "mouse up". However, I don't think it's a result of this PR. I suspect it might even be dev env only behavior, but I have noticed other drawing tools behave like this in the past, so this is just a note to say that if you're working in drawing tools code in the future, I've noticed some unexpected drag handle behavior and I don't think we have an Issue for it yet. If an active Zooniverse project is using a tool with this sticky handle behavior, please ping me or start an Issue to document! Here's an example of what I mean. After rotating, I've released my mouse click, but any hover over the handle sticks to it. Unfocusing and refocusing solves the strange behavior:

https://github.com/user-attachments/assets/ef172a5c-f367-4949-9e3c-202902131641

PR #7056, opened 2025-09-25
Node 22: Ignore sugarClient during SSR

eatyourgreens:
Dynamically import the Sugar client, so that it isn't loaded during SSR.


goplayoutside3 (approved):
Thank you!

Issue #2105, opened 2021-04-16
Disable Done and Done & Talk if task component could not be rendered


mcbouslog:
If a task can not be rendered (hasn't been converted from PFE to new classifier, invalid, etc.) disable the Done and Done & Talk buttons to prevent classification submission.

Example - https://frontend.preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify/workflow/3487?env=staging

goplayoutside3:
@kieftrav and @shaunanoordin noting that I restored the branch `disable-done-talk-on-task-render-error` from #7065, but you'll have to create a new PR on Github if continuing work toward this Issue.

For documentation, this Issue is still open because PR #7065 included the bug described in this [comment](https://github.com/zooniverse/front-end-monorepo/pull/7065#issuecomment-3503721110). There is different, unexpected behavior in the code between `pnpm dev` (dev mode) and `pnpm start` (production mode).

Furthermore, from the Slack convo https://zooniverse.slack.com/archives/CANKLB50E/p1777575341553119?thread_ts=1777574399.669049&cid=CANKLB50E. There was a bug on frontend.preview, so the PR was reverted.
> ...looking at https://frontend.preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify/workflow/3487?env=staging&demo=true and the Done buttons are enabled when they should be disabled.

I marked this Issue as "paused" in the FEM Planning spreadsheet for now.

PR #7065, opened 2025-10-07
Disable DoneButton and DoneAndTalkButton when Task component could not be rendered

kieftrav:

## Package
- [ ] lib-classifier

- Closes Issue #2105 

## Describe your changes
Add check in the Step & WorkflowStepStore for if a task couldn't be rendered (i.e. doesn't exist). Pull this into the DoneAndTalkButton and DoneButton so that we disable the button if there's a task count mismatch (i.e. a rendering issue). 

## How to Review
Load up the classifier and test out with a couple projects to see that the Done and Done&Talk buttons operate as expected.

- Workflow Disabled: https://local.zooniverse.org:8080/?project=markb-panoptes%2Ffocus-testing-project&env=staging&workflow=3487
- Workflow Enabled: (any other valid workflow): https://localhost:8080/?project=kieftrav%2Fdrawing-tools&env=staging&workflow=3867

# Checklist

## General

## New Feature

## Refactoring

## Maintenance

## Post-merge

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/DoneAndTalkButton/DoneAndTalkButtonConnector.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/DoneAndTalkButton/DoneAndTalkButtonConnector.jsx
@@ -11,6 +11,7 @@ function storeMapper(classifierStore) {
       active: subject
     },
     workflowSteps: {
+      hasUnsupportedTasks,
       shouldWeShowDoneAndTalkButton
     }
   } = classifierStore
@@ -26,6 +27,7 @@ function storeMapper(classifierStore) {
     }
 
     return {
+      hasUnsupportedTasks,
       onClick,
       talkURL: subject.talkURL,
       visible
@@ -36,9 +38,10 @@ function storeMapper(classifierStore) {
 }
 
 function DoneAndTalkConnector(props) {
-  const { onClick, talkURL, visible } = useStores(storeMapper)
+  const { hasUnsupportedTasks, onClick, talkURL, visible } = useStores(storeMapper)
+  const isDisabled = hasUnsupportedTasks || props.disabled
 
-  return visible ? <DoneAndTalkButton onClick={onClick} {...props} talkURL={talkURL} /> : null
+  return visible ? <DoneAndTalkButton {...props} disabled={isDisabled} onClick={onClick} talkURL={talkURL} /> : null
 }
 
 export default observer(DoneAndTalkConnector)
--- a/packages/lib-classifier/src/store/WorkflowStepStore/Step/Step.js
+++ b/packages/lib-classifier/src/store/WorkflowStepStore/Step/Step.js
@@ -42,6 +42,10 @@ const baseStep = types
       return isValid
     },
 
+    get hasUnsupportedTasks() {
+      return self.taskKeys.length !== self.tasks.length
+    },
+
     get isThereBranching () {
       // We return the first single choice task
       // It doesn't make sense to have more than one single choice task in a step
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/Task/Task.jsx
@@ -34,7 +34,7 @@ function Task ({
   task,
   ...props
 }) {
-  const TaskPlugin = task.type === 'dropdown-simple'? tasks.dropdownSimple : tasks[task.type]
+  const TaskPlugin = task.type === 'dropdown-simple' ? tasks.dropdownSimple : tasks[task.type]
   const TaskComponent = TaskPlugin?.TaskComponent
   let annotation
   let suggestions
--- a/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/DoneButton/DoneButtonConnector.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/DoneButton/DoneButtonConnector.jsx
@@ -12,7 +12,8 @@ function storeMapper(classifierStore) {
       completeClassification
     },
     workflowSteps: {
-      active: step
+      active: step,
+      hasUnsupportedTasks
     }
   } = classifierStore
 
@@ -37,6 +38,7 @@ function storeMapper(classifierStore) {
     }
 
     return {
+      hasUnsupportedTasks,
       hasNextStep,
       onClick
     }
@@ -46,8 +48,10 @@ function storeMapper(classifierStore) {
 }
 
 function DoneButtonConnector(props) {
-  const { hasNextStep, onClick } = useStores(storeMapper)
-  return hasNextStep ? null : <DoneButton onClick={onClick} {...props} />
+  const { hasUnsupportedTasks, hasNextStep, onClick } = useStores(storeMapper)
+  const isDisabled = hasUnsupportedTasks || props.disabled
+
+  return hasNextStep ? null : <DoneButton {...props} disabled={isDisabled} onClick={onClick} />
 }
 
 export default observer(DoneButtonConnector)
--- a/packages/lib-classifier/src/store/WorkflowStepStore/WorkflowStepStore.js
+++ b/packages/lib-classifier/src/store/WorkflowStepStore/WorkflowStepStore.js
@@ -48,6 +48,10 @@ const WorkflowStepStore = types
       return false
     },
 
+    get hasUnsupportedTasks() {
+      return Array.from(self.steps.values()).some(step => step.hasUnsupportedTasks)
+    },
+
     findTasksByType (type) {
       const tasksByType = Array.from(self.steps).map(([stepKey, step]) => {
         if (step?.tasks && step.tasks.length > 0) {


goplayoutside3 (commented):
While these changes do disable the Done and Done&Talk buttons by observing the workflow steps, it removes the helpful message rendered by Task.jsx when the classifier doesn't find the relevant TaskComponent. Is there a way to keep that helpful message in the task area?

For example, it appears at https://frontend.preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify/workflow/3487?env=staging, but not when run locally on this branch.
Screenshot 2025-10-16 at 3 00 43 PM

kieftrav:
> While these changes do disable the Done and Done&Talk buttons by observing the workflow steps, it removes the helpful message rendered by Task.jsx when the classifier doesn't find the relevant TaskComponent. Is there a way to keep that helpful message in the task area?
> 
> For example, it appears at https://frontend.preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify/workflow/3487?env=staging, but not when run locally on this branch. Screenshot 2025-10-16 at 3 00 43 PM

@goplayoutside3 - apologies for the 3 week delay on following up on this! I checked and none of the code changes touched that render path so I think it has to do with an env configuration. I deployed this branch after rebasing on master to the [fe-project-branch](https://fe-project-branch.preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify/workflow/3487?env=staging), wherein it's still displaying that text as desired.

project-branch-task-text

goplayoutside3:
Interesting 🤔 That is strange that the "unsupported" message only shows up in production mode. For instance, even when the default branch is run locally there's no message at https://localhost.zooniverse.org:3000/projects/markb-panoptes/focus-testing-project/classify/workflow/3487?env=staging with `pnpm dev`, but there _is_ a message with `pnpm start`.

While troubleshooting, I noticed that this PR branch doesn't disable the Done and Done&Talk buttons when run in production mode. You can see in your screenshot above and at the fe-project-branch link that the buttons are not disabled as desired.

Further work on this PR is needed, and it's definitely not required while you're on leave. Please notify me via comment here if you plan to follow-up, otherwise it will remain open until you're back or I find a solution through adjacent classifier dev work.

packages/lib-classifier/src/components/Classifier/components/TaskArea/components/Tasks/components/TaskNavButtons/components/DoneAndTalkButton/DoneAndTalkButtonConnector.jsx
```diff
@@ -26,6 +27,7 @@ function storeMapper(classifierStore) {
     }
 
     return {
+      disabled: hasUnsupportedTasks,
```
goplayoutside3: Rather than rename `hasUnsupportedTasks` in the `storeMapper`, I recommend passing this value named as `hasUnsupportedTasks` down into the component. Then, it's a much easier code read to understand `const isDisabled = hasUnsupportedTasks || props.disabled` inside the connector.

packages/lib-classifier/src/store/WorkflowStepStore/Step/Step.js
```diff
@@ -5,9 +5,6 @@ const taskModels = Object.values(tasks).map(task => task.TaskModel)
 
 function taskDispatcher (snapshot) {
   switch (snapshot.type) {
-    case 'dropdown': {
-      return tasks.dropdownSimple.TaskModel
-    }
```
goplayoutside3, line 10: There's some unintended consequences of editing the `dropdown` code in this PR. See the Material Cultural of Wills workflow 27828 on [deployed production](https://www.zooniverse.org/projects/hjsmith/the-material-culture-of-wills-england-1540-1790/classify/workflow/27828) versus run [locally](https://local.zooniverse.org:3000/projects/hjsmith/the-material-culture-of-wills-england-1540-1790/classify/workflow/27828?env=production&demo=true) on this branch (dev or production mode).

❗ That workflow has a combo task of transcription + dropdown, and the dropdown task is missing in the task area on this PR.

Rather than looking specifically at "dropdown" cases for unsupported task types, is there a broader check you can do in the store? Logically, is there one MST "view" that can be added that tells the TaskNavButtons a Task component failed to load? Without editing the existing `taskDispatcher`.

goplayoutside3 (dismissed):

shaunanoordin:
Ahoy hoy, just wanted to say that I'm hopping in to review this PR!

shaunanoordin (approved):
## PR Review

Package: lib-classifier
Affects: Workflows with ANY unsupported task types

This PR updates the Classifier so that volunteers can't submit Classifications if a Workflow has unsupported Tasks.

Functionality changes:

- Classifier, Task Area:
  - if a Workflow contains _any_ Tasks which is of an unsupported task type, then the Done and Done & Talk buttons will be disabled _when the user reaches the end of the Classifying process._ (I'm wording this very carefully. Bear with me. 🐻)

_Screenshot 1: currently, as see on [frontend.preview](https://frontend.preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify/workflow/3487?env=staging), if a Task isn't support, you can still submit a Classification (with 0 annotations)_

Image

_Screenshot 2: with the implementation of this PR [on localhost](https://local.zooniverse.org:3000/projects/markb-panoptes/focus-testing-project/classify/workflow/3487?env=staging), if a Task isn't supported, the Done and Done & Talk buttons are disabled. Ignore the fact that the "I have no idea wtf this Task type is" message has mysteriously disappeared. Don't worry about it._

Image

Code read:

- Workflow Step[^1] Store: now has the `hasUnsupportedTasks` read-only property, which returns false if the Workflow contains any supported Tasks.
- Done Button Connector: now checks for WorkflowStepStore.hasUnsupportedTasks.

The implementation is very straightforward and functional. 👍 I will note some wonky quirks and edge cases, but nothing that's blocking this PR.

[^1]: to be pedantic, shouldn't this be Workflow StepS (plural) ? Since it's actually a model for every Step in the in the workflow (`.steps`) in addition to the current active Step (`.active`).

### Dev Notes

**Quirk 1: "(Task Key) (Task Type) is not a supported task type" error message not appearing on localhost**

Yeah, I'm seeing the same quirk too. (See Screenshot 1 & 2) What?? This is probably a separate investigation, and low priority if it only affects devs, but it's weird as hey.

**Edge Case 1: "Why did my branching path fail?" possible confusion**

I created a branching WF that looks like this:

```
              +-> T1: invalid_type --> submit
T0: question -+
              +-> T2: question     --> submit
```

If I go down the T0 -> T1 path, the Done buttons are disabled as expected. ✔️ 

However, if I go down the T0 -> T2 path, _which seems legit from a volunteer's perspective,_ the Done buttons are also disabled... as expected? 🤔 

Give me a minute to argue with myself:

- On one hand, this is legitimate behaviour - if any Tasks in the workflow is invalid, then the workflow itself is invalid.
- On the other hand, this is very confusing for the volunteer. Imagine if I answered questions correctly down a valid path, then suddenly I can't submit. What?
- On the third appendage, project owners SHOULD catch this early on in testing, so volunteers should never ever encounter this edge case to begin with.

OK, come to think of it, I'm fine with this behaviour as-is.

**Edge Case 2: can I go past an unsupported Task?**

I haven't tested this yet, but suppose I create a Workflow that looks like...

```
T0: question --> T1: invalid_type --> T2: question --> submit
```

I'm guessing that I'l be able to answer T0, then click "Next" when I'm at T1, then when I answer T2, the Done buttons will remain disabled. This would be a variant of Edge Case 1, in the sense that it'd be confusing for volunteers, but project owners should catch this early so volunteers should never this edge case.

**Edge Case 3: someone added an unsupported Task but there's no way to get there but it blocks Done anyway.**

I haven't tested this either, but let's say I created a Workflow with 1 unsupported Task but it's not in any branching path...

```
T0: question --> T1: question --> submit

(unlinked) T2: invalid_task
```

Since volunteers will never see T2: invalid_task in the process of classifying, but the code DOES see T2, the volunteers will never understand WHY the Done buttons are disabled.

This is definitely reaching for the far edge of edge cases, but this is where my brain went. Again, this is something that'll be caught as long as pre-launch tests are performed.

**👿 Wait, hold up**

Sorry, Devil On Shaun's Shoulder wants to say something. Go ahead, buddy:

> 👿 DOSS: _sure, early testing will catch cases where volunteers run into the disabled Done button inexplicably, but are there any EXISTING projects that match, say, Edge Case 3? Because if there are, then volunteers will suddenly be unable to classify on live projects, meaning fiery brimstone wi-_

Okay buddy, you're worrying too much, time for a nap.

### Testing

Tested on app-project on localhost with macOS + Chrome.

**Baseline test: normal workflow**

- Loaded a test workflow _with only supported Task types:_ https://local.zooniverse.org:3000/projects/darkeshard/test-project-2025/classify/workflow/3850?env=staging
- Confirm that I can make classifications and submit correctly.

**Main test 1: workflow with 1 task, which is unsupported**

- Loaded a test workflow with only 1 Task, and that Task isn't supported (dropdown): https://local.zooniverse.org:3000/projects/markb-panoptes/focus-testing-project/classify/workflow/3487?env=staging
- Confirm that Done and Done & Talk buttons are disabled. (See Screenshot 2)

**Main test 2: branching workflow with 3 tasks, 1 of which is unsupported**

- Loaded a test workflow with branching structure described in Dev Notes: Edge Case 1 : https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022/classify/workflow/31770?env=production
- Walk down T0 -> T1 route.
- Confirm that Done buttons are disabled.
- Reset. Walk down T0 -> T2 route.
- Confirm that Done buttons are also disabled.

_Screenshots 3.1 and 3.2: following T0 -> T1 route_

Image Image

_Screenshots 4.1 and 4.2: following T0 -> T2 route_

Image Image

_Screenshot 5: note that even on Task T0, which isn't required, the Done buttons are disabled._

Image

### Status

LGTM 👍 

I say this PR does what it advertises to do, and the function to stop volunteers from submitting classifications on invalid workflows is a good function. (UX-wise it'd be better to stop volunteers from even _starting_ classifications on an invalid workflow, but as I mentioned, testing will prevent invalid workflows from going out to volunteers in the first place.)

shaunanoordin:
This PR can't be merged because there's a Change Request from @goplayoutside3 

However, both Delilah's CRs were addressed by Travis's recent changes:

- _Rather than rename hasUnsupportedTasks in the storeMapper, I recommend passing this value named as hasUnsupportedTasks down into the component._ --> hasUnsupportedTask is now a read-only property (view) of WorkflowStepStore and the Step model.
- _Rather than looking specifically at "dropdown" cases for unsupported task types, is there a broader check you can do in the store?_ --> this was addressed by making each Step check if its Tasks Keys match the number of its actual Tasks. (Invalid Tasks wouldn't be part of `step.tasks[]`)

I feel comfortable marking those CRs as resolved. 👍

(I'll uh, see how I can tell Github this. Brb, gonna press some buttons on the webpage.)

kieftrav:
@shaunanoordin - thanks for helping get this across the finish line! I've just rebased on main, will make sure all tests still pass, and then merge it in for deploy this week.

PR #7101, opened 2025-11-07
[DRAFT] Define SENTRY_PROJECT_DSN in Dockerfiles, kubernetes templates

mcbouslog:
[ ...incoming ]


mcbouslog:
This doesn't fix anything. I think the issue is that `SENTRY_PROJECT_DSN` is missing in the container/pod.

PR #7111, opened 2025-11-11
lib-classifier: Add initial GeoMapViewer component

### LINKED ISSUE #7098: cssi mapping nfn: Add basic OpenLayers viewer
## Package
- lib-classifier

Towards [Feature Request 1](https://docs.google.com/spreadsheets/d/1ifyRtdKmUqLBUoVcENiMQIsNn8InOM57IAcVBv3ubTQ/edit?usp=sharing), user story: "As a participant, I want to view a base map that is pre-set to a specific center view [based on lat/lon information from the data being ingested]."

Specifics
- add basic viewer, largely OpenLayers default configuration
- utilize OpenStreetMap source layer
- viewer name: `geoMap` or `GeoMapViewer`

Open Items
- [ ] check impact of adding OpenLayers library to build size, document difference, potentially move OpenLayers to lib-subject-viewers (and implement similar to Volumetric Viewer)

### LINKED ISSUE #7099: cssi mapping nfn: Add OpenLayers pan and zoom
## Package
- lib-classifier

Towards [Feature Request 2](https://docs.google.com/spreadsheets/d/1ifyRtdKmUqLBUoVcENiMQIsNn8InOM57IAcVBv3ubTQ/edit?usp=sharing), user story: "As a participant, I want to adjust the area of the map that I am viewing, including panning, zooming in and out, and resetting."

Specifics
- add or confirm OpenLayers default pan and zoom functionality operates as expected

mcbouslog:
## Package
- lib-classifier

- closes #7098 
- closes #7099 

## Describe your changes
- adds basic geographic map viewer component - GeoMapViewer
- adds OpenLayers (`ol`) dependency to lib-classifier
- refactors `getViewer`, `subjectViewers`, and `WorkflowConfiguration` for `geoMap` subject viewer
- adds `esmExternals: 'loose'` to app-project `next.config`

## How to Review
  - lib-classifier dev: https://local.zooniverse.org:8080/?project=2030
  - app-project: https://local.zooniverse.org:3000/projects/markb-panoptes/geomap-test-project-staging/classify/workflow/3874?env=staging
  - confirm subject viewer loads a map 
  - interact with map, confirming pan and zoom works as expected
  - http://localhost:6006/?path=/story/subject-viewers-geomapviewer--default

# Checklist

## General

## General UX

## New Feature

### DIFF
--- a/packages/lib-classifier/.storybook/preview.jsx
+++ b/packages/lib-classifier/.storybook/preview.jsx
@@ -3,6 +3,7 @@ import i18n from '@translations/i18n'
 import zooTheme from '@zooniverse/grommet-theme'
 import { Grommet } from 'grommet'
 import { StrictMode } from 'react'
+// import OpenLayers CSS for lib-classifier GeoMapViewer component styling
 import 'ol/ol.css'
 
 const background = {
--- a/packages/lib-classifier/dev/components/App/App.jsx
+++ b/packages/lib-classifier/dev/components/App/App.jsx
@@ -7,6 +7,7 @@ import { Component } from 'react'
 
 import Classifier from '@components/Classifier'
 import localeMenu from './localeMenu.js'
+// import OpenLayers CSS for lib-classifier GeoMapViewer component styling
 import 'ol/ol.css'
 
 const onAddToCollection = (subjectId) => console.log(subjectId)


packages/app-project/next.config.js
```diff
@@ -55,6 +55,7 @@ const nextConfig = {
   },
 
   experimental: {
+    esmExternals: 'loose',
```
mcbouslog, line 59: This experimental setting impacts how Next.js handles packages at build. Values could be: 
- `true` (default): strict ESM handling, throws errors for mismatched imports
- `false`: treats all externals as CommonJS
- `'loose'`: middle ground, more permissive ([Next.js ESM magic](https://nextjs.org/docs/messages/import-esm-externals)), if a package fails at server-side build it marks the package as external for run-time

I can't find much documentation about this experimental setting, which I don't like, but it fixes the app-project build issues due to OpenLayers being an ESM-only package.

packages/app-project/pages/_app.jsx
```diff
@@ -5,6 +5,7 @@ import { createGlobalStyle } from 'styled-components'
 import { appWithTranslation } from 'next-i18next'
 import { Grommet } from 'grommet'
 import zooTheme from '@zooniverse/grommet-theme'
+import 'ol/ol.css'
```
mcbouslog, line 9: The `ol.css` is required for OpenLayers map components to render properly. We could copy the `ol.css` file to the lib-classifier GeoMapViewer component folder, I'm certainly open to that, but then it might become out of sync with the installed OpenLayers version. It's not ideal to import it for the entire app (instead of only when GeoMapViewer is rendered), but it's small and I think importing here is within Next.js external css conventions. Open to suggestion!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/GeoMapViewer.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+  useEffect(function loadMap() {
+    let map
+
+    function createMap(target) {
+      const osmLayer = new TileLayer({
+        // preload tiles for 1 level of zooming
+        preload: 1,
+        // use OpenStreetMap as the base layer
+        source: new OSM(),
```
mcbouslog, line 24: Additional documentation regarding using OpenLayers and OpenStreetMaps will be included in #7112 .

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/README.md
```diff
@@ -0,0 +1,23 @@
+# Geographic Map Viewer
```
mcbouslog, line 1: This will continue to be updated as additional functionality and features are added, per https://github.com/orgs/zooniverse/projects/66 .

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/GeoMapViewer.jsx
```diff
@@ -0,0 +1,64 @@
...
+function GeoMapViewer() {
+  const mapRef = useRef()
+
+  useEffect(function loadMap() {
+    let map
+
+    function createMap(target) {
+      const osmLayer = new TileLayer({
+        // preload tiles for 1 level of zooming
+        preload: 1,
```
mcbouslog, line 22: This helps with initial load, with `preload: 1`, it only prefetches 1 additional zoom level. Map tiles are fetched progressively from OpenStreetMap servers as needed (user zooms/pans).

mcbouslog:
Adding the OpenLayers dependency is about 150-200KB minified. With tree-shaking and only minimal components used at this stage, I see no change in the lib-classifier build:
- from `/lib-classifier` I ran `du -sh dist/esm` on `main` and this branch `map-viewer`, both yield `4.7M`.

From `/app-project` Running `APP_ENV=production pnpm run build` also yielded the same Next.js size/first load info as noted:
- main:
main APP_ENV=production pnpm run build
- map-viewer:
map-viewer APP_ENV=production pnpm run build

Loading I Fancy Cats, dev tools/network footer:
- main:
main i_fancy_cats

- map-viewer:
map-viewer i_fancy_cats

Loading GeoMap Test Project, dev tools/network footer:
- map-viewer:
map-viewer geomap-test-project-staging

So I don't think this PR / adding OpenLayers with OpenStreetMaps adds significant size to the build size or client load, but I'm not confident I'm analyzing optimally, please let me know if there are other ways to check size/load/etc. or any suggestions or concerns.

mcbouslog:
I think the ImageToolbar should not be shown with the geoMap viewer. I think the geoMap viewer should look similar to the separate frames viewer (no ImageToolbar, move FieldGuide button below task area). I'll open an issue to document this potential code change (refactor SubjectViewerStore `separateFramesView` to something more generic?).

packages/app-project/pages/_app.jsx
shaunanoordin: To anyone else reading this: the reason `import 'ol/ol.css` is placed at the page-level instead of the component level (i.e. the import statement ain't in `GeoMapViewer.jsx`) is because Next.JS will throw a fit, with the error:

> `Global CSS cannot be imported from files other than your Custom <App>. Due to the Global nature of stylesheets, and to avoid conflicts, Please move all first-party global CSS imports to pages/_app.js. Or convert the import to Component-Level CSS (CSS Modules).`
> `Read more: https://nextjs.org/docs/messages/css-global`

The other alternative is to create a component called `GeoMapViewerCSS.jsx` that contains the CSS but as Mark pointed out, the tradeoff is that we'd need to manually sync that CSS with future OpenLayers updates.

❓  That said, I'm genuinely surprised that this worked on _this page_?? Because "ol" is **only** defined in _lib-classifier's package.json,_ yet somehow _app-project_ is able to resolve the "ol" package correctly?? pnpm, is this some weird magic you're doing?

packages/app-project/next.config.js
shaunanoordin: I'm fascinated by this experimental flag and I'm glad that you found it, though I too would prefer it if Next.JS actually added docs for it. This is one of those "small but insanely important build options" that we won't want to forget about in the future.

[documentation] I **strongly** suggest that you add an inline comment here that says something like `// Required for lib-classifier which uses ESM-only dependencies like OpenLayers aka "ol". See https://nextjs.org/docs/messages/import-esm-externals`

mcbouslog: Yes agreed! I added a code comment accordingly. 👍

packages/app-project/pages/_app.jsx
mcbouslog: That is a great point. I've added a comment in `_app.jsx` above the import and added OpenLayers (`ol`) as a dependency to app-project as well. It was working because the lib-classifier dependencies, notably `ol` in this case, were hoisted with app-project, but if `ol.css` is going to be imported into app-project's `_app.jsx` then `ol` should indeed be defined as a dependency in app-project's `package.json` as well. I don't think adding `ol` as a dependency to app-project has an impact on build size because I think modules are deduped and `ol` is ESM and Next.js uses tree-shaking to reduce `ol`'s impact on bundle size.

Issue #7113, opened 2025-11-13
cssi mapping nfn: Add ADR for geoMap drawing task and tools


mcbouslog:
_Issue documentation in progress_

Not Ready because:
- want to get proof of concept with #7104 and #7105  

Towards CSSI mapping efforts.

Document:
- `workflow.task` structure
- `annotation` structure (including `annotation.value`)
- ...more

PR #7118, opened 2025-11-24
Add ADR 63 for GeoMapViewer

### LINKED ISSUE #7112: cssi mapping nfn: Add ADR for GeoMapViewer
_Issue documentation in progress_

Towards CSSI Mapping efforts.

Add ADR documenting:
- mapping viewer requirements
- decision to use OpenLayers
- decision to default to OpenStreetMaps
- viewer named `GeoMapViewer` or `geoMap` as defined by `workflow.configuration.subject_viewer`

mcbouslog:
## Package
- lib-classifier

- closes #7112 

## Describe your changes
- adds ADR 63 for the GeoMapViewer
- there'll be subsequent PR for the mapping tasks and tools as per #7113 
- initial PR to add the GeoMapViewer component in #7111 

## Context
- linked to GitHub project https://github.com/orgs/zooniverse/projects/66
- towards [CSSI Mapping Infrastructure - Feature Requirements](https://docs.google.com/spreadsheets/d/1ifyRtdKmUqLBUoVcENiMQIsNn8InOM57IAcVBv3ubTQ/edit?usp=sharing)

### DIFF
--- a/docs/arch/adr-63.md
+++ b/docs/arch/adr-63.md
@@ -12,7 +12,7 @@ To support projects with complex geospatial datasets and provide a better volunt
 
 ## Decision
 
-Create a new `GeoMapViewer` subject viewer implemented with [OpenLayers](https://openlayers.org/) and a default OpenStreetMap (OSM) base tile layer. This viewer will:
+Create a new `GeoMapViewer` subject viewer implemented with [OpenLayers](https://openlayers.org/) and a default [OpenStreetMap](https://www.openstreetmap.org/) (OSM) base tile layer. This viewer will:
 
 - Be enabled via workflow configuration: `workflow.configuration.subject_viewer = 'geoMap'`
 - Utilize a default OSM base layer
@@ -20,7 +20,7 @@ Create a new `GeoMapViewer` subject viewer implemented with [OpenLayers](https:/
 
 ### OpenLayers mapping library
 
-After evaluating mapping libraries including Leaflet, Mapbox GL JS, and OpenLayers, we have chosen OpenLayers for the following reasons:
+After evaluating mapping libraries including Leaflet, Mapbox GL JS, and OpenLayers, we have chosen OpenLayers (package name `ol`) for the following reasons:
 
 - Open-source JavaScript library with permissive license
 - Active community and ongoing maintenance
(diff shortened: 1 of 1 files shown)


docs/arch/adr-63.md
```diff
@@ -0,0 +1,48 @@
...
+- Extensive features for handling vector and raster data
+
+### OpenStreetMap default base tile layer
+
+The OSM base tile layer provides a free, open-source, and widely used mapping layer that offers sufficient detail for most geospatial classification tasks. While OSM will be the default base layer, future work may include support for custom tile layers.
+
+## Consequences
+
+- A subsequent ADR will be created regarding geospatial annotation tasks and tools
+- This ADR will be updated or a subsequent ADR will be created for custom tile layer support
```
shaunanoordin, line 38: [optional] There are 2 additional consequences here that may be worth noting.

- OpenStreetMaps should probably be recognised or tagged as an _external dependency,_ i.e. we need it for this to work but we _don't have full control over it._
  - This would just be a theoretical understanding that if the OSM servers go down (or gets rate-limited, etc) then we can't do much.
  - In practice, we probably won't need to about this for the foreseeable future since OSM works fine as is, and in most cases we're already looking into _Zooniverse-hosted (I think?)_ custom tile layers. 👌 
- The addition of OpenLayers is making us update our build solution to favour ESModules.
  - For the moment I think this is fine as a footnote, though I can see the argument for making "we had to change our build solution to favour ESM" _into an ADR of its own,_ **if** that Next.JS `experimental.esmExternals` flag gets pulled in the future. 🤷

docs/arch/adr-63.md
```diff
@@ -0,0 +1,48 @@
...
+
+Create a new `GeoMapViewer` subject viewer implemented with [OpenLayers](https://openlayers.org/) and a default OpenStreetMap (OSM) base tile layer. This viewer will:
+
+- Be enabled via workflow configuration: `workflow.configuration.subject_viewer = 'geoMap'`
+- Utilize a default OSM base layer
+- Default to a minimal toolbar with panning and zooming capabilities
+
+### OpenLayers mapping library
+
+After evaluating mapping libraries including Leaflet, Mapbox GL JS, and OpenLayers, we have chosen OpenLayers for the following reasons:
```
shaunanoordin: [minor] I'm not 100% sure, but maybe it's worth explaining to anybody else that `ol` is the OpenLayers library? Like, if there's one thing I have gripes with, it's that OpenLayers didn't just name the library `openlayers`. 😬 [(Edit: THEY USED TO!?)](https://www.npmjs.com/package/openlayers)

docs/arch/adr-63.md
```diff
@@ -0,0 +1,48 @@
...
+
+Proposed
+
+## Context
+
+To support projects with complex geospatial datasets and provide a better volunteer experience we will create a mapping viewer within the classifier. The mapping viewer will support panning and zooming, and geospatial annotations (latitude and longitude coordinates).
+
+## Decision
+
+Create a new `GeoMapViewer` subject viewer implemented with [OpenLayers](https://openlayers.org/) and a default OpenStreetMap (OSM) base tile layer. This viewer will:
```
shaunanoordin: [minor] Linking to OpenLayers, that's good! 👍 We could also link to OSM: https://www.openstreetmap.org/

docs/arch/adr-63.md
mcbouslog: Yes, both great points - I've added additional lines regarding OSM dependency and ESM/`esmExternals`.

PR #7132, opened 2025-12-03
lib-classifier: Add GeoMapLayout for GeoMapViewer within classifier

### LINKED ISSUE #7116: cssi mapping nfn: Refactor layout for GeoMapViewer
Towards Feature Request 1, user story: "As a participant, I want to view a base map that is pre-set to a specific center view [based on lat/lon information from the data being ingested]."

Specifics
- for layout with GeoMapViewer, remove ImageToolbar, show FieldGuide button below task area
- refactor as necessary to show GeoMapViewer within VolumetricLayout
- if ^ works, rename VolumetricLayout to something more generic
- if doesn't work, create new layout, similar to VolumetricLayout

mcbouslog:
## Package
- lib-classifier

- closes #7116 

## Describe your changes
- adds new GeoMapLayout layout similar to volumetric layout
- refactors GeoMapViewer height per layout, with min-height
- OpenLayers Map component requires a defined height to render properly, these changes include a defined minimum height of 400px which gives a little space above and below on a mobile device to scroll (as scrolling/pinching within map will only change map view), otherwise height is defined by parent elements, or in this case the GeoMapLayout which...
- GeoMapLayout has a subject area height of 600px on medium and large screens, and a height of 400px on smaller screens
  - I think these are good starting heights, but I'm very much open to suggestion!
  - alternatively we could use a viewport height / `vh`; `80vh` looked ok, though maybe too much on large screens
  - the subject viewer height should certainly be revisited with design input at the noted design phase

## How to Review
  - lib-classifier: https://local.zooniverse.org:8080/?project=2030
  - app-project: https://local.zooniverse.org:3000/projects/markb-panoptes/geomap-test-project-staging/classify/workflow/3874?env=staging 
  - added story for layout: http://localhost:6006/?path=/story/layouts-geomap--default
  - related story for viewer: http://localhost:6006/?path=/story/subject-viewers-geomapviewer--default
- Are there plans for follow up PR’s to further fix this bug or develop this feature? yes! this might change significantly based on decisions about the task area, how the subject location data is shown, or design work, but for now I think it's a good starting point

# Checklist

## General

## General UX

## New Feature
  - Example of SlideTutorial [component](https://github.com/zooniverse/front-end-monorepo/blob/main/packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.js) with docgen comments, and its [story](https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/index.html?path=/docs/other-slidetutorial--default)

## Refactoring

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/README.md
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/README.md
@@ -21,3 +21,7 @@ The workflow must include a configuration for the GeoMapViewer:
 ### Subject
 
 TBD
+
+## Layout
+
+The GeoMapLayout is assigned to workflows with `configuration.subject_viewer` set as `geoMap`. The viewer container has a fixed height of 600px (400px on mobile). The task area is sticky positioned with a width of 20rem.


packages/lib-classifier/src/store/WorkflowStore/Workflow/Workflow.js
```diff
@@ -72,6 +72,9 @@ const Workflow = types
       if (self.configuration.layout) {
         return self.configuration.layout
       }
+      if (self.configuration.subject_viewer === 'geoMap') {
```
shaunanoordin, line 75: [documentation][moderate] If we're adding a new if-switch here that specifically looks at `workflow.configuration.subject_viewer`, it may be worth updating some of our READMEs.

- Definitely add something to the GeoMapViewer's README.
- The Layout Components's [README](https://github.com/zooniverse/front-end-monorepo/blob/6cdb6b/packages/lib-classifier/src/components/Classifier/components/Layout/README.md) would also need a few lines updated, to explicitly say "The GeoMapLayout is (blah blah), and it's activated automatically when (blah blah)"

The previous pattern/rule is to explicitly spell out custom layouts in `workflow.configuration.layout`, and if we're going to break that pattern/change that rule, we should do so consciously.[^1]

[^1]: ...but that said, I just read through the The Layout Components's README and I think we completely forgot to mention the implementation of `workflow.configration.layout` as part of the layout selection strategy. 😬

shaunanoordin (approved):
## PR Review

Package: lib-classifier
Part of: GeoMapViewer / Geo Mapping Projects

This PR adds a new layout (GeoMapLayout) to the Classifier, which is only activated when the GeoMapViewer is used.

- New GeoMapLayout added.
- `Workflow.(get)layout` modified with a new if-switch that _specifically_ checks 

Code read looks pretty straightforward, though I did identify some spots that will benefit from additional documentation, and documentation updates.

Testing looks mostly good, with some small caveats. (See Testing section)

### Dev Notes

❓ [random] Oh here's a long-term planning question I should've asked waaay earlier - do we ever plan to have mixed media Subjects? Like a multi-frame viewer, where the first frame is an image, and the second frame is a GeoMap?

Because I just realised that, if we do plan to do this, then the Subject Viewer selection logic and the Layout selection logic would need to be revised. 🤔 

But this is just a random question and not a blocker whatsoever, feel free to ignore.

### Testing

Tested on localhost with macOS + Chrome using app-project.

Testing Steps:
- Bootstrapped `app-project` and ran `pnpm dev`
- Open the Classify page of some projects that use the GeoMapViewer (i.e. `testWorkflow.configuration.subject_viewer = 'geoMap'`)
  - example 1:
  - example 2: 
- Observe that the OpenLayers-powered geographical map of the world is visible in the Subject Viewer (targeting no specific landmark, just lat=0 lon=0 at the centre and zoom=1)
- ✅ Confirm that new GeoMapLayout is being used (this can be confirmed by examining the div surrounding the GeoMapViewer component and ensuring it has height: 600px in its CSS)
- Confirm that the geographical map is displaying correctly and can be interacted with (pan & zoom) normally.
  - ✅ ...in desktop view.
  - ❓ ...in mobile view.

Results:
- ✅ The desktop view looks OK.
  - If I wanted to be nitty-gritty about it, I did notice that the OpenLayers map itself doesn't take up the full 600px height. I'm assuming this is because that 600px container also contains the space for the MetaTools and banner(s). (Screenshot: red border indicates 600px tall container. blue border indicates 558.2px tall OpenLayers map)
    Image
  - IF you meant that the OpenLayers map should be 600px tall, then you might want to change that up a bit so it's something like...
    ```
    <SubjectContainer>
      <Banners />
      <SubContainerWithHeight600px>
        <SubjectViewer />
      </SubContainerWithHeight600px>
      <MetaTools className='metatools-centered' />
    </SubjectContainer>
    ```
  - But if not, don't worry about it. 👌 
- The mobile view has a bit of an issue where the "Add to Favourites" and "Add to Collection" buttons start creeping onto (on top of) the map itself. 😬
  Image
  -  This happens due to the OpenLayers map fully taking up the 400px container height, as can be seen in the screenshot below with the debug borders enabled.  
    Image

### Status

PR looks good to me, actually - I've noted some things that need to be answered or addressed, but nothing's blocking this. 👍 The GeoMapViewer is a WIP, so I'm not too worried about making sure everything is perfect in one go. I'm happy to see improvements in later PRs.

Questions to answer:
- Question 1: in desktop view, did you mean to have the OpenLayers map be 600px?
  - If the answer is yes and you end up using that `<SubContainerWithHeight600px>` solution, that might also fix the issue with the mobile view.
- Question 2: just wanted to confirm that the intended layout selection logic for GeoMapLayout is that we're checking for 'geoMap' in workflow.configuration.subject_viewer and not workflow.configuration.layout, yes?[^1][^2]

Things to fix/do:
- [UI] Issue 1: mobile view's meta tools creep on top of the map. This shouldn't be the case. 
- [documentation] Docs required for GeoMapLayout selection logic.
  - GeoMapViewer's README needs to be updated to mention how having `workflow.configratuon.subject_viewer = 'geoMap'` also affects the layout.
  - Layout components' README needs to be updated with the overall layout selection strategy, but this may be its own PR. 😬 

OK, that's it from me!

[^1]: and by that, I mean, we want to tell project owners to "just set the custom workflow subject_viewer and don't worry about other settings, it's automatic!"
[^2]: although logically if you set `workflow.configuration.layout = 'geoMap'` you'd ALSO get the GeoMapLayout, but that gets confusing real quick.

Issue #7217, opened 2026-02-05
Volumetric Viewer: "undefined in SortedSet" crashing error


shaunanoordin:
## Crashing Issue

Follows: #7210's tests
Noted on PR 7210 _and_ on main (commit 837dadc)
Affects: Volumetric Viewer projects [(production example)](https://www.zooniverse.org/projects/cmcbrain/mind-mapper)

When a new Subject loads in the Volumetric Viewer, there's a random chance that the whole page _crashes_ with the error: `TypeError: Cannot read properties of undefined (reading 'data') at eval (SortedSet.js:76:1)`

<img width="1508" height="823" alt="image" src="https://github.com/user-attachments/assets/341794ce-8688-4929-a2c5-f17ef38c96b0" />

### Dev Notes

By inserting some debug code in `SortedSet.js`...

```
export const SortedSetUnion = ({ sets }) => {
  console.log('+++ sets: ', sets)
  const sortedSets = sets
    .map((s) => s.data)
    .sort((a, b) => a.length - b.length)
```

...I can see that _when_ the error triggers, we get the following in the dev console:

```
+++ sets:  (3) [undefined, undefined, undefined]
```

I believe that the sets should be 3x arrays that describe the X, Y, and Z planes (I think?) which makes it a puzzler as to why we're getting undefineds here.

Also note that I can trigger this problem on the live Zooniverse platform - https://www.zooniverse.org/projects/cmcbrain/mind-mapper/classify/workflow/28858?env=production&demo=true - so this isn't a quirk of React dev mode or anything.

### Testing

Replication steps:

- Load the Volumetric Viewer test project on localhost app-project, in demo mode: https://local.zooniverse.org:3000/projects/cmcbrain/mind-mapper/classify/workflow/28858?env=production&demo=true
- Repeatedly submit (empty) classifications. Check that next Subject loads correctly.
- Aim to submit about 10-20 empty classifications. (For reference, I'm able to trigger the error as soon as the first Subject (fails to) loads, or as late as the 11th Subject.

⚠️ NOTE: I suspect that the problem is related to the Subject, so it may be entirely possible that we'll be unable to replicate the problem once the workflow is retired.

Therefore, to follow up on this investigation, I'd suggest feeding the following Subjects to the Volumetric Viewer, and see if any of them trigger _consistently_ the error:

[...]

PR #7219, opened 2026-02-06
lib-subject-viewers: Fix SortedSet crash in VolumetricViewer

kieftrav:
## Package
lib-subject-viewers

Issue #7217 

## Describe your changes
Add defensive guards to prevent TypeError when SortedSet receives undefined sets, which can occur during component unmount or malformed subject data.

- Filter undefined/null sets in SortedSetUnion and SortedSetIntersection
- Add bounds checking in getPlaneSet to return empty SortedSet
- Guard renderPlanePoints against disposed viewer and null meshPlane refs
- Add unit tests for undefined set handling

## How to Review
- checkout main, make sure to run pnpm build in lib-subject-viewers and then in lib-classifier before running pnpm start in the lib-classifier (build introduces it's own issues with how it wraps all code).
- Visit classifier: https://local.zooniverse.org:8080/?project=cmcbrain%2Fmind-mapper&env=production&demo=true
- Click "Done" to see 25+ subjects with the dev console open. Analyze the console errors + monitor browser memory pressure + crashing.
- Compare with the PR branch

## General

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/lib-subject-viewers/src/VolumetricViewer/components/Cube.jsx
+++ b/packages/lib-subject-viewers/src/VolumetricViewer/components/Cube.jsx
@@ -261,6 +261,9 @@ export const Cube = ({ annotations, tool, viewer, orbitControlsEnabled = true })
   function renderPlanePoints () {
     // const t0 = performance.now()
 
+    // Guard against disposed or uninitialized viewer/component
+    if (!viewer || viewer.base <= 0 || !threeRef.current?.meshPlane) return
+
     const frames = viewer.planeFrameActive
     const sets = frames.map((frame, dimension) =>
       viewer.getPlaneSet({ dimension, frame })
@@ -268,9 +271,12 @@ export const Cube = ({ annotations, tool, viewer, orbitControlsEnabled = true })
 
     meshPlaneSet.current = SortedSetUnion({ sets })
 
+    const meshPlane = threeRef.current?.meshPlane
+    if (!meshPlane) return
+
     meshPlaneSet.current.data.forEach((point, index) => {
       drawMeshPoint({
-        mesh: threeRef.current.meshPlane,
+        mesh: meshPlane,
         meshPointIndex: index,
         point
       })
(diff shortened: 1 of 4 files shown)


shaunanoordin (approved):
## PR Review

Package: lib-subject-viewer
Affects: Volumetric Viewer
Fixes #7217 

This PR fixes an issue where the Volumetric Viewer would crash the entire page with an "undefined in SortedSet" error.

- **Note that this error appears to stem from malformed Subjects.**
- This PR fixes the crashing behaviour, it can't salvage the bad Subject data.

_Screenshot: the Volumetric Viewer displaying the "empty" representation of a bad Subject. Note how the page isn't crashing._

Image

Functional changes:
- Classifier no longer crashes when fed a bad VV Subject. 👍 
- NOTE: there is no indicator to volunteers that the Subject is bad/malformed, so we may need to help the research team communicate this on Talk or something.

Code read:
- The Volumetric Viewer's components & functions now have a more robust system of checks. It is paranoia central, _and it works._ 👌 

### Testing

Tested with app-project on localhost, with macOS + Chrome 144.

Preface:

- First, note the time of tests: 10 Feb 2026 at midnight GMT.
  - This is relevant because the error is Subject-dependent, and it's likely that future tests will not encounter the same results if the (live) workflow changes Subject Sets.
  - Some Subjects encountered in my tests: 113457183, 113457285, 113457298

Testing steps (baseline):

- On app-project, switch to `main` branch (commit c5d77fbca7e4a0a17bcfb46f10d6ab181726446b), then run _pnpm dev._
- Attempt to load the test project in demo mode: https://local.zooniverse.org:3000/projects/cmcbrain/mind-mapper/classify/workflow/28858?demo=true&env=production
- Observe that EVERY SINGLE SUBJECT now crashes the Classifier 😬  
  - When I was documenting issue 7217, at least SOME Subjects loaded properly. I can only assume that over the past few days, the "good" subjects were all retired and only the malformed ones remain to be served.

Testing steps (fix):

- On app-project, switch to this PR's branch, then run __pnpm dev._
- Load the test project in demo mode (same link.)
- Observe that the Classifier loads, albeit with an empty overall view, an empty y-plane, an empty z-plane, and a _blank_ x-plane. (See screenshot in intro, above.)
- Submit classifications. Observe that next (equally malformed) Subjects load into the Classifier without crashing anything.

Tests look good. 👍 

### Status

LGTM 👍 

This PR fixes the crashing bug from a code/front end point of view, but for the volunteer's sake, we may need to follow up with the research team to find out why there are bad Subjects, and how we can help.

PR #7223, opened 2026-02-12
lib-classifier: Refactor SingleImageCanvas with hasMounted

mcbouslog:
## Package
- lib-classifier

- Try to draw a freehand line at https://frontend.preview.zooniverse.org/projects/jenbays/beaver-dams/classify/workflow/31207?demo=true
- Note error
- Locally running app-project, I can only reproduce the bug with `pnpm build && pnpm start`, I can not reproduce the bug with `pnpm dev`

## Describe your changes
- in LineControls, add null check on canvas and default `{ width: 0, height: 0 }`

## How to Review
  - https://local.zooniverse.org:3000/projects/jenbays/beaver-dams/classify/workflow/31207?env=production&demo=true
  - try to draw a freehand line
- Which storybook stories should be reviewed? doesn't reproduce the bug, but http://localhost:6006/?path=/story/drawing-tools-freehand-line--drawing continues to work as expected

# Checklist

## General

## General UX

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/docs/arch/adr-64.md
+++ b/docs/arch/adr-64.md
@@ -26,23 +26,21 @@ The following is an example of a geoDrawing task annotation value with a geoPoin
 
 ```json
 {
-  "value": [
-    {
-      "type": "FeatureCollection",
-      "features": [
-        {
-          "type": "Feature",
-          "geometry": {
-            "type": "Point",
-            "coordinates": [2.2944810, 48.8583701]
-          },
-          "properties": {
-            "name": "Eiffel Tower"
-          }
+  "value": {
+    "type": "FeatureCollection",
+    "features": [
+      {
+        "type": "Feature",
+        "geometry": {
+          "type": "Point",
+          "coordinates": [2.2944810, 48.8583701]
+        },
+        "properties": {
+          "name": "Eiffel Tower"
         }
-      ]
-    }
-  ],
+      }
+    ]
+  },
   "task": "T0",
   "taskType": "geoDrawing",
   "tool": "geoPoint"
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/GeoMapViewer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/GeoMapViewer.jsx
@@ -1,6 +1,6 @@
 // dependencies
 import { Box } from 'grommet'
-import { arrayOf, shape, string } from 'prop-types'
+import { arrayOf, func, shape, string } from 'prop-types'
 import { useEffect, useRef } from 'react'
 import styled from 'styled-components'
 
@@ -13,12 +13,14 @@ import TileLayer from 'ol/layer/Tile'
 import VectorLayer from 'ol/layer/Vector'
 import OSM from 'ol/source/OSM'
 import VectorSource from 'ol/source/Vector'
+import { unByKey } from 'ol/Observable'
 
 // local imports
 import RecenterButton from './components/RecenterButton'
 import ResetButton from './components/ResetButton'
 import getFeatureStyle from './helpers/getFeatureStyle'
 import createModifyUncertaintyInteraction from './helpers/createModifyUncertaintyInteraction'
+import asMSTFeature from './helpers/asMSTFeature'
 
 const MapContainer = styled.div`
   height: 100%;
@@ -44,7 +46,9 @@ function fitViewToFeatures(map, features) {
 
 function GeoMapViewer({
   geoDrawingTask,
-  geoJSON = undefined
+  geoJSON = undefined,
+  onFeaturesChange = undefined,
+  onSelectedFeatureChange = undefined
 }) {
   // Map and layer refs: created once on mount, reused across feature updates
   const mapContainerRef = useRef()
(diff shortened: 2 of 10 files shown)


mcbouslog:
Thank you @sktw for noting the additional canvas/scale issue and suggested fix! I'll push related changes shortly and notify you for additional input.

mcbouslog:
This has been updated as suggested with `isMounted` and I think is looking good. A recap:

My understanding of the issue: when the SingleImageCanvas is rendered the first time, 
- `canvasLayer.current` is `null`, which sets SVGContext.canvas as `null`
- SVGContext.canvas as `null` causes the freehand line LineControls bug noted, and
- SVGContext.canvas as `null` results in `useScale` to return the default scale of `1`, which
- causes the noted rectangle scaling bug with a large subject image

In dev mode, which uses React StrictMode, SingleImageCanvas renders a second time. On second render the `canvasLayer` ref has mounted and is therefor defined when `canvasLayer.current` defines `SVGContext.canvas`. This results in none of the bugs noted above, as the freehand line LineControls have a defined `canvas` and `useScale` also has a defined `canvas` to compute the scale as expected.

The solution (thank you @sktw !): add `isMounted` state and a related effect that fires on mount. This effectively creates the second render so SVGContext.canvas is properly defined. The InteractionLayer (drawing fun) is prevented from rendering until isMounted is true to prevent improperly scaled drawing components from rendering.

Considerations: the drawing components are (slightly, I'd argue imperceptibly) delayed from rendering, and there's a related event handling gap, but I don't notice it and I think it's so small it's not a concern.

Links
- confirm bugs:
  - freehand line LineControls (select freehand line before any other action, try to draw): https://frontend.preview.zooniverse.org/projects/markb-panoptes/test-project-1-mb-fem-lab/classify/workflow/3846?env=staging&demo=true
  - improperly scaled rectangle: https://frontend.preview.zooniverse.org/projects/skeogh17/snail-scribe-archive/classify/workflow/29065?demo=true
- confirm fix:
  - freehand line LineControls (select freehand line before any other action, try to draw): https://fe-project-branch.preview.zooniverse.org/projects/projects/markb-panoptes/test-project-1-mb-fem-lab/classify/workflow/3846?env=staging&demo=true
  - improperly scaled rectangle, now properly scaled: https://fe-project-branch.preview.zooniverse.org/projects/skeogh17/snail-scribe-archive/classify/workflow/29065?demo=true
  
I've noticed a different freehand line LineControl bug I'll open an issue for.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.jsx
```diff
@@ -233,6 +233,7 @@ const LineControls = forwardRef(function LineControls({
               color={icon.color}
               size={`${icon.size}px`}
               focusable='false'
+              pointerEvents='none'
```
mcbouslog, line 236: The icons are capturing pointer events. To see this, 
1. go to https://fe-project-branch.preview.zooniverse.org/projects/projects/markb-panoptes/test-project-1-mb-fem-lab/classify/workflow/3846?env=staging&demo=true (which does not reflect this one line of code changes) and 
2. draw a freehand line
3. in the LineTools component, mouse directly over the garbage can icon, note the tooltip "Delete" disappears
4. click the garbage can icon, note yes/no delete prompt does not appear
5. move cursor few pixels away from garbage can icon, but still within delete button
6. click, note that the yes/no delete prompt does appear as expected

Adding `pointerEvents='none'` to the IconComponent here passes the pointer events up to the path/button.

mcbouslog:
In the SingleImageCanvas, I attempted something along the lines of:
```
const [canvasLayer, setCanvasLayer] = useState(null)
const canvasLayerRef = useCallback((node) => {
  setCanvasLayer(node)  // triggers a re-render
}, [])

<SVGContext.Provider value={{ canvas: canvasLayer, ... }}>
  <g ref={canvasLayerRef} ...>
```
But if I made a drawing annotation, left the workflow (i.e. to project home page), then returned to the workflow, the previously made drawing annotation rendered with incorrect scale, so I did not proceed accordingly to fix the issues noted.

mcbouslog:
> @mcbouslog You're welcome! I don't see any way around having the second render in `SingleImageCanvas`. To compute the scaling for the drag handles you need the rendered size of the canvas, which requires having the canvas element itself.
> 
> One issue does remain, although it involves an uncommon scenario:
> 
> 1. Navigate to https://fe-project-branch.preview.zooniverse.org/projects/skeogh17/snail-scribe-archive/classify/workflow/29065?demo=true
> 2. Resize the browser window, making it, say, roughly half the width of the screen.
> 3. Draw a rectangle
> 4. Maximize the browser window. Notice that the drag handles and delete button are now too large.
> 
> The problem is that the resize doesn't trigger a render, so the scaling is not updated for the new canvas size. This could be handled by adding a listener for the window's `resize` event, but, given the rarity of this scenario, it may not be worth it.

Yes, great catch @sktw ! I've opened #7231 to document and track accordingly.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageCanvas.jsx
```diff
@@ -23,6 +23,11 @@ function SingleImageCanvas({
   transformMatrix,
 }) {
   const canvasLayer = useRef()
+  const [isMounted, setIsMounted] = useState(false)
+
+  useEffect(() => {
+    setIsMounted(true)
+  }, [])
```
goplayoutside3: Not blocking, but just heads up we have a hook for this in LRC! https://github.com/zooniverse/front-end-monorepo/blob/main/packages/lib-react-components/src/hooks/useHasMounted.js

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.jsx
goplayoutside3: Thank you for fixing this UX. It's been a frustrating bug for awhile!

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageCanvas.jsx
mcbouslog: I didn't realize, thank you for noting! Refactored accordingly.

Issue #7222, opened 2026-02-11
Allow Hiding Marks toggle in PB: Unexpected behavior for marks on subjects with


lfortson:
## Package
FEM

## Describe the bug
showing up in project: https://www.zooniverse.org/lab/22358/workflows/29070

TL;DR - (1) when in the subject viewer, there seem to be a range of issues in marking on a subject when the “hide previous marks” toggle is on; (2) at the task editing level in PB, ticking the "allow hiding marks" does not seem to have a consistent effect in the subject viewer.

Fuller version - when the hide previous marks (HPM) toggle is on, it consistently hides them all just fine. But then, if I want to add a mark and click on the image anywhere, it (inconsistently) “reveals” some, but not all of the marks previously hidden - one at a time. But in one case, only the galaxy bulge and foreground stars (see screen shot). Then it got weirder, if I keep playing around, sometimes I click on the subject when HPM is on, and instead of showing a random mark that is supposed to be hidden, it will create a random mark that doesn’t even correspond to the “class” of the chosen marker  (see screen shot with large blue rectangle), but green marker was selected).

In addition, when I tick the “allow hiding marks” box at the top of the task “choices” section of the PB, it doesn’t seem to consistently apply whether I can hide/show previous marks in the subject viewer. 

## To Reproduce
1. Go to https://www.zooniverse.org/lab/22358/workflows/29070
2. Click on classify, go to a subject in subject viewer, click on "hide previous marks"
3. Select one of the task markers (e.g., "clumps")
4. Try to click on the subject in the subject viewer to make a new annotation
5. you may have to try a few times to induce the "bad" behavior

## Screenshots

<img width="1274" height="903" alt="Image" src="https://github.com/user-attachments/assets/9dc07193-6952-4837-bdbc-146683e25d2d" />

<img width="1285" height="953" alt="Image" src="https://github.com/user-attachments/assets/3cf85fd3-503e-4f40-8458-d4979df88fb9" />

<img width="1117" height="829" alt="Image" src="https://github.com/user-attachments/assets/a8f68fe1-5239-4b55-9cfa-360d539382b2" />

## Expected behavior
"Hidden" annotations should not randomly appear when clicking on subject viewer; a new mark with the correct task marker selected should be able to be made.

### Desktop (please complete the following information):
Replicated at minimum on Safari 26.2 and Chrome 144.0.7559.133

### Smartphone (please complete the following information):
Not tested

## Additional context
[...]

PR #7224, opened 2026-02-12
Drawing Tasks Previous Marks + Draw Bug

kieftrav:
## Package
`lib-classifier`

https://github.com/zooniverse/front-end-monorepo/issues/7222

## Describe your changes
When "Hide Previous Marks" is toggled, `togglePreviousMarks()` saved `hidingIndex = marks.length` as the visible/invisible cutoff. The filter `marks.slice(hidingIndex)` assumes new marks always append to the end of the flattened array... Yet the `marks` getter concatenates in **tool order** (`[tool0_marks, tool1_marks, tool2_marks, ...]`), so adding a mark to a non-last tool inserts it in the middle, shifting later marks' indices. This causes `slice(hidingIndex)` to show an old mark from a different tool instead of the newly drawn one.

Fix: Replace the positional `hidingIndex` (a single number) with `hiddenMarkIds` (an array of mark IDs captured at hide time). The filter becomes `marks.filter(m => !hiddenSet.has(m.id))`, which is immune to array reordering. This solution was already applied to `TranscriptionTask` and I'm applying it to `DrawingTask`.

## How to Review
Project: [Galaxy Zoo Clump Scout II, workflow 29070](https://www.zooniverse.org/projects/hughdickinson/galaxy-zoo-clump-scout-ii/classify/workflow/29070)

**Steps to reproduce the bug (before fix):**
1. Load the classifier with a subject that has Caesar machine predictions (multi-tool drawing task with 2+ tools)
2. Click "Hide Previous Marks" — all marks disappear
3. Draw a rectangle
4. Click "Show Previous Marks", then "Hide Previous Marks" again
5. Switch to a different drawing tool (e.g. "Foreground stars" / yellow)
6. Draw a rectangle — **Bug:** the visible mark is the wrong color (an old prediction from a different tool leaks through instead of your new mark)

**After fix:** Step 6 correctly shows the new mark in the selected tool's color.

**Local dev URL:** https://local.zooniverse.org:8080/?project=hughdickinson%2Fgalaxy-zoo-clump-scout-ii&workflow=29070&subject=110778508&env=production
** FE Branch** https://fe-project-branch.preview.zooniverse.org/projects/hughdickinson/galaxy-zoo-clump-scout-ii/classify/workflow/29070

## General

## Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/lib-classifier/src/plugins/tasks/drawing/models/DrawingTask.js
+++ b/packages/lib-classifier/src/plugins/tasks/drawing/models/DrawingTask.js
@@ -17,7 +17,8 @@ export const Drawing = types.model('Drawing', {
   activeToolIndex: types.optional(types.number, 0),
   annotation: types.safeReference(DrawingAnnotation),
   shownMarks: types.optional(types.enumeration(Object.keys(SHOWN_MARKS)), SHOWN_MARKS.ALL),
-  hiddenMarkIds: types.optional(types.array(types.string), []),
+  // This array contains the ID of every mark when the "Show Previous Marks" option is toggled. This will then be used to filter out which marks should be shown or hidden on the Subject Viewer
+  hiddenMarkIds: types.optional(types.array(types.string), []), 
   tools: types.array(GenericTool),
   type: types.literal('drawing')
 })


packages/lib-classifier/src/plugins/tasks/drawing/models/DrawingTask.js
```diff
@@ -99,7 +99,7 @@ export const Drawing = types.model('Drawing', {
         self.activeToolIndex = 0
         self.subTaskVisibility = false
         if (self.shownMarks === SHOWN_MARKS.NONE) {
-          self.hidingIndex = 0
+          self.hiddenMarkIds.clear()
```
shaunanoordin, line 103: [ramblings] At first I was confused that a JavaScript array had a `.clear()` function, but then I realised this was a MobX State Tree _observable array,_ which indeed has a [.clear()](https://mobx.js.org/api.html#observablearray) function. Derp!

packages/lib-classifier/src/plugins/tasks/drawing/models/DrawingTask.js
```diff
@@ -17,7 +17,7 @@ export const Drawing = types.model('Drawing', {
   activeToolIndex: types.optional(types.number, 0),
   annotation: types.safeReference(DrawingAnnotation),
   shownMarks: types.optional(types.enumeration(Object.keys(SHOWN_MARKS)), SHOWN_MARKS.ALL),
-  hidingIndex: types.maybeNull(types.number),
+  hiddenMarkIds: types.optional(types.array(types.string), []),
```
shaunanoordin: [documentation] I'd recommend adding a small comment here that says something like...

```
// This array contains the ID of every mark when the "Show Previous Marks" option is toggled. This will then be used to filter out which marks should be shown or hidden on the Subject Viewer.
```

goplayoutside3:
Reminder to add the exact phrase "Closes https://github.com/zooniverse/front-end-monorepo/issues/7222" to your PR description and the linked Issue will auto-close. Manually closing it now. Thanks for the bug fixing!

Issue #7203, opened 2026-01-28
Admin page link in ProjectHeader requires refreshing the page


goplayoutside3:
## Package
- app-project

## Describe the bug
The ProjectHeader component should display a link to a project's "admin page" when admin mode is on. This sometimes works, but often requires refreshing the whole page after admin mode is toggled.

## To Reproduce
Steps to reproduce the behavior shown in the video below:
1. Go to a Zooniverse project such as Gravity Spy.
2. Toggle admin mode on.
3. The ProjectHeader component links do not change.
4. Refresh the page. Now there's an admin mode link in ProjectHeader.
5. The same can be done in reverse (toggle admin mode off, the link doesn't disappear, refresh).

## Screenshots

https://github.com/user-attachments/assets/41d64085-14e4-4e6d-b31d-bcf8e0c0b36d

## Expected behavior
ProjectHeader should respond to changes in admin mode.

goplayoutside3:
Closed by https://github.com/zooniverse/front-end-monorepo/pull/7233

PR #7233, opened 2026-02-20
Admin Mode Simplification

kieftrav:
## Package
`app-project`

#7203 

## Describe your changes
- Move localStorage read/write into User store actions with a MobX reaction on isLoaded/isAdmin (follows UI.js afterAttach pattern)
- Add toggleAdminMode() store action (eliminates external _adminMode access)
- Strip useAdminMode hook to thin store wrapper (no useEffect/useState)
- Remove redundant adminMode prop-drilling through ProjectHeader/Nav/DropdownNav (these components already read the store via useStores)
- AdminContainer reads store directly instead of receiving props
- Update Storybook stories to set admin mode via store.user.setAdminMode()
- Add 10 unit tests for admin mode store behavior
- Props retained at cross-package boundaries where MobX Provider is unavailable (PageHeader=>ZooHeader, ClassifierWrapper=>Classifier)

## How to Review
Watch the video in the issue and test *that* on `main` vs this branch :)

# Checklist

## General

## Bug Fix
- [ ] The bug is fixed

## Refactoring


mcbouslog:
My apologies for the much delayed review! While these changes are great and effective, and the following is certainly a little hypocritical as I've used mst store with recent mapping efforts - I think if possible we're trying to use the mst stores less and lean on custom hooks and context when possible. In this case, app-root has a custom hook - [app-root's useAdminMode](https://github.com/zooniverse/front-end-monorepo/blob/main/packages/app-root/src/hooks/useAdminMode.js) - that I think could also be applied to app-project to close the linked issue? I'm not sure how urgent a fix is compared to preference for not using mst stores though.

goplayoutside3:
I want to add on to Mark's comment to clarify a bit. Also note that this FEM feature is lower priority than CSSI Mapping dev work, so please feel free to leave this PR as is, and pause review until time is available.

During FEM's development over the past few years, _particularly in app-project_, data fetching such as checking for a signed-in user has been moved to React hooks rather than relying on the MST store for component context. I don't have explicit documentation to point to without spelunking through years old PR comments, but roughly the reasoning behind moving _data fetching_ away from the MST store in app-project is because MST is a client-side strategy and data fetching functions should be removed from it to utilize Next.js architecture instead. In the longterm, serverside handling of a signed-in user is ideal, but can't be accomplished right now because it requires a dedicated Zooniverse architecture-level dev plan/work.

By contrast, lib-classifier is architected to be a client-side library. That's where you did your recent CSSI Mapping dev work @mcbouslog, and it is by design that the Classifier is a client-side state machine. It's not hypocritical to mention leaning away from the MST store in app-project here because app-project and lib-classifier use very different data management strategies. One is an app, and one is a React component.

In this PR, adding data fetching to the MST store would be a red flag since we'd just have to rip it out in a future refactor (hopefully once user data fetching is serverside someday). The only major modifications I see in this PR is that handing of `localStorage` is moved from a React hook to the User store model. As Mark mentioned, I think I would have reached for the `useAdminMode()` strategy in app-root as an example where it's setup as global context for every page. app-root is using a Next.js App Router, which looks a bit different from app-project's Next.js Pages Router, but eventually we'd like to use the same app-level strategies in both. Something to think about. 

Since this PR doesn't move actual data fetching into the MST store I'm ok with the strategy here as long as the UX works as expected. @kieftrav I'll just echo again that this FEM feature is lower priority than CSSI Mapping dev work, so please feel free to leave this PR as is, and pause review until time is available.

kieftrav:
@mcbouslog - I've incorporated your feedback and was able to cut out a decent chunk of code by creating a ContextProvider. Ready for re-review whenever you have time

mcbouslog (approved):
Changes look great, thank you! 

I've confirmed admin toggles as expected locally ✅

PR #7236, opened 2026-02-22
fix(a11y): screen-reader compatible survey task questions

### LINKED ISSUE #6947: a11y: survey task choice questions use invalid ARIA and HTML
https://github.com/zooniverse/front-end-monorepo/blob/186c97f4d26ac5ee303fda685a7eb3021d9d42c7/packages/lib-classifier/src/plugins/tasks/survey/components/components/Choice/components/Questions/Questions.js#L43-L55

Survey task choice questions use a h2 heading here, along with `aria-labelledby`, to add a question to a group of either radio buttons or checkboxes. This has a couple of problems:
- a screen reader like JAWS, in forms mode, will simply ignore the h2 and only read the checkboxes or radios. So you might hear "checkbox: 1, checkbox 2, checkbox: 3" etc. without knowing what the numbers refer to (How many animals?)
- HTML elements with the 'generic' role, like `<div>`, [cannot be labelled](https://benmyers.dev/blog/dont-use-aria-label-on-static-text-elements/) so, again, JAWS and NVDA will ignore `aria-labelledby` here. Desktop VoiceOver has a bug where it ignores the HTML spec and reads ARIA labels for any element, so you can’t test this with VoiceOver on a Mac.
>Because the generic role is nameless, the [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby) and [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label) attributes are prohibited. Because the role is generic, the [aria-roledescription](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-roledescription) and [aria-brailleroledescription](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-brailleroledescription) attributes are also prohibited.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/generic_role

There's an existing pattern for labelling groups of inputs in HTML: wrap them in a [`fieldset`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/fieldset) with a [`legend`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/legend). Here the legend should be the choice question. Both JAWS and NVDA will read the legend alongside each input label: "How many animals? 1, How many animals? 2, etc."

eatyourgreens:
Remove the invalid labelled divs from survey task questions. Replace them with a semantic, valid fieldset and legend for each checkbox/radio input group.

## Package
- lib-classifier

- fixes #6947.

## How to Review
You can review this in the classifier storybook. Visually, the survey task shouldn't have changed. In the accessibility tree, survey task questions should be correctly exposed to screen readers as labelled groups, where the group name is the question and the grouped items are the list of answers. I've added a test to check this too.

Screenshot of a mock survey task from the storybook, with the accessibility panel open in dev tools to show that a question is a labelled group in the accessibility tree, with labelled inputs as children.

Previously, the questions were added as labels to divs. Divs have [the generic role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/generic_role), thus cannot have ARIA labels. So the ARIA labels were ignored both in the accessibility tree and by screen readers.

## WCAG criteria to test

- [ ] [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html) (level A.)
- [ ] [3.3.2 Labels or Instructions](https://www.w3.org/WAI/WCAG22/Understanding/labels-or-instructions) (level A.)

## Checklist

### General

### General UX

### Bug Fix
- [x] The bug is fixed


eatyourgreens:
The tasks that render lists of radio buttons should group those inside fieldsets. 

Dropdowns aren't grouped, so don't need a fieldset. They do need an accessible name, in the accessibility tree, which they currently don't seem to have.

mcbouslog (approved):
This is great, thank you for the fix!

PR #7254, opened 2026-03-03
lib-react-components: Refactor Media for GeoJSON media preview

mcbouslog:
## Package
- lib-react-components

## Linked Issue and/or Talk Post
- closes #7110 

## Describe your changes
- add GeoJSON to JSONData types
- add Media story for GeoJSON subject media
- refactor useJSONData similar to lib-classifier useSubjectJSON
- refactor Data viewers with jsonData property (instead of deconstructed `{ data, chartOptions }`)
- add JSONViewer, with shared PreDisplay component from Text viewer

## How to Review
_Helpful explanations that will make your reviewer happy:_
- What Zooniverse project should my reviewer use to review UX?
  - https://local.zooniverse.org:3000/projects/markb-panoptes/geomap-test-project-staging/classify/workflow/3882?env=staging&demo=true
- What user actions should my reviewer step through to review this PR?
  - make a classification
  - note the Recent Classifications subject preview card below the classifier
- Which storybook stories should be reviewed?
  - http://localhost:6007/?path=/story/components-media--geo-json-data
  - other Data stories still work as expected
- Are there plans for follow up PR’s to further fix this bug or develop this feature? yes, per design review

# Checklist
_PR Creator - Please cater the checklist to fit the review needed for your code changes._
_PR Reviewer - Use the checklist during your review. Each point should be checkmarked or discussed before PR approval._

## General
- [ ] Tests are passing locally and on Github
- [ ] Documentation is up to date and changelog has been updated if appropriate
- [ ] You can `pnpm panic && pnpm bootstrap`
- [ ] FEM works in all major desktop browsers: Firefox, Chrome, Edge, Safari (Use Browserstack account as needed)
- [ ] FEM works in a mobile browser

## General UX
Example Staging Project: [i-fancy-cats](https://local.zooniverse.org:3000/projects/brooke/i-fancy-cats)
- [ ] All pages of a FEM project load: Home Page, Classify Page, and About Pages
- [ ] Can submit a classification
- [ ] Can sign-in and sign-out
- [ ] The component is accessible
  - Can be used with a screen reader [BBC guide to VoiceOver](https://bbc.github.io/accessibility-news-and-you/assistive-technology/testing-steps/voiceover-mac.html)
  - Can be used from the keyboard [WebAIM guide to keyboard testing](https://webaim.org/techniques/keyboard/#testing)
  - It is passing accessibility checks in the storybook

## New Feature
- [ ] The PR creator has listed user actions to use when testing the new feature
[...]

PR #7285, opened 2026-03-26
Add ADR 65 for GeoJSON subject

mcbouslog:
## Package
- lib-classifier

## Describe your changes
- add ADR 65 for GeoJSON subject

## How to Review

# Checklist

## General

### DIFF
--- a/docs/arch/adr-65.md
+++ b/docs/arch/adr-65.md
@@ -10,7 +10,7 @@ Proposed
 
 Since [ADR 28](adr-28.md), we support multiple JSON-based subject viewers. [ADR 48](adr-48.md) introduced a shared `JSONData` model and `JSONDataViewer` approach so JSON subject locations can be loaded and routed to the appropriate viewer.
 
-For geographic annotation projects, [ADR 64](adr-64.md) proposes a new `geoDrawing` task with annotation values stored as GeoJSON. We need a subject viewer that can render GeoJSON subject data directly from a location with MIME type `application/json`.
+For geographic annotation projects, [ADR 64](adr-64.md) proposes a new `geoDrawing` task with annotation values stored as GeoJSON. We need a subject viewer that can render GeoJSON subject data from a location with MIME type `application/json`.
 
 ## Decision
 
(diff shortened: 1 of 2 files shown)


docs/arch/adr-65.md
```diff
@@ -0,0 +1,56 @@
...
+    "locality": "City of Boulder Mountain Parks; Shanahan Ridge: base of Flatirons.",
+    "stateprovince": "Colorado"
+  }
+}
+```
+
+## Consequences
+
+- Existing JSON viewer behavior remains unchanged for non-GeoJSON payloads.
+- Projects can upload GeoJSON subject data as `application/json` and have it rendered by the GeoJSON viewer path.
```
mcbouslog, line 55: This will require a PR to refactor the existing chart viewers from props of `data` and `chartOptions` to the raw JSON data (PH TESS viewer has prop of the raw data). I think I'll open a related issue unless there's a suggestion otherwise.

goplayoutside3: Does handling GeoJSON subjects have a side effect on live projects like PH TESS? I think I'm a little confused by the wording of your highlighted bullet point in the ADR. Please explain more about why further dev work is needed beyond what you've already developed for the CSSI NfN project in order to "upload GeoJSON subject data as `application/json` and have it rendered by the GeoJSON viewer path".

docs/arch/adr-65.md
```diff
@@ -0,0 +1,56 @@
+# ADR 65: GeoJSON Subject Viewer
```
goplayoutside3, line 1:
```suggestion
# ADR 65: GeoJSON Subject
```

I think using the word "viewer" is too narrow for this ADR. "Viewer" implies classifier-only code (for now), so consider broadening the scope of this ADR to give an overview to someone who is new to the structure of a GeoJSON Zooniverse subject.

goplayoutside3 (commented):
Reading through this ADR, are you aiming to document the structure of a GeoJSON subject generally? Or how a GeoJSON subject is handled specifically in the classifier?

When you and I chatted in-office about adding an additional ADR for the subject type, GeoJSON, I was leaning toward broader documentation about it being a "json" subject, yet json subjects sort of have subtypes, so for example, GeoJSON subjects are handled differently than volumetric subjects or x,y point subjects. Your section of classifier behavior is a good starting point, but I recommend adding more, basically describing why you had to do https://github.com/zooniverse/front-end-monorepo/pull/7254.

The ADR outline could look like this:
1. Context ✅ 
2. Decision
    a. Structure of a GeoJSON subject (what its Javascript object looks like and a simple example). Briefly explain about `FeatureCollection` here, and this is where to explain what `reference_data` is too.
    b. How the classifier detects a GeoJSON subject. Note the subject is displayed in the `GeoMapViewer`. Further map viewer details aren't needed because ADR 63 elaborates on the classifier's subject viewer.
    c. How the Media component in LRC servers as a subject preview. Reference https://github.com/zooniverse/front-end-monorepo/pull/7254.
3. Consequences - see my comment in this section in response to your mention of PH TESS.

Let me know your thoughts on that example outline! There might be repetitive information in a, b, and c of the Decision section, but consider someone who's new to the GeoJSON subject type and learning about how FEM handles it big picture.

docs/arch/adr-65.md
mcbouslog: Currently, the the workflow configuration property `subject_viewer: "geoMap"` will also set `subject.viewer: "geoMap"` per the [Subject store](https://github.com/zooniverse/front-end-monorepo/blob/main/packages/lib-classifier/src/store/subjects/Subject/Subject.js#L55-L58), which allows the GeoMapViewer to load. So the admin toggle of the workflow configuration to "geoMap" is required for the GeoMapViewer to load.

If a subject has a location of valid GeoJSON but NOT the admin toggled workflow config to "geoMap," then the subject will be loaded with the [JSONDataViewer](https://github.com/zooniverse/front-end-monorepo/blob/main/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/JSONDataViewer/JSONDataViewer.jsx) and the GeoMapViewer will not be shown.

I think we want the GeoMapViewer to remain behind the necessary admin toggle for the short term? Once we're ready to open the GeoMapViewer to all, the JSONDataViewer can be refactored to add the GeoMapViewer similar to the other JSON viewers (BarChart, ScatterPlot, LightCurve, VariableStar). Currently the other JSON viewers (except LightCurve) destructure the JSON data for `chartOptions` and `data`, which can be moved from the parent JSONDataViewer component to the individual viewers, similar to [this PR for the Media component](https://github.com/zooniverse/front-end-monorepo/pull/7254/changes#diff-52898e16444613a47971d974c72f3dba70e57aedb469430c3a570077e3ce58cd).

goplayoutside3: Thank you that's a really good clarification. Please copy this text directly into the Consequences section of the ADR 👍

PR #7291, opened 2026-03-30
lib-classifier: Relocate LineControls outside of interaction layer

### LINKED ISSUE #7230: LineControls aren't scaled as expected
## Package
- lib-classifier

## Describe the bug
LineControls aren't scaled as expected with large (and presumably small) subject images.

## To Reproduce
1. Go to https://fe-project-branch.preview.zooniverse.org/projects/markb-panoptes/test-project-1-mb-fem-lab/classify/workflow/3847?env=staging&demo=true
2. Click on the freehand line drawing tool
3. Draw a freehand line
4. Note the LineControls rendered very small in the upper left corner of the subject viewer

## Expected behavior
The LineControls are scaled to a size as expected, similar to other drawing components, but remain positioned in the upper left (or if adjusted to upper right) corner of the subject viewer and do not adjust with the image as other drawing components do.

sktw:
Refactor the `LineControls` component so that it can be rendered outside of the interaction layer. This eliminates the need to apply inverse transformations to keep the controls static in the subject viewer.

## Package

- lib-classifier

Closes #7230

## Describe your changes

- Move mark deletion logic into a new `deleteMark` action on the `DrawingTask` model, and refactor `InteractionLayer`, `DrawingToolMarks` and `SubTaskPopup` to call it
- Remove transformations from the `LineControls` component, and add new `LineControlsContainer` to connect to the store
- Add new `ControlsLayer` component to logically contain the line controls, and update `SingleImageViewer` and `SingleVideoViewer` to render it
- Add tests and stories

### Workflow

- Draw a freehand line. Check that line controls render and that all buttons function correctly. Confirm that the controls remain static when panning, zooming and rotating the image.
- Draw a large cyan point. Confirm that the mark can be deleted using the delete button.
- Draw a small red point. Close the subtask popup without saving and confirm that the mark is deleted.
- Check that Hide / Show previous marks operates correctly.
- Resize browser window, and check that the line controls inset reduces to 15px below breakpoint.  

### Multi-image workflow

https://localhost:8080/?project=goplayoutside3%2Ffem-flipbook-viewer&workflow=24066&env=production

- Check that the line controls only display in the frame containing the active mark.
- If possible, also test in a workflow that has `multiImageCloneMarkers = true`, and confirm that line controls display in all frames.

### Storybook

- http://localhost:6006/?path=/story/drawing-tools-linecontrols--default
- http://localhost:6006/?path=/story/drawing-tools-freehand-line--drawing
- http://localhost:6006/?path=/story/subject-viewers-singlevideoviewer--with-drawing&args=toolType:freehandLine

# Checklist

## General
- [x] You can `pnpm panic && pnpm bootstrap`

## General UX

## Bug Fix
- [x] The bug is fixed

## Refactoring


packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.jsx
```diff
@@ -1,24 +1,35 @@
...
+const DELETE_CANCEL_COLOR = '#E45950'
+const DELETE_CONFIRM_COLOR = '#1ED359'
+const STROKE_WIDTH = 0.5
+const OUTER_RADIUS = 40
+const INNER_RADIUS = 20
+const CENTER_X = OUTER_RADIUS + 1
+const CENTER_Y = OUTER_RADIUS + 1
+const WIDTH = 2 * CENTER_X
+const HEIGHT = 2 * CENTER_Y
+const INSET = '30px'
```
goplayoutside3: Although Zooniverse’s freehand line isn’t the best experience on a small touchscreen, I think on `small` screens (<769px in our Grommet theme), the LineControls should sit closer to the edge of the subject viewer. How about 15px inset for `small` screens and 30px inset for all others? For example, https://local.zooniverse.org:3000/projects/eatyourgreens/-project-testing-ground/classify/workflow/3370 run locally via app-project on an iPhone, or the FreehandLine tool Storybook simulated on a mobile phone.

Referencing the example workflow in my review comment about separate frames for a multi image subject, maybe the inset could be defined by the width of its container parent, ControlsLayer. When a multi-image subject is displayed as a grid of separate frames, the images are often <300px wide. A CSS container query would satisfy small mobile screens and the separate frames grid. Example workflow: https://www.zooniverse.org/projects/goplayoutside3/fem-flipbook-viewer/classify/workflow/24066

goplayoutside3 (commented):
Thank you so much for the PR description with example Zooniverse workflows to test. The in-code comments, specs, and Storybook stories you’ve included are helpful, too!

Things are looking good for refactoring LineControls outside of the InteractionLayer component. There's one change in behavior I want to point out though, using this example workflow: https://localhost:8080/?project=goplayoutside3%2Ffem-flipbook-viewer&workflow=24066&env=production
- When a subject has multiple images, it’s shown in the flipbook viewer. The flipbook viewer also has a “separate frames” mode, which can be toggled on by default in the project builder. 
- With the LineControls no longer a child of the InteractionLayer, the LineControls component has no knowledge of the current “frame” of a multi image subject. My example workflow is configured to show the separate frames first, and drawn marks are _not_ cloned across frames. 
- Unexpected behavior on this PR branch = Draw a freehand line on one frame and the LineControls show on every frame, instead of only on the frame of the active `mark`. This is perhaps a very specific case of drawn marks on multi-image subjects and not cloned across frames, but I think best to address the UX now while touching LineControls.

Overall, please let me know if you prefer to make code changes to address my comments and notify for a second review, or if you prefer I take things from here in terms of code changes before merging. Either way, we really appreciate your work toward the LineControls enhancement! Simplifying its component code is very helpful.

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.jsx
sktw: Just to check - do we want to add a single css container query `@container (max-width: 768px) { /* set inset to 15px */}` to handle both small screens and small images?

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/InteractionLayer/InteractionLayerContainer.jsx
sktw, line 77: Apologies - I completely missed this bit of code! So for the LineControls we need to pass the frame down in props, and conditionally render the controls if the active mark isn't hidden and is in the current frame (unless `multiImageCloneMarkers` is `true`).

packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.jsx
goplayoutside3: I'm not super familiar with container queries because I've only used it as a feature a couple of times. My first instinct is yes there should be a single css container query to handle the LineControls inset. However, I think that means both the VideoWithDrawing component and SingleImageViewer component become the "container". In that case, we're not limited the Grommet `small` screen breakpoint of 768px viewport-width, and instead listen only for the container's width. The value doesn't have to be 768px.

I'll let you experiment with the container width that makes the most sense for adjusting `INSET` in LineControls, and please let me know if you'd like me to branch off of this PR to show an example!

sktw: For size queries the container is determined by setting a `container-type` property (and optionally a `container-name`) on an ancestor element. In this case, suitable containers would be the relatively positioned divs in SingleImageViewer and VideoWithDrawing.

Reducing the inset to 15px doesn't really change the ux much, but looks better for smaller images. Perhaps a breakpoint of 480px is about right. It would mean that in the three-column grid layout on an FHD screen the inset is set at the smaller value.

One other thought - currently, the line controls are displayed as soon as the user starts to draw and as long as the mark is active. An alternative would be to only display the controls when the user isn't drawing. This would be beneficial for smaller screens / images, where it's more likely that the controls will get in the way (although I'm not suggesting making this change size-dependent). The `FreehandLine` model already has an `isDragging` property that is true when the user is drawing, so the implementation would be straightforward.

goplayoutside3: Thanks for thinking through UX options for smaller images. A container size breakpoint of 480px sounds great.

If changing the show / hide behavior of the LineControls based on `isDragging`, I suggest branching that change into a separate PR. I'd want to test `isDragging` behavior on a few different touchscreen devices, and an isolated PR would make it easy to demonstrate the change to a few of the stakeholders on the Zooniverse team connected to the freehand line tool development.

Happy to review that PR if you choose to do so, but also no pressure because this PR is already a big enhancement for the LineControls!

sktw:
This is now ready for review, with the following changes:

- Add multi-frame and hidden marks handling
- Add variable inset to line controls
- Update tests and stories

I'll submit a separate PR with the `isDragging` changes, as suggested.

What is the current situation with regards to linting? Both `pnpm run lint` and `pnpm run eslint` seem to return errors in multiple files when run on main.

goplayoutside3:
Perfect, thank you for the update! No need to worry about linting commands. The `tools-standard` package was created in this codebase a long time ago, but never fully implemented so the current team doesn't enforce linting commands for code review, though long-term it's definitely something we should consider.

I plan to review by EOD Monday.

goplayoutside3 (approved):
This looks great. Thanks for changing the LineControls inset and considering the multi-frame subjects. I'll get this merged and we'll deploy it to production this week!

Issue #7076, opened 2025-10-22
Tweaks to processing of MachineLearnt reductions from Caesar


lcjohnso:
## Package
- lib-classifier

## Context and Broad Description
The result of PR #6039 was to facilitate the import of drawing annotations for multiple drawing tools, beyond an initial implementation of the freehand line tool. In the course of documenting what keys need to be included in the Caesar-hosted annotation for FEM to correctly render this expanded set of annotations (see https://github.com/zooniverse/caesar/pull/1612), a number of issues arose that could benefit from code edits and clarifications.

## Suggested Updates
1. Make clear what keys must be included in the reduction's `data` field. In addition to the tool/mark-specific keys (e.g., `x` and `y` for point marks), three additional keys are required for workflow/task processing purposes: `stepKey,` `taskIndex`, `toolIndex`.  This calls for a tweak to [useMachineLearntReductions.js#L36](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/hooks/useMachineLearntReductions.js#L36).
2. The `taskIndex` key is not currently used but should be -- in [useMachineLearntReductions.js#L41](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/hooks/useMachineLearntReductions.js#L41), `tasks[0]` can be changed to `tasks[newMark.taskIndex]`.
3. Update tests to enforce the smaller required set of annotation keys.

## To Reproduce
Please see this staging workflow where a new reduced set of annotation keys was used for a point mark: 
Lab = https://www.zooniverse.org/lab/2007/workflows/3789?env=staging
Test Subject = https://www.zooniverse.org/lab/2007/subject-sets/5155?env=staging
Caesar Subject View = https://caesar-staging.zooniverse.org/workflows/3789/subjects/253445

attn: @kieftrav

kieftrav:
Thanks @lcjohnso - PR #7319, when merged, will incorporate all of these suggestions.

PR #7319, opened 2026-04-14
Finalize MachineLearnt reductions

kieftrav:
## Package
- lib-classifier

- https://github.com/zooniverse/front-end-monorepo/issues/7076
- https://github.com/zooniverse/caesar/pull/1612

## Describe your changes
Addresses all three suggested updates from #7076:

1. **Use `taskIndex` from Caesar data instead of hardcoded `tasks[0]`** ([useMachineLearntReductions.js#L36-L42](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/hooks/useMachineLearntReductions.js#L36-L42)): The hook was destructuring `taskIndex` out of the Caesar mark but never using it - marks always routed to `tasks[0]`. Now uses `tasks[taskIndex]` for correct multi-task routing. `toolIndex` is also properly extracted before being passed to `createMark`.

2. **Default `stepKey`, `taskIndex`, and `toolIndex` when absent** ([MachineLearntReductions.js](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js)): As noted in [Cliff's comment](https://github.com/zooniverse/front-end-monorepo/issues/7076#issuecomment-3434484398), many correct-a-machine workflows have only one step, one task, and one tool. Rather than forcing every Caesar reduction to include `stepKey`, `taskIndex`, and `toolIndex`, these now default to `'S0'`, `0`, and `0` respectively when absent. This reduces the required keys in Caesar CSV uploads to just the tool-specific keys (e.g., `x`, `y` for points) plus `markId`, `toolType`, and `frame`.

3. **Updated tests** ([MachineLearntReductions.spec.js](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.spec.js)): Added test suites for multi-index routing (marks on different steps/tasks/tools) and default value behavior (marks with missing routing keys).

## How to Review
**Test project (staging):** Load the classifier with Cliff's staging point-mark workflow:
```
https://local.zooniverse.org:8080/?project=2007&workflow=3789&env=staging
```
Caesar Subject View: https://caesar-staging.zooniverse.org/workflows/3789/subjects/253445

**Test project (production, multi-tool):**
```
https://local.zooniverse.org:8080/?project=22358&env=production
```

**Steps:**
1. Load either project above — Caesar reduction marks should render on the subject
2. Verify marks appear on the correct tool
3. Run unit tests: `cd packages/lib-classifier && pnpm test -- --run src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.spec.js`

No storybook stories affected. No follow-up PRs planned — this is a self-contained fix.

### General

### DIFF
--- a/packages/lib-classifier/src/hooks/useMachineLearntReductions.js
+++ b/packages/lib-classifier/src/hooks/useMachineLearntReductions.js
@@ -33,13 +33,11 @@ export default function useMachineLearntReductions() {
     
     if (caesarMarks) {
       caesarMarks.forEach(caesarMark => {
-        const { stepKey, taskKey, taskType, toolType, taskIndex, toolIndex, ...newMark } = caesarMark
+        const { stepKey, taskKey, taskIndex, taskType, toolIndex, toolType, ...newMark } = caesarMark
         newMark.id = newMark.markId;
         delete newMark.markId;
 
-        const resolvedTaskIndex = taskIndex ?? 0
-        const resolvedToolIndex = toolIndex ?? 0
-        step?.tasks[resolvedTaskIndex]?.tools[resolvedToolIndex]?.createMark(newMark)
+        step?.tasks[taskIndex]?.tools[toolIndex]?.createMark(newMark)
       })
     }
 
--- a/packages/lib-classifier/src/hooks/useMachineLearntReductions.js
+++ b/packages/lib-classifier/src/hooks/useMachineLearntReductions.js
@@ -33,12 +33,13 @@ export default function useMachineLearntReductions() {
     
     if (caesarMarks) {
       caesarMarks.forEach(caesarMark => {
-        const { stepKey, taskIndex, taskKey, taskType, toolType, ...newMark } = caesarMark
+        const { stepKey, taskKey, taskType, toolType, taskIndex, toolIndex, ...newMark } = caesarMark
         newMark.id = newMark.markId;
         delete newMark.markId;
 
-        // we only get access to one task at a time in a step
-        step?.tasks[0].tools[newMark.toolIndex]?.createMark(newMark)
+        const resolvedTaskIndex = taskIndex ?? 0
+        const resolvedToolIndex = toolIndex ?? 0
+        step?.tasks[resolvedTaskIndex]?.tools[resolvedToolIndex]?.createMark(newMark)
       })
     }
 
--- a/packages/lib-classifier/src/hooks/useMachineLearntReductions.js
+++ b/packages/lib-classifier/src/hooks/useMachineLearntReductions.js
@@ -22,13 +22,14 @@ export default function useMachineLearntReductions() {
   if (loaded
     && getType(caesarReductions).name === 'MachineLearntReductions'
     && activeSubject.caesarReductionsLoadedForStep[stepIndex] !== true) {
-    
+
     // step.tasks is only the CURRENT TASK
     // Test project with 3 steps means we need to run this 3 times to check for reductions on each step
     // This is likely due to the potentially recursive nature of step navigation
 
     const caesarMarks = caesarReductions.findCurrentTaskMarks({
-      stepKey: step.stepKey
+      stepKey: step.stepKey,
+      workflow
     })
     
     if (caesarMarks) {
--- a/packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js
+++ b/packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js
@@ -23,8 +23,14 @@ const MachineLearntReductions = types
           let { data } = reduction.data   // Caesar expects data attribute to be an object
 
           data.forEach(datum => {
-            if (datum.stepKey === stepKey) {
-              caesarMarks.push(datum)
+            const mark = {
+              stepKey: 'S0',
+              taskIndex: 0,
+              toolIndex: 0,
+              ...datum
+            }
+            if (mark.stepKey === stepKey) {
+              caesarMarks.push(mark)
             }
           })
         })
--- a/packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js
+++ b/packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js
@@ -24,10 +24,10 @@ const MachineLearntReductions = types
 
           data.forEach(datum => {
             const mark = {
-              stepKey: 'S0',
-              taskIndex: 0,
-              toolIndex: 0,
-              ...datum
+              ...datum,
+              stepKey: datum.stepKey ?? 'S0',
+              taskIndex: datum.taskIndex ?? 0,
+              toolIndex: datum.toolIndex ?? 0,
             }
             if (mark.stepKey === stepKey) {
               caesarMarks.push(mark)
--- a/packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js
+++ b/packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js
@@ -12,7 +12,7 @@ const MachineLearntReductions = types
   })
   .views(self => {
     return {
-      findCurrentTaskMarks({ stepKey }) {
+      findCurrentTaskMarks({ stepKey, workflow }) {
         if (stepKey === undefined) {
           return
         }
@@ -25,7 +25,8 @@ const MachineLearntReductions = types
           data.forEach(datum => {
             const mark = {
               ...datum,
-              stepKey: datum.stepKey ?? 'S0',
+              // Stepless Caesar data attaches to the workflow's first step. See PR #7319.
+              stepKey: datum.stepKey ?? workflow?.steps?.[0]?.[0],
               taskIndex: datum.taskIndex ?? 0,
               toolIndex: datum.toolIndex ?? 0,
             }
--- a/packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.spec.js
+++ b/packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.spec.js
@@ -1,4 +1,4 @@
-import { reducedSubjectMocks } from './reducedSubjectMocks'
+import { reducedSubjectMocks, multiIndexMock, minimalKeysMocks } from './reducedSubjectMocks'
 import MachineLearntReductions from './MachineLearntReductions'
 
 let reductionsTaskStub = {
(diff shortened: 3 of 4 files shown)


packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js
```diff
@@ -23,8 +23,14 @@ const MachineLearntReductions = types
           let { data } = reduction.data   // Caesar expects data attribute to be an object
 
           data.forEach(datum => {
-            if (datum.stepKey === stepKey) {
-              caesarMarks.push(datum)
+            const mark = {
+              stepKey: 'S0',
+              taskIndex: 0,
+              toolIndex: 0,
```
lcjohnso: This is naive, but will these initialized definitions for `stepKey`, `taskIndex`, and `toolIndex` be overwritten if the datum includes these entries in the loaded reduction?

packages/lib-classifier/src/hooks/useMachineLearntReductions.js
```diff
@@ -33,12 +33,13 @@ export default function useMachineLearntReductions() {
...
     if (caesarMarks) {
       caesarMarks.forEach(caesarMark => {
-        const { stepKey, taskIndex, taskKey, taskType, toolType, ...newMark } = caesarMark
+        const { stepKey, taskKey, taskType, toolType, taskIndex, toolIndex, ...newMark } = caesarMark
         newMark.id = newMark.markId;
         delete newMark.markId;
 
-        // we only get access to one task at a time in a step
-        step?.tasks[0].tools[newMark.toolIndex]?.createMark(newMark)
+        const resolvedTaskIndex = taskIndex ?? 0
```
lcjohnso: My understanding is that `taskIndex` and `toolIndex` will already be set to 0 OR the reduction-defined value as a result of [MachineLearntReductions.js#L26-L30](https://github.com/zooniverse/front-end-monorepo/blob/fix-7076-caesar-ml-reductions-defaults/packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js#L26-L30), so maybe these `resolvedXIndex` values aren't needed?

packages/lib-classifier/src/hooks/useMachineLearntReductions.js
```diff
@@ -33,12 +33,13 @@ export default function useMachineLearntReductions() {
     
     if (caesarMarks) {
       caesarMarks.forEach(caesarMark => {
-        const { stepKey, taskIndex, taskKey, taskType, toolType, ...newMark } = caesarMark
+        const { stepKey, taskKey, taskType, toolType, taskIndex, toolIndex, ...newMark } = caesarMark
```
lcjohnso: Does order matter here? Otherwise good to group `task` and `tool` values together for readability?

lcjohnso (commented):
Awesome to see this! I have a few clarifying questions, but importantly I'm not able to truly evaluate the JS or tests so I tagged PR for frontend review.

packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js
kieftrav: Nope - this sets the default. It's the line below that overwrites those defaults.

packages/lib-classifier/src/hooks/useMachineLearntReductions.js
kieftrav: Yeah - I did these first to add a null-ish safeguard.. I realized how I can move these into the default-setting area so it's all in one place. Good call out!

packages/lib-classifier/src/hooks/useMachineLearntReductions.js
kieftrav: Nope - I'll reorder for readability. Good suggestion!

packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js
```diff
@@ -23,8 +23,14 @@ const MachineLearntReductions = types
           let { data } = reduction.data   // Caesar expects data attribute to be an object
 
           data.forEach(datum => {
-            if (datum.stepKey === stepKey) {
-              caesarMarks.push(datum)
+            const mark = {
+              ...datum,
+              stepKey: datum.stepKey ?? 'S0',
```
shaunanoordin: [query] I have a few questions here.

- Is it safe to assume that `"S0"` is the _default Step Key?_
- A realistic case to worry about is if the Workflow has only one Step, and it's actually titled `"P0"` (i.e. "Page 0", which was one of the changes sorta-introduced in the Pages Editor)
- Another case is if the Workflow has steps S0, S1... and then they delete S0, so the remaining "default" Step is S1.

Looking at [Cliff's comments](https://github.com/zooniverse/front-end-monorepo/issues/7076#issuecomment-3434484398)...

> I'll note that my dream would be that default values for stepKey, taskIndex, and toolIndex could be assumed when absent, rather than forcing every uploaded annotation to include those keys. Example: many correct-a-machine workflows will only have one page/step, one task, and one tool, so just load reductions assuming that's the case. I acknowledge that facilitating this will likely require a non-trivial refactor of the code.

...my understanding is that the _default Step Key's_ basic use case is to kick in when the Caesar setup just assumes there's one Step in the WF. i.e. _"if datum doesn't specify the Step Key, just assume it's for THIS Step. (Also assume this Step is the only Step in the Workflow.)"_

If this assessment is correct, would a better line be...

```
stepKey: datum.stepKey ?? stepKey,  // If Caesar doesn't specify which Step this data point is for, assume the current Step. This is mostly applicable for 1-Step Workflows with minimal configurations.
```

shaunanoordin (dismissed):
## PR Review

Package: lib-classifier
Affects: Subject Viewer -> Interaction Layer -> DrawingToolMarks

This PR updates how the DrawingToolMarks (sub-layer of the Interaction Layer) processes "aggregation data"[^1] from Caesar.

Functionality changes:

- useMachineLearntReductions() now checks if the aggregations data is meant for the current Step, instead of assuming the aggregations data is meant to be fed into _the first Step_ of the Workflow.
- MachineLearntReductions now has fallbacks/default values for Step Keys, Task Index, and Tool Index, if the aggregations data didn't provide those details.
  - Functionally, this means that the Classifier is more robust in processing aggregations data. This would only matter for future workflows/Caeasar configs though, since existing workflows don't require this level of robustness.

[^1]: kids these days called it "machine learnt data" from Caesar, but I still remember when this setup was originally designed for volunteer aggregations goshdangit.

Code changes:

- useMachineLearntReductions() has very straightforward changes.
- MachineLearntReductions.js applies default values for some properties. before adding aggregation data to the Caesar Marks array.
  - 🤔 **I added some comments about whether it's a good idea to assume that `"S0"` should be the "default Step Key value", and then proposed an alternative.**

### Testing

We're mostly testing to see that aggregated marks (or machine-learnt data from Caesar, whatever you call it) are still displayed correctly on the Subject Viewer, _for currently configure workflows + Caesar configs._ (I don't believe we have any live tests where the Caesar reductions don't return a .stepKey, .taskIndex, or .toolIndex value for testing the default values, for example.)

Tested on lib-classifier on localhost with macOS + Chrome.

- Load the test workflow, which has Caesar reductions (or aggregated data or machine-learnt data): https://local.zooniverse.org:8080/?project=22358&env=production
- Observe that the Subject is rendered with some aggregated data visible on the Subject (i.e. the "Caesar Marks")
- Check the Network tab of the browser to ensure these correspond with aggregation results from https://caesar.zooniverse.org/graphql
  - Confirm the shape of the aggregation data items, e.g. `{ "frame": 0, "width": 7.707611083984375, "height": 6.9209442138671875, "markId": "CDnXo7", "stepKey": "S0", "taskKey": "T0", "taskType": "drawing", "toolType": "rectangle", "x_center": 197.99998474121094, "y_center": 197.19638061523438, "taskIndex": 0, "toolIndex": 0 }`
    Image
- Make some edits. Make some new marks. Submit Classification (in Demo mode)

Tests work as expected.

### Status

LGTM. 👌 I'm giving this PR the green light, but please seriously consider my comments about whether using "S0" as the default Step Key is a good idea.

packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js
lcjohnso: Lets use [this Slack thread](https://zooniverse.slack.com/archives/CANKLB50E/p1777580864672709) to discuss the bigger picture question re: Step/Page keys, but here I have a specific question about your recommended change.

Would this solution work for the case of a two-step workflow where the step is not defined in the annotation?

Example: What if I had a drawing task in S0 for which I wanted to load a MachineLearnt annotation from Caesar, and then had a follow-up "is there anything else strange/notable" question task in S1. In the case the research team does not define the Step Key in the Caesar-stored annotation, I expect the following default behavior:
1. The MachineLearnt annotation from Caesar is loaded when a volunteer starts their classification in Step S0, because by default if no step is defined in the stored reduction then S0 is assumed.
2. The volunteer makes alterations to the annotation, and hits "next" to advance to Step S1 and its question task.
3. I expect the annotations to remain as-is in the viewer while answering the question task, and then the volunteer submits the classification.

Would your solution _re-load_ the MachineLearnt annotation for _every_ step?

lcjohnso: Noting @kieftrav suggested changing to position-based default to identify first step over in [Slack](https://zooniverse.slack.com/archives/CANKLB50E/p1777582241103059?thread_ts=1777580864.672709&cid=CANKLB50E): `workflow.steps[0][0]`

shaunanoordin: Popping in to note that I'm cool with Travis's plan of action. 👌 

- Instead of using "S0" or "current Step key" as the default value, use the _first Step._
- I would advise being careful for cases where the workflow doesn't have a first step for whatever reason[^1], so a `const defaultStepKey = workflow.steps[0]?.[0]` would be safer.

📖 Documentation: we're introducing a new implicit behaviour, which is: _"if the aggregated/machine-learnt/etc results from Caesar doesn't specify which Step the results belong to, assume that it belongs to the FIRST Step."_ This needs to be documented somewhere - at minimum, please add a comment in the code that points to this PR or something.

--------

<details>
<summary>➕ Additional technical nerdery</summary>

btw, in response to this question:

> Would this _(default Step key == current step)_ solution work for the case of a two-step workflow where the step is not defined in the annotation?

The answer is _nope._ It only works if the workflow has only one Step in the workflow.

As mentioned:

> ...my understanding is that the default Step Key's basic use case is to kick in when the Caesar setup just assumes there's one Step in the WF.

This actually applies to any implementation of a default step key = (whatever) solution, since as soon as you introduce a two-step workflow, the base assumption for a default workflow breaks. For example, I can imagine a technically valid _(if practically inadvisable)_ workflow which looks like...

Step 1: Does this page have words to transcribe? Yes -> Submit ; No -> go to Step 2
Step 2: OK, transcribe away! [pulls data from Caesar]

...and this would break the "default step key = first Step" solution. _And I'm totally OK with that,_ as long as we have a doc that says "if your Caesar reductions doesn't specify the Step Key, then... (etc etc)"

</details>

--------

<details>
<summary>➕ Bonus comments</summary>

Travis: as mentioned, I'm giving your plan a thumbs up 👍 If you want me to re-review this PR, please dismiss my current review, then ping me again when you're ready. 👍 (Bonus thumbs up!)

Cliff: in response to [your Slack comment](https://zooniverse.slack.com/archives/CANKLB50E/p1777582757264789?thread_ts=1777580864.672709&cid=CANKLB50E), I agree that our Step Keys are a bit of a Wild West situation! I _was_ going to say _"well ackshually"_ and link to a set of documents that carefully detail the prescribed structure of a Zooniverse Panoptes workflow and its tasks and steps, but this is the best I've got so far:

- [this old ADR uses "S1" and "T1" as first step and first task in its example without actually explaining the choice, wtf](https://github.com/zooniverse/front-end-monorepo/blob/main/docs/arch/adr-05.md)
- [TODO](#🤷-i-dunno-mate)

</details>

[^1]: 🤷 a workflow with Tasks but no Steps this should be _impossible_ in practice due to convertWorkflowToUseSteps(), but best to play it safe with some defensive programming.

kieftrav:
@lcjohnso & @shaunanoordin - I have rebased on top of the latest main, incorporated the feedback from the above discussion (and corresponding slack conversation). I'll dismiss the old review and re-request review. Thank you both!

packages/lib-classifier/src/store/subjects/Subject/MachineLearntReductions/MachineLearntReductions.js
```diff
@@ -23,8 +23,15 @@ const MachineLearntReductions = types
           let { data } = reduction.data   // Caesar expects data attribute to be an object
 
           data.forEach(datum => {
-            if (datum.stepKey === stepKey) {
-              caesarMarks.push(datum)
+            const mark = {
+              ...datum,
+              // Stepless Caesar data attaches to the workflow's first step. See PR #7319.
```
shaunanoordin, line 28: 👍

shaunanoordin (approved):
## PR Review (Update)

Follows [earlier review](https://github.com/zooniverse/front-end-monorepo/pull/7319#pullrequestreview-4199163672)

The updated PR has changed how the "default Step Key, if Step Key isn't specified by Caesar" behaviour works.

- Now, instead of assuming the default Step Key is "S0", FEM's MachineLearntReductions assumes the target of the Caesar aggregations is the _first_ Step.
  - This only applies if the Caesar config doesn't specify the step key, which should only apply if the project owner knows that they're doing.
- We have minimal documentation in the form of an inline comment explaining what/why this is happening, and links to this PR. Looks good to me. 👌 

### Testing

Same as previous review.

Additional tests:

- Ran app-project on localhost and loaded up a baseline test project: https://local.zooniverse.org:3000/projects/darkeshard/test-project-2022
- Confirmed that projects without Caesar reductions continue to work as normal.

### Status

LGTM! 👍 

I imagine that some of this information on the new "hey btw we can accept Caesar configs without specifying step keys" behaviour will need to be reflected in https://github.com/zooniverse/caesar/pull/1612

PR #7322, opened 2026-04-16
DataImageViewer: support multiple images

### LINKED ISSUE #7241: DataImageViewer: add support for multiple images
## DataImageViewer Enhancement

Package: lib-classifier

The [DataImageViewer](https://github.com/zooniverse/front-end-monorepo/blob/main/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/DataImageViewer/README.md) currently renders a JSON scatter plot and an image media file side by side, with pan and zoom support for the plot and image independently. This issue requests an enhancement to support a JSON scatter plot with _multiple_ images (e.g., three images; four total media = .json and 3 x .jpg).

In essence, can we swap in a Flipbook Viewer in place of the Single Image Viewer within the hybrid DataImageViewer? 

Driving Use Case: for Rubin Observatory projects (especially those based on the Alert Stream data), we want to support subjects with lightcurve information display in a JSON plot along with multiple separate image frames (e.g., three image files for reference, science, and difference image).

If this enhancement is prohibitively difficult, please reach out -- we may want to consider alternate methods of display (e.g., allowing the first frame of a flipbook viewer to be an interactive plot -- which actually might be harder....) rather than forcing a square peg into a round hole (i.e., forcing through a difficult solution).

shaunanoordin:
## PR Overview

Package: `lib-classifier`
Closes #7241

This PR allows the DataImageViewer to show multiple images.

image

Feature updates:

- DataImageViewer:[^1] if a Subject has 1 JSON and 1+ images, then those images will be displayed in the FlipbookViewer format. 🆕 
  - if a Subject has 1 JSON and 1 image, then that image will be displayed in the standard SingleImageViewer. (no change.)

[^1]: reminder - DataImageViewer is enabled by setting `workflow.configuration.subject_viewer = 'dataImage'`. The use of this viewer can't be inferred by the presence of a 'Data Image Subject', because there's no such thing as a 'Data Image Subject'.

Code changes:

- Data Image Viewer now expects all Subjects fed to it to have _exactly 1_ JSON file and _1 or more_ image files.
- **useDataImageSubject() now assumes that Subjects will NOT have a thumbnail image.**
- useDataImageSubject() now returns imageLocations (plural) instead of imageLocation.

Documentation updates:

- DataImageViewer's README (and by extension the "Data Image Subject"'s documentation) changed.
  - Clarified expected Subject structure.
  - Added additional dev notes on DataImageViewer's current narrow use case, and possible future scope expansion.

### Testing

Basic testing:

- Run Storybook for lib-classifier
- Confirm that the new Storybook Story (Subject Viewers -> DataImageViewer -> _Multi Image Subjects_ ) works correctly.
  - This new Story should allow you to see 1 JSON data on a scatterplot, and 3 images presented in a flipbook viewer.
- Confirm that the other Storybook Stories for the DataImageViewer work correctly too.
  - Other Stories should look/work the same as before this PR.

Note: the layout looks a bit funny on a narrow view, because the image toolbar hangs on the top-right while the actual image drops to the bottom, but this seems consistent with the current 1 JSON + 1 image design. 🤷 

image

Advanced testing:

- Run app-project
- Load  the test project/workflow, which is Cliff's "Data As Subject Testing" workflow on "LCJ Test" on _production._
  - https://local.zooniverse.org:3000/projects/lcjohnso/lcj-test/classify/workflow/23354?env=production&demo=true 
  - This project has 1 Subject (id 117885353) which contains `[ JSON, image, image, image ]`
- You should see a single scatterplot, next to a flipbook viewer showing 3 images.
    image
- Interact with the viewer to confirm you can view/explore/etc all 4 media items.
- Submit the classification (in demo mode) 

### Status

Ready for review! ~~If you'd rather wait until I create that full test project so you can test on app-project, that's fine too - just let me know.~~

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
@@ -27,7 +27,9 @@ const FlipbookViewer = ({
   rotation = 0,
   setOnPan = DEFAULT_HANDLER,
   setOnZoom = DEFAULT_HANDLER,
-  subject
+  subject,
+  zoomControlFn = null,
+  zooming = true
 }) => {
   const [currentFrame, setCurrentFrame] = useState(defaultFrame)
   const [playing, setPlaying] = useState(false)
@@ -93,6 +95,8 @@ const FlipbookViewer = ({
           setOnZoom={setOnZoom}
           src={currentMediaUrl}
           subject={subject}
+          zoomControlFn={zoomControlFn}
+          zooming={zooming}
         />
       )}
       {currentMediaType === 'video' && (
@@ -151,7 +155,9 @@ FlipbookViewer.propTypes = {
   /** Required. Passed from SubjectViewer component */
   subject: PropTypes.shape({
     locations: PropTypes.arrayOf(locationValidator)
-  }).isRequired
+  }).isRequired,
+  zoomControlFn: PropTypes.func,
+  zooming: PropTypes.bool,
 }
 
 export default FlipbookViewer
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
@@ -148,15 +148,17 @@ FlipbookViewer.propTypes = {
   playIterations: PropTypes.number,
   /** Passed from the subject viewer store. Needed in SingleImageViewer to handle transforming (rotating) the image */
   rotation: PropTypes.number,
-  /** Passed from the Subject Viewer Store */
+  /** Usually passed from the Subject Viewer Store. Can be overridden, such as when the FlipbookViewer is used in the DataImageViewer.  */
   setOnPan: PropTypes.func,
-  /** Passed from the Subject Viewer Store */
+  /** Usually passed from the Subject Viewer Store. Can be overridden, such as when the FlipbookViewer is used in the DataImageViewer. */
   setOnZoom: PropTypes.func,
   /** Required. Passed from SubjectViewer component */
   subject: PropTypes.shape({
     locations: PropTypes.arrayOf(locationValidator)
   }).isRequired,
+  /** OPTIONAL: only supply zoomControlFn and zooming if you want to allow users to manually enable/disable zoom & pan functionality. (As of Apr 2026, only used by DataImageViewer.) */
   zoomControlFn: PropTypes.func,
+  /** OPTIONAL: see zoomControlFn. */
   zooming: PropTypes.bool,
 }
 
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
@@ -28,7 +28,7 @@ const FlipbookViewer = ({
   setOnPan = DEFAULT_HANDLER,
   setOnZoom = DEFAULT_HANDLER,
   subject,
-  zoomControlFn = null,
+  zoomControlFn = undefined,
   zooming = true
 }) => {
   const [currentFrame, setCurrentFrame] = useState(defaultFrame)
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewerContainer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewerContainer.jsx
@@ -48,6 +48,8 @@ function FlipbookViewerContainer({
   loadingState = asyncStates.initialized,
   onError = DEFAULT_HANDLER,
   onReady = DEFAULT_HANDLER,
+  setOnPan: propSetOnPan = null,
+  setOnZoom: propSetOnZoom = null,
   subject,
   zoomControlFn = null,
   zooming = true
@@ -62,9 +64,12 @@ function FlipbookViewerContainer({
     playIterations,
     rotation,
     separateFramesView,
-    setOnPan,
-    setOnZoom
+    setOnZoom: storeSetOnZoom,
+    setOnPan: storeSetOnPan,
   } = useStores(storeMapper)
+
+  const effectiveSetOnPan = propSetOnPan || storeSetOnPan
+  const effectiveSetOnZoom = propSetOnZoom || storeSetOnZoom
 
   useEffect(function preloadImages() {
     subject?.locations?.forEach(({ type, url }) => {
@@ -103,8 +108,8 @@ function FlipbookViewerContainer({
           onReady={onReady}
           playIterations={playIterations}
           rotation={rotation}
-          setOnPan={setOnPan}
-          setOnZoom={setOnZoom}
+          setOnPan={effectiveSetOnPan}
+          setOnZoom={effectiveSetOnZoom}
           subject={subject}
           zoomControlFn={zoomControlFn}
           zooming={zooming}
@@ -123,6 +128,8 @@ FlipbookViewerContainer.propTypes = {
   onError: PropTypes.func,
   /** Passed from SubjectViewer and dimensions are added to classification metadata. Called after svg layers successfully load with `defaultFrameSrc`. */
   onReady: PropTypes.func,
+  setOnPan: PropTypes.func,
+  setOnZoom: PropTypes.func,
   /** Required. Passed from mobx store via SubjectViewer. */
   subject: PropTypes.shape({
     locations: PropTypes.arrayOf(locationValidator)
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewerContainer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewerContainer.jsx
@@ -48,7 +48,9 @@ function FlipbookViewerContainer({
   loadingState = asyncStates.initialized,
   onError = DEFAULT_HANDLER,
   onReady = DEFAULT_HANDLER,
-  subject
+  subject,
+  zoomControlFn = null,
+  zooming = true
 }) {
   const {
     defaultFrame,
@@ -104,6 +106,8 @@ function FlipbookViewerContainer({
           setOnPan={setOnPan}
           setOnZoom={setOnZoom}
           subject={subject}
+          zoomControlFn={zoomControlFn}
+          zooming={zooming}
         />
       )}
     </>
@@ -122,7 +126,9 @@ FlipbookViewerContainer.propTypes = {
   /** Required. Passed from mobx store via SubjectViewer. */
   subject: PropTypes.shape({
     locations: PropTypes.arrayOf(locationValidator)
-  }).isRequired
+  }).isRequired,
+  zoomControlFn: PropTypes.func,
+  zooming: PropTypes.bool
 }
 
 export default observer(FlipbookViewerContainer)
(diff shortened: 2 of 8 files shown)


shaunanoordin:
## PR (Non-)Update

The PR code hasn't been updated, but the PR _description_ has been! I've added [notes on Testing](https://github.com/zooniverse/front-end-monorepo/pull/7322#issue-4278611621) which explains that we DO have a fully functional project/workflow that utilises Data Image Viewer + Data Image Subject, which is Cliff's testing workflow with 1 Subject containing 1 JSON and 3 images:

app-project - https://local.zooniverse.org:3000/projects/lcjohnso/lcj-test/classify/workflow/23354?env=production&demo=true

Again, thanks to Cliff for setting up this project![^1]

### Status

Ready for review, [with full testing instructions.](https://github.com/zooniverse/front-end-monorepo/pull/7322#issue-4278611621)

[^1]: I was in the process of creating my own test project before I realised I was literally copying everything that Cliff already set up. I might continue developing my test project as a general [JSON Data Benchmark Project](https://www.zooniverse.org/lab/32752) though, but that's for another day.

goplayoutside3 (commented):
I ran app-project on this branch locally and visited the test workflow https://local.zooniverse.org:3000/projects/lcjohnso/lcj-test/classify/workflow/23354?env=production&demo=true.

A couple of UI things to note before I do a full code read. Are these intentional or unnoticed thus far?
1. When I use the image toolbar next to the flipbook viewer in the DataImageViewer, the "reset subject viewer" button does not always reset the zoom level of the flipbook. It does reset rotation, but something becomes frozen when I click the "reset subject viewer" button after enabling then disabling zoom on the scatterplot. Clicking through in this sequence replicates the bug 1) Load subject 2) Click on the "zoom out" button next to the flipbook 3) Click on the "rotate" button next to the flipbook 4) Click "enable zoom" above the scatterplot 5) Click "disable zoom" above the scatterplot 6) Click "reset subject view" next to the flipbook 6) Subject image rotation resets, its zoom does not reset, zoom buttons no longer functional.
2. When this test workflow is displayed on a mobile-sized screen, the image toolbar appears next to the scatterplot instead of the flipbook viewer even though the zoom is disabled on the scatterplot viewer by default. You already mentioned this layout with a screenshot in the PR description, but I'm echoing it here because I think it's a non-intuitive UI if a project launches publicly with this viewer. I think the bug above ^ is more important to address in this PR, but the enabling/disabling zoom when the viewer is stacked vertically is a potentially headache for future beta testers and worth discussing with Cliff.

shaunanoordin:
Thank you Delilah! Your feedback (and spotting of that major Image Toolbar bug) has been really helpful, and I had a follow-up chat with @lcjohnso a few hours ago.

Responses:

- re: When JSON Data Viewer's "Enable Zoom" is toggled, it disables Image Toolbar + general zoom functionality of Flipbook Viewer.
  - ⚠️ This is a major bug. I will need to address this in this PR.
- re: Image Toolbar UI/UX isn't intuitive on narrow screens.
  - You, me, and Cliff are in agreement on this. The UI/UX is rather unwieldy, but it's low priority compared to the bug above. One idea is to give the JSON data viewer its own Image Toolbar, similar to how the Separate Frames Viewer has a (pseudo) Image Toolbar for each image. (See below, Option 2.)

## PR Update

**This PR is now back in development mode.** New goals:

- Fix "toggling Enable Zoom on JSON Data Viewer" disables Flipbook Viewer's Image Toolbar + general zoom functionality. Two options:
  - Option 1: Re-implement the "Enable Zoom" toggle that was present for the Single Image Viewer.
    - This means volunteers have to manually toggle the "Enable Zoom" option on the appropriate viewer, to perform any zoom/pan/rotate/reset interaction, for that viewer.
  - Option 2: Implement two separate Image Toolbars, ones for the JSON Data Viewer, and one for the SingleImagViewer/FlipbookViewer.

Dev time will be the major in deciding which of the two options I'll go for; if Option 1 is easier, I'll queue the "Data Image Viewer Should Have Two Image Toolbars" idea (aka Option 2) as a possible future enhancement since Cliff & me think that's the better/more intuitive UX approach.

shaunanoordin:
## PR Update

Commits [5e058e8](https://github.com/zooniverse/front-end-monorepo/pull/7322/commits/5e058e80db65da934b75d3d76c61c60256457d94) to [696e4a3](https://github.com/zooniverse/front-end-monorepo/pull/7322/commits/696e4a3c09f8dddb352d5925a956d3fd7d2bdb3f)

A (surprisingly not-too-huge) update was added to this PR to address the issue of _"toggling Enable Zoom on JSON Data Viewer disables Flipbook Viewer's Image Toolbar + general zoom functionality"._

Now, when _a Subject has multiple images,_ the image viewer will use the same "Enable Zoom" toggle as when _a Subject has a single image._ (i.e. the Flipbook Viewer now uses the same zoomControlFn pattern as the Single Image Viewer.)

image

Functional changes:

- DataImageViewer: when subject has multiple images (i.e. using Flipbook Viewer), the **"enable/disable zoom" controls** now functions similarly to when a subject has one image (i.e. using SingleImageViewer)
  - 🆕 UI: Flipbook Viewer now shows the Enable Zoom / Disable Zoom button.
    - 🟰 "Enable Zoom" indicates whether the JSON data viewer OR the image(s) viewer reacts to the Image Toolbar buttons.
    - 🟰 "Enable Zoom" indicates whether the data/image viewer reacts to mouse input for panning and zooming. (Drag and mouse wheel)
    - 🟰 Only ONE viewer can have zoom enabled at a time. Enabling one will disable the other.
    - 🆕 For multi-images, Enable Zoom is enabled for all images (i.e. even when you switch the current image)
  - 🆕 When _(and only when)_ zoom is enabled for the flipbook viewer AND image toolbar is in Move mode:
    - mouse drag will pan image.
    - mousewheel will zoom image.
    - Image Toolbar's zoom in and zoom out buttons will zoom in/out of image
  - 🟰 [QUIRK] Image Toolbar's **rotate button** will ALWAYS rotate the image, regardless if the image viewer is zoom-enabled. 🤷  
  - 🟰 Image Toolbar's reset button will reset the zoom and pan of the current zoom-enabled viewer.
    - 🟰 [QUIRK] The reset button will ALWAYS reset the rotation of the image, regardless if the image viewer is zoom-enabled. 🤷 
  - 🟰 [QUIRK] there's no consistent behaviour on whether a the viewer's zoom/pan/rotation is reset when changing between zoom-enabled and zoom-disabled modes.[^1]
    - For the JSON data viewer, when zoom is disabled, the zoom & pan of the viewer resets to default.
    - For the image viewers (both SingleImageViewer and FlipbookViewer), the zoom, pan, and rotation of the viewer ~~sometimes resets, sometimes doesn't?~~ only resets if the "Enable/Disable Zoom" buttons is triggered twice in a row. 🤨 

(🟰 indicates existing behaviour)

[^1]: I'm investigating this as a separate issue as I'm noting that it's occurring even on the `main` branch.

Code changes:

- FlipbookViewer now accepts optional `zooming`, `zoomControlFn`, `onSetPan`, and `onSetZoom`
  - This mirrors what's used by SingleImageViewer, and in fact FlipbookViewer just passes these values down.
- [docs] Documentation also updated.

### Testing

- Run app-project
- Load the test project/workflow, which is Cliff's "Data As Subject Testing" workflow on "LCJ Test" on production.
  - https://local.zooniverse.org:3000/projects/lcjohnso/lcj-test/classify/workflow/23354?env=production&demo=true
  - This project has 1 Subject (id 117885353) which contains [ JSON, image, image, image ]
- You should see a single scatterplot, next to a flipbook viewer showing 3 images.
- 🆕 JSON Data Viewer:
  - Viewer should currently be in zoom-disabled mode.
  - Ensure you CAN'T pan & zoom using mouse; and that the Image Toolbar's zoom in/out and reset buttons DON'T work.
  - Click "Enable Zoom".
  - Ensure now you CAN pan & zoom using mouse; and that the Image Toolbar's zoom in/out and reset buttons DO work.
  - Click "Disable Zoom".
- 🆕 Images Viewer aka Flipbook Viewer:
  - Viewer should currently be in zoom-disabled mode.
  - Ensure you CAN'T pan & zoom using mouse; and that the Image Toolbar's zoom in/out and reset buttons DON'T work. _(Note: uh, image rotation can be reset though.)_
  - Click "Enable Zoom".
  - Ensure now you CAN pan & zoom using mouse; and that the Image Toolbar's zoom in/out and reset buttons DO work.
  - Ensure that the Image Toolbar's rotate button does work. _It will work._ You can't stop it. It doesn't care about your zoom-enabled rules. _The image will always rotate._
  - Click "Disable Zoom".
- 🆕 JSON Data Viewer _and_ Image(s) Viewer:
  - Enable the zoom on one viewer.
  - Now enable the zoom on the other viewer.
  - Ensure that the first viewer is now zoom-disabled.
- Submit the classification (in demo mode)

### Additional Testing

Testing must now include baseline tests to ensure that the FlipbookViewer functions normally OUTSIDE of the DataImageViewer context.

Baseline Testing for Flipbook Viewer:

- Test with app-project.
- Select a project with multi-image Subjects using the Flipbook viewer AND a Drawing Task, e.g.: https://local.zooniverse.org:3000/projects/marek-slipski/cloudspotting-on-mars/classify/workflow/16805?env=production&demo=true
  - Use the Image Toolbar to switch to Annotate Mode.
  - Add annotations by clicking on the Subject.
  - Use the flipbook controls to switch images/frames. Ensure annotations are the same across every frame.
  - Switch to Move Mode.
    - Pan the view of the Subject by using mouse drag.
    - Pan the view of the Subject by using keyboard keys.
    - Zoom using the mouse wheel.
    - Zoom using the Image Toolbar buttons.
  - Rotate the Subject using the Image Toolbar button.
  - Reset the view of the Subject using the Image Toolbar button.
  - Play the images in sequence by using the "Play" button.
- Submit classification in demo mode.
  - Ensure next Subject loads with default viewer settings (i.e. not zoomed in/out, not panned, not rotated)

### Status

This PR is now ready for review again, as the main blocking issue has been addressed _(I believe)._

I'm going to continue examining those two quirks with the "Rotate doesn't care about zoom-enabled state"[^1] and the "inconsistent view reset behaviour on image viewers (both Flipbook Viewer and Single Image Viewer) when toggling between zoom-enabled/disabled modes"[^3] and will open separate issues if I find anything useful.

[^2]: I suspect it's because there's never been any attempt to override the onRotate action - compared to how we have setOnPan and setOnZoom - so the image toolbar just uses the default rotate action of the store, which means changing the global `subjectViewer.rotation` value, which is something only the image viewers care about.
[^3]: current analysis is _heck if I know._ 🤷

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewerContainer.jsx
```diff
@@ -48,7 +48,11 @@ function FlipbookViewerContainer({
   loadingState = asyncStates.initialized,
   onError = DEFAULT_HANDLER,
   onReady = DEFAULT_HANDLER,
-  subject
+  setOnPan: propSetOnPan = null,
+  setOnZoom: propSetOnZoom = null,
```
goplayoutside3: As you noted in the PropTypes comments, `setOnPan` and `setOnZoom` won't be passed as a prop to FlipbookViewerContainer the majority of the time. When a React prop is not passed to a component, it is `undefined`. If you're going to set a default value for these functions, I think it should be `undefined` instead of `null`.

Same with `zoomControlFn` below.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
```diff
@@ -27,7 +27,9 @@ const FlipbookViewer = ({
   rotation = 0,
   setOnPan = DEFAULT_HANDLER,
   setOnZoom = DEFAULT_HANDLER,
-  subject
+  subject,
+  zoomControlFn = null,
```
goplayoutside3: As you noted in the PropTypes comment, `zoomControFn` won't be passed as a prop to FlipbookViewer the majority of the time. When a React prop is not passed to a component, it is `undefined`. If you're going to set a default value for this function, I think it should be `undefined` instead of `null`.

goplayoutside3 (approved):
Thank you for listing the steps for testing. Reading through the code changes and comparing the UX was easy 👍 This PR looks good to go. I just had two NB comments about using default values for missing React props.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewerContainer.jsx
shaunanoordin: This makes sense, I'll change the default values for setOnPan/Zoom and zoomControlFn to `undefined`. 👌

This pattern used `null` because it was the default used in the [SingleImageViewer](https://github.com/zooniverse/front-end-monorepo/blob/c680015/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/SingleImageViewer/SingleImageViewerContainer.jsx#L55), but upon closer inspection of that component, it would've made more sense to use `undefined` there too. 🤨

shaunanoordin:
Thanks for reviewing this, Delilah! 👍

PR #7326, opened 2026-04-20
FlipbookViewer: fix broken Image Toolbar, when first item isn't an image

### LINKED ISSUE #7325: Flipbook Viewer: image+video mode disables Pan & Zoom in Image Toolbar
## UI Issue

Package: lib-classifier
Component: Flipbook Viewer (possibly Separate Frames Viewer too)
Follows #7253

After the addition of image + video support in  7253, we [discovered an issue](https://github.com/zooniverse/front-end-monorepo/pull/7253#issuecomment-4268933937) where _the Pan & Zoom buttons on the image toolbar_ are disabled when viewing an Image file.

- The pan, zoom in, zoom out, rotate, and reset buttons on the image toolbar are disabled.
  - (Expected: these should be enabled and functional)
- Curiously, you can _still_ pan and zoom using the Ctrl + Mouse Drag action.

_Screenshot: Flipbook Viewer, while viewing an Image file. Note that the Image Toolbar has all its pertinent buttons disabled. BUT, the image itself has been zoomed & panned, notably using the Ctrl + Mouse Drag action._ [(URL)](https://frontend.preview.zooniverse.org/projects/sassydumbledore/chimp-and-see/classify/workflow/19245?demo=true)

Image

### Status

Investigation ongoing.

shaunanoordin:
## PR Overview

Package: lib-classifier
Fixes #7325 
Affects: FlipbookViewer _(and possible SeparateFramesViewer)_

This PR attempts to fix the Image Toolbar's Move/Pan/etc buttons on the Flipbook Viewer, which are disabled when _the first frame of the Subject isn't an image file._

- The Image Toolbar button's on/off settings are determined _by the current frame of the Subject._
- The Flipbook Viewer meanwhile uses _local state_ to keep track of its own separate "current frame".
- The Separate Frames Viewer does its own thing.

🔶 Functional changes:

- FlipbookViewer:
  - (fix) The Image Toolbar now correctly recognises what's the current frame.
  - When the current frame is a video...
    - The zoom, move, rotate, reset buttons on the Image Toolbar are disabled.
    - (fix) Invert Colours button on the Image Toolbar now properly inverts videos.
- SeparateFramesViewer:
  - (no changes, should work as normal)

🔶 Code changes:

- FlipbookViewer:
  - "current frame" value is now set at the _global store_ level, not at the local state level.
  - New params: `frame`, `setFrame` (provided
  - Change of params: `defaultFrame` is kept, but only used to determine the "size of the subject viewer".
- ResetButton, RotateButton, ZoomInButton, ZoomOutButton:
  - New `overrideDisabled` param. Usually, these buttons are only disabled if the Subject Viewer Store says so (as a result of the detected Subject type), but this param allows us to go "nuh-uh, you're enabled/disabled whether you like it or not!"
- SeparateFramesViewer: sets overrideDisabled=false for all buttons, when the attached media item is an image.

<details>
<summary>(Prev notes)</summary>

This PR attempts to implement said fix by:

- Removing FlipbookViewer's _local state_ "current frame", and adding the "current frame" (and "set frame" function) from the Subject Viewer Store.
  - This puts it in line with other viewers like the MultiFrameViewer and ImageAndTextViewer.
- ~~⚠️ TODO: SeparateFramesViewer~~

</details>

Dev Questions:

- Is there a reason why the FlipbookViewer use local state for keeping track of current frame, instead of using the Subject Viewer Store?
  - _Update: local state was used to manage performance. [(comment)](https://github.com/zooniverse/front-end-monorepo/pull/7326#issuecomment-4284021414) Also, it was done to prevent multiple entries to classifications.metadata.subject_dimensions. [(comment)](https://github.com/zooniverse/front-end-monorepo/pull/7326#issuecomment-4284117275) Upon investigatin, performance may not be an issue [(comments 1](https://github.com/zooniverse/front-end-monorepo/pull/7326#issuecomment-4734164459), [2)](https://github.com/zooniverse/front-end-monorepo/pull/7326#issuecomment-4744206848)_
- Why does the Flipbook Viewer's `defaultFrame` prop have a comment that says _"Fetched from metadata.default_frame or initialized to zero"_ ? This clearly hasn't been the case. What metadata.default_frame?
  - _Update: project owners can set subject.metadata.default_frame to set any frame as the "first frame". [(comment)](https://github.com/zooniverse/front-end-monorepo/pull/7326#issuecomment-4284153580)_ 
- [bug] 🎥 When a frame contains a video, the "Invert Colours" button on the Image Toolbar is enabled but doesn't do anything. Should I implement colour inversion, or disable the "Invert Colours" button, or ignore?
  - _Update: yes I should [(comment)](https://github.com/zooniverse/front-end-monorepo/pull/7326#issuecomment-4284170134), and yes I've done. [(comment)](https://github.com/zooniverse/front-end-monorepo/pull/7326#issuecomment-4284721050)_

### Testing

🔶 Test on localhost on app-project.

**Test 1: Flipbook Viewer: subjects with multiple images**

_There should be no behavioural change here. This scenario should have always been working, and should continue to work normally._

- Select a project/wf that uses the Flipbook Viewer, with Subjects consisting of _multiple images_
  - Try Snapshot Wisconsin: https://local.zooniverse.org:3000/projects/zooniverse/snapshot-wisconsin/classify/workflow/32164?env=production&demo=true
- Switch between frames using the Flipbook Controls.
  - Use the Play/Pause button to automatically flip through the Flipbook Viewer.
  - 👁️‍🗨️ Check that the _performance_ "feels OK". (No significant stutter, lag, etc compared to baseline.)
- 🖼️  Ensure the _frames with images_ can be panned/zoomed _using the Ctrl + Mouse Drag action._ 
- 🖼️  Ensure the _frames with images_ can be zoomed, rotated, and reset **using the Image Toolbar buttons.** 

**Test 2: Flipbook Viewer: subjects with images + videos**

- Select a project/wf that uses the Flipbook Viewer, with Subjects consisting of a combination of _images and videos_
  - Try Chimp & See: https://local.zooniverse.org:3000/projects/sassydumbledore/chimp-and-see/classify/workflow/19267?env=production&demo=true
  - **Ensure at least some Subjects contain a video as their _first frame._ This is important.**
- Switch between frames using the Flipbook Controls.
  - Use the Play/Pause button to automatically flip through the Flipbook Viewer.
  - 👁️‍🗨️ Check that the _performance_ "feels OK". (No significant stutter, lag, etc compared to baseline.)
- 🖼️  Ensure the _frames with images_ can be panned/zoomed _using the Ctrl + Mouse Drag action._ 
- 🖼️ ⭐ 🆕 Ensure the _frames with images_ can be zoomed, rotated, and reset **using the Image Toolbar buttons.** 
- 🎥 Ensure the _frames with videos_ CAN'T be panned/zoomed using the Image Toolbar buttons. 
- 🎥 Ensure the _frames with videos_ can have their colours inverted.

**Test 3: Separate Frames Viewer: subjects with images + videos**

- Select the same project as "Flipbook Viewer: subjects with images + videos". (We want to ensure the same testing criteria.)
- Switch from Flipbook Viewer to Separate Frames Viewer by pressing the mysterious button in the flipbook controls.
- 🎥 For EACH _frame with a video,_ ensure that the Image Toolbar doesn't exist.[^1]
- 🖼️ For EACH _frame with an image,_ ensure that the Image Toolbar is enabled.
  - Pan and zoom using Ctrl + Mouse Drag action
  - Zoom/reset/rotate/etc using Image Toolbar buttons.

[^1]: This does mean that, strangely, you can't invert the colour of videos on the Separate Frames Viewer. This is consistent with existing behaviour, but I'm not sure if this is intended or not.[^2]
[^2]: Y'see, the "ImageToolbar" in the SeparateFrame component is actually _pseudo toolbar_ - it's a collection of buttons arranged to _look_ like the actual ImageToolbar component, and it's _only_ added when the media type is 'image'. That said, it's simple enough for me to spin up a new PR to make the pseudo toolbar consistent for _both_ image and video types, if we want.

### Status

~~WIP~~

Update 2026.06.19: this PR is now ready for review! This PR has been updated "release-ready" notes, marked with 🔶.

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
@@ -36,7 +36,19 @@ const FlipbookViewer = ({
   setOnZoom = DEFAULT_HANDLER,
   subject
 }) => {
-  // const [currentFrame, setCurrentFrame] = useState(defaultFrame)
+  // Prior to Apr 2026, FlipbookViewer used to store the "current frame" value
+  // in local state, but this caused issues with the image toolbar which
+  // expected synced with the Subject Viewer Store's "current frame".
+  
+  // Also, Subject Viewer Store's resetSubject() ensures that the first `frame`
+  // fed to the FlipbookViewer will either be 0, or the default_frame as set by
+  // subject.metadata. As a result, the defaultFrame prop of FlipbookViewer is
+  // now ONLY used to determine the dimensions of the "subject frame". (i.e. to
+  // keep all images & videos displayed in a consistent width & height)
+  // Future devs, it should be fairly safe to that defaultFrame prop to a say
+  // const REFERENCE_FRAME = 0 ; it only matters that all frames are displayed
+  // in a consistent size.
+
   const [playing, setPlaying] = useState(false)
 
   useEffect(() => {
@@ -130,7 +142,7 @@ const FlipbookViewer = ({
 }
 
 FlipbookViewer.propTypes = {
-  /** Fetched from metadata.default_frame or initialized to zero ⚠️ WARNING: not actually implemented as of Apr 2026 */
+  /** Fetched from metadata.default_frame or initialized to zero. ONLY used to determine size of "subject frame" */
   defaultFrame: PropTypes.number,
   /** Determined per mobx store WorkflowStepStore via SubjectViewer. */
   enableInteractionLayer: PropTypes.bool,
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
@@ -18,6 +18,7 @@ const FlipbookViewer = ({
   enableInteractionLayer = false,
   enableRotation = DEFAULT_HANDLER,
   flipbookAutoplay = false,
+  frame = 0,
   invert = false,
   limitSubjectHeight = false,
   move = false,
@@ -25,11 +26,12 @@ const FlipbookViewer = ({
   onReady = DEFAULT_HANDLER,
   playIterations,
   rotation = 0,
+  setFrame = DEFAULT_HANDLER,
   setOnPan = DEFAULT_HANDLER,
   setOnZoom = DEFAULT_HANDLER,
   subject
 }) => {
-  const [currentFrame, setCurrentFrame] = useState(defaultFrame)
+  // const [currentFrame, setCurrentFrame] = useState(defaultFrame)
   const [playing, setPlaying] = useState(false)
 
   useEffect(() => {
@@ -64,8 +66,8 @@ const FlipbookViewer = ({
   // only for the sake of determining a consistent viewer width/height for
   // all media files. 
   // --------------------------------
-  const currentMediaType = subject?.locations[currentFrame]?.type
-  const currentMediaUrl = subject?.locations[currentFrame]?.url
+  const currentMediaType = subject?.locations[frame]?.type
+  const currentMediaUrl = subject?.locations[frame]?.url
   // --------------------------------
 
   const onPlayPause = () => {
@@ -80,7 +82,7 @@ const FlipbookViewer = ({
         <SingleImageViewer
           enableInteractionLayer={enableInteractionLayer}
           enableRotation={enableRotation}
-          frame={currentFrame}
+          frame={frame}
           imgRef={mediaElementRef}
           invert={invert}
           limitSubjectHeight={limitSubjectHeight}
@@ -110,9 +112,9 @@ const FlipbookViewer = ({
         />
       )}
       <FlipbookControls
-        currentFrame={currentFrame}
+        currentFrame={frame}
         locations={subject.locations}
-        onFrameChange={setCurrentFrame}
+        onFrameChange={setFrame}
         onPlayPause={onPlayPause}
         playing={playing}
         playIterations={playIterations}
@@ -122,12 +124,14 @@ const FlipbookViewer = ({
 }
 
 FlipbookViewer.propTypes = {
-  /** Fetched from metadata.default_frame or initialized to zero */
+  /** Fetched from metadata.default_frame or initialized to zero ⚠️ WARNING: not actually implemented as of Apr 2026 */
   defaultFrame: PropTypes.number,
   /** Determined per mobx store WorkflowStepStore via SubjectViewer. */
   enableInteractionLayer: PropTypes.bool,
   /** Function passed from Subject Viewer Store */
   enableRotation: PropTypes.func,
+  /** Passed from the subject viewer store. Determins the current frame of the Subject */
+  frame: PropTypes.number,
   /** Fetched from workflow configuration. Determines whether to autoplay the loop on viewer load */
   flipbookAutoplay: PropTypes.bool,
   /** Passed from Subject Viewer Store */
@@ -144,6 +148,8 @@ FlipbookViewer.propTypes = {
   playIterations: PropTypes.number,
   /** Passed from the subject viewer store. Needed in SingleImageViewer to handle transforming (rotating) the image */
   rotation: PropTypes.number,
+  /** Passed from the Subject Viewer Store */
+  setFrame: PropTypes.func,
   /** Passed from the Subject Viewer Store */
   setOnPan: PropTypes.func,
   /** Passed from the Subject Viewer Store */
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
@@ -1,6 +1,7 @@
 import { Box } from 'grommet'
 import PropTypes from 'prop-types'
 import { useEffect, useState } from 'react'
+import styled, { css } from 'styled-components'
 
 import { useKeyZoom, useSubjectImageOrVideo } from '@hooks'
 
@@ -12,6 +13,10 @@ import FlipbookControls from './components/FlipbookControls'
 const DEFAULT_HANDLER = () => true
 const DEFAULT_WIDTH = 800
 const DEFAULT_HEIGHT = 600
+
+const StyledVideo = styled.video`
+  ${props => props.invert && css`filter: invert(1)`}
+`
 
 const FlipbookViewer = ({
   defaultFrame = 0,
@@ -98,9 +103,10 @@ const FlipbookViewer = ({
         />
       )}
       {currentMediaType === 'video' && (
-        <video
+        <StyledVideo
           autoPlay={false}
           controls={true}
+          invert={invert}
           ref={mediaElementRef}
           src={currentMediaUrl}
           width={viewerWidth}
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
@@ -120,6 +120,7 @@ const FlipbookViewer = ({
           autoPlay={false}
           controls={true}
           invert={invert}
+          preload='metadata'
           ref={mediaElementRef}
           src={currentMediaUrl}
           width={viewerWidth}
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
@@ -38,7 +38,19 @@ const FlipbookViewer = ({
   zoomControlFn = undefined,
   zooming = true
 }) => {
-  // const [currentFrame, setCurrentFrame] = useState(defaultFrame)
+  // Prior to Apr 2026, FlipbookViewer used to store the "current frame" value
+  // in local state, but this caused issues with the image toolbar which
+  // expected synced with the Subject Viewer Store's "current frame".
+  
+  // Also, Subject Viewer Store's resetSubject() ensures that the first `frame`
+  // fed to the FlipbookViewer will either be 0, or the default_frame as set by
+  // subject.metadata. As a result, the defaultFrame prop of FlipbookViewer is
+  // now ONLY used to determine the dimensions of the "subject frame". (i.e. to
+  // keep all images & videos displayed in a consistent width & height)
+  // Future devs, it should be fairly safe to that defaultFrame prop to a say
+  // const REFERENCE_FRAME = 0 ; it only matters that all frames are displayed
+  // in a consistent size.
+
   const [playing, setPlaying] = useState(false)
 
   useEffect(() => {
@@ -134,7 +146,7 @@ const FlipbookViewer = ({
 }
 
 FlipbookViewer.propTypes = {
-  /** Fetched from metadata.default_frame or initialized to zero ⚠️ WARNING: not actually implemented as of Apr 2026 */
+  /** Fetched from metadata.default_frame or initialized to zero. ONLY used to determine size of "subject frame" */
   defaultFrame: PropTypes.number,
   /** Determined per mobx store WorkflowStepStore via SubjectViewer. */
   enableInteractionLayer: PropTypes.bool,
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
@@ -1,6 +1,7 @@
 import { Box } from 'grommet'
 import PropTypes from 'prop-types'
 import { useEffect, useState } from 'react'
+import styled, { css } from 'styled-components'
 
 import { useKeyZoom, useSubjectImageOrVideo } from '@hooks'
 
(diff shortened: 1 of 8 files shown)


goplayoutside3:
> Is there a reason why the FlipbookViewer use local state for keeping track of current frame, instead of using the Subject Viewer Store?

Hey @shaunanoordin just glanced at your PR and thanks for documenting dev considerations. There is a reason FlipbookViewer uses local state for current frame and it has to do with the classifier's Subject Viewer Store running certain functions every time `frame` updates. I'll find the discussion on an FEM PR from a couple of years ago, but essentially because the flipbook is meant to be "played" or "flipped" through frames, we have to be careful about adding frame state to the Subject Viewer Store, which then causes all components that "observe" it to re-render and re-run functions, etc.

goplayoutside3:
> Is there a reason why the FlipbookViewer use local state for keeping track of current frame, instead of using the Subject Viewer Store?

Here's the Github [Issue](https://github.com/zooniverse/front-end-monorepo/issues/5951) that resulted from moving `frame` to the Subject Viewer store for all viewers in PR #5754. In #5952, I moved `frame` back to a local state variable to fix.

The subject viewers have been refactored a lot between that bug incident and now, so it could very well be safe to move `frame` to a store variable instead of local state, but wanted to link the historical discussion here for consideration. If moving `frame` around, there's a lot of use cases to test like "to clone or not to clone" drawn marks, making sure the correct `frame` is recorded when drawing tools are combined with flipbook (and separate frames viewer), and double checking that the classifier's general performance doesn't take a hit when flipbook is "played" or "flipped".

goplayoutside3:
> Why does the Flipbook Viewer's defaultFrame prop have a comment that says "Fetched from metadata.default_frame or initialized to zero" ? This clearly hasn't been the case. What metadata.default_frame?

`metadata.default_frame` is read per subject and set by the researcher who uploaded the subject. The default frame displayed when a subject is loaded in the flipbook is read from `metadata.default_frame` from the Subject store model, or is initialized to zero. See an example Storybook here: https://github.com/zooniverse/front-end-monorepo/blob/0cfa27e8809f921759e3eace91749848eb697383/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.stories.jsx#L126-L128

Link to classifier Story: https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/index.html?path=/story/subject-viewers-flipbookviewer--with-default-frame

goplayoutside3:
> [bug] 🎥 When a frame contains a video, the "Invert Colours" button on the Image Toolbar is enabled but doesn't do anything. Should I implement colour inversion, or disable the "Invert Colours" button, or ignore?

Yes please implement the ability to invert colors when a video is in frame. With single video subjects, the invert button does work on production as long as the project team has enabled "Allow Users To Flip Image Color" in the project builder. Example: https://www.zooniverse.org/projects/goplayoutside3/fem-test-project/classify/workflow/28231

shaunanoordin:
@goplayoutside3 Thank you for answering those questions! Nyerp, looks like I've still got a lot of work left to do here.

> There is a reason FlipbookViewer uses local state for current frame and it has to do with the classifier's Subject Viewer Store running certain functions every time frame updates.

😱 Nyerp! _(nervous collar-pulling)_

Bart simpson nervously pulling his collar

> metadata.default_frame is read per subject and set by the researcher who uploaded the subject

Ah, that confirms what I thought! I saw `subject.metadata.default_frame` used in SubjectStoreViewer's resetSubject(). _Explicitly,_ only Flipbook Viewer ever looks for default_frame. _Implicitly/in practice,_ resetSubject() ensures default_frame is applied to every Subject.

> Yes please implement the ability to invert colors when a video is in frame.

👍 OK, this is a can do!

--------

Here's what I got up to until commit c5517b7:

- In FlipbookViewer,
  - I've replaced the local state "current frame" with the store's "current frame"...
    - **...but now I realise I need to check for performance issues,** and other things! _(nervous collar-pulling intensifies)_ 😱 
  - Video frames now have a functional invert colours button.
- In SeparateFramesViewer
  - All image frames now have functioning Image Toolbars[^1]

[^1]: Prior to this PR, you'd only really notice there's a problem if the first frame of the Subject is a video, because the ImageToolbar's buttons' enabled/disabled-ness only care about the media type of the first frame.

--------

Here's what I'm going to do next:

- In FlipbookViewer,
  - I'm going to read up on the past issues/PRs [(1)](https://github.com/zooniverse/front-end-monorepo/issues/5951) [(2)](https://github.com/zooniverse/front-end-monorepo/pull/5754) [(3)](https://github.com/zooniverse/front-end-monorepo/pull/5952) to understand what the problems caused by using the global currentFrame
  - I'm going to keep an eye out for performance issues when I hit the "Play" button.
  - 🎏 I'm going to check to see if classification.metadata.subject_dimensions has multiple entries. (Previous assumption: there should be one subject_dimensions item for EACH frame) (Current understanding: a classification for multi-image/video subject should have only ONE subject_dimensions item, corresponding to  the first loaded frame) (...I think?)[^2]
  - ▶️ If ANY checks fail, of I have ANY cause of concern, I will revert FlipbookViewer to use local state for "current frame", and figure out an alternative to controlling the ImageToolbar's disabled behaviour. (i.e. change SubjectViewerStore's 'disableImageToolbar' to look for a, uh, `overrideForDisableImageToolbar` or something.) 
- In SeparateFramesViewer
  - I should give Video frames an Image Toolbar, even if it only has an "invert colours" button. For consistency.

[^2]: I might need to check if ImageAndTextViewer and MultiFrameViewer submit one subject_dimensions entry per subject, or one entry per frame.

shaunanoordin:
⚠️ This PR will need to be rebased/merged due to multiple new changes in PR #7322 that affects the FlipbookViewer.

shaunanoordin:
This PR has been rebased on top of 7322, and I'm now re-inspecting this PR based on the [most recent notes](https://github.com/zooniverse/front-end-monorepo/pull/7326#issuecomment-4284721050).

Main question: does the use of global state vs local state for "current frame" affect browser performance?
- How would I measure performance?
- Is there a hybrid alternative to using local currentFrame and global currentFrame?

shaunanoordin:
## Investigation Update: Performance Metrics

Question: does using global currentFrame state have a _noticeable_ negative impact on performance? 

I've set up a performance benchmark to compare two different iterations:

- Case A: using local currentFrame state. (`main` branch, commit af5eca1, [using lib-classifier's Storybook](http://localhost:6006/?path=/story/subject-viewers-flipbookviewer--with-images-and-videos))
  - Case A2: similar to Case A, but use [zooniverse.github.io](https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/?path=/story/subject-viewers-flipbookviewer--with-images-and-videos), whatever the version commit is[^1]
- Case B: using global currentFrame state. (This PR, commit [53be034](https://github.com/zooniverse/front-end-monorepo/pull/7326/commits/53be034d608c9752021ef3e5329ed14c48c52966), [using lib-classifier's Storybook](http://localhost:6006/?path=/story/subject-viewers-flipbookviewer--with-images-and-videos)) 

[^1]: probably [7a4b352](https://github.com/zooniverse/front-end-monorepo/commit/7a4b352e539c5ca0a0446913f8aaf4360c8d3e9a)

**Test 1: Google Chrome Metrics**

See https://developer.chrome.com/docs/devtools/performance

Measurement tools: Chrome 149 + MacBook Pro (Apple M4 Pro, 24 GB, Sequoia)

Measurement method:

- Open Storybook Story for FlipbookViewer -> With Images and Videos
- Use Google Chrome -> Dev Console -> Performance tab
- Set CPU to ~~20x slowdown~~ 6x slowdown
- "Play" the Flipbook for at least 3 full loops 
- Record performance for the full 3 loops
- Observe differences in **frame rate,** since this will matter the most to users. (Look for number of FPS bottlenecks)
- Observe CPU usage. (Look for increases in usage.)

Measurement results:

> **Case A**
>
> image

> **Case B**
>
> image

<details>
<summary>How to read those charts</summary>

tmp-20260617

Horizontal value is time, increasing to the right.

A: FPS chart (the red bar) and CPU chart (yellow and grey waves). When the FPS chart shows a strong/bold/vibrant red section, that means the frame rate dropped for a bit there. That's bad. When the CPU chart's waves get pretty tall (and you see a lot of yellow), it means it's running a lot of the script.

B: what's the CPU is doing?

C: call stack.

</details>

Test conclusions:

- Surprisingly, there's little difference between Case A and Case B. 🤨 

**Test 2: Side-by-Side Feel**

Measurement method:

- Open Case A2 and Case B in two separate windows, side by side.
- Play both Flipbook viewers, complete 3 loops.
- Observe performance of flipbook animation.
- (Note: measurement won't be perfect since Case A2 is on github instead of the local machine, but eh, close enough.)

> image

Measurement results:

- The time for each Flipbook to complete 3 loops of playing was about the same. (Accounting for the small delay as my mouse switches windows) 
- There were no noticeable stutters or slowdowns.
- (Note: this could not tested on a Subject with say a hundred of images to play through, since that would've required deploying Storybook.)  
- (Note: the image and videos were already fetched in previous tests, so those files were retrieved from my disk cache instead of from network calls. This may be worth noting that actual users "playing" a new Subject for the first time might end up fetching a bunch of resources in one go, causing a one-time-slowdown of network traffic _per subject.)_

Test conclusions:

- There's no noticeable difference to the "feel" of playing through media files, for both Case A and Case B.

### Summary

As far as I can tell, there's no significant performance issue between using local state or global state.[^2]

[^2]: I _was_ able to _some_ differences in performance when the CPU throttle was at 20x slowdown, but the measurements were so inconsistent between both Case A and Case B (because small delays in my testing input were magnified 20-fold) that the results couldn't be used.

As such, I propose this PR continues with the current implementation of using global state (store) for managing currentFrame in FlipbookViewer. I'll perform some cleanup and mark this PR ready for review.

HOWEVER, I will also endeavour to create an alternative "use local state when Playing" in a separate "what if?" PR, that can be merged onto this PR if there's a strong argument for maintaining local state for performance. (e.g. if somebody has better performance measurements that shows a distinct difference. Maybe we'd need to crank the test Subjects to use 100 image files?) I'll approach that separately.

shaunanoordin:
## Investigation Update

Delilah made a good point that my performance tests only covered Stroybook Stories, which is a far lighter environment than the actual volunteer's classifying experience. I'm running comparisons for Case A and Case B again, but this time using the full FEM classifying experience.

**Test 3: Google Chrome Metrics on app-project**

See https://developer.chrome.com/docs/devtools/performance

Measurement tools: Chrome 149 + MacBook Pro (Apple M4 Pro, 24 GB, Sequoia)

Measurement method:

- Using **app-project,** open a Chimp & See workflow: https://local.zooniverse.org:3000/projects/sassydumbledore/chimp-and-see/classify/workflow/19267?env=production
- Use Google Chrome -> Dev Console -> Performance tab
- blah blah blah, the rest is the same as Test 1.

Measurement results:

> **Case A**
>
> image

> **Case B**
> 
> image

Test conclusions:

- Compared to Test 1, this test uses a larger Subject (1 video + 15 images, versus 1 video + 4 images) and has a lot more code surrounding the main classifier.
- Performance-wise, there is little difference between _Case A and Case B._
  - The CPU usage and frame rate looks mostly similar.
  - Case B's call stack seems a bit deeper though?
- However, there is a _lot_ of difference between _Test 1 and Test 3._
  - A 6x CPU slowdown is enough to cause way more frame rate hiccups (look for the solid blocks of red in the charts)
  - CPU calls for one frame often overflow into the next frame. 

### Updated Summary

There continues to be little difference between using local state vs global state for currentFrame, so this PR can continue as planned.

That said, we can now see (measure) that there's a _significant_ performance impact when the Flipbook Viewer is used in the full classifying environment - there's space to investigate _why_ there's such a huge difference.[^1]

[^1]: I'll try to find time later to take measurements on lib-classifier's `pnpm dev` mode, so I can compare the spectrum of lightest environment to the heaviest: Flipbook Viewer in storybook story (need to crank up to 1 video + 15 images for consistency) >-> Flipbook Viewer in lib-classifier dev mode >-> Flipbook Viewer in app-project classify page

shaunanoordin:
## PR Update

This PR is now ready for review! I've updated the [main post](https://github.com/zooniverse/front-end-monorepo/pull/7326#issue-4297546168) with full notes on functionality changes, code changes, and testing. (Anything in that post tagged with 🔶 was added as part of the "ready for review" notes)

### Status

Ready for review

shaunanoordin:
Wait, I forgot to edit the video elements to have `preload='metadata'`! One second...

[UPDATE:](https://github.com/zooniverse/front-end-monorepo/pull/7326/commits/973a57daaeebffb9bfb4b5f9baa533a0896e45e8) OK, done. 👌

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
```diff
@@ -13,25 +14,40 @@ const DEFAULT_HANDLER = () => true
 const DEFAULT_WIDTH = 800
 const DEFAULT_HEIGHT = 600
 
+const StyledVideo = styled.video`
+  ${props => props.invert && css`filter: invert(1)`}
```
mcbouslog, line 18:
```suggestion
  ${props => props.$invert && css`filter: invert(1)`}
```

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/FlipbookViewer.jsx
```diff
@@ -100,9 +116,11 @@ const FlipbookViewer = ({
         />
       )}
       {currentMediaType === 'video' && (
-        <video
+        <StyledVideo
           autoPlay={false}
           controls={true}
+          invert={invert}
```
mcbouslog:
```suggestion
          $invert={invert}
```

mcbouslog (approved):
This LGTM! Thank you @shaunanoordin for the investigation, performance metrics, and changes!!

I tested multiple location subjects with drawing marks on different frames, cloned and not cloned, on this staging workflow - https://local.zooniverse.org:3000/projects/markb-panoptes/fem-drawing-tools/classify/workflow/3363?env=staging&demo=true - without issue.

I've left a minor suggestion in-line related to styled-component props to prevent the console warning "Received `false` for a non-boolean attribute `invert`."

Unrelated to these changes, and I'm not even sure if there's any related follow up, but I noticed with video/images subject, in separate frame mode, the video doesn't show the toolbar with the (now working) invert button for the video.

shaunanoordin:
Thank you for reviewing this, Mark! 😄 👍 I've added your suggestion, thank you for spotting + fixing that warning!

Side note: omg I thought I merged this before flying to Chicago, my brain must've been friend. 😱

PR #7339, opened 2026-04-22
lib-classifier: Hide line controls while user is drawing

sktw:

## Package

- lib-classifier

## Context

Currently, the line controls are shown as soon as the user starts to draw. If the controls obscure the part of the subject that the user wants to mark, they have to stop drawing and move the controls before continuing. This PR changes the behaviour so that the controls are hidden whenever the user is drawing. The position of the controls is preserved between hide and show events.

This change is most beneficial for smaller screens / subjects, where it's more likely that the controls will interfere with drawing.

## Describe your changes

- Conditionally render the` LineControls` when `isDragging = false` on the `FreehandLine` mark model.
- Update tests.

### Project

1. Draw a freehand line. Check that the controls are not shown until the drag ends.
2. Extend the line. Check that the controls are hidden, and reappear when the drag ends.
3. Move the controls to the right. Extend the line. Check that the controls reappear on the right when the drag ends.

### Storybook

- http://localhost:6006/?path=/story/drawing-tools-freehand-line--drawing
- http://localhost:6006/?path=/story/subject-viewers-singlevideoviewer--with-drawing&args=toolType:freehandLine

# Checklist

## General

## General UX

### New Feature

### DIFF
--- a/packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.jsx
+++ b/packages/lib-classifier/src/plugins/drawingTools/experimental/components/LineControls/LineControls.jsx
@@ -1,4 +1,5 @@
 import { forwardRef, useState } from 'react'
+import { observer } from 'mobx-react'
 import PropTypes from 'prop-types'
 import styled, { withTheme } from 'styled-components'
 import { Tooltip } from '@zooniverse/react-components'
@@ -199,6 +200,16 @@ const LineControls = forwardRef(function LineControls({
 
   const buttons = (showDeleteButtons) ? deleteButtons : defaultButtons
 
+  /*
+   * Hide the controls if the user is drawing.
+   * This is handled here rather than in LineControlsContainer to preserve
+   * the position of the controls.
+  */
+  
+  if (mark.isDragging) {
+    return null
+  }
+
   return (
     <StyledSVG 
       width={`${WIDTH}`} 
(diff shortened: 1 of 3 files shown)


goplayoutside3 (commented):
Apologies for the delay in review while getting other dev efforts wrapped up.

@sktw this is a great idea to hide the controls when drawing a line. I tested this branch via lib-classifier's Storybook and the test project link you provided. Hiding the controls works most of the time, but about 1 out of 4 marks drawn results in `mark.isDragging` being stuck as `true` even when my mouse has left the drawn line. Because the LineControls are hidden when `mark.isDragging`, there's no way to delete the drawn mark when its `.isDragging` prop is stuck as `true`.

I'm not sure if this is the result of a broader problem with handing `isDragging` in the drawing tools, or related to directly reading the prop in LineControls. I'll provide a screen recording below:

https://github.com/user-attachments/assets/523246dd-e2ad-4c50-87c3-5e282e0ec854

sktw:
@goplayoutside3 I'm afraid that I've been unable to recreate this bug myself in either Chrome or Firefox. The only 'buggy' behaviour I've observed is that if you draw a line too quickly, it is deleted when you release the mouse. The reason is that it fails the minimum-points check in `finish`, presumably intended to remove very short lines drawn by accident. It does however ensure that the user has to draw the lines carefully! 

As you say, we need to know whether this is a problem in the FreehandLine model or in the LineControls component. From the recording I think that it may be the former. At about 10s in you draw a 'u' shaped line at the bottom of the screen, and the line controls appear as expected (`isDragging` must be `false` on this line). You then click the 'bad' line above to make it active, and the controls disappear. Changing the active mark rerenders the line controls (because it's passed in a prop), so `isDragging` must presumably still be `true` on bad line's mark model.

goplayoutside3:
Interesting! I'll have to dig in more to find what conditions caused the bug in my screen recording. Also agreed it's likely due to something in the Freehandline model.

@kieftrav when you have some time available, could you run this change locally to see if you can replicate the bug in my screen recording?

PR #7194, opened 2026-01-23
Classifier: add debug messages when "This Workflow Does Not Exist", add error message when no workflow selected

shaunanoordin:
## PR Overview

Supports: #7193 
Package: lib-classifier

This PR updates the Classifier with more verbose error messages.

New behaviour:
- 🆕 If NO workflow is selected[^1] => a new error message, "No workflow selected", is displayed.
- If an INVALID workflows is selected[^2] => the standard "This workflow does not exist or you do not have permission to view it" message is displayed, PLUS 🆕 additional debug data.
  - 🆕 Debug data includes: selected workflow ID, allowed workflow IDs, project ID, ~~project updated time,~~[^3] url, timestamp.
  - Q: why? A: if a volunteer encounters this message, we want to know WHY they received it. The debug data includes everything need (hopefully) to understand why `allowedWorkflowID` is null. (See 7193)
- Otherwise, display the Classifier if everything has loaded, or show "Loading the Classifier" if it's still loading.

Current behaviour:
- If NO workflow is selected[^1] => "This workflow does not exist or you do not have permission to view it"
- If an INVALID workflow is selected[^2] => "This workflow does not exist or you do not have permission to view it"
- Otherwise, display the Classifier if everything has loaded etc etc.

[^1]: For example, when a volunteer goes to the /classify page of a project with multiple workflows, the classifier will load in the background with no workflow selected, while the "pls select your workflow" modal appears in the foreground.
[^2]: For example, when a volunteer tries to look at a retired workflow. NOTE: if you view a workflow that does not exist, or exists for a different project, then you get different results depending on whether you're looking at it on lib-classifier or app-project. On the former, you'll get the "This workflow does not exist" error message. On the latter, you'll get a 404.
[^3]: Apparently, we're not capturing project updated time. Hmm, this might be worth adding if we want to test that cache problem.

### Testing

Test on localhost with _app-project._

Testing Steps (project with one workflow):
- Choose a test project with one workflow, e.g. Snapshot Wisconsin.
- Go to the Classify page: https://local.zooniverse.org:3000/projects/zooniverse/snapshot-wisconsin/classify?env=production
  - You should be auto-directed to the one active workflow, e.g. https://local.zooniverse.org:3000/projects/zooniverse/snapshot-wisconsin/classify/workflow/30961?env=production
  - You should see the Classifier functioning normally.
[...]

PR #7345, opened 2026-04-24
Classifier: add "No Workflow Selected" message

shaunanoordin:
## PR Overview

Package: lib-classifier (but test with app-project)
Improves on #7194

This is a small UI enhancement to the Classifier, which adds a better status message when no workflow is selected.

Currently, when you visit the root /classify page of a project (with either 0 or 2+ active workflows), you will see the error message/status message _"This workflow does not exist or you do not have permission to view it."_ Which is kinda wrong, since there's no _"this workflow"_ to begin with.

With this PR, the status message is a bit more accurate to the situation:

image

Behaviour:
- When visiting the root /classify page as a normal (non-owner) volunteer... (e.g. https://local.zooniverse.org:3000/projects/darkeshard/test-project-2026/classify?env=staging)
  - 🆕 ...if a project has 2 or more active workflows, the status message is "No workflow selected", and a Select Workflow popup will appear.
  - 🟰 ...if a project has exactly 1 active workflow, user will be redirected to that workflow.
  - 🆕 ...if a project has 0 active workflows, the status message is "No workflow selected", and a Select Workflow popup will appear (albeit with the message "nothing to Classify").
- 🟰 When visiting the /classify path of a workflow that doesn't exist[^1] OR the user doesn't have access to[^2], error message will be "This workflow does not exist or you do not have permission to view it."[^3]
- When visiting the /classify path of a valid, active workflow, Classifier will function as normal.

(🆕 indicates new behaviour introduced with this PR. 🟰 indicates no change in behaviour.)

[^1]: in practice, if you visit a workflow that doesn't exist - e.g. https://frontend.preview.zooniverse.org/projects/darkeshard/test-project-2022/classify/workflow/123456789 - then you'll just get a 404 page, since app-project will perform the redirect before lib-classifier can even load to show the error message.
[^2]: to be specific, this means that the workflow exists BUT is not active, and the user isn't logged in as a project owner or collaborator or tester. In practice, this is usually the only way you'll see the "workflow doesn't exist" error message.
[^3]: OK, one more caveat - if you visit a workflow that DOES exist but belongs to a DIFFERENT project, you'll also get a 404. This is a weird edge case though.

### Testing 

Test on localhost with **app-project.**

- Select a test project: https://www.zooniverse.org/lab/2026/workflows?env=staging
  - (For your testing convenience, please feel free to edit this Test Project 2026. Activate/deactivate workflows any which way you like.)
- Edit the test project so it has (say) **3 active workflows.**
  - Visit the root /classify page: https://local.zooniverse.org:3000/projects/darkeshard/test-project-2026/classify?env=staging
  - Ensure that you see 1. the "No workflow selected" message (in the back) and 2. the Select Workflow popup with 3 workflows.
      image
  - BONUS: choose any one workflow and try classifying on it.
- Edit the test project so it has **exactly 1 active workflow.**
  - Visit the root /classify page.
  - Ensure that you're redirected to that active workflow.
      image
- Edit the test project so it has **0 active workflows.**
  - Visit the root /classify page.
  - Ensure that you see 1. the "No workflow selected" message (in the back) and 2. the Select Workflow popup with "Nothing to Clasify".
      image
  - BONUS: try classifying on this workflow.
- Visit a workflow that EXISTS but is NOT active. (And make sure you're not logged in as a project owner!)
  - e.g. https://local.zooniverse.org:3000/projects/darkeshard/test-project-2026/classify/workflow/3913
  - Ensure that you see the "This workflow does not exist or you do not have permission to view it." error message
      image

NOTE: if you visit a workflow that actually DOES NOT EXIST, then you'll immediately go to a 404 page. You won't get a chance to see the "This Workflow doesn't exist" error message, ironically enough.

### Status

Ready for review. Super low priority, though.

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/ClassifierContainer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/ClassifierContainer.jsx
@@ -229,16 +229,22 @@ export default function ClassifierContainer({
   try {
     return (
       <Provider classifierStore={classifierStore}>
-      {!allowedWorkflowID && userHasLoaded ?
-        <Paragraph>This workflow does not exist or you do not have permission to view it.</Paragraph>
+        {!workflowID ? // If the workflow ID isn't specified, it's usually because the user is at the root /classify path.
+          <Paragraph>No workflow selected.</Paragraph>
+
+        : !allowedWorkflowID && userHasLoaded ?
+          <Paragraph>This workflow does not exist or you do not have permission to view it.</Paragraph>
+
         : classifierIsReady ?
-           <Classifier
-              onError={onError}
-              showTutorial={showTutorial}
-              subjectSetID={subjectSetID}
-              subjectID={subjectID}
-              workflowSnapshot={workflowSnapshot}
-            /> :
+          <Classifier
+            onError={onError}
+            showTutorial={showTutorial}
+            subjectSetID={subjectSetID}
+            subjectID={subjectID}
+            workflowSnapshot={workflowSnapshot}
+          />
+
+        :
           <Paragraph>Loading the Classifier</Paragraph>
         }
       </Provider>


goplayoutside3 (approved):
Great enhancement to the classify page. I tried out the various workflow scenarios on your test project 👍 

So far, these error messages haven't been setup with translations. However, this small PR might be a good place to add them to lib-classifier's English dictionary so they can be pulled into Lokalise. Would you like to do so in this PR? Or I'm happy to follow up with a secondary PR after this one is merged.

shaunanoordin:
Thanks for the review, Delilah! 👍 Let's get the translations in a separate PR; I'll be happy to take a stab at that.[^1]

[^1]: it's been a while since I last worked with translations, so it's a good idea for me to exercise these muscles to make sure I'm up to date.

PR #7405, opened 2026-05-05
lib-classifier: Refactor SVGContext to handle resize

### LINKED ISSUE #7231: Drawing marks and related controls not scaling on window resize
Per #7223, 

...One issue does remain, although it involves an uncommon scenario:

1. Navigate to https://fe-project-branch.preview.zooniverse.org/projects/skeogh17/snail-scribe-archive/classify/workflow/29065?demo=true
2. Resize the browser window, making it, say, roughly half the width of the screen.
3. Draw a rectangle
4. Maximize the browser window. Notice that the drag handles and delete button are now too large.

The problem is that the resize doesn't trigger a render, so the scaling is not updated for the new canvas size. This could be handled by adding a listener for the window's `resize` event, but, given the rarity of this scenario, it may not be worth it.

_Originally posted by @sktw in https://github.com/zooniverse/front-end-monorepo/issues/7223#issuecomment-3927456730_

sktw:

## Package

- lib-classifier

- Closes #7231

## Context

The `useScale` hook does not currently handle resize events caused by either a change window size (issue #7231) or layout. It also recomputes the scale on every render for each consumer.

This PR adds a resize detector (using the existing react-resize-detector package) to the root svg element, and stores the computed scale in the SVGContext. All consumers (per canvas) then read the scale from the shared context.

The resize detector is attached to the root svg element because it's the only element with a defined [content box](https://github.com/w3c/csswg-drafts/issues/4032#issuecomment-502475833). Fortunately, it is also easier to compute the viewport scaling from this element because it has a viewBox.

## Describe your changes

- Remove `rotation` and `transformMatrix` from SVGContext (no longer needed), and add `scale`
- Add new SVGCanvas component to encapsulate the canvas element, SVGContext provider and `hasMounted` logic introduced in #7223
- Add resize detector to SingleImageViewer and VideoWithDrawing to respond to width change on root svg element 
- Refactor SingleImageCanvas, VideoWithDrawing and SubjectGroupViewer to use SVGCanvas, providing an appropriate implementation of `scale` for each
- Refactor existing `useScale` consumers to read `scale` from context via a new `useSVGContext` hook
- Remove the `useScale` hook
- Add tests

### Project

Confirm fix of #7231:

1. Draw a rectangle. 
2. Apply zoom and rotation, and resize the browser window. Check that the drag handles remain the same size.

### Storybook

Stories for refactored components:

http://localhost:6006/?path=/story/subject-viewers-singleimageviewer--pan-and-zoom
http://localhost:6006/?path=/story/subject-viewers-singlevideoviewer--with-drawing&args=toolType:ellipse
http://localhost:6006/?path=/story/subject-viewers-subjectgroupviewer--default

All of the individual drawing tool stories could also be checked on different device sizes, if desired. One thing to bear in mind is that many of the control elements (e.g. [DeleteButton](https://github.com/zooniverse/front-end-monorepo/blob/c68001537ae1d797cd2eae78abae9fffdc798b5b/packages/lib-classifier/src/plugins/drawingTools/components/DeleteButton/DeleteButton.jsx#L34)) have a window-width breakpoint below which the size of the control reduces.

### Further development

Further to the comment above regarding the size breakpoints for the drawing control elements. It would be straightforward to add the width of the root svg element to the context, and use it define the breakpoints instead of the window / screen width. This would be beneficial for separate frames in particular, where the width of each viewer can be much smaller than the window.

## Checklist

### General

### General UX

### Bug Fix
- [ ] The bug is fixed

### Refactoring


sktw:
This is ready for review, but there is an administrative task outstanding. The react-resize-detector package is currently used by app-project and lib-react-components, but I haven't managed to find the right command to add it to lib-classifier. I've tried

```pnpm add react-resize-detector@~9.1.0 --filter @zooniverse/classifier```

but that resulted in a number of peer dependency errors and multiple additional changes to the lockfile.

goplayoutside3:
Thanks @sktw for the PR! We'll assign someone to review it, likely in the next week-ish.

What I typically do to install a new dependency or add a dependency already in the monorepo to another package/workspace is to double check the version (you mentioned is ~9.1.0 in app-project and lib-react-components), manually add `"react-resize-detector": "~9.1.0"` to lib-classifier's package.json, and then run `pnpm install` at the monorepo's root to update its pnpm-lock.yaml. I have no idea if that's the most efficient way to handle `pnpm` workspaces, but let me know if you give that a try and if it works.

sktw:
@goplayoutside3  Thanks - the problem seems to have been a pending update to next.js, which for some reason didn't get installed when I ran `pnpm run clean-install`. Running `pnpm install` at the project root has updated it.

I now have an updated pnpm-lock.yaml with react-resize-detector added to lib-classifier and a couple of new deprecation warnings for uuid <= v10. If that sounds right, I'll push an update to the PR branch.

goplayoutside3:
@sktw Yes that sounds right. I see the same warnings when I locally installed `react-resize-detector` in lib-classifier.

goplayoutside3 (approved):
This looks great. Thank you @sktw for the PR!

Issue #7330, opened 2026-04-20
Add partner logos to ZooHeader


seanmiller26:
A request to more prominently display our partner institutional logos. 

Match the layout of the recently implemented site footer group of logos (issue #7051), but spaced 20px apart and sized at 45px square for each logo. The logo group should always remain centered.

<img width="800" alt="ZooHeader Institutional logos" src="https://github.com/user-attachments/assets/79db4a50-b42e-4814-9ed3-d1683bdbb620" />

Mobile stacks on top of the existing links and gets an added 5px top padding.

Responsive behavior: transition to the mobile header when the horizontal gap between the logo and navigation links reaches ≤20px. That browser width will be different depending on a number of factors, including whether or not you are logged in, or are displaying the Admin option, as seen below.

<img width="1645" height="252" alt="Image" src="https://github.com/user-attachments/assets/71545173-5a8c-40a4-911a-fd04e2b53f9b" />

---

<img width="1104" height="527" alt="Image" src="https://github.com/user-attachments/assets/e621ad8e-5721-4a7b-af48-5682660b2cbd" />

---

<img width="369" height="537" alt="Image" src="https://github.com/user-attachments/assets/a8035733-f072-49b8-971d-f4d0a7af5042" />

---

Figma https://www.figma.com/design/6rXVi6arrGTmoM08x9tpex/UI-Kit?node-id=2-77&t=yIErfQCzHxWwlc3a-1

goplayoutside3:
A quick dev note while I'm thinking about this UX:
> Responsive behavior: transition to the mobile header when the horizontal gap between the logo and navigation links reaches ≤20px. That browser width will be different depending on a number of factors, including whether or not you are logged in, or are displaying the Admin option, as seen below.

Initially I thought we could use a [container query](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries) and listen for the size of the `div` that contains the three logos. With each logo 45px in height, the widths are 68px + 68px + 44px plus 20px padding in between ( so + 40px) and plus 20px on either side (so another + 40px) = container width of 260px is the container "breakpoint". However, the design requires a change of the overall ZooHeader _layout_ when comparing desktop to mobile, and heavily depends on whether a user a signed-in or signed-out, the text length of their username, and the chosen language. This type of responsiveness requires Javascript rather than native browser CSS.

[...]

PR #7407, opened 2026-05-07
lib-react-components: Refactor handling of ZooHeader institutional logos to CSS instead of a resize observer

goplayoutside3:
## Package
lib-react-components

Related to https://github.com/zooniverse/front-end-monorepo/issues/7330

When navigation between pages on production such as `/about` and `/projects`, the logos in ZooHeader layout-shift after every route navigation because the component-width measurement via Javascript can only be done client side. This was the initial solution to #7330 because the desktop vs mobile designs have different layouts for the logos. However, the repaint of ZooHeader's height and the logos position is an annoying UX and not worth the trade off of measuring exact widths of ZooHeader's components.

## Describe your changes
- Define a breakpoint in `rem` units for the logos layout breakpoint.

## How to Review
- Run one of the apps locally like app-root and navigate between pages. The logos position should not change between routes.
- Run lib-react-component's Storybook and look at the various cases in ZooHeader combine with various screen sizes simulated by Storybook.

### General

### Refactoring

### DIFF
--- a/packages/lib-react-components/src/ZooHeader/ZooHeader.jsx
+++ b/packages/lib-react-components/src/ZooHeader/ZooHeader.jsx
@@ -14,6 +14,18 @@ import NarrowMainNavMenu from './components/NarrowMainNavMenu/NarrowMainNavMenu'
 import UserNavigation from './components/UserNavigation/UserNavigation'
 import ZooniverseLogo from '../ZooniverseLogo/ZooniverseLogo'
 
+const defaultHandler = () => true
+
+// Best guess for when the nav menus should collapse to hamburger style so as not to overlap horiztonally
+// This could be refactored to use CSS and rem units. The legacy version of ZooHeader was initially built
+// with detecting client side viewport width to define isNarrow as a prop passed to the nav menus.
+const narrowNavBreakpoint = 990
+
+// Best guess for max screen width when a user is signed-in, has admin mode enabled, or viewing a locale with longer than average words like German
+// while still accounting for rem if someone has changed their browser font size. This resize cannot use CSS container queries because the space in
+// between the two navs in ZooHeader doesn't have a defined width and that's incompatible with a CSS container query.
+const logosBreakpoint = '80rem'
+
 export const StyledHeader = styled(Box)`
   color: #a6a7a9; // light-5
   font-size: 1em;
@@ -24,14 +36,29 @@ export const StyledHeader = styled(Box)`
   }
 `
 
+const TopLogosContainer = styled(Box)`
+  @media (width >= ${logosBreakpoint}) {
+    display: none;
+  }
+`
+
 const Relative = styled(Box)`
   position: relative;
+  padding-block: 20px;
+
+  @media (width < ${narrowNavBreakpoint}px) {
+    padding-block: 10px;
+  }
 `
 
 const Absolute = styled(Box)`
   position: absolute;
   left: 0;
   top: 0;
+
+  @media (width < ${logosBreakpoint}) {
+    display: none;
+  }
 `
 
 export const StyledLogoAnchor = styled(Anchor)`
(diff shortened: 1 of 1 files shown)


shaunanoordin (approved):
## PR Review

Package: lib-react-components (tested on app-project)
Affects: Zooniverse Header

This PR changes how the Zooniverse Header manages the layout of its elements - particularly in regards to the institution logos.

- Prior to this PR, the ZooHeader uses useResizeDetector() to dynamically detect the size of the left menu bits + right menu bits to deteremine when the institution logo in the middle would be squished, then adjust accordingly.
  - Problem: the calculation takes time and has to be re-triggered when traversing between certain pages, causing a "hiccup" in the early render on page load.
- With this PR, the ZooHeader uses CSS breakpoints.
  - Benefit: no more hiccups on page transitions/page load.
  - Cost: the CSS breakpoints optimises for 1. the default English locale, 2. users having reasonable usernames (or display names), and 3. users not modifying their default browser (font?) zoom too much. If any of these assumptions aren't met, small layout quirks appear.

image

image

> The layout for the ZooHeader elements remains the same. The "show either logos_top XOR logos_middle" rule remains the same. The change is how/when the switch between logos_top and logos_middle is calculated.
>
> ```
>                   [logos_top]
> [left_menu_bits] [logos_middle] [right_menu_bits]
> ```

Functional changes:

- ZooHeader: now switches between displaying the "top logos" and "middle logos" elements at 1280px viewport width.

Code changes:

- ZooHeader.jsx: pretty straighforward change from JS logic to CSS breakpoints.

### Dev Notes

The implementation of this PR is ultimately a choice between two different tradeoffs - either accept the hiccups, or accept that layout gets quirky in certain situations - and this PR chooses the latter.

I'm cool with this, but I will note some of the more egregious we may want to look into in the future in case, I dunno, CSS suddenly adds some functions that lets us solve the problem of displaying every menu element _the way the visual design demands it_ without having to resort to any tradeoffs.

**Quirky Layout 1: Russian Language**

I couldn't find a German project to test with, but I can definitely see the issue of overlapping menu bits when the Russian language is chosen, because the word are visibly longer.

image

**Quirky Layout 2: Super-long User Names**

In theory it's possible that a user with a looooooong username could throw the layout out of whack, but I never tested this beyond editing my browser details. For all I know, Panoptes has a character limit on display names. Anyway, this is what it'd look like:

image

### Testing

Test on localhost with macOS + Chrome

Standard tests: does the ZooHeader look good in normal conditions?

- Run `pnpm start` (because pnpm dev keep bringing up that dang screen-covering warning overlay that prevents me from seeing the header at load.)
- Select a test project, e.g. Galaxy Zoo https://local.zooniverse.org:3000/projects/zookeeper/galaxy-zoo
- Scenario 1: standard browser window view
  - Scenario 1a: view the page while logged in.
    - Observe that the header displays the left menu bits, the institute logos, and the right menu bits.
    - Resize the window, observing various breakpoints (notably at width=1280px)
    - Confirm that left menu bits, the institute logos, and the right menu bits are always visible/usable.
  - Scenario 1b: view the page while logged out.
    - Repeat all steps in scenario 1a, while logged out.
- Scenario 2: Chrome device emulation mode
  - Enable Chrome device emulation mode - responsive view, 50% zoom.
  - Repeat scenarios 1a and 1b, except now go for even more extreme breakpoints (e.g. width=100px, or width=2000px)

Note: layout breaks at below ~333px, but this is to be expected.

image

Edge case tests:

- Run `pnpm start`, select a test project.
- Scenario 3: different languages
  - Select a series of different languages. Notably, try Russian, French, and Chinese.
  - Repeat scenarios 1a and 1b.
  - Observe how for certain languages, visual length/width of words are very long (Russian), while others are shorter than usual (Chinese).

### Status

LGTM. 👍 As noted, there will be quirky layout scenarios that trigger when the baseline assumptions created by the CSS breakpoints aren't met, but this is a fine tradeoff for me.

I say we get this merged first to solve the "hiccup" issue.

_Note: I'm approving this first so this PR can be merged. I'll update my PR review later for documentation completness (19:47 BST)_

_Note: OK, documentation for this PR review is complete! (22:19 BST)_

goplayoutside3:
Thank you Shaun! If needing to test translated content in the future, you can always add `/de/` after `/projects` to any project URL and see the page in German with at least the platform-level translations.

PR #7425, opened 2026-05-13
Add zooniverse:deployed_app as custom metadata to the apps

goplayoutside3:
## Package
app-project
app-root

## Describe your changes
This is an enhancement to the Next.js apps' metadata to help with debugging CDN caching or preparing to switch routes from PFE to FEM via `curl`.

## How to Review
This PR has no effect on the functionality of either app. It adds `zooniverse:deployed_app` as a custom metadata property in the `<head>` of app-project (fe-project) and to the `<body>` of app-root (fe-root). The latter can't inject custom metadata into the `<head>` due to [Metadata Streaming](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#streaming-metadata) in Next.js's App Router since version 15+.

attn: @zwolf happy to discuss if this is actually helpful for sending requests to the deployed apps. See note ^ about `<head>` vs `<body>`.

### General

### DIFF
--- a/packages/app-project/src/components/Head/Head.jsx
+++ b/packages/app-project/src/components/Head/Head.jsx
@@ -51,6 +51,7 @@ function Head (props) {
 
       <meta name='zooniverse:deployed_commit' content={process.env.COMMIT_ID} />
       <meta name='zooniverse:deployed_ref' content={process.env.GITHUB_REF_NAME} />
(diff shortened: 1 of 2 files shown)


zwolf (approved):
This is great. For curling purposes, it would still be nice to have these values in response headers so that they can be read without grepping through the single-line, minified content. But this is perfect for confirming, thanks!

PR #7434, opened 2026-05-18
app-project: Modify addQueryParams to handle env and demo, only

goplayoutside3:
## Package
app-project

Related to https://github.com/zooniverse/front-end-monorepo/pull/7281

## Describe your changes
- Modify `addQueryParams()` to handle only the app-wide params `env` and `demo`. 
- Demo mode is really only for use in the classifier, but helpful when previewing the project pages on production and navigating to/from the classifier when you don't want to accidentally submit a test classification.
- This PR is toward building the project stats page because the stats page will use query params in the URL to track date ranges and we don't want those query params to carry over to other project pages.

## How to Review
- Use a test project such as https://localhost.zooniverse.org:3000/projects/brooke/i-fancy-cats. Add query params to its URL.
- When hovering over or clicking a link in the ProjectHeader, only the `env` and `demo` params should carry over. Any other param such as `?start_date=2026-05-18` should be deleted from the URL.

### General
- [x] You can `pnpm panic && pnpm bootstrap`

### General UX


packages/app-project/src/helpers/addQueryParams/addQueryParams.js
```diff
@@ -1,7 +1,26 @@
 function addQueryParams(url) {
   const isBrowser = typeof window !== 'undefined'
-  const query = isBrowser ? new URLSearchParams(window.location.search).toString() : ''
-  return query.length ? `${url}?${query}` : url
+  if (!isBrowser) return url
+
+  const query = new URLSearchParams(window.location.search)
+  const env = query.get('env')
+  const demo = query.get('demo')
+
```
mcbouslog, line 8: If I go to https://local.zooniverse.org:3000/projects/markb-panoptes/sltp-test-project-staging/talk/subjects/252638?env=staging and hover over a comment, the "Go to Comment" link has two `?`s, like `https://local.zooniverse.org:3000/projects/markb-panoptes/sltp-test-project-staging/talk/275/434?comment=2068?env=staging` instead of I think the expected `&` between query params. I think `comment` could be added similar to `env` and `demo`?

goplayoutside3: `comment` is a specific add-on query param for linking to a Talk comment. It's not an app-wide param like `env` and `demo` (especially `env`).

I think the solution here would be to create a specific helper function for FEM's SLTP > SubjectTalkData > TalkComment. I'll do so and re-request your review.

goplayoutside3: I added a specific helper in TalkComment for adding the `comment` param to certain links. Tested with https://local.zooniverse.org:3000/projects/markb-panoptes/sltp-test-project-staging/talk/subjects/252638 when in dev mode and added `?env=staging` in production mode.

I think in the future if there are more special cases for query params in app-project we could generalize `addQueryParams()` to take another variable and add it to the href if defined, but for now the one UI component with `comment` is ok with a specialized helper function.

mcbouslog (approved):
Reviewed code, tested locally - LGTM!

`addCommentQueryParam` fixes subject talk page links 👍 . I don't think there are any similar situations, and as noted we can revisit when necessary.

PR #7442, opened 2026-05-20
Custom calendar grommet bug

goplayoutside3:
This PR is a draft because there's a bug in Grommet's Calendar component when trying to control its `reference` externally (the month/year in view). See more in the review comments.

## Package
app-project

Builds on top of https://github.com/zooniverse/front-end-monorepo/pull/7281

Towards a request to make the "Custom" option in the stats bar chart date range easier to use when someone wants to see a date far from the current date. Grommet Calendar's default behavior is to show the current month/year, and the arrows used to navigate to the next or previous month only work one month at a time. That's not very useful for navigating to several years ago for a custom stats date range.

## Describe your changes
- Add two Select components to the CalendarModal. Their options are months of the year based on `locale` and years based on how long since the Zooniverse project was launched (or created at if not launched).
- Set `reference` for the calendar if the arrows are used for navigation or if the dropdowns are selected.
- Upgrade to the latest version of Grommet 2.54.0

## How to Review
Do not. This is just a demo PR.


packages/app-project/src/screens/ProjectStatsPage/components/ChartContainer/CustomCalendar.jsx
```diff
@@ -71,20 +116,37 @@ function CustomCalendar({
...
+            activeDate={reference}
+            animate={false}
             bounds={[getStatsDateString(launchDate), todayUTC]}
-            dates={[calendarStart, calendarEnd]}
+            // dates={[calendarStart, calendarEnd]}
+            // range='array'
+            reference={reference}
             onSelect={handleCalendarChange}
-            range='array'
+            onReference={date => handleOnReference(date)}
```
goplayoutside3, line 146: Handling `reference` is based on a Grommet Storybook example: https://storybook.grommet.io/?path=/story/visualizations-calendar-dual--dual. In that example, one Calendar controls the other, so in theory a Calendar's `reference` can be externally controlled to shift which month/year is visible. However, the code in this PR introduces a lot of UI bugs when Calendar is passed `reference` and `dates` and `onSelect` props at the same time. The bugs are caused by whatever is happening internally in Grommet's Calendar component.

I even tried to copy paste all of the code directly from the Calendar Storybook "Dual" example into the project stats page's CustomCalendar component, just to see if it would work as expected. It does not. I tried at Grommet versions 2.47.0 and 2.54.0. I'm not sure what magic is powering the live Grommet Storybook, but it can't be reproduced here so I'll look for a different solution.

goplayoutside3:
Closing in favor of https://github.com/zooniverse/front-end-monorepo/pull/7443

PR #7452, opened 2026-05-21
Project Stats: fix crash when start_date is after end_date

shaunanoordin:
## PR Overview

Package: app-project
Affects: Project Stats pages

This PR fixes an issue on the Project Stats page, which crashes if the selected date range is invalid - specifically, in the case when the start date comes _after_ the end date.[^1]

[^1]: the time cops disapprove of this time crime.

Issue: "If start_date is after end_date, then page crashes."

- Can be replicated by going to the project's stats page, opening the time range dropdown, choosing Custom, and setting (say) end date = 1 Jan '26, start date = 1 May '26
- Page crashes with console error: "Cannot read properties of undefined (reading 'length')"
- Error is in ChartContainer.jsx: when useProjectStats() is called with wonky date ranges, it returns `data` with the value `{error: 'Date range entered is not valid'}` instead of `undefined`, as the code expects. This causes the use of `data?.data.length` in the code to break.  

Solution: 

- Add existence checks.
- Page now fails gracefully with the error message: "Invalid date range, start date must be before end date"

image

### Testing

- [Choose a project, go to its stats page.](https://local.zooniverse.org:3000/projects/zookeeper/galaxy-zoo/stats?env=production)
- Open time range dropdown, choose Custom, set end date before start date.
- Page should now fail gracefully.
- You should be able to continue using the stats page, e.g. choosing a valid date range this time.

### Status

Ready for review! I won't be in on Friday, so please feel free to approve+merge or discard this PR while I'm away.

### DIFF
--- a/packages/app-project/src/screens/ProjectStatsPage/components/ChartContainer/ChartContainer.jsx
+++ b/packages/app-project/src/screens/ProjectStatsPage/components/ChartContainer/ChartContainer.jsx
@@ -338,10 +338,10 @@ function ChartContainer({ workflows }) {
             the data because Grommet `detail` uses `series.render` method. Crashes the page if it can't
             find the expected `period`.
           */}
-          {data?.data.length > 0 ? (
+          {data?.data?.length > 0 ? (
             <BarChart data={data?.data} dateRange={{ startDate, endDate }} type={type} />
           ) : null}
-          {!loadingOrValidating && data?.data.length === 0 ? <EmptyPlaceholder /> : null}
+          {!loadingOrValidating && data?.data?.length === 0 ? <EmptyPlaceholder /> : null}
         </Relative>
       </ContentBox>
     </>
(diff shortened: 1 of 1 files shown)


goplayoutside3 (approved):
Thank you for the date range fix Shaun! I tested the case where the page should show an error gracefully if the date range is invalid.

While running this branch locally I saw a familiar, elusive bug that's been haunting me over the past couple of weeks 👻 For some reason, when revalidating `data`, Grommet's DataChart component will throw an error from its internal `<Detail>`. I've never been able to replicate it reliably, it just seems to happen randomly when selecting different variables for the ERAS `query`, and then going back to "Last 7 Days". Grommet's error message is unhelpful and crashes the page :( 

~~I pushed another small change to this PR checking that props passed to BarChart are defined before it's rendered. Hopefully that bug will stop haunting us.~~ Fixed here https://github.com/zooniverse/front-end-monorepo/pull/7458

PR #7432, opened 2026-05-15
ScatterPlotViewer: allow data points to have custom sizes

shaunanoordin:
## PR Overview

Closes #7348
Follows #7404

This PR will allow the data points displayed on the ScatterPlotViewer (and by extension, the DataImageViewer and VariableStarViewer) to have different sizes. Specifically, project owners will be able to edit the size of the points for each _data series_ in the JSON data.

_Screenshot: lib-classifier showing scatter plot data with 3 different data series. Yellow diamonds have no specified size and therefore default to 25. Blue squares have pointSize="large" so are size 100. Magenta stars have pointSize=200._

image

<details>

<summary>previous in-development screenshot</summary>

_Screenshot: example of different data series with different (and exagerrated) sizes for their data points. The default size is "25" (whatever that means). The "diamonds" are size 25, the "circles" are size 250, and the "triangles" are size 500._

image

</details>

Context:

- The ScatterPlotViewer already allows a certain level of data customisation, e.g. each data series in a chart can have different colours and glyphs representing each data point.
- The ScatterPlotViewer component actually _does_ allow for a `dataPointSize` property to be specified, but according to the [README](https://github.com/zooniverse/front-end-monorepo/blob/238dba7/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer/README.md):
    > dataPointSize (number) Default: 25. The size of the SVG glyph icon. _Not available to configure via the subject location JSON's chartOptions property._
  - I'm not sure _why._

Functionality:

- ScatterPlot: 🔸
  - the size of each data point in a data series can/will be customised by setting `jsonData[n].chartOptions.pointSize = (number|"small"|"medium"|"large)`
  - chartOptions.pointSize can be any of the following values: "small", "medium", "large", or any positive number.
  - by default, the size of a data point will continue to be `25` (whatever that means) on the Scatter Plot Viewer.
- (NO CHANGE) Media: 🔸
  - The Media element, when displaying scatter plot data, will ignore chartOptions.pointSize and just display each data point at the default size of `10`. (Again, whatever that means) 

### Testing

Test with lib-classifier, and with app-project. The goal is to ensure that 1. Subjects can customise the visual size of their data points and 2. (baseline) the visual size of the data points retain their default value when nothing is specified.

On lib-classifier, Storybook:

- Run lib-classifier's Storybook.
[...]

PR #7463, opened 2026-05-27
ScatterPlotViewer: data series controls

### LINKED ISSUE #7347: ScatterPlot Enhancement: toggle each series on/off
## Package
lib-classifier

## Feature Request
As a volunteer, when I view lightcurve data in the ScatterPlotViewer and there are multiple series of data (say, corresponding to photometry from different filters: g, r, i, etc.), I wanted to be able to view/hide (activate/deactivate) each available series.

This functionality already exists for the multi-component VariableStarViewer, and this issue is requesting that the series-specific view/hide capability be made available in the stand-alone ScatterPlotViewer. See screenshot (from [Storybook](https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/index.html?path=/story/subject-viewers-variablestarviewer--default)) showing the desired "Highlight" tool (see magenta box), where all series are activated (ticked) by default, but series can be deactivated (unticked) to better investigate specific features in the data.

All series active:
Image

"C" series deactivated:
Image

## Example Data
- For testing, see Subject [107479754](https://www.zooniverse.org/projects/lcjohnso/lcj-test/talk/subjects/107479754) as part of SubjectSet 126846 on LCJ Test (Project ID = 8749) which has two series (labels: g, i)
- That SubjectSet is currently linked to test workflow on LCJ Test here: Workflow [31774](https://www.zooniverse.org/projects/lcjohnso/lcj-test/classify/workflow/31774)

Screenshot of Subject 107479754:
Image

## Design
_Not needed?_ To minimize dev effort, it might be convenient to make use of existing "Highlight" selector used by VariableStarViewer.

Preferences:
- Only show legend in case of multiple series present. i.e., don't show view/hide tickboxes for case of only one series.
- Keep the "figure legend" containing list of all series and their associated tick boxes out of the plot area, potentially sticking to the same "above the plot" positioning, but where the series are listed horizontally to minimize use of vertical space.
- Labels follow from `seriesOptions.label` property, but it is not required to provide one. We likely want to handle default behavior in the case a missing series label (e.g., use "Series 1" for the first series in the absence of user provided strings).
- Viewers: I also want this enhancement to trickle down to the DataImage viewer (JSON Data + Image) and its associated variants (i.e., JSON Data + Filpbook).
- Optional: provide the ability to disable the legend / tick boxes, in case this additional functionality isn't desired.

shaunanoordin:
## PR Overview

Closes #7347
Follows #7432

This PR updates the Scatter Plot Viewer with new controls which let users toggle the visibility of data series, for multi-series data sets.

Functionality updates:

- ScatterPlotViewer: now has a "Data Series Controls" component.
  - The DataSeriesControls appears in the ScatterPlotViewer ONLY when the data set has _more than 1_ data series.
    - There will be an option to disable this component, e.g. when the ScatterPlotViewer is used in the VariableStarViewer.
  - The DataSeriesControls will list every data series in the data set.
    - Each data series will have a toggle-able checkbox. Checkboxes indicate whether a data series is visible, and lets users toggle that visibility. (NOTE: when a data series is set to "not visible" it really means "semi-transparent".
    - Each data series will indicate its data symbol on the scatter plot chart.
    - Each data series will display its `seriesOptions.label` label value. If absent, the data series will show "Series {n}" as the default label, e.g. "Series 1".
- The following viewers will be affected by this change: ScatterPlotViewer (obviously), and DataImageViewer.
- The following viewers will NOT be affected: VariableStarViewer (because this viewer already has its own equivalent of the data series controls.)

Code changes:

- ScatterPlotViewer:
  - New property: `showDataSeriesControls`, true by default.
  - New component: DataSeriesControls.
    - This uses VariableStarViewer's `<Label>` btw
  - New layout rules:
    - If data contains multi-series data AND showDataSeriesControls=true, then DataSeriesControls component is displayed.
    - If DataSeriesControls is displayed, the Scatter Plot chart takes only takes up 90% of available vertical height, instead of the standard 100%.
  - 🙃 HTML: this component now has even mooore Boxes/divs flying around.
  - HTML: CSS class names added to help identify components when viewing in web browser.
- VariableStarViewer: now explicitly sets showDataSeriesControls=false.
- 🌏 Localisation:
  - added lib-classifier > components.json > ScatterPlotViewer > DataSeriesControls

IMPORTANT NOTE:

- ⚠️ Documentation has been updated to reflect that multi-series data Subjects _must have unique labels for each series._
- ⚠️ Documentation has been added to Scatter Plot Viewer regarding the `highlightedSeries`.
  - For the time being, I'm keeping the highlightedSeries system the same as the VariableStarViewer: this is an _array of strings (data series labels)._ Known issue: if two data series share labels, then toggling series on/off will look wonky.

_Screenshot: Scatter Plot Viewer with a multi-series data set [on app-project](https://local.zooniverse.org:3000/projects/darkeshard/test-project-2026/classify/workflow/3924)_

image

_Screenshot: same as above, but with some series toggled "off"._

image

_Screenshot: same as above, but on iPhone view_

image

<details>
<summary>Additional Screenshots</summary>

_Storybook: scatter plot with single series data. (No changes, Data Series Controls aren't show for single series data.)_

image

_Storybook: Data Image Viewer with single series data. (No changes.)_

image

_Storybook: Data Image Viewer with **multi-series** data. Data Series Controls is present._

image

_Storybook: Variable Star Viewer. (No changes.)_

image

</details>

<details>
<summary>(Old WIP screenshots)</summary>

_Screenshot: WIP showing how the core components._

image
</details>

### Testing

**Testing suite 1: Full Project Experience**

We want to ensure that the following conditions are true:

- The DataSeriesControls should only appear when it's supposed to. 
- The on/off functionality of the DataSeriesControls work.

Test on localhost, on app-project, with a project that uses the ScatterPlotViewer.

- Load a test project that uses _multi-series_ scatter plot data: https://local.zooniverse.org:3000/projects/darkeshard/test-project-2026/classify/workflow/3924
- Confirm the Subject is displayed as a scatter plot chart.
- Confirm that the chart displays multiple data series. (3 in the test project)
  - (OPTIONAL) check with the JSON data to ensure the series options are correct (i.e. the glyph + colour matches what the JSON data describes.)
- Confirm that the DataSeriesControls (the list of data series/labels below the chart) is visible and lists all 3 data series.
- Click on a label in the DataSeriesControls. Confirm that the checkbox toggles off, and the data points on the chart disappears accordingly.
  - Click on the same label, confirm checkbox toggles on, and the data points on the chart appear.
- "Browse"/"Examine" the Subject normally, using pan, zoom, etc.

Test on localhost, on app-project, with a project that uses the DataImageViewer.

- Load a test project that uses _single-series_ scatter plot data: https://local.zooniverse.org:3000/projects/darkeshard/test-project-2026/classify/workflow/3922 _(aka "sorta-kinda-PH-TESS")_
  - This project is configured to use the DataImageViewer, and each Subject contains 1 JSON file and 1 image file.
- Confirm that the DataSeriesControls DOESN'T appear.
  - Note: since this is a DataImageViewer, the ScatterPlotViewer is on the left, and the image viewer is on the right.
- "Browse"/"Examine" the Subject normally, using pan, zoom, etc.

If possible, please test with any other JSON data projects.

**Testing suite 2: Storybook Stories**

We want to ensure the following conditions are true:

- The DataSeriesControls (DSC) _shouldn't_ appear when it's _not_ supposed to.
- Other viewers that use the ScatterPlotViewer should continue functioning as normal.

Test on localhost, on lib-classifier, using Storybook.

- Examine the following Stories under the "Subject Viewers" category:
  - DataImageViewer > Default -- ensure there are no changes
  - DataImageViewer > Multi Image Subject -- ensure DSC works on the scatter plot viewer AND the image viewer works AND the pan and zoom works for both (remember to click "Enable Zoom" buttons)
  - ScatterPlotViewer > Default -- ensure there are no changes, and stories functions the same as before this PR.
  - ScatterPlotViewer > With Varying Data Series Properties -- ensure the DSC works.
  - ScatterPlotViewer > Pan and Zoom -- ensure DSC works AND that pan & zoom works correctly.
  - VariableStarViewer > Default -- ensure there are no changes

### Status

2026.05.27: WIP

[2026.06.05:](https://github.com/zooniverse/front-end-monorepo/pull/7463#issuecomment-4635523232) Ready for review!


shaunanoordin:
### Dev Notes

As of commit: [95e6bbb](https://github.com/zooniverse/front-end-monorepo/pull/7463/commits/95e6bbbafd64bff87fc8645358ed63df0ce76254)

<details>

<summary>(Details on progress)</summary>

(2026.06.05: this section was hidden for PR review brevity)

I'm going to need to step back and revise how I'm planning to approach this problem. Here's a record of everything up until now:

- As of the latest commit, I've got a basic "list all series" and show/hide functionality working.
    image
- However, I realise that my basic implementation of show/hide data points may not be ideal. (For example, if I remove the data series with the largest/smallest x/y values, the zoom-to-fit will automatically snap to another data series, which is a bit jarring.)
    image
- Note that with Variable Star Viewer, a "hidden" series is still semi-visible, like a spooky ghost of datums past.
    image
  - This is done by specifying the `highlightedSeries` value (an array of series labels), which indicates which series should be shown, which should be hidden.
- Logically, this means I should reuse as much code as I can from the VariableStarViewer. OK, makes sense. 🤔 
- **Except I'm now finding issues with the VariableStarViewer.** 😬 

Issues with Variable Star Viewer (and possibly Scatter Plot design in general):

- The `highlightedSeries` value is an array of data series labels, like `['a', 'b']`. Problem is, **data series labels aren't unique.** ⚠️
  - This should be an _array of indexes,_ which would be unique, and actually something I implement in the DataSeriesControls solution I currently have.
- The ScatterPlotViewer is now being overloaded with functions designed to cater to _optional parent viewers._
  - For example, the `highlightedSeries` value and all its supporting functions (e.g. getDataSeriesColor) and functionality (make certain data series ghostly) is tied very, very closely to Variable Star Viewer and not used anywhere else.
  - Another less complex example is the zoom control function that's built into the SPV, but only enabled when the SPV is in the DataImageViewer or VariableStarViewer.
  - Adding the DataSeriesControls component the way that I planned will add another level of specificity to what should be a general component, but I'm not sure there's a more elegant way around this. Eh.

</details>

Bigger questions:

- Q: Should I continue using `highlightedSeries` for showing/hiding data series?
  - A: Yes, it makes sense for the visual presentation than my current implementation.
- Q: Should I continue using `highlightedSeries` as an array of labels?
  - A: Heck no! This needs to be turned into an array of indexes.
- Q: Instead of adding DataSeriesControls into the already overloaded ScatterPlotViewer, should I create a new component called ScatterPlotViewerWithDataSeriesControls?
  - A: Eh, no? This just shuffles around the complexity without really making the hierarchy/responsibility of components clear. e.g. if I create ScatterPlotViewerWithDataSeriesControls (SPVWDSC), then the DataImageViewer would used the SPVWDSC, the "viewer selection logic" needs to swap out the ol' SPV for the SPVWDSC, and the only benefit is that VariableStarViewer (and potential other future viewers that might use the SPV) have a "clean" SPV as a child component.
  - ...but this will mean I will have to juggle some nonsense where I instantiate a copy of `highlightedSeries` _(with state management)_ within the SPV _if and only if_ there's no parent VariableStarViewer passing down its own highlightedSeries value
- Q: Should I improve on VariableStarViewer?
  - A: Not in this PR! (Asides from the changes required to make `highlightedSeries` sensible)

shaunanoordin:
Shaun TODO: (✔️ [done](https://github.com/zooniverse/front-end-monorepo/pull/7463/commits/d0cb0378117890d7a22c40df1dcc602eb9b09709))

> From [comment on 7404](https://github.com/zooniverse/front-end-monorepo/pull/7404#discussion_r3312502622):
> Speaking of modules, you can also change imports from lodash to import flatten from 'lodash/flatten' for better tree-shaking instead of referencing all of lodash in this older code.

Also, bonus test:

> I'll find time to test if switching lib-react-component's components from `@visx/vendor/d3-array` to `d3-array` results in any ESM/CJS issues and/or other problems.

shaunanoordin:
Note to self: actually, before I continue with the plan to rewrite `highlightedSeries` from an array of series labels to an array of series indexes, maybe check with team first if there are some _human_ failsafes. _If_ we want to continue with the current system, we  need to ensure 2 rules are followed:

- Every data series must have a label. 
- Every data series's label must be unique. 

Currently, _neither_ of these rules are enforced in code (hence why I want to make changes), but if we have human rules, the issue may be mitigated enough that I may want to fix the highlightedSeries rule _in a separate PR._

~~Here an example of how this could bork the code: Subject contains 2 data series, neither have labels. The code interprets this as series1.label = '', series2.label = ''. The highlightedSeries code is no longer able to differentiate either.~~ 🤷 

Update (21:54 BST): my apologies to the VariableStarViewer, I need to give the code more credit - there _is_ a fallback function that basically adds a fallback label, if a data series doesn't have one. This still doesn't solve the problem of duplicate labels, but at least there's some sort of safety here.

<details>
<summary>(See VariableStarViewerContainer.jsx's setupSeriesHighlight() )</summary>

```
  setupSeriesHighlight (scatterPlotJSON) {
    const { t } = this.props
    return scatterPlotJSON.data.map((series, index) => {
      if (series?.seriesData.length > 0) {
        const fallbackLabel = t('SubjectViewer.VariableStarViewer.label', { id: index + 1 })
        series.seriesOptions.label ??= fallbackLabel

        return series.seriesOptions.label
      }
    })
  }
```

It works, though I need to double check if modifying the source `scatterPlotJSON` object instead of returning a copy is a wise idea.

This safety check also only occurs at a high level (i.e. at the VariableStarViewerCointainer), and such safeties are not afforded further down the line (i.e. at the ScatterPlotViewer)

</details>

shaunanoordin:
## PR Update

As of commit: [ecdbf21](https://github.com/zooniverse/front-end-monorepo/pull/7463/commits/ecdbf21d7d2b699017caca8af5b239698ad061d1)

Yaaahooo~

image

Progress:

- ScatterPlotViewer:
  - now uses existing highlightedSeries mechanisms.
  - when a data series is un-highlighted, the data points on the scatter plot now go transparent instead of being completely hidden.
- DataSeriesControls:
  - now uses existing highlightedSeries mechanisms.
  - uses _VariableStarViewers's_ Label component. Notably, this causes glyphs to be shown.

### Dev Notes

- For _this PR,_ I'm **sticking** with the `highlightedSeries` as an _array of **strings** (data series labels),_ and putting the onus on project owners to ensure their subjects have unique data series labels.
- Known issue: "toggling series on/off" borks if two or more data series share a label.
- New _purposely-introduced_ issue: if a data series is missing a label, it won't have a fallback value in the highlightedSeries array. The label _displayed on the UI_ however will continue to have a fallback.[^1]

[^1]: I've opted not to use VariableStarViewer's setupSeriesHighlight(), which alters the raw JSON data's labels with fallback values. [(See comment.)](https://github.com/zooniverse/front-end-monorepo/pull/7463#issuecomment-4614566624) Instead, I'm purposely keeping "label as unique ID" and "label to be displayed" as separate things.

### Status

What's left?

<details>

<summary>2026.06.05: ✔️ Done</summary>

- Improve/reconsider UI. That DataSeriesControl can get preeetty tall, and there's a lot of empty space now that there are no other control elements like in VariableStarViewer. (✔️ [done](https://github.com/zooniverse/front-end-monorepo/pull/7463/commits/2cf8cb82755ea7acf94af958f09264b7637f50a7))
    image
- Use translations for labels. (See how VariableStarViewer does it.) (✔️ [done](https://github.com/zooniverse/front-end-monorepo/pull/7463/commits/eadb657eb4c7a9c71c81972d851bc9eed7ba940a))
- Flatten lodash imports. (✔️ [done](https://github.com/zooniverse/front-end-monorepo/pull/7463/commits/d0cb0378117890d7a22c40df1dcc602eb9b09709))

</details>

shaunanoordin:
Goshdangit, just when I thought I was at the finishing line, I encounter another issue. 😬 

_Screenshot: new presentation of DataSeriesControls, now appears underneath the scatter plot. Data series items are laid out horizontally, centre-aligned, and wrap around if there are too many items to fit. Unfortunately, DataSeriesControls appears too snug and close to the MetaTools._

image

<details>

<summary>TL;DR: the scatter plot chart takes up 100% of available subject viewer space. DataSeriesControls exceeds that space, causing it to be "pushed out" of the allocated space and onto the MetaTools.</summary>

(Update 2026.06.05: this was solved by [setting the scatter plot chart _container's_ height to 90%.](https://github.com/zooniverse/front-end-monorepo/pull/7463/commits/5914932a21fe001fe6c4c8577e6ab1fcba794e96))

The ScatterPlotViewer design updates work fine in Storybook, but viewed it on app-project (e.g. https://local.zooniverse.org:3000/projects/darkeshard/test-project-2026/classify/workflow/3924) and things get wonky. In the screenshot above, the Data Series Controls appears uncomfortably close to the Subject Info/Add To Favourites/etc component. Looking more closely:

image

...that's because the DataSeriesControls is being pushed completely outside of the ScatterPlotViewer container (blue outline) it shares with the ScatterPlot component (cyan outline), causing it to overlap with the MetaTools component (red outline).

Looking closer, I realise that the scatter plot (`<Plot>`) sets a specific width & height based on a parent container (@visx/responsive's `ParentSize`), but that means the Plot will always eat up 100% of the screen real estate reserved for the SubjectViewer space. 🤨 

(Side note: this also means that the `ZoomControlButton`, which, is e.g. enabled on the DataImageViewer, has been pushing the Plot outside of its allocated SubjectViewer space all this time. I should look into this.) 

I'm currently looking into ways to work around this gracefully, as any changes to layouts can have knock-on effects to other viewers.

</details>


<details>
<summary>Update 22:06 BST: this is more difficult than it seems</summary>

This is what happens when I wrap the `<ParentSize>` in a `<div>` 😬 

https://github.com/user-attachments/assets/64f90572-73d6-40a6-8e3f-2a985840a8ce

</details>

<details>

<summary> Update 22:50 BST: OK, some progress!</summary>

Alright, as of commit [5914932](https://github.com/zooniverse/front-end-monorepo/pull/7463/commits/5914932a21fe001fe6c4c8577e6ab1fcba794e96) I've wrapped the `<ParentSize>` element in another container that swaps between height: 100% and height: 90% depending on whether there's a presence of the DataSeriesControls.

image

This is a solution, but adds an ungodly amount of `<div>`s to our component code. 🫠 Is this an acceptable tradeoff?

</details>

shaunanoordin:
## Side Quest: Variable Star Viewer Is Too Variable

Wait, this is bizarre. I'm noticing a weird bug with the VariableStarViewer - under certain conditions, when toggling certain data series, the glyphs on the chart swap. In the screenshots below (using the lib-classifier Storybook Story, Subject Viewers > Variable Star > Default), see what happens when I change from showing " 🟥⭕ red circle, 🟦➕ blue cross" to just " 🟥⭕  red circle only". On the chart, the red circles are replaced with the 🟥➕ red cross. 🤨 

image

image

And this is on `main` (68a63f6), so I didn't break this one. (I'm already responsible for so many _other_ bugs, thanks.)

I'll file this for further investigation _later,_ and open an Issue if one doesn't already exist.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer/components/DataSeriesControls/DataSeriesControls.jsx
```diff
@@ -0,0 +1,76 @@
...
+      className='ScatterPlotViewer-DataSeriesControls'
+      direction='row'
+      gap='large'
+      justify='center'
+      wrap={true}
+    >
+      {data?.map((dataSeries, index) => {
+
+        const label = dataSeries.seriesOptions?.label
+        const labelToDisplay = label?.trim() || t('SubjectViewer.ScatterPlotViewer.DataSeriesControls.label', { id: index + 1 })
```
shaunanoordin, line 32: [dev] I initially used `t('SubjectViewer.VariableStarViewer.label')`, but that became "Filter 1", "Filter 2", etc. I've created a new translation entry for "Series 1", "Series 2", etc since that makes more sense for a generic label for a generic scatter plot.

If this proves to be too much of a pain to sync with our Lokalise though, I'm happy enough to switch back to the VariableStarViewer label - this is, after all, just a fallback label.

shaunanoordin:
## PR Update

Layout problems solved, and design updates updated!

_Screenshot: scatter plot viewer with a multi-series data set [on app-project](https://local.zooniverse.org:3000/projects/darkeshard/test-project-2026/classify/workflow/3924)_

image

_Screenshot: same as above, but on iPhone view_

image

<details>
<summary>Additional Screenshots</summary>

_Storybook: scatter plot with single series data. (No changes, Data Series Controls aren't show for single series data.)_

image

_Storybook: Data Image Viewer with single series data. (No changes.)_

image

_Storybook: Data Image Viewer with **multi-series** data. Data Series Controls is present._

image

_Storybook: Variable Star Viewer. (No changes.)_

image

</details>

### Status

This PR is now ready for review. I'll update the [main post](https://github.com/zooniverse/front-end-monorepo/pull/7463#issue-4534419634) with the complete details, testing steps, etc.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer/ScatterPlotViewer.jsx
```diff
@@ -1,32 +1,95 @@
...
+import DataSeriesControls from './components/DataSeriesControls'
 import ZoomControlButton from '../ZoomControlButton'
 
+const DEFAULT_HANDLER = () => {}
+
 const ScatterPlotViewer = forwardRef(function ScatterPlotViewer (props, ref) {
   const {
+    data,
+    highlightedSeries = undefined,  // Array of strings (the labels of the data series) OR undefined, indicating which data series should be visible on the chart. This is usually undefined, meaning the state of highlighted series is managed locally by this component and can be used in conjunction with the built-in DataSeriesControls. If this is an array, it means the state is managed by an external component, e.g. VariableStarViewer.
+    showDataSeriesControls = true,
```
goplayoutside3, line 16: Suggestion to add these props to `PropTypes` at the end of the file. Even though the prop-types library is no longer compatible with React 19, other tools like CoPilot do look at the type definitions and we might be able to replace prop-types with a future tool more easily if the types are already defined.

goplayoutside3: (Same with the other places you've added props in VariableStarViewer's > Label, and DataSeriesControls)

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer/components/DataSeriesControls/DataSeriesControls.jsx
```diff
@@ -0,0 +1,76 @@
+import { Box, CheckBox } from 'grommet'
+import { withTheme } from 'styled-components'
```
goplayoutside3: You could use the `useTheme` hook directly in Label to get `theme.global.colors` instead of this HOC / passing of props. See an example in Scatterplot.jsx:
https://github.com/zooniverse/front-end-monorepo/blob/1f1459786b4c1187fb5abde3b7015ecedc2f75a6/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer/components/ScatterPlot/ScatterPlot.jsx#L87-L92

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer/components/DataSeriesControls/DataSeriesControls.jsx
```diff
@@ -0,0 +1,76 @@
...
+import Label from '../../../VariableStarViewer/components/Controls/components/Label'  // This can be organised this better. Perhaps make this a shared component?
```
goplayoutside3: If anything it might make more sense to move Label into this folder with DataSeriesControls, and then import it into VariableStarViewer. I think of ScatterplotViewer as the more widely used, generic viewer and VariableStarViewer as a one-off, very special use case as a parent viewer. Therefore, VariableStarViewer's version of the series controls is the special case and Label belongs first to the ScatterPlotViewer > DataSeriesControls as the default, generic controls.

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer/ScatterPlotViewer.jsx
goplayoutside3, line 1: I read through your thoughts on how to handle the fact that ScatterPlotViewer is a child of the specialized parent viewers DataImageViewer and VariableStarViewer where the new DataSeriesControls below the plot are displayed in the DataImageViewer and the generic ScatterPlotViewer use cases, but VariableStarViewer has its own Controls already. VariableStarViewer is the specialized, one-off use case viewer.

Even though you decided not to create a container around ScatterPlotViewer with a series controls (SPVWDSC 😆 ), I actually think I would have chosen to do so. I'm going to approve this PR because the UX works as expected and code read is great with lots of in-code comments, but something to think about ~~ A SPVWDSC container around ScatterPlotViewer would remove the need for `highlightedSeries: highlightedSeries ?? localHighlightedSeries` handling (and the useEffect for handling useState, which is usually a red flag for React functional components..) by isolating the DataSeriesControls logic to one container component around ScatterPlotViewer. Then, all you'd have to do is 1) Change the default import in ScatterPlotViewer > [index.js](https://github.com/zooniverse/front-end-monorepo/blob/main/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer/index.js) and 2) Import ScatterPlotViewer.jsx directly into VariableStarViewer instead of the SPVWDSC container. This would separate concerns, leaving VariableStarViewer alone to do its specialized functionality from long ago while simplifying series controls in the more often used ScatterPlotViewer.

Let me know if you'd like to talk through a possible refactor, but also mentioning again that it's ok to go ahead and merge as is, if you need!

goplayoutside3 (approved):
The UX looks good to me! I left some non-blocking comments. Please let me know if you'd like to talk through any of them.

shaunanoordin:
Thank you Delilah! You've shared some very good feedback, and here's what I plan to do before I merge this PR:

- [x] `Label` will move from being a VariableStarViewer component to a ScatterPlotViewer component.
  - The rationale is sound. SPV _is_ the more generic, widely-used component. 👌 
- [x] `Label` will use useTheme().
  - VariableStarViewer's HighlighSeriesCheckboxes will be updated accordingly.
- [x] Label, ScatterPlotViewer, and DataSeriesControls will each get .propTypes/.propTypes updates

As for this comment[^1] -

> Even though you decided not to create a container around ScatterPlotViewer with a series controls (SPVWDSC 😆 ), I actually think I would have chosen to do so. I'm going to approve this PR because the UX works as expected and code read is great with lots of in-code comments, but something to think about

I agree this is definitely something to think about, and I appreciate that the ScatterPlot/SPVWDSC refactor can wait for after this PR. 👍 I'm going to open a new issue[^2] to keep track of this, so the plotted Scatter Plot improvements won't be so scattered!

[^1]: which is in response to this ["Q: Instead of adding DataSeriesControls into the already overloaded ScatterPlotViewer, should I create a new component called ScatterPlotViewerWithDataSeriesControls?"](https://github.com/zooniverse/front-end-monorepo/pull/7463#issuecomment-4579872751)

[^2]: in addition to other issues I need to open: enhancement - ScatterPlotViewer's highlightedSeries should use indexes, not labels. Bug - VariableStarViewer's glyphs shouldn't swap around. Man, I'mma gonna be writing a lot of text after this PR!

shaunanoordin:
Awright, updates done, let's get this merged![^1] 🎉 

[^1]: (Now to start writing up those issues)

PR #7510, opened 2026-07-01
lib-react-components: Refactor FavoritesIconButton checked state styling

mcbouslog:
## Package
- lib-react-components

## Describe your changes
- update IconActionButton with `pressedWhenChecked` prop to conditionally show styling accordingly
- update FavoritesIconButton with `pressedWhenChecked` to `false` so checked (favorited) button state is just pink heart without pressed fill color

## How to Review
  - https://local.zooniverse.org:3000/projects/markb-panoptes/sltp-test-project-staging/talk/subjects/252638?env=staging
  - press an unfavorited subject's favorite button, note pressed state styling
  - release, note favorite button checked state
  - press favorited subject's favorite button, note pressed state styling
  - http://localhost:6007/?path=/story/components-favoritesiconbutton--checked

## Checklist

### General

### General UX

### Refactoring

### DIFF
--- a/packages/lib-react-components/src/IconActionButton/IconActionButton.jsx
+++ b/packages/lib-react-components/src/IconActionButton/IconActionButton.jsx
@@ -2,6 +2,16 @@ import { Box, Button as GrommetButton, Text, Tip } from 'grommet'
 import { bool, func, node, object, oneOfType, shape, string } from 'prop-types'
 import styled from 'styled-components'
 
+const getPressedStateSelector = (pressedWhenChecked) => {
+  if (pressedWhenChecked) {
+    return `&:active:not(:disabled),
+  &[aria-pressed='true'],
+  &[aria-checked='true']`
+  }
+  return `&:active:not(:disabled),
+  &[aria-pressed='true']`
+}
+
 const StyledButton = styled(GrommetButton)`
   align-items: center;
   aspect-ratio: 1;
@@ -62,9 +72,7 @@ const StyledButton = styled(GrommetButton)`
     outline: none;
   }
 
-  &:active:not(:disabled),
-  &[aria-pressed='true'],
-  &[aria-checked='true'] {
+  ${props => getPressedStateSelector(props.$pressedWhenChecked)} {
     background-color: ${props => props.theme.global.colors['neutral-1']};
 
     > svg {
@@ -116,6 +124,7 @@ function IconActionButton({
   onPointerOut = DEFAULT_HANDLER,
   onPointerOver = DEFAULT_HANDLER,
   onPointerUp = DEFAULT_HANDLER,
+  pressedWhenChecked = true,
   role,
   ...props
 }) {
@@ -148,6 +157,7 @@ function IconActionButton({
         icon={icon}
         plain
         role={role}
+        $pressedWhenChecked={pressedWhenChecked}
         {...eventHandlers}
         {...props}
       />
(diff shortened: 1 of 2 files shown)


packages/lib-react-components/src/IconActionButton/IconActionButton.jsx
```diff
@@ -62,9 +72,7 @@ const StyledButton = styled(GrommetButton)`
     outline: none;
   }
 
-  &:active:not(:disabled),
-  &[aria-pressed='true'],
-  &[aria-checked='true'] {
+  ${props => getPressedStateSelector(props.$pressedWhenChecked)} {
```
shaunanoordin, line 75: Question: would the Styled Components' [`css` helper](https://styled-components.com/docs/api#css) need to be used here, or at the getPressedStateSelector() function?

I'm genuinely not sure as that `css` helper continues to confuse me.

shaunanoordin: Update + Adventures in AI: I asked [Copilot](https://github.com/copilot/) and they said _nah, no `css` helper is needed here._

<details>
<summary>Copilot's response</summary>

-- Prompt --

Copilot, please take a look at the changes in @zooniverse/front-end-monorepo/pull/7510 . Can you please tell me if the PR changes - which uses styled-components - should use Styled Components' css helper function?

-- Response --

Looking at the PR changes, no, the code should NOT use Styled Components' css helper function.

Here's why:

The PR introduces a function getPressedStateSelector() (lines 5-12 in the diff) that returns a CSS selector string:

> (code for getPressedStateSeelctor())

This is then embedded directly into the styled component using template literal interpolation:

> ```
> ${props => getPressedStateSelector(props.$pressedWhenChecked)} {
>   background-color: ...
> }
> ```

This approach is correct and doesn't require the css helper because:

1. It's already inside a styled component — The outer template literal context (the styled(GrommetButton)` block) already knows this is CSS
2. Simple string interpolation is sufficient — Since getPressedStateSelector() only returns a plain CSS selector string, there's no need for the css helper
3. The css helper is for complex scenarios — You typically use styled.css helper when you need to:
  - Return multiple CSS rules (not just selectors)
  - Reuse styles across multiple styled components
  - Dynamically generate complete CSS blocks with nested rules

In this case, the code is cleaner and more maintainable as-is. The refactoring extracts conditional selector logic into a helper function while keeping the styling definition unified in one place.

</details>

I can't argue either way since I don't fully understand [Styled Component's source documentation](https://styled-components.com/docs/api#css), so I'll leave it at that but with an eyebrow slightly raised at 20%. 🤨

shaunanoordin (approved):
## PR Review

Package: lib-react-components

This PR changes the style of the Favourite Icon Button. Now, when the "Fave button" is _☑️ checked_ (i.e. the item attached to this button has been selected as a favourite), the background is white (or black in dark mode) instead of _dark teal_ (i.e. the background colour when a button is pressed).

_FavoritesIconButton when item is favourited (light mode and dark mode):_

Image Image

_Previously:_

Image Image

UI & styling changes:

- FavoriteIconButton: when "☑️ checked", background is white (in light mode) or black (in dark mode).
- (no change) for _other_ buttons derived from IconActionButton, their "☑️ checked" state continues to have a dark teal background. (Same as when the button is pressed.)

Code changes:

- IconActionButton: now accepts `pressedWhenChecked` parameter. True by default - this maintains the "dark teal background when checked" behaviour.
- FavoritesIconButton: has pressedWhenCheck=false to apply styling change.

### Status

LGTM. 👍

PR #7511, opened 2026-07-02
lib-react-components: Update Media Video with poster property

mcbouslog:
## Package
- lib-react-components

- currently, often subject videos are downloaded in their entirety when only a single frame is shown
- videos can be large in kb size
- based on new Thumbnailer service that takes a Zooniverse video file and returns the first frame as an image
- we can utilize the video element poster property to show the first frame image and not download the video

## Describe your changes
- update Media Video with `showPoster` property to use src with origin as [video element poster property](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/video#poster) url, with [new Thumbnailer functionality](https://github.com/zooniverse/thumbnailer/pull/28)
- refactor SubjectCard with `showPoster={true}`

## How to Review
- What Zooniverse project should my reviewer use to review UX? n/a until SubjectCard implemented
  - currently, https://zooniverse.github.io/front-end-monorepo/@zooniverse/react-components/index.html?path=/story/components-subjectcard-video--florida-keys: load SubjectCard with videos, inspect network tab, filter by media, note videos loaded
  - on this branch, http://localhost:6007/?path=/story/components-subjectcard-video--florida-keys: same as above, note NO videos loaded only images 
  - http://localhost:6007/?path=/story/components-subjectcard-video--woodpecker-cavity-cam
- Are there plans for follow up PR’s to further fix this bug or develop this feature? maybe there are other instances

## Checklist

### General

### General UX

### Refactoring

### DIFF
--- a/packages/lib-react-components/src/Media/helpers/mediaPropTypes.js
+++ b/packages/lib-react-components/src/Media/helpers/mediaPropTypes.js
@@ -8,6 +8,7 @@ export const propTypes = {
   height: number,
   origin: string,
   placeholder: oneOfType([node, object]),
+  showPoster: bool,
   src: string.isRequired,
   width: number
 }
@@ -19,5 +20,6 @@ export const defaultProps = {
   fit: 'cover',
   flex: 'grow',
   origin: 'https://thumbnails.zooniverse.org',
+  showPoster: false,
   placeholder: null
 }
--- a/packages/lib-react-components/src/Media/Media.stories.jsx
+++ b/packages/lib-react-components/src/Media/Media.stories.jsx
@@ -32,6 +32,8 @@ const TEXT_URL =
   'https://panoptes-uploads.zooniverse.org/subject_location/f5506d1c-a0e9-4aba-a418-6a6c46a7731a.txt'
 const VIDEO_URL =
   'https://static.zooniverse.org/fem-assets/home-video.mp4'
+const VIDEO_ZOONIVERSE_URL =
+  'https://panoptes-uploads.zooniverse.org/subject_location/6d83e7bd-14d1-41cb-8245-8d70a4d3921a.mp4'
 const VOLUMETRIC_URL = 
   'https://panoptes-uploads-staging.zooniverse.org/subject_location/3c56e6a7-be27-4ab3-93a7-c41a3c867baf.json'
 
@@ -97,6 +99,10 @@ export function Video() {
       <Box width='540px' height='540px'>
         <Media alt='Zooniverse in a nutshell' fit='cover' src={VIDEO_URL} />
       </Box>
+      <Text>Zooniverse video src with showPoster</Text>
+      <Media alt='Zooniverse src with showPoster' controls={false} showPoster src={VIDEO_ZOONIVERSE_URL} width={270} />
+      <Text>External video src with showPoster, fallback to default video behavior</Text>
+      <Media alt='External src with showPoster' controls={false} showPoster src={VIDEO_URL} width={270} />
     </Box>
   )
 }
--- a/packages/lib-react-components/src/Media/README.md
+++ b/packages/lib-react-components/src/Media/README.md
@@ -10,6 +10,7 @@ The Media component dynamically renders an appropriate child media component bas
 - height: _(number)_ height in CSS pixels.
 - origin: _(string)_ The origin for the Zooniverse thumbnail service. Defaults to `'https://thumbnails.zooniverse.org'`
 - placeholder: _(node)_ React component or HTML to render as a child of the placeholder container while the image is loading
+- showPoster: _(bool)_ whether to show a poster thumbnail for video media. Defaults to false. When true, preload is set to `'none'`; when false, preload is set to `'metadata'`.
 - src: _(string)_ the source of the media. Required.
   - NOTE: MIME type and the corresponding child media component is determined by this string. If a MIME type can't be guessed, consider specifying a `defaultMimeType`
 - width: _(number)_ width in CSS pixels.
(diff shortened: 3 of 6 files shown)


packages/lib-react-components/src/Media/helpers/mediaPropTypes.js
```diff
@@ -19,5 +20,6 @@ export const defaultProps = {
   fit: 'cover',
   flex: 'grow',
   origin: 'https://thumbnails.zooniverse.org',
```
mcbouslog, line 22: Until https://github.com/zooniverse/thumbnailer/pull/28 is deployed to production, use `origin: 'https://thumbnails-staging.zooniverse.org/',`

shaunanoordin (approved):
## PR Review

Package: lib-react-components
Affects: Media -> Video type, Subject Card -> Video type

This PR modifies the Media and Subject Card components to better support video file types. For the Media component, the component now _optionally_ allows the video (IF it's hosted on Panoptes) to use the Zooniverse thumbnail service for its preview image. For the Subject Card component, it _will_ use the thumbnail service for a video's preview image.

_Screenshot: Subject Cards for video-type media files._

Image

- Visually, the screenshot _looks_ the same whether the preview image uses the Zooniverse thumbnail service (with this PR) or the video file metadata (prior to this PR).
- However, there's a major difference if we monitor the network.
- With the thumbnail service, we now download 3x thumbnail JPGs (6.1kb, 5.0kb, 3.9kb)
- With the video file metadata is download (476kb) and also the... full... video is downloaded... despite the preload=metadata... (2.4MB) wait what[^1] 🤨
  - <details>
      <summary>For real</summary>
        Image
      </details>
- Uh anyway, it just shows that the video thumbnail service is very good at reducing download sizes.

Functional changes

- Subject Card: when displaying video Subjects, the "preview image" now uses an actual thumbnail (IF the video file is hosted on Zooniverse), instead of the full video/video metadata's preview.

Code changes:

- `Media` component: now accepts optional `showPoster` parameter. False by default. When true, the preview image will use the Zooniverse thumbnail service.
  - ~~If thumbnail service is unavailable, the preview image will use the video file itself.[^3]~~ ❓ _I need to double check this._ 
- `SubjectCard` component: sets showPoster=true for video files.

### Testing

Tested with lib-react-components, on Storybook, on localhost, with macOS + Chrome.

Testing steps:

- Started LRC's Storybook.
- View the Subject Card -> Videos story: `http://localhost:6007/?path=/story/components-subjectcard-video--florida-keys`
  - Confirm the preview image displays a, uh, an image corresponding to the subject.
  - Confirm (via Network tab) that _no_ .mp4 files are downloaded, and _only_ the thumbnail .jpg files are downloaded.

Additional testing:
- View other Stories in the Subject Card category, including Application, Audio, Image, and Text.
- Confirm their previews display correctly.
- View other Stories in the Media category, including Image, Video, Audio, Data, and Text Media.
- Confirm their media files display correctly. Note that the Media -> Video Story does NOT use showPoster=true, and therefore will download the .mp4 file.

### Dev Notes

1️⃣ Plus plus points for updating the Media README! 👍 Small suggestion (optional, not blocking):

```
- showPoster: _(bool)_ whether to show a poster thumbnail for video media. Defaults to false. When true, preload is set to `'none'`; when false, preload is set to `'metadata'`.
  - NOTE: poster thumbnails are generated using the Zooniverse thumbnail service, and ensures the preview images downloaded are very small (compared to downloading the video metadata). This will only work if the video files are hosted on the Zooniverse, however - see the [Zooniverse thumbnailer repo](https://github.com/zooniverse/thumbnailer) for acceptable file hosts.
```

2️⃣ Medium-sized suggestion (but also optional, and non-blocking): since the primary changes are in the `Media` component, would it be worth updating the Media -> Video Story with an example that uses showPoster=true? 

- Fair warning: I said this is a medium-sized and _optional_ suggestion because I actually tried changing `Media.stories.jsx` by adding showPoster=true, only to run into a bunch of shenanigans including the fact that the current example video is on `static.zooniverse.org`. 🤦 
- If this suggestion takes more than 10 minutes, _ignore this suggestion._

3️⃣ (UPDATE) Bonus Shaun TODO: check what happens to the Subject Card if the video isn't hosted on Zooniverse, and therefore the thumbnail service doesn't work.
- Does it fall back to loading the video file's metadata (unlikely, since preload will still be 'none'), or does it display nothing at all?
- Based on tinkering with Media's Video Stories (where the video is home video hosted on static.zooniverse.org, and I just added showPoster=true), it seems like the latter is the most likely scenario.
- If confirmed, add this to the notes - this may have some effect on theoretical projects that use externally-hosted video files. (I can't think of any at the moment.)

### Status

LGTM! 👍 

[^1]: What?? OK if you look at say https://zooniverse.github.io/front-end-monorepo/@zooniverse/react-components/?path=/story/components-subjectcard-video--florida-keys and monitor the network on Chrome 150, I see that the network downloads the FULL video (200 response, 2.4MB) and THEN grabs the metadata (206 response, 476kb). But preload=metadata should only load the the metadata! Ah well, at least this PR fixes that.[^2]
[^2]: ...except when the file isn't hosted on the Zooniverse, in which case the preview image falls back to the video file's metadata, which is actually just the full video. Whaaat.[^3]
[^3]: Wait, I just realised I'm assuming this is the case. I need to test this behaviour somehow.

mcbouslog:
Excellent notes @shaunanoordin ! I dug into the potential issues with an external src when `showPoster = true`, which certainly could be an externally hosted subject location, and accordingly added a fallback to load the Video component with defaults so it would show as expected. I also added more stories (for `showPoster` and `showPoster` with the fallback) and more tests, and updated the README as suggested and more based on the changes noted in this comment. Thank you!

Issue #6416, opened 2024-11-01
Broken thumbnail images in the flipbook viewer


eatyourgreens:
## Package
- lib-classifier

## Describe the bug
The thumbnail images at the bottom of the flipbook viewer don't always show an image, for Squirrel Mapper subjects.

## To Reproduce
https://www.zooniverse.org/projects/bcosentino/squirrelmapper/classify/workflow/20808/subject/85284649

<img width=800 alt="A Squirrel Mapper subject with 3 image frames. All three thumbnails, beneath the main image, are empty in the image carousel." src="https://github.com/user-attachments/assets/c0d2fcb6-8545-4ade-a26e-27c21b22a458">

## Expected behavior
Subjects with multiple images should show a row of thumbnails beneath the main image, in the image carousel.

## Additional context
If an image errors, or otherwise fails to load, browsers are supposed to display the alt text instead, but these thumbnails are showing up completely empty (in Firefox.)

eatyourgreens:
This bug also affects collections images, which also use the thumbnail service. Here, the cover image for my Squireel Mapper favourites doesn't display, because the image URL is broken.

<img width=400 alt="A collections card for Squirrel Mapper. The cover image is blank." src="https://github.com/user-attachments/assets/ced9a4c3-f3fe-4293-91fb-2dd05b557993">

Cover image URL for that collection: https://thumbnails.zooniverse.org/220x270/static.inaturalist.org/photos/76275575/original.jpg

eatyourgreens:
This is iNaturalist URLs not working with thumbnails.zooniverse.org, although the subject previews on the project home page at least show alt text when an image doesn't download:
<img width=600 alt="Subject previews on the Squirrel Mapper home page. One image is missing, and replaced by alt text." src="https://github.com/user-attachments/assets/71c1d8c4-f2ee-45b9-92fc-409d9bf95e1c">

I'm not sure what the flipbook buttons should do for images that aren't hosted by Zooniverse. If iNaturalist projects are going to be a thing, then it might be worth making a small component that uses the iNaturalist image API for thumbnails.

PR #7523, opened 2026-07-23
lib-react-components: Update interactive SubjectCard with FlipbookControls

### LINKED ISSUE #7487: Create simplified flipbook controls for interactive SubjectCard
Create simplified flipbook controls for interactive SubjectCard. Somewhat similar to [FlipbookControls](https://github.com/zooniverse/front-end-monorepo/blob/main/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/components/FlipbookControls/FlipbookControls.jsx), however notably:
- if only images, include Play/Pause button (no speed input or Previous/Next buttons)
- if mixed media, no Play/Pause button
- if mixed media, see Figma media type icons for data, text, and video
- horizontally scrolls

mcbouslog:
## Package
- lib-react-components

- closes #7487 

## Describe your changes
- create (simplified) FlipbookControls for interactive subject card, similar to [lib-classifier/FlipbookControls](https://github.com/zooniverse/front-end-monorepo/blob/main/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/FlipbookViewer/components/FlipbookControls/FlipbookControls.jsx)
- create MultiMedia component for interactive subject card with media _and_ flipbook controls
- refactor InteractiveMedia component accordingly ^
- create shared MediaLink component for simple and interactive subject cards

## How to Review
  - mainly review stories
  - http://localhost:6007/?path=/story/components-subjectcard-interactive-image--logged-in-multi-image
  - http://localhost:6007/?path=/story/components-subjectcard-interactive-flipbook-controls--default
  - simple subject card stories should continue to work as expected (no changes)

## Checklist

### General

### General UX

### New Feature
  - Example of SlideTutorial [component](https://github.com/zooniverse/front-end-monorepo/blob/main/packages/lib-classifier/src/components/Classifier/components/SlideTutorial/SlideTutorial.js) with docgen comments, and its [story](https://zooniverse.github.io/front-end-monorepo/@zooniverse/classifier/index.html?path=/docs/other-slidetutorial--default)

### Refactoring

### Post-merge


packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/FlipbookControls/FlipbookControls.jsx
```diff
@@ -0,0 +1,168 @@
...
+
+import IconActionButton from '../../../../../IconActionButton'
+import { useTranslation } from '../../../../../translations/i18n'
+
+const THUMBNAIL_SIZE = 30
+
+const FrameList = styled(Box)`
+  overflow-x: auto;
+  overflow-y: hidden;
+  scrollbar-width: thin;
```
goplayoutside3, line 23: Cool! I didn't know you could style the scrollbar like this.

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/FlipbookControls/FlipbookControls.jsx
```diff
@@ -0,0 +1,168 @@
...
+
+function getThumbnailUrl(imageUrl) {
+  if (typeof imageUrl !== 'string' || imageUrl.length === 0) {
+    return imageUrl
+  }
+
+  if (imageUrl.startsWith('https://') || imageUrl.startsWith('http://')) {
+    const sourcePath = imageUrl.replace(/^https?:\/\//, '')
+    return `https://thumbnails.zooniverse.org/100x100/${sourcePath}`
+  }
```
goplayoutside3: While this should work 100% of the time for panoptes-uploads, how should we handle images with external origins. This Issue https://github.com/zooniverse/front-end-monorepo/issues/6416 describes a really good example from Squirrel Mapper where passing everything to the right of `https://` borks the thumbnailer service.

If a subject uses media (image or otherwise) that isn't compatible with our thumbnailer service, should the thumbnails fallback to something like an icon? I think I lean more toward that fallback rather than fetching the original image, which in the Squirrel Mapper case is 2mb. Let me know what you think!

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/MultiMedia/MultiMedia.jsx
```diff
@@ -0,0 +1,102 @@
...
+	const [playing, setPlaying] = useState(false)
+
+	const currentMediaSrc = useMemo(() => {
+		return mediaSources[currentFrame]
+	}, [currentFrame, mediaSources])
+
+	useEffect(() => {
+		setCurrentFrame(0)
+		setPlaying(false)
+	}, [mediaSources])
```
goplayoutside3: On the website, in what scenario would `mediaSources` change and this `useEffect` runs? Asking mainly for my understanding bc I'm a bit confused on first code read of this file.

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/MultiMedia/MultiMedia.jsx
```diff
@@ -0,0 +1,102 @@
...
+			>
+				<SimpleMedia
+					mediaSrc={currentMediaSrc}
+					placeholder={placeholder}
+					previewHeight={previewHeight}
+					showBackground={showBackground}
+					showTitle={false}
+					subjectIdTitle={subjectIdTitle}
+					width={width}
+				/>
```
goplayoutside3: I'm not totally sold on using SimpleMedia for multi-frame subjects. Because SimpleMedia uses Media to show an image and ThumbnailImage has a built-in "useProgressiveImage" hook, there's a flash of the teal `<Placeholder />` with a `fadeIn` animation on frame change in the InteractiveMedia flipbook UX. The teal placeholder is there when `loading` is true from useProgressiveImage, but even if the image has been cached, the fade animation plays every time. I'll attach a screen recording that shows one frame with the teal placeholder. I wasn't able to replicate after my machine already cached the images in the example subject.

https://github.com/user-attachments/assets/94f531b6-7f76-4669-8a92-a7ed526466da

I think a better UX would be to match the classifier flipbook, no animation, no placeholder (maybe the first frame?). Also, looking at the other media types in Media some handle `loading`, but they define the preview content as null instead of using a placeholder. 

What do you think about either 1) finding a way to turn off the placeholder and animation in Media, or 2) creating a special Image component for MultiMedia? Linked to this is considering how to categorize the subject type examples in Storybook. I noticed you have an "Image" category and within that is both MultiMedia and single-frame image subjects. I think about the categorization a little differently in that the parent category is interactiveMedia, and then the next category level is MultiMedia or SimpleMedia, and then each has examples of media types. The MultiMedia category will have examples of various combinations of images, videos, data, text to satisfy those multi-frame subject types on the website. Is the best choice for the MultiMedia UX the existing Media component with its `loading` placeholder animation disabled? Or better to create a special purpose thumbnail image component for MultiMedia?

goplayoutside3 (commented):
Nice! I like the separation you've done between InteractiveMedia and SimpleMedia. There's one comment to start a discussion about how to handle images hosted externally like Squirrel Mapper, so please see my comment about the thumbnails in FlipbookControls. And another comment to start a discussion about actually flipping through the flipbook. Please see my comment about using the generic Media component.

Once resolved, this'll be good to go!

mcbouslog:
Noting the following to dos per comments (not exclusive, but noting per discussion):
- refactor for externally hosted locations like Squirrel Mapper subjects
    - for flipbook controls use Grommet Image icon 
    - for large frame preview fetch full image (currently Media is trying to request from thumbnailer, failing, requesting full image - can likely cut failing requests)
- refactor with thumbnail image component specifically for multi-image subjects (duplicate thumbnailer service use in Media)

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/MultiMedia/MultiMedia.jsx
mcbouslog: Great point, it wouldn't! Refactored with [a68f35e](https://github.com/zooniverse/front-end-monorepo/pull/7523/commits/a68f35e105ab86b31767920ce164d4aa9778781a).

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/FlipbookControls/FlipbookControls.jsx
mcbouslog: 👍 Refactored to use an icon (Grommet Image icon for now, more icons in subsequent PRs) if the subject location is not hosted by the Zooniverse.

The Media component attempts to determine mime type per src url, which fails if there are params like some Squirrel Mapper subjects (see [Media comments](https://github.com/zooniverse/front-end-monorepo/blob/main/packages/lib-react-components/src/Media/Media.jsx#L13-L27) for additional detail). But since we're just dealing with subjects, and the subject locations structure is `{ mimeType: url }` we can use the locations item key/mimeType, and pass it along as defaultMimeType to Media when applicable.

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/MultiMedia/MultiMedia.jsx
mcbouslog: Created SubjectThumbnail component (open to suggest on naming!) to use with multi image subjects and the UX is much better.

Reorganized stories as suggested with the mix of media types to come in subsequent PR, just multi image (Zooniverse hosted and external hosted) for now.

mcbouslog:
I've added `getSubjectThumbnailSrc` in this PR. It checks if the subject source is hosted by the Zooniverse and uses the thumbnailer service if possible. This fixes the issue with the Media component requesting external hosted images from the thumbnailer service, failing, then requesting the external image. Unfortunately, because the simple SubjectCard and the interactive SubjectCard with a single image location uses Media, the issue noted still occurs in those instances. 

I'm not sure the Media component can be refactored to fix the external source request to thumbnailer fail then normal request issue because the thumbnailer service supports "All other paths will be proxied to Azure zooniversestatic storage account in the $web container (e.g. [www.sciencegossip.org](http://www.sciencegossip.org/), [www.galaxyzoo.org](http://www.galaxyzoo.org/))" so the list of supported hosts might be quite long.

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/FlipbookControls/FlipbookControls.jsx
```diff
@@ -0,0 +1,209 @@
...
+  const selectedButtonRef = useRef(null)
+
+  // Scroll the selected frame button into view when it changes
+  useEffect(() => {
+    selectedButtonRef.current?.scrollIntoView({
+      behavior: 'smooth',
+      block: 'nearest',
+      inline: 'center'
+    })
+    selectedButtonRef.current?.focus()
```
goplayoutside3: I'm not sure about this use of `.focus()` 🤔  If someone is using a keyboard to tab through the subject card, and they intentionally stop on the play/pause button, it receives focus. If they hit Enter or Space to _play_ the flipbook, they can't use Enter or Space to _pause_ because you've moved the tab order focus to a frame button.

Can this `useEffect` just be for scrolling into view, but leave the focus where the user left it? See an example of flipbook in the classifier like [Daily Minor Planet](https://www.zooniverse.org/projects/fulsdavid/the-daily-minor-planet/classify).

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/Image/Image.jsx
```diff
@@ -0,0 +1,79 @@
...
+	alt = '',
+	fit = 'cover',
+	height,
+	origin = 'https://thumbnails.zooniverse.org',
+	src,
+	width,
+	...rest
+}) {
+	const thumbnailSrc = getSubjectThumbnailSrc({ height, origin, src, width }) || src
+	const { error } = useProgressiveImage({ src: thumbnailSrc })
```
goplayoutside3: Non-blocking suggestion to just grab the `preloadImage()` logic from `useProgressiveImage()` and paste it here to access `img.onerror`. An opinion, but I don't think this Image.jsx file is the same functionally as a "progressive image" (on purpose bc we're not animating or handling `loading`), so the code read is a bit confusing seeing `useProgressiveImage()` imported. Also I'm pretty sure lib-react-component's file organization and build process means that anytime a function is imported from the `/hooks` folder, all of the code in that entire folder is included.

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/Image/Image.jsx
```diff
@@ -0,0 +1,79 @@
...
+				<span style={fallbackStyle}>
+					<img
+						alt={alt}
+						height='100%'
+						src={imageSrc}
+						style={{ objectFit: fit }}
+						width='100%'
+					/>
+				</span>
+			</noscript>
```
goplayoutside3, line 75: Do you plan to create a Story component for the fallback style of subjects whose media didn't load? Thinking particularly about subjects with images from inaturalist (that not infrequently) get removed from their server and we have no control over it. Definitely not a requirement to merging this PR, but suggestion to note a to-do when manually testing real Zooniverse subjects before launch.

goplayoutside3 (approved):
Approved! Changes we talked about last week look good. I left a couple of non-blocking comments that don't necessarily need to be addressed in this PR, but would be good to make a note of before page launch.

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/FlipbookControls/FlipbookControls.jsx
mcbouslog: Yes, not stealing focus is a much better UX! Change made.

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/Image/Image.jsx
mcbouslog: Yes, good suggestion - change made in https://github.com/zooniverse/front-end-monorepo/pull/7523/commits/388d975b919de57c3e9bfc7c2125e586024c2b52

packages/lib-react-components/src/SubjectCard/components/InteractiveMedia/components/Image/Image.jsx
mcbouslog: Yes an external image failure story is helpful for all of the related cards (simple, interactive single and multi) - added stories accordingly!

PR #7524, opened 2026-07-27
lib-classifier: Constrain GeoMapViewer to the subject extent with a bounds mask

### LINKED ISSUE #7373: lib-classifier: Add bbox field to GeoJSON model
## Package
- lib-classifier

Towards letting subjects declare their working extent. Beaver Dams uses a regular grid that divides Wisconsin into bounded subjects; the volunteer needs to know what area they're meant to annotate.

Considerations
- support the optional `bbox` field on a GeoJSON subject (per RFC 7946 section 5)
- bbox shape: [west, south, east, north] in EPSG:4326
- backward compatible: subjects without bbox continue to work as today

### LINKED ISSUE #7374: lib-classifier: Add SubjectExtent visual layer as OL rectangle
## Package
- lib-classifier

Towards making the subject extent visible to volunteers as they pan and zoom. Without a visual reference, it's easy to wander off the area you're meant to be annotating.

Considerations
- render the subject bbox as a non-interactive rectangle on the map
- styled as a dashed outline with no fill
- sits below the editable feature layer; not selectable, not translateable
- subjects without a bbox render no rectangle

### LINKED ISSUE #7375: lib-classifier: Wire subject bbox to initial map fit and Recenter button
## Package
- lib-classifier

Towards centering the volunteer on the subject extent rather than just the seeded features.

Considerations
- when the subject has a bbox: initial fit and the Recenter button use the bbox
- when the subject has no bbox: fall back to the seeded features extent (current behavior)
- works for empty subjects (no seeded features) too, given a bbox is present

### LINKED ISSUE #7376: lib-classifier: Extract subject bbox from container to GeoMapViewer
## Package
- lib-classifier

Plumbing: thread the subject bbox from the container down to the viewer and the SubjectExtent layer so the layer renders correctly.

Considerations
- container reads bbox from the subject GeoJSON
- passed through to the viewer as a prop
- visible result is the dashed rectangle showing on the map

kieftrav:
## Package
- lib-classifier

- Closes #7373
- Closes #7374
- Closes #7375
- Closes #7376

## Describe your changes
- add `bbox` to the `GeoJSON` store model so a subject's RFC 7946 bbox survives into the viewer
- constrain the view to the subject bbox in `loadGeoJSON`, falling back to the features extent when a subject has no bbox
- mask the map beyond the subject extent with a grey overlay and a light outline
- cap zoom-out and clamp panning so the subject box never leaves the viewport
- clamp point drags, teleports, translates, and line vertex edits to the extent; draw clicks outside it place nothing
- recenter refits to the same extent as the initial fit

## How to Review
Helpful explanations that will make your reviewer happy:

What Zooniverse project should my reviewer use to review UX?
- markbouslog/north-star-mapping-project, workflow 32087 "GeoDrawing SegmentedLine" (subject set 136882): `https://local.zooniverse.org:8080/?project=31513&workflow=32087&env=production`
- staging equivalent (kieftrav/geo-tools, subjects with an explicit bbox): `https://local.zooniverse.org:8080/?project=2050&workflow=3919&env=staging`

What user actions should my reviewer step through to review this PR?
- [x] run `pnpm dev` in `packages/lib-classifier`, open `https://local.zooniverse.org:8080/?project=31513&workflow=32087&env=production`
- [ ] the map loads fit to the subject box, with everything outside greyed behind a white outline
- [ ] pan hard in any direction; the subject box pins at the viewport edge and never leaves view
- [ ] zoom out; the view stops before the subject shrinks away
- [x] with a line tool active, click outside the box; no vertex is placed and no sketch dot renders there
- [x] draw a line inside, then drag a vertex outside and release; it snaps back to the boundary
- [x] drag the subject's point outside; it clamps at the edge. Click-teleport outside; the point stays put
- [x] pan away, then click recenter; the view refits to the subject box
- [x] confirm earlier-PR behaviors are intact: vertex grab/drag, line count and vertex bounds, delete affordance
- [x] run `pnpm test` in `packages/lib-classifier` and confirm `extentConstraint.spec.js` and `store/JSONData/index.spec.js` pass

Which storybook stories should be reviewed?
- n/a for this PR. The behavior is driven by the subject's GeoJSON rather than a tool config, so there is no story control to exercise; the staging workflow above is the review surface.

Are there plans for follow up PR's to further fix this bug or develop this feature?
- Yes. This PR completes the Subject Definition milestone (#7373, #7374, #7375; #7376's plumbing is satisfied by reading the bbox at load time in `loadGeoJSON`). Sequential PRs that follow this one (in merge order):
- lib-classifier: geoDrawing Point tool subject-driven rendering with min/max creation bounds (branch ready, PR to follow)
- Panoptes-Front-End: Add "Number of points" Min/Max editor to the GeoDrawing Point tool config (branch ready, PR to follow)

# Checklist

## General

## New Feature

### DIFF
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/hooks/useMapSelection.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/hooks/useMapSelection.js
@@ -5,6 +5,7 @@ import { unByKey } from 'ol/Observable'
 
 import asMSTFeature from '../helpers/asMSTFeature'
 import { FEATURE_HIT_TOLERANCE_PX } from '../helpers/createGeoLineStringInteraction'
+import { clampToSubjectExtent } from '../helpers/extentConstraint'
 import { isPixelNearPointCenter, POINT_CENTER_HIT_RADIUS_PIXELS, getFeaturePixelsAcrossWorldCopies } from '../helpers/hitTesting'
 import { clearSelectedFeature, selectFirstFeature } from '../helpers/mapSelection'
 
@@ -42,11 +43,21 @@ export default function useMapSelection({
       }
     })
 
+    const translateEndKey = translate.on('translateend', (event) => {
+      event.features.forEach((feature) => {
+        const geometry = feature.getGeometry?.()
+        if (geometry?.getType?.() !== 'Point') return
+        geometry.setCoordinates(clampToSubjectExtent(map, geometry.getCoordinates()))
+        feature.changed()
+      })
+    })
+
     map.addInteraction(select)
     map.addInteraction(translate)
     setInteractions({ select, translate })
 
     return () => {
+      unByKey(translateEndKey)
       map.removeInteraction(translate)
       map.removeInteraction(select)
       setInteractions({ select: null, translate: null })
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/hooks/useMapSelection.js
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/hooks/useMapSelection.js
@@ -49,6 +49,9 @@ export default function useMapSelection({
         if (geometry?.getType?.() !== 'Point') return
         geometry.setCoordinates(clampToSubjectExtent(map, geometry.getCoordinates()))
         feature.changed()
+        if (geoDrawingTask?.setActiveFeatureGeometry) {
+          geoDrawingTask.setActiveFeatureGeometry(geometry)
+        }
       })
     })
 
--- a/packages/lib-classifier/src/store/JSONData/GeoJSON.js
+++ b/packages/lib-classifier/src/store/JSONData/GeoJSON.js
@@ -14,6 +14,7 @@ const Feature = types.model('GeoJSONFeature', {
 
 export default types.model('GeoJSON', {
 	type: types.literal('FeatureCollection'),
+	bbox: types.maybe(types.frozen()),
 	features: types.array(Feature),
 	reference_data: types.maybe(types.frozen())
 })
--- a/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/GeoMapViewer.jsx
+++ b/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/GeoMapViewer.jsx
@@ -19,7 +19,7 @@ import ZoomInButton from './components/ZoomInButton'
 import ZoomOutButton from './components/ZoomOutButton'
 import { GEOJSON_READ_OPTIONS, ZOOM_ANIMATION_DURATION_MS } from './helpers/constants'
 import loadGeoJSON from './helpers/loadGeoJSON'
-import { fitViewToFeatures } from './helpers/mapSelection'
+import { fitViewToExtent } from './helpers/mapSelection'
 
 import useMapCursor from './hooks/useMapCursor'
 import useMapInteractions from './hooks/useMapInteractions'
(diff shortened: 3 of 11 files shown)


packages/lib-classifier/src/store/JSONData/GeoJSON.js
```diff
@@ -14,6 +14,7 @@ const Feature = types.model('GeoJSONFeature', {
 
 export default types.model('GeoJSON', {
 	type: types.literal('FeatureCollection'),
+	bbox: types.maybe(types.frozen()),
 	features: types.array(Feature),
```
mcbouslog, line 18: I think Beaver Dam subjects might not include any features initially, in which case - would this need to be?
```suggestion
	features: types.maybe(types.array(Feature)),
```

mcbouslog (approved):
Code changes look good! 👍 

I'm going to approve as I think a subsequent PR addresses the following and it does not impact the live NfN mapping project, but the initial classifier load is not what I expect as I'm having trouble seeing the tiles, but could be something on my end:

Image

packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/GeoMapViewer/hooks/useMapSelection.js
```diff
@@ -42,11 +43,21 @@ export default function useMapSelection({
       }
     })
 
+    const translateEndKey = translate.on('translateend', (event) => {
+      event.features.forEach((feature) => {
+        const geometry = feature.getGeometry?.()
+        if (geometry?.getType?.() !== 'Point') return
+        geometry.setCoordinates(clampToSubjectExtent(map, geometry.getCoordinates()))
+        feature.changed()
```
mcbouslog, line 51: In the `createMoveToClickInteraction` the drag handler eventually calls the geoDrawingTask.setActiveFeatureGeometry, is that also needed here?

```suggestion
        feature.changed()
        if (geoDrawingTask?.setActiveFeatureGeometry) {
          geoDrawingTask.setActiveFeatureGeometry(geometry)
        }
```

packages/lib-classifier/src/store/JSONData/GeoJSON.js
```diff
@@ -14,6 +14,7 @@ const Feature = types.model('GeoJSONFeature', {
 
 export default types.model('GeoJSON', {
 	type: types.literal('FeatureCollection'),
+	bbox: types.maybe(types.frozen()),
```
mcbouslog, line 17: Not necessary, but I think could be:
```suggestion
  bbox: types.maybe(types.array(types.number)),
```

packages/lib-classifier/src/store/JSONData/GeoJSON.js
kieftrav: Great idea! So good, MobX automatically does it for us :) 

`Note that since MST v3 types.array and types.map are wrapped in types.optional by default, with [] and {} set as their default values, respectively.`

https://mobx-state-tree.js.org/overview/types

packages/lib-classifier/src/store/JSONData/GeoJSON.js
kieftrav: Not a bad idea - the issue is if there is a malformed bbox the whole subject fails to load instead of degrading gracefully. I do the stronger type-driven verification in the [loadGeoJSON helper](https://github.com/zooniverse/front-end-monorepo/pull/7524/changes#diff-208bae3d816288fe81b2eed9cf7ad1d16ae85728a4e188f82e66a87a69b34076R8)

kieftrav:
> Code changes look good! 👍
> 
> I'm going to approve as I think a subsequent PR addresses the following and it does not impact the live NfN mapping project, but the initial classifier load is not what I expect as I'm having trouble seeing the tiles, but could be something on my end:
> 
> Image

I've noticed this too occasionally. I'm not sure what causes it, but the fact that it's intermittent is the thing I'm struggling with. It worked great during our last demo - I'm not sure if it's a specific tiling/mapping service or if it's something else. It is something I want to keep an eye on, especially if volunteers are experiencing it.

Issue #7117, opened 2025-11-20
Reset password page - FEM design


seanmiller26:
The reset password page allows users to request to change their password. I have outlined the steps of that process here.

Initial page (accessed by clicking FORGOT YOUR PASSWORD? on a Sign In modal

<img width="972" height="431" alt="Image" src="https://github.com/user-attachments/assets/01f4c006-b0a2-4712-8851-eb733fa37a93" />

After entering an email

<img width="786" height="293" alt="Image" src="https://github.com/user-attachments/assets/39956ecb-2784-4cf3-9479-5f6cd16fe6b5" />

A user will receive an email with a link. This is the page that opens after clicking the link in your email.

<img width="612" height="222" alt="Image" src="https://github.com/user-attachments/assets/825cd492-2ff9-44ad-8e39-f50699cc24ea" />

The current UX then redirects you to /projects with a Sign In modal open. We will keep that same behavior and use the FEM Sign In modal once the new /projects page is implemented.

---

**Mobile**

<img width="923" height="358" alt="Image" src="https://github.com/user-attachments/assets/435d0ec6-df41-4d32-af85-eec892bae89f" />

**Dark mode**

<img width="609" height="212" alt="Image" src="https://github.com/user-attachments/assets/0450492c-4a21-445c-a66e-5c0b225ff47a" />

**Error States**

<img width="1364" height="486" alt="Image" src="https://github.com/user-attachments/assets/945be972-4c64-40ea-9e62-b52ef7432ec9" />

<img width="1347" height="564" alt="Image" src="https://github.com/user-attachments/assets/b2251e4d-40f3-4415-95cb-0d672a70398f" />

Figma: https://www.figma.com/design/BieGKNqUqgEqdhvlcip02V/Reset-Password?node-id=4-269&t=FXFie7hHiRKHmnD6-1

shaunanoordin:
I forgot to share my dev notes from our meeting on 2026.07.02, so here they are: (also PS I've started work on the Reset Password page)

## Dev Notes

URLs:
1. Request state: https://www.zooniverse.org/reset-password
2. Confirmation state: https://www.zooniverse.org/reset-password?reset_password_token=1234567890

Design: https://www.figma.com/design/BieGKNqUqgEqdhvlcip02V/Reset-Password?node-id=4-269&t=3nadigfGiAewrq1X-0
Issue: https://github.com/zooniverse/front-end-monorepo/issues/7117

- Security: never confirm if an email address exists!
[...]

PR #7514, opened 2026-07-10
Unsubscribe Page pt1: add (non-functional) Unsubscribe Page

shaunanoordin:
## PR Overview

Part of: Unsubscribe Page (#7123) 
Package: lib-content, app-root

The Zooniverse has [an Unsubscribe Page](https://www.zooniverse.org/unsubscribe) but it's hosted on PFE. This PR is the first step to making an FEM Unsubscribe Page.

image

Functional changes:

- Unsubscribe (page): new page added.
  - This page is accessible (locally) at `https://local.zooniverse.org:3000/unsubscribe` (on app-root)
    - When visited normally, a "TODO: add form" message is shown.[^1]
    - When visited with the `?processed=true query param, an "Unsubscribe Successful message" is shown on this page.[^2]
  - 🚧 For this PR, _an incomplete Unsubscribe Form will be added._ i.e. no functional Unsubscribe action will be added.

[^1]: obviously, not the final state.
[^2]: In the _full version,_ this message should only show when either 1. the Unsubscribe action is completed successfully via an Unsubscribe Form, or 2. when the [`?processed=true` query parameter](https://www.zooniverse.org/unsubscribe?processed=true) is added.

Code changes:

- app-root:
  - New route: `/unsubscribe`
- lib-content:
  - `Unsubscribe` (screen): added
  - `UnsubscribeForm` added.
    - Shows a "TODO" message normally
    - When `processed=true` param is passed, it shows an "Unsubscribe successful" message.
  - `OtherLayout`: new page layout added. It's basically the About page, but the header has no links to other About pages.
  - PageLayout folder: additional documentation added.

### Testing

**Test 1: view on lib-content**

- Run lib-content's Storybook
- View Other -> Unsubscribe -> Default
  - You should only see a placeholder message here.
- View Other -> Unsubscribe -> After Redirect From Panoptes
  - You should see the "Unsubscribe successful" message here.
  - Ensure the design looks OK.

**Test 2: view on app-root**

- Run app-root on localhost. (`pnpm dev`)
- Open `https://local.zooniverse:3000/unsubscribe`
  - You should see the Unsubscribe Page with a TODO placeholder message.
- Open `https://local.zooniverse:3000/unsubscribe?processed=true`
  - You should see the Unsubscribe Page with the "Unsubscribe Successful" message.
  - Ensure this state conforms to the Figma design (see #7123)

**Test 2a: mobile & desktop views**

- While still viewing the Unsubscribe Page on app-root, change the browsing window to various sizes.
[...]

PR #7529, opened 2026-07-30
Reset Password Page pt1: add route, page, and forms

shaunanoordin:
## PR Overview

Part of: Reset Password Page (#7117)
Package: lib-content, app-root

The Zooniverse has [a Reset Password Page](https://www.zooniverse.org/unsubscribe) but it's hosted on PFE. This PR is the first step to making an FEM Reset Password Page.

image

image

image

Functional changes:

- Reset Password (page): new page added
  - This page is accessible (locally) at https://local.zooniverse.org:3000/reset-password (on app-root)
- Reset Password form: has ~~two~~ three modes.
  - **Mode 1: No Access** - users can't access the page _if they're logged in._
    - Message: "please logout to request a password reset."
  - **Mode 2: Request Reset**  - if user accesses the page with _no reset password token,_ (i.e. just https://www.zooniverse.org/reset-password), they'll see the Request Reset form.
    - The Request Reset form asks from an email address.
    - (See README in this PR for full behaviour, including ready/busy/success/error states upon submitting.)
  - **Mode 3: Commit Reset**  - if user accesses the page _WITH a reset password token,_ (e.g. https://www.zooniverse.org/reset-password?reset_password_token=1234567890) they'll see the Commit Reset form.
    - The Commit Reset form asks from a a new password + password confirmation.
    - (See README in this PR for full behaviour, including ready/busy/success/error states upon submitting.)
    - 🚧 NOTE: as per #7117, on success, users should be redirected to a theoretical sign-in page, or the home page with the sign in form open, and prompted to sign in. However, this behaviour will be added in a Part 2.

Code changes:

- app-root:
  - New route: `/reset-password`
- lib-content:
  - `ResetPassword` (screen): added
    - `RequestResetForm`: added 
    - `CommitResetForm`: added
    - `NoAccess`: added

### Testing

**NOTE: the full app-root testing can only be done on production.** I was never able to get the email from Panoptes on staging.

Setup: create a new account, or ask me for my `zootester202608A` account. I shall call this the "test account".

**Scenario 1: lib-content, Storybook**

- Boot up lib-content's Storybook.

- Examine the Story: Reset Password > Default
  - This should mimic the page when user visits with NO unsub token.
  - Confirm design looks OK.
  - Type in an _invalid email_ and submit.
    - Standard (browser-supported) error messages should trigger and prevent the submit.
  - Type in a _valid email_ and submit.
    - You should briefly see the Loader.
    - You should then see the "Request has been terminated" error message because _you can't submit requests via Storybook._ (The Panoptes API has CORS protection.)
 
- Examine the Story: Reset Password > With Reset Password Token
  - This should mimic the page when user visits WITH an unsub token.
  - Confirm design looks OK.
  - Type in some wonky input and try to submit, e.g.:
    - Password or confirmation less than 8 chars long. This should trigger a standard (browser-supported) error and not submit.
    - Password and confirmation don't match. This should trigger a custom error message and not submit.
  - Type in a _valid password & confirmation_ and submit.
    - You should briefly see the Loader.
    - You should then see the "Request has been terminated" error message because _you can't submit requests via Storybook._ (The Panoptes API has CORS protection.)
 
 - Examine the Story: Reset Password > User is Logged In
  - Confirm design looks OK. 

**Scenario 2: app-root with ?env=production**

- Boot up app-root and go to https://local.zooniverse.org:3000/reset-password?env=production
  - Visit the page while logged in. Ensure the page stops you from doing anything.
  - Visit the page while logged out, then continue the test scenario.
- You should see the **Request Reset form**.
- Type in the email for your test account and submit.
  - You should see a success message.

- ⏳ Wait a few minutes in realtime. Go grab some tea or something.
- 📩 Check your test account's email. Copy the `reset_password_token` in the "reset password" link.

- Go to `https://local.zooniverse.org:3000/reset-password?env=production&reset_password_token=(INSERT TOKEN HERE)`
- You should now see the **Commit Reset form**.
- Submit a valid new password + confirmation.
  - You should see a success message with a reminder to the user to sign in, _and nothing else._ (We're not implementing the "redirect user to sign in page" in this PR.)

Extra notes:
- You can also test for invalid input again.
- ⚠️ Be aware that Panoptes has a cooldown period between requesting password resets.
  - (AFAIK this is the only time a user can legitimately trigger a _server-side API error message._ Otherwise, it's usually client-side input error messages.)

### PR Checklist

_(This checklist is for the PR author)_

- [x] The component is accessible _(I hope.)_

### Status

2026.07.30: WIP

202608.06: Ready for review!

This PR needs a Part 2 follow up, to address the intended "redirect user to sign in page, after successful password reset" behaviour.


shaunanoordin:
## PR Update

This PR is now ready for review!

shaunanoordin:
Oh butts, the Github build & test actionss all went KO due to the service outages yesterday. I'll perform a rebase to re-trigger the jobs.

shaunanoordin:
## Adventures in AI

_As part of our dev research, we're keeping track of when we use AI systems and how they perform, so we can evaluate if/how this tool can fit into our work._

I asked [Copilot](https://github.com/copilot), via the Github webpage, to review this PR. Here are the main points

(✅ : good feedback, will be addressed. ✔️: good feedback, but no actions will be taken. ✖️: you're wrong, AI.)

> **1. RequestResetForm.jsx – Missing <label> for email input**
> The email input uses aria-label but lacks a corresponding <label> element. While aria-label will work for screen readers, a native <label> is better practice for form semantics and provides a larger click target.

✔️ I personally agree that forms should have explicit labels, but this is a design decision. I'm happy to revise the code if Sean wants to.

> **2. CommitResetForm.jsx – Form controls remain enabled after success**
> On line 145–146, the submit button and inputs are disabled when isComplete is true, which is good. However, the comment on line 66 says "Form should remain disabled?" suggesting this was uncertain. The current behavior (keeping it disabled) is correct for a success state, but consider adding explicit focus management—moving focus to the success message so screen reader users see it.

✔️ I should change that doc to say "Form should remain disabled!"

✖️ I'm dubious about changing focus management to the success message. For one, the success message is in an aria-role=status region explicitly to announce changes. For another, _programmatically changing focus ain't always a good idea._ [BBC accessibility guide: "Focus must only change when activated by the user. ... This guideline is particularly important when creating forms."](https://www.bbc.co.uk/accessibility/forproducts/guides/mobile/changing-focus/)

> **3. Missing ARIA labels on styled components**
> Both password inputs in CommitResetForm (lines 127–131, 142–147) use the password input type, which is good, but there's no explicit aria-describedby linking the help text (e.g., "Both fields must match") to the inputs. This leaves screen reader users without immediate feedback about validation rules.

✅ I'm missing aria-describedby!! Good catch.

> **4. HTML5 pattern validation may be too strict for some locales**
> CommitResetForm uses an HTML5 pattern=".{8,}" (lines 120, 141). This works for most locales, but some combining diacritics or multi-byte characters may behave unexpectedly. However, this follows the Panoptes backend rule, so it's defensible. Just be aware if you get reports from users with complex scripts.

❓ ...huh? That's fascinating - I never thought about locales. I can neither confirm nor deny that `.{8,}` (minimum 8 characters) might cause issues; this pattern is taken from our existing PFE Reset Password code, so it may need to be revised?

As far as I can tell, "おはよう" and "朋友你好" both still count as 4 character strings in JavaScript - but there may be computer-based setting that change that logic? I dunno.

> **5. No client-side password confirmation validation before form submission**
> In CommitResetForm.onSubmit(), the passwords are checked after the form fires the onSubmit event (line 87). The <Form> component should prevent invalid submissions due to the pattern and required attributes, but the "passwords must match" check is entirely custom and only runs in JavaScript. Consider adding a <FormField> wrapper or onBlur validation to provide immediate feedback.

✔️ Hmmm... let me think if this actually improves the UX.

> **6. API error response structure assumption may fail silently**
> In doCommitPasswordReset.js (line 38–40), the code assumes the error has a .body.errors[0].message structure.
> But if _error.body.errors is an empty array or missing, users see "Unknown API error", which is not helpful. Panoptes may return different error shapes. Consider logging the full error object or building a fallback message that includes the status code.

✖️ The `if (_error.req && error.status >= 400 && error.status <= 599)` ensures that error check applies only if the caught Error object isn't an Error object as is actually a XMLHttpRequest Response. This means an API error response from Panoptes, which means errors[0].message, and if doesn't, then it's genuinely an unknown API error.

> **7. doRequestPasswordReset.js inconsistent error handling pattern**
> doRequestPasswordReset() (lines 26–32) returns the raw error object, but it doesn't inspect the shape like doCommitPasswordReset() does. If Panoptes returns a response-like object instead of an Error, the component won't have a .message property. This could cause the error message display to fail silently. Keep error handling consistent across both helpers.

✔️ Valid comment, but the problem is that doCommitPassword() is the _exception,_ since it uses auth.resetPassword(), which throws Responses instead of Errors. doRequestPassword() is actually the correct standard.

> **8. HTTP 429 (rate limit) special-cased only in one form**
> The "too many requests" error (line 40 in RequestResetForm.jsx) is handled with a special translation key, but CommitResetForm doesn't check for error.status === 429. If users hit the Panoptes rate limit after requesting a reset, they'll see a generic error instead of the friendly message. Consider moving this check to the helper function.

✔️ Fair assessment, but 429s would only appear in RequestResertForm, since the one-off use limit of the unsub token means the CommitResetForm can't perform multiple password resets. (You'd get an "invalid token" error instead if you tried.)

> **9. Typo in comment in ResetPassword.jsx**
> Line 6 says "We need a wrapper around the <ResetPassword>" but it's actually in ResetPasswordContainer.js. The comment is a bit unclear about why the wrapper is needed. Suggest: "We need a container to check login status via PanoptesAuthContext, which is only available on the client side."

✖️ No, I meant what I said. If a human thinks I don't make sense, I will revise my docs.

Also did you make a typo with "Suggest:"? Did you mean "Suggestion:"?

> **10. Potential issue: useId() in CommitResetForm**
> Lines 64–65 call useId() which generates a fresh ID on each render. This is fine, but ensure that Grommet's FormField doesn't rely on stable IDs for validation message association. If it does, moving useId() calls outside the component or memoizing might be needed (though unlikely to be an issue in practice).

✔️ That's a good point about useId() changing IDs on every render. It's not applicable to this form, though.

✖️ Wait, why are you talking about FormField? I replaced FormField with input + label, how did you miss that?

✅ Goddammit, I forgot to delete the import. I'll fix this.

<details>
<summary>[Full AI interaction]</summary>

Query: _Hi Copilot, may I please ask you to review @zooniverse/front-end-monorepo/pull/7529 ? For brevity's sake, you don't need to repeat anything that's already written in the PR posts, since I'm the author of that PR. I'm keen to know if there are any obvious bugs, typos, accessibility issues, bad HTML form practices, and so on that can be fixed._

Response:

```
PR Review: Reset Password Page
PR Summary
Title: Reset Password Page: add route, page, and forms
Status: Open, not in draft
Scope: 17 files changed, +700 −1 lines
Packages: lib-content, app-root
Linked to: Issue #7117 (Reset Password Page epic)

Key Purpose: This is Part 1 of a multi-part feature to migrate the Zooniverse Reset Password Page from PFE to FEM. The PR adds a new route, page component, and three forms for handling user password reset workflows.

Merge Status: Mergeable, but currently behind main (rebaseable). No blocking reviews yet; the frontend team has been requested for review.

Core Changes
The PR introduces a new reset password flow with three distinct modes:

No Access Mode — Users already logged in are blocked with a clear message
Request Reset Mode — Unauthenticated users submit their email to request a password reset
Commit Reset Mode — Users with a reset token submit a new password
New files added:

Route handler: app-root/src/app/reset-password/page.js
Container: app-root/src/app/reset-password/ResetPasswordContainer.js (checks login status)
Main screen: lib-content/src/screens/ResetPassword/ResetPassword.jsx (renders the three modes)
Form components: RequestResetForm.jsx, CommitResetForm.jsx, NoAccessMessage.jsx
Helper functions: doRequestPasswordReset(), doCommitPasswordReset(), isNewPasswordValid()
Tests & stories for the new components
Translation keys in en.json
README documenting the feature
Other Changes
Minor fixes and refactoring:

Fixed typo in Unsubscribe/README.md ("Unsubscrube" → "Unsubscribe")
Added a TODO comment in doUnsubscribe.js noting the return pattern should be kept consistent with the new password reset helpers
Detailed Review
Accessibility & Form Practices
RequestResetForm.jsx – Missing <label> for email input
The email input uses aria-label but lacks a corresponding <label> element. While aria-label will work for screen readers, a native <label> is better practice for form semantics and provides a larger click target.

jsx
// Current (line 72)
<TextInput
  aria-label={t('ResetPassword.RequestResetForm.inputEmail')}
  ...
/>

// Suggested
<label htmlFor={emailInputId}>
  {t('ResetPassword.RequestResetForm.inputEmail')}
</label>
<TextInput
  id={emailInputId}
  aria-label={...} // or remove if label is used
  ...
/>
CommitResetForm.jsx – Form controls remain enabled after success
On line 145–146, the submit button and inputs are disabled when isComplete is true, which is good. However, the comment on line 66 says "Form should remain disabled?" suggesting this was uncertain. The current behavior (keeping it disabled) is correct for a success state, but consider adding explicit focus management—moving focus to the success message so screen reader users see it.

Missing ARIA labels on styled components
Both password inputs in CommitResetForm (lines 127–131, 142–147) use the password input type, which is good, but there's no explicit aria-describedby linking the help text (e.g., "Both fields must match") to the inputs. This leaves screen reader users without immediate feedback about validation rules.

HTML Form & Input Validation
HTML5 pattern validation may be too strict for some locales
CommitResetForm uses an HTML5 pattern=".{8,}" (lines 120, 141). This works for most locales, but some combining diacritics or multi-byte characters may behave unexpectedly. However, this follows the Panoptes backend rule, so it's defensible. Just be aware if you get reports from users with complex scripts.

No client-side password confirmation validation before form submission
In CommitResetForm.onSubmit(), the passwords are checked after the form fires the onSubmit event (line 87). The <Form> component should prevent invalid submissions due to the pattern and required attributes, but the "passwords must match" check is entirely custom and only runs in JavaScript. Consider adding a <FormField> wrapper or onBlur validation to provide immediate feedback.

Error Handling & UX
API error response structure assumption may fail silently
In doCommitPasswordReset.js (line 38–40), the code assumes the error has a .body.errors[0].message structure:

JavaScript
if (_error.req && error.status >= 400 && error.status <= 599) {
  const errorMessage = _error.body?.errors?.[0]?.message || 'Unknown API error'
}
But if _error.body.errors is an empty array or missing, users see "Unknown API error", which is not helpful. Panoptes may return different error shapes. Consider logging the full error object or building a fallback message that includes the status code.

doRequestPasswordReset.js inconsistent error handling pattern
doRequestPasswordReset() (lines 26–32) returns the raw error object, but it doesn't inspect the shape like doCommitPasswordReset() does. If Panoptes returns a response-like object instead of an Error, the component won't have a .message property. This could cause the error message display to fail silently. Keep error handling consistent across both helpers.

HTTP 429 (rate limit) special-cased only in one form
The "too many requests" error (line 40 in RequestResetForm.jsx) is handled with a special translation key, but CommitResetForm doesn't check for error.status === 429. If users hit the Panoptes rate limit after requesting a reset, they'll see a generic error instead of the friendly message. Consider moving this check to the helper function.

Typos & Minor Issues
Typo in comment in ResetPassword.jsx
Line 6 says "We need a wrapper around the <ResetPassword>" but it's actually in ResetPasswordContainer.js. The comment is a bit unclear about why the wrapper is needed. Suggest: "We need a container to check login status via PanoptesAuthContext, which is only available on the client side."

Potential issue: useId() in CommitResetForm
Lines 64–65 call useId() which generates a fresh ID on each render. This is fine, but ensure that Grommet's FormField doesn't rely on stable IDs for validation message association. If it does, moving useId() calls outside the component or memoizing might be needed (though unlikely to be an issue in practice).

Testing & Documentation
Unit tests look solid, checking all three modes.
Storybook stories are helpful for visual testing.
README is thorough and notes the TODO for Part 2 (redirect after reset).
Testing instructions are clear and pragmatic (noting that local Storybook can't submit due to CORS, production-only testing needed).
Merge Readiness Assessment
Status: Low risk, but address the form accessibility and error handling issues above before merge.

Non-blocking observations:

PR is behind main, but rebaseable. No conflicts noted.
No CI failures visible in the data provided (check GitHub Actions for actual results).
The "behind main" state is expected given the branch was created recently.
Blocking concerns to address:

The accessibility improvements (labels, ARIA descriptions) will improve compliance with WCAG standards.
Error handling inconsistency between the two reset helper functions should be normalized.
The HTTP 429 special case should be moved to the helper or applied to both forms.
```

</details>

### Dev Notes

The feedback provided by this AI review was more useful than [previous attempts,](https://github.com/zooniverse/front-end-monorepo/pull/7514#issuecomment-5026164364) but still short of being a game-changer.

There's one very pertinent piece of feedback I can use; several observations/feedback items that are valid but will be/must be ignored since Copilot can't see the bigger picture; and _only_ a small number of completely wrong assessments.[^1]

It's fine, I guess.

[^1]: this is like saying "my iceberg lettuce has _only_ some small traces of cyclospora."

packages/app-root/src/app/reset-password/page.js
```diff
@@ -0,0 +1,22 @@
...
+
+export const metadata = {
+  title: 'Reset Password',
+  description: ''
+}
+
+export default async function ResetPasswordPage (props) {
+
+  // Check if there's a ?reset_password_token=1234567890 in the URL's query
+  // string. If yes, it means the user clicked a link in an "Unsubscribe Me Pls"
```
goplayoutside3:
```suggestion
  // string. If yes, it means the user clicked a link in an "Reset password instructions"
```

packages/lib-content/src/screens/ResetPassword/components/ResetPasswordHeading/ResetPasswordHeading.jsx
goplayoutside3, line 1: Suggestion to rename this file and component to ResetPasswordHead _ing_. Easy to confuse the purpose of it as building a `<header>` but it is actually for the `<h1>` heading text content.

packages/lib-content/src/screens/ResetPassword/components/ResetPasswordHeader/ResetPasswordHeader.jsx
```diff
@@ -0,0 +1,26 @@
+import { Heading } from 'grommet'
+import styled from 'styled-components'
+import { useTranslation } from '@translations/i18n'
+
+const DecoDivider = styled('div')`
+  width: 100%;
+  height: 2px;
+  background: linear-gradient(90deg, rgba(255, 255, 255, 0) 0%, #A6A7A9 49.62%, rgba(255, 255, 255, 0) 100%);
```
goplayoutside3:
```suggestion
  background: linear-gradient(90deg, #ffffff 0%, #A6A7A9 50%, #ffffff 100%);
```

Sorry to be annoying about readable CSS 😆  but suggestion to make this change for human dev readability when copy-pasting something from Figma.

packages/lib-content/src/screens/ResetPassword/components/CommitResetForm/CommitResetForm.jsx
```diff
@@ -0,0 +1,175 @@
...
+`
+
+const HelpInfo = styled('span')`
+  ${props => props.theme.dark
+    ? css`color: ${props.theme.global.colors['dark-4']}`
+    : css`color: ${props.theme.global.colors['light-2']}`
+  }
+  font-weight: 300;
+  font-size: 0.6666666667em;
+  line-height: 1em;
```
goplayoutside3:
```suggestion
const HelpInfo = styled(Text)`
  ${props => props.theme.dark
    ? css`color: ${props.theme.global.colors['dark-4']}`
    : css`color: ${props.theme.global.colors['light-2']}`
  }
  font-size: 0.6666666667rem;
```

To match the patterns of most recently build FEM pages, please use the available Grommet Text component for a span. No need to touch the line-height or font weight. We've also been defining font size in `rem`. I noticed you've used `em` a lot for gaps, margins, etc which is generally fine, but I do want to catch consistency in our font units (`rem`).

Additionally, instead of using a `styled` component you could write this HelpInfo as the following Grommet component, although I think it's just preference in code reading 💁‍♀️ 
```
import { Text } form 'grommet'

 <Text
  color={{ light: 'light-2', dark: 'dark-4' }}
  size='0.67rem'
>
  {t('')}
</Text>
```

packages/lib-content/src/screens/ResetPassword/helpers/isNewPasswordValid.js
```diff
@@ -0,0 +1,27 @@
...
+
+  // Minimum length is checked via <input pattern=".{8,}" />
+
+  // Users shouldn't see this message since the browser should check for minimum length, and prevent a form submit.
+  if (!password || !confirmation) { return new Error(t('ResetPassword.CommitResetForm.status.errorInvalidInput')) }
+
+  // This is the most likely message a user would see.
+  if (!(password === confirmation)) { return new Error(t('ResetPassword.CommitResetForm.status.errorPasswordsDoNotMatch')) }
+
+  return 0
```
goplayoutside3: This suggestion if definitely non-blocking, but my opinion is that because the browser handles min password length, the only real reason the `setInputError` state updater in CommitResetForm is called is when the two password fields don't match. To simplify things, this function `isNewPasswordValid` could be single purpose "check if two strings match, return a boolean". Then you wouldn't need to pass `t`, and in CommitResetForm when `isNewPasswordValid` returns true, submit the API request, when `isNewPasswordValid` returns false, `setInputError(t('ResetPassword.CommitResetForm.status.errorPasswordsDoNotMatch'))`.

The returning `0` vs returning a translation string pattern here is overcomplicated for the actual use case. But it _does_ make sense to use the return `0` or return an Error object when making an actual API request in the other helper functions.

packages/lib-content/src/screens/ResetPassword/README.md
```diff
@@ -0,0 +1,47 @@
...
+- **Mode 2: Request Reset**
+  - Entry point: user clicks "Forgot Password" on a Login Form, and is directed to this Reset Password Page. (Or they access the URL directly.)
+  - If a user accesses the page URL _with no reset password token,_ (i.e. just https://www.zooniverse.org/reset-password), they will be shown a form that asks for their email address.
+  - The form will perform basic HTML checks for email validity.
+  - When the user submits a valid email address, the form transitions from an initial "ready" state to the following states:
+    - "busy" state: request is being sent to Panoptes. "Loader" (busy indicator) is shown. Form is disabled.
+    - "success" state: request has successfully been processed. Success message is show. Form remains disabled.
+    - "error" state: request has failed (API error). Error message is shown. Form is re-enabled.
+- **Mode 3: Commit Reset**
+  - Entry point: user receives a password reset email, and clicks on the link in that email. They're taken to this Reset Password Page _with a token._
```
goplayoutside3, line 21: When I received an email from `no-reply@zooniverse.org` with reset password instructions, the link is to `https://panoptes.zooniverse.org/users/password/edit?reset_password_token=xxxxxxxxxxxxxx`.

When I click, it's redirected to `/reset-password?reset_password_token=xxxxxxxxx`. Do you know if that redirect is defined in the `static` repo? Somewhere else?

packages/lib-content/src/screens/ResetPassword/helpers/doCommitPasswordReset.js
```diff
@@ -0,0 +1,49 @@
...
+  try {
+    await auth.resetPassword({ password, confirmation, token })
+    return 0
+  } catch (error) {
+    let _error = error
+    console.error(_error)
+
+    // For some reason, auth.resetPassword() sometimes throws the entire
+    // XMLHttpRequest's Response object instead of an Error object. If this
+    // happens, we need to adapt it.
```
goplayoutside3, line 40: 🤔  Interesting.. out of curiosity are there any notes in PJC's definition of `.resetPassword()` about this?

packages/lib-content/src/screens/ResetPassword/components/RequestResetForm/RequestResetForm.jsx
```diff
@@ -0,0 +1,100 @@
...
+        <InputBoxes>
+        
+          <TextInput
+            aria-label={t('ResetPassword.RequestResetForm.inputEmail')}
+            disabled={isBusy || isComplete}
+            name='email'
+            ref={inputEmail}
+            required
+            type='email'
+          />
```
goplayoutside3, line 96: I saw in your notes about Copilot feedback that it mentioned this input as missing a `<label>`. Agreed that while you've got an `aria-label` defined, you can also try wrapping `t('ResetPassword.RequestResetForm.body')` in a Grommet Text component (which is a `<span>`) and styling it with the correct font size (`1rem`). Then wrap that with a `<label>`. It would look the same as your current UI, but be closer to semantic HTML expected of a text `<input>`.

goplayoutside3 (approved):
Thank you for the helpful description of user actions to manually test! All is working as expected, and I was able to successfully reset the password of a test account. See my non-blocking comments for some additional thoughts, and happy to discuss more as needed 👍

eatyourgreens:
>✖️ I'm dubious about changing focus management to the success message. For one, the success message is in an aria-role=status region explicitly to announce changes. For another, programmatically changing focus ain't always a good idea. [BBC accessibility guide: "Focus must only change when activated by the user. ... This guideline is particularly important when creating forms."](https://www.bbc.co.uk/accessibility/forproducts/guides/mobile/changing-focus/)

Sorry to interrupt here, but the BBC accessibility guidance you've linked to isn't relevant to form submission success/error messages (it's for form inputs that change focus when you change their value, which fails [WCAG SC 3.2.2: On Input](https://www.w3.org/WAI/WCAG21/Understanding/on-input).) The LLM is correct here about managing focus after submitting a form. On form submission, the user's attention (whether sighted or visually-impaired) should be directed to the success/error message, so that they know immediately whether they need to try submitting the form again. Definitely check that focus is in a sensible place after a successful form submission, both for sighted keyboard users, visually-impaired screen reader and screen magnifier users, and that both those groups know whether their form submission worked or not. It's totally ok to manage focus programmatically from an `onSubmit` handler, since that is changing focus in response to a user action (submitting the form.)

From that same BBC accessibility guidance.
>Focus must only change when activated by the user.

packages/lib-content/src/screens/ResetPassword/README.md
eatyourgreens: There's a redirect defined in the Rails app, here:
https://github.com/zooniverse/panoptes/blob/e38d24ed1db45a56e65b457e87607da8db98e2e6/config/initializers/redirect_urls.rb#L6-L8

eatyourgreens: Oh, and here in the passwords controller too.
https://github.com/zooniverse/panoptes/blob/e38d24ed1db45a56e65b457e87607da8db98e2e6/app/controllers/passwords_controller.rb#L16-L22

packages/lib-content/src/screens/ResetPassword/helpers/doCommitPasswordReset.js
shaunanoordin: There's no documentation on PJC about _why_ this is the case, but I do know _how_ this is the case.

Looking at PJC's [resetPassword()](https://github.com/zooniverse/panoptes-javascript-client/blob/721b696dccecf86a35614244ab29d19d1a62c3b7/lib/auth.js#L269) function in auth.js, this is the ONLY function that just _returns_ `makeCredentialHTTPRequest()` like so...

```
async function resetPassword () {
  ...
  return makeCredentialHTTPRequest(...)
}
```

...which makes error handling the caller's problem.

For comparison, most other functions in auth.js wrap it in a try-catch:

```
async function _getBearerToken () {
  try {
    const response = await makeCredentialHTTPRequest(...)
    ...
  } else catch (err) {
    throw new Error(err) (OR) throw apiClient.handleError(err)
  }
}
```

...or is a wildcard loose cannon who plays by no rules:

```
async function changePassword () {  // As in, signed in user wants to change their password via the profile page
  ...
  await makeCredentialHTTPRequest(...);
  // Resetting the password changes the underlying cookie session data
  // need to sign out and back in to refresh
  await this.signOut();
  return this.signIn(...);
  ...
  // No try-catch!! I'm not sure what happens if the makeCredentialHTTPRequest() fails? Presumably 
}
```

shaunanoordin:
## PR Update

Thank you for your feedback, everyone! 👍 

Here are the feedback items that I plan to address:

- [x] Fix typo in reset-password/page.js's docs (✔️ merged suggestion)
- [x] Rename ResetPasswordHeader.jsx -> ResetPasswordHeading.jsx
- [x] Code: improve CSS readability (rgb(255,255,255) -> #ffffff) (✔️ merged suggestion)
- [x] HelpInfo: replaced styled span with Grommet Text.
- [x] Simplify isNewPasswordValid.js to just check for matching strings.
  - [Update:](https://github.com/zooniverse/front-end-monorepo/pull/7529/commits/8326bc9f96500b539d508ff6901cd716907ca10a) submit checks tweaked as well.
- [x] Wrap TextInput in RequestResetForm in visually hidden label.
- [x] Double-confirm success message is announced correctly after submit. If not, update page focus on success message.

shaunanoordin:
## PR Update

After addressing the feedback, this PR is now failing with an error in `app-root`: `reset-password/page.js doesn't have a root layout. To fix this error, make sure every page has a root layout.` ❓ 🤔 

I thought rebasing the stack would help, but no luck with that. For the record, tests were passing prior to the rebase up at commit [9d08dd7](https://github.com/zooniverse/front-end-monorepo/pull/7529/commits/9d08dd7b8b037b50c0141d8de96ab52f0cc2cdda), which was right before the feedback changes.

eatyourgreens:
@shaunanoordin you need to move `reset-password` to `[locale]/reset-password`.

shaunanoordin:
Thanks Jim! 👍 ⭐

eatyourgreens:
You're welcome (and working very late, when you should be playing Pokopia.)

An underwater scene from Pokémon Pokopia, on a large LED TV.

Issue #7472, opened 2026-06-02
Errors in ETC and completion percentages for project workflows


eatyourgreens:
## Package
- app-project

## Describe the bug
Estimated Time to Completion is showing as 0 for workflows ~~with small numbers of classifications and retired subjects~~ that are less than 3 days old. 

Workflows that have been set to use classification counts for completion still show the fraction of retired subjects, not the percentage of required classifications.

## To Reproduce
This workflow is 2 days old, has 35 classifications (out of 730 required) and one retired subject, but shows ETC as 0 (and 0% progress.)

Assuming a constant rate of 34 classifications per day, ETC is ~21 days. Progress is currently ~5% complete.

https://frontend.preview.zooniverse.org/projects/meteonetwork/redata/stats?start_date=2026-05-27&end_date=2026-06-02&type=count&workflow_id=32059

<img width=400 alt="Mantova 1901 with ETC 0 and progress 0% but one retired subject. It has 35 classifications." src="https://github.com/user-attachments/assets/3fa11573-4e76-4f7d-bfef-675abc52e611" />

## Expected behavior
ETC should reflect the number of days required to complete the workflow at the current rate of work. Progress towards completion should be calculated from the classification count (for those workflows that use classifications to calculate completion.)

## Additional context

The old statistics page didn't show ETC for new workflows, so it might be best to hide ETC when it can't be calculated.
https://pr-7406.pfe-preview.zooniverse.org/projects/meteonetwork/redata/stats

<img width=400 alt="Old completion statistics for two new ReData workflows, which don't have an ETC calculated yet." src="https://github.com/user-attachments/assets/ae581a16-9f28-40cb-adb7-8cf77a62f626" />

eatyourgreens:
Related to this, here's a workflow that is 83% complete (4928/5973 classifications) but the percentage completion incorrectly shows the fraction of retired subjects (59%.)

https://frontend.preview.zooniverse.org/projects/md68135/notes-from-nature-southeastern-us-biodiversity/stats?start_date=2019-07-23&end_date=2026-06-02&type=count

<img width=400 alt="Two NfN workflows, close to completion, but showing completion percentages based on the number of retired subjects instead of the number of classifications." src="https://github.com/user-attachments/assets/5bc64f8e-8789-42cb-9955-bbfd536f432c" />

You can see the correct completion percentages (83% and 89%) on the old statistics page. 

[...]

PR #7533, opened 2026-08-04
app-project: Adjust ETC for recently launched workflows and completeness for those configed to classification count stats instead of retirement count

goplayoutside3:
## Package
app-project

https://github.com/zooniverse/front-end-monorepo/issues/7472

## Describe your changes
- Adjust when to show ETC per workflow (only need 1 day bucket of stats instead of 2)
- Don't show ETC if it's `null` (happens if workflow is just launched today or workflow is complete)
- Remove unnecessary calc of eras response `total_count`
- Override `workflow.completeness` when its `stats_completeness_type` is set to "classification". The default `stats_completeness_type` is by retirement count, so we have to override it because toggling in the project builder does not recalc `workflow.completeness` on the backend. This prop is only for the project stats page UI.

## How to Review

Review the discussion in https://github.com/zooniverse/front-end-monorepo/issues/7472

Compare a project's stats page to its current page, PFE's old stats page, and running app-project locally:
1. Notes From Nature - Southeastern U.S. Biodiversity (wants to show stats by classification count, not the default by retirement count)
    i. https://pr-7406.pfe-preview.zooniverse.org/projects/md68135/notes-from-nature-southeastern-us-biodiversity/stats
    ii. https://localhost.zooniverse.org:3000/projects/md68135/notes-from-nature-southeastern-us-biodiversity/stats?env=production
    iii. https://www.zooniverse.org/projects/md68135/notes-from-nature-southeastern-us-biodiversity/stats
2. [E-I-E-I-O No! (shows stats by the default UI - retirement count)
    i. https://pr-7406.pfe-preview.zooniverse.org/projects/blerdapprentice/e-i-e-i-o-no/stats
    ii. https://localhost.zooniverse.org:3000/projects/blerdapprentice/e-i-e-i-o-no/stats?env=production
    iii. https://www.zooniverse.org/projects/blerdapprentice/e-i-e-i-o-no/stats
3. Nature Spam Filter (a finished project)
    i. https://pr-7406.pfe-preview.zooniverse.org/projects/dverissimo/nature-spam-filter/stats
    ii. https://localhost.zooniverse.org:3000/projects/dverissimo/nature-spam-filter/stats?env=production
    iii. https://www.zooniverse.org/projects/dverissimo/nature-spam-filter/stats

### General

### General UX

### Bug Fix
- [ ] The bug is fixed

### DIFF
--- a/packages/app-project/src/helpers/getProjectStatsPageProps/fetchWorkflowStatsHelper.js
+++ b/packages/app-project/src/helpers/getProjectStatsPageProps/fetchWorkflowStatsHelper.js
@@ -120,6 +120,23 @@ function calcDaysToCompletion(erasData, workflow) {
   return numDays
 }
 
+function calcCompleteness(workflow) {
+  if (workflow.completeness === 1) return 1
+
+  let completeness = workflow.completeness // the value returned from panoptes API
+
+  const totalCount = workflow.subjects_count * workflow.retirement.options.count
+
+  // This config prop only applies to stats page UI. Configuring stats_completion_type in
+  // the project builder as "Completeness statistic" does not recalc workflow.completeness on the backend.
+  // The default stats_completeness_type is by retirement count, so we have to override here.
+  if (workflow.configuration.stats_completeness_type === 'classification') {
+    completeness = workflow.classifications_count / totalCount
+  }
+
+  return completeness
+}
+
 async function fetchWorkflowStatsHelper(language = 'en', workflowIDs, env, workflowOrder) {
   // Fetch workflow data from Panoptes API
   const workflows = await fetchWorkflowData(workflowIDs, env)
@@ -151,8 +168,10 @@ async function fetchWorkflowStatsHelper(language = 'en', workflowIDs, env, workf
 
     if (workflowERASData?.data) {
       const workflowETC = calcDaysToCompletion(workflowERASData, workflow) ?? null
+      const workflowCompleteness = calcCompleteness(workflow)
       const workflowWithETC = {
         ...workflow,
+        completeness: workflowCompleteness,
         etc: workflowETC
       }
       return workflowWithETC
--- a/packages/app-project/src/helpers/getProjectStatsPageProps/fetchWorkflowStatsHelper.js
+++ b/packages/app-project/src/helpers/getProjectStatsPageProps/fetchWorkflowStatsHelper.js
@@ -61,15 +61,26 @@ async function fetchWorkflowClassificationStats(workflowID, env) {
 
   const today = new Date()
 
-  // Subtract one day because today's stats bucket is still accumulating
-  const yesterday = new Date(today)
-  yesterday.setDate(yesterday.getDate() - 1)
-  const endDate = yesterday.toISOString().substring(0, 10)
+  // Subtract one day because today's stats bucket is still accumulating (always UTC)
+  const yesterdayUTC = new Date(
+    Date.UTC(
+      today.getUTCFullYear(),
+      today.getUTCMonth(),
+      today.getUTCDate() - 1
+    )
+  )
 
-  // Get date of two weeks ago
-  const twoWeeksAgo = new Date(today)
-  twoWeeksAgo.setDate(twoWeeksAgo.getDate() - 14)
-  const startDate = twoWeeksAgo.toISOString().substring(0, 10)
+  const endDate = yesterdayUTC.toISOString().substring(0, 10)
+
+  // Get date of two weeks ago (always UTC)
+  const twoWeeksAgoUTC = new Date(
+    Date.UTC(
+      today.getUTCFullYear(),
+      today.getUTCMonth(),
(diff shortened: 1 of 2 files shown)


packages/app-project/src/helpers/getProjectStatsPageProps/fetchWorkflowStatsHelper.js
```diff
@@ -88,31 +99,44 @@ async function fetchWorkflowClassificationStats(workflowID, env) {
  * Returns estimate number of days based on ERAS query excluding the current day because it's not over yet.
  */
 function calcDaysToCompletion(erasData, workflow) {
-  if (workflow.completeness === 1) return 0
+  if (workflow.completeness === 1) return null
```
goplayoutside3, line 102: Return `null` instead of `0`. No need to show ETC when a workflow is complete.

packages/app-project/src/helpers/getProjectStatsPageProps/fetchWorkflowStatsHelper.js
```diff
@@ -88,31 +99,44 @@ async function fetchWorkflowClassificationStats(workflowID, env) {
...
-    // Sum the classifications per day the workflow has been live (up to 14 days)
-    const rate = classificationCountsArray.reduce((a, b) => a + b)
+  const dataLength = erasData.data.length // number of days the workflow has stats for
 
+  if (dataLength > 0) {
     // Estimate number of days needed to achieve the classification counts to retire remaining subjects
     numDays = Math.max(
       0,
-      Math.ceil((dataLength * (totalCount - workflow.classifications_count)) / rate)
+      Math.ceil((dataLength * (totalCount - workflow.classifications_count)) / erasData.total_count)
```
goplayoutside3, line 116: No need to calculate `total_count` via the older reduce method, which was confusingly named `rate` in PFE's original stats page (and mentioned in this PR's linked Issue).

But this also makes me wonder why this particular calculation still uses `workflow.classification_count`. Would it make sense to change this line to 
`Math.ceil((dataLength * (totalCount - erasData.total_count) / erasData.total_count)` ?

packages/app-project/src/helpers/getProjectStatsPageProps/fetchWorkflowStatsHelper.js
```diff
@@ -88,31 +99,44 @@ async function fetchWorkflowClassificationStats(workflowID, env) {
...
+  let completeness = workflow.completeness // the value returned from panoptes API
+
+  const totalCount = workflow.subjects_count * workflow.retirement.options.count
+
+  // This config prop only applies to stats page UI. Configuring stats_completion_type in
+  // the project builder as "Completeness statistic" does not recalc workflow.completeness on the backend.
+  // The default stats_completeness_type is by retirement count, so we have to override here.
+  if (workflow.configuration.stats_completeness_type === 'classification') {
+    completeness = workflow.classifications_count / totalCount
+  }
```
goplayoutside3, line 135: Fixes https://zooniverse.slack.com/archives/C0GR39WKB/p1784663224098389

packages/app-project/src/helpers/getProjectStatsPageProps/fetchWorkflowStatsHelper.js
sktw: The `erasData` in this function is just for the last 14 days, so it wouldn't make sense to replace `workflow.classifications_count` with `erasData.total_count`.

More generally, there is an important difference between the workflow and ERAS classification counts. The workflow classification count only includes classifications from the *current* subject set(s) attached to the workflow. ERAS on the other hand includes contributions from *every* subject set that has ever been attached. This means, for example, that if a project removes a completed subject set and attaches a new one, the ETC calculation for the new subject set will initially use the rate calculated from the old one.

goplayoutside3: Good point! I think the differences between the workflow and ERAS classifications counts is what my brain was flagging here. Indeed if a subject set is changed or replaced while a workflow is still active, the ETC calculation above is influenced by ERAS counting every classification over that period of 14 days regardless of attached subject set, and it also counts duplicate classifications, or classifications on retired subjects too (if a volunteer happens to see one in their classifier queue).

I don't know if there's a meaningful change we can make to ETC in this PR to increase accuracy, but my goal is to doc all of the nuances in the code of the FEM version of this stats page.

sktw: There's a trade-off here that I don't think can be avoided. Averaging the classification rate over 14 days has the advantage of smoothing out random fluctuations which would cause the ETC to jump around if only the previous day was used. However, if the classification rate genuinely changes, for example if a new, more challenging subject set is added, the ETC won't adjust completely for another 14 days.

kieftrav (approved):
Changes are reasonable / make sense and everything works as described. Good work!

